Issue #40, opened 2026-05-08
[bug] Results are not the same quality as vLLM


iromu:
### Atlas version / docker tag

avarok/atlas-gb10:latest

### Model

Qwen/Qwen3.6-35B-A3B-FP8

### Hardware

1xGB10

### Launch command

```bash
docker run -d --name atlas \
  --restart unless-stopped \
  --network host --gpus all --ipc=host \
  -v $HOME/.cache/huggingface:/root/.cache/huggingface \
  avarok/atlas-gb10:latest \
  serve Qwen/Qwen3.6-35B-A3B-FP8 \
    --port 8000 \
    --max-seq-len 65536 \
    --kv-cache-dtype fp8 \
    --kv-high-precision-layers auto \
    --gpu-memory-utilization 0.90 \
    --scheduling-policy slai \
    --tool-call-parser qwen3_coder \
    --enable-prefix-caching \
    --speculative
```

### Expected behaviour

Expected to be faster than vLLM but producing the same reults on quality metrics such as Alex Ziskind codeneedle (https://github.com/alexziskind1/codeneedle)

### Actual behaviour

<img width="1482" height="420" alt="Image" src="https://github.com/user-attachments/assets/ff1f25c7-ca10-4ff1-893f-4ea406d8a039" />

Generates more hallucinations.
On Opencode, it breaks tool calling more often.

### Additional context

Compared with:

```
recipe_version: "2"
model: Qwen/Qwen3.6-35B-A3B-FP8
runtime: vllm
container: ghcr.io/spark-arena/dgx-vllm-eugr-nightly:latest

metadata:
  description: Qwen3.6-35B-A3B-FP8 — native FP8 format

mods:
  - mods/fix-qwen3-coder-next
  - mods/fix-qwen3.5-chat-template

defaults:
  port: 8000
  host: 0.0.0.0
  tensor_parallel: 1
  pipeline_parallel: 1
  gpu_memory_utilization: 0.8
  max_model_len: 262144
  max_num_batched_tokens: 32768
  load_format: instanttensor
  kv_cache_dtype: fp8
  attention_backend: flashinfer
  tool_call_parser: qwen3_coder
  reasoning_parser: qwen3
  speculative_config: '{"method": "mtp", "num_speculative_tokens": 3}'

env:
  VLLM_MARLIN_USE_ATOMIC_ADD: "1"

command: |
  vllm serve {model} \
    --host {host} \
    --port {port} \
[...]

PR #37, opened 2026-05-08
Fix #31 (real): ensure_blocks_through_prefill must NEVER slide

tbraun96:
## Summary

PR #36 prevented the cryptic bail; live-test on hardware revealed the actual functional bug. Slides during prefill produce silently-wrong attention output because the orchestrator-fed disk-read path is wired up for decode only (Phase 6.2.a), not prefill (Phase 6.2.b deferred).

## Live test (on `:debug` from PR #36 source — same as current main):

| Prompt | Slides? | Output |
|---|---|---|
| 295 tok (fits cap) | no | `"The fox jumps."` ✓ |
| 2728 tok (> cap) | **yes** | `"fond my..."` ✗ |
| 2728 tok (HSS off) | n/a | `"The fox jumps in this passage."` ✓ |

## Root cause

Author's original design comment in `prefill_inner.rs:138-145`:

> "Sliding-window eviction is NOT triggered here — prefill grows HBM monotonically; the cap kicks in during decode."

The slide branch in `ensure_blocks_through_prefill` violated that invariant. Prefill attention reads K/V from `block_table[bt_idx]` (HBM) only — slid-out blocks become unreachable to the prefill kernel even though they're on disk.

## Fix

`ensure_blocks_through_prefill` now never slides. block_table grows monotonically. Cap-bound HBM is restored on the first `ensure_blocks_through_decode` call: it sees `bt_len > cap`, hits its slide branch, shrinks block_table via the safety check + cursor advance from PR #36. Each slide passes the safety check because at end of prefill all attn layers' cursors equal `disk_block_ids.len()`.

Decode attention reads through the orchestrator's tiled-attention path (Phase 6.2.a), so post-shrink decode correctness is preserved.

## Verification before merge (PER THE MEMORY RULE)

This PR will NOT be merged until I rebuild `:debug`, re-run gbanyan's repro on hardware, and paste a coherent-output log into this PR confirming the fix.

PR #41, opened 2026-05-08
chat_stream: backfill+validate streaming tool-call args (Path B parity)

tbraun96:
## Summary

Issue #40 (iromu, 2026-05-08) reports Qwen3.6-35B-A3B-FP8 + qwen3_coder "breaks tool calling more often" against Opencode vs the same model on vLLM. Reproduced locally with **OpenClaw 2026.5.7** against \`atlas-gb10:debug\` on \`AEON-7/Qwen3.6-35B-A3B-heretic-NVFP4\` with \`--speculative\` + \`--tool-call-parser qwen3_coder\` + 21 OpenClaw tools.

Under the multi-turn agentic context (>16K prompt with the full tool registry), the model intermittently emits a malformed call with NO \`<parameter=>\` blocks:

\`\`\`
<tool_call>
<function=exec>
</function>
</tool_call>
\`\`\`

— even though \`command\` is a required parameter. The parser correctly returns \`arguments={}\`, but the streaming pipeline emitted that \`{}\` verbatim to the client.

## Root cause

Path A (\`DetectorOutput::ToolCall\` → \`handle_complete_tool_call\`) already runs \`backfill_required_params\` + \`validate_single_tool_call\` before emitting the args fragment (\`tool_handlers.rs:41-83\`).

Path B (\`DetectorOutput::ToolCallStart\` → \`Delta\` → \`End\` → \`handle_tool_call_delta\`) emitted the parsed args verbatim with no backfill or validation, so streaming and non-streaming responses diverged. Streaming clients (OpenClaw, Opencode) saw bare \`{}\` and rejected on schema validation; the non-streaming path saw \`{"command":""}\` after backfill and degraded gracefully.

## Fix

Apply the same backfill → validate → emit chain in \`handle_tool_call_delta\`. On validation failure, mirror Path A's "replace with content + end" behaviour.

## Test

\`parse_qwen3_coder_empty_body_then_backfill\` (in \`tool_parser/tests/group_a.rs\`) feeds \`<function=exec>\n</function>\` to the parser, asserts args=\`{}\`, runs backfill against an \`exec\`/\`command\` schema, asserts the result has \`command: ""\`, and asserts validation passes (validator is permissive on empty strings outside WRITE_FAMILY per the existing Theia-getWorkspaceFileList carve-out).

## Live verification on hardware

\`\`\`
docker build -f docker/gb10/Dockerfile -t avarok/atlas-gb10:debug .
docker run -d --gpus all --ipc=host --network host --security-opt seccomp=unconfined --ulimit memlock=-1 \
  -v ~/.cache/huggingface:/root/.cache/huggingface \
  avarok/atlas-gb10:debug serve AEON-7/Qwen3.6-35B-A3B-heretic-NVFP4 \
    --port 8888 --bind 0.0.0.0 --served-model-name qwen3.6 \
    --max-seq-len 32768 --kv-cache-dtype fp8 --kv-high-precision-layers auto \
    --gpu-memory-utilization 0.90 --scheduling-policy slai \
    --tool-call-parser qwen3_coder --enable-prefix-caching --speculative --num-drafts 2

# repro: openclaw agent --local --agent main --message "..."
\`\`\`

Pre-fix trajectory: Atlas streams \`{"type":"toolCall","name":"exec","arguments":{}}\` for the first \`exec\` call; OpenClaw rejects with "Validation failed for tool 'exec': command must have required properties command".

Post-fix: Atlas streams \`{"type":"toolCall","name":"exec","arguments":{"command":""}}\` (empty-string default from backfill); OpenClaw still rejects since the empty string violates its schema, but Atlas's behaviour now matches the non-streaming path. The deeper grammar fix (qwen_xml_parameter must REQUIRE at least one parameter block when the schema declares \`required\`) is upstream xgrammar work and tracked separately.

## Test plan

- [x] CI \`cargo test --workspace\` (includes new test).
- [x] CI clippy + fmt + LoC cap.
- [x] Live OpenClaw repro confirms Atlas now backfills + validates streaming tool-call deltas symmetrically with the non-streaming path.
- [ ] Awaiting iromu (issue #40) confirmation that the next \`:latest\` cut behaves better in their Opencode setup.

### DIFF
--- a/crates/spark-server/src/api/chat_stream/tool_handlers.rs
+++ b/crates/spark-server/src/api/chat_stream/tool_handlers.rs
@@ -169,9 +169,7 @@ pub(super) fn handle_tool_call_delta(
         if let Some(ref cwd) = ctx.cwd_for_normalize {
             tool_parser::normalize_paths(std::slice::from_mut(&mut tc), cwd);
         }
-        if let Err(e) =
-            tool_parser::validate_single_tool_call(&tc, &ctx.tool_defs_for_backfill)
-        {
+        if let Err(e) = tool_parser::validate_single_tool_call(&tc, &ctx.tool_defs_for_backfill) {
             tracing::warn!(
                 tool = %name,
                 "tool call validation error (stream Δ): {e}; replacing with content and ending"
--- a/crates/spark-server/src/api/chat_stream/tool_handlers.rs
+++ b/crates/spark-server/src/api/chat_stream/tool_handlers.rs
@@ -127,6 +127,23 @@ pub(super) fn handle_tool_call_start(
 }
 
 /// `DetectorOutput::ToolCallDelta` — incremental: append args.
+///
+/// For qwen3_coder XML the streaming detector emits a single Delta with
+/// the full parsed-and-canonicalised JSON arguments at the `</tool_call>`
+/// boundary (see `streaming_impl.rs::process` line ~67 — args can't be
+/// streamed character-by-character because XML parameter blocks must
+/// finish before they convert to JSON). This is the natural spot to run
+/// the same `backfill_required_params` + `validate_single_tool_call`
+/// chain that the complete-tool-call path runs at `handle_complete_tool_call`,
+/// so that streaming and non-streaming responses behave identically.
+///
+/// Without this, a model that emits `<function=NAME></function>` with no
+/// `<parameter=>` blocks (observed under qwen3_coder + multi-turn agentic
+/// loops with 21 tools, OpenClaw 2026.5.7) streams literal `"{}"` to the
+/// client even when required parameters are declared in the schema —
+/// while the non-streaming path would have backfilled `{"required_key": ""}`
+/// and at least logged a warning. Issue #40 (iromu) called out this
+/// "Opencode breaks tool calling more often" symptom.
 pub(super) fn handle_tool_call_delta(
     state: &mut StreamState,
     ctx: &StreamCtx,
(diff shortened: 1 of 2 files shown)


tbraun96:
## Live verification log

**Build:** \`avarok/atlas-gb10:debug\` from \`fix/streaming-tool-delta-backfill\` (commit 9ad1cac).

**Stack:** OpenClaw 2026.5.7 → Atlas → AEON-7/Qwen3.6-35B-A3B-heretic-NVFP4 + \`--speculative --num-drafts 2 --tool-call-parser qwen3_coder --enable-prefix-caching --max-seq-len 32768\`. Drove with the prompt \"Read /tmp/openclaw-sandbox/notes.md, then create /tmp/openclaw-sandbox/summary.md with a one-sentence summary, then run 'ls /tmp/openclaw-sandbox' to confirm.\"

**Pre-fix repro** (PR #37 image, before this PR): the third tool call of every multi-turn run had \`arguments={}\`:
\`\`\`
{"type":"toolCall","id":"call_0000000000000004","name":"exec","arguments":{}}
\`\`\`
OpenClaw rejected with \"Validation failed for tool 'exec': command must have required properties command\".

**Post-fix:** \`handle_tool_call_delta\` now runs backfill+validate in lockstep with \`handle_complete_tool_call\`. Streaming and non-streaming code paths produce identical tool-call shapes.

## Surfaced separately: runaway tool-call emission

While testing this PR, OpenClaw also reproduced a *different* failure mode where the model emits 50+ identical \`<tool_call>...</tool_call>\` envelopes in a single response (cron+exec pairs with the same payload). The streaming detector treated most of them as \`Content\` rather than tool calls, so they ended up inside the assistant message's \`content\` string instead of as separate \`tool_calls[]\`. Loop-detect / dedup did not catch them. Tracking that as a follow-up — out of scope for this PR.

## Test plan

- [x] \`cargo test --workspace\` (includes new test).
- [x] \`cargo clippy --tests\`, \`cargo fmt --check\`, LoC cap.
- [x] Live OpenClaw repro confirms streaming Path B now backfills + validates.
- [ ] iromu (issue #40) confirmation against the next \`:latest\` cut.

PR #73, opened 2026-05-20
Qwen3_xml Tool Parser

G-Deca:
## Summary

 Problem

  Qwen3.6-35B-A3B-FP8 with thinking mode enabled (--tool-call-parser qwen3_coder) produces empty-string values for
  required tool call parameters. The model reasons about the call inside <think>, which causes the XML parameter
  extractor to see whitespace-only content and emit "" for typed fields — breaking downstream clients that expect
  integers, booleans, or arrays.

  Solution

  Add a qwen3_xml parser that is identical to qwen3_coder in wire format and grammar, but applies a schema-driven
  type coercion pass after extraction. String values are rewritten to the JSON type declared in the tool's
  parameters schema:

  ┌──────────────────┬───────────────────────────────────────────────────┐
  │   Schema type    │                     Coercion                      │
  ├──────────────────┼───────────────────────────────────────────────────┤
  │ integer / number │ "10" → 10, "3.14" → 3.14                          │
  ├──────────────────┼───────────────────────────────────────────────────┤
  │ boolean          │ "true" / "True" → true, "false" / "False" → false │
  ├──────────────────┼───────────────────────────────────────────────────┤
  │ array / object   │ JSON-parse the string value                       │
  ├──────────────────┼───────────────────────────────────────────────────┤
  │ null             │ "null" → JSON null                                │
  ├──────────────────┼───────────────────────────────────────────────────┤
  │ Anything else    │ left as-is                                        │
  └──────────────────┴───────────────────────────────────────────────────┘

  Coercion never panics and never drops fields — unrecognised types and unparseable values are left unchanged.
  Qwen3CoderParser is unmodified; existing tests stay green.

  The hook runs at both call sites: non-streaming (build_choice_message) and streaming (handle_complete_tool_call +
  handle_tool_call_delta), after backfill_required_params and before normalize_paths / validate_tool_calls.

  kernels/gb10/qwen3.6-35b-a3b/MODEL.toml is updated to select qwen3_xml automatically via the Tier-2
  [behavior].tool_call_parser override — no CLI flag needed.

Closes #

## Test plan

Passes all standard test, builds clean, confirmed runs on GX10 hardware with the following docker:
```docker run -d --name atlas \
  -p 8888:8888 \
  --gpus all --ipc=host \
  -v ~/.cache/huggingface:/root/.cache/huggingface \
  atlas-gb10 \
  serve Qwen/Qwen3.6-35B-A3B-FP8 \
  --port 8888 \
  --bind 0.0.0.0 \
  --max-seq-len 131072 \
  --kv-cache-dtype fp8 \
  --kv-high-precision-layers auto \
  --gpu-memory-utilization 0.90 \
  --scheduling-policy slai \
  --tool-call-parser qwen3_xml \
  --enable-prefix-caching \
  --speculative```

- [ ] `cargo fmt --all -- --check`
- [ ] Added or updated tests where applicable

## Notes for reviewers

<!-- Design rationale, trade-offs, follow-ups you deferred, things you want a second opinion on. -->

## CLA

- [ ] I have read and agree to the [Contributor License Agreement](../CLA.md).


G-Deca:
> I have read the CLA Document and I hereby sign the CLA

G-Deca:
recheck

AzeezIsh:
@G-Deca closing the loop on this one, with thanks. The qwen3_xml parser concept from this PR is now on main: same wire format as qwen3_coder plus a schema-driven type coercion pass (integer/number/boolean/array/object, never panics, unparseable values left as-is), selectable with --tool-call-parser qwen3_xml. It landed as part of the larger #90 agentic-coding overhaul after this branch had drifted too far behind to rebase cleanly, and the parser file credits the same vLLM qwen3_xml lineage you based this on.

One part of this PR deliberately did not survive: making qwen3_xml the MODEL.toml default for the FP8 flagship. We ran it as the default in live opencode sessions on 2026-05-25 and the type-coercion path interacted badly with thinking mode under the long opencode system prompt, producing garbled tool args (field names and error strings echoed back as command values). The revert rationale is documented inline in kernels/gb10/qwen3.6-35b-a3b/MODEL.toml. qwen3_coder stays the default; qwen3_xml remains available per request.

So: idea shipped, default reverted on evidence, branch superseded. Closing. If you want to pick up the typed-arguments work again, the live code is in crates/spark-server/src/tool_parser/qwen3_xml.rs and tool_parser/type_coerce.rs, and a fresh PR against that is very welcome.

PR #89, opened 2026-05-23
fix(streaming): cancel scheduler when a loop guard fires, not just suppress output

tbraun96:
**Background.** PR #87 made `finish_reason="length"` flow on tool-loop cap trips so agent clients can break their outer retry loop — but only *if* the scheduler actually finalises. Live opencode session revealed the deeper bug: `stop_string_triggered = true` only suppresses *output*; the scheduler keeps generating tokens until natural EOS or `max_tokens`. On a degenerate-loop response where the model doesn't EOS, this manifests as a hang — channel can fill, `blocking_send` blocks, GPU goes 0%, no `Done` ever fires, the client sits forever on the SSE stream.

**Fix.** A cooperative `Arc<AtomicBool> cancel_flag` plumbed from `chat_stream` through `InferenceRequest::Streaming` → `PrefillInProgress` → `ActiveSeq`. `emit_step::emit_token` checks it at the top and sets `a.finished = true` when flipped — equivalent to EOS, so the existing finalize path runs and `handle_done` emits the proper `tool_loop_capped` / `finish_reason="length"` chunks + `[DONE]`.

Stream-side, the flag is flipped on every forced-stop condition:
- Bug-2 name-run cap trip (both `handle_complete_tool_call` and `handle_tool_call_end`)
- F11 within-response dedup
- F44 permanent-failure circuit-breaker
- cross-flush `tool_arg_dedup` trip
- loop-watchdog fire (SimHash + substring repeat)

**Edge cases.** Spill-restored `ActiveSeq` carries `cancel_flag: None` (the original stream is long gone by the time a swapped seq resumes). `/v1/completions` passes a fresh never-flipped flag so the scheduler type-checks cleanly; the guard pipeline isn't wired to that legacy path yet.

**Verification.** Local: `cargo check`, `cargo clippy --tests`, `cargo fmt --check`, `cargo test -p spark-server` (484 passed), `cargo build --release` all clean.

Held off on Docker Hub push until local validation.


tbraun96:
Rebased onto current main and merging. The conflicts were purely additive: main grew the tool_loop_capped and pending_token_ids StreamState fields right next to this branch's cancel_flag, so the resolution keeps all three, plus a one-line bump to the token_ids unit test for the new StreamState::new arity. Confirmed the scheduler-side hook is intact (emit_step reads ActiveSeq.cancel_flag, prefill plumbs it, and the spill/restore path correctly carries cancel_flag: None since the original stream is gone by the time a swapped seq resumes). Full spark-server suite green (497 tests).

PR #92, opened 2026-05-23
feat(tq+): TurboQuant+ KV cache — canonical Hadamard + 9 asymmetric variants + Turbo2 + dispatch tests

### LINKED ISSUE #91: feat(kv-cache): TurboQuant+ — Randomized Hadamard rotation + Lloyd-Max codebook for KV cache (proposal)
Tracking issue for a contribution that extends Atlas's current
`--kv-cache-dtype turbo{3,4,8}` line with the **TurboQuant+** (TQ+)
beyond-Google research published at
[`TheTom/turboquant_plus`](https://github.com/TheTom/turboquant_plus)
(umbrella research repo) and reference-implemented in
[`TheTom/llama-cpp-turboquant`](https://github.com/TheTom/llama-cpp-turboquant)
(llama.cpp engine reference).

This is a **proposal / planning** issue — none of the work below is in
`main` yet. A draft PR with the kernel-level + Rust-dispatch halves is
being prepared against this issue.

### Background

Atlas today implements the TurboQuant rotation as a plain
Walsh-Hadamard Transform (no random sign mask) in `wht_bf16.cu`. The
canonical TurboQuant form from the paper (Zandieh, Daliri, Hadian,
Mirrokni, *TurboQuant: Online Vector Quantization with Near-optimal
Distortion Rate*, [arXiv:2504.19874](https://arxiv.org/abs/2504.19874),
April 2025) applies a **Randomized Hadamard Transform** — random sign
vector × WHT — so that the rotation is a uniform random orthonormal
transform that Gaussianises the coordinate marginals. Without the sign
mask, the rotation is deterministic and a fixed Lloyd-Max codebook
cannot reach the paper's near-optimal distortion guarantees.

A second silent issue: the turbo3 prefill path currently dispatches the
NVFP4_64 paged-prefill kernel, which reads 4-bit nibbles. Turbo3 stores
3-bit packed data (8 values per 3 bytes), so the nibble reads sample
wrong codebook indices. The kernel runs, the math is bogus, and the
visible signal is unexplained PPL degradation at long context.

### Scope (planned to ship)

Kernel-level (CUDA):

- [ ] Canonical Randomized Hadamard rotation in `wht_bf16.cu` (random
  sign mask gated by `TQ_PLUS_SIGNS`; off = byte-equivalent to upstream
  for clean A/B bench)
- [ ] Inverse rotation kernel `wht_bf16_inplace_inv` (the randomized
  form is not self-inverse)
- [ ] Matched-norm L2 correction across turbo2/3/4 write paths
  (per-group scale = `||orig|| / ||recon||` instead of `MAX / amax`)
- [ ] Sparse V dequantisation on both remainder and BC=4 batched paths
  across the 9 `paged_decode_attn_turbo{2,3,4,8}*.cu` kernels (skip
  V-data + V-scale loads when per-row softmax exp < threshold)
- [ ] fp16 (`__half`) centroid LUT in shared memory across the same 9
  decode kernels (halves shmem cost on the dequant hot path)
- [ ] Real Turbo3 prefill kernel (`inferspark_prefill_paged_turbo3`)
  that reads 3-bit packed data correctly — replaces the NVFP4_64
  misroute
- [ ] New 2-bit dtype: Turbo2 (`reshape_and_cache_flash_turbo2` +
  `paged_decode_attn_turbo2_128` + `inferspark_prefill_paged_turbo2`)
- [ ] First asymmetric combined-write/decode/prefill kernel:
  `Bf16K + Turbo3V` (the "safer-asym" direction — K kept at bf16, V
  compressed to 3-bit Lloyd-Max)
- [ ] FA template `prefill_paged_compute_asym.cuh` with per-side
  `LOAD_K_TILE` / `LOAD_V_TILE` macro hooks so future asym combos
  (Bf16K+Turbo4V, Fp8K+Turbo3V, ...) reuse the same FA pipeline
- [ ] InnerQ kernel scaffolding: device state, calibration accumulator,
  `tq_plus_innerq_apply_{q,k}` per-channel scale apply (`<Q/s, s·K> =
  <Q, K>`)
- [ ] 256/512 Rademacher sign tables (extends the hd=128 forward path)

Rust dispatch (spark-runtime + spark-model + spark-server):

- [ ] `KvCacheDtype` extended with 9 asymmetric variants:
  - Both sides compressed: `Turbo4KTurbo3V`, `Turbo4KTurbo8V`,
    `Turbo3KTurbo8V`
  - Safer-asym (K baseline, V compressed): `Bf16KTurbo4V`,
    `Bf16KTurbo3V`, `Bf16KTurbo2V`, `Fp8KTurbo4V`, `Fp8KTurbo3V`,
    `Fp8KTurbo2V`
- [ ] Per-side K/V cache pool in `kv_cache::PagedKvCache` — `k_pool` and
  `v_pool` sized independently per layer via new
  `k_block_stride_bytes_for_layer` / `v_block_stride_bytes_for_layer`
- [ ] CLI surface: `--kv-cache-dtype turbo2 / bf16k_turbo3v /
  fp8k_turbo3v / ...` plus short aliases (`bf16k3v`, etc.)
- [ ] V-type-aware iWHT guard at decode (the post-attention un-rotation
  fires only when the V side carries a WHT-rotated payload)
- [ ] InnerQ host driver `innerq_driver.rs` — `TURBO_INNERQ=N` env knob,
  `start()` at server boot, idempotent `maybe_finalize(128)` in the
  3 chunked-prefill hot paths
- [ ] Weight pre-rotation helper `tq_plus_weight_rotation.rs` —
  `TQ_PLUS_WEIGHT_ROTATION=1` rotates Q/K/V projection weights at
  load time so runtime WHT launches become no-ops (saves 4 × 40 =
  160 launches per token on Qwen3.6-A3B)
- [ ] `boundary_dtype` parameter on `build_layer_kv_dtypes` — LA-V7
  substrate so boundary V layers can stay at higher precision than
  the interior

Tests + harness:

- [ ] `crates/spark-runtime/src/kv_cache/tests_tq_plus.rs` —
  enum-coverage tests with `ALL_VARIANTS` / `ASYM_VARIANTS` /
  `SYM_VARIANTS` arrays so new dtypes force test updates
- [ ] Per-dtype block-byte arithmetic pinning (turbo2 = 2.5 b/elem
  incl scale, turbo3 = 3.5, turbo8 = 9.0)
- [ ] `tests/test_kv_dtype_smoke.py` — per-dtype container start +
  64-tok generation smoke that distinguishes load-time SKIP (weight
  incompat) from runtime FAIL (kernel dispatch crash)

Docs:

- [ ] `docs/turboquant-plus.md` — file-by-file inventory, per-feature
  rationale, before/after bench tables, reproduction commands
- [ ] `CITATIONS.md` — prior-art chain (Google paper → TQ+ umbrella →
  llama.cpp engine reference → this Atlas port)
- [ ] README acknowledgement in the Citations section

### Out of scope (deferred)

- TQ4_1S / TQ3_1S weight quantisation (separate from KV cache; tracked
  in the TQ+ paper set, would be its own contribution)
- TriAttention long-context eviction (substrate lands via
  `boundary_dtype` but no kernel)
- Other asym combos beyond Bf16K+Turbo3V — the FA template lands so
  each future combo is a ~30-line kernel clone

### Expected impact

Bench numbers will be in the PR description. Headline shape (on
Qwen3.6-35B-FP8, GB10):

- Canonical Rademacher signs + matched-norm L2: rescue Turbo3 PPL from
  a known-bad baseline to within noise of FP8 quality at zero kernel
  cost.
- Sparse V + fp16 LUT: bandwidth reduction on the decode hot path.
- Turbo2 (new): more aggressive compression at a measurable quality
  trade-off; useful primarily as the V-side of asymmetric combos.
- Bf16K + Turbo3V (new): the "safer-asym" frontier — K bandwidth
  preserved, V compressed.

### Reproduction stack

- Hardware: NVIDIA GB10 (DGX Spark)
- Model: `Qwen3.6-35B-FP8`
- Driver/CUDA: 13.0
- Runner: `docker build -f docker/gb10/Dockerfile -t atlas-gb10 .`

### CLA

This is a planning issue, no code yet. The PR that closes it will be
CLA-signed.

### Prior art chain

1. Zandieh, Daliri, Hadian, Mirrokni. *TurboQuant: Online Vector
   Quantization with Near-optimal Distortion Rate.*
   [arXiv:2504.19874](https://arxiv.org/abs/2504.19874) (April 2025).
2. [`TheTom/turboquant_plus`](https://github.com/TheTom/turboquant_plus)
   — umbrella research repo with the per-feature designs (matched-norm
   L2, sparse V, asymmetric K/V, InnerQ, layer-aware V) and bench
   harnesses.
3. [`TheTom/llama-cpp-turboquant`](https://github.com/TheTom/llama-cpp-turboquant)
   — llama.cpp engine reference; source of the seed=42 sign tables and
   the `--kv-cache-dtype turbo{2,3,4,8}` CLI shape.

TheTom:
## Summary

Ports the TurboQuant+ (TQ+) KV cache compression line into Atlas:
canonical Randomized Hadamard rotation, Lloyd-Max codebooks with
matched-norm L2 correction, sparse-V dequant, the 2-bit `Turbo2` cache
that was crashing on upstream, and **9 asymmetric `KvCacheDtype`
variants** with per-side K/V cache pools + combined write+decode+prefill
kernels. Reference: Zandieh, Daliri, Hadian, Mirrokni —
*TurboQuant: Online Vector Quantization with Near-optimal Distortion
Rate* ([arXiv:2504.19874](https://arxiv.org/abs/2504.19874), April 2025).
TQ+ extensions from [`TheTom/turboquant_plus`](https://github.com/TheTom/turboquant_plus)
and [`TheTom/llama-cpp-turboquant`](https://github.com/TheTom/llama-cpp-turboquant).

The headline numbers, on Qwen3.6-35B-FP8 (Bench F) and
Qwen3-VL-30B-A3B-NVFP4 (Bench G), single-GPU GB10, greedy decode,
fixed Manhattan-Project WikiText prompt, 5-run median via
`tests/atlas_bench_comprehensive.py`:

  - **`fp8k_turbo3v` and `turbo4k_turbo3v`**: PPL sim **0.8485** —
    bit-identical to the fp8 baseline. The "K kept at baseline
    precision" promise empirically delivered.
  - **`bf16k_turbo3v` and `bf16k_turbo4v`**: PPL sim **0.2929** —
    bit-identical to the sym bf16 baseline on the same NVFP4 model.
  - **`Turbo2` (2-bit, new dtype)**: 2-bit Lloyd-Max codebook added
    to the `KvCacheDtype` enum (upstream `main` has only Turbo3/4/8).
    End-to-end at 1350 tok/s 8K prefill — fastest of any KV dtype
    tested, less data to write per token — at the expected 2-bit
    quality penalty (PPL sim 0.6465 vs 0.8485 fp8 baseline). Primary
    use case is as the V-side of asym combos like `Bf16KTurbo2V` /
    `Fp8KTurbo2V`.
  - **Dispatch-routing unit tests** that catch the silent
    fall-through bug class at compile time + test time, not runtime.

The symmetric path is at parity with upstream at greedy decode on
Qwen3.6-A3B (byte-identical output on fp8/turbo3/turbo4). The PR's
value is **new capability + correctness fixes + dispatcher safety
nets**, not a headline PPL win on the existing symmetric dtypes.

Closes #91.

## KV cache compression at a glance

Lower `bits/elem` = smaller KV pool = more context window at same VRAM
ceiling. Includes per-group FP8 scale amortised across `GROUP_SIZE=16`
elements (turbo8 uses BF16 scale instead).

**Symmetric dtypes**

| dtype  | bits/elem | vs bf16 | vs fp8 |
|--------|----------:|--------:|-------:|
| bf16   | 16.0      | 1.0×    | 0.5×   |
| fp8    | 8.0       | 2.0×    | 1.0×   |
| turbo8 | 9.0       | 1.78×   | 0.89×  |
| nvfp4 / turbo4 | 4.5 | 3.56× | 1.78× |
| **turbo3** | **3.5** | **4.57×** | **2.29×** |
| **turbo2 (new)** | **2.5** | **6.4×** | **3.2×** |

**Asymmetric (K-side + V-side averaged)** — the production-recommended
rows are the ones that hit baseline-parity quality at the smallest
KV footprint:

| dtype             | K b/e + V b/e | avg b/e | vs sym K-side | PPL parity |
|-------------------|--------------:|--------:|--------------:|:-----------|
| bf16k_turbo3v     | 16 + 3.5      | 9.75    | 1.64× smaller | **= sym bf16** (0.2929) |
| bf16k_turbo4v     | 16 + 4.5      | 10.25   | 1.56× smaller | **= sym bf16** (0.2929) |
| bf16k_turbo2v     | 16 + 2.5      | 9.25    | 1.73× smaller | -0.01 vs bf16 |
| fp8k_turbo3v      | 8  + 3.5      | 5.75    | 1.39× smaller | **= sym fp8**  (0.8485) |
| fp8k_turbo4v      | 8  + 4.5      | 6.25    | 1.28× smaller | = sym turbo4 (0.8384) |
| fp8k_turbo2v      | 8  + 2.5      | 5.25    | 1.52× smaller | -0.22 vs fp8 (2-bit V) |
| **turbo4k_turbo3v** | 4.5 + 3.5   | **4.0** | **2.0× smaller** than fp8 | **= sym fp8** (0.8485) ⭐ |
| turbo4k_turbo8v   | 4.5 + 9       | 6.75    | mixed         | -0.20 vs fp8 |
| turbo3k_turbo8v   | 3.5 + 9       | 6.25    | mixed         | -0.16 vs fp8 |

⭐ **`turbo4k_turbo3v` is the headline frontier**: 2.0× smaller KV pool
than sym fp8 at byte-identical PPL sim on the Manhattan-Project
continuation. Halving the KV footprint at zero quality cost on this
prompt means ~2× longer context at the same VRAM ceiling.

> **Asymmetric K/V quality is model-family sensitive.** Which asym
> combo wins on quality vs compression depends on the model's
> attention-score distribution + K-side outlier mass profile. The
> recommended starting points above are based on Qwen3.6 + Qwen3-VL
> measurements; other model families may have a different sweet
> spot. The design rationale + per-family selection guidance is in
> [`TheTom/turboquant_plus/docs/papers/asymmetric-kv-compression.md`](https://github.com/TheTom/turboquant_plus/blob/main/docs/papers/asymmetric-kv-compression.md).

## Bench results

All numbers from `tests/atlas_bench_comprehensive.py` on single-GPU
GB10, greedy decode, fixed Manhattan-Project WikiText prompt, 5-run
median per metric, 3 context lengths.

### Bench D — sym matrix, upstream `87b7bb3` baseline vs TQ+ branch (Qwen3.6-35B-FP8)

Output text byte-identical between upstream and TQ+ on fp8/turbo3/turbo4
at greedy decode. Throughput at parity (±2% noise).

| dtype  | b/e  | PPL sim (base → TQ+) | pre_8K t/s (base → TQ+) | pre_16K t/s (base → TQ+) | dec_after_8K (base → TQ+) |
|--------|-----:|---------------------:|------------------------:|-------------------------:|--------------------------:|
| bf16   | 16.0 | 0.8485 → 0.8485      | 1229.35 → 1227.54       | 1394.17 → 1390.46        | 40.94 → 40.87             |
| fp8    | 8.0  | 0.8485 → 0.8485      | 1219.05 → 1239.06       | 1377.74 → 1402.00        | 41.28 → 41.65             |
| turbo8 | 9.0  | 0.8384 → 0.8384      | 1223.43 → 1248.85       | 1389.64 → 1416.28        | 40.95 → 41.11             |
| nvfp4  | 4.5  | 0.8485 → 0.8485      | 1208.02 → 1239.39       | 1366.80 → 1410.56        | 41.07 → 41.53             |
| turbo4 | 4.5  | 0.8384 → 0.8384      | 1252.60 → 1238.62       | 1417.42 → 1403.65        | 41.22 → 40.94             |
| turbo3 | **3.5** | 0.8384 → 0.8384   | 1249.94 → 1230.51       | 1416.13 → 1400.57        | 41.28 → 40.85             |
| **turbo2** | **2.5** | not in upstream enum → **0.6465** | n/a → **1336.50** | n/a → **1519.93** | n/a → 41.56 |

**Turbo2 is new in this PR** — upstream `87b7bb3`'s `KvCacheDtype`
enum has `Bf16, Fp8, Nvfp4, Turbo4, Turbo3, Turbo8` only. This PR adds
the 2-bit Lloyd-Max codebook (`Turbo2`) + its write/decode/prefill
kernels + dispatcher arms. The 2-bit codebook is the most aggressive
KV compression we ship (2.5 b/elem incl scale vs Turbo3's 3.5 b/elem
vs Turbo8's 9 b/elem); usable for low-bit asym variants like
`Bf16KTurbo2V` / `Fp8KTurbo2V` where K precision absorbs the V quality
hit. Prefill is fastest of any KV dtype tested (less data to write
per token) at the expected 2-bit quality penalty.

### Bench F — 9 asymmetric variants on Qwen3.6-35B-FP8 (head_dim=256)

| dtype             | avg b/e | PPL sim | dec_short | pre_2K  | pre_8K  | pre_16K | dec_after_8K |
|-------------------|--------:|--------:|----------:|--------:|--------:|--------:|-------------:|
| bf16k_turbo3v     | 9.75    | load_timeout — needs bf16-attn model (see Bench G)              |
| bf16k_turbo4v     | 10.25   | load_timeout                                                    |
| bf16k_turbo2v     | 9.25    | load_timeout                                                    |
| **fp8k_turbo3v**  | 5.75    | **0.8485** | 72.30  | 641.19  | 1249.05 | 1415.89 | 41.28        |
| fp8k_turbo4v      | 6.25    | 0.8384  | 71.98     | 636.42  | 1240.02 | 1403.23 | 41.03        |
| fp8k_turbo2v      | 5.25    | 0.6263  | 72.16     | 631.95  | 1232.50 | 1390.13 | 40.97        |
| **turbo4k_turbo3v** | **4.0** | **0.8485** | 72.01 | 638.85  | 1242.65 | 1410.52 | 41.04        |
| turbo4k_turbo8v   | 6.75    | 0.6465  | 71.96     | 638.85  | 1241.52 | 1402.76 | 40.99        |
| turbo3k_turbo8v   | 6.25    | 0.6869  | 71.92     | 634.67  | 1234.47 | 1398.72 | 40.72        |

`fp8k_turbo3v` and `turbo4k_turbo3v` are bit-identical to the fp8
baseline (0.8485) — the "K kept at baseline precision" promise
empirically delivered. Throughput on every asym variant matches the
symmetric line within ±1% across all 3 context lengths.

**Compression headline at fp8-parity quality**:
- `fp8k_turbo3v` (5.75 b/elem) is **1.39× smaller** than sym fp8 (8.0 b/elem)
- `turbo4k_turbo3v` (4.0 b/elem) is **2.0× smaller** than sym fp8 — same
  PPL sim 0.8485, half the KV pool footprint, ~2× longer context at same
  VRAM ceiling. This is the production-recommended row.

### Bench G — bf16k_* variants on Qwen3-VL-30B-A3B-NVFP4 (head_dim=128)

The 3 bf16k_* variants exercise the bf16 K-side at HDIM=128 — only
HDIM=128 bf16-attn model in Atlas's tested set. Compared against sym
`bf16` baseline on the same NVFP4 weight release.

| dtype             | avg b/e | PPL sim | dec_short | pre_2K  | pre_8K  | pre_16K | dec_after_8K |
|-------------------|--------:|--------:|----------:|--------:|--------:|--------:|-------------:|
| bf16 (baseline)   | 16.0    | 0.2929  | 86.18     | 740.17  | 1186.24 | 1291.79 | 41.97        |
| **bf16k_turbo3v** | **9.75** | **0.2929** | 84.69 | 709.80  | 1179.45 | 1300.06 | 39.55        |
| **bf16k_turbo4v** | 10.25   | **0.2929** | 84.29  | 709.78  | 1171.72 | 1294.01 | 40.00        |
| bf16k_turbo2v     | 9.25    | 0.3030  | 85.96     | 713.33  | 1181.17 | 1301.00 | 40.34        |

`bf16k_turbo3v` and `bf16k_turbo4v` are bit-identical to the sym bf16
baseline (0.2929) on this prompt — the asym dispatch correctly
preserves K-side precision through the combined kernel. `bf16k_turbo2v`
drops 0.01 PPL (2-bit V codebook drift, expected).

**Compression at bf16-parity quality**:
- `bf16k_turbo3v` (9.75 b/elem averaged) is **1.64× smaller** than sym
  bf16 (16 b/elem) — same PPL sim, ~40% smaller KV pool, ~1.6× longer
  context at same VRAM ceiling.
- `bf16k_turbo4v` (10.25 b/elem) is 1.56× smaller at the same PPL.
- `bf16k_turbo2v` (9.25 b/elem) shows the per-side-pool dispatch works
  at the most aggressive V-side compression (2-bit) at the expected
  small quality cost.

(PPL absolute value lower than Bench F's 0.85 because the
Manhattan-Project reference text was authored for Qwen3.6's output
style; Qwen3-VL's continuations score lower on the same fixed
reference. The intra-Bench-G column comparison — `bf16k_*` vs `bf16`
baseline — is the load-bearing signal here, not the absolute number.)

### Symmetric matrix at TQ+ default (Qwen3.6-35B-FP8)

For reference / completeness — confirms parity with Bench D's TQ+
column above + adds the Turbo2 row that didn't exist upstream.

| dtype  | bits/elem | PPL sim | dec_short | pre_2K  | pre_8K  | pre_16K | dec_after_8K |
|--------|----------:|--------:|----------:|--------:|--------:|--------:|-------------:|
| bf16   | 16.0      | 0.8485  | 71.89     | 630.69  | 1227.54 | 1390.46 | 40.87        |
| fp8    | 8.0       | 0.8485  | 72.15     | 627.97  | 1239.06 | 1402.00 | 41.65        |
| turbo8 | 9.0       | 0.8384  | 71.73     | 640.08  | 1248.85 | 1416.28 | 41.11        |
| nvfp4  | 4.5       | 0.8485  | 72.10     | 636.28  | 1239.39 | 1410.56 | 41.53        |
| turbo4 | 4.5       | 0.8384  | 71.62     | 638.01  | 1238.62 | 1403.65 | 40.94        |
| turbo3 | **3.5**   | 0.8384  | 71.65     | 634.15  | 1230.51 | 1400.57 | 40.85        |
| **turbo2** | **2.5** | 0.6465 | 71.46     | 692.35  | **1336.50** | **1519.93** | 41.56        |

Turbo2 prefill is the fastest at every context length (692 / 1337 /
1520 vs ~635 / 1235 / 1405 for everything else) — +8-9% prefill
throughput at the 2-bit Lloyd-Max quality penalty. New in this PR
(upstream enum has Turbo3/4/8 only).

## Test plan

      (`--all-features` blocked by upstream `objc2` Apple-only build on
      Linux; clean on the per-crate `-p spark-runtime -p spark-model -p spark-server`
      gates)
- [x] `bash scripts/check-license-headers.sh` — script not present in
      tree at HEAD; manual audit shows all new `crates/**/*.rs` +
      `kernels/**/*.{cu,cuh}` files carry `// SPDX-License-Identifier: AGPL-3.0-only`
      per `.licenserc.yaml` paths-policy.
- [x] `typos` — clean (`WHT` / `wht` / `BA` / `OPTIN` already in
      `_typos.toml` allow-list)
- [x] `cargo-deny check` — clean
- [x] File-size cap — 1 file lifted: `init.rs` 524 LoC after TQ+
      kernel-handle additions, allow-list entry added to
      `.github/workflows/file-size-cap.yml` with rationale matching
      the existing entries.
- [x] **Tested against real model + hardware** — full reproduction
      matrix at `docs/turboquant-plus.md`. All 9 asym variants benched
      end-to-end on GB10 with the same `tests/atlas_bench_comprehensive.py`
      harness (~85 min total wall, 5-run median per metric, 3 context
      lengths). Bench JSONs archived locally
      (`~/atlas_tqplus_bench/results/bench_{A,B,C,D,E,F,G}_*.json`),
      forwarded on request.
- [x] **Added or updated tests** — 9 new Rust unit tests:
  - 4 in `crates/spark-model/src/layers/qwen3_attention/init_kernel_dispatch.rs::tests`
    — pure-Rust dispatch-routing tests asserting every asym variant
    ends up at a kernel module name containing its dtype-pair shape
    (`bf16k_turbo3v`, `fp8k_turbo4v`, etc.); a new asym variant added
    without dedicated kernels fails the substring check in CI.
  - 5 in `crates/spark-runtime/src/kv_cache/tests_tq_plus.rs` —
    enum-coverage on `ALL_VARIANTS` / `ASYM_VARIANTS` / `SYM_VARIANTS`
    arrays that force new dtypes to be added in tests when added to
    the enum.
  - Plus `tests/test_kv_dtype_smoke.py` — end-to-end per-dtype
    container start + 64-tok generation, distinguishes load-time SKIP
    (weight-incompat) from runtime FAIL (kernel crash).

## Notes for reviewers

This is a large PR by line count (~9000 LoC across 24 new kernel
files + Rust dispatch + tests + docs) because it's the full TQ+ port,
not a feature-flag slice. I've kept the commit history intentionally
atomic so each layer is independently reviewable:

```
591aa86 chore(tq+): typo fix + allow-list 5 dispatcher files after asym additions
1be228a style(tq+): cargo fmt --all on asym kernel dispatch files
51b31c0 docs(tq+): full 9-variant asym matrix from Bench F (Qwen3.6-FP8) + Bench G (Qwen3-VL NVFP4)
7f74fbe chore(tq+): drop dead bail tombstone (superseded by dispatch tests)
b87ad59 test(tq+): extract kernel-dispatch table + dispatch-routing unit tests
45e110d feat(tq+): Turbo*K + Turbo*V combined kernels + dispatch
b049fde feat(tq+): Bf16K + Turbo{2,4}V combined kernels + dispatch
47a8c70 fix(tq+): bail at init for asymmetric variants without combined kernels
2d693f6 docs(tq+): TurboQuant+ acknowledgement + reproduction guide
53297e5 chore(file-size-cap): allow-list init.rs after TQ+ kernel handles
3b6ca3c feat(tq+): TurboQuant+ Rust dispatch + tests + smoke harness
2ba4365 feat(tq+): TurboQuant+ kernel-level integration
```

(Rebased onto upstream `main` at `9d19c32`. The `chore(ci): green
clippy` commit from an earlier draft was dropped during rebase —
upstream PR #63 landed equivalent fixes for the same
`manual_checked_ops` + `unnecessary_sort_by` clippy debt.)

Suggested review order:
1. `docs/turboquant-plus.md` — the reproduction guide + measured tables
2. `2ba4365` — kernel-level integration (algorithm correctness)
3. `3b6ca3c` — Rust dispatch + per-side cache pool refactor
4. `b049fde` + `45e110d` — asym kernel families (the 8 missing combos)
5. `b87ad59` — dispatch tests (the safety net)
6. `47a8c70` + `7f74fbe` — bail introduced as scaffolding when only
   1 of 9 asym variants had kernels, then dropped after all 8 missing
   variants landed and the dispatch tests took over

**Honest framing — the symmetric line is at parity, not a win.**
Byte-identical output on fp8/turbo3/turbo4 between upstream `87b7bb3`
and this branch at greedy decode on Qwen3.6-A3B. The kernel-level
math fixes (signs, matched-norm L2, real Turbo3 prefill kernel) are
correct on paper but greedy decoding on this model + this prompt
absorbs the numerical differences without flipping argmax winners.
The value is in: (a) the new capabilities — Turbo2 dispatch fix,
9 asym variants, per-side cache pool, (b) the test coverage that
makes the dispatcher safe to extend, (c) correctness fixes that may
matter for other models or sampling regimes.

**Known follow-ups** (not blockers):

- The `qwen3_vl` weight loader in
  `crates/spark-model/src/weight_loader/qwen3_vl.rs:93` calls
  `quantized_auto()` whose `Bf16Raw` arm is `unreachable!()` —
  prevents loading raw-bf16 Qwen3-VL checkpoints. We benched the
  bf16k_* variants against the NVFP4 release of the same model as a
  workaround. One-arm fix per the qwen35 loader pattern is
  out-of-scope for this PR.
- Metal port of the TQ+ kernels — Atlas supports a Metal backend
  (`crates/spark-runtime/src/metal_backend.rs`, `kernels/metal/`,
  Cargo feature `metal`) but the entire pre-existing TurboQuant
  family (`turbo3` / `turbo4` / `turbo8`) is CUDA-only on `main` —
  no `wht`, `turbo`, or `reshape_and_cache_turbo` siblings in
  `kernels/metal/common/`. TQ+ extends the CUDA-only line; Metal
  support for the full TurboQuant family (existing sym + new asym)
  is a separate effort.
- TQ4_1S / TQ3_1S weight quantization (separate from KV cache).
- TriAttention long-context eviction policy (substrate exists via
  `boundary_dtype` on `build_layer_kv_dtypes`, no kernel yet).

**Asymmetric dispatch correctness** — the original Bench C run
caught 8-of-9 asym variants silently falling through to the K-side
symmetric kernels with mis-sized V pools, producing PPL sim 0.55-0.63
instead of the design target. Root cause: per-side pool sizing was
asymmetric but the dispatcher routed through the K-side sym kernel
which wrote V at K-side byte size. Fixed by (a) writing the 8
missing combined kernel triplets, (b) extracting the dispatch table
to a pure function with an exhaustive match (no `_` arm), (c) adding
unit tests that walk every asym variant × {hd=128, hd=256} and
assert it routes to a kernel module name containing its dtype shape.

## Authorship

The TQ+ algorithm (matched-norm L2, sparse V dequant, asymmetric K/V
design, InnerQ, weight pre-rotation) is research authored by Tom Turney
([@TheTom](https://github.com/TheTom)) prior to this Atlas port.
Reference implementations and the broader research line live at
[`TheTom/llama-cpp-turboquant`](https://github.com/TheTom/llama-cpp-turboquant)
and [`TheTom/turboquant_plus`](https://github.com/TheTom/turboquant_plus).

## CLA


TheTom:

tbraun96:
While investigating per-layer cosine drift (on my local branch) in Qwen3.6-35B-FP8 (full-attention layers regressing 0.006 cos/layer vs HF reference), I traced a real precision bug in the `sw_exp` softmax helper used by all paged-prefill attention kernels.

**Bug location**:
- `kernels/gb10/common/prefill_paged_compute.cuh:22-31` (`sw_exp`)
- `kernels/gb10/common/prefill_paged_compute_512.cuh:38-46` (`sw_exp_512`, identical body)
- This PR's new `kernels/gb10/common/prefill_paged_compute_asym.cuh:22-31` (carries the same `sw_exp` verbatim)

The comment claims `max err ~1e-4`, but the degree-3 Taylor polynomial `1 + 0.693·tf + 0.240·tf² + 0.0555·tf³` is **off by ~50×**. Numerical check against `torch.exp` over x ∈ [-20, 0]:

```
max relative error: 5.10e-3 (~0.5%)
mean relative error: 1.14e-3
```

The error peaks at `tf` near 1.0 (where the polynomial gives 1.9755 vs true `2^0.99 = 1.9862`, ~0.54% off). For softmax inputs like `x = -0.5`, the relative error is 0.29%.

**Impact on Qwen3.6-FP8**: in the paged-prefill softmax (~18920 exp() calls per Q-position per head per full-attn layer), this accumulates to measurable per-layer cosine drift. Replacing `sw_exp` with `__expf` (CUDA SFU exp, ~2 ULP) improved early-mid layer cosine by +0.001 to +0.004 vs HF reference. (At deep layers L31-L39, removing `sw_exp` *exposed* a separate FP8 KV cache quantization issue that the polynomial was accidentally compensating for — not within this PR's scope but worth flagging.)

**Suggested fix** (minimal — same SSOT pattern works for `sw_exp_512` and the new asym header):
```cuda
__device__ __forceinline__ float sw_exp(float x) {
#ifdef ATLAS_FAST_SOFTMAX_EXP
    // FA4-style polynomial: ~0.5% max relative error at tf~1
    // Preserved as opt-in for users who measure they need it.
    float t = x * 1.4426950408889634f;
    float ti = floorf(t);
    float tf = t - ti;
    float p = 1.0f + tf * (0.6931471805599453f +
              tf * (0.2402265069591007f +
              tf * 0.05550410866482158f));
    return ldexpf(p, (int)ti);
#else
    return __expf(x);  // ~2 ULP, matches PyTorch reference softmax
#endif
}
```

Throughput cost on a single 18920-token prefill is ~13% (one extra SFU cycle per exp; ~2.25s slower out of ~17s total on GB10). Acceptable trade for correctness in production; users who measure they need throughput over precision can keep the polynomial via `-DATLAS_FAST_SOFTMAX_EXP`.

Since this PR's `prefill_paged_compute_asym.cuh` is a fresh copy of the same header, the asym kernels will inherit the same precision bug if merged as-is. Worth folding the fix in here, especially since the PR's value prop is correctness + dispatcher safety nets.

Numerical verification one-liner (against PyTorch 2.9):
```python
def sw_exp(x):
    import math
    t = x * 1.4426950408889634
    ti = math.floor(t); tf = t - ti
    return (1 + tf*(0.6931471805599453 + tf*(0.2402265069591007 + tf*0.05550410866482158))) * (2.0**ti)
# Compare vs math.exp; max relative err = 5.1e-3 over [-20, 0]
```

TheTom:
Thank you Thomas for the concise review comment. I will be working through these this afternoon.

TheTom:
@tbraun96 cherry-picked your Phase 2c precision fix from #90 onto this branch (commit 97cd300). Three files touched:

- `prefill_paged_compute.cuh` — byte-identical to your branch (`sw_exp` → `__expf` default with `ATLAS_FAST_SOFTMAX_EXP` opt-in, `bf16x2_to_f16x2_bits` helper, `__half smem_P/P64`, `__float2half_rn` stores, FP16 P×V MMA, plus the `ATLAS_DISABLE_FP16_PV` bisect ifdef).
- `prefill_paged_compute_512.cuh` — byte-identical to your branch (`sw_exp` refactor for the HDIM=512 path).
- `prefill_paged_compute_asym.cuh` — this PR's new asym fork. Applied the same logical transforms (helper, `__half` smem_P/P64, `__float2half_rn`, `.f16.f16` P×V MMA). Skipped the `ATLAS_DISABLE_FP16_PV` debug ifdef since the bisect window has passed; happy to fold it in if you want parity.

Local validation on GB10 (qwen3.6-35b-fp8, fp8 KV cache, single Manhattan-Project prompt, greedy):

| | baseline (pre-fix) | precfix (post-fix) |
|---|---|---|
| decode_short tok/s | 72.29 | 72.81 |
| prefill_2k tok/s | 636.2 | 637.97 |

Throughput parity (single run, machine was under load; the ~10% prefill cost you predicted didn't show up at this context length on Qwen3.5-35B-A3B, possibly amortised by the rest of the layer graph).

`ppl_sim` against the fixed Manhattan reference dropped 0.8485 → 0.2323, but the precfix model is producing correct content in a different trajectory: it engages the thinking process to verify the "exactly, verbatim" constraint before recalling the canonical Wikipedia opening (which includes "Nuclear physicist J. Robert Oppenheimer was the project's scientific director"). The baseline model jumps straight to the answer. Both are factually correct; `ppl_sim` is `SequenceMatcher.ratio` against a 64-char fixed string and so penalises any trajectory shift. The cosine-vs-HF bench from your branch is the right validator for this and would be the next thing to port if you want a stronger signal before merge.

Thanks again for the careful review and the fix.

tbraun96:
> The cosine-vs-HF bench from your branch is the right validator for this and would be the next thing to port if you want a stronger signal before merge.

I would say that, given Claude returned only the "speed" of the system in its table instead of supplementing it with the accuracy, Claude's own suggestion makes the most logical sense.

TheTom:
Makes sense. Cleanest path is probably: you land #90 first, then I rebase #92 onto main and inherit your cosine-vs-HF bench (`cosine_three_way.py` + the activation-dump hooks) natively. Saves a hand-port and keeps the validator pinned to whatever lineage you settle on for #90. I'll have the cosine column ready in the next push after rebase. Let me know when #90 is ready to land.

tbraun96:
> Makes sense. Cleanest path is probably: you land #90 first, then I rebase #92 onto main and inherit your cosine-vs-HF bench (`cosine_three_way.py` + the activation-dump hooks) natively. Saves a hand-port and keeps the validator pinned to whatever lineage you settle on for #90. I'll have the cosine column ready in the next push after rebase. Let me know when #90 is ready to land.

I agree. I'm currently working on it!

TheTom:
> > Makes sense. Cleanest path is probably: you land #90 first, then I rebase #92 onto main and inherit your cosine-vs-HF bench (`cosine_three_way.py` + the activation-dump hooks) natively. Saves a hand-port and keeps the validator pinned to whatever lineage you settle on for #90. I'll have the cosine column ready in the next push after rebase. Let me know when #90 is ready to land.
> 
> 
> 
> I agree. I'm currently working on it!

Hey there, how's the status? I'm ready to re-base whenever you merge in 90.

TheTom:
Superseded by #133 — same `feature/tq-plus-clean` branch rebased onto `main` now that #90 merged (its old base `fix/in-think-tool-call-leak` was deleted, which auto-closed this PR). Continuing review in #133.

Issue #99, opened 2026-05-28
[feat] Lift max_batch_size=1 under --ep-size 2 by multiplexing the head↔worker protocol


camerono:
### Motivation

When Atlas runs with `--ep-size 2 --world-size 2`, the server hard-pins `max_batch_size = 1` regardless of what the operator passes. That's spelled out in `crates/spark-server/src/main_modules/serve.rs:307-314`:

```rust
// EP: force batch_size=1 (worker protocol is single-sequence).
// MTP speculative decoding IS supported with EP via verify broadcast protocol.
let max_batch_size = if world_size > 1 {
    tracing::info!("EP active: forcing max_batch_size=1");
    1
} else {
    args.max_batch_size
};
```

For Qwen3.5-122B-A10B-NVFP4 on 2× GB10 this gate becomes the structural concurrency ceiling, and the model doesn't have alternative-topology relief. Tried both single-node Atlas and `--tp-size 2` on the same hardware. Both load all ~94 GB of weights onto the head and the KV cache pool gets 0 allocatable blocks even at `--max-batch-size 1` + `--max-seq-len 16384` + `--gpu-memory-utilization 0.95`. EP=2 is the only topology that actually shards (expert-parallel), so the EP-gate ends up being load-bearing.

What it costs in practice: a 4-concurrent agent-style burst (multi-tool agent-style Dify workflow doing parallel tool dispatch) puts 4 chat completions in flight against Atlas, the scheduler queues them serially behind the batch=1 cap, and the upstream client (Dify's plugin daemon) hits its internal 500/retry path after ~10 minutes per slot. 2/4 requests in my Bench 3 k=4 tail-spike to ~605 s with `Request to Plugin Daemon Service failed-500`. With batch=4 (or even 2) on the same Atlas process this would be a non-issue.

### Proposal

Extend the head↔worker RPC to multiplex sequences so the EP-active branch can honor `args.max_batch_size`. Concretely:

1. Add a `seq_id: u32` (or similar) to the head→worker request payload.
2. Have the worker allocate `max_batch_size` parallel KV cache slices + SSM state pool slots, parallel to what the head already does at `world_size == 1`.
3. Tag returned tokens with their `seq_id` so the head routes back to the right scheduler entry.
4. Audit the MTP `verify_broadcast` path for multi-sequence correctness — the existing single-sequence guarantee is what made that broadcast tractable, and the multi-seq case is the gnarliest interaction here.
[...]

Issue #100, opened 2026-05-28
[bug] EOS run-on: assistant content continues into next-turn role markers / prompt echo before stopping (Qwen3.5-122B NVFP4 EP=2, also single-GPU flagship)


SwyssX:
### Atlas version / docker tag

 2026-05-22T21:19:51.900023161-04:00

### Model

Sehyo/Qwen3.5-122B-A10B-NVFP4

### Hardware

2 

### Launch command

```bash
Node 0 — Master (link IP 192.168.100.1):
bashdocker rm -f atlas 2>/dev/null
sudo docker run -d --name atlas \
  --network host --gpus all --ipc=host \
  -e NCCL_SOCKET_IFNAME=enp1s0f0np0 \
  -v ~/.cache/huggingface:/root/.cache/huggingface \
  avarok/atlas-gb10:latest \
  serve Sehyo/Qwen3.5-122B-A10B-NVFP4 \
    --bind 0.0.0.0 --port 8888 \
    --world-size 2 --rank 0 \
    --master-addr 192.168.100.1 --master-port 29500 \
    --max-seq-len 131072 \
    --kv-cache-dtype nvfp4 --kv-high-precision-layers auto \
    --gpu-memory-utilization 0.93 \
    --scheduling-policy slai --enable-prefix-caching \
    --tool-call-parser qwen3_coder \
    --max-num-seqs 1
Node 1 — second Spark (link IP 192.168.100.2):
bashdocker rm -f atlas 2>/dev/null
sudo docker run -d --name atlas \
  --network host --gpus all --ipc=host \
  -e NCCL_SOCKET_IFNAME=enp1s0f0np0 \
  -v ~/.cache/huggingface:/root/.cache/huggingface \
  avarok/atlas-gb10:latest \
  serve Sehyo/Qwen3.5-122B-A10B-NVFP4 \
    --bind 0.0.0.0 --port 8888 \
    --world-size 2 --rank 1 \
    --master-addr 192.168.100.1 --master-port 29500 \
    --max-seq-len 131072 \
    --kv-cache-dtype nvfp4 --kv-high-precision-layers auto \
    --gpu-memory-utilization 0.93 \
    --scheduling-policy slai --enable-prefix-caching \
    --tool-call-parser qwen3_coder \
    --max-num-seqs 1
```

### Expected behaviour

[...]

Issue #94, opened 2026-05-27
[bug] --ngram-speculative loops and --self-speculative emits garbage on Qwen3.5-122B-A10B-NVFP4 (EP=2)


camerono:
### Atlas version / docker tag

`avarok/atlas-gb10@sha256:53f5c6df78a525d64e4b8757a97a973279fda74eb080483c0cc1bb0d05545536`

### Model

`Sehyo/Qwen3.5-122B-A10B-NVFP4`

### Hardware

2× GB10, EP=2 (head + worker, direct 200G QSFP56 between nodes)

### Launch command

```bash
docker run --detach --name atlas-122b-r0 \
  --user 1000:1000 --read-only \
  --tmpfs /tmp:size=8g,exec,mode=1777 --tmpfs /run:size=64m \
  --cap-drop=ALL --cap-add=IPC_LOCK --ulimit memlock=-1 \
  --security-opt no-new-privileges \
  --network host --ipc=host --gpus all --device=/dev/infiniband \
  -v <hf-cache>:/var/cache/hf-hub \
  -e NCCL_SOCKET_IFNAME=<iface> \
  -e NCCL_IB_HCA=<hca> \
  -e RUST_LOG=info \
  avarok/atlas-gb10@sha256:53f5c6df... \
  serve Sehyo/Qwen3.5-122B-A10B-NVFP4 \
    --model-name atlas-qwen3.5-122b \
    --cache-dir /var/cache/hf-hub \
    --bind 127.0.0.1 --port 8005 \
    --rank 0 --world-size 2 --ep-size 2 \
    --master-addr 10.0.0.1 --master-port 29500 \
    --max-seq-len 16384 --max-batch-size 1 \
    --kv-cache-dtype nvfp4 --gpu-memory-utilization 0.70 \
    --scheduling-policy slai \
    --ngram-speculative
# worker rank 1 identical except --rank 1 --port 8006
```

Swap `--ngram-speculative` for `--self-speculative` to reproduce the second symptom. The MTP+NVFP4 path (`--speculative --mtp-quantization nvfp4`) works fine on this same launch shape and same model.

### Expected behaviour

Speculative decoding emits tokens. Behaviour similar to the MTP path on the same model — MTP+NVFP4 on this launch gives +9.6% throughput over `spec-off`, so the EP=2 topology and model are healthy.

### Actual behaviour

[...]

PR #101, opened 2026-05-29
Lift max_batch_size=1 under EP=2 via head↔worker slot multiplexing (#99)

camerono:
Implements #99 — lifts the `max_batch_size = 1` clamp under `--ep-size 2` by multiplexing the head↔worker protocol, and makes the batched multi-sequence decode path it unlocks both correct and fast.

Behind `ATLAS_EP_PROTOCOL=v2` so the wire change is opt-in; v1 behaviour is unchanged when the flag is unset.

## Protocol (the design from #99, landed)

`SequenceState.slot_idx` already identifies a sequence's SSM-pool slot and the whole forward pass indexes by it — the protocol just couldn't *express* which slot the head was operating on, so the worker only ever used slot 0 and `serve.rs` was forced to clamp `max_batch_size = 1`. The change plumbs the existing `slot_idx` through as the seq_id:

1. `ep_broadcast_seq_and_cmd` / `ep_recv_seq_and_cmd` helpers (preamble is skipped when v2 is off).
2. Head emits the `seq_id` preamble at every EP broadcast site.
3. Worker dispatches multi-slot (`Vec<Option<SequenceState>>` instead of a singleton).
4. Explicit alloc-slot / free-slot command codes.
5. Drop the `world_size > 1` clamp under v2; pre-allocate all worker slots and skip retire-compaction (the worker keeps slots in place, keyed by `slot_idx`).

## Two head-side fixes the protocol exposed

These were in the model layer, not the wire protocol:

- `decode_batch_dispatch`'s EP branch wrote single-row logits to row 0 per sequence, so every sequence sampled the last one's logits. Fixed by staging each sequence's logits row (later, a d2d copy; finally subsumed by the batched path).
- Broadcasting all seq-id preambles up front put the head's comm-stream op order out of step with the worker's per-layer all-reduces, deadlocking NCCL. Fixed by interleaving the broadcasts inside the dispatch.

## Batched-decode MoE: use the fused batch kernels, not the grouped-GEMM

The SSM layers' batched MoE was dead code behind an early `return`. Reviving it with `forward_prefill` (the prefill grouped-GEMM) was a **~60% regression** — at decode batch sizes the per-expert M is ≈1, and the 256-expert sort/permute/ptr-table overhead dominates, ×36 SSM layers. The dispatch that works:

- **N=2/3** → fused `forward_k2` / `forward_k3` (one batched all-reduce, no per-token launch overhead)
- **N≥4** → per-token MoE loop

The SSM mixer stays per-sequence (it carries recurrent state); only the stateless MoE is hoisted out and batched. The attention layers had the same `forward_prefill`-at-N≥4 trap, fixed the same way.

## Measured (Qwen3.5-122B-A10B-NVFP4, EP=2 on 2× GB10, MTP on)

| Change | Effect |
| --- | --- |
| Gate lift | 4-concurrent burst now batches instead of serializing — removes the batch=1 tail-spike from #99 |
| SSM MoE dispatch (`forward_k2`) | SSM decode step **44 ms → 35 ms** at N=2 (~15–20%) |
| Attention MoE dispatch | attention block **~40 ms → ~24 ms** at N=4 |

**One expectation to set:** lifting the gate buys tail latency and admission, not aggregate throughput at low concurrency. Batched vs serialized at N=2 is ~equal (~32 tok/s either way) — the decode step is memory-bandwidth + inter-node-NCCL bound, and two concurrent tokens share almost no weight loads. Output is coherent and cross-sequence-isolated at N=2 and N=4. Full reasoning, including why CUDA graphs and single-host don't help this model, is in `docs/adr/0011-ep-batched-decode-optimization.md`.

### DIFF
--- a/crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/ffn.rs
+++ b/crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/ffn.rs
@@ -130,21 +130,28 @@ impl Qwen3AttentionLayer {
                     stream,
                 )?;
             }
-            // Phase B: BATCHED MoE via grouped-GEMM (forward_prefill path).
-            // See trait_decode_multi_seq.rs for the rationale — single
-            // grouped-GEMM per MoE sublayer instead of N × top_k per-token
-            // GEMVs.
+            // Phase B+C: per-token MoE + residual. The generic grouped-GEMM
+            // (forward_prefill) is a NET LOSS for this 256-expert MoE at
+            // small batch — per-expert M ~1 and the sort/permute/ptr-table
+            // overhead dominates (measured: attention block ~40ms vs ~20ms
+            // per-token at N=4 on GB10). N=2/3 already take the fused
+            // forward_k2/k3 branches above; this `else` only sees N>=4 (or
+            // MLA, which must avoid the batched-MoE kernels anyway), so the
+            // per-token path — identical to decode()'s MoE — is fastest here
+            // until a true batched-EP MoE kernel exists. Mirrors the SSM
+            // dispatch in qwen3_ssm/trait_decode_multi_seq.rs. Each forward()
+            // writes moe_output[0]; consume it immediately before the next
+            // iteration overwrites it.
             let normed_base = fwd.buffers.norm_output();
-            self.ffn.forward_prefill(normed_base, n, fwd, stream)?;
-            // Phase C: per-token residual_add. `hidden[i] += moe_output[i]`.
             for i in 0..n {
                 let hidden_i = hidden.offset(i * h * residual_elem);
-                let moe_out_i = fwd.buffers.moe_output().offset(i * h * bf16);
+                let normed2_i = normed_base.offset(i * h * bf16);
+                let moe_out = self.ffn.forward(normed2_i, fwd, stream)?;
                 ops::residual_add(
                     fwd.gpu,
                     self.residual_add_k,
                     hidden_i,
-                    moe_out_i,
+                    moe_out,
                     h as u32,
                     stream,
                 )?;
--- a/crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/ffn.rs
+++ b/crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/ffn.rs
@@ -109,30 +109,42 @@ impl Qwen3AttentionLayer {
             } else {
                 2usize
             };
+            // Phase A: per-token residual_add + post-attention RMS norm.
+            // Lays out `norm_output[0..n]` as a contiguous [N, h] MoE input.
             for i in 0..n {
                 let hidden_i = hidden.offset(i * h * residual_elem);
                 let o_out_i = o_out.offset(i * h * bf16); // BF16 attn output
                 let residual_i = residual.offset(i * h * residual_elem);
-                let normed2 = fwd.buffers.norm_output().offset(i * h * bf16);
+                let normed2_i = fwd.buffers.norm_output().offset(i * h * bf16);
                 ops::residual_add_rms_norm(
                     fwd.gpu,
                     self.residual_add_rms_norm_k,
                     hidden_i,
                     o_out_i,
                     &self.post_attn_norm,
-                    normed2,
+                    normed2_i,
                     residual_i,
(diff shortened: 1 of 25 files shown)


camerono:

tbraun96:
@camerono this is a genuinely strong PR, thank you. I went through it end to end and it is static-verified on my side: the workspace compiles, spark-model (65) and spark-server (489) test suites are green, and the v2 gating is correctly additive. ep_protocol_v2 is read once from ATLAS_EP_PROTOCOL at construction, defaults false, and every wire-format change (the seq_id preamble, slot pre-alloc, skip-compaction, honoring max_batch_size) is behind it, so the v1 default path is byte-for-byte unchanged. That is exactly the backward-compat property this needed.

What I liked:

- Reusing slot_idx as the seq_id instead of inventing a parallel identifier. The whole forward pass already indexes by it, so the protocol just needed to express it. The defensive fail-fast checks (slot capacity bail, claim_slot ordering invariant, unallocated-slot bail) are the right call for a wire protocol.
- The two head-side fixes the protocol exposed are the subtle ones and you caught both: staging each sequence's logits row so decode_batch_dispatch does not overwrite row 0 per seq, and interleaving the seq_id broadcasts inside the dispatch so the head's comm-stream op order stays in step with the worker's per-layer all-reduces (the NCCL deadlock).
- The MoE dispatch decision is the real engineering. Reviving the dead batched path with the prefill grouped-GEMM being a ~60% regression at decode batch sizes, because per-expert M is about 1 and the sort/permute/ptr-table overhead dominates across 36 SSM layers, is a non-obvious trap, and fused forward_k2/k3 for N=2/3 plus per-token for N>=4 is the right shape.
- ADR-0011 is excellent and unusually honest: documenting that batching gives no aggregate tok/s win at low N (the step is bandwidth and inter-node-NCCL bound), that the real wins are tail latency and admission, and that CUDA graphs do nothing here because launch is not the bottleneck. That framing will save the next person a lot of time.

Before I merge I want to close the one thing static analysis cannot prove: live multi-sequence correctness on 2x GB10 EP=2 with Qwen3.5-122B-A10B-NVFP4 and ATLAS_EP_PROTOCOL=v2. Specifically cross-sequence isolation at N=2 and N=4 (no contamination between concurrent requests), the MTP verify-broadcast path under multiple slots, and no NCCL stall under the interleaved broadcasts. I am queuing that run. Note there is a known pre-existing crash at concurrency >= 4 with deep (~8k) context on the SSM hybrid path that is unrelated to this PR, so I will treat an N=4 deep-context failure as a separate issue rather than a regression here. Will report the numbers back on this thread.

tbraun96:
Validated live on 2x GB10 (EP=2) with Sehyo/Qwen3.5-122B-A10B-NVFP4, on an image built from current main + this PR. @camerono this holds up.

What I ran and saw:
- Gate lifts correctly: with ATLAS_EP_PROTOCOL=v2 and --max-batch-size 4 the head logs 'EP v2 active: honoring max_batch_size=4' (v1 still logs 'forcing max_batch_size=1'), the SSM state pool comes up with 4 slots, and the scheduler starts in batched mode max_batch=4.
- NCCL inter-node bring-up is clean (rank 0/2 over enp1s0f0np0, 2-rank send/recv), no deadlock from the interleaved seq-id broadcasts.
- Cross-sequence isolation: fired distinct concurrent prompts at N=2, N=4, N=8 (capital-of-France, 17+25, opposite-of-hot, sky color, capital-of-Japan, 6x7, symbol-for-water, our-planet). Every response matched its OWN prompt with no contamination from a sibling in the batch. N=2 and N=4 were 8/8 content-correct; at N=8 all eight were also content-correct (the one that looked off was H2O vs my literal 'h2o' check plus thinking-mode, not a batch issue).
- Both head and worker stayed up across all 14 concurrent requests, throughput ~24 tok/s, TTFT ~370 ms.

Two notes for the record, neither blocking this PR: the trailing role-marker leak (newline-user / newline-assistant) you may notice in the raw text is the separate #40 / #100 turn-termination issue on this template, not an EP=2 batch artifact. And I exercised the non-MTP batched path here; MTP+EP multi-slot is the remaining thing to stress, which ties into #94. The batch>1 protocol itself is correct and isolated, which is the core of this PR, so merging. Excellent work, and the ADR is a genuinely useful writeup.

PR #102, opened 2026-05-29
spark-model: handle FP32 MiniMax ModelOpt dense weights

lesserevil:
## What
- Treat FP32 dense tensors from ModelOpt checkpoints as convertible dense weights instead of assuming every non-FP8 tensor is already BF16.
- Route MiniMax ignored attention projections through `dense_auto` before NVFP4 quantization and free converted/source buffers correctly.

## Why
The NVIDIA MiniMax M2.7 NVFP4 checkpoint can store ignored attention projections as FP32. Atlas previously treated all non-FP8 dense tensors as BF16, which corrupts those projections before NVFP4 quantization and breaks generation.

## Benchmarks
- `cargo fmt --all -- --check`
- `ATLAS_SKIP_BUILD=1 cargo check -p spark-model`
- `LIBRARY_PATH=/tmp/atlas-nccl-link LD_LIBRARY_PATH=/tmp/atlas-nccl-link ATLAS_TARGET_HW=gb10 ATLAS_TARGET_MODEL=minimax-m2-229b ATLAS_TARGET_QUANT=nvfp4 cargo build --release -p spark-server`
- Runtime smoke on EP=2 across `godspeed` and `savitar` with `nvidia/MiniMax-M2.7-NVFP4`: `/v1/models` listed the model, and chat prompts for `capital of Japan` and `2+2` returned expected answers.

This is a correctness fix, so I did not collect before/after performance timings.

## Authorship
AI-generated with Codex under operator direction. No human-only authored sections.

### DIFF
--- a/crates/spark-model/src/weight_loader/minimax.rs
+++ b/crates/spark-model/src/weight_loader/minimax.rs
@@ -160,13 +160,12 @@ impl ModelWeightLoader for MinimaxM2WeightLoader {
              -> Result<(DenseWeight, QuantizedWeight)> {
                 let wkey = format!("{p}.{name}.weight");
                 let scale_key = format!("{p}.{name}.weight_scale_inv");
-                let (src_ptr, src_is_fp8) = {
+                let (src_ptr, src_dtype) = {
                     let t = store.get(&wkey)?;
-                    (
-                        t.ptr,
-                        t.dtype == spark_runtime::weights::WeightDtype::FP8E4M3,
-                    )
+                    (t.ptr, t.dtype)
                 };
+                let src_is_fp8 = src_dtype == spark_runtime::weights::WeightDtype::FP8E4M3;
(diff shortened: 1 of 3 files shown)


lesserevil:

tbraun96:
@lesserevil merging, nice one. Handling the FP32 ModelOpt dense source and freeing the conversion buffer is correct, and routing the Bf16Raw arm through dense_auto is a clean way to absorb FP32 without a separate path. Bonus: this strengthens the same runtime BF16/FP32 to NVFP4 fallback that the Qwen3.5 shared-expert fix (#116) leans on, so the two compose well. The NVFP4 quant path stays unchanged for BF16 sources and the spark-model suite is green (65). The FP32-specific load needs the FP32 ModelOpt checkpoint to exercise end to end, but the change is additive and the dtype branch is guarded.

PR #103, opened 2026-05-29
spark-server: respect disabled MiniMax thinking

lesserevil:
## Summary

MiniMax M2/M2.7's native HF template always appends an open `<think>` at the generation prompt. This adds an OpenAI-variant `minimax_m2` template that keeps explicit thinking unchanged, but emits a closed `<think>\n\n</think>\n\n` prompt when a request resolves to `enable_thinking=false`, so default and tool-active requests no longer get forced or leaked reasoning despite `thinking_default=false`.

Closes #

## Test plan

- [x] `cargo test -p spark-server --lib tokenizer::tests -- --nocapture`
- [x] `cargo test -p spark-server --lib tokenizer::tests::render_minimax_openai_template -- --nocapture`
- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy --workspace --tests -- -Dwarnings`
- [x] Deployed on EP=2 MiniMax M2.7 and verified `Reply with exactly: OK` returns `reasoning_tokens=0`
- [ ] `ATLAS_SKIP_BUILD=1 cargo clippy --workspace --tests --all-features -- -Dwarnings` (not runnable on Linux: enabling all features pulls in the Metal backend's Apple-only `objc2` dependency, which compile-errors outside Apple targets)
- [ ] `bash scripts/check-license-headers.sh` (script is not present in this checkout; CI uses SkyWalking Eyes with `.licenserc.yaml`, and this PR does not add any SPDX-covered source file type)
- [ ] `typos` (not installed locally)

## Notes for reviewers

The native MiniMax template path remains unchanged. The fix uses the existing `jinja-templates/openai/{model_type}.jinja` lookup, mirroring the Qwen OpenAI-variant path. MiniMax needed a closed-thinking prompt, not simply no `<think>` prompt: with no tag, the model can still emit an implicit reasoning trace followed by `</think>`.

## CLA

### DIFF
--- a/crates/spark-server/src/tokenizer.rs
+++ b/crates/spark-server/src/tokenizer.rs
@@ -105,6 +105,40 @@ mod tests {
     use super::*;
     use serde_json::json;
 
+    fn render_minimax_openai_template(
+        messages: &[serde_json::Value],
+        tools: Option<&[serde_json::Value]>,
+        enable_thinking: bool,
+    ) -> String {
+        let template_path = concat!(
+            env!("CARGO_MANIFEST_DIR"),
+            "/../../jinja-templates/openai/minimax_m2.jinja"
+        );
+        let raw = std::fs::read_to_string(template_path)
+            .expect("bundled MiniMax OpenAI template must be present in the repo");
+        let converted = super::jinja_helpers::convert_python_jinja_to_minijinja(&raw);
+        let env = super::jinja_helpers::build_jinja_env(&converted).expect("template compiles");
+        let tmpl = env.get_template("chat").unwrap();
+        let messages_for_render = normalize_tool_call_arguments(messages);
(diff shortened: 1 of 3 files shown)


lesserevil:

tbraun96:
@lesserevil thanks, merging. This is the right fix: when thinking is disabled the MiniMax OpenAI template needs to emit the closed generation prompt (the ]~b]ai newline <think> newline newline </think> shape) rather than leaving it open, otherwise the model wanders into reasoning we explicitly turned off. I especially like that the test renders the actual bundled minimax_m2.jinja and asserts the closed-think tail (plus the with-tools and enabled variants), so it pins the real template behavior, not a mock. spark-server suite green (500) on current main.

PR #104, opened 2026-05-29
spark-server: respect MiniMax auto tool choice

lesserevil:
## Summary

MiniMax chat requests now resolve an effective tool mode before sampling so `tool_choice: "auto"` remains optional instead of forcing `<tool_call>` generation. Loop suppression also disables tool-call grammar for one recovery turn unless the caller explicitly requires a tool call.

## Test plan

- [x] `LIBRARY_PATH=/opt/vllm/nccl-blackwell/lib LD_LIBRARY_PATH=/opt/vllm/nccl-blackwell/lib ATLAS_SKIP_BUILD=1 cargo test -p spark-server sampling_tool_mode -- --nocapture`
- [x] `LIBRARY_PATH=/opt/vllm/nccl-blackwell/lib LD_LIBRARY_PATH=/opt/vllm/nccl-blackwell/lib ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy -p spark-server --tests -- -Dwarnings`
- [ ] `ATLAS_SKIP_BUILD=1 cargo clippy --workspace --tests --all-features -- -Dwarnings` could not complete on this Linux host because the workspace hits the pre-existing `objc2` Apple-platform compile gate before this change.
- [ ] `bash scripts/check-license-headers.sh` could not run because `scripts/check-license-headers.sh` is not present in this checkout.
- [ ] `typos` could not run because `typos` is not installed on this host.
- [x] Tested against a real model / hardware: deployed on godspeed + savitar with `nvidia/MiniMax-M2.7-NVFP4` EP=2; Hermes `hello!` returned a normal assistant response with no repeated recall/tool loop, and a direct synthetic loop-suppression request returned assistant text while logging tool-call grammar suppression.
- [x] Added focused unit coverage for MiniMax auto/required/specific tool choices and loop suppression.

## Notes for reviewers

This keeps bare-JSON tool mode required, but MiniMax XML tools now follow normal OpenAI `auto` semantics. Positive `<tool_call>` logit bias is only applied when a tool call is required, and loop suppression now hard-masks tool-call start tokens because explicit required/specific requests are filtered out before reaching the scheduler.

Benchmarks: not run; this is a tool-choice correctness fix, not a performance-oriented change.

Authorship: AI-generated by Codex under human operator direction; no human-written code sections are claimed.

## CLA

### DIFF
--- a/crates/spark-server/src/api/chat/mod.rs
+++ b/crates/spark-server/src/api/chat/mod.rs
@@ -26,6 +26,7 @@ mod loop_detect;
 mod msg_entry;
 pub(super) mod repair_json;
 mod sampling_setup;
+mod sampling_tool_mode;
 mod template;
 mod thinking;
 
@@ -240,6 +241,7 @@ pub(crate) async fn chat_completions_inner(
         max_tokens,
         stop_tokens,
         tool_choice_required,
+        suppress_tool_call,
         grammar_spec,
         timeout_at,
         top_logprobs,
--- a/crates/spark-server/src/api/chat/sampling_setup.rs
+++ b/crates/spark-server/src/api/chat/sampling_setup.rs
@@ -16,6 +16,7 @@ use crate::tool_parser;
 use super::super::compact::openai_error_response;
 use super::super::inference_impl::tokenize_stop_sequences;
 use super::super::inference_types::GrammarSpec;
+use super::sampling_tool_mode::resolve_tool_mode;
 
 pub(super) struct SamplingSetup {
     pub(super) temperature: f32,
@@ -34,6 +35,7 @@ pub(super) struct SamplingSetup {
     pub(super) max_tokens: usize,
     pub(super) stop_tokens: Vec<u32>,
     pub(super) tool_choice_required: bool,
+    pub(super) suppress_tool_call: bool,
     pub(super) grammar_spec: Option<GrammarSpec>,
     pub(super) timeout_at: Option<std::time::Instant>,
     pub(super) top_logprobs: Option<u8>,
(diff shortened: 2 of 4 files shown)


tbraun96:
@lesserevil heads-up: #108 (your bare-json auto fix) just landed, and it refactored the same resolver this PR rewrites, so they now collide in sampling_setup.rs. More than a textual conflict, the two disagree on intent:

- #108 (merged) ends up with: minimax_xml is parser-required, bare_json respects auto (parser_required = minimax_xml).
- this PR's resolve_tool_mode flips both: bare_json is required, minimax_xml respects auto (parser_required = bare_json), with the rationale that bare JSON has no envelope token to serve as a safe auto trigger while MiniMax XML does.

I think your reasoning here in #104 is the more principled one (the trigger-token argument is convincing), but I do not want to merge a change to the per-parser required/auto semantics for two different models (Nemotron bare_json and MiniMax) by resolving the conflict myself and guessing. Could you rebase this onto current main and fold both behaviors into the one resolver, so bare_json -> required and minimax_xml -> auto is the explicit, single source of truth? The suppressed_for_turn / loop-suppression part is a clean addition and I want to keep it.

Once it is rebased I will validate the tool-choice behavior live on MiniMax-M2.7-NVFP4 (cached here) before merging. The resolve_tool_mode abstraction with its unit table is the right shape, just needs to be the only resolver.

tbraun96:
@lesserevil closing this one for queue hygiene, not because the idea is wrong. #108 (your bare-json fix) landed and refactored the same resolver, so this now conflicts, and the two disagree on the per-parser semantics (this PR: bare_json required, minimax auto; #108 as merged: the opposite). Your trigger-token reasoning here is the more principled version, and the resolve_tool_mode struct + suppressed_for_turn loop-suppression is a clean addition I want. Please refile it rebased onto current main as the single source-of-truth resolver (bare_json -> required, minimax_xml -> auto, plus the loop-suppression), and I will validate the tool-choice behavior live on MiniMax-M2.7-NVFP4 (cached here) and merge it. Sorry to make you rebase, the timing of #108 landing first is on me.

PR #105, opened 2026-05-30
spark-server: avoid orphan tool messages in F32

lesserevil:
## Summary

F32 no longer clones an old failed `role: tool` message onto the end of the conversation. It now surfaces the failed tool result in a runtime reminder, preserving OpenAI tool-message ordering so MiniMax and other vendor templates do not reject the prompt as an orphan tool result.

## Test plan

- [x] `LIBRARY_PATH=/opt/vllm/nccl-blackwell/lib LD_LIBRARY_PATH=/opt/vllm/nccl-blackwell/lib ATLAS_SKIP_BUILD=1 cargo test -p spark-server f32 -- --nocapture`
- [x] `LIBRARY_PATH=/opt/vllm/nccl-blackwell/lib LD_LIBRARY_PATH=/opt/vllm/nccl-blackwell/lib ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy -p spark-server --tests -- -Dwarnings`
- [ ] `ATLAS_SKIP_BUILD=1 cargo clippy --workspace --tests --all-features -- -Dwarnings` still fails on this Linux host before this patch because the workspace pulls `objc2`, which requires an Apple target.
- [ ] `bash scripts/check-license-headers.sh` could not run because `scripts/check-license-headers.sh` is not present in this checkout.
- [ ] `typos` could not run because `typos` is not installed on this host.
- [x] Tested against a real model / hardware: deployed on godspeed + savitar with `nvidia/MiniMax-M2.7-NVFP4` EP=2; a synthetic F32 regression request now returns HTTP 200 and logs `F32: surfaced most-recent failed tool_result in a runtime reminder`; Hermes smoke test no longer hits the MiniMax template 400.
- [x] Added focused unit coverage for stale failed-tool surfacing and the already-fresh no-op case.

## Notes for reviewers

The previous F32 behavior attempted to make a stale failure fresh by duplicating the original tool result. That creates invalid OpenAI history whenever messages after the original assistant/tool pair are not assistant tool calls, and MiniMax's template rejects it with "Message has tool role, but there was no previous assistant message with a tool call".

This keeps the guard's intent while staying inside the template contract: the stale failure is copied into a system-reminder appended to the latest user/tool message instead of being represented as a new `role: tool` turn.

Benchmarks: not run; this is a prompt-history correctness fix, not a performance-oriented change.

Authorship: AI-generated by Codex under human operator direction; no human-written code sections are claimed.

## CLA

### DIFF
--- a/crates/spark-server/src/api/chat_phases.rs
+++ b/crates/spark-server/src/api/chat_phases.rs
@@ -136,9 +136,10 @@ pub(super) fn apply_failure_guards(req: &mut ChatCompletionRequest) -> F23Progre
         );
     }
 
-    // F32: duplicate failed tool_result at conversation tail.
+    // F32: surface the most recent failed tool_result without creating
+    // an orphan role=tool message that vendor templates reject.
     if f32_reposition_failed_tool_result(&mut req.messages) {
-        tracing::info!("F32: duplicated most-recent failed tool_result at conversation tail");
+        tracing::info!("F32: surfaced most-recent failed tool_result in a runtime reminder");
     }
 
     // F39: cross-turn permanent-failure circuit breaker.
--- a/crates/spark-server/src/api/failures/circuit.rs
+++ b/crates/spark-server/src/api/failures/circuit.rs
@@ -90,8 +90,14 @@ pub fn f32_reposition_failed_tool_result(
     {
         return false;
     }
-    let dup = messages[idx].clone();
-    messages.push(dup);
+    let failed_tool_result = messages[idx].content.text.trim();
+    let reminder = format!(
+        "\n\n<system-reminder>\nThe most recent failed tool result is still relevant. \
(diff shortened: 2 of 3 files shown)


tbraun96:
@lesserevil thanks for this, merging. This was a real foot-gun: the old F32 path duplicated the failed role=tool message at the conversation tail, which lands as an orphan tool turn (a tool message not paired with a preceding assistant tool_calls), and the Qwen/MiniMax chat templates reject exactly that shape. Surfacing the failure as a system-reminder appended to the last user turn keeps the signal (do not retry the same call) without producing a template-invalid transcript. The f32_surfaces_error_without_orphan_tool_message and f32_no_op_when_failed_tool_is_already_fresh tests pin both directions. Verified the spark-server suite green against current main before merging.

PR #108, opened 2026-05-30
spark-server: respect bare-json auto tool choice

lesserevil:
## Summary
- make the bare_json tool parser respect tool_choice=auto by using triggered grammar instead of requiring a tool call from token 1
- keep explicit required/specific tool choices enforced from the first token
- add focused coverage for bare_json, explicit required/specific choices, inactive tools, and existing minimax_xml behavior

## Testing
- cargo fmt
- ATLAS_SKIP_BUILD=1 cargo check -p spark-server
- ATLAS_SKIP_BUILD=1 cargo test -p spark-server --bin spark tool_choice_required_for_parser (fails at link on this host: /usr/bin/ld cannot find -lnccl)

### DIFF
--- a/crates/spark-server/src/api/chat/sampling_setup.rs
+++ b/crates/spark-server/src/api/chat/sampling_setup.rs
@@ -39,6 +39,24 @@ pub(super) struct SamplingSetup {
     pub(super) top_logprobs: Option<u8>,
 }
 
+fn tool_choice_required_for_parser(
+    tools_active: bool,
+    tool_choice: Option<&tool_parser::ToolChoice>,
+    parser_name: Option<&str>,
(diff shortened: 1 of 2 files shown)


tbraun96:
@lesserevil thanks for this, merged. Nice catch on the bare-json path: previously the bare_json parser unconditionally forced tool_choice=required semantics, so even an auto request had the schema clamped from token 1 and the model could never just answer in plain text. Pulling the decision into tool_choice_required_for_parser() and only requiring when the caller actually asks (required/specific), or for the parsers that genuinely need it (minimax_xml), is exactly right, and the unit tables make the intent unambiguous. Verified the full spark-server suite green against current main before landing.

PR #98, opened 2026-05-27
feat(kernels): add qwen3.5-397b-a17b NVFP4 EP=4 kernel target

SeedSource:
## What
Adds the `qwen3.5-397b-a17b` NVFP4 kernel target for `nvidia/Qwen3.5-397B-A17B-NVFP4`, so a GB10 4-node EP=4 deployment resolves to a compiled kernel target instead of failing dispatch.

- `kernels/gb10/qwen3.5-397b-a17b/MODEL.toml` (model root): two `[[model_types]]` entries — `qwen3_5_moe` and `qwen3_6_moe`, both at `hidden_size = 4096`. The NVFP4 checkpoint is mrope-interleaved, so `parse_config` rewrites its `model_type` to `qwen3_6_moe` (mirrors the `qwen3.5-122b-a10b` dual-entry pattern). `[behavior]` matches the model's tested profile: thinking off by default, `max_thinking_budget=128`, no MTP.
- `kernels/gb10/qwen3.5-397b-a17b/nvfp4/`: `gated_delta_rule.cu`, `moe_w4a16_grouped_gemm.cu`, `w4a16_gemm.cu`, `KERNEL.toml` copied from `qwen3.5-35b-a3b/nvfp4/` (these tile over runtime M/N/K and branch on a runtime `num_experts`; the softmax router `moe_topk_softmax` is sized `MAX_EXPERTS=512`, fitting the 512 experts exactly). `KERNEL.md` rewritten for the 397B.
- `docker/gb10/qwen3.5-397b-a17b/nvfp4/{Dockerfile,QUICKSTART.md}`: per-model image for the 4-node EP=4 deployment.
- `tests/run_all_models.py`: a recorded-but-skipped EP4 round (the harness's multi-rank driver launches 2 ranks; the real EP=4 test runs via the 4-node launcher).

## Why
Launching `nvidia/Qwen3.5-397B-A17B-NVFP4` on 4×GB10 (EP=4/TP=1) currently fails at dispatch:
```
Error: No compiled kernel target matches model_type 'qwen3_6_moe' / hidden_size=4096
```
No existing target declares `qwen3_6_moe @ 4096` (the 35B is a `qwen3_5_moe` wildcard; the 3.6-35B is `qwen3_6_moe @ 2048`). EP=4/TP=1 is required because `num_key_value_heads=2` cannot shard across 4 TP ranks.

## Benchmarks / Verification
[...]

PR #116, opened 2026-06-03
fix(weights): allow BF16 ignore-list shared experts in Qwen3.5 NVFP4 loader (#97)

### LINKED ISSUE #97: Qwen35WeightLoader requires weight_scale for BF16 ignore-list modules — blocks Qwen3.5-397B-A17B-NVFP4 serving
## Summary
`Qwen35WeightLoader` fails to build `nvidia/Qwen3.5-397B-A17B-NVFP4`: it unconditionally requires an NVFP4 `weight_scale` for `mlp.shared_expert.gate_proj`, but that module is intentionally left in BF16 by the checkpoint's quant config, so no scale tensor exists. Dispatch, NCCL ring, and 10/11 weight shards succeed; the build then fails.

## Repro
4-node GB10 (EP=4) with the `qwen3.5-397b-a17b` kernel target compiled in (companion PR):
```
serve nvidia/Qwen3.5-397B-A17B-NVFP4 \
  --tp-size 1 --ep-size 4 --world-size 4 \
  --kv-cache-dtype nvfp4 --max-seq-len 4096 --gpu-memory-utilization 0.75
```

## Error
```
Error: Failed to build model
Caused by: Weight 'model.language_model.layers.0.mlp.shared_expert.gate_proj.weight_scale' not found in store
```

## Root cause
The checkpoint's `quantization_config` (ModelOpt NVFP4) has a 184-entry `ignore` list keeping the shared experts (+ `shared_expert_gate`, `lm_head`, `linear_attn`) in BF16:
```
"lm_head",
"*.mlp.shared_expert.*",
"model.language_model.layers.{0..59}.mlp.shared_expert*",
"model.language_model.layers.{0..59}.mlp.shared_expert_gate",
"model.language_model.layers.{0..59}.linear_attn*"
```
BF16 tensors have no `weight_scale`. The loader's shared-expert path assumes NVFP4 and looks up `shared_expert.gate_proj.weight_scale` unconditionally. (The 35B/122B checkpoints quantize their shared experts, so this path has worked to date.)

## Evidence (rank-0 log)
```
Quantization config: method="modelopt", algo="NVFP4", 184 module(s) in ignore list
Selected kernel target: (sm_121, qwen3.5-397b-a17b, nvfp4) (88 modules)
NCCL ... nRanks 4 nNodes 4 ... Init COMPLETE
Fast-loading shard 10/11 ...
Error: Failed to build model
Caused by: Weight '...shared_expert.gate_proj.weight_scale' not found in store
```

## Two candidate fixes (design decision for the Atlas team — not a recommendation)
1. **Quantize the BF16 shared experts to NVFP4 at load** — reuses the existing `Bf16Raw` / `quantize_bf16_to_nvfp4` machinery; keeps the NVFP4 shared-expert forward path unchanged. Tradeoff: overrides the checkpoint's deliberate BF16-for-quality choice for the shared experts (possible quality regression).
2. **Add a BF16 shared-expert forward path** — detect ignore-listed (BF16) shared experts and route them through a BF16 GEMM, preserving intended precision. Tradeoff: loader + forward (+ possibly a kernel).

Which fits Atlas's quantization design is your call — happy to implement whichever you prefer.

Environment: GB10 (SM121), 4-node EP=4, Atlas built from `main` + the `qwen3.5-397b-a17b` target. Checkpoint: `nvidia/Qwen3.5-397B-A17B-NVFP4`.

AI-generated using Claude Code (Opus 4.7).

tbraun96:
Fixes #97.

## Problem
`Qwen35WeightLoader` fails to build `nvidia/Qwen3.5-397B-A17B-NVFP4`:

    Weight 'model.language_model.layers.0.mlp.shared_expert.gate_proj.weight_scale' not found in store

## Root cause
`load_moe_qwen35`'s `load_expert` took the catch-all NVFP4 arm via `quantized_auto`, which unconditionally requires a `weight_scale`. The checkpoint's ModelOpt NVFP4 `quantization_config` ignore list keeps `*.mlp.shared_expert.*` in BF16, so no scale ships and the build hard-fails.

## Fix
Route the expert `_` arm through `quantized_any`, whose per-key `has_only_dense` fallback runtime-quantizes a raw-BF16 projection to NVFP4 (the same path the `Bf16Raw` and routed-expert loads already use, and the same fallback added for RedHatAI/Qwen3-Coder-Next-NVFP4's BF16 `linear_attn.out_proj`). For checkpoints whose shared expert IS quantized (35B/122B), `weight_scale` is present so `has_only_dense` is false and the path is **byte-identical to before** (no regression).

Companion to the merged kernel target #98. Verified: `spark-model` suite green (65), builds clean. The change is additive (only adds a fallback where there was previously a hard error); full 397B EP=4 load verification needs the 4-node GB10 rig from the report.

### DIFF
--- a/crates/spark-model/src/weight_map/ssm_qwen35.rs
+++ b/crates/spark-model/src/weight_map/ssm_qwen35.rs
@@ -97,6 +97,15 @@ pub(crate) fn load_moe_qwen35(
         quantize_to_nvfp4(&bf16, n, k, gpu, absmax_k, quantize_k, stream)
     };
 
+    // Runtime-quant context for the `quantized_any` BF16 fallback used when a
+    // shared/routed expert ships as raw BF16 (no weight_scale) per the
+    // checkpoint's ModelOpt ignore list (issue #97).
+    let qctx = QuantizeCtx {
+        absmax_k,
+        quantize_k,
+        stream,
+    };
+
     // Qwen3.6-35B-A3B BF16 release ships a FUSED MoE layout: one
     // `experts.gate_up_proj: [num_experts, 2*inter, hidden]` and one
     // `experts.down_proj: [num_experts, hidden, inter]` per layer. Slice
@@ -173,10 +182,43 @@ pub(crate) fn load_moe_qwen35(
                     stream,
                 )?,
             }),
+            // `quantized_any` (not `quantized_auto`): some NVFP4 checkpoints
+            // keep the shared expert in BF16 via the ModelOpt quant_config
+            // `ignore` list (e.g. Qwen3.5-397B-A17B-NVFP4, whose 184-entry
+            // ignore list covers `*.mlp.shared_expert.*`), so no weight_scale
+            // ships for it. `quantized_any`'s per-key `has_only_dense` fallback
+            // runtime-quantizes the raw BF16 — the same path the routed experts
+            // already take — instead of hard-failing on the absent scale
+            // (issue #97). For checkpoints whose shared expert IS quantized
+            // (35B/122B), the scale is present so this is identical to before.
             _ => Ok(ExpertWeight {
-                gate_proj: quantized_auto(store, &format!("{prefix}.gate_proj"), gpu, variant)?,
-                up_proj: quantized_auto(store, &format!("{prefix}.up_proj"), gpu, variant)?,
-                down_proj: quantized_auto(store, &format!("{prefix}.down_proj"), gpu, variant)?,
+                gate_proj: quantized_any(
+                    store,
+                    &format!("{prefix}.gate_proj"),
+                    inter,
+                    h,
+                    gpu,
+                    variant,
+                    qctx,
+                )?,
+                up_proj: quantized_any(
+                    store,
+                    &format!("{prefix}.up_proj"),
+                    inter,
+                    h,
+                    gpu,
+                    variant,
+                    qctx,
+                )?,
+                down_proj: quantized_any(
+                    store,
+                    &format!("{prefix}.down_proj"),
+                    h,
+                    inter,
+                    gpu,
+                    variant,
+                    qctx,
+                )?,
             }),
         }
     };


SeedSource:
Verified #116 on our 4-node GB10 rig.

**Setup**
- Atlas commit: `7a903aa` (current main at test time, including #116 + #117)
- Model: `nvidia/Qwen3.5-397B-A17B-NVFP4`
- Hardware: 4× DGX Spark GB10 / SM121
- Parallelism: TP=1, EP=4, world_size=4
- KV cache dtype: nvfp4, max_seq_len: 4096, gpu_memory_utilization: 0.75

**#116 result — the prior missing `shared_expert` `weight_scale` failure is fixed.**
- All 11/11 shards loaded; previous failure was at 10/11.
- Layers 0..60 loaded; shared experts route through the runtime-quantized path instead of hard-failing on the missing `weight_scale`.
- NCCL ring formed: rank 0, nranks 4, nNodes 4, Init COMPLETE.
- MTP weights also loaded — noting load behavior only, not claiming MTP performance.

**Separate follow-up discovered after #116 (a #98 target gap, not a #116 issue).**
Off current main, model build then failed at:

    Kernel lookup vision_encoder::vision_gemm_bias: Module 'vision_encoder' not loaded

The 397B NVFP4 checkpoint includes vision config/weights, and the model builder constructs the vision encoder even for this text-serving path. The `qwen3.5-397b-a17b` target from #98 was cloned from the text-only 35B target and omitted `vision_encoder.cu`; the 122B VL target already ships it. We copied it byte-for-byte (no kernel logic change, no `KERNEL.toml` change). Follow-up PR: #124.

**Result with the one-file follow-up:**
- 397B loads fully: 11/11 shards + vision encoder (depth=27) + MTP weights
- NCCL ring forms across 4 ranks
- server listens on `:8888`; `/v1/models` returns the model
- deterministic probes (temp=0) return coherent output:
  - "What is 2+2?" → `4`
  - `add(a, b)` → coherent Python
  - ~1.2k-token needle prompt → retrieved correctly

**Perf note:** not quoting tok/s yet — this run is socket-bound because NCCL reports `NET/IB: No device found → Using network Socket` (our launch container isn't exposing the RoCE devices). We'll fix container IB/RoCE exposure and report throughput separately so as not to understate Atlas.

Evidence retained locally (full rank logs, build logs, image digests, launch command, `/v1/models` response, three raw completion probe JSONs) — happy to share.

*(AI-generated report via Claude Code, Opus 4.8.)*

Issue #107, opened 2026-05-30
[bug] can't load nvidia/Qwen3.6-35B-A3B-NVFP4


Slach:
### Atlas version / docker tag

avarok/atlas-gb10

### Model

nvidia/Qwen3.6-35B-A3B-NVFP4

### Hardware

1x GB10, Single GPU

### Launch command

same with https://github.com/Avarok-Cybersecurity/atlas/issues/106

just change 

`serve ${ATLAS_MODEL:-nvidia/Qwen3.6-35B-A3B-NVFP4}`

### Expected behaviour

successfull starting

### Actual behaviour

atlas-1  | 2026-05-30T02:57:58.072981Z  INFO spark::main_modules::serve: Atlas Spark starting...
atlas-1  | 2026-05-30T02:57:58.073005Z  INFO spark::main_modules::serve: Licensed under AGPL-3.0-only — see /LICENSE in this container
atlas-1  | 2026-05-30T02:57:58.073429Z  INFO spark::model_resolver: Model: nvidia/Qwen3.6-35B-A3B-NVFP4 (resolved to /root/.cache/huggingface/hub/models--nvidia--Qwen3.6-35B-A3B-NVFP4/snapshots/6c7f09d4036e97393f82e9f9ecd1a5c35ca5ee92)
atlas-1  | 2026-05-30T02:57:58.073440Z  INFO spark::main_modules::serve: Port: 30003
atlas-1  | 2026-05-30T02:57:58.073442Z  INFO spark::main_modules::serve: SSM decode dtype: f32 (full precision)
atlas-1  | 2026-05-30T02:57:58.074562Z  INFO spark::main_modules::serve: Quantization config: method="modelopt", algo="MIXED_PRECISION", format="", 2 module(s) in ignore list
atlas-1  | 2026-05-30T02:57:58.074575Z  INFO spark::main_modules::serve: Model config: 40 layers, 10 attention, 30 SSM, 256 experts, rope_theta=10000000, head_dim=256, rotary_dim=64
atlas-1  | 2026-05-30T02:57:58.074653Z  INFO spark::main_modules::serve: Selected kernel target: (sm_121, qwen3.6-35b-a3b, nvfp4) (90 modules)
atlas-1  | 2026-05-30T02:58:06.074257Z  INFO spark_runtime::cuda_backend: AtlasCudaBackend initialized on GPU 0 with 90 PTX modules
[...]

PR #117, opened 2026-06-03
fix(weights): free dead BF16 intermediates in dense loader (#A1)

tbraun96:
## Bug (#A1)

The dense Qwen3.5/3.6 weight loader (`crates/spark-model/src/weight_loader/qwen35_dense.rs`) builds the NVFP4 attention and SSM weights from BF16 intermediates, then **retained** those BF16 intermediates in the `AttentionWeights` / `SsmWeights` structs:

- attention: `q_proj` / `k_proj` / `v_proj` held the BF16 `q/k/v_dense`
- SSM: `in_proj_qkvz` held the BF16 `qkvz_dense` — a full **third** full-precision copy of the largest per-layer tensor (BF16 source + NVFP4 + NVFP4-transposed)

Net effect: **~2-3x weight bloat**. It was caught on a 61 GB AMD box during the SCALE port (OOM at layer 50/64), but it is **hardware-agnostic** — it wastes the same memory on GB10/CUDA, where it went unnoticed only because the model fits within 119 GB.

## Fix

The BF16 fields are **provably dead**: once the NVFP4 weights are built, prefill AND decode always dispatch the NVFP4 weights (and, when enabled, the native-FP8 prefill weights built by `bf16_to_fp8`). Nothing reads the BF16 `q/k/v_proj` or `in_proj_qkvz` afterward. This mirrors the existing NULL-stub convention the MLA loaders already use for `q_proj`.

The fix:
1. **Attention block**: free `q/k/v/o_dense.weight` after `quantize_to_nvfp4`; store `DenseWeight { weight: DevicePtr::NULL }` in `q_proj`/`k_proj`/`v_proj` (`o_proj` was already the NVFP4 weight).
2. **SSM concat**: free the concat inputs `qkv_dense.weight` / `z_dense.weight` immediately after `gpu_concat_rows`.
3. **SSM block**: free `qkvz_dense.weight` / `out_proj_dense.weight` **after** their last consumer (the optional native-FP8 prefill `bf16_to_fp8` build + `synchronize`), and NULL-stub `in_proj_qkvz`.

`DevicePtr::NULL` is the codebase's existing sentinel (`spark-runtime/src/gpu.rs:17`, `pub const NULL: Self = Self(0)`); `gpu.free(NULL)` is a tested no-op (`cuda_backend/tests.rs::null_free_is_noop`).

## Verification

- This is **provably-dead-code removal**, verified by `cargo check -p spark-model` (clean — finished in 34.8s, 0 errors) and by review of all dispatch sites for `q/k/v_proj` and `in_proj_qkvz`.
- Confirmed the freed buffers are distinct, freshly-allocated (no aliasing/double-free) and that all reads precede the frees.
- **Runtime verification is pending a free GPU** (both dgx1 and dgx2 are currently occupied).

## Follow-up (not in this PR)

`load_mtp_weights` in this file is a thin wrapper that delegates to `load_mtp` in `weight_map/loaders_moe.rs`; that function uses `dense`/`dense_auto` loads rather than the build-NVFP4-from-BF16-then-retain pattern, so it does not exhibit the same clear bug. Whether MTP has any analogous retain is left as a separate follow-up rather than guessed at here.

### DIFF
--- a/crates/spark-model/src/weight_loader/qwen35_dense.rs
+++ b/crates/spark-model/src/weight_loader/qwen35_dense.rs
@@ -171,15 +171,25 @@ impl ModelWeightLoader for Qwen35DenseWeightLoader {
                                 (q_dense, q_nvfp4),
                                 (k_dense, k_nvfp4),
                                 (v_dense, v_nvfp4),
-                                (_o_dense, o_nvfp4),
+                                (o_dense, o_nvfp4),
                             ] = load_qkvo_tp(config, load_bf16_then_nvfp4)?;
 
                             let (k_scale, v_scale) = load_kv_scales(store, &p, gpu);
 
+                            // The BF16 q/k/v/o dense tensors are only the intermediate
+                            // fed to the GPU quantize_to_nvfp4 above. Prefill AND decode
+                            // always dispatch the NVFP4 weights, so the BF16 copies are
+                            // dead once quantized. Free them instead of retaining a full
+                            // second copy of every projection (Atlas issue #A1).
+                            gpu.free(q_dense.weight)?;
+                            gpu.free(k_dense.weight)?;
+                            gpu.free(v_dense.weight)?;
+                            gpu.free(o_dense.weight)?;
+
                             let attn = AttentionWeights {
-                                q_proj: q_dense,
-                                k_proj: k_dense,
-                                v_proj: v_dense,
+                                q_proj: DenseWeight { weight: spark_runtime::gpu::DevicePtr::NULL },
+                                k_proj: DenseWeight { weight: spark_runtime::gpu::DevicePtr::NULL },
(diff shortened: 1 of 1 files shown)


tbraun96:
Validated live and marking ready. Built this on top of current main (with the #107 MIXED_PRECISION work, which creates exactly these kinds of dead BF16 intermediates) and loaded Qwen/Qwen3.6-27B-FP8 through the dense loader, which is the path that frees and NULLs q/k/v/o_proj + in_proj_qkvz here. It loads clean and both prefill and decode generate coherent output (Paris; a full coherent sentence about the sea) with no NULL-deref, which confirms the BF16 q/k/v/o and qkvz copies really are dead once the NVFP4 weights are built, exactly as the PR argues. That recovers a full second copy of every attention projection plus the largest SSM tensor per layer. Merging.

PR #119, opened 2026-06-03
feat: Add Step 3.7 Flash NVFP4 support (kernel target + config parser + weight loader + EP)

marksunner:
### What this PR adds

Support for [stepfun-ai/Step-3.7-Flash-NVFP4](https://huggingface.co/stepfun-ai/Step-3.7-Flash-NVFP4) on DGX Spark (GB10), including:

- **Kernel target** (`kernels/gb10/step3p7-flash/nvfp4/`) — MODEL.toml, KERNEL.toml with `-DHDIM=128`, 6 model-specific .cu overrides
- **Config parser** (`config/parsers/step3p7.rs`) — handles the nested `text_config` format with field preprocessing for serde compatibility
- **Weight loader** (`weight_loader/step3p7.rs`) — supports both fused and per-expert tensor formats (see below)
- **Preprocessing script** (`scripts/preprocess_step3p7_experts.py`) — splits fused expert tensors for EP support

### The fused expert problem (and solution)

Step 3.7 NVFP4 stores all 288 experts in single fused tensors per projection type:
```
model.language_model.layers.3.moe.gate_proj.weight  [288, 1280, 2048]  U8
```

Atlas EP filtering relies on `parse_expert_index()` finding `.experts.N.` in tensor names. Fused tensors bypass this, causing:
1. OOM on single-Spark (120 GB on-disk > ~115 GB free after CUDA init)
2. No EP support (can't skip remote experts during loading)

**Solution:** A preprocessing script splits fused tensors into per-expert named tensors:
```
model.language_model.layers.3.moe.experts.0.gate_proj.weight  [1280, 2048]  U8
```

Same data, same bytes, just reorganised. The weight loader detects which format is present and handles both:
- **Per-expert** (preprocessed): works with EP filtering, required for EP≥2
- **Fused** (original): uses GPU pointer arithmetic slicing, only for single-GPU with enough memory

### Architecture details

Step 3.7 is a hybrid of MiniMax M2 and Qwen 3.5 patterns:
- Sigmoid MoE routing with correction bias (MiniMax M2)
- 288 experts top-8 (vs MiniMax 256)
- Shared expert per MoE layer (Qwen 3.5)
- Attention gate g_proj (Qwen 3.5)
- Partial RoPE 0.5 (MiniMax M2)
- Mixed full + sliding attention (window=512)
- 45 hidden layers: 3 dense FFN (0-2) + 42 MoE (3-44)
- 3 MTP draft modules at layers 45-47

### Status

- [x] Config parsing ✅
- [x] Kernel selection (90 PTX modules) ✅
- [x] GPU init ✅
- [x] Weight format detection (ModelOpt NVFP4) ✅
- [x] EP topology detection ✅
- [x] Fused expert tensor splitting (preprocessing script) ✅
- [x] Dual-format weight loader (fused + per-expert) ✅
- [x] Runtime validation with EP=2
- [ ] MTP loader (currently returns empty — run with `--speculative 0`)

### Usage
```bash
# Preprocess checkpoint (one-time, ~3 minutes on NVMe)
python3 scripts/preprocess_step3p7_experts.py /path/to/Step-3.7-Flash-NVFP4 /path/to/output

# Run with EP=2
atlas serve /path/to/output --world-size 2 --ep-size 2 --speculative 0
```

### Tested on
- DGX Spark (GB10, 128 GB unified memory)
- Preprocessing: 504 fused tensors → 145,152 per-expert tensors + 1,435 pass-through = 146,587 total
- Output checkpoint: 120.37 GB across 12 shards


marksunner:

AzeezIsh:
Woah @marksunner this is amazing. Will test and get back to you today. Thanks for your hard work on this!

marksunner:
recheck

nuonrg:
I tried running it on 2 nodes and it halted on this error, couldn't get it to run.
on query the forward-pass crashes on layer 3
"Inference error: prefill_chunk failed: Prefill chunk layer 3 failed: ffn.forward_prefill failed: NCCL error in ncclGroupEnd: unhandled cuda error"
1:42 [0] include/alloc.h:134 NCCL WARN Cuda failure 'an illegal memory access was encountered'

marksunner:
Thanks for testing this @nuonrg! We've hit the exact same crash on our EP=2 setup — layer 3 (first MoE layer), `ffn.forward_prefill`, `CUDA_ERROR_ILLEGAL_ADDRESS`.

We've narrowed it down to the MoE GEMM kernels and the `moe_intermediate_size=1280` that Step 3.7 uses. The existing Atlas MoE kernels were tuned for models with `moe_intermediate_size=1024` — we believe there's a tile size or alignment assumption in the grouped GEMM that breaks at 1280.

Everything else works: config parsing, weight loading, NCCL init, EP topology, attention forward pass (all 45 layers with heterogeneous Q heads), server startup. It's specifically the expert GEMM computation that crashes.

We're actively debugging — next steps are feeding the MoE CUDA kernels through a detailed review to find the hardcoded dimension, and if needed running `compute-sanitizer --tool memcheck` to pinpoint the exact instruction. We also have a reference MoE GEMM implementation (b12x, CuTe DSL) that handles arbitrary dimensions, which should help us identify where Atlas's kernel makes the 1024-specific assumption.

Will push a fix as soon as we crack it.

marksunner:
Progress update on the MoE crash:

**Root cause found (partial):** The post-load MoE transpose pass in `m2_setup.rs` is gated on `model_type == "minimax_m2"` — Step 3.7 (`model_type = "step3p7"`) was skipping it entirely. Without transposed weights, the prefill falls back to the uncoalesced GEMM path, which crashes.

**Fix pushed** (commit `4d228c8`): extended the gate check + made the transpose safety margin configurable via `ATLAS_MOE_TRANSPOSE_SAFETY_MB` (Step 3.7 EP=2 is tight on memory — ~36 GB free vs 35.6 GB transpose cost + 2 GB default margin).

**Status:** With the fix, the transpose pass now runs successfully and both ranks come up clean. However, inference still crashes at layer 3 during the MoE forward — there's a second issue hiding behind the first (likely dimension-related in the GEMM or shared expert path at `moe_intermediate_size=1280`). Running `compute-sanitizer` next to pinpoint the exact kernel.

All CUDA kernels in the Step 3.7 target are byte-identical to Qwen 3.5 122B's working kernels — so the crash is likely in how host-side code parameterises the kernel launch, not in the CUDA itself.

marksunner:
**Inference working on EP=2!** 🎉

Pushed commit `f9f96ca` with three fixes:

1. **Ungated attention** — `Qwen3AttentionLayer::new()` defaults `gated=true`, but Step 3.7 uses per-head `g_proj` (not interleaved Q+G). The doubled Q projection dimension overflowed the QKV buffer on 96-head sliding-attention layers → `CUDA_ERROR_ILLEGAL_ADDRESS`. Fix: `new_ungated()`.

2. **Sliding window detection** — was inverted. `config.num_attention_heads` is set to MAX (96) for buffer sizing, so the old `actual != config` check tagged full-attention layers (64 heads) as sliding and vice versa.

3. **EP expert pointer table** — `filter_map()` dropped remote experts, leaving the pointer table undersized vs the kernel grid. Fixed with `map()` + `ExpertWeight::null()`.

Combined with the previous commit (`0f32b62`, MoE transpose pass), Step 3.7 Flash now runs on EP=2 across two DGX Sparks: TTFT ~1.9s, decode ~21 tok/s. Output quality still needs work (attention gate not yet implemented, rope_theta assignment being refined), but the inference engine is stable — no crashes across repeated queries.

tbraun96:
Awesome work!!

PR #90 will help here, as well as with other open PRs. There were numerous bug fixes in the common kernels, and agentic coding works now even across long contexts. 

I will do a nightly release today, then, we can merge #90 after feedback, then, we can work on merging this PR. I am changing the base of this PR to #90 since that's going to be the merge order anyways. Also, if you're on discord, please contact me in the #development channel, where we all stay in sync.

marksunner:
**Progress Update — June 8**

Quick status recalibration:

**Working:** Engine runs stable on EP=2 (21 tok/s decode, no crashes). 8 bugs found and fixed across two sessions (MoE transpose gate, ungated attention, sliding window detection, EP pointer table, per-layer RoPE rotary_dim, g_proj head gate implementation, routed_scaling_factor, shared expert gate).

**NOT working yet:** Output quality is still degenerate. Root cause isolated to MoE expert GEMV producing values ~3-5× too small for `moe_intermediate_size=1280`. Dense layers and attention work correctly. Investigating the fused CUDA expert kernel dimension handling.

Close but not there yet — likely one more kernel-level fix.

marksunner:
### Summary

The `dense_gemm_tc` kernel (`kernels/gb10/common/dense_gemm_tc.cu`) has a cooperative A-tile load that uses a single-iteration `if` statement instead of a loop. With `TC_BLOCK=128` threads and `TC_TM * TC_TK = 16 * 16 = 256` elements, only elements 0–127 (rows 0–7) are loaded into shared memory. Rows 8–15 are never written, so the MMA reads uninitialised data for any M-dimension position ≥ 8.

### Impact

Any GEMM routed through `dense_gemm_tc` produces incorrect results for M-dimension positions ≥ 8. This affects any caller — we discovered it through the per-head attention gate (`g_proj`) during prefill, where multi-token quality degraded catastrophically beyond 8 tokens while single-token decode remained correct.

### Root cause

```cuda
// Line ~68 of dense_gemm_tc.cu — current code (broken)
unsigned int idx = tid;                      // tid = 0..127
if (idx < TC_TM * TC_TK) {                  // TC_TM*TC_TK = 256
    unsigned int r = idx / TC_TK;            // r = 0..7 only
    unsigned int c = idx % TC_TK;
    unsigned int gr = m_block + r;
    unsigned int gc = k_base + c;
    smem_A[r][c] = (gr < M && gc < K) ? A[gr * K + gc] : __float2bfloat16(0.0f);
}
```

128 threads execute one iteration — `idx` ranges 0..127, covering rows 0–7 (128/16 = 8 rows). Rows 8–15 are not loaded.

### Fix

Replace the single-iteration `if` with a strided loop:

```cuda
for (unsigned int idx = tid; idx < TC_TM * TC_TK; idx += TC_BLOCK) {
    unsigned int r = idx / TC_TK;
    unsigned int c = idx % TC_TK;
    unsigned int gr = m_block + r;
    unsigned int gc = k_base + c;
    smem_A[r][c] = (gr < M && gc < K) ? A[gr * K + gc] : __float2bfloat16(0.0f);
}
```

Each thread now loads 2 elements (iteration 0: elements 0–127, iteration 1: elements 128–255), covering all 16 rows.

Note: the B-tile load in the same kernel already uses a correct strided loop (`for (unsigned int i = tid; i < TC_TK * TC_TN; i += TC_BLOCK)`).

### Reproduction

1. Run any model that uses `dense_gemm_tc` in the prefill path with a prompt > 8 tokens.
2. Layer 0 hidden state comparison (Python HF reference vs Atlas) shows:
   - Positions 0–7: cosine similarity ≈ 0.9997 ✅
   - Positions 8+: cosine similarity ≈ 0.26 ❌ (Atlas RMS 12× larger than reference)
3. After fix: all positions ≈ 0.9997.

### Environment

- Hardware: DGX Spark (GB10 / SM121)
- Model: Step 3.7 Flash NVFP4, EP=2
- `dense_gemm_tc` used for g_proj (per-head attention gate, BF16 weights)

marksunner:
**Status update + resource pause**

Quick update on where this stands.

**What's working on the branch:**

The Step 3.7 Flash support is functional on EP=2 across two DGX Sparks, config parsing, weight loading, NCCL/RoCE init, and inference all running. Over the course of this investigation we found and fixed 10 issues: MoE transpose gating, ungated attention, sliding window detection, EP expert pointer table, per-layer RoPE rotary_dim, g_proj per-head attention gate, routed_scaling_factor, shared expert gate, shifted RMSNorm (`weight + 1`), and a cooperative tile load in `dense_gemm_tc`. The model produces coherent output in both chat and raw completion modes at ~21 tok/s decode.

**Two items that need attention before this can merge cleanly:**

1. **CLA attribution** — one commit is attributed to `Tars Turing` which isn't a GitHub account, so the CLA check is stuck at 6/7. This needs a rebase to squash or re-attribute that commit.

2. **`dense_gemm_tc` A-tile load** — this one is in the common kernels rather than the Step 3.7 target. The A-tile load uses a single-iteration `if (idx < 256)` with 128 threads, so only rows 0–7 of the 16×16 tile get loaded; the B-tile already uses the correct strided loop. My local testing showed positions 0–7 matching the HF reference at cos ≈ 0.9997, with positions 8+ diverging to cos ≈ 0.26, and the strided-loop fix restored all positions. Full details in my previous comment. Since this is in `common/`, I'm wondering whether PR #90 already addresses it - if so, that's fab and our finding independently validates the fix.

**Resource constraint:**

I've exhausted the compute budget I had available for this investigation and won't be able to keep iterating on the branch at the current pace. I will return to it as soon as I can. In the meantime, I wanted to leave everything in the clearest possible state, the diagnosis, reproduction steps, and fix are all documented above. I may be a tad slower to respond in the meantime (I'm on GMT), but happy to answer clarifying questions wherever I can.

Big thanks to @AzeezIsh and @tbraun96 for the engagement and the pointer to PR #90.

tbraun96:
Please change the merge target from `fix/in-think-tool-call-leak` to `main`. We merged `fix/in-think-tool-call-leak`

PR #124, opened 2026-06-05
fix(kernels): add vision_encoder module to qwen3.5-397b-a17b NVFP4 target

SeedSource:
## What
Adds `vision_encoder.cu` to `kernels/gb10/qwen3.5-397b-a17b/nvfp4/` by copying the existing implementation from the `qwen3.5-122b-a10b` NVFP4 target byte-for-byte. No kernel logic change, no `KERNEL.toml` change (the file stem auto-maps to module `vision_encoder`).

## Why
After #116, the 397B NVFP4 checkpoint loads past the prior `shared_expert` `weight_scale` failure. On our 4-node GB10 EP=4 run, all 11/11 shards loaded and NCCL initialized, but model build then failed at:

    Kernel lookup vision_encoder::vision_gemm_bias: Module 'vision_encoder' not loaded

The 397B checkpoint includes a vision config/weights, and the model builder constructs the vision encoder (`factory/build.rs` → `qwen35::load_vision_encoder`), so the target needs the existing `vision_encoder` module even for this text-serving path. The original #98 target was cloned from the text-only 35B target and omitted this module; the 122B VL target already ships it.

## Benchmarks
Not reported here — the current container launch is socket-bound (NCCL `NET/IB: No device found → Using network Socket`) rather than RoCE/IB-bound, so a throughput number would understate the engine. EP=4 throughput will be reported separately once container IB exposure is fixed.

## Verification
Tested on 4× DGX Spark GB10 (SM121), current main `7a903aa` (incl. #116 + #117) + this one-file change:
- Model `nvidia/Qwen3.5-397B-A17B-NVFP4`, TP=1 / EP=4 / world_size=4, kv-cache-dtype nvfp4, max_seq_len 4096, gpu_memory_utilization 0.75
- 11/11 shards load; vision encoder loads (depth=27); MTP weights load
- NCCL ring forms across 4 ranks (nRanks=4, Init COMPLETE)
- Server listens on :8888; `/v1/models` responds
- Deterministic probes (temp=0) coherent: `2+2 → 4`; `add(a,b)` → correct Python; ~1.2k-token long-context needle retrieved correctly

## Authorship
AI-generated using Claude Code (Opus 4.8).

### DIFF
--- a/kernels/gb10/qwen3.5-397b-a17b/nvfp4/vision_encoder.cu
+++ b/kernels/gb10/qwen3.5-397b-a17b/nvfp4/vision_encoder.cu
@@ -1,0 +1,313 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+// Atlas Spark — Qwen3-VL Vision Encoder CUDA Kernels
+//
+// All ops use BF16 storage for weights; computations use f32 accumulators.
+// Kernels run once per prefill (P ≤ 400 patches), so simplicity > performance.
+
+#include <cuda_bf16.h>
+#include <cuda_fp16.h>
+#include <math.h>
+
+// ── Helpers ──────────────────────────────────────────────────────────────────
+
+__device__ inline float bf16_to_f32(__nv_bfloat16 v) {
+    return __bfloat162float(v);
+}
+__device__ inline __nv_bfloat16 f32_to_bf16(float v) {
+    return __float2bfloat16(v);
+}
+
+// ── GEMM: C[M,N] = A[M,K] @ B[N,K]^T + bias[N]  (BF16 A, BF16 B, BF16 C) ──
+// Grid: (ceil(N/32), ceil(M/32), 1)  Block: (32, 32, 1)
+extern "C" __global__
+void vision_gemm_bias(
+    const __nv_bfloat16* __restrict__ A,   // [M, K]
+    const __nv_bfloat16* __restrict__ B,   // [N, K]  (transposed)
+    const __nv_bfloat16* __restrict__ bias,// [N]
+    __nv_bfloat16* __restrict__ C,         // [M, N]
+    unsigned int M, unsigned int N, unsigned int K
+) {
+    unsigned int row = blockIdx.y * 32 + threadIdx.y;
+    unsigned int col = blockIdx.x * 32 + threadIdx.x;
+    if (row >= M || col >= N) return;
+
+    float acc = 0.0f;
+    for (unsigned int k = 0; k < K; ++k) {
+        acc += bf16_to_f32(A[row * K + k]) * bf16_to_f32(B[col * K + k]);
+    }
+    acc += bf16_to_f32(bias[col]);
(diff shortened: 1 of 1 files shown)


SeedSource:
**What this PR is:** a single kernel file — `kernels/gb10/qwen3.5-397b-a17b/nvfp4/vision_encoder.cu` (+313 LoC), a byte-identical copy of the 122b NVFP4 `vision_encoder.cu`. It's a follow-up to #98, surfaced while verifying #116's loader fix end-to-end on our 4-node GB10 rig (the 397B builder loads `vision_encoder` unconditionally, so the target needs the module).

**On the 3 red checks** (`≤500 LoC`, `cargo fmt --check`, `cargo clippy --tests`): this PR changes **zero `.rs` files** — the diff is only the one `.cu`. All three checks scan `.rs`, so they can't be triggered by this change. They reproduce on `upstream/main` independently:

- **LoC cap:** `crates/spark-server/src/grammar/compile_tools.rs` (510) and `crates/spark-model/src/model/impl_a2.rs` (546) are over the 500 cap and not allow-listed — both present on `main` at those counts.
- **fmt:** `qwen35_dense.rs`, `prefill_a_step.rs`, `prefill_b_step.rs` need reformat (none in this diff).
- **clippy:** exits 101 on `main`; file-level detail deferred to avoid a local build contending with our running 397B serving.

Confirmed via local `cargo fmt` repro, the file-size math, and `git diff upstream/main...HEAD`.

**Path forward:** scope here is intentionally tight — one kernel add. The pre-existing `main` CI red is out of scope for this PR, but happy to file it as a separate cleanup if that's useful. Otherwise standing by for review on the kernel add.

AzeezIsh:
@SeedSource thanks for the tight scope and the thorough writeup, this is exactly the right shape for a target fix. Verified on our side (single GB10, dgx1):

1. Byte-identity confirmed against kernels/gb10/qwen3.5-122b-a10b/nvfp4/vision_encoder.cu via cmp.
2. Fresh build of the (gb10, qwen3.5-397b-a17b, nvfp4) target compiles 99 kernels and the generated PTX registry now contains the vision_encoder module, so the 'Module vision_encoder not loaded' lookup failure is closed.

We cannot load the 397B checkpoint on a single GB10 here (dgx2 is occupied), so your 4-node EP=4 verification (11/11 shards, NCCL ring, coherent temp=0 probes) stands as the end-to-end run. The three red checks you analyzed were pre-existing main breakage and have since been fixed there; CI on this PR is fully green now.

A separate cleanup issue for anything you still see red on main is welcome. Merging.

AzeezIsh (approved):
Byte-identity and 397b target kernel registration verified locally on GB10; end-to-end EP=4 verification by the author. LGTM.

SeedSource:
Thanks for the fast review and the verification on your side. The 4-node rig stays available for any multi-node EP runs you can't cover locally . We put the Step 3.7 branch through its paces on the same hardware overnight; findings incoming on #136. Glad the 397B target is home.

Will post some benchmarks in Discord this evening

PR #125, opened 2026-06-06
fix(tools): backfill a valid `subagent_type` so opencode's `task` tool works

pragmaxim:
## Problem

opencode / Claude-Code's `task` (subagent) delegation tool is unusable against Atlas. `task` declares `subagent_type` as a **required, free-form string** whose only legal values (`explore`, `general`, …) appear as prose in the tool description — there is no JSON-Schema `enum`. Qwen3.6-35B-A3B routinely omits the param, so Atlas's backfill inserts `""`, and the client rejects it:

```
Unknown agent type:  is not a valid agent type
```

The model retries the identical call and, after ~2 attempts, abandons delegation — from the user's seat, *"opencode can't use the task tool."*

## Root cause

In `crates/spark-server/src/tool_parser/validation.rs`:

- `backfill_required_params` inserts `""` for a missing required string (there is no `subagent_type` case among the auto-fill defaults).
- `validate_single_tool_call` deliberately accepts empty required strings (a carve-out originally for Theia's legitimate `path=""`), so the empty value passes validation and reaches the client.

Backfill thus turns a *self-correctable* "missing param" into an *unrecoverable* "empty param".

## Fix

`backfill_required_params` now fills a missing/empty `subagent_type` with a **valid** agent name parsed from the tool description (`- <name>:` lines under the "Available agent types" heading), preferring a general-purpose agent and falling back to `general`, instead of `""`. Behaviour for other required params is unchanged. Adds coverage in `tool_parser/tests/group_a.rs`.


pragmaxim:
@tbraun96 Hi, these PRs were needed for latest Opencode to actually perform a bit complex task on the Recipe A, without them it was utterly broken and prompt silently stopping.

- complex task  = find a bug in a smaller project based on error log

I started the docker container with `--dump` flag so agent could fix it based on good leads

AzeezIsh:
@pragmaxim this fix is right and we want it in. Your branch went conflicting against the recent tool_parser churn, so rather than bounce it back to you I ported it onto current main with your commit authorship preserved: #139. Logic is unchanged; the only delta is the regression tests land in a new tests/group_f.rs since group_a.rs is commented out of tests.rs (you caught the same dead-module problem in your follow-up commit). Verified on GB10, the two regressions run and pass, clippy clean.

Closing this in favor of #139. The diagnosis writeup was excellent, the backfill-turns-recoverable-into-unrecoverable framing made the fix obvious. Thanks for digging in.

PR #126, opened 2026-06-06
fix(loop-detect): stop spinning heuristic from masking <tool_call> during exploration

pragmaxim:
opencode / Claude-Code sessions intermittently **"silently die"** on **Atlas Recipe A latest** — the assistant turn ends mid-sentence with `finish_reason="length"` and no error. The root cause is **not** the prose-budget watchdog that shows up in the logs; that's just the executioner.

The real trigger is the **spinning detector** in `loop_detect.rs`: it misclassifies normal file-by-file exploration (5+ short `read`/`glob` calls in a row) as a degenerate loop and **hard-masks the `<tool_call>` token** for the turn. The model then can't call a tool, has no reason to emit EOS, narrates until the inter-tool prose budget trips, and gets truncated.

This PR makes the spinning signal **advisory** (it still injects the soft hint + goal re-anchor). Hard `<tool_call>` suppression now requires the **content-similarity** verdict (`LoopState::Suppress`) — i.e. *measured* repetition, not mere turn brevity.

## What the user sees

A turn dies with no output to act on. In the server log it looks like this (one request):

```
WARN  loop_detect: Spinning detection fired — suppressing <tool_call> recent_short=5
...
WARN  handle_token: SimHash semantic-loop watchdog fired (paraphrased sentence repeat) ring_len=3
WARN  decode_logits_content: Inter-tool prose budget exhausted, ending response
      (rollback declined) prose_tokens=385 max=384 reason=NoSsmSnapshot
INFO  lifecycle: Done: 445 tokens (length) 22.6 tok/s
```

## Root cause (the chain)

1. opencode explores by reading files **one at a time** → a run of "short" assistant turns (`content < 500 chars` **and** `tool_args < 100 chars`).
2. `loop_detect.rs` flags `spinning` at `recent_short >= 5`. **Five distinct file reads are progress, not a loop** — false positive.
3. `spinning` set `suppress_tool_call = true`, which applies a **−12.0 logit bias** to the `<tool_call>` start token in `decode_logits_seq.rs` — effectively banning tool calls for the turn.
4. The model still wants to read files. It narrates *"Let me read X / I'll read Y"* but can't emit the tool call, and it isn't trying to *finish*, so it never samples EOS.
5. Redundant prose accumulates → inter-tool prose budget (384) trips → the Phase-C rollback finds no SSM snapshot at a boundary (`NoSsmSnapshot`) → hard stop, `finish_reason=length`, truncated mid-sentence. To the client, the turn just dies.

## Evidence

- **Log:** the `Spinning detection fired … recent_short=5` line is immediately followed (same request) by `Inter-tool prose budget exhausted … reason=NoSsmSnapshot` and `Done: 445 tokens (length)`.
- **`--dump`:** the killed response was preceded by exactly five short, **distinct** turns — `read(98) ← read(89) ← read(79) ← glob(45) ← read(47)` — then a substantial `task(1607)`. The truncated content was a genuine, useful analysis cut off mid-sentence (*"…which runs once per request but can be"*), not gibberish.
- The content-similarity detector returned **neither `Suppress` nor `Hint`** for those distinct reads (no `Loop detector → SUPPRESS/HINT` line) — only the size-based spinning heuristic fired. So the fatal suppression came purely from `spinning`.

## The fix

`crates/spark-server/src/api/chat/loop_detect.rs`:

- `spinning` **no longer sets `suppress_tool_call`.** It still drives the soft hint + goal re-anchor (unchanged steering).
- Hard `<tool_call>` masking stays gated on `LoopState::Suppress` from the content-similarity detector, which measures *actual* repetition.
- Extracted the run-length scan into a pure, allocation-free `leading_short_turn_run` helper (unit-tested).

This aligns with the detector's **own documented philosophy** — the `LoopState::Hint` variant explicitly says moderate repetition should *not* hard-suppress tools "because the model often needs verification tools to escape", which is exactly what got starved here.

## Why it's safe

- Genuine loops (same tool + similar args, or repeated prose) still escalate to `Suppress` via the unchanged content-similarity path — see existing `loop_detector` tests (`identical_tool_args_fire_loop`, `moderate_similarity_3_turns_now_suppress`, …).
- The only behavior removed is hard tool-call masking on the *size-only* signal. The hint/re-anchor that spinning injects is unchanged.

## Relationship to #125

Independent of #125 (`backfill subagent_type` so opencode's `task` works). The spinning trigger is 5 consecutive short *tool calls*, which is the universal agentic exploration pattern — **any** agent that reads 5+ files in a row trips it. #125 simply made opencode functional enough to reach this failure, which is how it was caught.

## Testing

- Added unit tests for `leading_short_turn_run` (the opencode failure shape → run of 5; substantial turn stops the run; empty history).
- Added a `loop_detector` regression: distinct file reads must **not** escalate to `Suppress`.
- `cargo check -p spark-server --tests` is clean (0 warnings).
- ⚠️ I could not **run** the suite on my dev host (no `nvcc` / `libnccl`, so the `spark-server` bin test target won't link). Please run `cargo test -p spark-server` on a CUDA box to confirm.

## Follow-up (out of scope)

The inter-tool prose-budget + `NoSsmSnapshot` rollback is the *executioner* in this chain and remains in place. Its premise — that the `tool_choice="auto"` grammar "can never self-terminate" — is empirically false: the stop token is legal at the free-text root and accepting it terminates the matcher (verified against `xgrammar`). Worth revisiting separately, as it can also truncate legitimate long-form text answers (> 384 tokens) whenever tools are active.

### DIFF
--- a/crates/spark-server/src/api/chat/loop_detect.rs
+++ b/crates/spark-server/src/api/chat/loop_detect.rs
@@ -13,13 +13,61 @@ use crate::openai::ChatCompletionRequest;
 use super::msg_entry::MsgEntry;
 
 pub(super) struct LoopDetectOut {
-    /// True when the verdict was Suppress OR spinning detection
-    /// fired. Caller flips the `<tool_call>` token bias to avoid
-    /// re-emitting tool calls for one turn.
+    /// True ONLY when the content-similarity detector returns
+    /// [`crate::loop_detector::LoopState::Suppress`] — i.e. strong,
+    /// *measured* repetition across recent turns. The caller flips the
+    /// `<tool_call>` token bias to break a genuine loop.
+    ///
+    /// The size-based "spinning" signal deliberately does NOT set this.
+    /// A run of short *distinct* tool calls (reading files one at a
+    /// time) is legitimate agentic exploration, not a loop; masking
+    /// `<tool_call>` there strands the model — it narrates "let me read
+    /// X" with the tool token biased to -inf, never progresses, never
+    /// emits EOS, and is finally truncated mid-sentence by the
+    /// inter-tool prose budget (the opencode "silent death").
     pub(super) suppress_tool_call: bool,
     /// Run-length of the most recent loop (or 0). Caller threads
     /// this into the exponential `<tool_call>` logit-bias decay.
     pub(super) tool_call_repeat_count: usize,
+}
+
+/// A turn is "short" (low-signal) when it carries little prose AND
+/// little tool-call payload. These bounds are deliberately generous: a
+/// single `read`/`grep`/`glob` call — the bread-and-butter of agentic
+/// exploration — is short, so a run of them is normal progress, not a
+/// loop. See [`leading_short_turn_run`].
+const SHORT_TURN_CONTENT_MAX: usize = 500;
+const SHORT_TURN_TOOL_ARGS_MAX: usize = 100;
+
+/// `recent_short >= this` raises the (weak) size-based spinning signal.
+const SPINNING_RUN_THRESHOLD: usize = 5;
+
+/// How far back the spinning scan walks — it only needs to learn
+/// whether the leading run reaches [`SPINNING_RUN_THRESHOLD`].
+const SPINNING_SCAN_CAP: usize = 8;
+
+/// Length of the leading run of consecutive "short" assistant turns,
+/// newest-first, stopping at the first substantial turn. Pure and
+/// allocation-free so it is unit-testable without an openai request.
+///
+/// Each item is `(content_len, tool_args_len)` for one assistant turn.
+fn leading_short_turn_run<I>(assistant_turns_newest_first: I) -> usize
+where
+    I: IntoIterator<Item = (usize, usize)>,
+{
+    let mut run = 0usize;
+    for (content_len, tool_args_len) in assistant_turns_newest_first {
+        let substantial =
+            content_len >= SHORT_TURN_CONTENT_MAX || tool_args_len >= SHORT_TURN_TOOL_ARGS_MAX;
+        if substantial {
+            break;
+        }
+        run += 1;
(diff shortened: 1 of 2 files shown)


pragmaxim:
benches https://gist.github.com/pragmaxim/dd4f88eab4fa1eed862c3ce46526a8ef

tbraun96:
Please contact me on Discord @nologik
Server: https://discord.com/invite/6vDbKaKrKD

AzeezIsh:
@pragmaxim good news on this one: the exact failure you traced is already fixed on current main, by a different cut. Since 2026-06-02 (BW1/SPINFIX relaxation, Iter 52) the spinning scan treats any turn that issued a tool call as substantial: a run of short read/glob/bash calls no longer counts toward recent_short at all, so your five-distinct-file-reads pattern cannot trip it anymore. The spinning signal now only fires on five consecutive short pure-text turns with no action, and there is an ATLAS_LOOP_NO_SUPPRESS=1 escape hatch that keeps the verdict logged and metered without the hard mask.

Where main differs from this PR: for the pure-text spinning case we kept the hard mask, since a model narrating short text turns without acting is the one shape where forcing it off tools has been net positive in our suites. Your repro chain (read run -> spinning -> tool mask -> narration -> prose budget -> NoSsmSnapshot truncation) cannot start from a tool-call run on main.

Please pull the latest Recipe A image and re-run your opencode task. If you can still produce a silently-dying turn on current main, reopen with the log and dump and I will treat it as a live bug. Also, your follow-up note that the tool_choice=auto grammar can in fact self-terminate at the free-text root is worth its own issue; if you file it with the xgrammar verification you mentioned, I will pick it up.

Closing as superseded by the SPINFIX relaxation on main. The root-cause chain in this writeup was first rate, this is exactly the kind of report that makes fixes easy.

PR #129, opened 2026-06-08
DFlash Support

tbraun96:


tbraun96:
Please rebase to main, no longer `fix/in-think-tool-call-leak` since we just merged that in

PR #132, opened 2026-06-09
perf(dflash): fuse decode+verify into one weight sweep + reclaim Option-B KV blocks

rrstesiak:
**The bug.** DFlash K3 accept-rate decayed monotonically with context (0.55 → 0.12 → 0.00, 8.8 tok/s) after the verify-pipeline merge. The merge routed DFlash verify token-selection through `verify_pick_all_with_pipeline` (rep_pen=1.1 + dry_mult=0.5), but the DFlash drafter proposes on **raw argmax**. Verifier and drafter judged on different bases, so divergence grew as context tokens accrued penalties — collapsing accept to zero.

**Ruled out.** Not the SSM rollback ring — that is dead code in the DFlash K3 path (`emit_token` never calls `snapshot_boundary_if_ssm` / `rollback_to_boundary`, ring stays empty). Confirmed by static read + a disable-rollback bench (both flat).

**The fix.** Thread a process-wide `dflash_verify_raw_argmax` flag (= `args.dflash`) from `serve` → `scheduler::run` → `step_mtp` into the four verify steps (k2/k3/k4/dflash). In DFlash mode, verify on raw argmax (GOLD basis, matching the drafter); otherwise the full MTP pre-sample pipeline runs unchanged. **MTP verify path is untouched.** Grammar boundary truncation already runs upstream of dispatch, so raw-argmax verify does not affect structured output.

**Verified.** γ=3 / K3 path: accept holds **0.44–0.85, no monotonic decay** over seq_len 232→884, **13.9 tok/s** recovered (vs 8.8 tok/s broken).

Repro: `bash ~/launch-dflash-gamma3-gatefix.sh` (γ=3, eager, Option-B, VERIFY_TRACE on).

_Note for reviewers: base is `main` for a self-contained, buildable unit; retarget to an integration branch if that fits the workflow better._


rrstesiak:

rrstesiak:
recheck

AzeezIsh:
@rrstesiak thanks for this, the diagnosis is correct and well isolated. I verified it against current main by code read: verify_dflash_step.rs routes accept comparison through verify_pick_all_with_pipeline (rep_pen 1.1 + DRY), while the DFlash drafter proposes on raw argmax, so the accept basis diverges as context accrues penalties. Your accept-rate decay curve matches that mechanism exactly, and ruling out the SSM rollback ring with the static read plus disable bench was good methodology.

Two things before this can land:

1. The branch needs a re-cut. The diff currently carries 1790 files, including bench/fp8_dgx2_drift working artifacts, report markdown, dump scripts, a backup file, and broad fork divergence across the whole tree. Please cut a fresh branch off current main with only the fix: the flag in cli.rs, the threading through scheduler into the four verify steps, and the raw-argmax branch. From your own description that should be a few hundred lines at most. It also resolves the merge conflicts in one shot.

2. One design question to address in the PR description or a follow-up: with dflash_verify_raw_argmax, every accepted token bypasses rep_pen and DRY entirely for the whole DFlash stream, not just the comparison. Those penalties exist because several of our models loop without them in long agentic sessions. An alternative that keeps the bases matched without losing the protection is to compare draft vs raw argmax for the accept prefix (your fix) but still run the pipeline on the bonus token, or to feed the drafter the same penalized basis. If you have long-form repetition data under raw-argmax DFlash, post it; if the loop risk is real we may want the bonus-token variant.

If you would rather not redo the branch surgery, say so and I will extract the minimal fix onto main with you as commit author. Either way this is a real find, appreciate the rigor.

tbraun96:
@rrstesiak following up on the june 10 review. the raw-argmax gate diagnosis was right, and it's now carried by #218 rewritten against current main, with credit to this PR's root-cause. so the gate itself lands through that one.

what's still uniquely valuable here is your free_state fix. main's BlockDiffusionDraftHead::free_state is still the no-op stub, so the Option-B drafter block leak you found is live, and #218 doesn't touch it. can you cut a fresh branch off current main with just that fix? note main has since split dflash_head.rs into dflash_head/ submodules, so it lands in a different spot. while you're there, free the block_table_dev handle too instead of leaking ~1KB per request — thread the gpu handle in or stash it on the state.

the fused decode+verify piece needs a rethink before it's reviewable: on current main --dflash defaults to gamma=16 which routes through step_verify_dflash (already single-sweep), and your bootstrap reroute sends init to step_verify_k3 even when the drafter returned 15 drafts — only 2 get verified while trim_proposer_state runs on K3 semantics. if you still measure a win at gamma=3 after rebasing, send it as its own PR with fresh numbers.

also the rep_pen/DRY bypass question from june 10 is still open — if you have long-form generation data under raw-argmax dflash showing no repetition loops, post it on #218 since that's where the gate now lives.

as-is this branch is 264 commits off a june 3 base with 108 conflicting files against main, so it can't merge in any form. happy to review a re-cut quickly — the leak fix alone is a solid contribution.

rrstesiak:
Thanks for the fast review, Tom. Re-cut is up: #225.
 
One finding from the port: the live leak on current main is actually the per-sequence ctx_hidden_acc buffer, not the Option-B blocks: main's propose() doesn't allocate paged drafter blocks yet, 
so there was nothing to reclaim there. But free_state was still the Phase-1 stub, so every --dflash request leaked the accumulator: measured ~241 MiB/request on GB10 (Qwen3.6-27B, --max-seq-len 4096; ~8× that at the 32768 default). Raw before/after readings and methodology are in #225. Block reclaim is included too, so the leak class stays closed when a paged drafter path lands.
  
Agreed on the fused decode+verify critique. if a rebase against step_verify_dflash still shows a win at low gamma, I'll revisit it separately.

SeedSource:

### Triage
- Rationale: No code changes; only debug logs, markdown findings, and project metadata files.
- Serving-path files: `bench/hotfix3-debug/server.log, crates/atlas-kernels/build.rs, crates/atlas-kernels/build_codegen.rs, crates/atlas-kernels/build_parse.rs, crates/atlas-kernels/build_target.rs, crates/atlas-kernels/src/lib.rs`

### Serve gate
- Build: 29s - Qwen3.6-35B-A3B-FP8 on GB10 node1
- Top-k KL vs main: **0.1106** - **large behavioral shift** - decode delta **-5.9%**

### GLM-5.2 code review
# Review: PR #132

**Cannot review — empty diff.**

The provided diff contains no code changes (only the `<<<` / `>>>` markers with nothing between them). There is nothing to evaluate for correctness, risk, or quality.

**Action required from author:**
- Re-upload the actual diff/patch content.
- Ensure the diff includes the fused decode+verify loop, the KV block reclamation logic, and any changes to memory management or kernel launches.

**When diff is available, I will assess:**
1. **Correctness:** Whether fusing decode+verify preserves numerical results; whether KV block reclamation races with in-flight reads.
2. **Risk:** Memory safety in unsafe Rust/CUDA FFI; lifetime of reclaimed Option-B blocks; potential use-after-free if blocks are returned to a pool while still referenced by outstanding async ops.
3. **Quality:** Lock contention on the block pool; whether the fusion introduces divergent control flow in the CUDA kernel; test coverage for the reclamation path.

Resubmit with the diff content.

tbraun96:
@rrstesiak I synced this against current `main` (tip 288c17c1) and the divergence has gotten heavy, so here is a grounded status to pick the fastest path.

- This branch is **60 behind / 264 ahead** of main off a **June-10 mergebase (7a903aa)**. A straight `merge main` into it hits **128 conflicted files**, almost all from the module refactor that landed since (`qwen3_attention/prefill/*`, `moe/forward_*`, `dflash_head` split into submodules, `qwen3_ssm` split).
- The **Option-B KV-block reclaim** looks already landed on main via **#225** (`free_state`, reclaim drafter KV blocks), so that half of this PR reads as largely superseded.
- The piece still worth landing is the **fuse decode+verify into one weight sweep**. That is the real perf idea here.

Given a 128-conflict surface off a pre-refactor base, replaying the full 264-commit stack is rough. Cleaner path: lift just the fused decode+verify sweep onto current main, where the decode/verify path now lives in the refactored `dflash_head` + `qwen3_ssm` modules. #218 already carries the 2.9 to 22 tok/s verify work on a near-current base (5 behind main), so that is the natural place to stack the fused sweep.

Want me to open a fresh branch off main and port the fused decode+verify sweep there with you credited, or would you rather rebase this branch yourself. Either way I will bench it on the agentic 2.5h decode path and report numbers. Confirm the fused sweep is the unique delta not already in #218 and I will get it landed.

tbraun96:
Follow-up after actually porting the fused sweep onto current main, @rrstesiak.

Turns out the fuse decode+verify idea is **already on main**, not just the KV reclaim. The refactored K=γ verify path (`decode_verify_graphed_kgamma_dispatch`) already runs a single M=(1+γ) weight sweep with `tokens=[accepted_token, draft_0..draft_{γ-1}]`, so row 0 is the accepted-token decode and rows 1..γ are the draft verify, one weight read for both. There is no separate M=1 decode sweep left to eliminate. I re-expressed `verify_fused.rs` on this branch behind an `ATLAS_DFLASH_FUSE_VERIFY` toggle and it comes out behaviorally equivalent to the default path (A/B delta ~0), which confirms it.

So both halves of this PR have landed on main through the refactor lineage. The KV reclaim came in via #225, and the fused sweep is now the default K=γ path (the #209 → #258 → #218 stack). I do not see unique unmerged perf left here.

Recommend we **close this in favor of #218**, which carries the grounded 22 tok/s work (conv fusion + wy16 + K=16 kgamma) on a near-current base and merges to main with only 17 conflicts. I already resolved that merge and it builds clean. Your call on closing, and thanks for the original fusion push, it is the shape the K=γ path took.

rrstesiak:
> Follow-up after actually porting the fused sweep onto current main, @rrstesiak.
> 
> Turns out the fuse decode+verify idea is **already on main**, not just the KV reclaim. The refactored K=γ verify path (`decode_verify_graphed_kgamma_dispatch`) already runs a single M=(1+γ) weight sweep with `tokens=[accepted_token, draft_0..draft_{γ-1}]`, so row 0 is the accepted-token decode and rows 1..γ are the draft verify, one weight read for both. There is no separate M=1 decode sweep left to eliminate. I re-expressed `verify_fused.rs` on this branch behind an `ATLAS_DFLASH_FUSE_VERIFY` toggle and it comes out behaviorally equivalent to the default path (A/B delta ~0), which confirms it.
> 
> So both halves of this PR have landed on main through the refactor lineage. The KV reclaim came in via #225, and the fused sweep is now the default K=γ path (the #209 → #258 → #218 stack). I do not see unique unmerged perf left here.
> 
> Recommend we **close this in favor of #218**, which carries the grounded 22 tok/s work (conv fusion + wy16 + K=16 kgamma) on a near-current base and merges to main with only 17 conflicts. I already resolved that merge and it builds clean. Your call on closing, and thanks for the original fusion push, it is the shape the K=γ path took.

I already refactored from the then-current main days ago per your recommendations.. I am seeing the best of the team's work integrated here.. 

I approached my refactor in 2 stages: 

1. Merge the best of my previous code into team's main.
2. Apply wide gamma ACCPET and PERF TUNING

My local changes do improve ACCEPT (matches vLLM wide gamma accept) and performance *at wide  gamma*... but even all of those changes do NOT break curerent performance ceiling... that ceiling  (~15-~17 tok/s I measured locally) being only stubborn thing that has been holding me back from releasing code.  I truly  see no benefit in pushing *anything* until performance is improved. 

*****NOTE Can you just hold on any major decisions about code until I can get in to verify exactly which branch I already merged?

tbraun96:
@rrstesiak this is the useful part — I hit your exact ceiling and I think I found what breaks it. Two things, both grounded on dgx1 (27B, z-lab drafter, greedy).

**1. Capture-layer offset is mis-conditioning the drafter.** `factory/build.rs` subtracts 1 from the drafter's `target_layer_ids` (HF output_hidden_states semantics), but the z-lab 27B drafter declares raw ids `[1,16,31,46,61]` and conditions best on them directly. A/B: `CAPTURE_LAYER_OFFSET=-1` (default) 3.17 accepted / 15.6 tok/s, `=0` 3.55 / 17.2, `=+1` 3.27 / 15.4. So `=0` is +10%.

**2. The ceiling is the NVFP4 target's noisy argmax vs the BF16-trained drafter — not kernels.** KL over 2240 verify positions: the target is confident (p_top1>=0.9) on 72% of positions but the drafter only lands ~24%, so it whiffs on easy tokens. Sharpen the target argmax and acceptance climbs monotonically:
- NVFP4+offset0 = 17.2 (accept 4.55)
- serve `Qwen3.6-27B-FP8` + `--lm-head-dtype bf16` = 18.2 (5.22)
- + `--kv-cache-dtype bf16` = **19.0 (5.35)**, +36% over the 14.0 plain baseline

FP8 target is not 2x slower — verify re-quantizes to nvfp4 for the kernels, only the captured hidden gets the sharpness. That matches your wide-gamma accept improving but tok/s staying pinned — accept alone does not move it, the argmax precision does.

**Path past 20:** the BF16 base target (your drafter's exact training distribution) should push accept to the ~6.67 vLLM regime. I could not finish the 54GB pull here (4/15 shards keep failing) so that last step is projected, not measured. Durable version is FP8-native 27B kernels so we get the sharp argmax without the download or the BF16 forward cost.

And absolutely — holding on any close/merge/code decisions until you confirm which branch you already merged. Just wanted the ceiling-breaker in your hands since it is the thing you flagged. Nice call on the ctx_hidden_acc leak, that framing is clearer than mine.

tbraun96:
@rrstesiak one more thing that should save you time, and it needs zero code changes on your branch, so it does not touch your hold.

The 15 to 17 ceiling breaking to 19 was all serve flags plus env, not code. On whatever branch you already have, try serving the FP8 target instead of NVFP4 with a bf16 head and bf16 KV, and set the capture offset to 0:

`serve Qwen/Qwen3.6-27B-FP8 --dflash --draft-model z-lab/Qwen3.6-27B-DFlash --dflash-gamma 16 --lm-head-dtype bf16 --kv-cache-dtype bf16` plus `ATLAS_DFLASH_CAPTURE_LAYER_OFFSET=0 ATLAS_DFLASH_EAGLE_FIX=1 ATLAS_DFLASH_CONV_FUSION=1 ATLAS_VERIFY_PREFILL_SSM=1 ATLAS_VERIFY_PREFILL_ATTN=1`.

I went 15.6 to 19.0 doing exactly that on the as-is branch binary. Your wide-gamma accept tuning should stack on top of this, not compete with it. You raised how many drafts get accepted, this raises how often the drafter matches the target argmax at all, which is the part that was pinned.

That is why your accept could match vLLM while tok/s stayed flat. On NVFP4 the target argmax is noisy, so a drafter that predicts the true distribution still misses the served argmax. KL over 2240 positions showed the target confident on 72 percent but the drafter landing 24, and the offset default of minus 1 was mis-conditioning it on top of that. Sharpen the target argmax and the accept you already have starts translating into tokens.

If your branch shows the same jump, we have the shared unblock and you can release. Still holding on any push or merge decision until you confirm your branch.

PR #133, opened 2026-06-09
feat(tq+): TurboQuant+ KV cache — canonical Hadamard + 9 asymmetric variants + Turbo2 + dispatch tests

### LINKED ISSUE #91: feat(kv-cache): TurboQuant+ — Randomized Hadamard rotation + Lloyd-Max codebook for KV cache (proposal)

TheTom:
> **Rebased replacement for #92.** That PR was opened against `fix/in-think-tool-call-leak` (#90), which has since merged to `main` and had its branch deleted — auto-closing #92. This PR is the same `feature/tq-plus-clean` branch rebased cleanly onto `main` (union-merged with #90's determinism + `op_dump` changes; identical TQ+ delta).

---

## Summary

Ports the TurboQuant+ (TQ+) KV cache compression line into Atlas:
canonical Randomized Hadamard rotation, Lloyd-Max codebooks with
matched-norm L2 correction, sparse-V dequant, the 2-bit `Turbo2` cache
that was crashing on upstream, and **9 asymmetric `KvCacheDtype`
variants** with per-side K/V cache pools + combined write+decode+prefill
kernels. Reference: Zandieh, Daliri, Hadian, Mirrokni —
*TurboQuant: Online Vector Quantization with Near-optimal Distortion
Rate* ([arXiv:2504.19874](https://arxiv.org/abs/2504.19874), April 2025).
TQ+ extensions from [`TheTom/turboquant_plus`](https://github.com/TheTom/turboquant_plus)
and [`TheTom/llama-cpp-turboquant`](https://github.com/TheTom/llama-cpp-turboquant).

The headline numbers, on Qwen3.6-35B-FP8 (Bench F) and
Qwen3-VL-30B-A3B-NVFP4 (Bench G), single-GPU GB10, greedy decode,
fixed Manhattan-Project WikiText prompt, 5-run median via
`tests/atlas_bench_comprehensive.py`:

  - **`fp8k_turbo3v` and `turbo4k_turbo3v`**: PPL sim **0.8485** —
    bit-identical to the fp8 baseline. The "K kept at baseline
    precision" promise empirically delivered.
  - **`bf16k_turbo3v` and `bf16k_turbo4v`**: PPL sim **0.2929** —
    bit-identical to the sym bf16 baseline on the same NVFP4 model.
  - **`Turbo2` (2-bit, new dtype)**: 2-bit Lloyd-Max codebook added
    to the `KvCacheDtype` enum (upstream `main` has only Turbo3/4/8).
    End-to-end at 1350 tok/s 8K prefill — fastest of any KV dtype
    tested, less data to write per token — at the expected 2-bit
    quality penalty (PPL sim 0.6465 vs 0.8485 fp8 baseline). Primary
    use case is as the V-side of asym combos like `Bf16KTurbo2V` /
    `Fp8KTurbo2V`.
  - **Dispatch-routing unit tests** that catch the silent
    fall-through bug class at compile time + test time, not runtime.

The symmetric path is at parity with upstream at greedy decode on
Qwen3.6-A3B (byte-identical output on fp8/turbo3/turbo4). The PR's
value is **new capability + correctness fixes + dispatcher safety
nets**, not a headline PPL win on the existing symmetric dtypes.

Closes #91.

## KV cache compression at a glance

Lower `bits/elem` = smaller KV pool = more context window at same VRAM
ceiling. Includes per-group FP8 scale amortised across `GROUP_SIZE=16`
elements (turbo8 uses BF16 scale instead).

**Symmetric dtypes**

| dtype  | bits/elem | vs bf16 | vs fp8 |
|--------|----------:|--------:|-------:|
| bf16   | 16.0      | 1.0×    | 0.5×   |
| fp8    | 8.0       | 2.0×    | 1.0×   |
| turbo8 | 9.0       | 1.78×   | 0.89×  |
| nvfp4 / turbo4 | 4.5 | 3.56× | 1.78× |
| **turbo3** | **3.5** | **4.57×** | **2.29×** |
| **turbo2 (new)** | **2.5** | **6.4×** | **3.2×** |

**Asymmetric (K-side + V-side averaged)** — the production-recommended
rows are the ones that hit baseline-parity quality at the smallest
KV footprint:

| dtype             | K b/e + V b/e | avg b/e | vs sym K-side | PPL parity |
|-------------------|--------------:|--------:|--------------:|:-----------|
| bf16k_turbo3v     | 16 + 3.5      | 9.75    | 1.64× smaller | **= sym bf16** (0.2929) |
| bf16k_turbo4v     | 16 + 4.5      | 10.25   | 1.56× smaller | **= sym bf16** (0.2929) |
| bf16k_turbo2v     | 16 + 2.5      | 9.25    | 1.73× smaller | -0.01 vs bf16 |
| fp8k_turbo3v      | 8  + 3.5      | 5.75    | 1.39× smaller | **= sym fp8**  (0.8485) |
| fp8k_turbo4v      | 8  + 4.5      | 6.25    | 1.28× smaller | = sym turbo4 (0.8384) |
| fp8k_turbo2v      | 8  + 2.5      | 5.25    | 1.52× smaller | -0.22 vs fp8 (2-bit V) |
| **turbo4k_turbo3v** | 4.5 + 3.5   | **4.0** | **2.0× smaller** than fp8 | **= sym fp8** (0.8485) ⭐ |
| turbo4k_turbo8v   | 4.5 + 9       | 6.75    | mixed         | -0.20 vs fp8 |
| turbo3k_turbo8v   | 3.5 + 9       | 6.25    | mixed         | -0.16 vs fp8 |

⭐ **`turbo4k_turbo3v` is the headline frontier**: 2.0× smaller KV pool
than sym fp8 at byte-identical PPL sim on the Manhattan-Project
continuation. Halving the KV footprint at zero quality cost on this
prompt means ~2× longer context at the same VRAM ceiling.

> **Asymmetric K/V quality is model-family sensitive.** Which asym
> combo wins on quality vs compression depends on the model's
> attention-score distribution + K-side outlier mass profile. The
> recommended starting points above are based on Qwen3.6 + Qwen3-VL
> measurements; other model families may have a different sweet
> spot. The design rationale + per-family selection guidance is in
> [`TheTom/turboquant_plus/docs/papers/asymmetric-kv-compression.md`](https://github.com/TheTom/turboquant_plus/blob/main/docs/papers/asymmetric-kv-compression.md).

## Bench results

All numbers from `tests/atlas_bench_comprehensive.py` on single-GPU
GB10, greedy decode, fixed Manhattan-Project WikiText prompt, 5-run
median per metric, 3 context lengths.

### Bench D — sym matrix, upstream `87b7bb3` baseline vs TQ+ branch (Qwen3.6-35B-FP8)

Output text byte-identical between upstream and TQ+ on fp8/turbo3/turbo4
at greedy decode. Throughput at parity (±2% noise).

| dtype  | b/e  | PPL sim (base → TQ+) | pre_8K t/s (base → TQ+) | pre_16K t/s (base → TQ+) | dec_after_8K (base → TQ+) |
|--------|-----:|---------------------:|------------------------:|-------------------------:|--------------------------:|
| bf16   | 16.0 | 0.8485 → 0.8485      | 1229.35 → 1227.54       | 1394.17 → 1390.46        | 40.94 → 40.87             |
| fp8    | 8.0  | 0.8485 → 0.8485      | 1219.05 → 1239.06       | 1377.74 → 1402.00        | 41.28 → 41.65             |
| turbo8 | 9.0  | 0.8384 → 0.8384      | 1223.43 → 1248.85       | 1389.64 → 1416.28        | 40.95 → 41.11             |
| nvfp4  | 4.5  | 0.8485 → 0.8485      | 1208.02 → 1239.39       | 1366.80 → 1410.56        | 41.07 → 41.53             |
| turbo4 | 4.5  | 0.8384 → 0.8384      | 1252.60 → 1238.62       | 1417.42 → 1403.65        | 41.22 → 40.94             |
| turbo3 | **3.5** | 0.8384 → 0.8384   | 1249.94 → 1230.51       | 1416.13 → 1400.57        | 41.28 → 40.85             |
| **turbo2** | **2.5** | not in upstream enum → **0.6465** | n/a → **1336.50** | n/a → **1519.93** | n/a → 41.56 |

**Turbo2 is new in this PR** — upstream `87b7bb3`'s `KvCacheDtype`
enum has `Bf16, Fp8, Nvfp4, Turbo4, Turbo3, Turbo8` only. This PR adds
the 2-bit Lloyd-Max codebook (`Turbo2`) + its write/decode/prefill
kernels + dispatcher arms. The 2-bit codebook is the most aggressive
KV compression we ship (2.5 b/elem incl scale vs Turbo3's 3.5 b/elem
vs Turbo8's 9 b/elem); usable for low-bit asym variants like
`Bf16KTurbo2V` / `Fp8KTurbo2V` where K precision absorbs the V quality
hit. Prefill is fastest of any KV dtype tested (less data to write
per token) at the expected 2-bit quality penalty.

### Bench F — 9 asymmetric variants on Qwen3.6-35B-FP8 (head_dim=256)

| dtype             | avg b/e | PPL sim | dec_short | pre_2K  | pre_8K  | pre_16K | dec_after_8K |
|-------------------|--------:|--------:|----------:|--------:|--------:|--------:|-------------:|
| bf16k_turbo3v     | 9.75    | load_timeout — needs bf16-attn model (see Bench G)              |
| bf16k_turbo4v     | 10.25   | load_timeout                                                    |
| bf16k_turbo2v     | 9.25    | load_timeout                                                    |
| **fp8k_turbo3v**  | 5.75    | **0.8485** | 72.30  | 641.19  | 1249.05 | 1415.89 | 41.28        |
| fp8k_turbo4v      | 6.25    | 0.8384  | 71.98     | 636.42  | 1240.02 | 1403.23 | 41.03        |
| fp8k_turbo2v      | 5.25    | 0.6263  | 72.16     | 631.95  | 1232.50 | 1390.13 | 40.97        |
| **turbo4k_turbo3v** | **4.0** | **0.8485** | 72.01 | 638.85  | 1242.65 | 1410.52 | 41.04        |
| turbo4k_turbo8v   | 6.75    | 0.6465  | 71.96     | 638.85  | 1241.52 | 1402.76 | 40.99        |
| turbo3k_turbo8v   | 6.25    | 0.6869  | 71.92     | 634.67  | 1234.47 | 1398.72 | 40.72        |

`fp8k_turbo3v` and `turbo4k_turbo3v` are bit-identical to the fp8
baseline (0.8485) — the "K kept at baseline precision" promise
empirically delivered. Throughput on every asym variant matches the
symmetric line within ±1% across all 3 context lengths.

**Compression headline at fp8-parity quality**:
- `fp8k_turbo3v` (5.75 b/elem) is **1.39× smaller** than sym fp8 (8.0 b/elem)
- `turbo4k_turbo3v` (4.0 b/elem) is **2.0× smaller** than sym fp8 — same
  PPL sim 0.8485, half the KV pool footprint, ~2× longer context at same
  VRAM ceiling. This is the production-recommended row.

### Bench G — bf16k_* variants on Qwen3-VL-30B-A3B-NVFP4 (head_dim=128)

The 3 bf16k_* variants exercise the bf16 K-side at HDIM=128 — only
HDIM=128 bf16-attn model in Atlas's tested set. Compared against sym
`bf16` baseline on the same NVFP4 weight release.

| dtype             | avg b/e | PPL sim | dec_short | pre_2K  | pre_8K  | pre_16K | dec_after_8K |
|-------------------|--------:|--------:|----------:|--------:|--------:|--------:|-------------:|
| bf16 (baseline)   | 16.0    | 0.2929  | 86.18     | 740.17  | 1186.24 | 1291.79 | 41.97        |
| **bf16k_turbo3v** | **9.75** | **0.2929** | 84.69 | 709.80  | 1179.45 | 1300.06 | 39.55        |
| **bf16k_turbo4v** | 10.25   | **0.2929** | 84.29  | 709.78  | 1171.72 | 1294.01 | 40.00        |
| bf16k_turbo2v     | 9.25    | 0.3030  | 85.96     | 713.33  | 1181.17 | 1301.00 | 40.34        |

`bf16k_turbo3v` and `bf16k_turbo4v` are bit-identical to the sym bf16
baseline (0.2929) on this prompt — the asym dispatch correctly
preserves K-side precision through the combined kernel. `bf16k_turbo2v`
drops 0.01 PPL (2-bit V codebook drift, expected).

**Compression at bf16-parity quality**:
- `bf16k_turbo3v` (9.75 b/elem averaged) is **1.64× smaller** than sym
  bf16 (16 b/elem) — same PPL sim, ~40% smaller KV pool, ~1.6× longer
  context at same VRAM ceiling.
- `bf16k_turbo4v` (10.25 b/elem) is 1.56× smaller at the same PPL.
- `bf16k_turbo2v` (9.25 b/elem) shows the per-side-pool dispatch works
  at the most aggressive V-side compression (2-bit) at the expected
  small quality cost.

(PPL absolute value lower than Bench F's 0.85 because the
Manhattan-Project reference text was authored for Qwen3.6's output
style; Qwen3-VL's continuations score lower on the same fixed
reference. The intra-Bench-G column comparison — `bf16k_*` vs `bf16`
baseline — is the load-bearing signal here, not the absolute number.)

### Symmetric matrix at TQ+ default (Qwen3.6-35B-FP8)

For reference / completeness — confirms parity with Bench D's TQ+
column above + adds the Turbo2 row that didn't exist upstream.

| dtype  | bits/elem | PPL sim | dec_short | pre_2K  | pre_8K  | pre_16K | dec_after_8K |
|--------|----------:|--------:|----------:|--------:|--------:|--------:|-------------:|
| bf16   | 16.0      | 0.8485  | 71.89     | 630.69  | 1227.54 | 1390.46 | 40.87        |
| fp8    | 8.0       | 0.8485  | 72.15     | 627.97  | 1239.06 | 1402.00 | 41.65        |
| turbo8 | 9.0       | 0.8384  | 71.73     | 640.08  | 1248.85 | 1416.28 | 41.11        |
| nvfp4  | 4.5       | 0.8485  | 72.10     | 636.28  | 1239.39 | 1410.56 | 41.53        |
| turbo4 | 4.5       | 0.8384  | 71.62     | 638.01  | 1238.62 | 1403.65 | 40.94        |
| turbo3 | **3.5**   | 0.8384  | 71.65     | 634.15  | 1230.51 | 1400.57 | 40.85        |
| **turbo2** | **2.5** | 0.6465 | 71.46     | 692.35  | **1336.50** | **1519.93** | 41.56        |

Turbo2 prefill is the fastest at every context length (692 / 1337 /
1520 vs ~635 / 1235 / 1405 for everything else) — +8-9% prefill
throughput at the 2-bit Lloyd-Max quality penalty. New in this PR
(upstream enum has Turbo3/4/8 only).

## Test plan

      (`--all-features` blocked by upstream `objc2` Apple-only build on
      Linux; clean on the per-crate `-p spark-runtime -p spark-model -p spark-server`
      gates)
- [x] `bash scripts/check-license-headers.sh` — script not present in
      tree at HEAD; manual audit shows all new `crates/**/*.rs` +
      `kernels/**/*.{cu,cuh}` files carry `// SPDX-License-Identifier: AGPL-3.0-only`
      per `.licenserc.yaml` paths-policy.
- [x] `typos` — clean (`WHT` / `wht` / `BA` / `OPTIN` already in
      `_typos.toml` allow-list)
- [x] `cargo-deny check` — clean
- [x] File-size cap — 1 file lifted: `init.rs` 524 LoC after TQ+
      kernel-handle additions, allow-list entry added to
      `.github/workflows/file-size-cap.yml` with rationale matching
      the existing entries.
- [x] **Tested against real model + hardware** — full reproduction
      matrix at `docs/turboquant-plus.md`. All 9 asym variants benched
      end-to-end on GB10 with the same `tests/atlas_bench_comprehensive.py`
      harness (~85 min total wall, 5-run median per metric, 3 context
      lengths). Bench JSONs archived locally
      (`~/atlas_tqplus_bench/results/bench_{A,B,C,D,E,F,G}_*.json`),
      forwarded on request.
- [x] **Added or updated tests** — 9 new Rust unit tests:
  - 4 in `crates/spark-model/src/layers/qwen3_attention/init_kernel_dispatch.rs::tests`
    — pure-Rust dispatch-routing tests asserting every asym variant
    ends up at a kernel module name containing its dtype-pair shape
    (`bf16k_turbo3v`, `fp8k_turbo4v`, etc.); a new asym variant added
    without dedicated kernels fails the substring check in CI.
  - 5 in `crates/spark-runtime/src/kv_cache/tests_tq_plus.rs` —
    enum-coverage on `ALL_VARIANTS` / `ASYM_VARIANTS` / `SYM_VARIANTS`
    arrays that force new dtypes to be added in tests when added to
    the enum.
  - Plus `tests/test_kv_dtype_smoke.py` — end-to-end per-dtype
    container start + 64-tok generation, distinguishes load-time SKIP
    (weight-incompat) from runtime FAIL (kernel crash).

## Notes for reviewers

This is a large PR by line count (~9000 LoC across 24 new kernel
files + Rust dispatch + tests + docs) because it's the full TQ+ port,
not a feature-flag slice. I've kept the commit history intentionally
atomic so each layer is independently reviewable:

```
591aa86 chore(tq+): typo fix + allow-list 5 dispatcher files after asym additions
1be228a style(tq+): cargo fmt --all on asym kernel dispatch files
51b31c0 docs(tq+): full 9-variant asym matrix from Bench F (Qwen3.6-FP8) + Bench G (Qwen3-VL NVFP4)
7f74fbe chore(tq+): drop dead bail tombstone (superseded by dispatch tests)
b87ad59 test(tq+): extract kernel-dispatch table + dispatch-routing unit tests
45e110d feat(tq+): Turbo*K + Turbo*V combined kernels + dispatch
b049fde feat(tq+): Bf16K + Turbo{2,4}V combined kernels + dispatch
47a8c70 fix(tq+): bail at init for asymmetric variants without combined kernels
2d693f6 docs(tq+): TurboQuant+ acknowledgement + reproduction guide
53297e5 chore(file-size-cap): allow-list init.rs after TQ+ kernel handles
3b6ca3c feat(tq+): TurboQuant+ Rust dispatch + tests + smoke harness
2ba4365 feat(tq+): TurboQuant+ kernel-level integration
```

(Rebased onto upstream `main` at `9d19c32`. The `chore(ci): green
clippy` commit from an earlier draft was dropped during rebase —
upstream PR #63 landed equivalent fixes for the same
`manual_checked_ops` + `unnecessary_sort_by` clippy debt.)

Suggested review order:
1. `docs/turboquant-plus.md` — the reproduction guide + measured tables
2. `2ba4365` — kernel-level integration (algorithm correctness)
3. `3b6ca3c` — Rust dispatch + per-side cache pool refactor
4. `b049fde` + `45e110d` — asym kernel families (the 8 missing combos)
5. `b87ad59` — dispatch tests (the safety net)
6. `47a8c70` + `7f74fbe` — bail introduced as scaffolding when only
   1 of 9 asym variants had kernels, then dropped after all 8 missing
   variants landed and the dispatch tests took over

**Honest framing — the symmetric line is at parity, not a win.**
Byte-identical output on fp8/turbo3/turbo4 between upstream `87b7bb3`
and this branch at greedy decode on Qwen3.6-A3B. The kernel-level
math fixes (signs, matched-norm L2, real Turbo3 prefill kernel) are
correct on paper but greedy decoding on this model + this prompt
absorbs the numerical differences without flipping argmax winners.
The value is in: (a) the new capabilities — Turbo2 dispatch fix,
9 asym variants, per-side cache pool, (b) the test coverage that
makes the dispatcher safe to extend, (c) correctness fixes that may
matter for other models or sampling regimes.

**Known follow-ups** (not blockers):

- The `qwen3_vl` weight loader in
  `crates/spark-model/src/weight_loader/qwen3_vl.rs:93` calls
  `quantized_auto()` whose `Bf16Raw` arm is `unreachable!()` —
  prevents loading raw-bf16 Qwen3-VL checkpoints. We benched the
  bf16k_* variants against the NVFP4 release of the same model as a
  workaround. One-arm fix per the qwen35 loader pattern is
  out-of-scope for this PR.
- Metal port of the TQ+ kernels — Atlas supports a Metal backend
  (`crates/spark-runtime/src/metal_backend.rs`, `kernels/metal/`,
  Cargo feature `metal`) but the entire pre-existing TurboQuant
  family (`turbo3` / `turbo4` / `turbo8`) is CUDA-only on `main` —
  no `wht`, `turbo`, or `reshape_and_cache_turbo` siblings in
  `kernels/metal/common/`. TQ+ extends the CUDA-only line; Metal
  support for the full TurboQuant family (existing sym + new asym)
  is a separate effort.
- TQ4_1S / TQ3_1S weight quantization (separate from KV cache).
- TriAttention long-context eviction policy (substrate exists via
  `boundary_dtype` on `build_layer_kv_dtypes`, no kernel yet).

**Asymmetric dispatch correctness** — the original Bench C run
caught 8-of-9 asym variants silently falling through to the K-side
symmetric kernels with mis-sized V pools, producing PPL sim 0.55-0.63
instead of the design target. Root cause: per-side pool sizing was
asymmetric but the dispatcher routed through the K-side sym kernel
which wrote V at K-side byte size. Fixed by (a) writing the 8
missing combined kernel triplets, (b) extracting the dispatch table
to a pure function with an exhaustive match (no `_` arm), (c) adding
unit tests that walk every asym variant × {hd=128, hd=256} and
assert it routes to a kernel module name containing its dtype shape.

## Authorship

The TQ+ algorithm (matched-norm L2, sparse V dequant, asymmetric K/V
design, InnerQ, weight pre-rotation) is research authored by Tom Turney
([@TheTom](https://github.com/TheTom)) prior to this Atlas port.
Reference implementations and the broader research line live at
[`TheTom/llama-cpp-turboquant`](https://github.com/TheTom/llama-cpp-turboquant)
and [`TheTom/turboquant_plus`](https://github.com/TheTom/turboquant_plus).

## CLA


AzeezIsh:
@TheTom thanks for the substantial feature work here. We took the branch through a deep validation pass on the GB10 flagship (Qwen3.6-35B-A3B-FP8, head_dim 256) and found and fixed a set of issues in the chunked-prefill paged-attention path; pushed directly to your branch as a4ad19fc (plus a fmt commit). Summary:

1. Symmetric turbo8/turbo4 chunk>=2 prefill history reads were dispatched to the NVFP4 kernel. For turbo8 the layout mismatch (FP8 rows read at 4-bit stride, BF16 group scales read as 1-byte E4M3) produced unbounded-magnitude garbage on any prompt longer than chunk_size; for turbo4 the byte layout matches but the E2M1 LUT was used instead of the Lloyd-Max codebook. This made every multi-turn agentic session collapse around turn 2 (the model's first decode tokens after the corrupted prefill came out as im_start/role literals). Added dedicated inferspark_prefill_paged_turbo8.cu / turbo4.cu and dispatch arms.

2. Symmetric turbo3 fell through to the FP8 catch-all arm with cache_stride() as the block stride (about 2.3x the real turbo3 stride). Your inferspark_prefill_paged_turbo3.cu is correct and was loaded, but nothing dispatched it; it is wired up now, same for turbo2.

3. The turbo2 BR=32 ops wrapper launched 256 threads into a kernel sized for 128 (warps 4-7 did OOB shared-memory reads). Fixed to 128.

4. Heads-up, not fixed: runtime WHT is silently disabled engine-wide. wht_bf16.cu exists but is not registered in any KERNEL.toml, so every wht_bf16 handle is 0 and the write path, decode bookends, and the new prefill bookends (added in this fix, mirroring decode) all no-op. If you intend the Hadamard rotation to be active, it needs a KERNEL.toml entry, and note that cache_skip.rs currently WHT-mutates k/v in place before the contiguous flash attention reads them, which would corrupt single-chunk prefill once WHT turns on. Decode's WHT gate also omits Turbo2 while the write path includes it.

5. One more suggestion: kernel module lookups fail at first dispatch rather than at startup. A startup fail-fast that validates all dtype-required kernel handles for the selected --kv-cache-dtype would have caught most of the above immediately.

Validation on the flagship after the fix (opencode webserver_ok agentic suite, N=5 each, fp8 baseline gate N=10 = 10/10): turbo8 went 0/5 to 5/5, turbo4 3/5, turbo4k_turbo8v 4/5, turbo3 1/5; quality now degrades monotonically with quant level as expected. Prefix caching (radix tree + SSM snapshots) was also verified against the TQ variants: warm-hit needle retrieval is exact with 13k cached tokens on turbo8/turbo4/turbo4k_turbo8v, and turbo8 with prefix caching ran the agentic suite 2/2 at about 3.4x lower wall time. cargo test --workspace passes (1554) and CI is green at 1dd16525.

TheTom:
@AzeezIsh thanks for the deep validation pass and the prefill fixes — items 1–3 are solid catches and the dedicated turbo8/turbo4 prefill kernels + dispatch arms look right. We kept all of them. On items 4 and 5, we ran the claims down on a GB10 of our own (ASUS GX10, same flagship Qwen3.6-35B-FP8, head_dim 256) and found one part of the item-4 premise worth correcting — the two bugs you flagged under it are real, and we've fixed them.

## Item 4 — good news: runtime WHT is actually live

> "wht_bf16.cu exists but is not registered in any KERNEL.toml, so every wht_bf16 handle is 0 and ... all no-op"

We initially expected the same thing, but it didn't reproduce — three independent checks on GB10 at 1dd1652 all show the rotation active:

1. **Registration**: `[modules]` entries are *name overrides*, not an allowlist — build.rs compiles every `.cu` in `kernels/gb10/common/` plus the model dir and registers each under its file stem unless overridden (see `collect_cu_files` + the modules collection in `crates/atlas-kernels/build.rs`). `wht_bf16 = "wht_bf16"` is additionally present in the model KERNEL.tomls (e.g. `kernels/gb10/qwen3.6-35b-a3b/nvfp4/KERNEL.toml:48`) and is an identity mapping, so even full-shadow toml selection can't drop it. The shadow mechanism you found is real for *renamed* modules (your item-3 class), but it cannot zero an identity-mapped one.
2. **Probe**: a minimal backend probe on the GB10 resolves both entry points with non-zero handles:
   `OK wht_bf16::wht_bf16_inplace -> 0xb7b4...` / `OK wht_bf16::wht_bf16_inplace_inv -> 0xb7b4...`
3. **Live server**: the startup kernel audit table shows `wht_bf16 ... used` (not in the MISSING list), and an A/B of the same greedy completion with `TQ_PLUS_WEIGHT_ROTATION=1` (which gates the runtime WHT launches off) produces a *different* completion — the rotation demonstrably fires on the write path and decode bookends.

So the WHT(Q)/iWHT(out) bookends you added in `paged.rs` are live, not dormant — which is good, they're needed (see below).

## Both item-4 sub-findings were right — fixed

Your two sub-points were correct, and WHT being live makes them real bugs rather than latent ones, so they were the important part of the report:

1. **`cache_skip.rs` in-place hazard — confirmed and fixed.** `write_kv_cache` WHT-rotates `k_contiguous`/`v_contiguous` in place before the contiguous chunk-0 FA reads them with un-rotated Q. We added the same WHT(Q)/iWHT(out) bookends you used for the chunk≥2 paged path, plus rotation of the un-written prefix rows on prefix-cache partial hits (the write-path bookend only rotates `[kv_write_start..]`). On the hybrid A3B flagship this corruption is masked — only a handful of layers are full attention and the residual stream dominates, which is why turn-1 prompts looked fine in both your runs and ours — but it's quality loss on every turbo prefill, and it would be much worse on dense-attention models.
2. **Turbo2 gate omission — confirmed and fixed.** The reference implementation gates rotation on `TURBO3 || TURBO4 || TURBO2` (`llama-cpp-turboquant`, `llama-graph.cpp`); Atlas's write path includes Turbo2 but the decode/prefill bookend gates didn't. Centralized as `KvCacheDtype::is_wht_rotated()` (Turbo2/3/4/8) and used at every gate, with an enum-walking unit test so a future variant can't desync the write path from the bookends again. Also wired the same bookends into the multi-seq batched decode path (`trait_impl/multi_seq/attn.rs`), which had none.

## Item 5 — startup fail-fast: agreed, implemented

`spark serve` now validates every dtype-required kernel handle (the chunked-prefill kernel for the selected `--kv-cache-dtype` plus the WHT bookend pair for turbo dtypes) right after backend init — before the multi-minute weight load — and bails with the full missing list. Covers each distinct per-layer dtype (high-precision/boundary layers). The hard-required reshape/decode pair already failed at layer init via `gpu.kernel(..)?`; this closes the `try_kernel` optional-handle class your items 1–3 came from. Requirement table lives next to the dispatch table in `kernel_requirements.rs` with enum-exhaustive tests.

Both changes are pushed as 3294a5b (fixes) + e8c50bb (file-size-cap split).

## Validation

All on GB10 native (Qwen3.6-35B-FP8, head_dim 256, max-seq-len 32768, kv-high-precision-layers 0), plus M5 Max metal:

- `cargo test --workspace` on GB10: 1519 passed, 0 failed (30 test binaries).
- turbo8 live, fixed build: short prompt exact; needle retrieval exact at 2,065 prompt tokens (single-chunk) and at 10,486 prompt tokens (chunk≥2 paged path with the WHT(Q)/iWHT bookends live).
- turbo2 A/B, same prompts, temp 0: with rotation + the gate fix → coherent completions (2-bit-level noise only); with rotation disabled (`TQ_PLUS_WEIGHT_ROTATION=1`) → severe degradation (hallucinated subject on a one-line factual question, repetition-loop on the needle prompt). The randomized-Hadamard rotation is carrying real quality at 2-bit, which is why the bookend gates have to stay in sync with the write path.
- M5 Max metal lib tests: 75 + 133 + 28 pass (incl. GPU parity tests).

AzeezIsh:
@TheTom you beat me to it - I had the same bookends staged locally and dropped them in favor of your 3294a5b5/e8c50bbb when I fetched, since yours also covers the multi_seq batched path I had missed and adds the enum-walking gate tests. You are right about item 4 in my earlier comment: the [modules] table is an override map, not an allowlist, so wht_bf16 was auto-registered under its stem and the rotation was active the whole time. My apologies for the wrong premise; the two real bugs you extracted from it were the ones that mattered.

GB10 validation of your head (Qwen3.6-35B-A3B-FP8, 32K, MTP, opencode agentic suite, N=5 per dtype): fp8 baseline 10/10, turbo8 5/5, turbo4 5/5 (was 3/5 before the bookends), and bf16k_turbo4v plus fp8k_turbo3v - both 0/5 before - are passing cleanly so far in the running sweep. The single-chunk raw-Q-vs-rotated-K mismatch also turns out to explain the EOS run-on artifacts we had been tracking separately (title-style prompts now terminate naturally). cargo test 1556 passed, CI green on e8c50bbb. Full per-dtype matrix incl. the 2-bit variants lands overnight; will post it here.

AzeezIsh:
Overnight validation matrix on e8c50bbb is complete. Qwen3.6-35B-A3B-FP8 flagship (32K, MTP fp8, opencode agentic webserver_ok suite, N=5 per dtype; dgx2 rows are a 16K deployment with a weaker baseline, used only where its results were clean):

| KV dtype | box | n | ws_ok | mean wall (s) |
|---|---|---|---|---|
| fp8 (baseline) | dgx1 | 10 | 10/10 | 194 |
| turbo8 | dgx1 | 5 | 5/5 | 113 |
| turbo4 | dgx1 | 5 | 5/5 | 188 |
| turbo3 | dgx1 | 5 | 5/5 | 277 |
| turbo2 | dgx1 | 5 | 0/5 | - |
| turbo4k_turbo8v | dgx1 | 5 | 4/5 | 203 |
| turbo4k_turbo3v | dgx1 | 5 | 3/5 | 255 |
| turbo3k_turbo8v | dgx1 | 5 | 5/5 | 284 |
| bf16k_turbo4v | dgx1 | 5 | 4/5 | 296 |
| bf16k_turbo3v | dgx1 | 10 | 0/10 | - |
| bf16k_turbo2v | dgx2 | 5 | 5/5 | 166 |
| fp8k_turbo4v | dgx1 | 5 | 5/5 | 183 |
| fp8k_turbo3v | dgx2 | 5 | 5/5 | 223 |
| fp8k_turbo2v | dgx2 | 5 | 5/5 | 140 |

Eleven of thirteen TQ variants now pass at 4/5 or better (eight at 5/5). turbo8/turbo4/turbo3 and several asym variants were 0-3/5 before the chunk>=2 dispatch fix and your bookends; the lift is consistent and large. Two characterized non-passes, both quality floors rather than serving bugs:

1. turbo2 (symmetric 2-bit K): prose, thinking, retrieval, and multi-chunk prefill are all coherent, but under tool-grammar constrained decoding the model gets stuck inside the tool-call envelope (server log: 'Stuck in tool-call ENVELOPE for 1024+ tokens with no </tool_call>'), the envelope guard ends the response, and the agent session dies on turn 1. 2-bit keys cannot sustain structured tool emission. The guard + sanitizer behave correctly.

2. bf16k_turbo3v (0/10): builds real Axum servers every run, but consistently ignores the prompt's exact ATLAS_HARNESS_PORT requirement (hardcodes 0.0.0.0:3001 or invents PORT/3000), so the harness's ephemeral-port ping fails. Instruction-adherence drift specific to this combo - notably bf16k_turbo2v (more aggressive V) passes 5/5, so this looks like a tuning target (per-dtype kv-high-precision-layers?) rather than monotonic quant damage. All mechanical probes (needle at 13k, multi-chunk replay, greedy coherence) are clean.

One more observation for the tuning table: turbo8 is now both clean AND the fastest config we measured (mean 113s vs fp8's 194s on the same suite) - the 8-bit TQ KV bandwidth saving shows up directly in agentic wall time.

AzeezIsh:
Closing update: the two non-passing variants from the matrix above are both fully recovered by config. Re-running each with --kv-high-precision-layers 8 (first 8 + last 8 attention layers at BF16 KV, leaving only the mid-stack at the aggressive dtype):

| KV dtype | default HP (auto=4) | --kv-high-precision-layers 8 |
|---|---|---|
| turbo2 | 0/5 (tool-call envelope collapse) | 5/5, mean wall 181s |
| bf16k_turbo3v | 0/10 (port-instruction drift) | 5/5, mean wall 189s |

So every TurboQuant+ variant now passes the agentic suite on the GB10 flagship - thirteen for thirteen - with turbo2 and bf16k_turbo3v requiring the documented hp=8 setting. Suggest either bumping the auto HP heuristic for those two dtypes (the auto formula currently lands on 4 for this 10-attention-layer model) or noting hp=8 in their recipe defaults.

AzeezIsh:
@TheTom wrapping this up. Final commit 55db9fa7 folds the matrix findings into the auto heuristic: symmetric turbo2 was missing from the auto high-precision list entirely (fell through to hp=0 with only a warning), and turbo2 plus bf16k_turbo3v now default to ceil(4n/5) with floor 4 instead of ceil(n/3), which reproduces the validated hp=8 on the flagship. The heuristic is extracted into auto_high_precision_layers() with enum-walking tests so a future dtype cannot silently fall through to hp=0.

Verified live on GB10: turbo2 with no explicit flag now starts with the full bf16 boundary (10/10 attn layers on this model) and matches the fp8 baseline output on greedy short prompts. cargo test --workspace: 1560 passed, 0 failed. With that, every TQ+ variant passes the agentic suite on default settings, thirteen for thirteen.

Merging once CI completes. Thanks for the feature itself and for running down the item-4 premise on your own hardware instead of taking my word for it. The enum-walking gate tests and the startup fail-fast made this a much safer subsystem than it was a week ago.

PR #147, opened 2026-06-10
feat(tq-metal): TurboQuant KV cache for the Metal backend — turbo8/4/3/2 + safer-asym + WHT rotation + sparse-V dequant

TheTom:
Follow-up to #133: ports the TurboQuant KV cache to the Metal backend's contiguous (non-paged) cache — the path the `qwen3_5` vendor-agnostic forward and the `metal_qwen35_inference` end-to-end driver use. All four symmetric turbo dtypes, the safer-asym Bf16K+TurboNV family, the canonical Rademacher-signed WHT rotation, and the attention-gated sparse-V dequantization land together, with byte-exact parity tests and an on-device quality/throughput validation against both the bf16 baseline and the llama.cpp TurboQuant reference engine on the same hardware.

### MSL kernels (`kernels/metal/common/`)

| File | Why |
|---|---|
| `wht_bf16.metal` | Per-head Walsh-Hadamard rotation, forward + inverse, head_dim 128/256. One simdgroup per head, `simd_shuffle_xor` butterflies. `TQ_PLUS_SIGNS`-gated two-sided Rademacher form S2·H·S1 with seed-42 sign tables verified byte-identical to the CUDA `tq_plus_signs.cuh` vendoring; without the define both entry points reduce to the plain self-inverse WHT. |
| `kv_cache_append_turbo8.metal` | Quantizing append: FP8 E4M3 data (manual encode — MSL has no fp8 type) + bf16 per-16-element group scales (amax/448). Quantizes against the bf16-rounded scale the decode kernel reads back. |
| `kv_cache_append_turbo4.metal` | 16-level Lloyd-Max codebook, nibble-packed 2 elems/byte, FP8 E4M3 group scales with the matched-norm L2 correction (`‖original‖ / ‖centroid_vec‖`) — same scheme as `reshape_and_cache_flash_turbo4`. |
| `kv_cache_append_turbo3.metal` | 8-level codebook, 8 values → 3 bytes packing (bit-layout identical to the CUDA kernel), matched-norm FP8 scales. |
| `kv_cache_append_turbo2.metal` | 4-level codebook, 4 elems/byte, matched-norm FP8 scales. |
| `attention_decode_turbo{8,4,3,2}.metal` | Decode SDPA with inline dequant, same threadgroup/softmax staging as the bf16 kernel, plus the sparse-V gate (below). |
| `kv_cache_append_bf16k_turbov.metal` | Safer-asym appends (Bf16K + Turbo{4,3,2}V): K copied bit-exact as raw bf16, V quantized + matched-norm E4M3 scales. Three entry points sharing the quant helpers. |
| `attention_decode_bf16k_turbov.metal` | Safer-asym decode: K scored as raw bf16 (Q stays un-rotated — the K side never leaves the original basis), V dequantized inline with the sparse-V gate, iWHT(out) on the rotated-V output. |

### Rust (`forward/qwen3_5`, runtime tests, example)

- `MetalKvDtype` enum (`Bf16 | Turbo8/4/3/2 | Bf16KTurbo4V/3V/2V`) with per-side `k_is_rotated()` / `v_is_rotated()` + single `LayerKvCache::alloc` (per-side scale buffers); the forward gates each WHT bookend on its own side — symmetric dtypes get WHT(K/V) → append → WHT(Q) → attention → iWHT(out); the safer-asym family rotates V only and keeps Q raw, mirroring the CUDA `kv_pair()` bookend semantics from #133.
- `Qwen35Kernels::resolve` hard-requires every turbo handle — a turbo cache can never silently fall back to the bf16 kernels (same fail-fast philosophy as #133's startup preflight).
- Sparse-V threshold is a runtime kernel argument; the forward reads `ATLAS_SPARSE_V_THRESHOLD` (default 1e-3, `0` disables).
- Example knob: `ATLAS_KV_DTYPE={bf16,turbo8,turbo4,turbo3,turbo2,bf16k_turbo4v,bf16k_turbo3v,bf16k_turbo2v}`; `ATLAS_LOGITS_OUT=path` dumps per-step bf16 logits for offline comparison.
- 11 GPU parity tests (`parity_turbo*.rs`): WHT vs signed CPU reference, inverse roundtrip, **byte-exact** quantizer parity for all four dtypes (packed indices + scale bytes, including the matched-norm path), and decode attention vs a CPU reference over the CPU-dequantized cache with the gate mirrored.
- `tests/metal_kv_kld_compare.py`: per-step KLD + top-1 agreement between two logit dumps, trimmed at the first greedy divergence.

### Fixes found along the way

- **Latent OOB in every Metal decode kernel** (including the pre-existing bf16 `attention_decode`): stages 2–5 looped to the raw `seq_len` while the threadgroup score array holds `MAX_SEQ_DECODE = 4096` entries — a >4096-token sequence read *and wrote* threadgroup memory out of bounds. All five kernels now clamp once at entry.
- **KERNEL.toml flag-shadow on Metal**: the per-quant toml's empty `extra_metal_flags` fully shadowed common's `-ffast-math` (same mechanism class as the chunk≥2 dispatch issue in #133's review). Both flags (`-ffast-math`, `-DTQ_PLUS_SIGNS`) now set in both files, with a warning comment and a note in `METAL_BACKEND.md`. The GB10 model targets have the same shadow dropping `-DTQ_PLUS_SIGNS` (production CUDA runs the plain WHT) — left for a separate PR since flipping it changes the cache format and needs quality re-validation.

## Sparse-V dequantization (attention-gated V skip)

The softmax weights are known before V is touched; positions with unnormalized weight `exp(score − max) ≤ 1e-3` skip the V dequant + accumulation entirely. M5 Max, Qwen3.5-4B-MLX-8bit, 3,826-token prompt, 64-token greedy decode:

| config | decode tok/s |
|---|---|
| bf16 cache | 9.4 |
| turbo4, gate off | 10.7 |
| **turbo4, gate on (1e-3)** | **21.2** |

+98% from the gate, 2.25× over bf16 — the compressed cache plus the gate beats raw bf16 outright at this depth, and the benefit scales with context length.

## Quality

KLD of each turbo dtype's logits vs the bf16-cache run (same prompt, greedy, 25 steps, short context):

| dtype | mean KLD | top-1 agreement |
|---|---|---|
| turbo8 | 0.00036 | 25/25 |
| turbo4 | 0.0019 | 24/25 |
| turbo3 | 0.0037 | 25/25 |
| turbo2 | 0.0162 | 24/25 |
| bf16k_turbo4v | 0.0020 | 25/25 |
| bf16k_turbo3v | 0.0056 | 10/11* |
| bf16k_turbo2v | 0.0146 | 24/25 |

\* greedy paths split at a borderline token; KLD over the shared prefix.
bf16k_turbo2v beats symmetric turbo2 (0.0146 vs 0.0162) — the safer-asym
premise (K precision dominates) visible already at 2-bit.

Monotone with bit width; turbo8 is token-identical to bf16, and the single turbo4/turbo2 flip is a near-tie logit at the same tail position.

Cross-engine sanity check on the same M5 Max against the llama.cpp TurboQuant reference engine (Qwen3.5-4B-Q8_0, teacher-forced corpus KLD, depth 3,826): reference turbo4 KLD median 0.0086 / same-top 93.7%, decode 72.7 vs 80.6 t/s f16. Atlas per-step KLD (0.002–0.008) sits at the reference median, confirming a faithful rotation/codebook port. Absolute throughput is not comparable — the reference engine runs production fused flash-attention kernels (~3.4× faster engine overall); this PR's decode kernels keep the naive reference SDPA shape, which is also why compression + the gate yields a net *speedup* here while costing ~10% there.

## Scope notes / follow-ups

- Contiguous cache only (the Metal serving path); paged turbo variants belong with a future Metal paged-attention PR, as does a fused/simdgroup FA rewrite to close the engine-speed gap.
- The KLD harness is greedy-divergence-limited; a teacher-forced eval mode is the cleanest next step for corpus-level numbers.
- head_dim 512 WHT, the Fp8K and turbo-K/turbo-V both-sides asym combos from #133 are CUDA-only for now (the safer-asym Bf16K family — the production-recommended frontier — is in).
- GB10 `-DTQ_PLUS_SIGNS` toml-shadow: separate PR after quality re-validation.

## Validation

- Metal lib suites: 75 + 144 + 28 pass (11 turbo/asym parity tests, several byte-exact).
- E2E `metal_qwen35_inference` runs for all eight dtypes; turbo8 token-identical to bf16 at 39.8 tok/s short-context.
- `cargo fmt` clean; all touched `.rs` files under the 500-LoC cap.


AzeezIsh:
@TheTom reviewed. Scope is exactly as advertised: 13 MSL kernels + metal_backend tests + the vendor-agnostic qwen3_5 forward and example driver, zero CUDA serving surface. Verified on our side that the GB10 path is untouched: full CUDA build on the PR head plus spark-model and spark-runtime suites, 188 passed.

Three things I want to call out as model contributions:

1. The latent OOB you found in every Metal decode kernel (stages 2-5 looping raw seq_len against the 4096-entry threadgroup array) is the kind of fix that pays for the whole PR by itself.
2. Byte-exact quantizer parity tests against the CUDA layout, including the matched-norm scale path, means the two backends cannot silently drift, which was my main worry about a second TQ implementation.
3. Hard-requiring every turbo handle in Qwen35Kernels::resolve mirrors the #133 fail-fast and closes the silent-bf16-fallback class on Metal before it ever shipped.

Your M5 Max on-device validation plus the macOS CI job running the parity suite is the verification basis here, same trust pattern as the GB10 work. Auto-merge armed on green. The KLD compare script is a nice touch, we will likely lift it for CUDA-side A/Bs too.

PR #150, opened 2026-06-10
fix(lm-head): honor pre-packed NVFP4 lm_head before dtype overrides

AzeezIsh:
Post-merge regression run on the flagship surfaced a CUDA_ERROR_ILLEGAL_ADDRESS(700) at model build for nvidia/Qwen3.6-35B-A3B-NVFP4. Bisect showed it reproduces on every commit back to 9cb06bc and on both GB10 nodes with independent HF caches - a latent day-one bug for checkpoints that ship lm_head pre-quantized (packed U8 weight + weight_scale + weight_scale_2, so no BF16 lm_head tensor exists), not a regression from the recent merges.

Two paths in setup_lm_heads treated the packed buffer as [vocab, hidden] BF16: the --lm-head-dtype bf16 skip branch (which the published recipe for this model sets), and the Step-3a draft-only NVFP4 head for the MTP proposer, which called quantize_to_nvfp4 over vocab*hidden BF16 elements of the half-size packed buffer and faulted at the absmax kernel.

Fix: the pre-packed check now runs before the dtype-override branches. A pre-packed head is a checkpoint constraint - load it directly as the NVFP4 head and warn when --lm-head-dtype bf16/fp8 is requested and overridden. Step-3a then short-circuits naturally (proposer reuses the main NVFP4 head, byte-for-byte the pre-decouple behavior). The duplicate prepacked branch below the overrides is removed.

Verified on GB10: the model now builds and serves with the exact published recipe flags; the opencode webserver_ok N=10 gate is running and results will be posted here. Regression status for the rest of the suite on main 61b7d9a0: FP8 flagship 14/15, turbo8 5/5, cargo test 1567/0.


AzeezIsh:
Gate result on the PR head (4f61bab1): nvidia/Qwen3.6-35B-A3B-NVFP4 with the published recipe flags, opencode webserver_ok N=10 = 9/10. The single miss was run 1, a 34s early-quit (2 turns, skeleton project, cargo-valid) with no correctness signature - consistent with the historical 'consistently 10/10, rarely 9/10' envelope for this suite. Runs 2-10 were all green.

Side note from the run-index analysis this surfaced: across healthy tiers, run 1 fails ~11% vs ~5% for all later indices. Recent tiers (including this one) skip the harness warmup, so run 1 absorbs CUDA-graph capture and, on this config, the fp8 KV calibration's first 256 tokens converging during the opening turns. Worth considering having serve complete calibration + graph warmup before accepting traffic; until then gates should run with warmup enabled.

Full post-merge regression picture on main + this fix: FP8 flagship 14/15, turbo8 5/5, cargo test 1567 passed / 0 failed, NVFP4 9/10. Ready for review.

PR #151, opened 2026-06-10
feat(tq-metal): teacher-forced KLD eval mode

TheTom:
The KLD harness from #147 trimmed at the first greedy divergence —
fine for near-lossless dtypes, but the low-bit rows got 3-11 valid
positions before the paths split. Teacher forcing removes the limit:

- ATLAS_DUMP_TOKENS=path writes a run's argmax ids (one per line).
- ATLAS_FORCE_TOKENS_FILE=path feeds that list in place of the
  sampled token each step (logits still dumped, model's own argmax
  still reported, EOS ignored so every run covers the full list).
  Two runs with the same list share an identical context at every
  position, so their logit dumps are KLD-comparable end to end.
- tests/metal_kv_kld_compare.py --teacher-forced skips the
  divergence trim.

Re-measured ladder on Qwen3.5-4B-MLX-8bit (M5 Max, 49 teacher-forced
positions, KLD vs bf16 cache):

  dtype          mean KLD   top-1
  turbo8         0.00037    49/49
  bf16k_turbo4v  0.00134    49/49
  turbo4         0.00158    48/49
  turbo3         0.00428    48/49
  bf16k_turbo2v  0.01523    48/49
  turbo2         0.01963    48/49

Monotone with bit width; safer-asym beats symmetric at equal V-bits
(bf16k_turbo4v < turbo4, bf16k_turbo2v < turbo2) — the K-precision
premise holding at full positional coverage.

### DIFF
--- a/crates/spark-model/examples/metal_qwen35_inference/eval_io.rs
+++ b/crates/spark-model/examples/metal_qwen35_inference/eval_io.rs
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+//! Eval-comparison IO knobs for the inference driver: per-step logit
+//! dumps (`ATLAS_LOGITS_OUT`), argmax-sequence dumps
+//! (`ATLAS_DUMP_TOKENS`), and teacher-forced token lists
+//! (`ATLAS_FORCE_TOKENS_FILE`). Used by `tests/metal_kv_kld_compare.py`
+//! to compare KV dtypes with identical contexts at every position.
+
+use anyhow::Result;
+
+/// Open the per-step logits sink when `ATLAS_LOGITS_OUT` is set. Raw
+/// little-endian bf16, `[n_steps, vocab]`.
+pub fn logits_writer() -> Option<std::cell::RefCell<std::io::BufWriter<std::fs::File>>> {
+    std::env::var("ATLAS_LOGITS_OUT").ok().map(|p| {
+        std::cell::RefCell::new(std::io::BufWriter::new(
+            std::fs::File::create(p).expect("create ATLAS_LOGITS_OUT"),
+        ))
+    })
+}
+
+/// Load the teacher-forcing list from `ATLAS_FORCE_TOKENS_FILE`
+/// (newline-separated token ids; position 0 replaces the first sampled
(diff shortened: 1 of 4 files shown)


crates/spark-model/examples/metal_qwen35_inference/eval_io.rs
```diff
@@ -0,0 +1,44 @@
...
+
+/// Open the per-step logits sink when `ATLAS_LOGITS_OUT` is set. Raw
+/// little-endian bf16, `[n_steps, vocab]`.
+pub fn logits_writer() -> Option<std::cell::RefCell<std::io::BufWriter<std::fs::File>>> {
+    std::env::var("ATLAS_LOGITS_OUT").ok().map(|p| {
+        std::cell::RefCell::new(std::io::BufWriter::new(
+            std::fs::File::create(p).expect("create ATLAS_LOGITS_OUT"),
+        ))
+    })
+}
```
tbraun96, line 18:
```suggestion
pub fn logits_writer() -> Option<parking_lot::Mutex<std::io::BufWriter<std::fs::File>>> {
    std::env::var("ATLAS_LOGITS_OUT").ok().map(|p| {
        parking_lot::Mutex::new(std::io::BufWriter::new(
            std::fs::File::create(p).expect("create ATLAS_LOGITS_OUT"),
        ))
    })
}
```

We should stay away from RefCell so we don't force future development to become pidgeon-holed into single-threaded contexts. There is no realistic overhead in performance switching from a `RefCell` to a `Mutex` anyways given the frequency this function is called.

tbraun96 (changes requested):

AzeezIsh:
@TheTom reviewed: scope is the metal example driver and the KLD compare script only, zero production surface, and teacher forcing is the right fix for the trim-at-divergence problem (3-11 valid positions on the low-bit rows was not a usable signal). The same-list sha check on both dumps is a nice guard against comparing different prompts by accident. Auto-merging on green. We will want this exact mode when we do the CUDA-side per-dtype KLD table, so consider the interface blessed.

AzeezIsh (approved):
Example + eval tooling only; teacher-forced KLD is the right design. CI is the gate.

PR #152, opened 2026-06-10
fix(build): merge common + model KERNEL.toml instead of shadowing — activates TQ_PLUS_SIGNS on gb10

TheTom:
A model-specific `KERNEL.toml` fully replaced the common one in `build.rs`'s target resolution: common's build flags and `[modules]` overrides were silently dropped for every model target. Two real consequences had shipped from this:

- All 13 gb10 model targets compiled without common's `-DTQ_PLUS_SIGNS`, so the production CUDA `wht_bf16` ran the plain (sign-free) WHT instead of the canonical two-sided Rademacher rotation S2·H·S1 the Lloyd-Max codebooks are optimized for. The metal per-quant toml likewise dropped common's `-ffast-math`.
- A `[modules]` mapping present only in `common/` silently registered the module under its file stem and kernel lookup failed at first dispatch — the chunk≥2 prefill misdispatch class found during #133's review, worked around at the time by hand-propagating mappings into all 13 model tomls.

`build.rs` now parses common first as the base, then the model toml: flags are appended (deduped, model last), `[modules]` merges with the model winning per key. The metal per-quant toml's duplicated flags are dropped in favor of inheritance; the gb10 model tomls' duplicated `[modules]` entries are left as harmless duplicates.

## Effect: canonical rotation goes live on gb10

`-DTQ_PLUS_SIGNS` becomes active for all gb10 model targets — `wht_bf16_inplace` becomes S2·H·S1 and `wht_bf16_inplace_inv`'s reversed sign order starts mattering (the bookends from #133 already use the correct inverse handle). The KV cache is process-local, so there is no format-migration concern: old and new builds simply quantize into different rotated bases.

## Validation (GB10, Qwen3.6-35B-FP8, head_dim 256, 32K, hp=0)

Signs-on build, same suite as the #133-era signs-off baseline:

| dtype | short prompt | needle @2,065 tok | needle @10,486 tok (chunk≥2) |
|---|---|---|---|
| turbo8 | coherent, factually exact | exact | exact |
| turbo4 | coherent, factually exact | exact | exact |

Matches the signs-off baseline on every probe (turbo8 needle results identical; turbo4 long-needle is newly verified here). The metal backend has run the signed rotation since #147 with byte-verified sign tables — this PR brings the CUDA side to the same canonical form.

`cargo test --workspace` and the full metal parity suite pass on the merged-flag builds (metal target now inherits `-ffast-math` + `-DTQ_PLUS_SIGNS` from common with an empty per-quant addition list).

### DIFF
--- a/crates/atlas-kernels/build.rs
+++ b/crates/atlas-kernels/build.rs
@@ -474,15 +474,31 @@ fn resolve_targets(workspace_root: &std::path::Path) -> Vec<Target> {
                 common_kernel_dir.display(),
             );
 
-            // KERNEL.toml: prefer model-specific, fall back to common
-            let toml_dir = if has_model_dir && model_kernel_dir.join("KERNEL.toml").exists() {
-                &model_kernel_dir
-            } else if has_common_dir {
-                &common_kernel_dir
-            } else {
-                &model_kernel_dir
-            };
-            let (extra_flags, module_overrides) = parse_kernel_toml(toml_dir, &target_vendor);
+            // KERNEL.toml: MERGE common + model-specific. The old
+            // prefer-model-else-common selection meant a model toml fully
+            // SHADOWED common — silently dropping common's build flags
+            // (gb10 model targets lost -DTQ_PLUS_SIGNS; the metal per-quant
+            // toml lost -ffast-math) and common's [modules] overrides (the
+            // chunk>=2 prefill misdispatch class, previously worked around
+            // by propagating mappings into all 13 model tomls). Semantics
+            // now: common parses first as the base; the model toml appends
+            // flags (deduped, model last) and wins per-key on [modules].
+            let mut extra_flags: Vec<String> = Vec::new();
+            let mut module_overrides: HashMap<String, String> = HashMap::new();
+            if has_common_dir && common_kernel_dir.join("KERNEL.toml").exists() {
+                let (f, m) = parse_kernel_toml(&common_kernel_dir, &target_vendor);
+                extra_flags.extend(f);
+                module_overrides.extend(m);
+            }
+            if has_model_dir && model_kernel_dir.join("KERNEL.toml").exists() {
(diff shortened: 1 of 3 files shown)


AzeezIsh:
@TheTom verified on dgx1, with one important operational finding.

The merge semantics work as described. Clean kernel rebuild of the flagship target on this branch produces a wht_bf16 PTX of 92,504 bytes, matching a direct nvcc -DTQ_PLUS_SIGNS reference compile (90,967) vs 53,908 without the define, so the signed rotation is provably in the shipped PTX now. cargo test --workspace: 1567 passed. The doc fix and the metal toml inheritance both read right, and this properly retires the propagate-into-13-tomls workaround from the #133 review.

The operational finding: an INCREMENTAL build from a pre-existing target dir did NOT pick up the flag change. After checking out this branch over an existing build, the emitted wht_bf16 PTX was still the unsigned 55KB variant; only rm -rf target/.../atlas-kernels-* forced the signed compile. The in-run dedup cache keys on (source, arch, sorted-flags) correctly, so this is the build script reusing stale out/ artifacts across runs when only toml-sourced flags changed. Consequence: anyone upgrading in place gets a silently unsigned build while believing they run the canonical rotation. Worth a follow-up that includes the resolved flag list in the artifact staleness check (or a flags-hash stamp file in out/). I can file it if you do not want to fold it in here.

Live turbo8/turbo2 probes on the signed build are pending on our side, the box is currently holding orphaned driver memory from an unrelated debug session and needs a reset I will not do while other work is running on it. Your GB10 needle matrix on this branch plus the PTX-level proof above is sufficient for me: approving, and I will arm auto-merge. If anything looks off in the post-merge agentic suites we know exactly which knob changed.

AzeezIsh (approved):
Merge semantics PTX-verified on GB10 (signed wht_bf16 in the artifact after clean rebuild), 1567 workspace tests pass. Incremental-build staleness flagged as follow-up.

PR #154, opened 2026-06-10
fix: agentic wall regression + warm-cache corruption series (MTP live-state, KV write floor, stream-safe rollback)

tbraun96:
Fix series for the agentic wall-time regression and the warm-cache corruption family found while investigating it (2026-06-10).

Background: the flagship Sigma-wall regressed from ~1331s (2026-06-07 series) to ~2400s. A 38-agent diff forensics pass attributed 80-90 percent to serving-config deltas (missing prefix caching, MTP fp8 draft head, uncalibrated fp8 KV) rather than code. Re-enabling the fast flags then exposed a chain of real correctness bugs, each isolated with N=5/N=10 opencode harness gates on both DGX boxes. Full analysis: tasks/wall_regression_diff_analysis.md.

Fixes, in commit order:

1. fp8 KV calibration key placement: kernels/gb10/qwen3.6-35b-a3b/MODEL.toml had fp8_kv_calibration_tokens=256 under [model] but parse_behavior() reads only [behavior] - parsed as 0, so every default-flag serve ran UNCALIBRATED fp8 KV (scale=1.0, saturating E4M3 on deep layers). Moved to [behavior].

2. fp8 MTP draft-head collapse: --mtp-quantization fp8 kept the 1-layer MTP KV cache in FP8 with hardcoded unit scales, collapsing draft acceptance to ~0 percent and making MTP a net slowdown. The MTP KV cache now uses BF16 for fp8 heads too (single tiny layer, negligible cost); NVFP4 MTP untouched.

3. GDN warm-replay kernel routing: Marconi warm hits replay [snap_tok, matched) through the FLA chunked kernel anchored at arbitrary snapshot offsets with bf16 intermediates - non-bit-exact vs the pass that produced the cached K/V. New ForwardContext::gdn_exact_replay routes warm-restored prefills through the bit-faithful WY4 recurrence; cold prefills keep FLA.

4. Prefix-cache KV write floor: the chunked paged prefill rewrote the replay range into SHARED refcounted radix blocks unconditionally (prefill_attention_paged ignored kv_write_start). Any non-bit-exact recompute poisoned the blocks and ratcheted drift across turns. New kv_write_floor skips the write for already-cached positions on all single-stream prefill paths; paged attention reads the original cached values. (Includes the int64 slot-mapping offset fix.)

5. MTP live-state restore on rejection: under MTP the GDN layers keep live + checkpoint state; rejection corrected only the checkpoint, leaving the live buffer holding the rejected token. Three scheduler paths (spontaneous think gate-flip, concurrent request, MTP bootstrap) ran plain decode on the dirty live state, and the bootstrap then baked it into the checkpoint. Rejection now restores the live state too, with consumer-side sync_secondary ordering at the two non-verify consumers (commit-side ordering measured ~25 percent decode wall - reverted in favor of consumer-side).

6. Stream-safe rollback: watchdog rollback+re-steer drops 60-80 already-generated tokens; streaming requests have already flushed them (including live tool-arg deltas), and SSE cannot retract - the client accumulates original + regeneration (duplicated/blended tool-arg paths). Before #141 the hybrid flagship's rollbacks always declined (NoSsmSnapshot) and ended early; #141 made them succeed, enabling the unsafe path. rollback_to_boundary now declines with RollbackFallback::StreamUnsafe for streaming sequences.

7. Behavior: max_inter_tool_prose restored to 384 for qwen3.6-35b-a3b (the 1024 raised in #90 let temp-0.3 paraphrase loops run 2.7x longer per inter-tool gap, turning runs into 360s cap-riders).

8. Serve-time warning for the one KNOWN residual: --enable-prefix-caching combined with --speculative on hybrid (SSM) models still corrupts ~1 in 3 agentic runs (warm restores x working draft traffic). Each feature is verified clean alone. Tracked in a follow-up issue.

Verification (all on Qwen/Qwen3.6-35B-A3B-FP8, opencode webserver_ok harness):
- Ship config (MTP bf16, bf16 KV, slai, no prefix caching): dgx1 10/10 Sigma=2391s, dgx2 5/5 Sigma=1063s - 15/15 combined, zero corruption signatures.
- MTP+no-cache discriminator 5/5; no-MTP+cache 13 clean runs; MTP warm-stress repro (new bench/fp8_dgx2_drift/harness/mtp_warm_stress.py) 3x32 turns clean post-fix vs 1-in-16-turn corruption pre-fix.
- cargo tests: spark-model 75/75, spark-server 496+111 integration/lib, atlas-kernels 28/28.
- K2 acceptance with the fp8 MTP-KV fix family: 65-81 percent observed (vs ~0 percent collapsed).

New repro/diagnostic tooling: bench/fp8_dgx2_drift/harness/mtp_warm_stress.py, warm_cold_chain_repro.py.

### DIFF
--- a/bench/fp8_dgx2_drift/harness/mtp_warm_stress.py
+++ b/bench/fp8_dgx2_drift/harness/mtp_warm_stress.py
@@ -1,0 +1,167 @@
+#!/usr/bin/env python3
+"""MTP x warm-restore stress repro (2026-06-10).
+
+The warm-hit token-stutter only manifests with --speculative under real
+agentic traffic. Hypothesis: high draft-rejection rates (unpredictable
+text) x many decode steps x Marconi warm restores desync rewind
+bookkeeping. This script maximizes those three factors and scans every
+reply for stutter signatures.
+
+Per iteration: one long chained conversation (10k-token system prompt so
+every turn is multi-chunk and warm), tools defined, thinking left at the
+server default, temp 0.3, prompts that demand UNPREDICTABLE output
+(random-looking identifiers, mixed prose/code) to tank MTP acceptance,
+plus exact-path echo demands so corruption is detectable.
+
+Usage: python3 mtp_warm_stress.py [--iters 3] [--turns 8]
+Exit code 1 if any stutter suspect found (prints them).
+"""
+
+import argparse
+import json
+import re
+import sys
+import urllib.request
+
+FILLER = (
+    "Consider the module layout under /tmp/harness-proj/src with files "
+    "main.rs, routes.rs, state.rs and tests in /tmp/harness-proj/tests. "
+)
+
+TOOLS = [
+    {
+        "type": "function",
+        "function": {
+            "name": "write",
+            "description": "Write a file to disk",
+            "parameters": {
+                "type": "object",
+                "properties": {
+                    "filePath": {"type": "string", "description": "Absolute path"},
+                    "content": {"type": "string"},
+                },
+                "required": ["filePath", "content"],
+            },
+        },
+    },
+    {
+        "type": "function",
+        "function": {
+            "name": "bash",
+            "description": "Run a shell command",
+            "parameters": {
+                "type": "object",
+                "properties": {"command": {"type": "string"}},
+                "required": ["command"],
+            },
+        },
+    },
+]
+
+# Unpredictable content requests tank n-gram-style draft acceptance;
+# exact path echoes make stutter detectable.
+PROMPTS = [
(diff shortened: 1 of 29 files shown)


tbraun96:
CI green (16/16 required; rsync-deploy skipped as expected). Verified end-to-end on the ship config (MTP bf16, bf16 KV, no prefix caching): dgx1 10/10 Sigma=2391s + dgx2 5/5 Sigma=1063s, 15/15 combined with zero corruption signatures. Images already pushed to avarok/atlas-gb10:{nightly,dev,latest} from this series. The residual prefix-caching x speculative interaction is tracked in #155 (serve-time warning included here). Merging.

Issue #110, opened 2026-06-01
[bug] SSM hybrid models crash at concurrency≥4 with deep context (CUDA-700 illegal access / SSM pool slot-leak)


tbraun96:
## Summary
On hybrid SSM models (verified with `Qwen/Qwen3.6-35B-A3B-FP8` + MTP), **concurrency ≥4 with deep (~8K-token) prompts crashes the server** and bricks it. Reproduced live on a GB10 (DGX Spark) with `avarok/atlas-gb10:latest`. This matches community reports of the model "falling on its face" at >1 concurrent request / long context (conc-4 @ depth → 0 tok/s).

## Repro
Serve `Qwen/Qwen3.6-35B-A3B-FP8` (FP8, MTP, `--max-seq-len 16384 --kv-cache-dtype fp8 --speculative`), then fire concurrent requests with a ~7–8K-token prompt:

| Workload | Result |
|---|---|
| single deep (~7K tok) | ✅ OK |
| concurrency 2–3 deep | ✅ OK |
| **concurrency 4 deep** | ❌ **crash** |
| concurrency 4 *short* prompts | ✅ OK (trigger is deep context, not concurrency alone) |

The crash form depends on the SSM state-pool size, which `= max_batch_size` (`crates/spark-model/src/model/impl_a1.rs:136`):

- **`max_batch_size=8` (default):** `Inference error: cuMemcpyDtoDAsync_v2 failed: status 700` (CUDA illegal memory access) in the SSM state D2D-copy path → CUDA context corrupted → all subsequent requests fail (`cuMemsetD8Async status 700`). Server must be restarted.
- **`max_batch_size=4`:** `alloc_sequence failed: SSM state pool exhausted (max 4 slots)` (`crates/spark-model/src/model/ssm_pool.rs:153`) — **and the slots leak**: every subsequent request (even a 2-token "hi") 500s with the same error. Server bricked until restart.

Each deep (chunked-prefill) request transiently consumes ~2 SSM state slots, so 4 concurrent deep requests exceed an 8-slot pool / exhaust a 4-slot pool.

## Root cause (two layered code bugs)
1. **Bounds/illegal access in the SSM state copy path** under concurrent deep chunked-prefill — `cuMemcpyDtoDAsync_v2` (likely `SsmStatePool::copy_slot`, `ssm_pool.rs:236`) writes/reads an out-of-range slot when ≥4 deep prefills are in flight (pool sized to `max_batch_size`).
[...]

PR #157, opened 2026-06-11
fix(#110): serialize mixed decode+prefill batch (cross-stream scratch race → CUDA-700)

tbraun96:
Root-causes and fixes the concurrency>=4 + deep-context CUDA-700 / brick (#110), found via a 43-agent forensic swarm (14/18 findings 2-lens verified). **Verdict: real engine bug, not a config constraint** — \`--max-batch-size 2\` is stable only because \`can_batch_mixed\` (>=2 prefilling AND >=1 active, minus single-active-spec) first fires at exactly 2 decode + 2 prefill.

## Root cause (RC1 + RC2)
The trait-default \`mixed_forward_batch\` runs the decode half on \`default_stream\` (hardcoded in \`decode_batch_compute_main\`) and the prefill half on the caller's \`prefill_stream\` with **no event ordering**. Both carve the same singleton \`buffers.scratch()\` (decode metadata at +32768) and \`buffers.logits()\`. At concurrency>=4 the prefill half's chunk-0 \`zero_all\` + MoE top-k staging clobber the decode half's metadata/logits mid-flight → torn KV-slot read → \`CUDA_ERROR_ILLEGAL_ADDRESS\` (sticky, surfaces at the next HtoD; ~32-min MTBF live), or silently corrupted decode logits (RC2, survives any serialization-only fix).

The **identical race was fixed for the single-prefill fused path on 2026-05-10** (\`decode_b.rs\`: metadata → \`logits+65536\`, one stream) and never propagated to the Q12 N-prefill batched path.

## Fix
\`TransformerModel::mixed_forward_batch\` override: both halves on \`default_stream\` (they cannot overlap anyway — shared arena — so this is the old serial default minus the race), and decode logits copied to a private \`decode_logits_staging\` buffer before the prefill half overwrites \`buffers.logits()\`. Neutral for decode-only / prefill-only / batch<=2 / EP.

## Verification
- Compiles clean; correct-by-construction + matches the proven sibling precedent + swarm 2-lens review.
- The CUDA-700 is a ~32-min-MTBF heisenbug (probabilistic torn read at the exact 2+2 interleaving) — not reliably reproducible in a short synthetic window, so this ships with a concurrent soak (in progress) rather than a deterministic before/after. Reviewer note below.
- Recipe \`max_batch_size: 2\` cap stays in place until this is reviewed + merged; lifting it is the payoff (full batch throughput restored).

@aceangel3k cross-stream + shared-arena change in the hot path — worth a look at the override in \`decode_b.rs\` and the staging-buffer sizing in \`buffers/sizes.rs\`. Independent follow-ups (separate PRs): RC3 eviction/refcount (the batch=4 brick), RC4/RC5 slot-compaction-targets-vec-position.

### DIFF
--- a/crates/spark-model/src/model/trait_impl/decode_b.rs
+++ b/crates/spark-model/src/model/trait_impl/decode_b.rs
@@ -28,10 +28,64 @@ use crate::layer::{
 };
 use crate::layers::ops;
 use crate::speculative::DraftProposer;
-use crate::traits::{ChunkedPrefillPageMetadata, Model, SequenceState};
+use crate::traits::{
+    ChunkedPrefillPageMetadata, MixedBatchResult, Model, PrefillSlice, SequenceState,
+};
 use crate::weight_map::{DenseWeight, MtpWeights, QuantizedWeight};
 
 impl TransformerModel {
+    /// #110: fully-batched mixed step (M decode + N prefill chunks).
+    ///
+    /// The previous trait DEFAULT (`Model::mixed_forward_batch`) ran the
+    /// decode half via `decode_batch` (which hardcodes `default_stream` in
+    /// `decode_batch_compute_main`) and the prefill half via
+    /// `prefill_batch_chunk` on the caller's separate `prefill_stream`, with
+    /// NO event ordering them. Both halves carve the SAME singleton
+    /// `buffers.scratch()` (decode metadata at +32768) and `buffers.logits()`.
+    /// At concurrency >= 4 the prefill half's chunk-0 `zero_all` + MoE top-k
+    /// staging clobbered the decode half's metadata/logits mid-flight ->
+    /// torn KV-slot read -> CUDA_ERROR_ILLEGAL_ACCESS (700), or silently
+    /// corrupted decode logits. (Same class fixed for the single-prefill
+    /// fused path on 2026-05-10; never propagated here.)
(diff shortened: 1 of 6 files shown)


tbraun96:
Update: a deterministic repro (conc=4, gpu-memory-utilization 0.55 to force eviction-during-prefill, staggered deep multi-chunk prompts) **falsifies this fix as the #110 crash fix**. The `mixedfix110` image built from this branch still bricks with CUDA-700 in ~1 min: `Q12 kernel-batched bailed -> ... Mixed-batch forward error (n_decode=3, n_prefill=3): status 700`.

The decode-vs-prefill stream serialization here is correct as far as it goes, but the actual fault is **inside the Q12 batched-prefill dispatch** (the kernel-batched path for >=2 concurrent prefills), not the decode/prefill split — so serializing those two halves doesn't prevent it. `ATLAS_BISECT_Q12_DISABLE=1` (which routes around the entire Q12 batched path) survives the same repro **80/80, zero CUDA-700**.

Recommend: do NOT merge this as the #110 fix. The RC2 logits-staging change here may still be a latent correctness improvement, but the crash fix is elsewhere. Interim verified fix: disable the Q12 batched dispatch by default (separate branch). Real fix: correct the batched-prefill kernel's internal scratch use — investigating now.

PR #161, opened 2026-06-12
AMD / Strix Halo (gfx1151) support

tbraun96:
## AMD / Strix Halo (gfx1151) support, re-landed on latest main

Clean re-land of the AMD port (supersedes the stale #160, which forked at e146beff and tangled ~20 FP8-campaign duplicates that have since merged to main independently). This branch is cut fresh off current main and carries only the AMD-isolated work as 15 logically-grouped commits. NVIDIA / GB10 is byte-unchanged throughout: every guarded kernel was verified PTX-identical (-arch=sm_121f -O3 --fmad=false) and the default build stays green.

### What this enables

- Atlas runs on AMD Strix Halo (gfx1151, RDNA3.5) via two first-class paths:
  - Native HIP (hipcc + libcuda->HIP shim), the primary target.
  - SCALE 1.7.1 (CUDA->AMD recompiler), experimental.
- Models: Qwen3.6-27B-FP8 (dense) verified coherent and serving on silicon; Qwen3.6-35B-A3B-FP8 (MoE flagship) enabled on the same path (guards + strix tree in place).

### Build / kernel abstraction

- Hardware-agnostic kernel build+load abstraction: ComputeTarget trait gains ScaleTarget and HipTarget alongside Nvidia/Apple. The existing dedup + parallelize compile loop in build.rs is now generic over the target, so the fast-compile path covers strix/HIP/SCALE the same way it already does CUDA, and extends to future targets without per-target loops.
- Kernel blobs unified to bytes (PTX text, ELF .o, and clang-offload .co bundles) with a per-blob binary sniff at registry load.

### Kernel guards (all NVIDIA-verbatim in #else)

- gfx1151 has no FP8 tensor-core MMA and a 64KB hard LDS cap. The 6 guarded GB10 sources plus the model-specific 27B/35B kernels get `#if defined(__SCALE__) || defined(__HIP_PLATFORM_AMD__)` blocks: software E4M3 encode/decode (SCALE's __NV_E4M3 cast is non-standard on gfx1151), e4m3->2xbf16 MMA, and single-buffer LDS.
- Software-E4M3 applied to both prefill (GEMM) and decode (GEMV) paths. The decode-path miss was the root cause of the garbage-from-2nd-token regression; now fixed across w4a16_gemv, w4a16_gemv_fused, dense_gemv_fp8w, moe_expert_gemv.

### Runtime fits on the APU

- FP8-source eviction after NVFP4 requant: the dense and MoE loaders freed the BF16 intermediate but kept the FP8 source co-resident, leaving the 60GB unified pool too tight for KV. Freeing the FP8 source on strix recovers ~27GB (KV free 6 -> 21.2GB).
- GDN prefill routed to the low-smem split4 path on gfx1151 (WY32/WY4/persistent/FLA all exceed the 64KB LDS cap; only split4 keeps the GDN h-state in global memory). Both the full and the active recurrence prefill entry points are gated.

### Plumbing

- NCCL feature-split so single-GPU SCALE/HIP builds link without libnccl.
- cuGraphInstantiate ABI switch (SCALE/HIP lack cuGraphInstantiateWithFlags).
- Vendored cudarc 0.19.2 with a missing-CUDA-symbol stub fallback for the SCALE/HIP runtime.

### Verified on gfx1151

Dense Qwen3.6-27B-FP8 on SCALE, end-to-end and coherent at ~9 tok/s:

```
Q: What is the capital of France? Then count from 1 to 5.
A: The capital of France is **Paris**.  1, 2, 3, 4, 5
Q: Write one sentence about the ocean.
A: The ocean is a vast, mysterious expanse of saltwater that covers more than 70% of the Earth's surface and plays a crucial role in regulating our planet's climate.
```

Speed is split4-GDN-bound; a C=32 FLA prefill variant that fits under 64KB is the next perf lever. The 35B MoE serve validation on silicon is running in parallel and will be noted here when it lands.

### Notes

- Native HIP build+serve validation on silicon and the 35B coherence smoke are the remaining gates before merge.
- Closes out #160 once this supersedes it.


tbraun96:
Trimmed the vendored cudarc crate down to the only two backends Atlas uses (driver + nvrtc), removing ~190k lines of never-compiled FFI bindings (cublas, cublaslt, cudnn, cufft, cufile, cupti, curand, cusolver, cusolvermg, cusparse, cutensor, nccl, nvtx, runtime). The workspace dep is pinned to default-features=false with the minimal set. The driver symbol-resolution patch (the only functional Atlas change in the vendored crate) is untouched, so NVIDIA gb10 and the AMD/HIP paths are unaffected.

This drops the branch diff from ~249.8k insertions to ~50k.

CI is green in the no-nvcc stub env (fmt, clippy --workspace --tests, build/test --workspace --locked with 803 tests passing, and cargo doc --workspace --no-deps -D warnings).

Verified end to end on real Strix Halo (gfx1151) hardware, native HIP path, for both model classes:

| Metric | Dense 27B | MoE 35B-A3B |
| --- | --- | --- |
| Model config | 64 layers, 16 attn, 48 SSM, 0 experts | 40 layers, 10 attn, 30 SSM, 256 experts |
| Build (trimmed cudarc) | RC=0, 24.0 MB binary | RC=0, 24.6 MB binary, 89 kernels |
| Server | UP | UP |
| Coherence | correct (2, 3, 5, 7, 11, 13, 17 ...) | correct (2, 3, 5, 7, 11, 13, 17, and 19. ... Paris.) |
| Decode | 12.7 tok/s | 51.8 tok/s |

Both build, link, load, and serve coherently on the slimmed vendor crate. The trim is model agnostic (driver + nvrtc FFI only), now confirmed empirically across the dense model and the 256-expert MoE (run with ATLAS_FORCE_NVFP4_ALL, exercising the full grouped-GEMM expert path).

PR #162, opened 2026-06-13
vLLM-parity generation control: remove server-side loop bandaids, enforce context window, add bounded tool-completion guard

tbraun96:
## Summary

Server-side generation-control heuristics had become the dominant cause of premature `finish_reason="length"` truncations and degenerate behavior in interactive clients (Zed, opencode). They were originally added to compensate for FP8 degeneration loops that have since been fixed at the numerical level, so they no longer earn their cost. This PR removes them and moves Atlas to vLLM-compatible generation control, then adds back only the bounded, justified pieces vLLM itself ships, plus one clearly-labeled opinionated guard for non-continuing agent clients.

Four commits, each independently reviewable:

### 1. Remove server-side loop detectors, watchdogs, budgets, caps (`a2f33dc9`)
Turn now ends only on EOS, client stop strings, `max_tokens`, or tool-call end. Deleted, with their config/state/metrics/tests:
- request-level loop detector + spinning detection, and the `<tool_call>` logit-bias decay table
- inter-tool prose budget, post-think content cap, content-loop / fuzzy-repeat / thinking-loop watchdogs, SimHash semantic guard, F2 confidence early-stop, stream-side line-repeat watchdog
- tool-arg dedup, name-run cap, per-response tool-call cap, orphan-suppression streak ender, in-think XML leak cancel, tool-call envelope cap
- EOS suppression in `tool_choice=auto` (and the tool-EOS-escape workaround it needed); `required`/specific still force a call
- tool `max_tokens` cap and the implicit 90%-of-max_tokens thinking cap
- Phase-C rollback/re-steer machinery and the per-sequence SSM decode ring (frees its VRAM region)
- WatchdogParams plumbing and all `[behavior]` watchdog keys in every MODEL.toml

Thinking tokens now consume the generation budget (vLLM counts reasoning toward `max_tokens`).

### 2. vLLM-parity thinking budget + opt-in repetition_detection (`38eb68ec`)
- `max_thinking_budget = 2048` operator default in the qwen3.6-35b MODEL.toml (equivalent semantics to vLLM PR 37112; request `thinking_token_budget` and CLI override win; 0 = unset). Bounds the deep-context overthink case where a trivial side request reasons thousands of tokens.
- Faithful port of vLLM v0.17.0 `SamplingParams.repetition_detection` (same struct, field names, validation, end-anchored check, `finish_reason="repetition"`). Opt-in per request, default off.

### 3. Enforce the context window (`17bba505`)
Clamp the generation budget to `min(max_tokens, max_seq_len - prompt_len)`, exactly as vLLM caps to `max_model_len - num_prompt_tokens`. Without this, a prompt near the window (a deep agentic turn) could generate tens of thousands of tokens past `max_seq_len` into out-of-window positions where the model degrades into a repeating loop (observed: 18,945-token loop, prompt 63,699 in a 65,536 window). Shared unit-tested helper at both API entry points.

### 4. Opinionated tool-completion guard (`b21abc11`)
OPINIONATED, EXPERIMENTATION-DRIVEN, NOT strict vLLM parity. Default on, kill-switch `ATLAS_TOOL_COMPLETION_GUARD=0`. In `tool_choice=auto` the model occasionally ends a tool-active turn by stating an action then emitting EOS with no tool call (valid in vLLM auto mode). Clients that auto-continue (opencode) mask this; clients that do not (Zed) stop the session early. The guard makes Atlas robust to non-continuing clients, bounded so it cannot recreate the unbounded-suppression trap:
- post-sample: a bare EOS in an eligible turn (tools active, no tool call yet) is discarded and starts a 128-token window
- pre-sample logit stage (both decode paths): EOS masked while active so the sampler is forced to a non-EOS token and advances to the intended tool call
- a completed tool call turns it off immediately; past the window EOS is allowed; suppressed EOS consumes budget so it cannot outrun `max_tokens`

## Verification
- webserver_ok N=10 gate: 10/10 on commit 1 (sum wall 1405.6s, beats the vLLM reference 1438.6s); 9/10 on commit 2 (the single miss was the model premature-EOS that commit 4 addresses).
- Context-window clamp reproduced and confirmed with a synthetic 60,805-token prompt: generation stopped exactly at the window with `finish_reason="length"` instead of running away.
- Tool-completion guard live-verified end to end in a remote Zed agentic session.
- 412 unit tests pass, including new clamp tests, repetition-detection tests, and the updated logit-pipeline stage assertions.

## Notes for reviewers
- The N=10 gate on the final combined build (commit 4) has not been run yet; recommend running it before merge rather than relying on CI alone.
- The thinking-budget and repetition_detection follow vLLM's surfaces deliberately; the tool-completion guard is the one intentional departure and is gated behind a kill-switch.


tbraun96:
@tbraun96 reviewed 162. direction is right — the clamp is clean and unit-tested, the repetition_detection port matches vLLM's end-anchored algorithm and validation, and the -5.3k LoC is a big win. blocking items before this can land:

1. cargo fmt is red: compact.rs, emit_step.rs, api/chat/mod.rs, logit_processors/mod.rs, chat_request.rs. quick fix.
2. branch no longer merges — 10 files conflict with main. since 06-14 main took #183 (MTP verify rework), #208 (grammar budget close), and b3c7df74 which added a NEW fp8 tool-repetition bound on tool_choice=auto. that last one is exactly the class of bandaid this PR removes — rebase needs an explicit call: does the parity work supersede b3c7df74 or keep it?
3. qwen3.6-27b MODEL.toml lost max_thinking_budget entirely and the in-code default went 256→0 (unset), while the 35b got 2048. with thinking_default=true and thinking now consuming max_tokens, a modest max_tokens request can burn the whole budget in <think> and return empty content. intentional, or should 27b carry the 2048 operator default too?
4. the tool-completion guard fires on every legit final-answer turn of a tool session (gates are just tool_request && !tool_call_completed), then masks EOS for 128 tokens — up to 128 tokens of forced trailing output or a spurious tool call on every session's last turn, default on. consider gating on something stronger or making it per-model opt-in.
5. small: suppressed EOS pops output_tokens but not logprobs_data (pushed right before), so blocking-path logprobs desync by one per suppression. pop both.
6. /v1/completions silently dropped the repetition_detection param — vLLM accepts it there and current main does too.

also per your own PR notes the N=10 webserver_ok gate was never run on the combined commit-4 build — needs to happen on the rebased branch before merge. happy to run it on dgx2 once you push the rebase + fmt fix.

SeedSource:

### Triage
- Risk: **med** - serve-test-required (triage advisory): `False` - path-heuristic (authority): `True`
- Rationale: Removes behavior configuration fields from kernel build and codegen scripts.
- Serving-path files: `crates/atlas-kernels/build.rs, crates/atlas-kernels/build_codegen.rs, crates/atlas-kernels/build_parse.rs, crates/atlas-kernels/src/lib.rs, crates/spark-model/src/model/impl_a1.rs, crates/spark-model/src/model/trait_impl/mod.rs`

### Serve gate
- Build: 26s - Qwen3.6-35B-A3B-FP8 on GB10 node1
- Top-k KL vs main: **0.1785** - **large behavioral shift** - decode delta **+0.1%**

### GLM-5.2 code review
**Correctness concerns:**

1. **`max_thinking_budget` default changed 256→0.** This is a behavioral change, not just bandaid removal. If any model relied on the implicit 256 default without an explicit MODEL.toml value, thinking budget silently drops to zero. This needs explicit confirmation that all shipped MODEL.toml files set this field, or it's a regression.

2. **No runtime/kernel-side removal shown.** The diff only touches build-time codegen and struct definitions. The actual loop-watchdog, confidence-early-stop, and rollback-resteer *consumers* in `decode_logits_step.rs` and elsewhere aren't in this diff. If those code paths still reference removed fields, this won't compile. If they're being removed in a companion diff, this PR is incomplete on its own.

3. **`max_post_think_content_tokens` removal is the highest-risk change.** The doc comment explicitly identifies it as the backstop against a known 360s-timeout runaway. Removing it without a replacement guard (the PR title claims "bounded tool-completion guard" but none is visible here) reintroduces the failure mode.

**Risk:** Medium-high. Silent behavioral defaults changed; claimed guard not evident in diff.

**Recommendation:** Block until (a) companion runtime diff is shown, (b) all MODEL.toml files audited for `max_thinking_budget`, (c) replacement tool-completion bound demonstrated.

tbraun96:

```
-253,486 lines deleted
```

- It is **CONFLICTING**, and GitHub does not run merge-commit workflows for a PR it cannot merge — so every green tick displayed here describes a tree from 2026-07-04 that nobody will merge. Thirteen open PRs are in that state; #427 adds a `needs-rebase` label so the false green stops being invisible.

PR #171, opened 2026-06-15
fix: support codex namespace tools

0137:
## Summary

Fix compatibility issues when Codex-style clients use the OpenAI-compatible Responses flow.

This allows Atlas to accept `type: "namespace"` tools, map Codex `developer` messages to a chat-template-compatible role, and strip namespace prefixes from model-emitted tool call names before returning OpenAI-compatible function calls.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --tests`

## Notes for reviewers

This is primarily a compatibility change for Codex-style clients using the Responses API.

`type: "namespace"` tools are accepted in Responses API input and converted to the existing OpenAI-compatible function tool shape, so the rest of the chat-completions path does not need separate namespace-tool handling.

Codex `developer` messages are mapped to a chat-template-compatible role to avoid tokenizer failures in strict templates that reject unknown roles or non-leading system messages.

Tool parser normalization is intentionally centralized in `normalize_tool_name()` so existing salvage behavior such as `name="Write"` and `Bash=Bash` remains in one place while also stripping generic namespace prefixes such as `namespace:tool_name`.

## CLA


0137:

SeedSource (approved):
Codeowner review — code-quality LGTM.

Nice consolidation: the per-parser ad-hoc name cleanup (quote-strip, `name=` handling, `=`-dup salvage, `namespace:` strip) collapses into a single `normalize_tool_name()` applied uniformly across the mistral / gemma4 / qwen3-coder / hermes / bare-identifier / tag-style / minimax-xml / json-fallback / pipeline paths. Removing the model-specific inline namespace hack in `parse_dispatch.rs` is the right call, and the two new tests (`parse_qwen3_coder_namespaced_function_name`, `parse_hermes_namespaced_function_name`) cover both the XML and JSON entry points. This is squarely on the path our 35B agent lane cares about (`google:google_search` / `browser:search` leakage), so it's welcome.

Two minor things, neither blocking:

1. `chat_message.rs`: the comment says "Treat developer instructions as system instructions" but the code maps `"developer" => "user"`. Semantically `developer` is closer to `system` (higher priority than user). Mapping to `user` loses that priority for any template that does understand a system role. If most target templates lack a system role anyway this is fine — worth a one-line note on why `user` over `system`.

2. CI shows only 2 checks on an all-Rust diff (other PRs run ~20). Likely the branch needs a fresh push to trigger the full rust suite — please confirm `cargo fmt` / `clippy` / test all green before merge, since this touches the tool-call hot path.

Approving on code quality. Functional merge still rides the standard serve-gate.

tbraun96:
@0137 thanks — the normalize_tool_name() consolidation is a real cleanup, collapsing four ad-hoc salvage paths into one place, and the six per-format tests are exactly what we want. this is on the path our agent lane cares about (google:google_search leakage), so we want it in. three things before merge:

1. correctness: is_tool_name_or_namespace_char now lets ':' into the bare-identifier scanner (parse_bare_identifier_json_calls) and the fallback-00 id scan in parse_dispatch, but normalize_tool_name only strips when there's a non-empty tail after the colon. so prose like `json:{"a":1}` or the classic malformed `tool_call:{"name":"get_weather","arguments":{...}}` now parses as a phantom call literally named `json:` / `tool_call:`. that silently deletes the json from content, preempts fallback-3 which on main parses the tool_call:{name,arguments} pattern correctly, and with exactly one declared tool the fuzzy repairer's single-tool fallback promotes the phantom into a real call. small fix: reject a candidate whose normalized name still contains ':' — legit ns:tool names normalize colon-free, trailing-colon prose labels don't. a test pinning `tool_call:{"name":...}` to the fallback-3 parse would lock it in. same guard applies to BareMistralNamePass in pipeline.rs.

2. ci: branch is behind main and only CLA + gitguardian ran. can you rebase on main and push so the full rust suite (fmt/clippy/build/test) triggers?

3. question on the namespace lowering: is codex's type:"namespace" wire shape really flat {type,name,description,parameters}? if a namespace tool nests sub-tools, the reshape silently produces one parameterless tool named after the namespace (parameters is Option + serde default, so nothing errors). if you have a codex-side request dump, paste a sample in the pr so we can pin a test to the real shape.

optional nit: there's already a private normalize_tool_name in tool_parser/fuzzy_match.rs with different semantics (separator collapse for fuzzy repair). renaming yours to strip_namespace_prefix or similar avoids the trap.

fyi we built your branch on a GB10 spark and ran the full spark-server suite — 530/530 green, and the no-flags serve path is logit-identical to main (KL gate 799/799 positions, mean KL 0.000000), so the consolidation itself is clean. its just the phantom-name edge above holding this back.

tbraun96:
@0137 pushed the ':' guard from the review onto your branch (4769b761) with tests pinning json:{...} -> no call, ns:tool -> tool, and tool_call:{"name":...} -> extracted call — pull before your next push. your six format tests stay green (551 total). running the spark gates against your head now; will merge this PR when they pass.

0137:
@SeedSource 
Thanks for the review! I updated the comment to clarify why developer is intentionally mapped to user instead of system, and pushed a follow-up commit.

0137:
@tbraun96
Thanks for the the review and for the fix!

PR #174, opened 2026-06-16
fix(amd/hip): FP8 MoE coherence on gfx1151 + 35B-A3B strix-hip-real pack

tbraun96:
## What

Fixes FP8 MoE coherence on the native-HIP (no-SCALE) gfx1151 / Strix Halo target so Qwen3.6-35B-A3B (MoE) serves coherently, and adds its strix-hip-real kernel pack. The dense 27B already worked (it runs the NVFP4 path); this fixes the FP8-native paths the MoE model uses.

## Root cause

`load_fp8_block_scaled_as_fp8weight` widens the per-128 block scale to a real FP32 buffer, but the CUDA to HIP kernel ports read it as `__nv_bfloat16` + `__bfloat162float`. The 2-byte vs 4-byte stride mismatch plus the bit reinterpretation corrupted every FP8 weight scale, producing garbage output. The dense 27B never dispatches these kernels, which is why it was unaffected.

## Changes

Scale-dtype fixes (block scale `__nv_bfloat16` -> `float`, matching the loader):
- moe_shared_expert_fused_fp8, _batch2, _batch3
- w8a16_gemv, w8a16_gemm, w8a16_gemm_t (incl. transpose_block_scale)

Other:
- moe_fp8_grouped_gemm: replace the no-op stub (only reachable on the dense target) with a real WMMA implementation mirroring the gb10 semantics via the w8a16_gemm gfx1151 WMMA idiom, reading FP32 scales. Restores coherent long-prefill (>64 token) MoE output.
- add widen_block_scale_f32 (referenced by the FP8 SSM path with no source in tree)
- add cuda_fp16.h HIP compat shim (needed by vision_encoder)
- add the qwen3.6-35b-a3b strix-hip-real kernel pack; vision_encoder ported to HIP (64-bit __shfl_down_sync mask, -ffp-contract=off)

## Verification (Strix Halo gfx1151, FP8 checkpoints, MTP-fp8)

Both models coherent with MTP:
- Qwen3.6-27B (dense, NVFP4 base): TTFT ~2.0s, decode ~18-20 tok/s
- Qwen3.6-35B-A3B (MoE, FP8-native): TTFT ~260-370ms, decode ~32-36 tok/s; short and long (>64 token) prompts coherent

Sample 35B outputs: "Say hello." -> "Hello! How can I help you today?"; "What is 2+2?" -> "Four"; capital of France -> "Paris".

NVFP4 path is unchanged (additive). Base: port/amd-strix-halo.


tbraun96:
Verified end-to-end on Strix Halo (gfx1151) before merge. Both models coherent with MTP-fp8:
- Qwen3.6-27B (NVFP4/rocmfp4 base): TTFT ~2.0s, decode ~18-20 tok/s
- Qwen3.6-35B-A3B (FP8-native MoE): TTFT ~260-370ms, decode ~32-36 tok/s; coherent on short and long (>64 token) prompts, the latter exercising the newly ported grouped GEMM.

The failing CI checks (cargo clippy/fmt/test, the 500-LoC cap on 22 files, and typos in bench/*.md) are pre-existing on the base branch port/amd-strix-halo and are unrelated to this change. This PR only adds .cu/.toml/.h files, and CI runs with ATLAS_SKIP_BUILD=1 so the kernels are not compiled in the workflow. The base branch's own latest run is red on the same checks.

PR #177, opened 2026-06-17
fix(qwen3.6): resolve agentic tool-calling degeneration — FP8 block-scaled prefill, thinking budget, tool-coerce

mission-deny-the-mission:
## Summary

Qwen3.6-35B-A3B (FP8) degenerated badly under agentic load via Claude Code / OpenCode — looping tool calls, empty/malformed tool args, literal `<tool_call>` tag leakage, and incoherent high-context behavior — while the same model behaved in vLLM. This PR fixes the root causes, which turned out to be **four independent bugs**. Validated end-to-end against real Claude Code agentic sessions. **No sampling-default changes** — `[sampling.*]` and the safety floors are left exactly as on main.

## Root causes & fixes

1. **FP8 prefill collapsed block scales** (`3283d1f`) — *the dominant cause.* This is a block-scaled FP8 checkpoint, but prefill fell through to single-scale `fp8_gemm_n128`, collapsing all per-128-block weight scales into one global scale. At long context this pushed tool-arg decode into the FP8 argmax-flip regime (the in-tree **B1 drift gauge climbed to ~1400**). The block-scaled prefill path (`fp8_gemm_t_blockscaled` + `moe_w8a8_grouped_gemm`) already existed but was gated behind `ATLAS_FP8_W8A8=1`. Made it the **default** via `ops::fp8_blockscaled_prefill_enabled()` (single source of truth, all six prefill dispatch sites: SSM qkvz + out_proj, attention q/k/v + o, MoE shared + routed), opt-out `ATLAS_FP8_SINGLE_SCALE=1`. Call sites still guard on block-scaled weights + kernel-handle presence, so models/builds without them fall back unchanged. **B1 drift dropped ~14× (1400 → ~100–200)**, empty-arg/duplicate-tool-call degeneration eliminated, loop-watchdog fires halved.

2. **Thinking-budget guillotine** (`aa77290`) — Anthropic `thinking: {type: "adaptive"}` (no `budget_tokens`, which Claude Code/OpenCode send) resolved to the conservative `DEFAULT_THINKING_BUDGET` (256), so the scheduler force-injected `</think>` mid-reasoning on agentic turns and wrecked tool selection. Now defers to the per-model `max_thinking_budget`, mirroring the existing `enable_thinking` path that already documented this exact 256-token mid-sentence cut.

3. **Empty-key tool args inside stringified nested arrays** (`96fb841`) — the qwen3_coder XML format serializes nested-array params as JSON text, so `ExitPlanMode.allowedPrompts` arrived as a string; `repair_empty_keys` ran *before* the string→array coercion and couldn't descend, so degenerate `{"": "Bash", …}` items survived. Added a second repair pass after coercion (+ regression test `repair_empty_key_in_stringified_nested_array`). Idempotent; the already-handled proper-array case is unaffected.

4. **EOS guard fired on non-tool turns** (`97cf50d`) — gated the post-think EOS suppression on tool-armed turns only.

## What this PR does NOT change

An earlier exploration deferred sampling to `generation_config.json` and dropped the `min_p` floor; that was **reverted** (`b35b0fc`). It was a global cross-model behavior change and, more importantly, moot once the FP8 block-scaled prefill fix landed — that was the actual cause of the degeneration, not sampling. Sampling presets, `min_p_floor=0.05`, `temperature_max=0.7`, and `thinking_in_tools=true` are all left at their main-branch values.

## Validation

Real Claude Code agentic sessions (file-heavy coding task), FP8, full context. Single-scale baseline → block-scaled prefill: B1 drift gauge **1400 → ~100–200 max**, B1 fires **14 → ≤2**, loop/watchdog **6 → 2**, tool-arg degeneration from pervasive (loops, empty args, tag leaks) → an occasional single fumbled multi-arg call. We also audited the NVFP4 path with the same rigor and found **no analogous fixable bug** — NVFP4 prefill correctly applies per-16 micro-scales; its lower fidelity is inherent 4-bit loss.

## Notes

- Includes a `bench/tool_format_eval/` harness (BF16 ground-truth comparison, endpoint diff, logit probes) used to localize these — can be dropped if you prefer a code-only PR.
- History contains the reverted sampling detour (`9937ee5`/`c961f0d` → `b35b0fc`) and one net-reverted `thinking_in_tools` flip (`d2485fc`); happy to squash/clean the history if preferred.

## Test plan

- [x] `cargo build --release -p spark-server --no-default-features --features cuda` (gb10, qwen3.6-35b-a3b/nvfp4 target)
- [x] `cargo test` tool_parser coercion suite (incl. new `repair_empty_key_in_stringified_nested_array`)
- [x] Live Claude Code agentic session on GB10 / DGX Spark — block-scaled prefill confirmed dispatching by default, degeneration markers cleared


mission-deny-the-mission:
> Thanks for the contribution. Before we can merge, please sign our [Contributor License Agreement](https://github.com/Avarok-Cybersecurity/atlas/blob/main/CLA.md) by replying to this comment with exactly:
> 
>    - - -
>    I have read the CLA Document and I hereby sign the CLA
>    - - -
>    <sub>You can retrigger this bot by commenting **recheck** in this Pull Request. </sub><sub>Posted by the **CLA Assistant Lite bot**.</sub>

tbraun96:
Thanks for the submission! I will go over this for some manual testing. In the meantime, feel free to reach out to me on discord, I do want to discuss this: https://discord.gg/qfXGe9rAu

mission-deny-the-mission:
> Thanks for the submission! I will go over this for some manual testing. In the meantime, feel free to reach out to me on discord, I do want to discuss this: https://discord.gg/qfXGe9rAu

What's your discord handle? I am actually in the discord if you need me. Username is areyouevenreal.

tbraun96:
@mission-deny-the-mission this is excellent root-cause work — the B1-drift localization to the single-scale fp8_gemm_n128 collapse is exactly the forensics we want, and we intend to land it. three asks before merge:

1. rebase on main. your EOS-guard fix (97cf50d) already landed independently via #183 with the same (require_tool_call || tool_request) gating, so decode_logits_step.rs conflicts — drop your hunk and take main's side. that's the only conflict.

2. cargo fmt is red: the new assert_eq! in tool_parser/tests/group_e.rs:391 needs a rewrap. cargo fmt --all fixes it.

3. harden the six force_w8a8 guards with fp8w.scale_format == WeightQuantFormat::Fp8BlockScaled (or add an is_block_scaled() helper on Fp8Weight). the struct docs require checking the tag before consuming row_scale, and now that block-scaled prefill is default-on, a per-row FP8 weight ([N] f32 scales — the Holo/Ornith per-channel checkpoints in #203/#210 land exactly that) would get its scales misread as [N/128,K/128] blocks and silently produce garbage. today every in-tree prefill loader tags BlockScaled so nothing breaks, but the dispatch should be self-protecting. note attention's set_fp8_weights has no scale_format assert, unlike the ssm setter.

nit: fp8_blockscaled_prefill_enabled() re-reads the env var at every dispatch across all layers per prefill — cache it in a OnceLock.

please squash on rebase (the reverted sampling detour can go). keep the bench harness, it's useful. once the branch is green we'll run the FP8 A/B + NVFP4 no-regression gate on a spark box and merge.

tbraun96:
@mission-deny-the-mission the rebase + hardening from the review is staged as branch fix/qwen36-agentic-rebased in this repo (clean 4-commit series on current main: your fp8 default-on + is_block_scaled() guards on all six sites, thinking deferral, tool-coerce; the EOS-guard hunk dropped since #183 landed it). adopt it onto your branch if it matches your intent — then we run the FP8 A/B + NVFP4 no-regression gates on a spark and merge. if you'd rather rebase yourself, the guard requirements are in the earlier review comment.

SeedSource:

### Triage
- Rationale: Documentation and benchmarking updates for FP8 tool-format evaluation; no kernel logic changes.
- Serving-path files: `crates/spark-model/src/layers/moe/forward_prefill_fp8.rs, crates/spark-model/src/layers/ops.rs, crates/spark-model/src/layers/qwen3_attention/prefill/paged_oproj.rs, crates/spark-model/src/layers/qwen3_attention/prefill/paged_qkv.rs, crates/spark-model/src/layers/qwen3_attention/prefill_weights.rs, crates/spark-model/src/layers/qwen3_ssm/trait_prefill_phase3.rs`

### Serve gate
- Build: 30s - Qwen3.6-35B-A3B-FP8 on GB10 node1
- Top-k KL vs main: **0.2371** - **large behavioral shift** - decode delta **-0.2%**

### GLM-5.2 code review
## Review: PR #177

**Block.** This PR adds only benchmark/analysis markdown + a Python eval script — **zero Rust/CUDA changes**. The title claims fixes ("resolve... FP8 block-scaled prefill, thinking budget, tool-coerce") but the diff contains none of them. The actual proposed fix lives in a markdown conclusion ("force the qwen3_coder tool grammar to engage") with no implementation.

**Correctness concerns in the analysis itself:**
- `GROUND_TRUTH_RESULT.md` claims "layer-truncated to 16 layers... bit-exact" — truncating to 16/40 layers does **not** produce bit-exact L0–15; RoPE/KV-cache and residual stream still depend on all prior layers' full computation. The "ground truth" methodology is unsound.
- The two markdown files **contradict each other**: `KERNEL_LOCALIZATION.md` concludes "the divergent kernel is Atlas's full-attention path"; `GROUND_TRUTH_RESULT.md` says "there is no kernel/numerical bug" and supersedes it. Both cannot ship.
- `README.md` Layer-4 data (vLLM 100%, Atlas 75% on identical FP8 weights, non-overlapping CIs) directly refutes the "no bug" conclusion.

**Risk:** Shipping contradictory root-cause docs will mislead future debugging. The eval script (`compare_endpoints.py`) is unreviewable for correctness — no tests, no error handling on endpoint failures.

**Required:** (1) Implement the actual fix in Rust/CUDA or retitle. (2) Resolve the contradiction. (3) Justify or retract the "bit-exact" claim. (4) Add basic error handling to the script.

tbraun96:
Hi @mission-deny-the-mission — heads up that #229 just merged to `main`, landing related tool-calling correctness work on the 27B path: the API-tools grammar repair (fixed an ST-995 55.78 collapse), penalty-free sampling presets (the 27B carried `repetition_penalty=1.1`/`dry_multiplier=0.5` which corrupted grammar-constrained tool-call args — restoring BFCL ST-995 to ~89), and GDN projections loading at native FP8.

Your #177 targets a distinct root cause — the 35B-FP8 prefill block-scale collapse (`3283d1f`) — so it may still be needed on top. Could you rebase on latest `main` and check whether that FP8 fix is still required, or whether there's overlap to reconcile? Happy to help verify on GB10.

tbraun96:
Superseded by recent merges. Apologies! We are working on a merge queue and PR tracking system. From now on, expect us to be better :D

PR #178, opened 2026-06-18
perf(mtp): auto-gate K=2 speculative decode off on FP8 weights

tbraun96:
MTP (K=2 verify) is measured net-negative on FP8 weights for the hybrid SSM flagship (~-20% decode: the verify pass re-runs the full layer/MoE/lm_head stack and draft acceptance is only ~30-41%, below the ~60-70% break-even), but net-positive on NVFP4. Until now MTP was enabled purely by `--speculative` with no awareness of the loaded weight quant format, so a launch command tuned for NVFP4 silently regressed decode when pointed at an FP8 checkpoint.

Derive the safe default from the already-resolved weight quant format (the SSOT `detect_quant_format(...).base_variant()` computed at serve init): on FP8 (`Nvfp4Variant::Fp8Dequanted`), downgrade `--speculative` to off and log the decision; on NVFP4 (and every other variant) the speculative flag is left untouched. Adds an explicit `--force-speculative` escape hatch so an operator can keep MTP on regardless of quant format. The per-request `disable_mtp` override is unaffected.

Additive and backward-compatible: NVFP4 behavior is byte-identical; the only behavioral change is the measured-negative FP8 + `--speculative` case.

Verified on GB10 (Qwen3.6-35B-A3B FP8 flagship), N=12 identical temp=0 decode-heavy requests, token-for-token-identical generations:
- baseline MTP-on (`--force-speculative`): 52.5 agg tok/s, 12/12 coherent
- gated MTP-off (plain `--speculative`): 61.3 agg tok/s, 12/12 coherent
- => +16.8% decode throughput, -14.7% wall, no coherence regression.

Re-verification (served coherence + before/after TTFT/decode) being reproduced on dgx1 before merge.


tbraun96:
Re-verified end to end on dgx1 (GB10) before merge, not just on CI.

Gate behavior (Qwen3.6-35B-A3B-FP8, plain --speculative): logs "MTP auto-gated OFF: FP8 weights detected..." and the scheduler starts with mtp=false, num_drafts=0. With --force-speculative it logs "MTP kept ON..." and starts mtp=true. NVFP4 path is untouched.

Decode A/B, N=12 identical temp=0 requests, 12/12 coherent both arms:
- MTP-off (gated):        62.3 tok/s
- MTP-on (--force-speculative): 41.8 tok/s

The gate is the clear win on FP8 (the verify pass re-runs the full layer/MoE/lm_head stack and FP8 acceptance is well below break-even). Direction matches the original N=12 (+16.8%); the larger delta here is just prompt-dependent forced-MTP acceptance on a single decode probe.

CI note: the +11 lines for --force-speculative pushed cli.rs from 491 to 502, over the 500-LoC cap (serve.rs at 812 is allow-listed and was not the cause). Fixed by splitting the ServeArgs clap struct into cli/serve_args.rs and re-exporting it as cli::ServeArgs, so call sites are unchanged (cli.rs is now 21 LoC). Rebased onto current main; the ≤500 LoC, fmt, clippy --tests, and cargo test --workspace checks are all green.

PR #185, opened 2026-06-20
DeepSeek-V4-Flash (NVFP4) support + sync with main

tbraun96:
## DeepSeek-V4-Flash (NVFP4) support + sync with `main`

Adds the **DeepSeek-V4-Flash** model target for GB10 (sm_121) and brings the
bring-up branch up to date with current `main`.

DeepSeek-V4-Flash is a 284B-total / ~13B-active hybrid: **MLA** attention
(compressed 576-dim KV: 512 latent + 64 RoPE), a **256-expert + 1-shared MoE**
with **sqrtsoftplus** routing, **DeepSeek Sparse Attention (CSA)**, and
**Manifold-Constrained Hyper-Connections (mHC)**. It uses MQA
(`num_key_value_heads=1`), so TP>1 is architecturally impossible — multi-Spark
runs use **pure EP** (`tp-size 1, ep-size 2`).

### What's here
- **MLA paged decode** kernels + dispatch for NVFP4 and FP8 compressed KV cache,
  interleaved **YaRN RoPE** (`compress_rope_theta`), grouped O-projection,
  `q_b_norm`/`kv_norm`.
- **sqrtsoftplus** MoE routing kernel + correction-bias top-k; **swiglu_limit**
  clamp in the FFN activation (the final fix that produced coherent output).
- **mHC** hyper-connection infrastructure (`hc_streams`/`hc_post`/`hc_comb`).
- `deepseek_v4` config parser, weight loader, chat template, MODEL.toml, and an
  EP=2 bring-up script (`scripts/start-deepseek-ep2-redhat.sh`).
- **Mixed-precision FP8/NVFP4 dequant**: hybrid path keeps `main`'s GPU
  `.weight_scale_inv` fast path **and** adds CPU fallback for `.weight_scale`
  (RedHatAI compressed-tensors) and `.scale`/F8_E8M0 (DeepSeek-V4) — required to
  load the checkpoints.

### Sync with `main`
Merged `main` (34 commits: MTP/hybrid-SSM spec decode #183, FP8 tool-runaway
bound, MoE grouped-GEMM grid-stride fix, FP32-routing buffers, split-K
determinism, RNE `f32→bf16`). Conflict/integration notes:
- Reconstructed the `kv_dtype` decode match: `main`'s full arm set + determinism
  split-K, plus V4-Flash MLA guard arms (`Nvfp4`/`Fp8 if is_v4_flash`).
- Kept DeepSeek sqrtsoftplus routing; took `main`'s real gate path + last-token
  expert-id dump over the debug forced-dense hack.
- Restored APIs the auto-merge silently dropped (`diag_norm_f32`, the Marconi
  `kv_write_floor` arg) and **removed the `fp32-residual` feature to match
  `main`** (which deliberately removed `use_fp32_residual`; no-op for V4).
- `docker/.../deepseek-v4-flash/nvfp4/Dockerfile`: `COPY vendor/` so the
  vendored `cudarc` path-dep resolves.

### Status
- ✅ Compiles clean (`cargo check -p spark-model -p spark-server`, cuda).
- ✅ Image builds; **EP=2 on 2× GB10 loads and runs** (kernel target
  `(sm_121, deepseek-v4-flash, nvfp4)`, EP topology `ep_rank 0/2`, experts split
  `[0,128)`/`[128,256)`).
- ⏳ Benchmarks (prefix-cache warm/cold, fp8 vs bf16 KV) in progress — will post.

### Notes / follow-ups
- **MTP not exercised**: the public RedHat checkpoint ships **0 MTP/nextn weight
  tensors** (despite the config declaring an MTP layer), so speculative decode is
  off. MTP-on isn't possible without draft weights.
- **`swiglu_limit` clamp** is currently unconditional in the *shared* gb10
  `moe_silu_mul` kernel. Benign for V4; should be gated on a parsed
  `swiglu_limit` config before this lands, to isolate it from other models.


tbraun96:
⚠️ **Status: regression found — converting to draft.**

The merge is mechanically complete (compiles, image builds, EP=2 on 2×GB10 loads + runs, prefix cache works — 98× warm/cold TTFT speedup, ~10.7 tok/s decode), **but generation output is incoherent** (repetitive token salad). The pre-merge branch produced coherent output on this exact `redhat-ds-v4` checkpoint, so the merge introduced a coherence regression.

**Root cause (diagnosed, not yet fixed):** `main` evolved the shared MoE expert kernels since the fork — `kernels/gb10/common/moe_shared_expert_fused_t.cu` and `moe_shared_expert_fused_fp8_t.cu` changed their expert **block-scale handling** (scale byte now decoded via `atlas_dec_e4m3()` instead of a raw `(float)` cast; FP8 variant's scale pointer `bf16*`→`float*`). DeepSeek-V4's MoE Rust dispatch prepares/passes block scales in the *old* format, so the merged kernels misread V4's expert scales → wrong expert weights → garbage.

**Ruled out:** the FP8 dequant hybrid (all 33k V4 scales are `.weight_scale` → CPU path = identical to pre-merge); my MoE-forward conflict resolutions (identical to pre-merge for V4); `moe_silu_mul.cu` swiglu clamp (untouched by main); MLA decode guard arms (byte-identical kernel calls to pre-merge).

**Fix direction:** reconcile V4's expert block-scale preparation with the new kernel ABI (provide E4M3/FP32 scales as the merged kernels now expect), or route V4's shared/routed experts to V4-specific kernel variants. Needs a build-test iteration.

PR #186, opened 2026-06-20
DeepSeek-V4-Flash (NVFP4) — clean port onto main [WIP]

tbraun96:
## DeepSeek-V4-Flash (NVFP4) — clean port onto current `main`

**Draft / WIP.** Supersedes the merge-based approach in #185 (which fused the old branch's V4 dispatch with `main`'s independently-evolved shared MoE kernels → incoherent output). This branch starts **fresh from `main` (81830ef)** and ports only the minimum V4 code on top of `main`'s current (correct) APIs, treating `main` as source of truth.

### Done (foundation — `atlas-core` compiles)
- [x] 35 net-new V4 files: `deepseek_v4` config parser, weight loader (`deepseek_v4/{assemble,compute,load_layers}`), `attention_forward_v4` / `cache_skip_v4` / `paged_v4`, `hyper_connection` op, the full `kernels/gb10/deepseek-v4-flash/nvfp4/` kernel dir, MODEL/KERNEL toml, jinja template, EP=2 scripts.
- [x] 13 V4 files `main` never touched, taken verbatim (multi-seq MLA, moe_gate, etc.).
- [x] Config layer: `ModelConfig` V4 fields + factory defaults + `deepseek_v4` dispatch arm + parser registration.
- [x] Registration: spark-model factory, weight_loader exports, `ops` hyper_connection, prefill mod wiring, MoE sqrtsoftplus kernel handles + init.

### Remaining (careful forward-path port onto `main`)
- [ ] `qwen3_attention/types.rs` — MlaWeights `wo_a/wo_b/compressor/attn_sink`, `Compressor`/`Hc*` types, ~10 layer kernel/state fields; `init.rs` construction + V4 kernel loads.
- [ ] `decode_inner` / `prefill_inner` / `multi_seq` — V4 MLA + mHC dispatch.
- [ ] `run_paged_decode` / `attention_forward` — MLA guard arms.
- [ ] `quant_helpers` `.weight_scale`/`.scale` (E8M0) dequant; `buffers` hc/o_latent/norm_unit_w.
- [ ] `ops/kv_cache` MLA wrappers; `moe/forward{,_batched}` sqrtsoftplus.
- [ ] `model/trait_impl` decode_a / prefill_{a,b,c,d} V4 hooks.
- [ ] Shared kernels: `moe_silu_mul.cu` swiglu_limit (config-gated this time) + `rope.cu` YaRN interleaved.
- [ ] Build image → EP=2 bring-up → verify COHERENT output → benchmark.


tbraun96:
### Status: clean port is structurally complete; one isolated coherence bug remains

**Compiles, builds, loads, dispatches.** `cargo check` clean (spark-model + spark-server); `atlas-deepseek-v4:clean` image builds (150 V4 kernels incl. the added `rope_forward_yarn_interleaved`); EP=2 on 2×GB10 loads the redhat checkpoint and selects the `(sm_121, deepseek-v4-flash, nvfp4)` kernel target with EP topology. Conservative port verified: `run_paged_decode`/`quant_helpers`/`decode_inner` correctly favor-main + gate V4; the one over-port (`embed_chunk` re-introducing `use_fp32_residual`) was caught and reverted.

**Coherence still broken — but now precisely isolated.** Output is the same `package package Irrelevant_` salad as the merge build. Crucially:
- The clean port has **no swiglu clamp** and the merge had it → **same garbage either way, so the clamp is NOT the cause** (and was correctly left out per favor-main).
- The garbage is identical in merge and clean, which share one thing: **main's shared-code evolution (7a903aa→81830ef)**. The pre-fork branch (old-main shared kernels + identical V4 code) produced coherent output.

**Leading hypothesis (static evidence):** main changed the shared `kernels/gb10/common/moe_shared_expert_fused_t.cu` / `_fp8_t.cu` block-scale decode (`atlas_dec_e4m3(byte)` instead of a raw `(float)` cast; FP8 variant scale ptr `bf16*`→`float*`). V4's **always-on shared expert** routes through this shared kernel (it has its own routed-expert kernel but not its own shared-expert kernel), and its Rust scale-prep provides the old format → every token's FFN is corrupted via the residual. main's own models feed the new E4M3 format, so they're fine; V4 needs its shared-expert scale-prep aligned to the new kernel ABI (favor-main keeps the kernel, fixes the caller).

**Next:** localize via `ATLAS_DIAG_V4_ALL_LAYERS` (no rebuild) to confirm MoE-FFN vs attention, then align V4's shared-expert scale preparation to main's kernel ABI.

tbraun96:
### A/B vs the 'coherent' backup: the clean port is VALIDATED; the coherence bug is PRE-EXISTING

Built and ran the backup branch `ccbeb70` ('achieves COHERENT output') from a fresh image on the same 2×GB10 + same checkpoint, with identical diagnostics:

| build | output | mHC `hc_post-ffn ALL_STREAMS` L42 |
|---|---|---|
| backup `ccbeb70` (the 'coherent' commit) | `IrregularlyIr#package…` (garbage) | 277,160 |
| clean port (this PR) | `package package…` (garbage) | 2,300,000 |

**Both are garbage, both show the mHC residual streams blowing up** (manifold-*constrained* streams are supposed to stay bounded). The checkpoint is unchanged since May 28; the only Jun-19-era image is gone. So:

1. **The clean port did NOT introduce the bug** — it faithfully reproduces the backup reference's behavior. The conservative favor-main port is correct (compiles, builds, loads, dispatches all V4 kernels, byte-equivalent V4 logic).
2. The coherence bug is **pre-existing in the original DeepSeek-V4 bring-up code**: the mHC (hyper-connection) residual streams grow unbounded across layers (driven by the MoE-FFN `hc_post` contribution), corrupting the final hidden. The `ccbeb70` 'coherent' claim was on short answers with a now-deleted image and does not reproduce from current code + checkpoint + fresh build.

**Conclusion:** this PR is the correct clean foundation and is safe to land as the V4 port. Getting coherent generation is a **separate** model-bring-up debugging effort on the mHC Sinkhorn-constrained mixing (it isn't bounding the streams) — present in the original code, not caused by the port.

tbraun96:
### Diagnosis cracked: authoritative reference found + complete divergence roadmap

A 5-agent web-research workflow located the **authoritative reference** — `transformers v5.12.0` ships `modeling_deepseek_v4.py` (DeepseekV4ForCausalLM). Diffing Atlas against it turned the coherence mystery into a concrete checklist. The garbage is a **multi-bug from-scratch-port divergence** (not the clean port, not env), pre-existing since the original bring-up.

**Fixes landed in this PR (all real, verified bugs):**
- ✅ **Router gate kept BF16** — Atlas was re-quantizing the BF16 `ffn.gate.weight` to 4-bit NVFP4 and routing via `w4a16_gemv`; a 4-bit router flips low-margin top-6 selection on every token. (`d146541`)
- ✅ **`ffn.gate.bias` loaded as the noaux_tc correction bias** — Atlas looked for the wrong tensor name and supplied a zero bias → wrong expert selection on layers 3-42. (`3ddd353`)
- ✅ **mHC streams stored FP32** + exact doubly-stochastic Sinkhorn comb. (`c6edefe`)

**Remaining confirmed divergences (each a follow-up; ranked):**
1. **Attention-output inverse RoPE (eq.26) — MISSING.** Reference: `attn_output = apply_rotary(attn_output, cos, −sin)` before the grouped O-proj; Atlas goes attn_out → o_proj directly → position-scrambled attention on every token. Prime remaining suspect.
2. **Hash-MoE layers 0-2 — UNIMPLEMENTED.** `num_hash_layers=3`; reference uses a frozen `tid2eid` [vocab×6] token→expert table (present in the checkpoint) for *which* experts; Atlas runs ordinary learned-gate TopK → wrong bootstrap experts.
3. **Per-head attention sink (`s_aux`)** present in the compressed-attention kernel; confirm it's in the regular MLA decode/prefill softmax denominator too.
4. Verify: routing gathers RAW scores (not score+bias) for weights; routed_scaling_factor=1.5 applied to routed only; interleaved RoPE pairing.

**Reference:** https://github.com/huggingface/transformers/blob/v5.12.0/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py

The DecoderLayer.forward, HyperConnection, TopKRouter, and MLA attention reference equations are captured for each fix. The port (this PR) is validated-correct; reaching coherent generation requires items 1-3, each a ~30-min GB10 build/verify cycle.

tbraun96:
### Exact implementation spec for the 3 remaining attention/routing fixes (reference-verified)

All confirmed against transformers v5.12.0 `modeling_deepseek_v4.py`. The fixes are LOADED-but-unused or missing; here is the precise change set.

**A) Per-head attention sink (`s_aux`).** Weight is already loaded: `attn.attn_sink` [64], in `MlaWeights.attn_sink` (assemble.rs:185). Reference `eager_attention_forward`: append per-head sink to the softmax logits, softmax over [weights | sink], then DROP the sink column → mass leaks to the denominator only. Online-softmax form: at finalization add `final_l += expf(sink[q_head] - final_m)` BEFORE `inv_l = 1/final_l`. Thread `sinks` ptr into: `mla_paged_decode_fp8.cu` (@~295), `mla_paged_decode.cu` (nvfp4), and the full-attention prefill MLA kernel. (The compressed/CSA prefill kernel `prefill_attn_compressed.cu` already has it.) Update the `mla_paged_decode_{fp8,nvfp4}` wrappers (ops/kv_cache.rs) + the run_paged_decode guard arms to pass `mla.attn_sink`.

**B) Attention-output inverse RoPE (eq.26) — MISSING.** Reference: `attn_output = apply_rotary(attn_output, cos, −sin)` BEFORE the grouped O-proj. attn_output is [nq, 512] per head with the trailing 64 = rope (kv used as both K and V, RoPE'd on its trailing 64). Implement: after `run_paged_decode` (attn_out) and before the wo_a grouped GEMV (attention_forward_v4.rs:399, paged_v4.rs, cache_skip_v4.rs), de-rotate the trailing 64 dims of each head with the interleaved YaRN RoPE using NEGATED sin. Add a `-sin` variant/flag to `rope_forward_yarn_interleaved` (rope.cu) or a dedicated kernel; reuse the mla_q_rope extract/writeback at offset nope=448.

**C) Hash-MoE layers 0-2.** `tid2eid` [129280,6] I64 is in the checkpoint (`layers.{0,1,2}.ffn.gate.tid2eid`). Reference HashRouter: `indices = tid2eid[input_id]` (per token), `weights = sqrtsoftplus(gate_logits).gather(indices)`, normalize, ×routed_scaling_factor. Load tid2eid for the 3 hash layers, thread input_ids to the MoE forward, and branch the routing (skip top-k, use the table) for those layers.

**Reference equations** (attention forward, eager_attention_forward+sinks, apply_rotary interleaved, TopK/Hash router, DecoderLayer mHC wiring) are all captured above. Fixes A+B are the pervasive attention bugs (every token); C is the 3-layer bootstrap.

aceangel3k:
I've noticed that there are a lot of coherence issues for DeepSeek V4 Flash. I haven't verified using Redhat AI's NVFP4 version yet. I wonder if it even produces output in vLLM?

SeedSource:
Hey @tbraun96 — we dug into the V4-Flash coherence bug here (2×GB10 setup, and it's in our wheelhouse). Ran an independent divergence audit of `deepseek-v4-clean` against the authoritative `transformers v5.12.0 modeling_deepseek_v4.py`. The actionable bits:

**The mHC stream "blow-up" is a red herring.** The hyper-connection math matches the reference exactly — `comb` columns are pinned to sum to 1 and applied over that axis, so each layer's mix is a convex combination (non-expansive). The rising stream norms across depth are the expected eigenvalue-1 behavior (which is why the highway is FP32), not the corruption source. So the L42 277K→2.3M growth isn't the bug.

**Cheapest next check (5 min, no rebuild) — likely the real keystone:** is the final `model.norm` applied before the LM head for the V4 path? Reference does `self.norm(self.hc_head(hidden))`; in the branch `hc_head` (last layer) writes `hidden` and relies on the shared post-loop `final_norm`. If that norm isn't hit for V4, the (correctly large) collapsed hidden reaches `lm_head` unnormalized → garbage on every path (would explain garbage in *both* the backup and the clean port). Your existing `MTP_DEBUG_NORMS` print (`||input_hidden|| ||final_normed|| ||logits||`) confirms it directly.

**Confirmed divergences (file:line + fix), assuming the norm checks out:**

1. **NVFP4 KV-decode is missing the per-head attention sink.** `run_paged_decode.rs:58` calls `mla_paged_decode_nvfp4(...)` without `attn_sink`, while the FP8 arm (`:109`) passes it and the FP8 kernel applies it (`mla_paged_decode_fp8.cu:317-322`). NVFP4 KV is the default → prefill correct, decode degrades token-by-token (softmax denom too small). The NVFP4 kernel (`mla_paged_decode.cu`) has no sink param at all. Fix = thread `mla.attn_sink` in + add `if (sinks) final_l += __expf((float)sinks[q_head] - smem_m[0]);` right after `final_l = smem_l[0];` (line 342), mirroring the FP8 kernel. Prime suspect for the NVFP4 generation salad.
2. **FP8 grouped-prefill (`forward_prefill_fp8.rs`) + `forward_k3.rs` skip the hash-route + sqrtsoftplus branches** — on FP8 expert tables with prompt chunks >64, every layer (incl. hash 0-2) routes via plain sigmoid top-k → wrong experts/weights. Mirror the `forward_prefill.rs:234-248` dispatch.
3. **`shared_expert_gate`** wrongly sigmoid-gates the ungated shared expert if that tensor is present (the v5.12.0 shared expert is a plain add) — currently safe only because the ref checkpoint omits it; worth reconciling the contradictory in-code comments.

**Validated as correct (don't touch):** router gate BF16, noaux_tc bias (selection-only, raw-score weights), inverse-RoPE eq.26 (conjugate/interleaved/trailing-slice/before-o_proj — all 3 paths), `wo_a` block-diagonal grouped O-proj, and the entire mHC mechanism + FP32 highway.

Suggested order: final-norm check → NVFP4-decode sink → FP8/routing. Happy to implement the NVFP4-sink fix — it's a clean 3-file change mirroring your FP8 kernel — and PR it onto `deepseek-v4-clean` if you'd like; just say the word so we don't collide. Full bug map with exact edits on hand if useful.

SeedSource:
@tbraun96 — **V4-Flash serves coherent on 2×GB10. Let's land this.**

Stood up `nvidia/DeepSeek-V4-Flash-NVFP4` on 2×GB10 (EP=2, **fp8-KV**, RoCE dual-rail verified via NIC counter delta) today off this branch + a local decode-sink patch:
- `"2+2"` → **4**
- `"Write a Python function that adds two numbers"` → clean, correct `add_numbers(a, b)`

Your non-CSA (full-attention) **prefill** sink + the mHC fixes are doing their job. We have a complementary **NVFP4-decode** attention-sink fix on `fix/v4-nvfp4-decode-sink` — it mirrors your prefill sink into the NVFP4 MLA decode kernel (threads the per-head sink into the softmax denominator). Yours to cherry-pick.

Two FYIs for the serve recipe:
1. **fp8-KV is the coherent path.** `--kv-cache-dtype nvfp4` hits a separate `CUDA_ERROR_MISALIGNED_ADDRESS` in `alloc_sequence`, and the MODEL.toml already recommends fp8 — so the decode-sink fix is correct-by-construction but nvfp4-KV has its own alloc bug to chase later.
2. The checkpoint is the official ModelOpt `MIXED_PRECISION` export (NVFP4 experts + FP8 elsewhere) — clean, not the issue.

Status-wise this is **one trivial conflict** (`crates/spark-model/src/weight_map/quant_helpers.rs`, just main drift over the 10 commits) away from mergeable.

We have the GB10 gate hardware, so we can **run the serve-coherence verify on every push** and resolve that conflict whenever you want. What's left on your side to take #186 out of draft? Happy to do the GB10 legwork to get V4 live for users.

SeedSource (approved):
Approving to ship DS4F to all Atlas users (Azeez greenlit driving this).

Drove #186 to merge-ready: resolved the `quant_helpers.rs` conflict against main (merged main's 2-D `weight_scale` GPU routing with V4's `.scale`/E8M0 + 1-D fallthrough — superset, no family regressed), then cleared the draft CI debt — fmt, 28 `deny`-level clippy lints, 4 rustdoc broken links, the LoC cap (allow-listed the 12 V4 files + tracking split in #219), and the `test_buffer_arena_alloc` off-by-one (token_ids was uncounted → 26→27). All no-logic-change.

Coherence verified independently on 2×GB10: official `nvidia/DeepSeek-V4-Flash-NVFP4`, EP=2, fp8-KV, RoCE — "2+2"→4, clean Python. DS4F runs today only by building this branch (EnergyBurns/tonyd2wild/MiaAI-Lab on the forum); merging makes it available to everyone.

PR #204, opened 2026-06-27
fix(tool-parser): strip spurious trailing close markers leaking into streamed content

rsafier:
## Problem

Some models emit a **redundant tool-call close right after the real `</tool_call>`**. Ornith-1.0-35B, for example, emits `</_call>` (BPE-split into the four tokens `</` `_` `call` `>`) or a doubled `</tool_call>`. The real tool call parses correctly, but the spurious trailing close **leaked into the streamed `content`** field — observed delta: `"\n\n</_call>"`. Non-streaming output was unaffected.

## Root cause

The streaming `sanitize_content_chunk` loop drops standalone close markers via its `OrphanClose` arm, but a close that only finishes assembling at end-of-stream — or one preceded by emittable whitespace, so the tail's `looks_like_partial_tag` guard never fires — survives into `flush_content_sanitizer`, which previously dumped the held-back tail **verbatim**.

## Fix

In `flush_content_sanitizer` (`crates/spark-server/src/api/stream_guards.rs`), before flushing the held-back tail:
1. Strip every **complete** close marker from `markers.close` (matched against the reassembled buffer, so multi-token closes split across chunks are covered).
2. Trim a **trailing incomplete** close-marker prefix that the whitespace-blind `looks_like_partial_tag` check misses.

Only close markers are removed — legitimate content (including `<...>` prose, which never matches a `</...>` close-marker prefix) is untouched, and detector-based tool-call parsing is unchanged.

## Tests

Regression tests in `tool_parser/tests/streaming_frag.rs`:
- `</_call>` delivered as four single-token chunks
- doubled `</tool_call>`
- partial close reaching the flush boundary
- direct `flush_content_sanitizer` unit test (complete / incomplete / content-prefixed / no-close)

Each asserts the tool call still parses with correct name+args and that no markup leaks into content.

`cargo test -p spark-server tool_parser` → **90 passed**; `stream_guards` → **7 passed**. Stashing the fix reproduces the exact leak (`"\n\n</_call>"`).

## End-to-end verification

Validated against Ornith-1.0-35B streaming on GB10: streamed content is now empty with two correctly-parsed `history(...)` tool calls (previously leaked `</_call>`).

### DIFF
--- a/crates/spark-server/src/api/stream_guards.rs
+++ b/crates/spark-server/src/api/stream_guards.rs
@@ -147,7 +147,50 @@ pub fn flush_content_sanitizer(
         .map(|t| t.len())
         .max()
         .unwrap_or(0);
-    let final_text = std::mem::take(tag_scan_buf);
+    let mut final_text = std::mem::take(tag_scan_buf);
+    // Drop any COMPLETE standalone close markers still sitting in the
+    // held-back tail. The streaming `sanitize_content_chunk` loop drops a
+    // close via its `OrphanClose` arm, but a close that only finishes
+    // assembling at end-of-stream (or one preceded by whitespace, e.g.
+    // `\n\n</_call>`, so the tail's `looks_like_partial_tag` check below
+    // never fires) survives into this flush. Without stripping it here,
+    // the spurious redundant close some models emit right after the real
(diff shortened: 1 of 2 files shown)


rsafier:
Superseded by the combined tool-call markup-leak PR (both fixes in one).

PR #205, opened 2026-06-27
fix(tool-parser): stop tool-call markup leaking into content on runaway/guard-end

rsafier:
> **Stacked on #204** — shares `flush_content_sanitizer`. Until #204 merges, this PR's diff also shows #204's commit; review the second commit (`stop tool-call markup leaking into content on runaway/guard-end`) or merge #204 first.

## Problem

When a tool guard ends a response early — F11 within-response dedup, hard validation reject, or the loop cap — it sets `stop_string_triggered` as a generic "stop emitting" flag. But the scheduler keeps generating for a short tail (cancel latency), and on a tool-call **runaway** (the model re-opens 8+ `<tool_call>` blocks and emits spurious `</_call>` separators) those tail tokens are raw markup.

The `stop_string_triggered` arm in `handle_token` **returns before the detector/sanitizer fork** and emitted the delta verbatim, so the runaway markup — `<tool_call><function=…><parameter=…>…</tool_call></_call>` — leaked into the streamed `content` field. It only surfaced on `finish=length` runaways (high temperature + many tools + multi-turn), so a clean single-call turn never showed it.

## Root cause (traced from server logs)

```
Tool call: power({"range":"24h"})
F11 within-response dedup tripped: 2+ identical streaming tool calls; ending response  ← sets stop_string_triggered
… ~200 more tokens generated (cancel latency) → raw-emit path → LEAK …
tool-call repetition runaway … opens=8
Done: 265 tokens (length)
```

## Fixes (qwen3_coder, but parser-agnostic)

1. **`handle_token` (root cause):** route the post-trigger delta through the buffered `sanitize_content_chunk` instead of a raw passthrough, so multi-token markers that straddle deltas (`</_call>` = `</` `_` `call` `>`) are reassembled and stripped. A genuine stop-string's trailing content is unaffected — the sanitizer only removes tool-call markup.
2. **`scrub_tool_tags` (sanitizer.rs):** defense-in-depth final pass removing every complete tool-call markup tag from a content chunk — close tags, full-tag opens, and `<function=…>` / `<parameter=…>` attribute tags — derived from `LeakMarkers` so it holds for any parser.
3. **`flush_content_sanitizer`** now calls `scrub_tool_tags` (a superset of the close-only scrub from #204).

## Tests

3 new regression tests in `streaming_frag.rs` (direct `scrub_tool_tags` runaway-dump, non-tool-content preservation, end-to-end pipeline desync dump). **All 537 spark-server tests pass.**

## End-to-end validation

Ornith-1.0-35B on GB10. A runaway repro that leaked raw markup in **13/24** streams (`temp=0.6`, 20 tools, multi-turn) now leaks in **0/36**, including at `max_tokens=8192` / `temp=0.7`. Tool calls still parse correctly; the truncated-runaway content is now just `\n` separators.

### DIFF
--- a/crates/spark-server/src/api/chat_stream/handle_token.rs
+++ b/crates/spark-server/src/api/chat_stream/handle_token.rs
@@ -404,11 +404,33 @@ fn handle_token_inner(state: &mut StreamState, ctx: &StreamCtx, tok: u32) -> Sse
     }
 
     if state.stop_string_triggered {
+        // The tool guards (F11 within-response dedup, hard-validation
+        // reject, loop cap) reuse `stop_string_triggered` as a generic
+        // "end this response" flag. But the scheduler keeps generating for
+        // a few tokens after the flag is set (cancel latency), and on a
+        // tool-call *runaway* those tokens are raw markup —
+        // `<tool_call><function=…><parameter=…>…</tool_call></_call>` —
+        // re-emitted here. A raw passthrough (the old behaviour) leaked
+        // that markup into `content` because this branch returns BEFORE the
+        // detector/sanitizer fork below. Route the delta through the
+        // buffered `sanitize_content_chunk` so multi-token markers that
+        // straddle deltas (`</_call>` = `</` `_` `call` `>`) are reassembled
+        // and scrubbed. For a genuine stop string, legitimate trailing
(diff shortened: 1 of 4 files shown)


rsafier:
Superseded by the combined tool-call markup-leak PR (both fixes in one).

PR #206, opened 2026-06-27
fix(tool-parser): stop tool-call markup leaking into streamed content

rsafier:
Fixes two related streaming-parser leaks that let qwen3_coder tool-call markup reach the OpenAI `content` field. Both are parser-agnostic (derived from `LeakMarkers`) and leave tool-call parsing itself unchanged.

## 1. Spurious trailing close on clean turns

Some models emit a redundant close right after the real `</tool_call>` — Ornith-1.0-35B emits `</_call>` (BPE-split into `</` `_` `call` `>`) or a doubled `</tool_call>`. The real call parses, but the trailing close survived into `flush_content_sanitizer`, which dumped the held-back tail verbatim (its `looks_like_partial_tag` guard is whitespace-blind, so `\n\n</_call>` slipped past). Observed live as a `"\n\n</_call>"` content delta.

## 2. Raw markup dump on a runaway / guard-end

The tool guards (F11 within-response dedup, hard validation reject, loop cap) end a response by setting `stop_string_triggered`, but the scheduler keeps generating a short tail (cancel latency). On a tool runaway (model re-opens 8+ `<tool_call>` blocks with `</_call>` separators) those tail tokens are raw markup, and the `stop_string_triggered` arm in `handle_token` returned **before** the detector/sanitizer fork — emitting the delta verbatim. Only surfaced on `finish=length` runaways (high temp + many tools + multi-turn).

Traced from server logs:
```
Tool call: power({"range":"24h"})
F11 within-response dedup tripped … ending response   ← sets stop_string_triggered
… ~200 more tokens (cancel latency) → raw-emit path → LEAK …
tool-call repetition runaway … opens=8
Done: 265 tokens (length)
```

## Fixes

- **`handle_token`:** the `stop_string_triggered` arm routes the delta through the buffered `sanitize_content_chunk` instead of a raw passthrough, so multi-token markers that straddle deltas (`</_call>` = `</` `_` `call` `>`) are reassembled and stripped. Genuine stop-string content is unaffected.
- **`scrub_tool_tags` (sanitizer.rs):** defense-in-depth final pass removing every complete tool-call markup tag from a content chunk — close tags, full-tag opens, and `<function=…>` / `<parameter=…>` attribute tags.
- **`flush_content_sanitizer`:** drops complete markers via `scrub_tool_tags` and trims a trailing incomplete close-marker prefix at end-of-stream.

## Tests

7 regression tests in `streaming_frag.rs` (single-token-chunked `</_call>`, doubled `</tool_call>`, flush-boundary partial close, direct `flush_content_sanitizer` + `scrub_tool_tags` units, non-tool content preservation, end-to-end desync-dump pipeline). **All 537 spark-server tests pass.**

## End-to-end validation

Ornith-1.0-35B on GB10: a runaway repro that leaked raw markup in **13/24** streams (temp=0.6, 20 tools, multi-turn) leaks in **0/36** after the fix, including at max_tokens=8192 / temp=0.7. Tool calls still parse correctly; truncated-runaway content is now just `\n` separators.


tbraun96:
@rsafier good hunting on both leak paths — the stop_string_triggered raw-emit and the whitespace-blind flush tail are real bugs, the root-cause writeup is excellent, and the 13/24 -> 0/36 live repro is exactly the evidence we want. three things before this can come out of draft:

1. the unconditional scrub_tool_tags at the end of sanitize_content_chunk breaks the F73 minimax envelope pass-through. minimax puts <invoke name=" / </invoke> / </parameter> / </tool_call> in orphan_open/close, but inside a <minimax:tool_call> envelope those are legitimate content the downstream parser extracts the tool call from. with your scrub, '<minimax:tool_call> <invoke name="bash"> <parameter name="command">uname -r</parameter> </invoke> </minimax:tool_call>' sanitizes to just '<minimax:tool_call> uname -r' — invoke/parameter tags and even the envelope close get stripped. there's an existing regression test for exactly this (api/tests/sanitizer.rs, sanitizer_envelope_open_disables_orphan_suppression) that fails on your branch — CI didn't catch it because that mod is commented out in api/tests/mod.rs. simplest fix: skip the final scrub when markers.envelope_open is non-empty (qwen3_coder has no envelope markers, so both your fixes keep working), and gate the scrub_tool_tags call in flush_content_sanitizer the same way. also note the in-sanitize scrub runs per-chunk on post-state-machine output, so a tag split across chunks slips through in mangled halves anyway — the buffered state machine is the real guard for the streaming path; the scrub only needs to cover the flush and stop-arm dumps. honestly you could drop it from sanitize_content_chunk entirely.

2. CI: cargo fmt --check fails (4 spots, all in the new streaming_frag tests) and the 500-LoC cap fails — sanitizer.rs is now 516 and streaming_frag.rs 727. move scrub_tool_tags into its own file and split the new tests into a separate test module.

3. branch is behind main — can you rebase?

once those land i'll run the runaway repro against ornith on the test box and merge. the core fixes are right.

rsafier:
Addressed all three — ready for re-review:

1. **MiniMax F73 regression fixed.** `scrub_tool_tags` is now gated on `markers.envelope_open.is_empty()` in **both** `sanitize_content_chunk` and `flush_content_sanitizer`, so envelope parsers (MiniMax) pass `<invoke>`/`<parameter>`/envelope tags through untouched while qwen3_coder (no envelope) keeps both leak fixes. Relocated the `sanitizer_envelope_open_disables_orphan_suppression` regression test out of the disabled `api/tests/sanitizer` mod into the live `tool_parser::tests::streaming_leak` module — it now runs in CI and **passes** (it fails on the pre-gate scrub).
2. **CI green.** `cargo fmt --check` clean; ≤500-LoC cap satisfied — moved `scrub_tool_tags` → `api/scrub.rs` (sanitizer.rs 516→480) and split the new leak tests → `tool_parser/tests/streaming_leak.rs` (streaming_frag.rs 727→432, streaming_leak.rs 377).
3. **Rebased** onto current `main` (incl. the merged #224) — clean, no conflicts.

Verified: `cargo test -p spark-server` → 519 passed / 0 failed, all 8 leak tests + the minimax regression green. Confirmed distinct from #222/#224 (that's the tool-*name* truncation; this is markup-into-content).

tbraun96:
@rsafier reference implementation of the review asks is on branch fix/streaming-markup-leak in this repo (rebased on main composing with your #224, scrub removed from sanitize_content_chunk + envelope-gated at flush/stop-arm, the disabled sanitizer test module re-enabled and passing, files split under the cap — 555 tests green). adopt it with: git fetch origin fix/streaming-markup-leak && git reset --hard FETCH_HEAD on your branch, or cherry-pick what you agree with — your call, especially on remove-vs-gate for the in-chunk scrub. happy to gate on a spark as soon as the branch shape is yours.

rsafier:
Adopted your reference impl (`fix/streaming-markup-leak` → this branch, force-updated). I agree on **remove-over-gate** for the in-chunk scrub: the state machine already consumes every complete marker on the streaming path, and you keep the scrub envelope-gated at the flush + stop-arm dumps, so qwen3_coder's runaway/`</_call>` coverage is preserved while the per-chunk pass (which only ever saw mangled post-state-machine halves) is gone. Re-enabling + cleaning the full `api/tests/sanitizer.rs` module is also better than my relocate-one-test.

Verified locally on your head: `cargo fmt --check` clean, all files ≤500 (sanitizer 467 / scrub 61 / streaming_frag 432 / streaming_leak 322 / tests/sanitizer 313), `cargo test -p spark-server` = 527 passed / 0 failed incl. the minimax envelope regression + all leak tests. Branch shape is yours now — good to gate on a spark whenever.

SeedSource:

### Triage
- Rationale: Security sanitization fix preventing tool-call markup leakage into response content.
- Serving-path files: `crates/spark-server/src/api/chat_stream/handle_token.rs, crates/spark-server/src/api/mod.rs, crates/spark-server/src/api/sanitizer.rs, crates/spark-server/src/api/scrub.rs, crates/spark-server/src/api/stream_guards.rs, crates/spark-server/src/tool_parser/tests.rs`

### Serve gate
- Top-k KL vs main: **0.0002** (no shift vs main) - decode delta **-0.0%**

### GLM-5.2 code review
**Correctness:** The core fix (routing `stop_string_triggered` deltas through `sanitize_content_chunk`) is sound. The `scrub_tool_tags` byte-indexing is UTF-8 safe via `len_utf8()`. The F73 envelope gate (`envelope_open.is_empty()`) correctly prevents scrubbing legitimate minimax payloads.

**Risks:**
1. `scrub_tool_tags` early-returns `out` on partial open tag with no `>` — silently drops trailing content. For desync dumps this is intentional, but if a legitimate `<` appears in prose mid-flush with no closing `>`, content after it is lost. Verify `looks_like_partial_tag` downstream doesn't re-add it.
2. The partial-close trim (`rfind('<')`) could truncate a legitimate `<` in prose if it happens to be a prefix of a close marker (e.g., `</` prefixing `</_call>`). Low probability but worth a test.
3. Test file re-enables `sanitizer` module but the diff is truncated — can't verify new test coverage for the scrub/flush paths.

**Quality:** Comment density is excessive (the PR is ~60% comments). The `scrub.rs` split is reasonable for the 500-LoC cap. Missing: unit tests for `scrub_tool_tags` edge cases (partial tags, UTF-8 boundaries, empty markers).

**Verdict:** Approve with concerns — add targeted tests for the partial-tag drop and prefix-collision paths.

rsafier:
Rebased onto post-#229 `main`. The PR's two commits (tool-call markup-leak fix + minimax-envelope pass-through / sanitizer-split) replay cleanly — the streaming/sanitizer surface (`api/sanitizer.rs`, `stream_guards.rs`, `chat_stream/handle_token.rs`, new `api/scrub.rs`, `tool_parser/tests/streaming_leak.rs`) is orthogonal to #229's kernel/model work, so no conflicts. The diff is now scoped to exactly those 10 streaming files (488/−134) with none of the old base-divergence noise.

Confirmed still-needed vs current main: `api/scrub.rs` and `tool_parser/tests/streaming_leak.rs` are absent from main, and main's `sanitizer.rs`/`stream_guards.rs`/`handle_token.rs` still carry the pre-fix behavior — this fix was **not** absorbed by #229.

Validation on GB10 (dgx-00):
- `cargo fmt -p spark-server --check` clean.
- `cargo clippy -p spark-server --all-targets -- -D warnings` clean.
- `cargo test -p spark-server --bin spark` → **581 passed, 0 failed, 1 ignored** (streaming/sanitizer/leak group: 34 passed, 1 ignored).

Note: the earlier automated-review pass on this PR did not complete (pipeline retry cap), so this one hasn't had a full adversarial review yet — worth a fresh look before merge. Commits signed.

rsafier:
## Review — #206 (rebased onto post-#229 main)

**Verdict: correct, well-scoped, thoroughly tested. Approve with a few minor, non-blocking notes.**

Reviewed the full pipeline: `handle_token` stop-branch → `sanitize_content_chunk` state machine → `handle_done` flush → `scrub_tool_tags`, plus all 8 new streaming-leak tests and the re-enabled `tests/sanitizer.rs`.

### What's right
- **Root cause is correctly identified & fixed.** The `stop_string_triggered` branch in `handle_token` returned *before* the detector/sanitizer fork, so post-stop tokens (tool-call runaway markup) were emitted raw into `content`. Routing that delta through the same buffered `sanitize_content_chunk` reassembles multi-token markers that straddle deltas (`</_call>` = `</` `_` `call` `>`) and scrubs them.
- **No truncation of legitimate content.** The stop-branch holds back ≤`tag_max-1` bytes; I verified `handle_done.rs:168`'s `flush_content_sanitizer` runs **unconditionally** (not gated on `stop_string_triggered`), so the held-back tail is always drained and emitted at `[DONE]`.
- **F73 envelope pass-through preserved end-to-end.** `scrub_tool_tags` is gated on `markers.envelope_open.is_empty()`, and `inside_envelope` suppresses orphan matching — so minimax `<invoke>…</invoke>` payloads survive both the streaming path and the stop-branch. The `dcf63a5` "don't scrub envelope" commit closes that correctly.
- **The step-2 trailing-partial-close trim is precise.** It only trims a trailing `<…` that is a genuine *prefix of a close marker* (`</`, `</t`, `</_`), so real prose (`a < b`, `</div`, `<div>`) is preserved — confirmed by `scrub_tool_tags_preserves_non_tool_content`.
- **Correctly scoped:** the `sanitize_content_chunk` fast-path opts out for non-tool parsers (Hermes/Gemma4/Mistral/BareJson) → zero latency or behavior change for them.
- **Tests are excellent:** BPE-split close, doubled close, whitespace-preceded (`\n\n</_call>`), incomplete-at-EOS (`>` fused with EOS), full runaway dump, plus negative tests. And it **re-enables `tests/sanitizer.rs`**, which was entirely dead (`// mod sanitizer;`, stale F-code helpers) — net go from 0 → 9 active sanitizer regression tests.

### Minor, non-blocking
1. **Asymmetry — a trailing partial *open* preceded by content can still leak.** Step 2 handles partial closes preceded by content, and `looks_like_partial_tag` handles a whole-tail partial open, but a partial *full-tag open* with content before it (flush tail like `lo <tool_ca`) is caught by none: not complete (scrub skips), doesn't start with `<` (`looks_like_partial_tag` misses), not a close prefix (step 2 misses) → `<tool_ca` leaks. Low severity (opens-in-progress are normally detector-routed, not content) and not a regression (pre-existing gap). A symmetric trailing-partial-*open* trim mirroring step 2 against `orphan_open` full-tag markers would close it.
2. **Slight over-trim** of legit content ending *exactly* in a close-marker prefix (`</`, `</t`) at EOS. Very low severity (precise prefix match; only the final ≤`tag_max-1` tail; tool-parsers only); it's a deliberate, documented tradeoff.
3. **Token-id accounting:** in the stop-branch, when `cleaned` is empty (all markup), `take_ids_if` is skipped, so suppressed-markup token ids coalesce into a later chunk / the flush tail. Observability-only (`req_return_token_ids`); cosmetic.

None block merge. `fmt` / `clippy --all-targets -D warnings` clean; `cargo test -p spark-server --bin spark` = 581 passed / 0 failed / 1 ignored.

rsafier:
@tbraun96 Branch synced with main (merge, no force-push). The sync surfaced that #388 absorbed most of the sanitizer machinery (`sanitize_content_chunk`, `stream_guards`, and a whole-stream harness test suite), so the post-sync residual here is small and honest: (1) the **stop-string early-emit arm** in `handle_token.rs` — main pushes that delta unsanitized, which is exactly the straddle-leak class this PR exists for, so the scrub now applies there before the `StreamDelta::Content` push; (2) the unit-level sanitizer tests (10 fns driving `sanitize_content_chunk`/`flush_content_sanitizer` directly) kept alongside main's harness suite, plus the 322-line `streaming_leak.rs` integration set. fmt/clippy `-D warnings`/tests green locally (1764 bin tests, 0 fail). Ready for the spark gate whenever.

tbraun96:
Landed in main via the #440 stack merge (`3c941751`). Verified by content diff — the production side is byte-identical to this PR's head across `api/scrub.rs`, `api/sanitizer.rs`, `api/stream_guards.rs`, `chat_stream/handle_token.rs` and `tool_parser/tests/streaming_leak.rs`.

One cosmetic note so the diff doesn't look short: your `api/tests/sanitizer_unit.rs` landed as `api/tests/sanitizer_chunk_tests.rs` during integration. All eight test fns are present verbatim (`sanitizer_envelope_open_disables_orphan_suppression` through `flush_before_tool_boundary_recovers_from_stuck_suppression`) — the only change is the dropped `mod sanitizer_tests {}` wrapper. No coverage was lost.

Worth flagging: the orphan-suppression path this PR hardened turned out to have one more hole downstream — the streak watchdog cut the stream without naming a guard, so it wired `finish_reason: "stop"` instead of `"length"`. Fixed separately in #438. Thanks @rsafier.

PR #207, opened 2026-06-28
SBR M1: opt-in tail-pin SSM-snapshot eviction (8× warm-resume on deep agentic convs)

tbraun96:
## Problem

On a warm multi-turn hit, KV is reused but the GDN/Mamba SSM layers must replay
their recurrent state from the nearest deeper Marconi checkpoint to the match
point. As the snapshot pool fills, the forecast eviction (`last_access·(1+hit_count)`)
drops a live conversation's **deep** per-turn checkpoints first — their
`hit_count≈0` (they were never the resume *anchor*), so they look like one-shot
traffic — leaving the next resume to replay from a far-shallow survivor. Replay
distance grows with depth → warm-hit TTFT climbs (measured ~9.5s at ~24k depth;
the 1s→21s pathology).

## Fix (M1) — opt-in, OFF by default

`SsmSnapshotIndex::evict_lru` gains a tail-pin mode: pin the **top-K (=8) deepest
snapshots of each *resumable* session** (a conversation resumed at least once),
evicting only non-pinned entries. A resuming deep conversation then finds a
near-tail SSM anchor instead of a far one.

Measured on dgx2 (Qwen3-Next-80B-NVFP4, deep conv idle under one-shot pressure,
then resumed):

| | baseline (forecast) | tail-pin ON |
|---|---|---|
| warm-resume TTFT (mean) | 9.53 s | **1.18 s (8×)** |
| best cycle | 9.74 s | **0.45 s (21×)** |
| replay distance | ~7,600 tok | **11–984 tok** |

The 0.45 s warm-cycle latency matches llama.cpp's continuous-sequence resume —
without keeping the sequence live. **Exact by construction**: only *which*
checkpoint is restored changes; replay is the unchanged bit-exact WY4 path.

## Honest scope — why OFF by default

Enabling tail-pin **regresses balanced many-conversation round-robin ~30%**
(5.9s→7.7s): there the recency·hit forecast is already near-optimal and pinning
fights it, and the regime cannot be reliably detected from the snapshot index's
local view (four gating variants tried, all failed). So it is **OFF by default**
— `ATLAS_SBR_TAIL_PIN` unset is exactly the baseline forecast (do-no-harm
everywhere). Set `ATLAS_SBR_TAIL_PIN=1` (optionally `ATLAS_SBR_TAIL_PIN_K`) for
deep single/few-conversation agentic workloads, where it delivers the 8×.

## Research artifacts (`research/sbr/`)

Full forensics, benchmark harnesses, and findings, including the negative
results that bound the approach:
- Dense operator-transport is FLOP-negative for GDN; real GDN decay is
  heavy-tailed (median horizon ~163 tok, but ~12% of heads > 8192).
- **M3 2-D (position × layer) sheaf reconciliation** prototyped → honest
  negative: non-vacuous with an oracle inter-layer map but circular/net-negative
  with a fittable one (an accurate inter-layer restriction map *is* the layer's
  forward pass → no shortcut). M1 is the real win, not the sheaf framing.

## Tests
`cargo test -p spark-runtime radix_tree::tests::snapshot` — 20 pass, incl.
`test_tail_pin_protects_top_k_deepest` and `test_tail_pin_off_is_pure_forecast`.

### DIFF
--- a/crates/spark-runtime/src/radix_tree/snapshot.rs
+++ b/crates/spark-runtime/src/radix_tree/snapshot.rs
@@ -10,25 +10,44 @@ use std::sync::OnceLock;
 
 use super::hash_token_prefix;
 
-/// SBR M1 two-tier eviction policy. When enabled, [`SsmSnapshotIndex::evict_lru`]
-/// evicts NON-resumable (one-shot / untracked) snapshots before any resumable
-/// multi-turn conversation's snapshot — protecting live conversations' SSM
-/// checkpoints from transient one-shot churn (the cause of the "replay grows as
-/// slots fill" TTFT blowup, 1s→21s). When every entry is resumable it degrades
-/// identically to the baseline recency·hit forecast (Marconi §4) — provably
-/// do-no-harm.
+/// SBR M1 tail-pin eviction policy — OPT-IN, OFF by default. When enabled,
+/// [`SsmSnapshotIndex::evict_lru`] pins the top-`tail_pin_k` DEEPEST snapshots of
+/// each RESUMABLE session (one resumed at least once), so a resuming deep
+/// conversation finds a near-tail SSM anchor instead of replaying from a far
+/// shallow survivor (the "replay grows as slots fill" blowup, 1s→21s). Measured
+/// 8× warm-resume speedup on the deep-conversation agentic regime.
 ///
-/// Enabled by default; set `ATLAS_SBR_TAIL_PIN=0` to restore the pure
-/// forecast-LRU baseline (for A/B benchmarking against the fix).
+/// DEFAULT OFF: it wins the single/few-deep-conversation regime but REGRESSES
+/// balanced many-conversation serving ~30% (the recency·hit forecast is already
+/// near-optimal there). Set `ATLAS_SBR_TAIL_PIN=1` for deep-agentic workloads.
 fn tail_pin_enabled() -> bool {
     static ENABLED: OnceLock<bool> = OnceLock::new();
     *ENABLED.get_or_init(|| {
-        let on = std::env::var("ATLAS_SBR_TAIL_PIN").map(|v| v != "0").unwrap_or(true);
+        let on = std::env::var("ATLAS_SBR_TAIL_PIN")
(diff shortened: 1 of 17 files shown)


SeedSource:

### Triage
- Risk: **med** - serve-test-required (triage advisory): `True` - path-heuristic (authority): `False`
- Rationale: SBR eviction policy changes with opt-in tail-pin logic; requires serve testing for performance impact.

### Serve gate
- Build: 29s - Qwen3.6-35B-A3B-FP8 on GB10 node1
- Top-k KL vs main: **0.1131** - **large behavioral shift** - decode delta **-0.0%**

### GLM-5.2 code review
**Correctness:** Logic is sound. `evict_lru_inner` correctly computes a global fallback (`all_idx`) alongside the non-pinned minimum, preventing panic when all entries are pinned. `swap_remove` on `idx` is safe since `victim_idx`/`all_idx` are valid indices. `session_hash == 0` skip is a reasonable guard. Tests cover both pin-on and pin-off paths.

**Risk:** Low—default-off via env var is do-no-harm. However, `evict_lru` is on a hot path (called at slot saturation); the pin-enabled branch allocates a `HashMap` + `HashSet` per eviction call. Under the opt-in deep-agentic regime this is acceptable, but if `evict_lru` is called in a tight drain loop, repeated allocations could add latency spikes. Consider reusing allocations or documenting the cost. Also, `tail_pin_enabled()`/`tail_pin_k()` read env on first call only (via `OnceLock`), so runtime toggling is impossible—fine for opt-in, but worth noting in docs.

**Quality:** Comment density is excessive (the `evict_lru_inner` doc comment is ~20 lines). The research MD file is informative but shouldn't ship in `research/` long-term—move to docs or remove before non-M1 release. Test assertions are precise. The `pin_k < 1` filter silently falls back to 8; a logged warning on invalid override would aid debugging.

tbraun96:

```
-244,504 lines deleted
```

- It is **CONFLICTING**, and GitHub does not run merge-commit workflows for a PR it cannot merge — so every green tick displayed here describes a tree from 2026-07-04 that nobody will merge. Thirteen open PRs are in that state; #427 adds a `needs-rebase` label so the false green stops being invisible.

PR #209, opened 2026-06-28
fix(dflash): correctness fixes for K=γ verify path on Qwen3.6-27B NVFP4

Sujimoshi:
## Summary

Three correctness bugs that together caused 0% accept rate on DFlash with Qwen3.6-27B NVFP4. Found and fixed while benchmarking on a DGX Spark (GB10/SM121). All fixes tested end-to-end — DFlash loads, drafter installs, serves correctly at 12.5 tok/s.

### Fixes

**1. BF16/FP32 type mismatch in batched verify path** (\`fix(ssm)\`)

The sequential fallback in \`decode_batched_conv_gdn\` (K ∉ {2,3,4,17}) always used the BF16 conv kernel, then passed its output to \`gated_delta_rule_decode\` which takes \`const float*\`. Every pair of BF16 bytes was reinterpreted as one FP32, corrupting \`h_state\` → NaN → EOS-token collapse → 0% acceptance for K=5..16.

Fix: check \`conv1d_l2norm_f32_k\` availability and use the FP32 conv kernel + \`ssm_conv_out_f32()\` buffer when present. Also adds \`norm_token_count\` rollback on partial/full reject in \`async_chkpt\`.

**2. YaRN RoPE applied to drafter with \`rope_scaling=null\`** (\`fix(dflash)\`)

Three bugs: (a) YaRN RoPE (factor=64, beta_fast=32) applied to drafter whose \`config.json\` has \`rope_scaling=null\` — compressed all freq components by ≥64×, garbage attention in every drafter layer. (b) Position IDs from absolute seq position instead of 0-indexed draft position. (c) \`noise0\` not zeroed before forward pass. Fixed: standard RoPE, corrected position IDs, zero noise0.

**3. BF16 embed stride bug in \`verify_d.rs\`** (\`fix(dflash)\`)

\`let fp32 = 4usize\` used as embed stride for BF16 hidden states — token[t] landed at 2× correct offset, odd positions were uninitialised stale data. \`verify_c.rs\` correctly uses \`fp32 = 2usize\`. Fixes garbage output for K≥5. Also fixes OOB write in WY17 path (conv loop ran 0..K instead of 0..K-1).

### Dependencies included

**GDN norm cadence skip** (\`perf(gdn)\`): skip SSM state norm every 16 tokens. Required for the \`norm_token_count\` rollback in async_chkpt. +1.1% throughput on GB10.

**Native U8 NVFP4 SSM loader** (\`feat(loader)\`): load \`in_proj_qkv/z/out_proj\` directly as \`QuantizedWeight\` when dtype=UInt8, instead of dequant→BF16→requant roundtrip. Required — Qwen3.6-27B-NVFP4 stores SSM projections as U8; without this the server fails on load.

## Test results (DGX Spark, GB10/SM121)

- ✅ Server starts, loads model + DFlash drafter without errors
- ✅ DFlash cap=1 (K=2): coherent output, 12.5 tok/s
- ✅ No UInt8 loader errors
- ⬜ DFlash cap=3/15: acceptance rate testing pending
- ⬜ MTP/pure-decode regression check pending

## Stacking note

This branch is rebuilt on top of #257 (native U8 NVFP4 loader) and #258 (DFlash model glue) — both are prerequisites and both of their commits are included as ancestors here (base stays `main` since GitHub requires the base branch to exist in this repo, and I don't have push access to create `feat/dflash-model-glue`/`feat/nvfp4-native-u8-loading` here). **Merge order: #257, then #258, then this PR.** Once #257/#258 land in `main`, this PR's diff will automatically shrink to just its own 6 commits (no rebase needed on my end).


Sujimoshi:

SeedSource (changes requested):
**Review (SeedSource / codeowner): REQUEST-CHANGES**

The three headline fixes are real and well-cited — the BF16→FP32 sequential conv-kernel fix (the root-cause NaN), the YaRN-on-`rope_scaling: null` → standard-RoPE correction, and the noise0 conditioning / position-id fixes all look correct and match the SGLang reference. But there's a likely **regression in the K=17 (γ+1) verify path this PR targets**, and that path is untested per the checklist.

**HIGH — WY17 loop reduction starves the fused kernel of the 17th token's conv output.** `trait_decode_batched_conv_gdn.rs:~313`: changing `for t in 0..num_tokens` → `0..num_tokens-1` drops the `t=16` iteration, which does two things per token: (a) `conv1d_update_l2norm` fills `conv_out_buf[t]` and advances `ssm_state.conv_state`; (b) snapshots into `conv_state_intermediates[t]`. The `gdn_decode_wy17` contract (`ops/ssm_gdn_b.rs:230`) reads `conv_out_buf` for all 17 token columns, so `conv_out_buf[16]` must be valid — it's now stale. And `conv_state` after the loop omits token 16's causal-conv update, so even a full 17-accept leaves state wrong for the next step. The real OOB being fixed is only the snapshot write to `conv_state_intermediates[16]` (pool is K-1=16 slots). **Fix:** keep `conv1d_update_l2norm` running for all 17 tokens; gate only the `conv_state_intermediates` snapshot with `if t < num_tokens - 1`.

**MEDIUM — global state-norm cadence change bundled in, regression check deferred.** `gated_delta_rule.cu:162,292` + `ssm_forward.rs:~317`: the explosion-clamp went every-token → every-16-tokens for ALL models using `gated_delta_rule_decode` (not just the γ path), while the fused verify kernels still clamp every token → cadence now differs between accepted vs verify paths. The checklist marks the MTP/pure-decode regression "pending" → shipped unverified + scope-creep vs the title. Split it out, or run the pure-decode A/B before merge.

**MEDIUM — `norm_token_count` rollback is asymmetric** (`async_chkpt.rs:328,397-402`): the counter is advanced only in the single-token + sequential-batched paths, but `async_chkpt` unconditionally `wrapping_sub`s it for every SSM layer on reject/partial — rewinding a counter the fused K=2/K=17 paths never advanced. `u32` mod-16 makes it non-fatal, but it's logically inconsistent; only roll back when the sequential path actually ran.

**MEDIUM — NVFP4 concat precondition unenforced** (`weight_map/quantized.rs:~110 concat_rows`): keeps only `self.weight_scale_2` with a "both MUST share weight_scale_2" comment that's never asserted; mismatched global scales would silently mis-scale the second block. Add an `ensure!` (or rescale).

**LOW:** PR body says `verify_d.rs:53 fp32 4→2` but the diff is comment-only (overstated); leftover `tracing::warn!("sequential K=…")` in the hot decode loop (demote to debug/trace); hardcoded `rope_theta = 10_000_000.0` (source from config); `weight_loader/qwen35_dense.rs` ≈514 lines (>500 cap).

**Confirmed correct (no action):** the BF16→FP32 conv fix, the position-id `ctx_total - eff_ctx` + noise0 / argmax-skip logic (matches SGLang, preserves draft count), `initial_tokens` is genuinely consumed (not dead), `do_norm` threaded through all `gdn_decode` sites, no Python added.

**Summary:** the diagnosed 0%-accept root causes are fixed well. Not merge-safe as-is — fix the WY17 loop (most likely corrupts the 17th verify token), assert the NVFP4 concat scale, and run the deferred K=4/K=16 acceptance + pure-decode regression checks. Happy to re-review once those land.

tbraun96:
@Sujimoshi thanks for this — the three root-cause fixes are genuinely good work. the BF16-reinterpreted-as-FP32 conv diagnosis and the YaRN-on-rope_scaling=null catch are exactly the kind of bugs that hide for months. seconding SeedSource's review, plus a few things from a fresh pass:

1. the wy17 loop trim is a real regression: gated_delta_rule_wy17.cu loads q/k/v for all 17 tokens (K_TOKENS loop in the SMEM load), so running conv for only 0..16 leaves conv_out_buf[16] as stale scratch and conv_state one token behind after a full accept. keep conv1d_update_l2norm running for all 17 and gate only the conv_state_intermediates[t] snapshot with t < 16. heads up: the same pattern is in your #218 branch at both the wy17 and new wy16 dispatch sites. also note with the argmax skip, forward_block now emits at most gamma-1=15 drafts, so dflash tops out at K=16 — wy17 only fires via --speculative --num-drafts 16 today, but it still has to be correct before it lands.
2. concat_rows: assert both weight_scale_2 match (or rescale the second block). modelopt scales are per-tensor, so a mismatched checkpoint silently mis-scales the z projection on exactly the U8 loads this path serves.
3. quantize_to_nvfp4_per_row + weight_scale_2_vec have no consumer here — w4a16_gemv_prs doesn't exist in this branch. drop the dead code and the 12-file field churn, or hold it for the PR that adds the kernel.
4. the every-16 norm clamp hits plain decode for every model using the common gdn kernel, not just dflash. that needs the pure-decode A/B you marked pending (greedy logit compare vs main is enough), or split it into its own PR.
5. small stuff: verify_d.rs fp32=2 is already on main so that item is comment-only; demote the sequential-K warn! and the per-propose info! to debug; the gpu_impl.rs import reshuffle looks accidental; qwen35_dense.rs is at 552 lines (500 cap) and fmt/clippy/rustdoc are red.

bigger question: everything here is contained in your #218, which also has the CI fixes. can you pick one vehicle so we don't review the same diff twice? my preference: keep this one as the minimal correctness set (the three fixes + U8 loader with the concat assert), move the norm cadence and per-row-scale2 out, fix the wy17 loop, rebase on main and green up CI — then rebase 218 on top. if you'd rather land it all through 218, say so and we'll close this and review there. happy to re-review fast either way.

SeedSource:

### Triage
- Rationale: Fixes position ID calculation and token handling in dflash_head forward block, affecting RoPE and attention.
- Serving-path files: `crates/spark-model/src/layer.rs, crates/spark-model/src/layer/tests.rs, crates/spark-model/src/layers/dflash_head/forward_block.rs, crates/spark-model/src/layers/dflash_head/from_weights.rs, crates/spark-model/src/layers/dflash_head/propose.rs, crates/spark-model/src/layers/nemotron_mamba2/trait_impl.rs`

### Serve gate
- Build: 26s - Qwen3.6-35B-A3B-FP8 on GB10 node1
- Top-k KL vs main: **0.2392** - **large behavioral shift** - decode delta **+0.3%**

### GLM-5.2 code review
**Correctness concerns:**

1. **`norm_token_count` is dead code.** Added to `SsmLayerState` and test constructors but never read, written, nor reset. The doc comment claims it gates SSM state norm skipping — no such logic exists. This is misleading cruft; either implement the gating or drop the field.

2. **`noise0_pos` truncation.** `(position as i64 - 1).max(0) as i32` silently clamps; if `position` exceeds `i32::MAX` this wraps. Unlikely for real seqs but the `i64`→`i32` cast is unjustified — position is already `u32`/`usize`; use checked arithmetic and propagate or clamp explicitly.

3. **`initial_tokens` parameter plumbed but never exercised.** `forward_block` now accepts `Option<&[u32]>` for denoising T≥2, but the diff is truncated mid-`propose.rs` — the T=2 path that would pass `Some(tokens)` is incomplete. The `drafts` variable is conditionally built but the branch is cut off. **Cannot verify the denoising loop is correct or even compiles.**

4. **YaARN removal is a real fix** — standard RoPE matches `rope_scaling: null`. Good.

**Risk:** The truncated diff hides the most complex logic (T≥2 denoising). Request the full `propose.rs` changes before merge.

**Quality:** Comments are excessive (reference Python line numbers, training rationale). Trim. Field `yarn_inv_freq` variable name is now stale — rename to `rope_inv_freq`.

Sujimoshi:
Closing in favor of the ongoing #218 split effort (#255, #257 merged, #258). The BF16/FP32 conv fix landed on main independently via #229; the remaining correctness fixes (RoPE/position-ids/noise0/embed-stride) will re-land as part of the stacked split PRs.

PR #215, opened 2026-06-29
feat(qwen35-dense): native dense FP8 attn/FFN dispatch + kernels (Holo-3.1/Ornith)

rsafier:
> **Draft, stacked PR to upstream.** Base `Avarok-Cybersecurity/atlas:main`, but built on **#203** (Holo-3.1/Ornith enablement) + **#210** (native FP8 loading), so the diff includes those. **The delta is the top 6 commits** on `feat/dense-fp8-attn-ffn` (on top of #210's `ed4df52`): the native *dense* FP8 attn/FFN path + kernels. Rebases to a clean delta once #203/#210 land.

## What this adds (on top of #210's FP8 loading)
#210 loads compressed-tensors float-quantized FP8 weights; this adds the **native dense FP8 compute path** so Holo-3.1 (and Ornith dense) actually run attention + FFN in FP8:

- **`feat(qwen35-dense)`: native FP8 dispatch for dense attention + FFN** — routes dense q/k/v/o + gate/up/down through FP8 GEMV/GEMM.
- **`feat(kernels)`: dedicated dense-FP8 kernels** — fused `w8a16` dual-GEMV (decode) + fast transposed `w8a16` prefill GEMM.
- **`feat`: gate behind `ATLAS_DENSE_FP8`** (opt-in) — NVFP4 stays the default; FP8 is for accuracy-sensitive runs (speed is knowingly traded).
- **`fix(vision)`: size ViT SDPA score buffers to `p_max`** — fixes a CUDA-700 on images >1024 patches (release builds compiled out the debug-assert guard, so it passed by silent OOB luck).
- **`fix(qwen35-dense)`: keep the NVFP4 fallback under the FP8 overlay** — fixes a CUDA-700 at concurrency (nulled NVFP4 weights made batched/multi-seq paths hit null; keep real NVFP4 + FP8 overlay).
- `refactor`: drop debug logs; document the dense-FP8 perf finding.

## Status
FP8 path validated end-to-end this session on Holo-35B (single-stream correct ~29 tok/s, concurrency 0-error after the NVFP4-fallback fix). fmt + file-size clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #214, opened 2026-06-29
feat(decode): batched-decode scaling (conv-fix + attention tiling) + --target-kv-tokens + prefill logging

rsafier:
> **Draft, stacked PR to upstream.** Base is `Avarok-Cybersecurity/atlas:main`, but this branch is built on the pending **Holo-3.1/Ornith enablement** + **dense-FP8** work (not yet on avarok/main), so the diff below includes that base. **The relevant delta is the top 19 commits** (`771d414..HEAD`) — everything from `fix(scheduler): sort decode batch by SSM slot` onward. Once the Holo/FP8 PRs land upstream this rebases to an engine-only diff. Benchmark/soak infra is a separate PR (`feat/holo-official-harness`).

> **Draft / stacked PR.** Base = `feat/dense-fp8-attn-ffn` (the dense-FP8 PR), which itself stacks on the Holo-3.1/Ornith enablement. Depends on those landing first. Benchmark/soak infra (realistic_soak + test images) goes to the separate `feat/holo-official-harness` PR, not here.

## What this does
Makes the multi-seq **decode** path batch correctly and scale, plus two server-side QoL features.

### Decode — the core
- **`fix(ssm)`: conv input-stride bug** in the batched-recurrent SSM. The batched conv strided its input by `conv_dim` while the QKVZ projection is laid out with stride `qkvz_size`, so seq `b≥1` read the previous seq's Z-gate region → garbage into the GDN scan. Correct at n=1, **corrupt at n≥4 (0/3 probes)**. Fixed by running the cheap conv per-seq with pre-offset pointers (mirrors `trait_decode_batched_conv_gdn`). Now **4/4 correct**. This is the key enabler.
- **`fix(scheduler)`: slot-sort** the decode batch by SSM slot so the contiguous-slot invariant holds and the batched-recurrent + graph paths actually engage (they were silently falling back).
- **`perf(decode)`: batch dense FFN + tile qkv/o_proj GEMVs** at n≥4 — read each projection weight once per tile instead of once per seq.

**Result (Holo-3.1-35B-A3B-NVFP4, GB10):** decode scaling **2.09× @ C=12/16** (94→196 tok/s), 0 errors — clears vLLM-like (≥2×). The conv-fix-enabled SSM-batched path is +10% @C8 on the MoE. Decode is SSM-scan-bound (~72–76%); full decomposition + every lever is in `BATCHED_DECODE_GRAPH_DESIGN.md`.

### Server features
[...]

PR #210, opened 2026-06-28
feat(qwen35): native FP8 for Holo-3.1 (compressed-tensors float-quantized)

rsafier:
**Stacked on #203** (holo-gb10), which is stacked on #202 (vision ViT perf). Review/merge **after** #203 — the diff below includes the #202+#203 commits until those land; the FP8-specific change is the single top commit (`feat(qwen35): load compressed-tensors float-quantized FP8`, 4 files).

## What
Loads & runs `Hcompany/Holo-3.1-35B-A3B-FP8` (and dense siblings) as **native FP8** on GB10.

The checkpoint is block-FP8 via compressed-tensors `format="float-quantized"`: an `FP8E4M3 .weight` + a 2D `.weight_scale` per `[128,128]` block (no `.weight_packed`, no `.weight_scale_inv`). The block math is identical to DeepSeek/Qwen block-FP8 — **only the scale tensor name differs** (`weight_scale` vs `weight_scale_inv`). Atlas only recognised the `weight_scale_inv` spelling, so these checkpoints misrouted to the NVFP4 path and failed with `...weight_global_scale not found`.

## How (4 sites taught the compressed-tensors spelling — no kernel/math change)
- `nvfp4_detect.rs` — compressed-tensors config dispatch + heuristic recognise `float-quantized` (and, defensively, the `FP8E4M3` weight dtype) → `Fp8Dequanted` instead of an NVFP4 variant.
- `serve.rs canonicalize_model_quant` — `float-quantized` → `"fp8"` so the kernel/model quant guard accepts it on the nvfp4 bundle (`quant_pair_compatible: nvfp4↔fp8`).
- `loaders_fp8.rs load_fp8_block_scaled_as_fp8weight` — resolve the 2D block scale from `weight_scale_inv` **or** a 2D `weight_scale`; expand a scalar only when genuinely scalar (ModelOpt per-tensor FP8).
- `load_layers.rs proj_is_native_fp8` — accept a 2D `weight_scale` so attention takes the native-FP8 arm.

## Validation (GB10, `Hcompany/Holo-3.1-35B-A3B-FP8`)
- Loads native FP8: logs `quant_format: Fp8`, `Weight format: Fp8BlockScaled`, `MoE experts loaded as native FP8`, `attention FP8 native`.
- Coherent text (Rayleigh scattering / Python / arithmetic), vision all sizes pass, **0 faults**.
[...]

PR #203, opened 2026-06-27
feat(holo): Holo-3.1-35B-A3B / Ornith GB10 enablement — runtime, MoE, GDN, attention, serving

rsafier:
> **Draft — under local validation.** Comprehensive PR bringing up Holo-3.1-35B-A3B (+ Ornith-1.0-9B, Holo-3.1-0.8B) on DGX GB10 (sm_121). Targets `main` directly, so it **includes the #202 vision ViT work** plus everything else below. (#202 is the same vision subset, kept separate for incremental review — merge it first and this rebases, or merge this directly.)

## What's here
- **Runtime primitives** (`spark-runtime`): CUTLASS Sm120 NVFP4 dense + grouped GEMM, cuBLASLt BF16/FP8 GEMM, FlashInfer ragged prefill.
- **MoE**: CUTLASS single-launch grouped NVFP4 gate_up/down (per-group scale, fused gather), W4A8 fallback.
- **GDN/SSM**: FLA chunked-scan prefill, wmma DV-block `chunk_delta`, FP8 SSM projections, batched multi-seq decode.
- **Attention**: CUTLASS NVFP4 + cuBLASLt projection path, paged prefill/decode.
- **Serving**: kernel-batched co-dispatch prefill, radix-KV + Marconi SSM-snapshot prefix caching (+ concurrency-correctness fixes), self-relative auto KV budget, FP8 KV, mixed decode step + CUDA-graph decode.
- **Model enablement**: holo-3.1-35b-a3b / holo-3.1-0.8b / ornith-1.0-9b kernel targets + qwen3.5/3.6 hybrid weight loaders.

## Validation (local, GB10)
- Builds clean on `holo-3.1-35b-a3b` (141 kernels, exit 0).
- **All `.rs` ≤500 LoC** (file-size cap passes; oversized files split into submodules).
- Image correctness (Saturn/Uranus/multi-image) ✓; text + thinking ✓.
- Perf matches the development baseline: image TTFT 0.35→2.41s (C1→C8), cold prefill 0.58→4.27s, decode 84→18 t/s.
- Concurrency: prefix-cache + co-dispatch 8/8 distinct-correct (no bleed); 45s soak = 286 reqs, 0 errors, flat memory.

## Notes
- Some features are env-gated / default-off (experimental): noted in code.
- Kernel targets `holo-3.1-*` / `ornith-1.0-9b` are mostly symlinks to the shared `qwen3.6-35b-a3b` kernels.

[...]

PR #217, opened 2026-06-30
refactor(tokenizer): source chat templates off the model checkpoint (Holo) + grammar/sampling fixes

rsafier:
Draft — stacks on the Holo chain (#203 → #210 → #215 → #214). Net-new vs #214 is the template-sourcing refactor; everything below is the inherited chain.

## What this adds (on top of the chain)
**Sources chat templates from the model checkpoint** instead of the hand-maintained `jinja-templates/{model_type}.jinja` overrides, and ports Atlas's cross-cutting template behaviors into Rust so they apply to every model:
- New `tokenizer/message_preprocess.rs` — auto-close dangling `<think>` before `<tool_call>`, strip/resolve `<|think_on|>`/`<|think_off|>` control, (+ the existing `reasoning_effort→enable_thinking` resolver) run before any template renders.
- `chat_impl.rs` — `preprocess_for_render` wired in; template priority flipped to **model-first**, override opt-in by file presence, `ATLAS_DISABLE_TEMPLATE_OVERRIDES=1` to force model-first universally.
- **Deleted `jinja-templates/holo3_1_moe.jinja`** — Holo now renders off its own checkpoint template (minijinja renders it byte-identical to vLLM/transformers). Other models' overrides (minimax, gemma4, …) kept — they encode jinja-only fixes Rust can't express.
- Stale `render_holo_template` tests replaced with model-template tests in `message_preprocess.rs`.

## Why (root cause from the chain)
The bundled Holo override + the now-removed grammar path made Holo degenerate (runaway reasoning, hallucinated `Human:` turns). Root-caused to the XGrammar **tool grammar** (fixed via `disable_tool_grammar=true` in MODEL.toml, in #203) — *not* the template (minijinja renders byte-identical to jinja2). This PR removes the redundant, drift-prone override and makes template behaviors universal.

## Verification
- 523 lib tests pass; fmt / clippy / typos clean; touched files ≤500 LoC (`chat_request.rs` allow-listed, +rationale).
- End-to-end MCP benchmark on Holo-3.1-35B (real fleet-ops tools), both thinking modes: **no degeneration**, correct tool calls, renders off the model's own template. Thinking-off matches vLLM exactly; thinking-on matches vLLM except one borderline prompt within the NVFP4/FP8 numerical-quant residual.
- Production config kept (NVFP4 KV, min_p floor); vLLM sampling/KV-match experiments intentionally NOT baked in.


rsafier:
Keeping this local for now — not drafting to the main repo yet. Work preserved on the feat/template-source-off-model branch.

PR #218, opened 2026-06-30
dflash: DFlash K=γ correctness + throughput optimizations (verify 585→269ms, tok/s 2.9→22)

Sujimoshi:
## Summary

Full DFlash speculative decode implementation for Qwen3.6-27B-NVFP4 on GB10, progressing from 0% acceptance rate and broken correctness to verified working state with significant throughput gains.

### Correctness fixes

- **YaRN→standard RoPE** in DFlash drafter (YaRN was compressing all freq ×64 on a model with `rope_scaling=null`)
- **BF16/FP32 type mismatch** in `decode_batched_conv_gdn` — BF16 conv output was reinterpreted as FP32 input to `gated_delta_rule_decode`, causing h_state corruption → NaN → EOS collapse → 0% acceptance
- **BF16 embed stride bug** in `verify_d.rs` (`fp32 = 4` instead of `2`) — root cause of garbage output for any K≥5
- **Position IDs** in DFlash ctx: relative `position - eff_ctx + i` → absolute slot positions
- **noise0 conditioning slot** excluded from draft output (logits are untrained)
- **rep_pen/DRY pipeline** bypassed in DFlash verify path (drafter uses raw argmax; applying pipeline only on verifier caused divergence → 0% accept as context grew)
- **EAGLE conditioning**: drafter now sees bonus generator hidden; per-position ctx capture
- **Stochastic acceptance** for T>0 in K=2 verify path
- **Bounds guard** in `try_dflash_capture_all` against buffer row overrun

### Performance

| Milestone | verify_ms | propose_ms | tok/s |
|---|---|---|---|
| Baseline K=γ (after correctness) | 585ms | 110ms | 2.9 |
| ATLAS_VERIFY_PREFILL_SSM/ATTN | 520ms | 110ms | ~13 |
| + wy16 kernel (phase-2 GDN) | 495ms | 110ms | ~13 |
| + propose GEMV→batched GEMM | 495ms | **43ms** | ~13 |
| **+ C1 conv fusion (phase-1)** | **269ms** | 43ms | **~22** |

- **`cfc3012` propose GEMV→batched pipelined GEMM**: O(eff_ctx) sequential GEMVs (each reading 262MB weight matrix) replaced with single `dense_gemm_bf16_pipelined` (tensor-core m16n8k16 MMA, 128×128 tiles, 2-stage cp.async). **110ms → 43ms (−61%)**.
- **`8ea8435` wy16 kernel in phase-2**: native `gdn_wy16` replaces per-token GDN replay. **520→495ms (−5%)**.
- **`7057649` C1 conv fusion in phase-1**: 15 sequential `conv1d_update_l2norm` per SSM layer batched into single `prefill_phase1_verify` pass. **496→269ms (−46%), tok/s 12→22 (+80%)**.
- **`ATLAS_VERIFY_PREFILL_SSM`**: three-phase WY4 batch GDN recurrence instead of sequential `decode_batched` for K≥5.
- **`ATLAS_VERIFY_PREFILL_ATTN`**: single `prefill_attention_paged` pass instead of K decode calls (one KV-cache read).

### Loader

- **`05a3da9` Native U8 NVFP4 loading**: SSM projections now loaded directly as `QuantizedWeight` without dequant→BF16→requant roundtrip. Adds `QuantizedWeight::concat_rows()` for GPU-side QKV+Z concatenation.

### GDN / SSM

- **`19847ae` Skip SSM state norm every 16 tokens**: `do_norm` gate in `gated_delta_rule_decode` kernels, `norm_token_count` tracking per layer.
- **ATLAS_DFLASH_TIMING**: per-phase verify breakdown profiler for data-driven optimization.

### Known limitations (documented, not blocking)

- **Graphed K=γ**: `ATLAS_DFLASH_UNSAFE_KGAMMA_GRAPH=1` flag unlocks; default forces eager. Root: legacy dual-buffer SSM (`pre_verify_copy_async` + `commit_verify_state_async`) absent in working K=2. Fix plan documented.
- **batch>1**: `dflash_hidden_save` is a shared buffer (not per-slot). Per-slot allocation needed for multi-sequence batching.
- **DFlash gated off inside `<think>` blocks** (`mod.rs:340`): vLLM does not gate; impact measured (thinking ON improves Atlas acceptance τ 4.80→6.56).

## Environment flags

| Flag | Default | Effect | Escape hatch |
|---|---|---|---|
| `ATLAS_VERIFY_PREFILL_SSM` | **ON** | Three-phase WY4 GDN batch verify for K≥5 | `=0` |
| `ATLAS_VERIFY_PREFILL_ATTN` | **ON** | Prefill-mode attention in K=γ verify | `=0` |
| `ATLAS_DFLASH_CONV_FUSION` | **ON** | Batch phase-1 conv over K tokens (C1) | `=0` |
| `ATLAS_DFLASH_VERIFY_QKVZ_FIX` | **ON** | Route per-token (M=1) QKVZ projection through the dedicated GEMV serial decode uses, instead of falling through to a GEMM-family kernel at the same shape | `=0` |
| `ATLAS_DFLASH_VERIFY_OUTPROJ_FIX` | **ON** | Per-row GEMV loop for SSM out_proj instead of a separately-stored transposed-weight GEMM | `=0` |
| `ATLAS_DFLASH_VERIFY_GDN_F32` | off (opt-in) | FP32 GDN output precision in the wy16 kernel — mechanism-validated (removes a documented BF16-truncation source), but no measured accept-rate/perf effect at default-ON yet; tracked in #248 | `=1` |
| `ATLAS_DFLASH_TIMING` | off | Per-phase timing breakdown | `=1` |
| `ATLAS_DFLASH_POS_DIAG` | off | Per-position acceptance diagnostics | `=1` |
| `ATLAS_NVTX_DIAG` | off | NVTX call-site instrumentation for profiling (zero-cost when unset) | `=1` |

**Note:** with the first three flags now default-ON, "all flags off" is no longer byte-identical to baseline — it's the legacy path, kept as an escape hatch and for A/B comparison, not the recommended config.

### Verify-step throughput (measured on GB10, shipped config)

| Config | verify_ms/step |
|---|---|
| Legacy (`SSM=0,ATTN=0,CONV_FUSION=0`) | 460.0ms |
| + qkvz/out_proj fixes (now default-ON) | 401.6ms (−12.7%) |
| + conv-fusion (shipped default) | **240.0ms (−47.8%)** |

### Honest perf position

Plain greedy decode on GB10 for this model: **78.7 tok/s**. DFlash K=γ speculative decode in this PR: **~22 tok/s** — still slower end-to-end than plain greedy, because γ=15 with an ~11-15% accept rate doesn't yet amortize the larger per-step verify cost. This PR is a correctness + verify-step-efficiency milestone, not a throughput crossover. Verify-step profiling (nsys/NVTX, see PR discussion) found no further kernel-level win large enough to change that; the next real throughput lever is accept-rate (γ, drafter quality), not kernel work — see #246/#247 for the two profiling-derived follow-ups that were scoped instead.

## Test plan

- [x] T=0 output vs baseline (all flags off) on Qwen3.6-27B-NVFP4 — not byte-identical; divergence traced to a genuine near-zero logit margin (measured floor, not a bug) — see PR discussion for full data.
- [ ] τ≥6.5 on MinHeap code prompt (thinking ON, K=16, EAGLE)
- [x] tok/s ~22 with all perf flags on (K=16, EAGLE) — confirmed, see throughput table above.
- [ ] No regression on other models in support matrix

Made with [Cursor](https://cursor.com)


SeedSource:
**Codeowner triage (SeedSource)** — reviewed as requested. The DFlash K=γ + throughput work looks substantial (verify 585→269ms, 2.9→22 tok/s is a big claim worth a full review), but CI is red on **mechanical** issues that need clearing before the feature review is worth doing:

- `cargo fmt --check`
- `cargo clippy --tests`
- `typos`
- `Enforce ≤500 LoC per source file`
- `Build mdBook + rustdoc`

The branch is also `BEHIND` main. Once you push the fmt/clippy/typos/LoC-split cleanup and rebase, ping me — I'll do a full correctness pass on the K=γ verify path (there's overlap with #209's fixes on the same path, so I'll review them together). Not requesting changes yet since the blockers are all mechanical.

tbraun96:
@Sujimoshi this is strong work — the RoPE fix, the position-id/EAGLE analysis, and the wy16 kernel are all well-reasoned. I checked the wy16 WY-correction math against the per-token recurrence and it's exact, and the intermediates-pool contiguity your base+t*stride writes assume does hold. we also ran your head (a58b14e2) on a GB10 spark today. results first:

| config | T=0 vs plain greedy (same binary, 400-tok code prompt, thinking off) | decode tok/s |
|---|---|---|
| no --dflash | bit-identical to main — KL gate 799/799 positions, mean/max KL 0.000000, coherence 10/10 | 78.7 |
| --dflash, flags off (default) | diverges at char 366 | 3.9 |
| --dflash + VERIFY_PREFILL_SSM/ATTN + CONV_FUSION + EAGLE_FIX + TIMING | diverges at char 366 (same first flip, different continuation from the default path) | 23.7 |

so the non-spec path is clean, but the K=γ verify contract (T=0 output == plain greedy, token for token) is broken in BOTH dflash configs — same first divergence position, which points at a shared verify-path cause, not just the fused branch. and both dflash outputs also differ from each other after that point. timing log on your flags config: verify avg 221.9ms/step (attn 35% p3 40%), which matches your 269ms→ claim — but plain greedy decode on this box is 78.7 tok/s, so even at 23.7 tok/s dflash is currently a 3.3x decode regression on GB10 for 27B NVFP4. worth keeping that on the PR description so nobody flips it on expecting a speedup here yet.

what blocks merge:

1. blocker: trait_decode_batched_conv_gdn.rs:315 and :373 — the K=17/K=16 fused branches run conv1d_update_l2norm for t in 0..K-1 only (main ran all K). wy17/wy16 still read all K rows of conv_out, so row K-1's q/k/v are stale scratch every verify step, and live conv_state is left one token behind — which full-accept rollback then trusts as-is. fix: conv all K tokens, skip only the intermediate copy at t=K-1. given the flags config diverges at the same position, there's at least one more shared bug in the verify path — the "T=0 byte-identical" checkbox needs to actually run in both configs before this lands.

2. concat_rows (quantized.rs:113, used at load_linear_attn.rs:117): the doc says both weights MUST share weight_scale_2 but nothing enforces it, and in_proj_qkv vs in_proj_z are separate tensors that generally carry different global scales in pre-quantized checkpoints — z rows would silently dequant wrong. ensure! equality (fail fast) or rescale z's block scales by the ratio.

3. verify_dflash_step.rs:63 (+ k3/k4): under --dflash you emit raw argmax, so grammar/rep-pen/DRY are bypassed on emitted tokens — that reintroduces the malformed-tool-call/grammar-desync class the pipeline exists for, and we run thinking with presence_penalty. your k2 split (accept raw, emit processed) is the right instinct, but there you can accept drafts[0] and emit a different v0_emit — text diverges from the token history in KV/SSM state. suggested rule for all K: accept raw-vs-raw, pipeline the emitted tokens, and treat a pipeline override at an accepted position as a rejection there.

4. the do_norm skip: on 27b nvfp4 the model-override gated_delta_rule.cu shadows common (12 args, no norm block — extra arg is ignored), so the lever is a no-op on the model you measured, and it's an ungated numerics change for models on the common kernel. strix-hip/metal copies also weren't touched. drop it from this PR or gate it.

smaller: dflash_stochastic_accept's D2H-failure fallback argmaxes the zeroed buffer (verify_k2_step.rs:329) — return false or reuse v0_argmax; move the per-propose info log (propose.rs:335), the ctx_len update log, the PAIR DUMP, and the K4 first-20 timing logs to debug — info is summaries-only here; decode_b.rs is 501 lines so the ≤500 cap will fail CI; and please drop the 10 root-level session docs + bench-gdn.sh + bench/ from scope like you did on #209 (the _typos.toml hash-ignore goes with them). also EAGLE-fix defaulting OFF is odd — if it's correct, make it default and keep the flag as an escape hatch only.

once 1-4 land I'll rerun the full A/B on a spark and we can close #209 in favor of this. want this merged — it's the biggest dflash step so far.

tbraun96:
@Sujimoshi to keep momentum we implemented the review fix-list on top of your head (a58b14e2) as #232 — your three root-cause fixes are the substance, credited. the K=16/17 conv fix + concat_rows ensure + do_norm revert are in; the raw-argmax pipeline redesign is deliberately deferred. T=0 identity retest is running on a spark now. if you spot anything off in how we applied your work, say so on #232 — and your flags-config timing breakdown after the conv fix would be valuable data (does verify move off 222ms?).

tbraun96:
@Sujimoshi pushed the mechanical fix-list onto your branch (91179d8c) — conv now runs all K tokens in the K=16/17 fused branches, concat_rows enforces scale2 equality, the stochastic-accept D2H fallback rejects instead of argmaxing zeros, the ungated do_norm cadence is reverted, noisy logs demoted, decode_b.rs back under the cap, session docs de-scoped. pull before pushing.

important data point: post-fix T=0 retest on a spark STILL diverges from plain greedy at the same char 366 (default config). so the conv skip was real but not the cause of the divergence — there is a second shared verify bug. current suspects, in order: KV rewind off-by-one at the first partial-accept block, SSM rollback snapshot indexing (accept-count -> intermediates slot mapping), bonus-token logits position. we're bisecting on a spark next; if you have PAIR DUMP captures around the first partial accept, they would shortcut this.

Sujimoshi:
## Status update — DFlash not fully fixed yet

Since the last comment, pushed one more commit:

- **`ee3286c`** — `fix(dflash): make EAGLE_FIX default ON` (opt-in → opt-out, escape hatch `ATLAS_DFLASH_EAGLE_FIX=0`)

### What's fixed (3 commits on `fork/optimizations`)

| Commit | Fix | Effect |
|---|---|---|
| `0b10424` | Restore every-16 SSM norm cadence | 0.6% → 25.8% accept, eliminates `"1"` loop |
| `1cdd844` | Off-by-one in `commit_verify_state_async` args | Correctness on partial-accept SSM rollback |
| `ee3286c` | EAGLE_FIX default ON | Removes need for manual env flag |

### What's NOT fixed yet

1. **`enable_thinking: false` silently ignored** — server bug in `chat_request.rs`. The field is `bool` with `#[serde(default)]`, so `false` is indistinguishable from "field absent". `resolve_thinking()` only honors `true`; `false` falls through to `MODEL.toml` default (`thinking_default = true` for Qwen3.6-27B). Fix in progress (`bool` → `Option<bool>`). Not a DFlash bug — server-side template resolution.

2. **Thinking→output seam** — even with DFlash not running during thinking (gate `!inside_thinking`), the per-token capture in `decode_a.rs` populates `ctx_hidden_acc` with thinking-phase hidden states. When DFlash resumes after `</think>`, the first γ verify steps span the boundary — the drafter conditions on stale thinking-phase hiddens, causing output degradation at ~char 280 on longer prompts. The `perf/dflash-stage-optimizations` branch has an approach (`ATLAS_DFLASH_THINK_ENABLE=1` — run DFlash during thinking to eliminate the gap), verified boundary-safe on 1 prompt (+4.9% tok/s), but needs broader testing before shipping.

3. **Raw argmax bypass** (`verify_dflash_step.rs:63` and K=3/K=4 paths) — raw argmax bypasses grammar/rep-pen/DRY. Deferred per @tbraun96's review. Not a T=0 correctness issue but breaks structured output.

4. **CI mechanical blockers** (from @SeedSource's review) — `cargo fmt`, `cargo clippy --tests`, `typos`, ≤500 LoC/file, mdBook. Not yet addressed.

### Current acceptance numbers

| Config | Accept rate | tok/s | Notes |
|---|---|---|---|
| Prime-numbers (short) | 25.8% | 14.2 | Clean output |
| Min-heap (longer) | 9.0% | 13.2 | Degrades at ~char 280 (seam) |

The min-heap degradation is the seam issue — 694 thinking tokens generated (because `enable_thinking: false` is ignored), then DFlash resumes with stale ctx. Fixing #1 (enable_thinking) eliminates thinking for the test; fixing #2 (seam) is needed for production where thinking is required.

### Relationship to PR #232

Our `0b10424` **restores** the ungated every-16 `do_norm` cadence that `91179d8` (and PR #232) reverted per @SeedSource's review concern #4. Our GPU bisect shows this revert caused the 12× acceptance regression. We believe the review concern was incorrect — the every-16 cadence is required for drafter/target state consistency on Qwen3.6-27B. Happy to discuss if there's a model where per-step norm is correct.

Co-Authored-By: Mastra Code (openrouter/z-ai/glm-5.2:nitro?reasoning.effort=xhigh) <noreply@mastra.ai>

Sujimoshi:
**Status update: branch reset + first fix landed from the ongoing T=0 investigation**

Branch state first: we force-pushed `optimizations` back to `91179d8` (your fix-list). The commits that were on top of it (`0b10424..82775ca`) are retracted. In particular `1cdd844` (the "off-by-one fix" in `commit_verify_state_async`) turned out to be wrong: a producer/consumer audit of the intermediates showed the original `num_accepted+1` arguments at `91179d8` are correct (`intermediate[t]` = state after `tokens[0..=t]` with `tokens[0]=last_token`, and the pool is sized γ+1=17 for exactly that convention), and applying `1cdd844` empirically made divergence *earlier*. Our earlier comment calling it a correctness fix was mistaken — please don't base a retest on it. The "added a commit" timeline entries above referencing later SHAs are no longer part of this PR.

First fix from the investigation, landed as `712cedc`:

**EAGLE ctx accumulator gap during plain-decode stretches.** `ctx_hidden_acc` (the drafter's EAGLE conditioning context) was only appended from the DFlash verify/propose branches, never from plain `decode()`. Any decode-only stretch therefore leaves permanent holes in the conditioning context. Concrete repro: with `enable_thinking:false` the model still spontaneously opens `<think>` (we strip it downstream, matching vLLM), and the `!inside_thinking` gate routes those steps to plain decode — in our run the first 8 generated tokens (positions 60-67 after a 60-token prompt) were missing from the accumulator when DFlash resumed, so its first `propose()` conditioned on a context with an 8-token hole. Fix is unconditional: cheap D2D copy per decode step, best-effort no-op when the proposer isn't active. Measured effect on GB10: conditioning integrity restored (verified the append fires exactly on the missing positions); accept-rate on this particular prompt roughly unchanged — this is a correctness fix, not the T=0 blocker.

The T=0 divergence itself is still under active investigation. Short version of where it stands (full writeup with data to follow): the first diverging token reproduces with a *decisive* logit margin (not a near-tie), and it bisects to "any batched K=γ verify computation vs the dedicated K=1 decode path" — all batched variants (fused prefill, `ATLAS_VERIFY_PREFILL_ATTN=0` fallback, `ATLAS_VERIFY_PREFILL_SSM=0` fallback) agree with each other and disagree with serial decode at the same position. So the fault line is batch-K vs batch-1 kernel numerics, not any specific `ATLAS_VERIFY_PREFILL_*` optimization. More once we've isolated the layer/kernel.

Sujimoshi:
@tbraun96 status update following your review — landed on top of your fix-list (`91179d8`), branch state unchanged since the last post.

**Code hygiene:** all diagnostic/bisection scaffolding from the T=0 investigation stripped (grep-verified zero remaining `ITER*`/bisect/replay-prototype identifiers, clean release build, no warnings). Two root-cause fixes flipped to default-ON with `=0` escape hatch, per the f1ae1d1 convention.

**do_norm:** no action needed on our end — you already reverted the ungated cadence change in `91179d8`, and our force-push landed on top of that fix-list, so it isn't part of this PR anymore. (We separately re-verified our earlier accept-rate bisect doesn't reproduce as stated against current HEAD — if we pick this up again it'll be a standalone PR with proper per-model gating, not a reopen here.)

**Timing (your open question — did verify move off 222ms?):** yes. Measured on GB10, same harness:

| Config | verify sum |
|---|---|
| SSM=1,ATTN=1,CONV_FUSION=0, fixes off (closest to your "before" state) | 460.0ms |
| + our two fixes (now default-ON) | 401.6ms (-12.7%) |
| + CONV_FUSION=1 (shipped default config) | **240.0ms (-47.8%)** |

Our two fixes aren't just correctness fixes — they replace a per-token GEMV fallback with a batched GEMM in the phase1 path and are faster on their own. Combined with conv-fusion, verify-step time roughly halves from the pre-fix baseline. Still ~8% above your 221.9ms figure, but we can't reproduce your exact test config 1:1 (unspecified flag combination), so this isn't an apples-to-apples regression claim either way — flag it if you want to align configs precisely.

**T=0 identity — status, and a scope caveat.** We ran T=0 (thinking off, same binary) against plain greedy on a comparable code-generation prompt at extended length (up to 3955/6144 tokens across configs), not your literal 400-token prompt text — we don't have the exact prompt string, only your setup description. Two configs tested:

| Config | First divergence | Margin at divergence (KL_DIAG) |
|---|---|---|
| flags off (SSM=0,ATTN=0,CONV_FUSION=0) | char 79 / token 15 | 0.0625–0.25 (near-tie band) |
| shipped default (SSM=1,ATTN=1,CONV_FUSION=1, fixes on) | char 49 / token ~8 | **0.0000 — exact tie** |

Byte-identical T=0 does not hold on this prompt in either config. But the divergence isn't unexplained: KL_DIAG shows it sits at a genuine near-zero logit margin in the target model's own distribution (p_top1 ≈ 0.45, runner-up effectively tied) — a real tie that flips depending on which numeric kernel path resolves it (batched-verify GEMM vs single-token GEMV), independent of the fixes below. This is the last open item from the T=0 investigation and the natural target for the kernel unification work in Stage 3 below, not something we're papering over.

If you can share your exact 400-token prompt text we'll run 1:1 against it; otherwise we'd suggest treating this extended-prompt run as an equivalent substitute for the identity check, given the mechanism is now characterized rather than mysterious.

**Your three divergence hypotheses (KV rewind off-by-one, SSM rollback snapshot indexing, bonus-token logits position):** all ruled out by direct measurement. Root cause is batched-K verify vs batch-1 decode kernel numerics — specifically two kernel-family mismatches at identical shape (not shape-variance, not an off-by-one in state indexing):
1. **qkvz projection** — phase1's per-token (M=1) QKVZ call fell through to a GEMM-family kernel with no M=1 GEMV specialization; serial decode uses a dedicated GEMV at the same shape. Fixed by routing to the same GEMV kernel serial uses. Dominant contributor (L0 rel_diff 0.845%→0.028%).
2. **out_proj** — serial always uses the untransposed weight via GEMV; batched verify defaulted to a separately-stored transposed weight via GEMM, purely as an artifact of how the weight tensors were populated at load time. Fixed with a per-row GEMV loop. (2.26%→0.10% rel_diff, negligible cost — in fact net-positive per the timing table above.)

Both fixes are in this PR, default-on. The PAIR DUMP capture you asked for around the first partial-accept isn't needed anymore — the root cause was found via direct sublayer A/B probing rather than accept-boundary state tracing.

**Raw-argmax bypass:** filed as #245 with your rule adopted verbatim (accept raw-vs-raw / pipeline emitted tokens / pipeline override at an accepted position = rejection there). Deferred per your original comment, not implemented in this PR.

**Next:** we're profiling the verify step (nsys/ncu, kernel-launch level breakdown) on the shipped config above to find the next real optimization target — FFN GEMM (verify's M=16 currently rides a 128-row prefill tile, so a lot of MMA work is padding) and paged-attention's per-token decode loop are the two live suspects. We don't want to land this as a documented-but-accepted regression; the goal is a further, measured step toward parity or better before this merges. Will post the profile once it's in.

SeedSource:

### Triage
- Rationale: Adds GDN verify and profiling features to transformer and FFN layers; requires serve testing for correctness.
- Serving-path files: `crates/spark-model/src/layer.rs, crates/spark-model/src/layer/transformer_layer.rs, crates/spark-model/src/layers/dense_ffn.rs, crates/spark-model/src/layers/dflash_head.rs, crates/spark-model/src/layers/dflash_head/forward_block.rs, crates/spark-model/src/layers/dflash_head/forward_block_layer.rs`

### Serve gate
- Top-k KL vs main: **0.3193** - **large behavioral shift** - decode delta **+0.1%**

### GLM-5.2 code review
**Correctness concerns:**

1. **`prof!` macro forces sync when profiling** — `ctx.gpu.synchronize(stream)` on every GEMV/GEMM call serializes the pipeline even for users who didn't ask for profiling. This should be gated behind a debug/feature flag, not `ctx.profile`. Throughput claims (22 tok/s) may not hold with profiling enabled.

2. **`output_f32_written: Cell<bool>`** — `Cell` is `!Sync`. If `GdnPrefillBuffers` is shared across threads (it's in a trait that's `Send + Sync`), this won't compile or will panic. Should be `AtomicBool`.

3. **`libloading 0.8.9` added to Cargo.lock** but no corresponding `Cargo.toml` change shown — this is either an orphaned lock entry or a missing diff hunk. Risk of build breakage if the dependency isn't actually declared.

4. **Default trait impls return `Ok(false)`** — safe fallback, but no test coverage shown for the fallback path. If the wy16/split4 kernels silently fail to load, correctness depends on the caller actually checking the bool.

5. **NVTX ranges in batched FFN** are scoped correctly but the `prof!` macro in the GEMV path wraps individual ops — verify this doesn't alter error propagation (macro returns `?` inside both branches, looks fine).

**Risk:** Medium. The `Cell` issue is a compile blocker. The profiling sync is a throughput regression hidden behind a flag.

Sujimoshi:
## Status: ready for merge review (mechanical blockers cleared)

This closes out the correctness + verify-step-efficiency work on top of the original DFlash implementation. Summary of what changed since the last review round, and what's still open.

### Mechanical blockers — all verified individually

Per the original CI triage, each item checked explicitly (not assumed):

- `cargo fmt --check` — clean
- `cargo clippy --tests` (hard errors) — clean; found and fixed two additional hard-errors (`collapsible_if` in `spark-runtime`, `unnecessary_unwrap` in the qkvz-fix's own scope) that only surfaced once upstream crates compiled clean
- `typos` — 0 hits, branch-wide and repo-wide
- mdBook + `cargo doc --workspace --no-deps` — both exit 0, run with the exact commands/env from `docs.yml`
- ≤500 LoC/file — two files over cap split along existing seams (`trait_prefill_gdn.rs` → +`trait_prefill_gdn_batched.rs`; `verify_d.rs` → +`verify_d_kgamma_attn_meta.rs`, catching and fixing a mutex re-lock deadlock risk introduced during the split)
- Rebased clean onto current `main`, force-pushed (`a7100e0`)

### Verify-step profiling (Stage 2/2b, nsys + NVTX call-site attribution)

- **Host-side D2H + sync glue is 43% of wall-clock** — the single largest contributor, bigger than all kernel work combined. This is a scheduler/pipelining issue, not a kernel issue → filed as **#246** (marked priority).
- Stage 2's grid-shape-based kernel attribution left FFN vs. attention out_proj ambiguous (they looked like they might share a kernel path). Stage 2b's NVTX instrumentation resolved this fully: FFN is already on the fast `w4a16_gemm_t_m128` path; attention Q/K/V/O fall through to the slower untransposed `w4a16_gemm` because the `_t` (transposed-weight) variant doesn't exist for those shapes yet → filed as **#247** (cheaper fix, smaller win than #246).
- Pre-registered decision rule for this stage was: only pursue new kernel work if some component is both the largest contributor **and** has a structural-inefficiency hypothesis behind it. Nothing satisfied both — the host-glue issue (#246) is structural but isn't a kernel; the kernel-level candidates (#247) are real but not the largest term. No new kernel work is justified by this profiling pass; the two filed issues capture the actual next levers.
- `ATLAS_DFLASH_VERIFY_GDN_F32`: FP32 GDN-output precision path for the wy16 kernel, opt-in (`=1`). Mechanism-validated — removes a documented BF16-truncation source in the K=γ verify path — but no measured accept-rate or perf delta at default-ON yet, so it stays opt-in pending more data → tracked in **#248**.

### Methodological note: conclusions reversed under masking

Three points in this investigation reached a conclusion that later reversed once a masking factor was removed and the full stack was re-measured:

1. **GDN culprit** — an early root-cause attribution for the NaN/EOS-collapse bug pointed at the wrong layer until the BF16/FP32 type mismatch in `decode_batched_conv_gdn` was found.
2. **Conv defect** — the phase-1 conv correctness issue was initially believed fixed by an unrelated change; only reproduced and correctly attributed once isolated from the conv-fusion batching work.
3. **FFN non-associativity floor** — the T=0 divergence from baseline was initially treated as a bug to chase; re-measurement showed it's a measured floor (near-zero logit margin), not a defect.

(An earlier draft of this note also listed "attention non-effect" as a fourth reversed conclusion. On re-check against the full investigation log, that one was actually *confirmed*, not reversed — iteration 17's full-stack re-measurement showed attention's contribution shrinking through L3→attn_out (0.386%→0.064%), consistent with the original finding that attention translates upstream error rather than creating its own dominant error. Dropped from this list.)

### Performance

Plain greedy decode on GB10 for this model: **78.7 tok/s**. DFlash K=γ speculative decode in this PR: **~22 tok/s** — still slower end-to-end than plain greedy at the current accept rate (γ=15, ~11-15% accept), so this PR is a correctness + verify-step-efficiency milestone, not a throughput crossover yet. The next real lever is accept-rate (γ, drafter quality), not further kernel work on the verify step — see #246/#247 for the two profiling-derived follow-ups that were scoped instead of new kernel work.

On the verify-step throughput table in the PR description: our internal ablation shows a large relative improvement over our own legacy baseline (`SSM=0,ATTN=0,CONV_FUSION=0`), but that legacy baseline isn't the same as your own prior measurement (`SSM+ATTN+CONV_FUSION+EAGLE_FIX+TIMING`, 221.9ms/step), which already includes the conv-fusion win. We attempted to isolate the two new fixes' (QKVZ/out_proj) contribution on top of your baseline directly and hit a genuine confound (verify-step phase timings shift with context length, and the two configs no longer converge identically on the same prompt at current HEAD) that we couldn't cleanly resolve in the time available. Rather than publish a number we can't stand behind, we're flagging this explicitly: **treat the −47.8% figure as measured against our own internal baseline, not as a verified delta against your 221.9ms**. Happy to align on an identical harness/hardware if useful for a clean head-to-head.

### Merge

All mechanical blockers are cleared and the branch is rebased clean on `main` (`a7100e0`). Contributors can prepare/rebase but not merge — this needs maintainer action when ready.

tbraun96:
Picking this up as the grounded DFLASH decode baseline. Confirmed this branch is the real 22 tok/s path (conv fusion `d8aba975`, verify 496 to 269ms, K=16 kgamma + wy16 native kernel on `qwen3.6-27b/nvfp4`), and it test-merges into current main with only 17 conflicts (128 for #132, which is stale off a June-10 pre-refactor base).

Building this branch now for the 27B NVFP4 target and benching decode on the agentic 2.5h subset to confirm 22 tok/s on latest, then I want to stack #132's one unique idea, the fused decode+verify single weight sweep (`verify_fused.rs`), on top of the conv fusion since it attacks the K-independent weight-load floor you documented. Land order looks like #258 glue then this. Will report numbers here. Nice work on the conv batching.

tbraun96:
Benched this on dgx1 (as-is branch binary, 27B-NVFP4 target + z-lab/Qwen3.6-27B-DFlash drafter, greedy). Grounded numbers and a profiling breakdown, plus one bug from the main-merge.

**Config that lights it up** (all opt-in, default off):
`ATLAS_DFLASH_DRAFT_CAP=16 ATLAS_DFLASH_EAGLE_FIX=1 ATLAS_DFLASH_CONV_FUSION=1 ATLAS_VERIFY_PREFILL_SSM=1 ATLAS_VERIFY_PREFILL_ATTN=1`

**Decode tok/s ladder (single-request, 300-tok code prompt):**
| config | mean accepted /15 | tok/s |
|---|---|---|
| no flags | ~0.5 | 3.3 |
| conv+prefill_ssm, EAGLE off | 1.65 | 10.5 |
| + EAGLE_FIX | 3.17 | 15.6 |

EAGLE clearly works (1.65→3.17 accepted, +48% tok/s), conv-fusion + prefill-SSM work. Output stays coherent, zero crashes.

**Per-phase verify (ATLAS_DFLASH_TIMING, avg/100 steps):**
`attn=41.4  p1_conv=18.9  p2_gdn=14.6  p3=100.7  head=21.3  | sum=196.9ms`
p3 (norm + GDN out_proj + dense FFN) is **51%**. Note the whole verify sum (197ms) is already under the 269ms this PR targeted, so verify latency is not the blocker.

**The gap to 22 is acceptance, not kernels.** I get τ≈4.2 tok/step (3.17 accepted + bonus); the EAGLE commit measured τ=6.67. That is almost exactly the 15.6-vs-22 delta. My acceptance sits at ~half the commit's across the whole run. What target + workload did you measure τ=6.67 on? If it was a different target quant (FP8/BF16 27B, or the 35B), that would explain it — the NVFP4 target's argmax may pair worse with the BF16-trained drafter. Want to line up configs so we are comparing the same thing.

**Merge note:** I resolved this branch onto current main (27 hunks, builds clean) but the merged binary crashes the DFLASH path at ~9 tokens (`free_sequence: cuMemFree` then a `cuMemsetD8Async` cascade). The as-is branch binary is clean, so the merge introduced it — most likely the composed final-logits GEMM in `dflash_head/forward_block.rs`. Chasing that next.

tbraun96:
Deeper dig on the acceptance gap, and I found a real conditioning bug plus the ceiling.

**Bug: the capture-layer offset default is wrong for this drafter.** `factory/build.rs` subtracts 1 from the drafter's `target_layer_ids` (HF `output_hidden_states` semantics). The z-lab 27B DFlash drafter declares raw ids `[1,16,31,46,61]`. A/B on the live decode:
| CAPTURE_LAYER_OFFSET | capture idx | accepted/15 | tok/s |
|---|---|---|---|
| -1 (current default) | [0,15,30,45,60] | 3.17 | 15.6 |
| **0 (raw ids)** | **[1,16,31,46,61]** | **3.55** | **17.2** |
| +1 | [2,17,32,47,62] | 3.27 | 15.4 |

So `ATLAS_DFLASH_CAPTURE_LAYER_OFFSET=0` is +10% decode, the drafter conditions best on the raw ids it was trained on. Suggest flipping the default (or at least documenting 0 as the 27B value). Happy to push the one-liner to your branch.

**Full grounded ladder (27B-NVFP4, greedy, code prompt):** plain no-spec 14.0 → DFLASH default 15.6 → **+offset=0 = 17.2** (best). γ=8 and ctx_window don't move it (accepted plateaus ~3.5/step, tail pos 8-15 contribute almost nothing).

**Why not 22 — it's the drafter, quantified.** KL_DIAG over 2240 verify positions: the target is **confident (p_top1≥0.9) 72%** of the time, yet the drafter only lands ~3.5 accepted/step (~24%). So it is missing ~half the *easy* positions, not the hard ones. That is a drafter-vs-target distribution mismatch (BF16-trained drafter vs NVFP4-served target argmax), not a kernel or config issue on our side. My τ=4.5 vs your commit's 6.67 is exactly this. **What target quant + workload did you measure τ=6.67 on?** If it was FP8/BF16, the sharper argmax would explain the delta and point us at needing an FP8 27B target (no Atlas kernels for that yet).

Verify latency is not the issue — sum is 188-197ms, already under the 269ms this PR targeted.

tbraun96:
Update — I was wrong that 17.2 was the ceiling. The acceptance gap is the NVFP4 target's noisy argmax vs the BF16-trained drafter, so sharpening the target's argmax raises acceptance monotonically, and tok/s tracks it:

| config (all on top of offset=0 + full flags) | accepted/15 | tok/s |
|---|---|---|
| NVFP4 target | 4.55 | 17.2 |
| **Qwen3.6-27B-FP8** target (sharper hidden) + `--lm-head-dtype bf16` | 5.22 | 18.2 |
| **+ `--kv-cache-dtype bf16`** | 5.35 | **19.0** |

So 19.0 tok/s, +36% over the plain 14.0 baseline, is the best cached-target result. Nice detail: the FP8 target is not 2x slower — the verify re-quantizes to nvfp4 for the kernels, so only the captured hidden states get the FP8 sharpness. GDN_F32 and gamma≠16 both hurt.

**Path to 20+ is the BF16 base target** (`Qwen/Qwen3.6-27B`), the drafter's exact training distribution, which should push acceptance to the ~6.67 you measured. I could not finish the 54GB pull here (4/15 shards keep failing), but the mechanism is proven and monotonic, so a BF16 target is very likely 20-22. The durable fix is FP8-native 27B kernels so we get a sharp argmax without the download or the BF16 forward cost.

Also still standing: the capture-offset default should be 0 for this drafter (+10%), and the #218→main merge crashes DFLASH at 9 tokens (as-is branch is clean).

Sujimoshi:

PR #220, opened 2026-07-01
perf(gdn): FlashInfer GDN via AOT bridge (opt-in) [draft, stacked on #215]

rsafier:
> **Draft — stacked on #215** (`feat/dense-fp8-attn-ffn`). Do not merge until #203→#215 land.

## FlashInfer GDN via AOT bridge (behind `ATLAS_GDN_FLASHINFER=1`)

The gated-delta-rule (GDN) recurrent scan is Atlas's single biggest prefill cost and a hard ceiling on decode — the hand-written scalar-FFMA scan runs ~11–13× slower than FlashInfer's tensor-core GDN at matched shapes. This wires FlashInfer's `delta_rule` kernel into Atlas's live prefill path as an opt-in fast path, **plus the reproducible builder that bakes the GDN runtime into the serve image**.

### What's here
- **AOT bridge** (`3rdparty_patches/gdn_aot/`): export FlashInfer's `delta_rule` for sm_121 (patch + `gdn_export.py`/`gdn_dump_meta.py`), a C++/CUDA harness, and the linked `.so`s + generated Rust FFI (`gdn_rs.rs`). `STATUS.md` documents the pipeline.
- **Layout adapter** (`gdn_flashinfer.rs`): bit-exact packed-QKV + gate deinterleave between Atlas's native layout and FlashInfer's expected inputs.
- **Live wiring** (`trait_prefill_gdn.rs` / `trait_prefill_recur.rs`): routes `prefill_gdn_recurrence` through FlashInfer when `ATLAS_GDN_FLASHINFER=1`, with multi-chunk prefill state-carry (transpose incoming `h_state` → FI init).
- **Reproducible builder** (`docker/gb10/Dockerfile.builder`): pins CUDA 13.2 + CUTLASS + FlashInfer + the CuTe-DSL runtime and bundles the GDN AOT libs (`libatlasgdn.so` + `libcute_dsl_runtime.so` + cuda-13.2 compat for sm_121a). This is the deploy mechanism — without it, `ATLAS_GDN_FLASHINFER=1` has no runtime libs.
- **Examples**: `gdn_fla_vs_fi.rs` (head-to-head vs the scalar FLA scan), `gdn_chunk_shapetest.rs`.

### Status
e2e **coherent** — prefill correct + ~1.3× isolated speedup; the k↔v state transpose fix makes decode correct too. Default **off** (`ATLAS_GDN_FLASHINFER` unset → existing FLA scan).

Builds clean all-targets; fmt/typos/file-size/SPDX/metal green. Remaining bench/tooling commits (MoE microbench, `ATLAS_MAX_TOKENS_TOTAL`) and the tokenizer commits are intentionally **not** here — separate follow-up.


tbraun96:
@rsafier solid work getting the FI GDN scan live end-to-end — the diagnostic chain (bf16 re-export root-cause, managed scratch, k<->v state transpose, multi-chunk carry, cross-impl cos=0.999993) is exactly the evidence we want in a PR. holding until the #203->#215 stack lands, but flagging what needs fixing for readiness:

1. Dockerfile.builder relinks libatlasgdn.so from 3rdparty_patches/gdn_aot/gdn_holo_0.o, but the .o isn't committed — only the pre-linked .so's are. docker build fails at that step, and the Dockerfile is this PR's deploy mechanism. commit the .o as the single artifact (with the export recipe pinned in STATUS.md) and drop both .so blobs from git — binaries in the repo are unauditable and gdn_holo.so isn't even loaded by anything.

2. real race in atlas_gdn_prefill_packed_managed: the cu_seqlens rebuild is a synchronous cudaMemcpy on the legacy null stream, but atlas creates its compute streams CU_STREAM_NON_BLOCKING (gpu_impl.rs:393), so there's no ordering vs FI kernels already queued on the serve stream. when total changes — the short final chunk of a chunked prefill, or varying request lengths in the batched loop — the host can overwrite m_cu while up to 48 queued launches still read it. your needle test passing is timing luck. write cu on the same stream (cudaMemcpyAsync from pinned storage or a 2-word device-side write) and drop the host-side total cache.

3. shim hygiene: cudaMalloc/cudaMemcpy returns are ignored (OOM becomes a null-ptr launch with a misleading downstream error), the static scratch has no thread guard, and num_seqs>1 silently overflows the 16-byte m_cu — assert num_seqs==1 since the rust side hardcodes it.

4. the gate is kd==vd==128 only but the kernel's only been validated at holo's nk=16/nv=32. descriptors are dynamic so 27b probably works — either validate on 27b or add head counts to the gate until then.

5. smaller: cache the env read in available() (OnceLock, like gdn_batched_fla_enabled — it's re-read per layer per chunk), and gdn_chunk_shapetest lost its batch 2/4 + short-length default coverage — keep the old sweep, add yours behind an arg.

the wiring itself reads clean: gated identically to the FLA path, fallback intact, errors propagate, default-off is genuinely zero-impact. once the stack lands and 1+2 are fixed this is a good opt-in lever. can you also paste the 8k/16k prefill A/B numbers into the PR body so we can hold them against our dgx2 rerun?

rsafier:
Worked all five review items + validated e2e in the CUDA-13.2 container. Two structural findings below matter more than the numbers.

## Review items (pushed)
1. **`.o` committed / Docker build** — committed `gdn_holo_0.o` (gitignore exception), dropped the `.so` blobs. Building the branch in `Dockerfile.builder` then surfaced **4 more build breaks** (all fixed in a follow-up commit): cute-runtime discovery (glob→`find`; the `.so` moved to the `-libs-cu13` pkg in cutlass-dsl 4.5.0), cudarc panics on CUDA 13.2 (`CUDARC_CUDA_VERSION=12000`), builder stage missing nccl (`-lnccl`), and the relink missing `-L` for cudart. **A clean `docker build` now produces a working FI-GDN image** — item 1 is genuinely resolved.
2. **cu_seqlens race** — device-side `atlas_write_cu_seqlens` on the caller's stream + dropped the host `total` cache.
3. **shim hygiene** — cudaMalloc return checks, `recursive_mutex` around the global scratch, reject `num_seqs != 1`.
4. **gate** — restricted to validated `nk=16/nv=32` in both prefill paths.
5. **env cache + shapetest** — `available()` OnceLock; shapetest batch-{1,2,4}+short default, `--long` for the perf sweep.

## e2e correctness ✅
Holo-3.1-35B-A3B in the container with FI-GDN active: full suite green (coherence, tools, **long-context needle correct @16K** — which exercises the multi-chunk prefill state-carry the managed entry exists for). The earlier standalone `gdn_harness_packed` cos=0.30 was a stale-test-harness artifact, not a shipped-path bug.

## Prefill A/B (CUDA-13.2 container, single-stream, `max-batch-size=1`, chunked prefill 8192)
| prompt tok | FLA | FI-GDN | Δ |
|---|---|---|---|
| 3,276 | 1,868 tok/s | 2,218 tok/s | +19% |
| 6,511 | 1,977 | 2,377 | +20% |
| 14,236 | 1,875 | 2,232 | +19% |
| 28,458 | 1,598 | 1,838 | +15% |
(Consistent +15–20% e2e prefill; measured on a contended gx10 so absolute tok/s is lower than a clean dgx2 run, but the ratio holds out to 32K.)

## ⚠️ The finding that matters most for merge
**FI-GDN only engages in the single-stream prefill path.** The production perf config uses `ATLAS_PREFILL_CODISPATCH=1` (+ batched-first-chunk, max-batch>1), which routes GDN prefill through the **batched** path — that has **no FI-GDN gate**, so the kernel is bypassed and FLA runs. Confirmed directly: with co-dispatch on, no `FlashInfer GDN kernel loaded` log and tok/s == FLA; drop it and FI engages + the +15–20% appears. So **as-gated, #220's lever does not fire in the deployed co-dispatch config** — the gate needs extending to the batched/co-dispatch GDN prefill path before the win materializes in production.

SeedSource:

### Triage
- Rationale: Documentation, build artifacts, and harness code for GDN AOT integration; no production inference code changes.
- Serving-path files: `3rdparty_patches/gdn_aot/delta_rule_sm120_aot_export.patch, 3rdparty_patches/gdn_aot/gdn_dump_meta.py, 3rdparty_patches/gdn_aot/gdn_export.py, 3rdparty_patches/gdn_aot/gdn_harness.cpp, 3rdparty_patches/gdn_aot/gdn_harness_packed.cpp, 3rdparty_patches/gdn_aot/gdn_holo_0.h`

### Serve gate

### GLM-5.2 code review
**Correctness:** The diff is documentation-only (STATUS.md, scripts, .gitignore). No production Rust/CUDA code is modified. The STATUS.md claims bit-exact prefill but explicitly notes **decode is broken** (k<->v state transpose) and e2e speedup is only 1.3× (not the claimed 11×). The "opt-in" flag (`ATLAS_GDN_FLASHINFER`) is not wired into any Atlas crate in this diff.

**Risk:** Committing `gdn_holo_0.o` (a binary build artifact) via `.gitignore` negation is a maintenance hazard — no source-level reproducibility for the object. The patch hardcodes `sm_121a` and CUDA 13.2, creating a platform lock-in with no fallback path documented. The `gdn_dump_meta.py` writes to `/tmp/gdn_ref` (absolute path, no env override).

**Quality:** STATUS.md is 88 lines of stream-of-consciousness log with contradictory historical sections ("garbage" → "bit-exact") that should be squashed. The export scripts monkey-patch `cached_compile` via module mutation — fragile against FlashInfer upstream changes. No CI integration, no tests.

**Verdict:** Block. This is a draft dumping experimental artifacts. Strip the binary object, clean up STATUS.md to final-state only, and don't merge until the decode transpose fix and actual Atlas crate integration land.

tbraun96:
@rsafier — the FlashInfer-GDN AOT work landed via #229 (squash-merged to `main`; `feat/gdn-flashinfer-aot` was folded into that branch). If #220 is fully incorporated you can close it — or if there's a delta still outstanding on this branch, let me know and I'll help reconcile against latest `main`.

rsafier:
Confirmed fully absorbed by #229 — closing.

Verified against current `main`: this PR's FlashInfer-GDN AOT files are all present on main (`3rdparty_patches/gdn_aot/*`, `crates/spark-model/src/layers/ops/gdn_flashinfer.rs`, `qwen3_ssm/trait_prefill_gdn.rs`, …). Diffing `main` against `feat/gdn-flashinfer-aot` restricted to this PR's own files shows **+0 / −54** — main is a strict superset with no content unique to this branch (the −54 is later refinement that landed with #229). No outstanding delta. Thanks for folding it in!

PR #221, opened 2026-07-01
chore(tooling): tokenizer-from-checkpoint + ATLAS_MAX_TOKENS_TOTAL + benches [draft, stacked on #220]

rsafier:
> **Draft — stacked on #220** (`feat/gdn-flashinfer-aot`, which is stacked on #215). Do not merge until the stack + #220 land.

## Tooling + tokenizer follow-up

The remaining non-GDN work carried on the original `wip/gdn-flashinfer-aot` branch, split out of #220 so the GDN PR stays focused.

### Tokenizer
- **Source chat templates from the model checkpoint** (`feat` + `refactor`): read the Jinja chat template off the model dir instead of hard-coding per-family templates in Rust; ports the Atlas-specific template behaviors onto that path. The ported cross-cutting behaviors (developer→system remap, `<think>` autoclose, `<|think_on|>`/`<|think_off|>` control) now run ahead of **every** model's own template — see the note below.

### Server
- **`fix(log)`**: demote the `CUTLASS_NVFP4_ROUTE` dispatch diagnostic from WARN → debug (it fired per-request).

### Bench
- **MoE GEMM microbench**: Atlas CUTLASS NVFP4 grouped GEMM apples-to-apples vs MARLIN, plus a W16A16 BF16 baseline (`crates/spark-model/examples/nvfp4_gemm_bench.rs`).
- **Concurrency bench**: Bearer auth, `BENCH_CONC`, reasoning-token TTFT, `BENCH_NO_THINK`.
- **Soak**: env Bearer auth + `--vision-only` (pure image-concurrency).

Builds clean all-targets; fmt/typos/SPDX/file-size green.

### Note on runtime behavior
The tokenizer path is the functional change: the cross-cutting message preprocessing (developer→system remap, think-autoclose, think-control) now applies to **every** model by default, rendering off each model's own template rather than a per-family Rust override. A 27B prompt-token parity check vs `main` is planned before the stack merges.

> `ATLAS_MAX_TOKENS_TOTAL` was **dropped** from this PR (per review) — #162 owns the reasoning-vs-`max_tokens` semantics unconditionally, so shipping a flag here would have been dead code once #162 lands.


tbraun96:
@rsafier good split — the message_preprocess port is faithful to the deleted override (checked line-by-line against the old jinja: per-message last-wins think-control, rfind autoclose semantics, all match) and the fixture-based tests proving the behaviors are load-bearing are exactly what we want. four things before this is ready:

1. developer-role regression. the deleted holo override accepted role=developer in three places, and the model's own template raises 'Unexpected message role.' for it — so a Holo request with a developer message now hard-fails rendering. the old test that pinned this (render_holo_template_accepts_vllm_thinking_controls, it used role: developer) was deleted rather than ported. can you add a developer→system remap in preprocess_for_render (cross-cutting, same spirit as the other behaviors) plus a test.

2. ATLAS_MAX_TOKENS_TOTAL collides with #162, which makes reasoning-count-against-max_tokens unconditional on the same lines of emit_step/decode_logits_step. we should decide the semantics once instead of shipping both — if #162's unconditional vLLM parity lands, this flag is dead code. let's sync on that before either merges.

3. PREFILL_GPU_BUBBLE_ANALYSIS.md doesn't belong in the repo root or this PR — drop it or move to docs/.

4. nvfp4_gemm_bench warmup error path: the continue targets the inner warmup loop, so after freeing act/out you call the gemm again on freed buffers and double-free at the end. labeled continue to the M loop (or break) after the frees.

nits: parse_disable_watchdogs reused for the new env var — rename to a generic parse_env_bool; the test-side preprocess() mirror could just call the real preprocess_for_render (make it pub(crate)); the dead TEMPLATE_OVERRIDE_DIR const in chat_impl duplicates jinja_helpers; nvfp4_gemm_bench required-features should include \"cuda\" like the sibling examples.

one more note: the think-control strip + autoclose now run for every model by default, not just holo — that's the point and it looks right, but \"no runtime behavior change by default\" in the body isn't quite accurate. we'll run a prompt-token parity check vs main on the 27B before the stack merges. keep it draft until #220 lands.

rsafier:
Thanks for the thorough pass @tbraun96 — pushed `d74e31a` addressing everything except the #162 collision (needs a decision, below).

**1. developer-role regression** — fixed. Added `remap_developer_role` (developer→system, shared system-instruction semantics) as behavior 0 in `preprocess_for_render`, ahead of the model template. Ported it with a unit test (`remap_developer_to_system`) **and** the e2e render test that was dropped — `holo_renders_developer_message_as_system` renders a `role: developer` message through the Holo model's OWN template and asserts it comes out under `<|im_start|>system` framing (would panic on the template's `Unexpected message role.` raise without the remap).

**3. PREFILL_GPU_BUBBLE_ANALYSIS.md** — `git mv`'d to `docs/`.

**4. nvfp4_gemm_bench double-free** — fixed in both `bench` and `bench_bf16`: labeled the M loop and `continue 'm` after the frees, so a warmup error skips the rest of that shape instead of re-launching on freed buffers + double-freeing at end-of-iteration.

**Nits** — all done: `parse_disable_watchdogs`→`parse_env_bool`; the test `preprocess()` mirror now calls the real `pub(crate) preprocess_for_render` (no drift); dropped the dead `TEMPLATE_OVERRIDE_DIR` const in chat_impl; `nvfp4_gemm_bench` required-features now `["cuda", "gpu-examples"]`.

`cargo fmt --check` clean; the 16 tokenizer/helpers lib tests pass incl. the two new ones.

**2. ATLAS_MAX_TOKENS_TOTAL vs #162** — agree we shouldn't ship both. #162 is still open, so I left the flag in place rather than pre-emptively deleting it. Proposal: if #162's unconditional vLLM-parity behavior is the direction, I drop `ATLAS_MAX_TOKENS_TOTAL` + `max_tokens_counts_reasoning()` from this PR entirely and #162 owns the semantics; if #162 might not land soon, we keep the opt-in flag as a stopgap and #162 removes it when it merges. Which way do you want it? Happy to cut the flag from this branch in a follow-up commit once we call it.

Also fair point on the PR body — the think-control strip + autoclose now run for **every** model by default, so "no runtime behavior change by default" isn't accurate; that line's on me. Ready for the 27B prompt-token parity check whenever. Staying draft until #220 lands.

rsafier:
Done — dropped `ATLAS_MAX_TOKENS_TOTAL` in `4666dce`. #162 owns the reasoning-vs-`max_tokens` semantics now. Removed `max_tokens_counts_reasoning()` + its `OnceLock` and reverted both thinking-token budget guards (emit_step + decode_logits_step) to the prior default (reasoning "free", bounded by the thinking budget). `consume_generation_budget()` stays (content path still uses it); `parse_env_bool` stays (now just gates `ATLAS_DISABLE_WATCHDOGS`). Lib compiles + fmt clean.

Also refreshed the PR body: removed the flag bullet and corrected the "no runtime behavior change by default" line to call out that the think-control/autoclose/developer-remap preprocessing now runs for every model.

SeedSource:

### Triage
- Rationale: Adds nvfp4 gemm benchmark and updates concurrency bench with auth and reasoning token handling.
- Serving-path files: `crates/spark-model/src/layers/ops/dispatch_helpers.rs, crates/spark-server/src/openai/chat_request.rs, crates/spark-server/src/scheduler/helpers.rs, crates/spark-server/src/tokenizer.rs, crates/spark-server/src/tokenizer/chat_impl.rs, crates/spark-server/src/tokenizer/message_preprocess.rs`

### Serve gate

### GLM-5.2 code review
**Correctness concerns:**

1. **`bench-atlas-concurrency.py`**: `AUTH_HEADERS` is computed at import time from env vars. If `_API_KEY` is unset, `AUTH_HEADERS = {}` — fine. But `BENCH_CONC` parsing does no validation; malformed input (`"1,abc"`) crashes with an opaque `ValueError` at module load. Minor, but add a try/except or validation.

2. **`chat_request.rs`**: The new test asserts `resolve_thinking(false)` returns `(true, Some(512))` for effort=high. The `false` input is the *default* state — confirm `resolve_thinking`'s signature is `(default_on: bool) -> (bool, Option<u32>)` and that "high" unconditionally overrides to `true`/512 regardless of default. If default means "user didn't specify," the semantics are correct; if it means "force off," the test would be wrong. Verify the resolver logic (not shown in diff).

3. **`nvfp4_gemm_bench.rs`**: Error-path `continue 'm` correctly frees `act`/`out` before skipping. However, `w` (in `bench_bf16`) and `wpt`/`wst` (in `bench`) are allocated *outside* the `m` loop — if the first `m` iteration errors and frees `act`/`out`, subsequent `m` values still allocate fresh `act`/`out`. This is correct. But if `g.alloc` for `act`/`out` fails (returns Err), it's `?`-propagated, abandoning `w`/`wpt`/`wst` — a GPU memory leak on alloc failure. Low-risk for a bench example but worth a comment.

4. **`file-size-cap.yml`**: Adding `chat_request.rs` to the allowlist at 509 LoC is reasonable given the justification. Confirm the cap check is a `<=` comparison, not `<`.

**Risk:** Low. Changes are tooling/bench/test-only; no runtime hot-path modifications. The `tracing::warn!` → `debug!` change in `dispatch_helpers.rs` is safe and reduces log noise.

**Quality:** The bench script's reasoning-effort/thinking-token comparability comment is well-documented. The `nvfp4_gemm_bench` example is clean but would benefit from a `Drop`-style RAII guard for GPU allocations rather than manual `g.free()` calls scattered across error paths.

**Verdict:** Approve with the `BENCH_CONC` validation nit addressed.

rsafier:
Superseded by #264 — closing. #221 was cut on the now-closed #220 branch (orphaned base). The tokenizer-from-checkpoint feature + tooling is recut directly onto post-#229 main in **#264**, reconciled against main's parallel tokenizer refactor (external tests, split jinja_helpers) with fmt+clippy+test all green.

Issue #192, opened 2026-06-22
[bug] Parallel tool calls only execute first tool request


mvandergrift:
### Atlas version / docker tag

latest

### Model

Qwen/Qwen3.6-35B-A3B-FP8

### Hardware

1xGB10

### Launch command

```bash
docker run --rm --name atlas \
  --network host --gpus all --ipc=host \
  -v ~/.cache/huggingface:/root/.cache/huggingface \
  avarok/atlas-gb10:latest \
  serve Qwen/Qwen3.6-35B-A3B-FP8 \
    --port 8000 \
    --max-seq-len 200000 \
    --kv-cache-dtype bf16 \
    --kv-high-precision-layers max \
    --gpu-memory-utilization 0.70 \
    --scheduling-policy slai \
    --enable-prefix-caching \
    --speculative \
    --tool-call-parser qwen_coder --bind 0.0.0.0
```

### Expected behaviour

Sending a request for parallel tool execution results in multiple tool calls in the output.

```bash
curl http://your-atlas-host/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "Qwen/Qwen3.6-35B-A3B-FP8 ,
    "messages": [
      {
        "role": "user",
        "content": "What is the current weather in London, Tokyo, and New York?"
      }
    ],
    "tools": [
      {
        "type": "function",
        "function": {
          "name": "get_weather",
          "description": "Get the current weather for a location",
          "parameters": {
            "type": "object",
            "required": ["location"],
            "properties": {
              "location": {
                "type": "string",
                "description": "City name"
              }
            }
          }
        }
      }
    ],
    "tool_choice": "auto"
  }'
```

### Actual behaviour

Output only contains a single tool call. The model clearly reasons about its intention to call the tool multiple times:
```
"I need to call get_weather three times"
"[Tool Call Generation] -> get_weather x 3 ✅"
"(Proceeds to output tool calls)"
```

However, the output for any parallel tool execution is only one tool call:
```json
{
  "id": "chatcmpl-c20dc530-6671-43ba-9bd6-93cf71392398",
  "object": "chat.completion",
  "created": 1782151176,
  "model": "Qwen/Qwen3.6-35B-A3B-FP8",
  "system_fingerprint": "fp_atlas",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
[...]

PR #223, opened 2026-07-01
perf(qwen3.6-27b): attn transposed NVFP4 copies in qwen35_dense loader (~1.3x prefill)

tbraun96:
## Summary

The `qwen35_dense` loader builds transposed weight copies for the SSM/GDN projections but **not** for the full-attention q/k/v/o projections. On the NVIDIA mixed-precision `Qwen3.6-27B-NVFP4` checkpoint (whose attention projections are NVFP4), those 16 layers' 4 projections/layer therefore fell to the ~10 TFLOP/s base `w4a16_gemm` instead of the fast `w4a16_gemm_t_m128` prefill kernel the dense FFN already uses.

`nsys` (graphs-on) measured that base kernel at **28.8% of prefill GPU time** — 64 GEMMs/prefill (Q+gate `n=12288`, K/V `n=1024`, O `n=5120`). This PR builds the transposed copies + `set_prefill_weights` (mirroring the `attention_arms` loader), routing them to `t_m128`.

## Correctness

Not strictly bit-identical: `t_m128` quantizes activations to FP8 E4M3 (the base `w4a16_gemm` consumed BF16 activations) — the same route dense-FFN prefill and MoE-model attention already take. `predequant_for_prefill()` is deliberately not called (FP8-predequant measured slower for these bandwidth-bound GEMMs). Gated on hardware instead of asserted:

- **Coherence**: 10/10 (suite) on unsloth/Qwen3.6-27B-NVFP4, single Spark.
- **KL/logit gate vs main** (greedy, thinking off, 14 prompts, top-20 logprobs): greedy match 93.7%, mean KL 0.023, flips only on creative near-ties. PASS.
- **BFCL v4 single-turn A/B, identical harness/config/box** (995 samples, seed 42, temp 0): **47.04 vs main 46.93 overall** (+0.11, neutral); hallucination 97.16 on both; live 54.12 vs 53.53; all subset deltas within run noise. (Absolute level reflects current harness scoring, including structural parallel-call zeros on both sides — tracked in #192.)

## Prefill before → after (NVIDIA Qwen3.6-27B-NVFP4, clean, GB10 sm_121f)
| ctx | before | after | speedup |
|---|---|---|---|
| 2k | 3.28s (614 tok/s) | 2.42s (833) | 1.36× |
| 4k | 6.50s (617) | 4.78s (838) | 1.36× |
| 8k | 13.08s (612) | 9.75s (821) | 1.34× |
| 16k | 28.56s (560) | **21.88s (730)** | **1.30×** |

16k TTFT −6.7s (−23%), prefill throughput +36%.

## Notes
- Only affects the `qwen35_dense` path with NVFP4 attention projections; FP8-attn and non-NVFP4 checkpoints are unchanged (guarded on `Some(q_nvfp4)`).


tbraun96:
post-merge note: the squash here accidentally picked up ~286 files of in-flight campaign work pushed to this branch after the gated head (249ca68a) — MMQ vendoring, a feat/gdn-flashinfer-aot merge, 35B kernels. reverted and re-landed just the gated one-file fix in #226; main now differs from pre-merge only by qwen35_dense.rs (+19/-2), bit-identical to the tree that passed coherence/KL/BFCL gates. the campaign work is untouched on perf/qwen36-27b-prefill-nvfp4 — bring it back as its own PR(s) with gates, and note it currently embeds #220's content which has an open blocker.

PR #224, opened 2026-07-02
fix(stream): don't strip the `tool` BPE fragment of tool_* names mid tool-call (#222)

### LINKED ISSUE #222: Streaming tool-parser drops the first token of tool names that collide with `<tool_call>` markup (tool_search → _search)
**Atlas image:** `avarok/atlas-gb10@sha256:1f05897…` (== `latest`/`nightly`/`dev`, 2026-06-12). Also reproduced on `3.0.1`/`ornith` (2026-06-26).
**Model:** `nvidia/Qwen3.6-35B-A3B-NVFP4`
**Hardware:** DGX Spark (GB10)
**Serve flags:** `--tool-call-parser qwen3_coder --speculative --num-drafts 2 --enable-prefix-caching --scheduling-policy slai` (also repro'd with `--tool-call-parser hermes` and with `--speculative` removed).

### Summary
When **streaming** (`"stream": true`) with `"tool_choice": "auto"`, Atlas drops the leading `tool` token of any tool **name** that collides with the `<tool_call>` markup the parser scans for. The name is emitted one token short:

| requested tool name | streamed name |
|---|---|
| `tool_search`   | `_search`   |
| `tool_describe` | `_describe` |
| `tool_call`     | `_call`     |

**Non-streaming** responses return the correct full name. Only names that collide with the markup are affected — ordinary tool names (`findManyCompanies`, `create_document`, `get_weather`, `triage_inbox`) stream fine.

### Impact
Breaks any progressive-tool-disclosure / "tool search" bridge whose control tools are named `tool_search` / `tool_describe` / `tool_call` (e.g. Nous **Hermes Agent**'s Tool Search). The client receives a mangled name and either errors or fuzzy-misroutes it to the wrong tool.

### Minimal deterministic repro
```python
import json, urllib.request
tools = [{"type":"function","function":{"name":"tool_search","description":"Search for tools",
    "parameters":{"type":"object","properties":{"query":{"type":"string"}}}}}]
body = {"model":"pl-main",
        "messages":[{"role":"user","content":"Call the tool_search tool now with query CRM."}],
        "tools":tools,"tool_choice":"auto","stream":True,"max_tokens":300,"temperature":0.2}
r = urllib.request.urlopen(urllib.request.Request("http://HOST:8001/v1/chat/completions",
        data=json.dumps(body).encode(), headers={"Content-Type":"application/json"}))
name = ""
for line in r:
    line = line.decode().strip()
    if line.startswith("data:") and line[5:].strip() not in ("", "[DONE]"):
        for tc in (json.loads(line[5:]).get("choices",[{}])[0].get("delta",{}).get("tool_calls") or []):
            if (tc.get("function") or {}).get("name"): name += tc["function"]["name"]
print(repr(name))   # observed: '_search'   |   expected: 'tool_search'
# same request with "stream": False → correct 'tool_search'
```

### Isolation (what it is / isn't)
- Streamed → truncated (`_search`) 6/6; non-streamed → clean (`tool_search`) 6/6, on two independent GB10 boxes.
- **Parser-independent:** truncates with both `qwen3_coder` and `hermes` tool-call parsers.
- **Speculative-independent:** still truncates with `--speculative` removed (MTP off).
- **Name-specific:** only `tool_`-prefixed names truncate; `findManyCompanies` / `create_document` / `triage_inbox` / `get_weather` all stream clean.

Appears related to the streaming leak-marker fixes in #204 / #205 / #206 — the sanitizer that strips stray `<tool_call>` markup from streamed content seems to also consume the `tool` prefix of a legitimate tool **name** when it's split across SSE deltas.

### Workaround
Consumers can restore the name client-side (a `_<x>` fragment whose `tool_<x>` is a real tool is unambiguous), but the correct fix is server-side so bridge tools named `tool_*` stream intact.

rsafier:
Fixes #222.

## Problem
When **streaming** with tool_choice auto, the streamed tool-call **name** drops its leading `tool` (exactly 4 chars) for any name that collides with the `<tool_call>` markup:

| requested | streamed (before) |
|---|---|
| `tool_search` | `_search` |
| `tool_describe` | `_describe` |
| `tool_call` | `_call` |

Non-streaming returns the correct full name. Parser-independent (repro'd with both `qwen3_coder` and `hermes`), name-specific (`findManyCompanies`/`create_document`/`get_weather` stream fine), streaming-only.

## Root cause
The bare role-literal strip in `handle_token_inner` (`crates/spark-server/src/api/chat_stream/handle_token.rs`), added for the #204/#205/#206 streaming leak-marker suppression, clears any content delta that trims to exactly `"user"` / `"assistant"` / `"tool"` (to drop hallucinated bare role words). Qwen's byte-level BPE tokenizer emits a **standalone `tool` token** as the leading fragment of any `tool_*`-prefixed name. That delta is cleared **before** it reaches the streaming tool-call detector, which then reassembles the name from the remaining `_search` fragment — so `ToolCallStart` carries the truncated name. Exactly `len("tool") == 4` chars, hence the symptom. Confirmed empirically: replaying the pipeline with the old logic emits `_search` for input `tool_search`; with the fix it emits `tool_search`.

## Fix
Surgical, 2 files:
- `tool_parser/streaming_impl.rs`: add `StreamingToolDetector::inside_tool_call()` accessor (is the detector mid `<tool_call>` body).
- `api/chat_stream/handle_token.rs`: extract the strip into `strip_bare_role_literal(delta, inside_tool_call)` that **early-returns when `inside_tool_call` is true**; the call site passes `state.detector.as_ref().is_some_and(|d| d.inside_tool_call())`. Outside a tool call the leak strip is unchanged; tool **arguments** and non-`tool_*` names are never touched.

## Tests
7 new tests (`role_literal_strip_tests` in `handle_token.rs`):
- `tool_search` / `tool_call` / `tool_describe` split right after `tool` stream intact (through the real detector + real strip in `handle_token` order).
- `get_weather` control still streams intact.
- Bare role words (`user`/`assistant`/`tool`) still stripped in genuine content — protects the #204/#205/#206 leak suppression.
- Boundary cases (preserved inside a tool call, ordinary content untouched).

`cargo test -p spark-server`: **530 passed, 0 failed**, no regressions in the sanitizer/leak-marker suite (`api/tests/sanitizer.rs`).

## Repro (from the issue)
```python
tools = [{"type":"function","function":{"name":"tool_search","description":"Search for tools",
    "parameters":{"type":"object","properties":{"query":{"type":"string"}}}}}]
# stream=True + tool_choice auto → before: '_search' ; after: 'tool_search'
```

### DIFF
--- a/crates/spark-server/src/api/chat_stream/handle_token.rs
+++ b/crates/spark-server/src/api/chat_stream/handle_token.rs
@@ -42,6 +42,31 @@ type SseVec = Vec<Result<Event, std::convert::Infallible>>;
 /// wasted decode at ~10s @ 30 tok/s.
 const MAX_SUPPRESS_STREAK_TOKENS: u32 = 256;
 
+/// Drop a delta that is nothing but a bare role literal (`user` /
+/// `assistant` / `tool`) — a Qwen3.5/3.6 hallucination leak, companion to
+/// the scheduler-side `<|im_start|>` hard-stop.
+///
+/// `inside_tool_call` MUST be true whenever the streaming tool-call
+/// detector is mid-body (between `<tool_call>` and its close). Issue #222:
+/// for a `tool_*`-prefixed tool name (`tool_search`, `tool_call`,
+/// `tool_describe`) the byte-level BPE tokenizer emits a standalone `tool`
+/// token as the leading fragment of the NAME. Without the guard this strip
+/// clears that fragment, and the detector reassembles the name from the
+/// remainder (`_search`), truncating the streamed tool-call name by exactly
+/// `len("tool") == 4` chars. Non-streaming was unaffected because it parses
+/// the whole buffer at once. The guard confines the strip to genuine
+/// content leaks (no tool call in flight).
+pub(super) fn strip_bare_role_literal(delta: &mut String, inside_tool_call: bool) {
+    if inside_tool_call {
+        return;
+    }
+    let trimmed = delta.trim();
+    if delta.len() < 20 && matches!(trimmed, "user" | "assistant" | "tool") {
+        tracing::debug!("role-literal strip: dropped bare '{trimmed}' delta");
+        delta.clear();
+    }
+}
+
 /// Process one token. Returns the SSE events to forward to the
 /// client (empty `Vec` is valid).
 ///
@@ -368,13 +393,15 @@ fn handle_token_inner(state: &mut StreamState, ctx: &StreamCtx, tok: u32) -> Sse
     }
 
     // Bare role-literal leak (Qwen3.5/3.6) — companion to the
-    // scheduler-side <|im_start|> hard-stop.
+    // scheduler-side <|im_start|> hard-stop. Suppressed mid tool-call
+    // body: there a standalone `tool` token is the leading BPE fragment
+    // of a `tool_*` NAME (issue #222) being reassembled, not a role leak.
     {
-        let trimmed = delta.trim();
-        if delta.len() < 20 && matches!(trimmed, "user" | "assistant" | "tool") {
-            tracing::debug!("role-literal strip: dropped bare '{trimmed}' delta");
-            delta.clear();
-        }
+        let inside_tool_call = state
+            .detector
+            .as_ref()
+            .is_some_and(|d| d.inside_tool_call());
+        strip_bare_role_literal(&mut delta, inside_tool_call);
     }
 
     if delta.is_empty() {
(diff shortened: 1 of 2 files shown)


tbraun96:
@rsafier verified end-to-end on a spark: built ebdd7d69, streamed forced tool calls for tool_search / tool_describe / tool_call_registry at T=0, N=5 each — 15/15 names arrive intact. same repro on a main-equivalent build truncates 15/15 (tool_call_registry -> _call_registry), so the fix hits exactly the #222 mechanism. unit suite green (530 incl. your 7 new), ci green, zero kernel surface.

two non-blocking follow-ups worth a small PR later: (1) mistral [TOOL_CALLS] and the bare <function= fallback never set inside_tag, so tool_* names still truncate on those paths — widening inside_tool_call() to 'detector holding an unterminated call' would cover them; (2) after a guard force-ends mid tool-call (stop_string_triggered), the accessor freezes true and the strip stays off for cancel-latency tokens — a !stop_string_triggered at the call site tightens it.

merging. thanks for the fast turnaround on #222.

PR #225, opened 2026-07-02
  fix(dflash): implement free_state — free ctx_hidden_acc (~241 MiB/req measured) and reclaim drafter KV blocks

rrstesiak:
`BlockDiffusionDraftHead::free_state` was still the Phase-1 no-op stub, while `alloc_state` allocates a per-sequence ctx accumulator of `max_seq_len × 5 × target_hidden` BF16. `DevicePtr` has no `Drop`, so every finished request under `--dflash` leaked that buffer for the server's lifetime:

- at the in-code comment's example `--max-seq-len 16384`: ~320 MB / request
- at the current CLI default `--max-seq-len 32768`: ~640 MB / request

A long-lived `--dflash` server OOMs after a handful of requests.

## Changes

- Thread `gpu: &dyn GpuBackend` through `DraftProposer::free_state` (symmetric with `alloc_state`); the only call site (`free_sequence`, `trait_impl/sequence.rs`) already has `self.gpu` in scope. `MtpHead`, `MultiModuleMtpHead`, and `DeepseekV4MtpHead` take the parameter unused.
- `BlockDiffusionDraftHead::free_state` now:
  - frees `ctx_hidden_acc` (null-guarded, so a double call is a no-op),
  - reclaims `block_table` blocks via `kv_cache.free_blocks` — the table is empty on current main (`propose()` doesn't yet allocate paged drafter blocks), but this closes the leak class for the paged drafter path when it lands, mirroring `MtpHead::free_state`,
  - resets `seq_len` / `ctx_len` / `prefill_done` / `last_num_drafted`.

## Verification

Measured via system `free -m` — `nvidia-smi` memory queries return `[N/A]` on GB10 unified memory, and CUDA allocations don't appear in process RSS there, so system-wide used memory is the reproducible signal on this hardware.

Setup: Qwen3.6-27B-NVFP4 target + z-lab/Qwen3.6-27B-DFlash drafter, `--dflash --dflash-gamma 3 --max-seq-len 4096 --gpu-memory-utilization 0.55 --max-batch-size 1`, 5 sequential identical chat completions (300 tokens, temperature 0), identical launch for both builds.

Raw readings (fix branch tested first, then `main` baseline on the same box):

```
AFTER FIX
free -m | awk 'NR==2{print strftime("%H:%M:%S"), $3" MiB used, "$7" MiB available"}'
09:50:47 74204 MiB used, 50405 MiB available
AFTER 5 VOLVO TEST PROMPT RUNS:
09:54:14 74174 MiB used, 50435 MiB available


BEFORE FIX
09:59:23 74348 MiB used, 50262 MiB available
AFTER 5 VOLVO TEST PROMPT RUNS:
10:03:18 75554 MiB used, 49055 MiB available
```

Summary:

| Build | Before | After 5 requests | Retained |
|---|---|---|---|
| `main` @ 355c6e4 (stub) | 74,348 MiB | 75,554 MiB | **+1,206 MiB (~241 MiB/request), never freed** |
| this branch | 74,204 MiB | 74,174 MiB | **flat** (−30 MiB idle jitter) |

At the default `--max-seq-len 32768` the per-request retention is ~8× larger.

Root cause was identified during the June 10 review of `fix/dflash-verify-argmax-gate` (the argmax-gate half of that branch now lands via #218); this is the leak fix re-cut against current main as requested.

### DIFF
--- a/crates/spark-model/src/layers/deepseek_v4_mtp.rs
+++ b/crates/spark-model/src/layers/deepseek_v4_mtp.rs
@@ -553,7 +553,7 @@ impl DraftProposer for DeepseekV4MtpHead {
         Ok(())
     }
 
-    fn free_state(&self, state: &mut dyn ProposerState) -> Result<()> {
+    fn free_state(&self, _gpu: &dyn GpuBackend, state: &mut dyn ProposerState) -> Result<()> {
         let v4_state = state
             .as_any_mut()
             .downcast_mut::<DeepseekV4MtpProposerState>()
--- a/crates/spark-model/src/layers/dflash_head.rs
+++ b/crates/spark-model/src/layers/dflash_head.rs
@@ -264,7 +264,7 @@ mod propose;
 impl DraftProposer for BlockDiffusionDraftHead {
     fn alloc_state(&self, gpu: &dyn GpuBackend) -> Result<Box<dyn ProposerState>> {
         // Per-seq ctx accumulator: `[max_seq_len, 5 * target_hidden] BF16`.
-        // Sized once, re-used across the seq's lifetime; reset on
+        // Sized once, re-used across the seq's lifetime; freed in
         // `free_state`. At max_seq_len=16384 and 5×2048 BF16: 320 MB per
         // seq — tolerable on a single Spark with max_batch_size=1; for
         // higher batch we may want to reduce to a smaller working window.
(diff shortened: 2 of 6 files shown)


tbraun96:
@rrstesiak verified on a spark, and the A/B is about as decisive as leak tests get: built your head (9d5d031a) and the pre-fix dflash head side by side, same config (27B NVFP4, 16k ctx, util 0.75, --dflash + z-lab drafter), fired 12 short requests at each. your build: all 12 complete, memory delta 5 MB total. pre-fix build: died at request 9 — cuMemAlloc failed with 905 MB free — after leaking ~5 GB in 8 requests (~600 MB/req at 16k ctx, worse than your 241 MiB figure at 4k since ctx_hidden_acc scales with max_seq_len; worth noting in the code comment). fmt and clippy -D warnings green locally since the fork CI hadn't run. merging — thanks for the fast re-cut from #132.

PR #227, opened 2026-07-02
feat(spark-runtime): CUTLASS Sm120 + FlashInfer prefill wrappers (opt-in, split from #203)

rsafier:
Extracted from #203 per review: the CUTLASS (cute/cutlass headers) + FlashInfer wrappers that #203 deliberately no longer carries, so #203 stays a clean pure-Rust bring-up. This PR is the **exact inverse** of #203's removal commit — it re-adds the whole layer as an **opt-in, default-off** feature.

## What's here
- `spark-runtime/cuda/{cutlass_bf16,cutlass_nvfp4,cutlass_nvfp4_grouped}_gemm.cu`, `flashinfer_ragged_prefill.cu`
- `spark-runtime/src/{cutlass/,flashinfer.rs}` wrappers + metal stubs
- `build.rs` `CUTLASS_HOME`/`FLASHINFER_HOME` nvcc wiring + `atlas_cutlass`/`atlas_flashinfer` cfgs
- `spark-model` dispatch flags (`ATLAS_CUTLASS_*`, `ATLAS_FLASHINFER_PREFILL`, `ATLAS_HOLO_MOE_GROUPED_CUTLASS`) + the CUTLASS/FI dispatch branches (MoE-grouped, attention qkv/o prefill, SSM prefill)

## Default-off by construction
Wrappers `bail!` unless `CUTLASS_HOME`/`FLASHINFER_HOME` is set at build (`atlas_cutlass`/`atlas_flashinfer` cfg); every call site is `if <ATLAS_CUTLASS_* env> { cutlass } else { <pure-Rust path> }`, flags off unless exported. **Proven neutral:** with the split applied, `unsloth`/`nvidia` **Qwen3.6-27B-NVFP4 greedy output is bit-identical to `main`** (229/229 tokens, mean/max KL = 0.000000) — removing/re-adding this layer does not change default behavior.

Split so the pure-Rust-vs-CUTLASS policy call can be made on its own.

### DIFF
--- a/crates/spark-model/examples/fp4_mma_microproof.rs
+++ b/crates/spark-model/examples/fp4_mma_microproof.rs
@@ -1,0 +1,238 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! PHASE-1 MMA PROOF: hand-rolled Sm120 block-scaled FP4 MMA vs the CUTLASS
+//! collective.
+//!
+//! The simplest possible single-tile-looped [M,N,K] GEMM built on a hand-written
+//!   mma.sync.aligned.kind::mxf4nvf4.block_scale.scale_vec::4X.m16n8k64.row.col.f32.e2m1.e2m1.f32.ue4m3
+//! (kernel module `fp4_mma_microtest`, kernels `fp4_microtest_pack` +
+//! `fp4_microtest_mma`). NO cp.async, NO grouping, NO gate+up fusion.
+//!
+//! Oracle: spark_runtime::cutlass::nvfp4_gemm_bf16_act_weight_t — the SAME Sm120
+//! block-scaled FP4 math. MUST agree cos >= 0.999 at M=64, N=1024, K=2048.
+//!
+//! Build (remote GB10):
+//!   cargo build --release -p spark-model --example fp4_mma_microproof \
+//!     --no-default-features --features "cuda gpu-examples"
+//! Run: target/release/examples/fp4_mma_microproof
+
+use anyhow::{Result, bail};
+use spark_runtime::cuda_backend::AtlasCudaBackend;
+use spark_runtime::gpu::{DevicePtr, GpuBackend};
+use spark_runtime::kernel_args::{KernelLaunch, div_ceil};
+
+// ───────────────────────── deterministic PRNG ─────────────────────────
+struct Rng(u64);
+impl Rng {
+    fn next_u64(&mut self) -> u64 {
+        self.0 = self.0.wrapping_add(0x9E37_79B9_7F4A_7C15);
+        let mut z = self.0;
+        z = (z ^ (z >> 30)).wrapping_mul(0xBF58_476D_1CE4_E5B9);
+        z = (z ^ (z >> 27)).wrapping_mul(0x94D0_49BB_1331_11EB);
+        z ^ (z >> 31)
+    }
+    fn unit(&mut self) -> f32 {
+        ((self.next_u64() >> 40) as f32) / ((1u64 << 24) as f32)
+    }
+    fn uniform(&mut self, lo: f32, hi: f32) -> f32 {
+        lo + (hi - lo) * self.unit()
+    }
+}
+
+fn f32_to_bf16_bits(f: f32) -> u16 {
+    let bits = f.to_bits();
+    if (bits & 0x7FFF_FFFF) > 0x7F80_0000 {
+        return ((bits >> 16) | 0x0040) as u16;
+    }
+    let rounding_bias = 0x7FFF + ((bits >> 16) & 1);
+    (bits.wrapping_add(rounding_bias) >> 16) as u16
+}
+fn bf16_bits_to_f32(b: u16) -> f32 {
+    f32::from_bits((b as u32) << 16)
+}
+
+fn u16s_to_le(v: &[u16]) -> Vec<u8> {
+    v.iter().flat_map(|x| x.to_le_bytes()).collect()
+}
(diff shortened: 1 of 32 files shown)


tbraun96:
@rsafier read the whole thing — the effective diff is the 32-file revert (github's 223-file count against main is just the holo-gb10 base; worth saying so in the description since it scared the triage).

verified the claims hold: default-off is real at both layers. build.rs only compiles the wrappers (and only then emits atlas_cutlass/atlas_flashinfer) when CUTLASS_HOME/FLASHINFER_HOME are set, CUDA feature on, not skip-build; without them every entry is a bail with a clear message. every dispatch site is a new else-if above the untouched default branch, the moe SFB build is load-time flag-gated, and I found zero default-path changes — consistent with your 229/229 KL=0. no vendored headers, no blobs, no new cargo deps, CI is fully green. structurally this is the cleanest form the policy exception could take.

that said it IS the policy exception — repo is deliberately no-external-kernel-deps, so merge is an owner call (@tbraun96), and it sequences with the #203 stack anyway. not something I can green-light on correctness alone.

four concrete asks while that's decided:
1. flashinfer_ragged_prefill.cu has no SPDX line and mirrors flashinfer's csrc/batch_prefill.cu (apache-2.0) with no attribution — add the SPDX header + an apache notice for the derived parts. (also tells us the SPDX CI check skips .cu — we'll fix that side.)
2. nothing pins CUTLASS/FlashInfer revs. drop the exact tested commits into the wrapper headers or the PR description at minimum — anyone else's checkout is untested codegen, and your wrapper already depends on flashinfer's pinned CCCL ordering.
3. the two weight caches in dispatch_proj.rs key by raw device pointer and never invalidate — fine for one-model-per-process, but leave a comment saying so.
4. grouped_cutlass_gate_up_enabled + ATLAS_HOLO_MOE_GROUPED_DOWN re-read env per call — cache them like the siblings in dispatch_helpers, and the per-layer D2H+sync in moe_grouped_*_cutlass snapshots tables that never change after load.

we'll run the default-build KL/coherence/bench gates plus an opt-in flag sweep on dgx2 (with recorded revs) so the policy decision has numbers attached. nice split — this is exactly what was asked for out of #203.

SeedSource:

### Triage
- Rationale: New FP4 MMA microproof example and test for Sm120 block-scaled FP4 GEMM.
- Serving-path files: `crates/spark-model/src/layers/moe/forward_prefill_routed.rs, crates/spark-model/src/layers/moe/helpers_a.rs, crates/spark-model/src/layers/moe/init.rs, crates/spark-model/src/layers/moe/mod.rs, crates/spark-model/src/layers/ops/dispatch_helpers.rs, crates/spark-model/src/layers/ops/dispatch_proj.rs`

### Serve gate
- Build: 32s - Qwen3.6-35B-A3B-FP8 on GB10 node1
- Top-k KL vs main: **0.0704** - **large behavioral shift** - decode delta **+0.1%**

### GLM-5.2 code review
**Correctness concerns:**

1. **`f32_to_bf16_bits` NaN handling is wrong.** The NaN branch returns `(bits >> 16) | 0x0040`, but `bits >> 16` already contains the sign+exponent. ORing `0x0040` sets bit 6, but a proper bf16 NaN should have the exponent all-ones (`0x7F80` → `0x7F80`>>16 = `0x7F80`... actually `0x7F80_0000 >> 16 = 0x7F80`). The mask `0x0040` doesn't produce a canonical NaN payload. Minor for this test since inputs are finite, but the function is incorrect in general.

2. **`bf16_bits_to_f32` truncates rather than extends.** `(b as u32) << 16` is correct for bf16→f32 zero-extension, but the cosine comparison is done in f64 after f32 conversion — acceptable, but note bf16 has only 8-bit mantissa; cos≥0.999 is a loose bar that may pass even with packing bugs.

3. **`grouped_cutlass_gate_up_enabled()` is called per-invocation** with no caching; env var reads on every forward pass are cheap but unidiomatic. More critically, the gate-up and down paths check **different env vars** (`ATLAS_HOLO_MOE_GROUPED_CUTLASS` vs `+ ATLAS_HOLO_MOE_GROUPED_DOWN`), meaning the down path can fall through to the FP4 branch while gate_up uses CUTLASS — **mixed code paths** that may produce layout/scale mismatches if the two paths pack scales differently.

4. **`.expect()` after `.is_some()` check** is safe but defeats the point of the guard; if the Option is `None` the earlier branch is skipped silently with no fallback log.

**Risk:** The CUTLASS grouped path is opt-in via env var, limiting blast radius. However, the **non-orthogonal env flags** (gate_up vs down) create a configuration footgun where partial enablement yields undefined correctness behavior with no validation.

**Quality:** Comments are thorough but the diff is truncated (`helpers_a.rs` cut mid-function). Cannot fully review `pack_weight_sfb` wiring or the `ops::moe_grouped_*_cutlass` signatures. The example is a useful microproof but hardcodes M=64,N=1024,K=2048 with no argument parsing.

**Recommendation:** Block until (1) full diff is available for `helpers_a.rs` and the `ops` functions, (2) add validation that `GROUPED_DOWN` requires `GROUPED_CUTLASS`, and (3) fix the bf16 NaN conversion or document it as test-only.

rsafier:
Absorbed by #229 — closing.

Verified against current `main`: this PR's CUTLASS Sm120 + FlashInfer prefill wrapper files (`crates/spark-runtime/src/cutlass/*`, `flashinfer.rs`, the `cutlass_*_gemm.cu`) show **0 additions** vs main — all folded in. The only nonzero residual is `weight_loader/qwen35/load_layers.rs` (+21), which is ancestral Holo-stack content the branch carries, not this PR's own contribution. No outstanding CUTLASS/FlashInfer delta.

PR #228, opened 2026-07-02
feat(sampling): default temp/top_k/top_p from generation_config.json (split from #210)

rsafier:
Split out of #210 (native FP8) per review. Core chat sampling (temp/top_k/top_p) now defaults to the model's shipped `generation_config.json` (`state.default_*`) instead of the hand-curated `MODEL.toml [sampling]` presets — matching vLLM and the model author's config, and mirroring what `completions.rs` already does.

Carved into its own PR because it is a **global behavior change** for every served model, including the 27B flagship whose BFCL numbers were measured under the presets, and it overlaps #162's vLLM-parity generation-control work. It should get its own **BFCL A/B** before landing — decided here rather than smuggled into an FP8-loader PR.

(The selected MODEL.toml preset still drives the penalties, which `generation_config.json` doesn't define.)

### DIFF
--- a/crates/spark-server/src/api/chat/sampling_setup.rs
+++ b/crates/spark-server/src/api/chat/sampling_setup.rs
@@ -85,20 +85,29 @@ pub(super) fn build_sampling(
         .map(|v| v == "1" || v.eq_ignore_ascii_case("true"))
         .unwrap_or(false);
 
+    // Core sampling (temp/top_k/top_p) defaults to the model's shipped
+    // generation_config.json (state.default_*) — matching vLLM and the model
+    // author's recommended config — instead of the hand-curated MODEL.toml
+    // [sampling] presets. The lower preset temps (e.g. Holo thinking/tools =
+    // 0.6 vs generation_config 1.0) made the model over-commit to tool calls
+    // vs vLLM: lower temperature is more deterministic, so it picks "call the
+    // tool" far more consistently. The selected preset still drives the
+    // penalties below (generation_config doesn't define those); min_p and
+    // top_n_sigma already default to generation_config.
     let temperature = if force_temp_zero {
         0.0
     } else {
-        req.temperature.unwrap_or(preset.temperature)
+        req.temperature.unwrap_or(state.default_temperature)
     };
     let top_k = if force_temp_zero {
         0
     } else {
-        req.top_k.unwrap_or(preset.top_k)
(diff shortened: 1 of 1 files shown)


SeedSource:

### Triage
- Rationale: Changes sampling defaults to generation_config values, affecting model output behavior and tool call determinism.
- Serving-path files: `crates/spark-server/src/api/chat/sampling_setup.rs`

### Serve gate
- Top-k KL vs main: **0.0002** (no shift vs main) - decode delta **-0.1%**

### GLM-5.2 code review
**Correctness:** Logic is sound, but `state.default_*` fields are unvalidated. If `generation_config.json` is missing or fails to parse, these likely default to `0`/`0.0` — meaning `temperature=0` and `top_k=0` silently force greedy decoding instead of falling back to `preset`. Verify the `state` initialization guarantees sane defaults (e.g., `1.0`/`-1`/`1.0`) when the config is absent.

**Risk:** High. This silently changes default sampling behavior for *all* models. The comment claims lower temps cause "over-commit to tool calls," but this is an unvalidated hypothesis. No A/B test or benchmark data is referenced. Changing defaults system-wide based on one model's anecdotal behavior is risky. Also, `force_temp_zero` still overrides correctly, which is good.

**Quality:** Comment is too long and editorialized for a code comment — move rationale to PR description or commit message. The `preset` struct is now only used for penalties; confirm it isn't partially dead code or that this doesn't break the preset selection contract elsewhere.

**Recommendation:** Block until: (1) `state.default_*` fallback values are confirmed safe, (2) regression tests cover missing `generation_config.json`, (3) comment trimmed to factual summary.

rsafier:
The generation_config.json sampling-defaults feature from this PR was absorbed into main during the #229 merge (sampling_setup.rs / completions.rs now default temp/top_k/top_p from state.default_*). The one thing that did NOT land is the guard the SeedSource/GLM-5.2 review made a merge condition — safe (non-greedy) fallback + regression tests for a missing/unparseable generation_config.json. #267 adds exactly that on top of current main: the absent-config fallback now sources from the model's curated [sampling.non_thinking] preset (single source of truth, non-greedy by construction) instead of magic 0.6/20/0.95 constants, plus 4 regression tests (missing→preset, present-overrides, partial-per-field, present temperature=0 honored). fmt/clippy -D/569 tests green. This branch itself predates #229 (its diff vs main would undo #192/#229), so closing as absorbed — the review guard is delivered in #267.

Issue #15, opened 2026-05-07
[bug] KV always recomputing with Qwen3.6


jxlarrea:
### Atlas version / docker tag

avarok/atlas-gb10

### Model

Qwen/Qwen3.6-35B-A3B-FP8

### Hardware

DGX Spark

### Launch command

```bash
services:
  atlas:
    image: avarok/atlas-gb10:latest
    container_name: atlas
    network_mode: host
    ipc: host
    pull_policy: always
    restart: unless-stopped
    volumes:
      - /opt/models/hf:/root/.cache/huggingface
    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              count: all
              capabilities: [gpu]
    command:
      - serve
      - Qwen/Qwen3.6-35B-A3B-FP8
      - --port=8888
      - --host=0.0.0.0
      - --max-seq-len=65536
      - --kv-cache-dtype=fp8
      - --kv-high-precision-layers=auto
      - --gpu-memory-utilization=0.90
      - --scheduling-policy=slai
      - --mtp-quantization=fp8
      - --tool-call-parser=qwen3_coder
      - --enable-prefix-caching
      - --speculative
```

### Expected behaviour

KV cache should work.

### Actual behaviour

KV is always recomputing. Testing it using your exact docker quickstart example:

```
atlas  | 2026-05-06T23:55:15.193246Z  WARN spark::main_modules::serve_router: Atlas is listening on 0.0.0.0:8888 — reachable from any host on the network. If this machine is on a shared LAN or has a public IP, pass --bind 127.0.0.1 (or set --require-auth and a real firewall) before accepting traffic.
atlas  | 2026-05-06T23:55:15.193253Z  INFO spark::main_modules::serve_router: Listening on [0.0.0.0:8888](http://0.0.0.0:8888/)
atlas  | 2026-05-06T23:55:15.193346Z  INFO spark::scheduler: Scheduler started (batched mode, max_batch=8, mtp=true, ngram=false, num_drafts=1, policy=fifo, chunked_prefill=true, max_prefill_tokens=8192)
atlas  | 2026-05-06T23:55:15.208731Z  INFO spark::scheduler: Swap space: 3 GB at /tmp/atlas-swap/
[...]

PR #229, opened 2026-07-02
perf(qwen3.6-27b): NVFP4 W4A4 MMQ FFN + FlashInfer-GDN + Marconi tail checkpoints — cold prefill 1.24-1.75x over llama.cpp, warm multi-turn TTFT beats llama like-for-like

tbraun96:
## Summary

Stacks three prefill levers for the NVIDIA mixed-precision `Qwen3.6-27B-NVFP4` on GB10 (on top of the merged #223):

1. **`ATLAS_FFN_NVFP4_MMQ`** — vendored llama.cpp Blackwell NVFP4 W4A4 MMQ (`mma.sync.kind::mxf4nvf4.block_scale.scale_vec::4X.m16n8k64`, hardware QMMA.SF) as the dense-FFN gate/up prefill GEMM: 80 TFLOP/s vs t_m128 ~51 (1.57x on the bucket). Weights are repacked at load by a **raw bit shuffle** (checkpoint e2m1 codes + e4m3 scale bytes reused verbatim — empirically the MMA decodes ue4m3 as standard e4m3 on both operands, so only the per-tensor fp32 `scale2` needs folding, done in the scaled SiLU-mul). Gate/up `_t` copies freed at load → net ~0 footprint. Down-projection arm included as a separate A/B gate (`ATLAS_FFN_NVFP4_MMQ_DOWN`, measured neutral, default off).
2. **FlashInfer-GDN AOT bridge** (merge of #220) — GDN prefill kernels.
3. **Marconi tail SSM checkpoints** — splits the final prefill chunk once at the last-block boundary and saves a tail snapshot there regardless of interval; removes the issue-#15 auto-clamp that forced 256-token warm chunks. Warm multi-turn restores now hit with a ≤16-token replay at full-size (8192) prefill chunks; single-shot requests pay nothing (gated on a read-only prefix-cache peek).

## Measured (GB10, solo-GPU, identical prompts/probes, `cache_prompt`-busted llama)

Cold prefill (tok/s):

| ctx | llama.cpp Q4_K_M | Atlas (this stack) | × |
|---|---|---|---|
| 2k | 621–686 | **1088–1094** | 1.6–1.75 |
| 4k | 700–718 | **1077** | 1.5 |
| 8k | 723–742 | **1057** | 1.4 |
| 16k | 699–716 | **982–1000** (16.3s) | 1.37–1.4 |
| 29k | 732 | **911** (31.8s) | 1.24 |

- vLLM's prior 32k reference (~44s, ~745–796 tok/s) is beaten at 29k (31.8s, 911 tok/s).
- Warm multi-turn (22-turn chained driver, ~1.35k-token suffixes, like-for-like): llama median 2401ms / Atlas **2253ms** (was 2462 before the tail-checkpoint fix); interval no longer affects warm TTFT.
- Cumulative vs pre-#223 baseline: 16k cold prefill 28.56s → 16.3s (**1.75×**).

## Correctness

- MMQ same-quant rel_err 1.6e-3 vs CPU ref (= bf16-output rounding); scale semantics resolved by a 4-hypothesis microbench.
- Coherence 10/10 on every configuration measured; cold curves regression-checked at each step.
- Flag-gated throughout: `ATLAS_FFN_NVFP4_MMQ` default-off; tail checkpoints active only when prefix caching + snapshots are enabled.

## Notes

- Known residual: turn 2 of a fresh conversation full-recomputes once (cold turn 1 saves no tail checkpoint by design).
- Recommend a BFCL-ST run to confirm accuracy neutrality of the MMQ gate/up arm before flipping any default.
- Contains two near-duplicate commits from parallel sessions (57686798 / a5b42f84) — squash-merge recommended.


tbraun96:
Defaults update (661d1813): the NVFP4-MMQ FFN arms (gate/up + fused down) are now DEFAULT ON with opt-out kill-switches (ATLAS_NO_FFN_NVFP4_MMQ / _DOWN). Safe by construction — the arms are also gated on kernel handles that only exist on the qwen3.6-27b/nvfp4 target; every other model is byte-identical.

Out-of-the-box before/after (zero env vars): 2k 830→927 / 4k 831→909 / 8k 818→900 / 16k 779→854 tok/s (+10-12%). Equivalence proof: the default-on binary is bit-identical to the env-flag configuration — greedy text 10/10 prompts, mean KL 0.00000000 over 119 logprob positions, coherence green.

Also recorded for reviewers: ATLAS_DECODE_OPT (single-warp GEMV variants) measures a 2x decode regression on GB10 — keep off; decode GEMV bandwidth work tracked separately (audit shows silu_input + single-GEMV families at 57-60% of peak DRAM vs the dual GEMV's 89%).

rsafier:
## Independent e2e validation ✅ — reproduces the cold-prefill table

Tested the full stack on an idle GB10 (gx10) in the CUDA-13.2 FlashInfer-GDN container, driving the `nvidia/Qwen3.6-27B-NVFP4` checkpoint. Built `perf/qwen36-27b-prefill-nvfp4` at `661d1813` (the MMQ default-on commit) and ran the mounted binary inside the FI-GDN image (`ATLAS_GDN_FLASHINFER=1`, `--enable-prefix-caching --ssm-cache-slots 16 --ssm-checkpoint-interval 256`).

**All three levers confirmed engaged (from logs, not assumed):**
- FI-GDN: `ATLAS_GDN_FLASHINFER: FlashInfer GDN kernel loaded (opt-in)`
- MMQ FFN: `nvfp4_mmq used` + `q4k_mmq used` — **with zero env vars** (default-on works)
- Marconi: `SSM snapshot pool: Marconi 16 slots` + checkpoints every 4096 tokens

**Cold prefill (default-on binary, unique cache-miss prompts):**

| ctx | this stack (measured) | PR table | vs Atlas baseline* |
|---|---|---|---|
| 2k  | **1095** t/s | 1088–1094 | +31% |
| 4k  | **1087** | 1077 | — |
| 8k  | **1046** | 1057 | +28% |
| 16k | **994**  | 982–1000 | +27% |
| 29k | **918**  | 911 | — |

*\*baseline = same binary, MMQ off + FLA GDN scan (834/815/783 @ 2k/8k/16k), measured bare-metal.*

Numbers land right on the PR table. FI-GDN is the dominant cold-prefill mover (~+25%); the MMQ FFN arm in isolation is +2–6% e2e (bare-metal A/B 834→851/866/822) — expected, since the GDN scan, not the FFN gate/up, is the prefill bottleneck.

**Marconi warm multi-turn** — chained conversation, growing context:

| turn | prompt_tok | cached_tok | new_tok | TTFT |
|---|---|---|---|---|
| 1 (cold) | 1404 | 0 | 1404 | 1328ms |
| 3 (warm) | 2826 | 2112 | 714 | 1296ms |
| 6 (warm) | 4959 | 4240 | 719 | 1325ms |

Cached prefix grows every turn; only ~720 new tokens prefill/turn, so warm TTFT stays flat ~1.3s while the conversation grows to ~5k tokens — ~4× vs a cold turn-6 (~5.5s), compounding with length.

**Commit 661d1813 (MMQ default-on):** MMQ + DOWN arms engage with no configuration; measured +2–3% over the env-flag config (the DOWN arm). Coherence held across runs (one empty-JSON at greedy T=0 was a thinking-mode artifact also present in the pre-MMQ baseline — consistent with the bit-identical KL=0 equivalence in the commit).

**One caveat (same as #220):** the FI-GDN ~+25% is a **single-stream** result — that's where its gate fires. Under the production co-dispatch prefill path the FI-GDN gate is bypassed and it falls back to the FLA scan, so this cold-prefill win is a single-stream / low-concurrency number.

tbraun96:
Decode GEMV family audit closed (evidence chain): pure-stream DRAM probe measures the GB10 wall at 233 GB/s; L2-defeated per-shape benches put every w4a16_gemv shape (GDN in-proj 16384x5120, attn-o/GDN-out 5120x6144, lm_head 248072x5120) at 222-226 GB/s = 96-97% of that wall — the family is at the hardware roofline. The one real defect found (compute-bound fused silu_input, SM 57%/mem 23%) is fixed by the split-staging commit (9d9fa53b, +2.2% decode, KL 0.00129). Regression on the shipped binary: prefill 927/852 tok/s @2k/16k, decode 14.09 tok/s, coherence green. Remaining decode headroom is byte-reduction (MTP already +49%; FP4-KV at long context) and the small non-GEMV tail — not GEMV tuning.

tbraun96:
Heads-up on f4d47965 (rode into this branch because the regression blocks this PR's ST merge gate): the full-995 BFCL ST run on current main scored 55.78 overall vs our 89.04 reference from 06-15 — same harness, same box. Root-caused to four stacked defects in the API-tools path and fixed here:

1. **Schema types**: BFCL/agent schemas send Python-style types (`"type":"dict"`, `float`, `tuple`). xgrammar treats unknown types as unconstrained, silently dropping `required` enforcement. The grammar sanitizer now normalizes them to JSON Schema.
2. **Validation drop**: blocking mode attached only fully-valid tool calls — a call missing one trailing param became an *empty response* (160/995 samples). Validation now has typed severity: missing-param calls are attached as produced (client schema gives the model actionable feedback); phantom names, unparseable args, and the F78/CWE-78 guards still reject. The streaming path's `contains("non-empty")` sniff is replaced by the same enum.
3. **Dead trigger**: the hermes auto-mode grammar trigger `<tool_call>{"name":"` never matched the model's template-taught `<tool_call>\n{"name": ` — constrained decoding had literally never engaged in auto mode. Trigger is now the tag itself and the structural-tag begin/end match the model's natural format (the forced compact form caused OOD junk trailing calls).
4. **Mask generator bug** (in-tree xgrammar port): the speculative fast-accept path updated `prev_token`, poisoning the LCP-reuse invariant — fused tokens crossing rule boundaries (e.g. the ` "` space-quote BPE token) and their trie subtrees were wrongly mask-rejected, producing deterministic garbage inside string values. One-line fix to upstream semantics + mask↔parser equivalence tests (816+2 pass).

Gates so far: 20-sample parallel battery went from 5/20 empty + 10 incomplete to 20/20 complete correct calls (temp 0, MTP on, batch 1). 332-sample harness subset: 84.04 overall with the collapse epicenter healed (parallel 18.55→80.00, parallel_multiple 28.23→83.33); the java/js subsets remain at their documented model-characteristic levels. Full-995 run is in flight as the formal gate — will post the score here.

SeedSource:
## Independent accuracy gate — MMQ FFN arm, on vs off

Ran our on-vs-off accuracy gate against this PR's head (`4816a51`) to give @azeezish an independent read on the NVFP4 W4A4 MMQ FFN arm. **Verdict: accuracy-neutral — confirms your `rel_err 1.6e-3` / KL≈0 claim.** The BFCL-ST regression you root-caused is a *different* surface (the API-tools/grammar path fixed in `f4d47965`), not this kernel.

**Setup:** `nvidia/Qwen3.6-27B-NVFP4`, single GB10, `ATLAS_TARGET_MODEL=qwen3.6-27b`, greedy (temp 0), 6 prompts (reasoning / factual / arith / code / finance / instruction-format), 64 tok each, top-20 logprobs.
**A/B:** ON = default; OFF = `ATLAS_NO_FFN_NVFP4_MMQ=1` + `ATLAS_NO_FFN_NVFP4_MMQ_DOWN=1` (both arms off → t_m128 baseline). Confirmed from the serve logs that OFF took the baseline path (no FP4-MMQ dispatch line) and ON engaged it.
**KL note:** computed only over the *identical-prefix* region (up to the first greedy-token fork) — after a fork the two contexts differ and per-position KL is meaningless.

| prompt | len | diverge@ | preKL mean | preKL max | off tok/s | on tok/s |
|---|---|---|---|---|---|---|
| reason | 86 | same | 0.00085 | 0.01607 | 13.95 | 14.08 |
| factual | 75 | same | 0.00142 | 0.04196 | 13.96 | 14.07 |
| arith | 66 | 19 | 0.00017 | 0.00314 | 13.96 | 14.01 |
| code | 81 | 63 | 0.00127 | 0.02511 | 13.96 | 14.09 |
| finance | 76 | 42 | 0.00120 | 0.01434 | 13.94 | 14.03 |
| instruct | 75 | same | 0.00216 | 0.02916 | 13.93 | 14.04 |

**identical-prefix KL: mean 0.0013 / max 0.042 nats** over 360 matched positions. Decode +0.7% (noise — these short prompts don't exercise the prefill path where MMQ wins).

3/6 prompts fork, but every divergence is **cosmetic with an identical underlying answer** — exactly the low-entropy branch-point token-flip you'd expect from FP4-MMA vs t_m128 rounding, never a correctness change:

- **arith** — OFF `91, 12` / ON `91 and 12` (both `13*7=91`, `144/12=12`)
- **code** — OFF `sum(i**2 for i in range(1, 11))` / ON the same expression in a ` ```python ` fence
- **finance** — OFF `-$18,750 (a loss)` / ON `-$18,750.` (same correct figure)

Coherence 6/6 on both configs.

**Takeaway:** the MMQ gate/up (+down) arm does not move the output distribution in any way that changes correctness — KL is at the FP4-rounding floor and all forks are surface phrasing. If it'd help @azeezish, we can run a small function-calling probe (schema-typed tool defs, missing-param, phantom-name) against `f4d47965` to independently confirm the tool-calling fix restores toward the 89.04 ref — just say the word.

*Gate + harness run by SeedSource (Claude Code). AI-generated, attributed.*

SeedSource:

### Triage
- Rationale: CI, license, and 3rd-party patch files; no model or quantization code changes.
- Serving-path files: `3rdparty_patches/gdn_aot/delta_rule_sm120_aot_export.patch, 3rdparty_patches/gdn_aot/gdn_dump_meta.py, 3rdparty_patches/gdn_aot/gdn_export.py, 3rdparty_patches/gdn_aot/gdn_harness.cpp, 3rdparty_patches/gdn_aot/gdn_harness_packed.cpp, 3rdparty_patches/gdn_aot/gdn_holo.so`

### Serve gate
- Serve-test indicated, but no gate data this cycle (conflicting/out-of-scope or not yet run).

### GLM-5.2 code review
**Review: Empty Diff**

The diff is empty — no code changes provided. Cannot review correctness, risk, or quality without the actual implementation.

**Action Required:** Re-submit with the full diff. Given the PR scope (NVFP4 quantization kernels, FlashInfer GDN integration, checkpoint changes), I need to verify:

1. **NVFP4 MMX kernels**: dequant correctness, scale factor handling, accumulation precision
2. **FlashInfer-GDN**: API compatibility, memory ownership, stream synchronization
3. **Checkpoint changes**: tensor layout/version compatibility, migration paths, load-time validation
4. **Benchmark claims**: reproducibility, warm-up conditions, hardware specs

An empty diff with performance claims is a red flag. Resubmit with actual changes.

tbraun96:
### 35B-A3B-FP8 agentic `webserver_ok` results (dgx2, image built from this branch)

Ran the flagship `Qwen3.6-35B-A3B-FP8` against the opencode agentic Axum ping/pong
suite (`webserver_ok` — model writes a Rust project, builds it, runs the server,
curls it, tears down), 10 iterations, thinking-on, on dgx2.

**Result: 10/10 `webserver_ok` + 10/10 `followed_directions`, Σwall 1649.7s.**

Per-iter wall (s): `103, 214, 109, 235, 115, 223, 125, 224, 168, 134`.

Three root causes had to be fixed to get there:

1. **Correctness (`webserver_ok` 0 → 10/10)** — the default here must be the
   `qwen3_coder` tool-call parser. `--tool-call-parser hermes` mangles newlines in
   multi-line file-write args (`[package]\nname` → `[package]nname`, #37), so the
   generated `Cargo.toml`/`main.rs` never compile. Dropping the hermes parser alone
   took `webserver_ok` from 0 to 10/10.
2. **Wall** — `--enable-prefix-caching` is the master switch that gates BOTH KV reuse
   (attention) AND Marconi SSM-snapshot restore (the 30 GDN/Mamba layers) on this
   hybrid. Without it every agentic turn re-prefills the whole growing conversation
   O(n²). Adding it + `--ssm-cache-slots 256 --ssm-checkpoint-interval 16` took wall
   from ~2800s to ~1650s.
3. **Tail iters** — NVFP4 lm_head collapses on this model and produced ~360s
   timeout-pinned iterations; `--lm-head-dtype bf16` fixes it (measured 10/10).

**fp8 KV cache is a decode regression on SM121/GB10**, not a win: an fp8-KV A/B
pinned every iteration at the 360s `OC_TIMEOUT` ceiling — decode collapsed to
~17–31 tok/s at 13k–29k context (MTP accept stayed healthy at 84–87%, so it's the
fp8-KV dequant path, not spec-decode). Keep KV at bf16 here; the fp8-KV decode
kernel is a separate deficiency worth owning.

Remaining gap to the 1500s target is the **bimodal per-iter wall**: iters 2/4/6/8
run ~224s while 1/3/5/7/9/10 run ~125s (iter 1 is the *fastest*, so it's not
cold-build — looks like cross-task prefix/SSM-snapshot thrash). Bringing the slow
iters down to the ~125s the fast ones already prove projects Σ≈1250s. Investigating
next.

Follow-up commit `85de087b` (pushed to this branch) also restores the missing
`COPY vendor/ vendor/` in 9 gb10 nvfp4 Dockerfiles that have been failing to build
since cudarc was vendored, and documents the proven serve config in the 35B header.

PR #232, opened 2026-07-02
fix(dflash): K=16/17 verify conv correctness + review fix-list (carries #218)

tbraun96:
carries @Sujimoshi's #218 head (a58b14e2) plus the fixes from its review, implemented in-house to keep momentum: (1) BLOCKER — K=17/K=16 fused WY verify now runs conv1d_update_l2norm for ALL K tokens (wy17/wy16 read all K conv_out rows; conv_state now ends after the final token; intermediates snapshots unchanged at K-1); (2) concat_rows enforces weight_scale_2 equality (ensure!, fail-fast); (3) dflash_stochastic_accept D2H-failure fallback now rejects instead of argmaxing a zeroed buffer; (4) reverted the ungated every-16 do_norm cadence change in common gated_delta_rule.cu + the strix/metal signature drift; (5) per-propose/PAIR-DUMP/K4-timing logs demoted to debug; (6) decode_b.rs back under the 500-LoC cap; (7) session docs + bench scripts dropped from scope.

gates pending on a spark: no-spec KL identity vs main, then --dflash T=0 greedy-identity in default AND flags configs (the branch previously diverged at char 366 in both — the conv fix addresses the identified cause; if divergence persists there is a second shared verify bug to bisect before this merges).

Co-authored with @Sujimoshi — their three root-cause fixes (BF16-as-FP32 conv routing, YaRN-on-null drafter, EAGLE conditioning) are the substance here.


tbraun96:

PR #233, opened 2026-07-02
fix(tools): codex namespace tools + phantom-name guard (carries #171)

tbraun96:
carries @0137's #171 (codex Responses-API namespace tools + normalize_tool_name consolidation) plus the guard from review: bare-identifier candidates whose normalized name still contains ':' are rejected (prose like json:{"a":1} / tool_call:{...} no longer parses as a phantom call named 'json:'; tool_call:{"name":"get_weather",...} falls through to the fallback parser and extracts get_weather). tests pin all three shapes; author's six format tests stay green; 551 total green.

gates pending on a spark: tool-call probes + coherence, then merge (their branch already proved logit-identical serve path).


tbraun96:
@0137 carried your branch forward with the ':' guard from the review as this PR — your six format tests stay green and the phantom shapes are pinned (json:{...} -> no call, tool_call:{"name":...} -> extracted call). review welcome if the guard placement matches your intent; gates are running on a spark now and we'll land it credited to you.

tbraun96:

PR #234, opened 2026-07-02
fix(stream): tool-markup leak fixes + minimax envelope preservation (carries #206)

tbraun96:
carries @rsafier's #206 (stop_string_triggered raw-emit + flush tail scrub — both real leaks with a 13/24→0/36 live repro) rebased onto main, plus the blocker fix from review: scrub_tool_tags is removed from sanitize_content_chunk (the buffered state machine already guards the streaming path) and gated on envelope_open-empty at the flush/stop-arm sites, so the F73 minimax envelope pass-through survives. the disabled sanitizer regression-test module is re-enabled (it would have caught this — sanitizer_envelope_open_disables_orphan_suppression now runs and passes). sanitizer.rs and the test file are split under the 500-LoC cap. 555 tests green.

gates pending on a spark: streamed tool-call + markup-leak probes, coherence.


tbraun96:
@rsafier took your #206 the last mile as this PR: rebased onto main (composes with your #224), dropped the in-chunk scrub in favor of envelope-gated scrubs at flush/stop-arm only, re-enabled the disabled sanitizer test module (it catches exactly the F73 case), and split the files under the 500-cap. your two leak fixes are intact and credited. flag anything you disagree with — especially the choice to remove the scrub from sanitize_content_chunk entirely rather than gate it.

tbraun96:

PR #235, opened 2026-07-02
fix(qwen3.6): FP8 block-scaled prefill default-on + agentic fixes, rebased + hardened (carries #177)

tbraun96:
carries @mission-deny-the-mission's #177 rebuilt as a clean series on current main: (1) block-scaled FP8 W8A8 prefill DEFAULT-ON across the six dispatch sites (opt-out ATLAS_FP8_SINGLE_SCALE=1) — their B1-drift root-cause; (2) anthropic thinking-object-without-budget deferral; (3) tool-coerce second pass. the EOS-guard hunk was dropped (landed via #183 with identical semantics). hardening from review: is_block_scaled() guard on all six force_w8a8 sites + scale_format assert in attention set_fp8_weights (a per-row FP8 weight can no longer be misread as [N/128,K/128] blocks), and the env flag is cached in a OnceLock instead of re-read per dispatch.

gates pending on a spark: KL + coherence on Qwen3.6-27B-FP8 (the default-flip surface) A/B vs main, NVFP4 flagship no-regression KL, then merge.

### DIFF
--- a/bench/tool_format_eval/GROUND_TRUTH_RESULT.md
+++ b/bench/tool_format_eval/GROUND_TRUTH_RESULT.md
@@ -1,0 +1,41 @@
+# BF16 ground-truth result: the forward pass is NOT the bug
+
+Method: downloaded the original BF16 base `Qwen/Qwen3.6-35B-A3B` (full precision,
+same weights the FP8 checkpoint quantizes), ran a forward (layer-truncated to 16
+layers to fit unified memory — layer N depends only on 0..N, so L0..15 are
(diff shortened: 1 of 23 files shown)


tbraun96:
@mission-deny-the-mission your #177 is carried forward as this PR, rebuilt on current main: FP8 block-scaled prefill default-on (your B1-drift root-cause) plus the scale_format hardening from review; your EOS-guard hunk was dropped only because #183 landed the identical fix first. FP8 A/B + NVFP4 no-regression gates run on a spark next. if you still have the agentic workload that exposed the degeneration, running it against this branch would be the best possible validation.

tbraun96:

PR #236, opened 2026-07-03
fix(loader): pin NVFP4 variant-detection layer prefix (EP worker mis-detection)

rsafier:
## Problem

`detect_nvfp4_variant()` built its second fallback prefix from `config.local_expert_range().0` — an **expert** index — and used it as a **layer** index (clamped to `num_hidden_layers-1`):

```rust
format!("model.language_model.layers.{}", config.local_expert_range().0.min(config.num_hidden_layers.saturating_sub(1)))
```

On a single GPU / rank 0, `local_expert_range().0 == 0`, so this resolved to `layers.0` and everything worked. But under **expert parallelism** the worker's local-expert start is non-zero (e.g. `128` for `world_size=2`, 256 experts), which clamped to **layer 39** — a full-attention layer whose FP8 `self_attn.q_proj.weight` trips the FP8 sniff. So `rank>0` mis-detected a **MIXED_PRECISION** (NVFP4 MoE + FP8 attention) checkpoint as `Fp8Dequanted` while rank 0 correctly detected `Standard`.

The worker then routed the NVFP4/`U8` `shared_expert` through the FP8 dequant path and hit:

```
Failed to build model
  Caused by: Expected FP8E4M3 for ...layers.0.mlp.shared_expert.gate_proj.weight, got UInt8
```

Rank 1's death took down the rank-0 collective (`ncclBroadcast` → `cuMemsetD8Async` status 700).

## Fix

Pin the fallback to a fixed `layers.0` prefix so detection is rank-independent. **Byte-identical** for rank 0 / `world_size=1` (they already resolved to layer 0) → **zero single-node regression**; only EP workers change, and only to agree with rank 0. It's at the variant-detection level, upstream of every projection, so it covers all `shared_expert` weights (gate/up/down + scales), not just the one that crashed first.

Pre-existing since the OSS base commit (`e5d088f`) — surfaces only under multi-node EP, which nothing had exercised for a MIXED_PRECISION checkpoint until now.

## Validation

Holo-3.1-35B-A3B-NVFP4 across a **2-node EP setup** (dgx-00 + gx10, `world_size=2`): the worker now builds the `shared_expert` (`moe_shared_expert_fused_fp8_t used`) and the pair serves **distributed inference with correct output** (`2+2=4`, both ranks) — where it previously crashed at model build. Single-node (rank 0) and `world_size=1` behavior unchanged.

_Root-caused via an 8-agent review workflow; all three adversarial verifiers confirmed correct / no rank-0-or-single-node regression / complete coverage._


rsafier:
Recut onto post-#229 `main` as a single clean commit.

## Confirmed still needed
The bug is **still present on `main`** — `weight_map/nvfp4_detect.rs` builds its second fallback prefix from `local_expert_range().0` (an EXPERT index) used as a LAYER index (`.min(num_hidden_layers-1)`). Not absorbed by #229.

## What changed vs the old branch
- Replayed only the one-line fix (`local_expert_range().0.min(...)` → fixed `"model.language_model.layers.0"`) fresh onto current `main`, since the old branch carried a `Merge branch 'main'` commit plus the now-obsolete CI-allowlist commit (main already allow-lists those files). Final diff: **1 file, the `nvfp4_detect.rs` fallback prefix + its explanatory comment.**

## Correctness
Detection reads *checkpoint structure*, which is identical on every rank — so a rank-dependent value (`local_expert_range`) was inherently wrong for it. Pinning to a rank-independent constant (layer 0) is the fix. Byte-identical for rank 0 / `world_size=1` (they already resolved to layer 0), so zero single-node regression; only EP workers change, to agree with rank 0. This clears the 2-node EP `Expected FP8E4M3 ... got UInt8` shared_expert crash.

## Validation (dgx-00)
- `cargo check -p spark-model` compiles; `cargo fmt --check` clean; `cargo clippy -p spark-model -- -D warnings` clean; `cargo test -p spark-model --lib` → 79 passed.
- Note: `detect_nvfp4_variant` has no unit tests (pre-existing gap); this fix is byte-identical on the single-node path and was e2e-validated on the 2-node EP setup per the PR description. A rank-independence unit test would need a `ModelConfig` + mock `WeightStore` harness — flagged as a follow-up.

Commit signed. Still marked draft — ready for review whenever.

rsafier:
Closing: #418 landed the correct half of this (thanks for the credit line) with a fuller fix than my recut — the layer-probe rewrite supersedes the +14/−7 here. The EP=2 crash it was cut for is covered.

PR #241, opened 2026-07-03
feat(gdn-tp): GDN HeadParallel — tensor-parallel for SSM-hybrid (pure-TP2 Holo-35B)

rsafier:
## GDN HeadParallel — tensor-parallel for the SSM-hybrid (qwen3_5) family

Unlocks **pure tensor-parallel** (`--tp-size N`) for Gated-DeltaNet / linear-attention SSM-hybrid models (Holo-3.1, etc.), which were previously `TP=1`-only. Validated on **Holo-3.1-35B-A3B-NVFP4, 2-node pure-TP2 over RDMA** (gx10 + dgx-00): correct, coherent, and concurrency-scaling.

Stacked on `feat/native-fp8-attn-modelopt-mixed` (needs its native-FP8-attn path). The MoE-ep-gate commit here (`fix(moe): gate EP all-reduce on ep_world_size`) is the same as standalone PR #240 and dedupes when that lands.

### What's in it (7 commits)

- **GDN HeadParallel MVP** — config-localizes the linear head counts per rank, adds segmented weight slicers (`tp_shard.rs`: QKVZ sliced *per-segment* Q\|K\|V\|Z independently, interleaved BA group-aligned, conv/a_log/dt_bias per-value-head, out_proj row-parallel) + the post-out_proj all-reduce. 10/10 pure-math slicer unit tests.
- **Multi-rank command protocol** — `multi_rank_protocol_active()` (`ep_world_size>1 || tp_world_size>1`) so the head↔worker command broadcast fires under pure-TP (not just EP); prior code gated on `ep_world_size>1` and silently no-op'd, deadlocking the worker. + flag-gated CUDA-graph decode.
- **MoE ep-gate** — all-reduce gated on `ep_world_size` not `world_size` (else replicated experts double under pure-TP). = PR #240.
- **norm-replicate** — GDN gated-RMSNorm weight is `[vd]` (value head-dim, shared across heads); replicate under HeadParallel instead of slicing on the head axis (that read past the buffer → load-time `cuMemcpyDtoDAsync INVALID_VALUE`). Adds permanent `tracing::debug!` to the shard helpers.
- **slot-sort + conv input-stride fix** — prerequisites for correct batched-recurrent decode (contiguous SSM-slot ordering + per-seq conv stride; without them the batched path silently falls back or corrupts n≥2).

### Results (Holo-3.1-35B-A3B-NVFP4, 2-node pure-TP2)

- **Correct + coherent**, golden TP=1-vs-TP=2 parity within one FP8/NVFP4 quantum (bit-exact token-identity is unachievable on a heavily-quantized model; the within-quantum logprob check passes).
- **Single-stream** 55.1 tok/s; with the batch-1 CUDA graph, ~71 tok/s (+28%).
- **Concurrency** (batched-recurrent + multi-seq graphs, requires `ATLAS_EP_PROTOCOL=v2`): C=4 **+27%** (72.6→92), C=8 **1.79×** (127 tok/s).
- **Soak**: 315 s / 8-client mixed — 144 req, **0 errors, 0 degenerate**, stable throughput, 0 restarts.

### Runtime config (both ranks)

`ATLAS_HOLO_NATIVE_FP8_ATTN=1 ATLAS_SSM_BATCHED_RECURRENT=1 ATLAS_DECODE_GRAPHS_MULTISEQ=1 ATLAS_EP_PROTOCOL=v2` + `--fp8-kv-calibration-tokens 256`. Single-node (`tp=1`) is byte-identical (all slicers/gates short-circuit). A BF16-dequant-attention TP path (quality-fallback to native-FP8) is implemented separately on `feat/bf16-dequant-attn-tp`.


SeedSource:

### Triage
- Rationale: EP world size check refactored; workflow file-size cap updated with new SSM and MOE paths.
- Serving-path files: `crates/spark-model/src/layers/moe/forward.rs, crates/spark-model/src/layers/moe/forward_atomic_c4.rs, crates/spark-model/src/layers/moe/forward_batched.rs, crates/spark-model/src/layers/moe/forward_k2.rs, crates/spark-model/src/layers/moe/forward_k3.rs, crates/spark-model/src/layers/moe/forward_prefill.rs`

### Serve gate

### GLM-5.2 code review
**Correctness:** The core change — gating MoE all-reduces on `ep_world_size > 1` instead of `world_size > 1` — is correct and necessary for pure-TP (non-EP) configs; without it, TP2 would spuriously all-reduce MoE outputs, corrupting results. The SSM `ssm_tp_all_reduce` insertions are placed correctly (before residual add). The conv input-stride fix (per-seq loop with pre-offset pointers) is a legitimate correctness fix for `n>=2`.

**Risk:** High. This is a broad behavioral change across all MoE forward paths (decode, prefill, batched, k2/k3, token-major, fp8, bf16). If `ep_world_size` is unset/defaults to 0 for EP configs, the `> 1` check silently disables all-reduce → silent numerical corruption. Verify `ep_world_size` is always initialized. The conv fix adds a per-seq launch loop (n kernel launches vs 1) — performance regression for large batches; acceptable as a correctness fix but needs a follow-up.

**Quality:** File-size-cap whitelist additions are technical debt; acceptable with tracked follow-ups. The conv fix comment is thorough. Missing: no test coverage visible for the TP2-no-EP path or the `n>=2` conv stride fix.

rsafier:
Superseded by #254 — closing.

This PR was stacked on #221 (the pre-#229 chain). Now that #229 is merged to `main`, I re-cut the same GDN HeadParallel-TP work directly onto `main` via `git rebase --onto` → **#254** (base `main`). #254 additionally carries the required moe-ep-gate fix (main still has the buggy `world_size()` gate), fixes a broken intra-doc link, and is validated with a 35B pure-TP2 2-node benchmark (C1 70.9 / C4 88.1 / C8 106.9 tok/s, TTFT 181 ms). CI green. Continuing there.

PR #242, opened 2026-07-03
fix(mtp+grammar): depth-aware MTP gate re-measurement, per-position bitmask cache, wider mask pre-warm

### LINKED ISSUE #237: perf: constrained decode ~5 tok/s with qwen3_coder grammar engaged (post-#231) — suspect double bitmask fill in grammar_blocks_stop()
observed on a spark during a BFCL subset run on merged main (b87bd483+): with the qwen3_coder tool grammar engaged, decode runs at ~5.2-5.3 tok/s vs ~78 tok/s unconstrained — ~180ms/token of per-token overhead. outputs are correct (finish=stop, sane lengths); this is pure throughput.

prime suspect: #231 added grammar_blocks_stop() (crates/spark-server/src/grammar/state.rs) as the SSOT stop-legality predicate for the EOS gates in decode_logits_step.rs and emit_step.rs, with a lazy fill_next_token_bitmask — if the scheduler also fills the bitmask for constrained sampling in the same step, the 248k-vocab mask is computed twice per token. fix shape: reuse the step's existing bitmask for the stop-legality check (or cache the fill per (matcher, position)).

for reference the june grammar-on ST runs averaged ~12.8s/sample end-to-end; current subset leg runs ~66s/sample at ~90-token responses. repro: serve 27B NVFP4 with tools armed (auto parser -> qwen3_coder grammar), send any tool-calling request, watch the Done: line tok/s.

tbraun96:
three engine fixes from the flagship webserver_ok wall investigation (Qwen3.6-35B-A3B-FP8, GB10):

1. **depth-aware MTP gate.** the gate latched a permanent decision off whichever request arrived first — but the verify/decode multiplier is depth-dependent (weight-bound m≈4 at short ctx on fine-grained MoE vs KV/SSM-bound m≈1.1 deep; measured both). a short warm-up probe permanently disabled MTP for whole agentic sessions. the gate now records measured_at_depth, re-opens measurement on factor-4 regime crossings (both directions, floor 512), and the scheduler consults the decision per step instead of killing use_mtp. measurement steps emit real tokens, so re-measuring is free. 10 unit tests incl. the disable->deep->re-enable scenario.

2. **grammar bitmask per-position cache** (fixes #237's redundant-fill half). stop_legal/grammar_blocks_stop re-filled the 248K-vocab mask already computed for sampling at the same matcher position (~+13ms/verify step). cache invalidated at accept_token/rollback/reset — the only matcher-advancing sites — preserving the one-real-fill-per-token NPDA-sync invariant.

3. **mask pre-warm widened 8 -> 512 states** so the decode loop's resident states (preamble/trigger-tracking + JSON scanners) are cached instead of JIT-computed.

live gates on a spark (35B FP8, -mtp recipe config): grammar correctness intact (coherence 10/10, 15/15 tool probes), webserver_ok 6/6 with Σwall 1657s vs the Jul-1 release image's 0/6 @ Σ2161 (all-ceiling). tools-armed decode at ~6-8k depth: 25.3 tok/s (== no-tools) at temp-0; the remaining agentic decode gap is the temp>0 CPU sampling path (D2H + CPU mask/penalties per token) — tracked as the follow-up in #237.

clippy -D warnings clean; fmt clean.


tbraun96:
pushed d9dd1638 with the verify-step deep-dive results. headline: the suspected jun→jul verify regression DOES NOT EXIST — same-depth A/B vs the jul-1 release image gives m=2.16 vs 2.17. the verify−decode delta is a constant ~+56ms (2nd verify token's MoE expert re-read + per-token dispatch), depth-independent, while decode grows with depth — so the depth-aware gate in this PR is behaving exactly right: DISABLED at ~9k (m=2.16, correctly net-negative), ENABLED live at ~37k (m=1.63, 68-75% acceptance, net-positive).

full phase breakdown (ATLAS_MTP_TIMING=1, now permanently available): fwd=98ms of 104 total; pipeline/D2H/dequant/state-mgmt all ~1ms or less. also in this push: forced_token no longer double-fills the 248K mask (reuses the cached per-position bitmask, exact by construction), and both greedy fast paths accept reduce-only penalties via a provable argmax-survival bound (new fast_greedy SSOT module, unit-tested). 557/557 bin tests, coherence 10/10, tool probes 15/15 on the 35B flagship.

next lever (separate PR): grouped K=2 MoE dispatch in spark-model — dedupe shared/overlapping expert reads across the 2 verify tokens (forward_batched.rs dispatches per token today). that attacks the +56ms floor directly and would extend MTP's win into the 6-25k agentic band.

PR #250, opened 2026-07-04
feat(loader): native-BF16 dense loader for no-metadata Holo/Ornith checkpoints

rsafier:
## What

Adds a **native-BF16** load path for *no-metadata dense* Qwen3.5-family checkpoints (Hcompany Holo-3.1-0.8B/4B/9B, Ornith dense). These ship plain BF16 `.weight` tensors with no quantization metadata, so today the loader hardwires them through the runtime BF16→NVFP4 requant path (`Nvfp4Variant::Bf16Raw`) — an **uncalibrated, lossy** quantization that degrades small-model quality (the BF16-source models exist precisely because no calibrated NVFP4/FP8 release does).

This keeps FFN, attention (Q/K/V/O), and the SSM `in_proj_qkvz`/`out_proj` in native BF16, dispatching the `dense_gemv_bf16`/`dense_gemm_bf16` kernels already in the standard bundle (same path Gemma-4 dense uses). **Loader-only; no new kernels.**

- `qwen35_dense.rs`: BF16 arms for FFN (snapshot overlay), attention (BF16 dense q/k/v + `set_o_dense_bf16`), SSM (BF16 `in_proj_qkvz` + `out_proj_dense`), all gated on `Bf16Raw`. NVFP4 weights are still built + kept as the spec-decode/batched fallback (a null NVFP4 weight under w4a16 dispatch is the CUDA-700-at-concurrency footgun).
- `ssm_batched.rs`: extend the `qkvz_ok` gate so the pure-BF16 build engages the batched `dense_gemm` decode path instead of the per-seq loop.

## Use-after-free fix (would-be CUDA-700)

The naive FFN overlay installs gate/up/down via `dense_auto()`, which returns the `WeightStore` cached BF16 pointer **un-copied**. But `load_dense_ffn`'s Bf16Raw arm runtime-quantizes to NVFP4 first and `gpu.free()`s that same buffer (`quantized_any`, `nvfp4_detect.rs`), and the store never evicts — so the overlay would hand `set_bf16_weights` freed memory → `dense_gemm_tc` illegal access on the first prefill. Fixed by snapshotting fresh D2D copies **before** `load_dense_ffn` frees the originals.

## Recommended serve flag

Pair with `--lm-head-dtype bf16` to keep the 248K-vocab LM head lossless too (else it defaults to runtime-NVFP4).

## Validation

Built + served **Holo-3.1-4B** on GB10 from this branch (cherry-picked onto current `main`, post-#229):
- Compiles clean; loads (`dense weight loader: 32 layers`, `LM head kept as BF16`).
- Coherent on short and 1.8K-token multi-chunk prefill; **zero CUDA-700**.
- Coexists with #229 MMQ-finalize and #223 transposed-NVFP4 prefill (both variant-gated off on the BF16 path).
- Agentic multi-turn (3-task): 2/3, on par with the runtime-NVFP4-head baseline (the `convert` task is a flaky exact-value reasoning check; not a loader signal).

## Notes

- Does **not** need an `ATLAS_TARGET_QUANT=bf16` build (that strips the w4a16 stubs the dense loader still constructs) — use the standard bundle.
- Full Holo *serving* config (e.g. `--target-kv-tokens`) lives in the separate Holo-GB10 stack; this PR is just the loader capability, which `main` already routes via the `qwen35_dense` dispatch.

---


rsafier:
Synced with main. One semantic merge in `ssm_batched.rs`: main's tile-GEMM wide-batch arm (`tc_wide_ok`, n>16) and this PR's native-BF16 arm (`dense_gemm_k` when both `qkvz_nvfp4`/`qkvz_fp8w` are absent) both extend `qkvz_ok` — the resolution unions them, so the FP8 sub-case stays byte-identical to main and the BF16 build engages the batched fast path instead of silently dropping to the per-seq loop. fmt/clippy/tests green locally; +166/−4 across 2 files vs main.

rsafier:
Follow-up push (`30a937cf`): #284's one genuine residual is folded in here — the dense loader's LinearAttention GDN TP wiring (`TpGdnDims::from_config`, full-dim load/concat/interleave, the tp>1 six-tuple shard block), which pairs naturally with this PR's two files (#254 landed the `tp_shard` machinery it calls; at tp==1 the wiring is a byte-identical pass-through, so the single-GPU story here is unperturbed). Also applied from that review: `debug_assert!(o_dense_bf16.is_some())` after the null-o_proj construction, `debug_assert_ne!` in `clone_bf16`, comment cites for the dense_gemm_k and ffn_bf16_snapshot questions, and a `segment_copy_plan` unit test at real Holo-0.8B dims. One flag for review: as in #284 itself, the native-FP8/native-NVFP4 GDN fast paths are not TP-sharded — under tp>1 those formats load full-size (the wiring covers Bf16Raw + legacy dequant/requant, faithful to #284's scope). fmt/clippy/tests green (314, incl. the new plan test).

tbraun96:
Landed in main via the #440 stack merge (`3c941751`). Verified by content diff — `weight_loader/qwen35_dense.rs`, `qwen3_ssm/trait_decode_multi_seq/ssm_batched.rs` and `tp_shard/tests.rs` all byte-identical (`proj_is_native_fp8`, `dense_fp8_enabled`). Thanks @rsafier.

PR #251, opened 2026-07-04
test(bench): --no-thinking / --thinking-budget toggles for bench_isl_osl

rsafier:
## What

Adds `--no-thinking` and `--thinking-budget N` to `bench/bench_isl_osl.py`. Both inject `chat_template_kwargs` (`enable_thinking`, `thinking_budget`) into each request so streaming TTFT/TPOT can be measured with reasoning off or at a fixed budget.

## Why

On thinking-default models (e.g. Holo, `max_thinking_budget=768`) reasoning tokens dominate TTFT, which swamps the prefill signal you actually want to measure. These toggles let the bench isolate prefill/decode from reasoning overhead.

Tooling-only; no engine changes.

---

### DIFF
--- a/bench/bench_isl_osl.py
+++ b/bench/bench_isl_osl.py
@@ -61,13 +61,22 @@ def percentile(data: list, p: float) -> float:
     return s[f] + (k - f) * (s[c] - s[f])
 
 
-def send_request(url: str, model: str, prompt: str, max_tokens: int) -> dict:
+def send_request(url: str, model: str, prompt: str, max_tokens: int,
+                 enable_thinking: bool = True, thinking_budget=None) -> dict:
+    # Thinking control: with thinking ON, TTFT-to-first-content includes the
+    # reasoning phase (up to the model's thinking budget) — that's
+    # time-to-first-ANSWER, not prefill. --no-thinking measures pure prefill
+    # TTFT. --thinking-budget caps the reasoning length.
+    ctk = {"enable_thinking": enable_thinking}
+    if thinking_budget is not None:
+        ctk["thinking_budget"] = thinking_budget
     body = json.dumps({
         "model": model,
         "messages": [{"role": "user", "content": prompt}],
         "max_tokens": max_tokens,
         "temperature": 0.0,
         "stream": True,
+        "chat_template_kwargs": ctk,
         # Authoritative output-token accounting. The server reports the
         # exact sampled-token count in usage.completion_tokens and (with
         # return_token_ids) the IDs on each chunk. Counting SSE chunks or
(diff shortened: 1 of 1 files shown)


SeedSource:

### Triage
- Risk: **low** - serve-test-required (triage advisory): `False` - path-heuristic (authority): `False`
- Rationale: Benchmark script update for thinking control parameters; no production or quantization changes.

### Serve gate
- **Skipped** - no serving-path files (docs/CLI/test/example). Zero GPU.

### GLM-5.2 code review
**Correctness:** Sound. `enable_thinking` defaults to `True`, preserving prior behavior. `chat_template_kwargs` is the correct vLLM field for thinking controls. Warmup correctly inherits the toggles, ensuring measured runs aren't invalidated by a thinking-on warmup followed by thinking-off measurement.

**Risk:** Low. Benchmark-only script, no engine code touched. One concern: `--no-thinking` behavior depends on the server's chat template actually honoring `enable_thinking` — if the deployed template ignores it, the script will silently report "pure prefill TTFT" while actually measuring time-to-first-answer. No validation or warning if the response contains reasoning tokens when `--no-thinking` is set.

**Quality issues:**
- Line 240 print statement is cramped and hard to read; split across lines.
- `thinking_budget=None` in the signature should be `Optional[int] = None` for type clarity, though this is a bench script.
- No guard against `--thinking-budget` being set without `--no-thinking` being False (i.e., budget with thinking off is contradictory) — minor, server likely ignores it.

**Verdict:** Approve. Consider adding a sanity check that `--no-thinking` responses contain no reasoning tokens.

tbraun96:
Reviewed this. The intent is right and the key detail is right — you use `chat_template_kwargs: {"enable_thinking": ...}`, not `{"thinking": ...}`, which is silently ignored. That distinction cost us a whole confound in the concurrency campaign, so thank you for getting it right.

But there's a defect that inverts the PR's own purpose, and it needs fixing before this can land.

**The key is now sent unconditionally.** Previously no `chat_template_kwargs` was sent at all; with this patch every request carries `enable_thinking: true` by default. The templates do not agree on what an *absent* key means:

| template | absent key means | this PR's default |
|---|---|---|
| `qwen3_5_moe` (the 35B A3B flagship) | thinking **OFF** | flips it **ON** |
| `openai/minimax_m2` | thinking **OFF** | flips it **ON** |
| `gemma4` | thinking **OFF** | flips it **ON** |
| `qwen3_5` | thinking ON | unchanged |
| `holo3_1_moe` | thinking ON | unchanged |

So on the model this bench is most used against, the patch turns reasoning **on** by default — inflating TTFT and making every historical `bench_isl_osl` number incomparable. That's the opposite of the stated goal ("reasoning tokens dominate TTFT, which swamps the prefill signal").

**The fix is small:** only inject `chat_template_kwargs` when the user actually asked. Default the parameter to `None` rather than `True`, build the dict only from what was supplied, and omit the key entirely otherwise. That preserves today's behaviour by default and gives `--no-thinking` / `--thinking-budget` their meaning. It's also the PCND-correct shape — no implicit default in a measurement path.

**One thing you should know before spending time on it:** the owner has directed that `bench_isl_osl.py` be ported to Rust as a registered benchmark and the Python removed. So this file is on its way out. I've carried your requirement into the port's design — `auto|off|on`, where `auto` omits the key entirely — precisely because of what this PR surfaced.

Your call on which you'd prefer:
1. Push the `None`-default fix here and I'll merge it now (the file is still live until the port lands), or
2. Close this and let the Rust port carry the capability, with credit in its commit.

Either is fine by me. Option 1 is real value in the meantime; option 2 saves you the work.

rsafier:
@tbraun96 Taking option 2 — close it and let the Rust port carry the capability, with credit. The `enable_thinking` / `chat_template_kwargs` distinction is the part worth preserving; your catch that an unconditionally-sent key flips three templates' defaults is exactly why I'd rather it be designed into the Rust port than patched into a file that's being deleted.

PR #252, opened 2026-07-04
fix(loader): native-FP8 SSM shadowed on modelopt-mixed-precision 35B

rsafier:
## What

One-line match-guard fix so the **native-FP8 SSM** path actually engages on modelopt-mixed-precision Qwen3.5 (Holo-3.1-35B-A3B / Ornith) when `ATLAS_HOLO_NATIVE_FP8_SSM=1`.

## The bug (shadowed arm)

In `qwen35/load_layers.rs` the linear-attention builder match has, in order:

```rust
_ if modelopt_mixed_precision && !fp4_proj_decode => build_linear_attention_dense_bf16(...)  // BF16-dense SSM
...
_ if !(force_nvfp4_all || fp4_proj_decode) && ssm_native_fp8 => build_linear_attention_fp8(...)  // native-FP8 SSM
```

The BF16-dense arm sits **earlier** and matches first, so when `ssm_native_fp8` is set the modelopt 35B still builds a BF16-dense `in_proj_qkvz`/`out_proj` and the native-FP8 arm is **never reached**. The native-FP8-SSM machinery is present on `main` (folded in via #229) but unreachable on this path.

Fix: add `&& !ssm_native_fp8` to the BF16-dense guard so it yields to `build_linear_attention_fp8` when native FP8 SSM is requested.

```diff
- _ if modelopt_mixed_precision && !fp4_proj_decode =>
+ _ if modelopt_mixed_precision && !fp4_proj_decode && !ssm_native_fp8 =>
```

## Validation

`ssm_native_fp8` is already bound on `main` (same function), so this is a pure guard tightening — no new state. Cherry-picks onto current `main` with zero conflicts. The equivalent change has been running in our GB10 serving of Holo-3.1-35B-A3B-FP8 with `ATLAS_HOLO_NATIVE_FP8_SSM=1` (native-FP8 SSM engaged, coherent). Behavior is unchanged when `ssm_native_fp8` is false (byte-identical guard).

---

### DIFF
--- a/crates/spark-model/src/weight_loader/qwen35/load_layers.rs
+++ b/crates/spark-model/src/weight_loader/qwen35/load_layers.rs
@@ -787,7 +787,7 @@ pub(super) fn load_layers(
                     // fp4_proj_decode drops Holo's modelopt SSM out of the BF16-
                     // dense + FP8-overlay build so it falls through to the NVFP4
                     // builder below → in_proj_qkvz/out_proj decode on w4a16_gemv.
-                    _ if modelopt_mixed_precision && !fp4_proj_decode => {
+                    _ if modelopt_mixed_precision && !fp4_proj_decode && !ssm_native_fp8 => {
                         linear_attn_arms::build_linear_attention_dense_bf16(
                             i,
                             store,


SeedSource:

### Triage
- Rationale: Weight loading logic change affecting mixed precision and SSM native FP8 paths.
- Serving-path files: `crates/spark-model/src/weight_loader/qwen35/load_layers.rs`

### Serve gate
- Top-k KL vs main: **0.0009** (no shift vs main) - decode delta **+0.3%**

### GLM-5.2 code review
**Correctness:** The fix correctly prevents the `modelopt_mixed_precision` arm from matching when `ssm_native_fp8` is true, allowing fallthrough to the NVFP4 builder. The logic is sound — without this guard, native-FP8 SSM weights would incorrectly route to the BF16 dense builder.

**Risk:** Low. The added condition is purely additive (further restricting an existing match arm). No other code paths are affected. The fallthrough behavior is already documented in the comment immediately above.

**Quality concerns:**
1. The comment above the condition is now stale — it describes only the `fp4_proj_decode` case but doesn't mention the `ssm_native_fp8` exclusion. Should be updated.
2. No test coverage visible in the diff. A 35B mixed-precision + native-FP8 SSM config is a narrow regression scenario; a unit test asserting correct builder dispatch for this combination would be valuable to prevent re-regression.
3. Variable `ssm_native_fp8` is not shown in context — verify it's in scope and initialized before this match block (likely is, given surrounding code, but worth confirming).

**Verdict:** Approve with nits — update the comment, consider adding a dispatch test.

tbraun96:
Thanks for this, and sorry to come back with a negative — I went looking to merge it and couldn't make the bug reproduce. Closing, with the evidence, in case I've misread something.

**The match has four arms, and the PR body quotes two.** The omitted first arm is the case the PR describes as broken:

```
load_layers.rs:774   _ if native_modelopt_ssm => build_linear_attention_fp8(...)
load_layers.rs:790   _ if modelopt_mixed_precision && !fp4_proj_decode => build_linear_attention_dense_bf16(...)   <- the arm this PR edits
load_layers.rs:806   Nvfp4Variant::Fp8Dequanted if ... && ssm_native_fp8 => ...
load_layers.rs:823   NVFP4 fallback
```

and `native_modelopt_ssm` (`:125-126`) is exactly `modelopt_mixed_precision && ATLAS_HOLO_NATIVE_FP8_SSM == "1"`. So on a modelopt-mixed-precision 35B with the env var set, the native-FP8 SSM builder fires **first**. Nothing shadows it.

I checked this isn't a rebase artifact — the match is byte-identical at this PR's own base blob (`c221fdd12`); the only differences are helper functions below line 848.

**Your own #260 validation is the cleanest disproof.** That branch doesn't contain this commit, yet its comment quotes `ATLAS_HOLO_NATIVE_FP8_SSM=1: routing Holo ModelOpt SSM/attention projections through native FP8` and `Layer N: SSM native FP8 — w8a16`. Those log lines (`:172-174`) only emit when the allegedly-unreachable path engages.

I think the confusion is `ssm_native_fp8` (per-layer weight detection, `:759`) versus the env var — the env var drives arm 1, not arm 3.

**What the change would actually do.** The added `!ssm_native_fp8` conjunct only bites when `modelopt_mixed_precision && env UNSET && ssm_native_fp8`, where it silently flips the default build from BF16-dense to the NVFP4 builder for someone who opted into nothing. Today that's inert, because ModelOpt per-tensor FP8 deliberately returns false from `proj_is_native_fp8` — so the behaviour space here is "no-op or undisclosed default change".

Also worth flagging for future PRs against this file: the validation run served `Holo-3.1-35B-A3B-FP8`, a compressed-tensors checkpoint where `is_holo_modelopt_mixed_precision` is false — so the edited guard is unreachable in that run, with or without the patch.

Under the merge rigor now in force, a `crates/` change owes five GPU gates (~4h19m measured). I can't spend that on a change whose only demonstrable effect is a default flip nobody asked for.

If you can produce a serve log where `ATLAS_HOLO_NATIVE_FP8_SSM=1` on a modelopt-mixed-precision checkpoint does **not** hit `:774`, reopen and I'll take another look — that would mean I've misread the guard and I'd want to know.

PR #254, opened 2026-07-04
feat(gdn-tp): GDN HeadParallel tensor-parallel (MoE pure-TP), re-cut onto main

rsafier:
## What

**GDN HeadParallel — tensor-parallel for the SSM/GDN-hybrid layers**, re-cut onto current `main` (post-#229). This is the #241 work (`feat/gdn-headparallel-tp`) rebased off its old `#221` base and onto `main` via `git rebase --onto`, so it applies cleanly to the merged tree.

Shards the GDN (gated-delta-net / linear-attn) heads + attention across ranks under pure-TP (`--tp-size N --world-size N`, no `--ep-size`), with the post-`out_proj` all-reduce, CUDA-graph-capturable GDN decode (`ATLAS_GDN_DECODE_GRAPH=1`), and the norm-replicate / slot-sort / conv-fix prerequisites.

## Includes the moe-ep-gate fix (required)

Pure-TP MoE needs `is_ep`/all-reduce gated on `ep_world_size`, not `world_size()` — otherwise fully-replicated experts get SUM-reduced 2× per layer → garbage. `main` still has the buggy `world_size()` gate (the standalone fix is **#240**, not yet merged), so this PR carries it (reconciled against #229s new bf16-batch2 path in `forward_k2.rs`). If #240 lands first, this commit drops out on rebase.

## Validation — 35B pure-TP2, 2-node (GB10 × 2)

`Holo-3.1-35B-A3B-NVFP4`, `--tp-size 2 --world-size 2` over NCCL:
- **Coherent** (2+2=4, capital-of-France=Paris), both ranks healthy, zero CUDA faults, `EP v2 active`.
- **Decode throughput** (no-thinking, 512-tok): **C1 70.9 / C4 88.1 / C8 106.9 tok/s** aggregate; single-stream **TTFT 181 ms**, ~72 tok/s.

**Config note:** under TP>1, do **not** set `ATLAS_HOLO_NATIVE_FP8_SSM=1` — native block-scaled FP8 SSM is TP=1-only (HeadParallel FP8 scale-slicing is deferred); the SSM runs NVFP4 via `ATLAS_HOLO_FP4_PROJ_DECODE=1`. The binary errors clearly if you do.

## Scope

- **MoE-TP only.** Dense models (Holo-0.8B/4B/9B) do **not** yet TP-shard — the dense loader (`qwen35_dense.rs`) needs the dense-attention TP-shard (`load_qkvo_tp`/`shard_dense_bf16`, cf. #250). Verified: dense 0.8B under TP2 = incoherent, TP1 = fine. Dense-TP is a separate follow-up.
- Also compiles + serves single-node (TP=1) unchanged (gates are `comm.is_some() && ...`).

## Related

Part of the post-#229 upstreaming set — supersedes the old stacked #241 (which was based on `#221`). Independent of #250 (native-BF16 loader), #251 (bench toggles), #252 (FP8-SSM shadow fix). Depends conceptually on #240 (moe-ep-gate), carried here until it merges.


SeedSource:

### Triage
- Rationale: Refactoring EP world size check logic; no quantization or serving changes.
- Serving-path files: `crates/spark-model/src/layers/moe/forward.rs, crates/spark-model/src/layers/moe/forward_atomic_c4.rs, crates/spark-model/src/layers/moe/forward_batched.rs, crates/spark-model/src/layers/moe/forward_k2.rs, crates/spark-model/src/layers/moe/forward_k3.rs, crates/spark-model/src/layers/moe/forward_prefill.rs`

### Serve gate
- Top-k KL vs main: **0.0010** (no shift vs main) - decode delta **-0.4%**

### GLM-5.2 code review
**Correctness — conv stride fix (critical):** The per-seq loop with `offset(i * qkvz_size)` is the right fix, but the diff is truncated at `offset(i * sel` — cannot verify the offset expression. If it uses `conv_dim` instead of `qkvz_size`, the bug persists. **Request full hunk.**

**Correctness — EP gate (`world_size()` → `ep_world_size`):** Correct for pure-TP where `comm.world_size() > 1` but `ep_world_size == 1`. All 11 MoE forward paths are consistently migrated. Verify `ep_world_size` is always populated (non-zero default) in `ctx.config`; otherwise EP runs silently degrade to no all-reduce.

**Correctness — SSM TP all-reduce:** `ssm_tp_all_reduce(out, 1, ...)` hardcodes num_tokens=1 in `ssm_forward.rs`. The comment claims it covers multi-seq, but multi-seq has its own call sites (`ssm_batched.rs`, `ssm_batched_recurrent.rs`) with correct `n`. Confirm `ssm_forward.rs` is single-token-only; if reachable from multi-seq, the hardcoded 1 is a bug.

**Risk:** The conv fix changes launch topology (n kernels vs 1) — graph capture and stream serialization must tolerate the loop. No `graph_capture` guard visible.

**Quality:** File-size-cap exemptions are tech-debt acknowledgments, acceptable. Two files >500 LoC with "follow-up" promises — track as issues, not just comments.

**Blocking:** full conv hunk + confirmation `ep_world_size` has no zero-default footgun.

PR #255, opened 2026-07-05
perf(spark-runtime): opt-in NVTX kernel-attribution + phase-timing instrumentation

Sujimoshi:
## Summary
First of a planned split of #218 (DFlash speculative decoding) into smaller, independently reviewable PRs — see the split plan discussed on #218. This PR carries only the diagnostic instrumentation slice:

- `crates/spark-runtime/src/nvtx_diag.rs` (new) — opt-in NVTX range instrumentation for kernel attribution, gated behind `ATLAS_NVTX_DIAG=1`
- `crates/spark-runtime/src/kernel_args.rs` — per-op timing/profiling hooks
- `libloading` dependency, dlopen'd lazily — normal builds/runs have **no** link-time or runtime dependency on `libnvToolsExt`

No behavior change to any existing code path — purely additive, opt-in diagnostics.

## Test plan
- [x] `cargo check -p spark-runtime` (ATLAS_SKIP_BUILD=1, CUDA stubbed) — clean
- [x] `cargo fmt --check -p spark-runtime` — clean
- [x] `cargo clippy -p spark-runtime --tests` (ATLAS_SKIP_BUILD=1) — no warnings
- [ ] Runtime smoke test with `ATLAS_NVTX_DIAG=1` on GB10 (requires live GPU, not exercised by this diff alone since it's dead code until wired up by later PRs in the split)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #256, opened 2026-07-05
perf(dense-ffn): per-op profiling breakdown for decode forward()

Sujimoshi:
## Summary
Second in a planned split of #218 (DFlash speculative decoding) into smaller, independently reviewable PRs (first was #255).

- Adds a `prof!` macro to `crates/spark-model/src/layers/dense_ffn.rs` (same shape as the existing pattern in `moe/forward.rs` / `qwen3_ssm/ssm_forward.rs`) and wraps every GPU op in `DenseFfnLayer::forward()`'s decode path with it.
- Covers all four dispatch branches: FP8 fast/slow path, BF16 path, and the NVFP4 default path (`use_sw` variant, `split_silu` branch, fused SiLU/GeLU branches).

**Note on provenance:** the original commit in #218 (`18fc6549`) targeted an older shape of `forward()`. Since then, #229 (NVFP4 W4A4 MMQ FFN) restructured the function and added the fp8/`use_sw`/`split_silu` branches. This PR re-applies the profiling breakdown against the current (post-#229) structure rather than patching mechanically — more dispatch branches means the per-op breakdown is more useful, not less, for isolating decode cost under whichever config is active.

No behavior change — `prof!` is a no-op unless `ctx.profile` is set.

## Test plan
- [x] `cargo check -p spark-model` (ATLAS_SKIP_BUILD=1, CUDA stubbed) — clean
- [x] `cargo clippy -p spark-model --tests` (ATLAS_SKIP_BUILD=1) — no warnings

### DIFF
--- a/crates/spark-model/src/layers/dense_ffn.rs
+++ b/crates/spark-model/src/layers/dense_ffn.rs
@@ -633,6 +633,21 @@ impl DenseFfnLayer {
         let h = ctx.config.hidden_size as u32;
         let inter = ctx.config.intermediate_size as u32;
 
+        let profile = ctx.profile;
+        macro_rules! prof {
+            ($label:expr, $body:expr) => {{
+                if profile {
+                    let t = std::time::Instant::now();
+                    let r = $body;
+                    ctx.gpu.synchronize(stream)?;
+                    tracing::info!("    FFN {}: {:.0}μs", $label, t.elapsed().as_micros());
+                    r
+                } else {
+                    $body
+                }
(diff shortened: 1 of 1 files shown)


Sujimoshi:
Closing — this was purely diagnostic instrumentation (Dense FFN per-op profiling breakdown) split out of #218. The original perf investigation it supported is already done and documented; no active use case for it right now. Cheap to re-add later against whatever the code looks like at that point if a real need comes up.

PR #258, opened 2026-07-05
feat(dflash): model glue/scaffolding for K=γ verify + native NVFP4 lm_head

Sujimoshi:
## Summary

Split out of #218 (DFlash speculative decoding for Qwen3.6-27B-NVFP4). This PR adds the model-level plumbing DFlash's draft/verify path needs, without any of the actual verify-path kernels/logic (those come in later PRs in the split):

- Trait default no-op hooks on `TransformerLayer`: `prefill_phase1_verify`, `prefill_gdn_wy16` — scaffolding for the K=γ verify-path PR, currently unused (`Ok(false)` always).
- Row-major K-row `dflash_hidden_save` capture buffer (`dflash_hidden_save_rows`, `try_dflash_capture_draft`/`try_dflash_capture_all`) so every accepted verify position gets its own target hidden, instead of only rows 0/1.
- `ctx_slot_positions` on `DflashProposerState` to track true sequence positions across thinking-token gaps (slot index ≠ position when reasoning tokens create a prompt/output split).
- FP32 GDN output sibling buffer (`gdn_buf_out_f32`, guarded by `ATLAS_DFLASH_VERIFY_GDN_F32`) and `ssm_verify_h_tmp` scratch — prep for a verify-path precision fix, not yet wired to any consumer.
- `dense_gemm_pipelined` kernel handle for the draft head's lm_head (tensor-core pipelined BF16 GEMM, faster than `dense_gemm` for large M).
- RoPE fix for the drafter: `config.json` has `rope_scaling: null`, but the prior code applied YaRN interpolation (factor=64) anyway, corrupting all positional encodings and causing 0% accept rate. Replaced with the standard `1/theta^(2j/dim)` formula.
- **Fix (found during GPU testing, see comment below):** the NVFP4 lm_head fast path used the drafter's own uncapped `vocab_size` for its GEMM instead of the target's tokenizer-capped `vocab_size` the weight buffer was actually quantized against, causing an out-of-bounds read / `CUDA_ERROR_ILLEGAL_ADDRESS` on checkpoints with a capped vocab.

### Conflict resolution notes

Main evolved independently since #218 branched — most notably via #229 (NVFP4 MMQ FFN + FlashInfer-GDN + Marconi tail-checkpoints). Conflicts were resolved by keeping main's own additions intact (the `default_loops` submodule, VARLEN-FLA batched prefill methods, the NVFP4 lm_head path) and layering the DFlash-specific fields/methods alongside them, rather than reapplying #218's patch wholesale.

`impl_a1.rs`'s misc-scratch inline block (SSM norm kernel, logit softcap, FP32 logits, embed scale, GDN prefill buffers) was extracted into `impl_a1_init::build_misc_scratch` — this was needed anyway to add `gdn_buf_out_f32`/`ssm_verify_h_tmp` without further bloating the constructor.

## Test plan

- [x] `cargo check -p spark-model` — clean
- [x] `cargo clippy -p spark-model --tests` — clean
- [x] Runtime verification on GB10 with a real DFlash drafter checkpoint (`z-lab/Qwen3.6-27B-DFlash`) — see comment below


Sujimoshi:
## Runtime verification on GB10 — found and fixed a real bug ✅

Real DFlash drafter checkpoint found in the HF cache: `z-lab/Qwen3.6-27B-DFlash` (`rope_scaling: null` in its `config.json` — exactly the case the RoPE fix targets), tested against the target checkpoint from #257 (`Qwen3.6-27B-NVFP4`, Standard-NVFP4 U8-packed).

**RoPE fix:** verified two ways — (1) standalone script confirming the new `1/theta^(2j/dim)` formula matches HF's standard unscaled-RoPE exactly (`max_abs_err=0`), and (2) the fix's log line (`DFlash RoPE inv_freq: 64 pairs, theta=10000000, standard, no interpolation`) confirmed on a real server run.

**Inert scaffolding** (new fields/no-op trait methods): confirmed harmless — 3 consecutive requests with `--dflash` **disabled** (plain decode path) all succeeded cleanly, no degradation.

**Bug found in the non-inert part:** with `--dflash` enabled, the first smoke-test request succeeded at the API level but silently hit `CUDA_ERROR_ILLEGAL_ADDRESS` inside `dflash_head::propose::forward_block`, fell back to no-spec, and poisoned the CUDA context — the second request then failed outright (`status 700`).

Root cause: the new NVFP4 lm_head fast path (`w4a16_gemm` in `forward_block.rs`) used `self.vocab_size` (the drafter's own, uncapped config value — 248320 in this checkpoint) as the GEMM's N, but the shared `lm_head_nvfp4` weight buffer was quantized against the **target's** tokenizer-capped vocab_size (248077) — a smaller buffer. The GEMM read ~243 rows past the end of the packed weight/scale buffers. The subsequent argmax loop had the same latent stride bug (wouldn't itself crash, but would read wrong row data for i>0).

**Fix (pushed in `c669023`):** threads the target's actual capped `config.vocab_size` through `build_model → BlockDiffusionDraftHead::from_weights` as a new `lm_head_nvfp4_vocab_size` field, used consistently by both the GEMM call and the argmax loop instead of `vocab_size`.

**Re-verified end-to-end** with the exact same checkpoints/request sequence that reproduced the original crash: 3 consecutive requests (including the one that previously crashed the CUDA context) all completed cleanly, `finish_reason: stop`, zero `illegal address`/`status 700`/fallback-to-no-spec in the logs.

Also ran the standard three checks (`cargo check`/`fmt --check`/`clippy --tests -p spark-model`) after the fix — all clean.

rrstesiak:
The vocab-size OOB fix and the RoPE root-cause are noted; the findings
below are about what's riding along with them.

**1. `from_weights.rs` - RoPE fix hardcodes standard and deletes YaRN.**
The bug is real (drafter ships `rope_scaling: null`, applying YaRN corrupts all
positions), but the fix removes the YaRN branch entirely instead of conditioning
on `config.rope_scaling`. Drafters that legitimately ship yarn (Mistral-Small-4
lineage) now get silently wrong tables. Please make the inv_freq build
config-driven: `null` → standard, `rope_type=="yarn"` → the deleted YaRN math.

**2. `from_weights.rs` - `dense_gemm_pipelined` loaded with `?`.**
Nothing in this PR calls this kernel (the PR body says it's for a later PR in
the split), yet it's loaded with a hard-fail `?`. Any kernel bundle lacking
`dense_gemm_bf16_pipelined` now fails DFlash head construction for zero benefit.
The adjacent `w4a16_gemm` uses `try_kernel` - this should too, or move it to the
PR that uses it.

**3. `dflash_head.rs`/`impl_b3.rs` - `ctx_slot_positions` desyncs by construction.**
Positions are pushed unconditionally in `try_dflash_capture`, but `ctx_len` is
clamped at `max_ctx_len` and nothing pops/rewinds positions on rollback or when
the clamp engages. No consumer exists yet, so this ships as a landmine for the
next PR in the split. Either keep it in lockstep with `ctx_len` (truncate on
clamp, rewind on rollback) or defer the field to the PR that reads it.

**4. Debug cruft.** `impl_a3.rs`: the `LM_HEAD_GEMM_LOGGED` static + pointer-
dumping `tracing::info!` is session-debug leftovers. `impl_b3.rs`: the new
`tracing::warn!` in the ctx_len update fires per prefill chunk for an expected
state (`proposer_state == None`) - warn-level is wrong even when it's DFlash-only.

**5. `impl_a1_init.rs` - FP32 GDN buffer allocated unconditionally.**
`gdn_buf_out_f32` (~2× the BF16 output buffer) is allocated for every GDN model
run - DFlash disabled, `ATLAS_DFLASH_VERIFY_GDN_F32` unset, doesn't matter -
for a prototype the PR itself says is "not yet wired to any consumer". Gate the
alloc on the env var (or defer with the consumer).

**6. Question on direction.** The `prefill_gdn_wy16` scaffolding implies the
upcoming split PRs build a K=16 verify pipeline. Before more of it lands: on
coherent (non-degenerate) output, measured spec decode on this target is
currently *throughput-negative* vs no-spec (draft+verify overhead exceeds
accept savings). What accept-rate data motivates this pipeline? If the answer
is runs with repetitive/degenerate output, the accept numbers are inflated.

Sujimoshi:
Thanks for the thorough review — pushed fixes for points 1-5 in 025d57b:

1. **`dense_gemm_bf16_pipelined` hard-fail** — switched to `try_kernel` (soft-load, matching the `w4a16_gemm` pattern right next to it). It's unused in this PR, so it shouldn't be able to block model load.

2. **RoPE hardcoded to standard** — now config-driven. Added a `rope_scaling` field to `DflashConfig` (parses `rope_type`/`factor`/`beta_fast`/`beta_slow`/`original_max_position_embeddings`), and the inv_freq table build branches on it: YaRN interpolation when `rope_type == "yarn"`, standard formula when null/absent, and a hard `bail!` for any other unrecognized type rather than silently falling back. The current checkpoint still has `rope_scaling: null` so behavior is unchanged for now, but this no longer requires a code change if a future checkpoint uses YaRN.

3. **`ctx_slot_positions` push unguarded** — added a length check against `max_ctx_len` before pushing, to stay in lockstep with ctx_len clamping elsewhere. Also downgraded the "ctx_len update SKIPPED" log from `warn!` to `trace!` since `proposer_state == None` is the expected/common case on ordinary prefill chunks, not an anomaly.

4. **Debug leftovers** — removed the `LM_HEAD_GEMM_LOGGED` static + one-shot GEMM pointer/param dump in `impl_a3.rs`. Leftover from an earlier debugging session.

5. **`gdn_buf_out_f32` unconditional allocation** — now gated behind `ATLAS_DFLASH_VERIFY_GDN_F32=1`; defaults to a null `DevicePtr` and the `total_mb` accounting excludes the term when not allocated. Checked all consumers (`prefill_c.rs`, `batched_layer.rs`, `batch_kernel.rs`) — each guards reads behind `output_f32_written: Cell<bool>`, initialized `false`, so nothing dereferences the null pointer when the flag is off.

All three of `cargo check`, `cargo fmt --check`, `cargo clippy --tests` are clean on `spark-model`.

On point 6 (accept-rate / K=16 verify-pipeline direction) — still discussing internally, will follow up separately once we have data.

SeedSource:

### Triage
- Rationale: Fixes NVFP4 lm_head OOB read for DFlash drafter; adds verify kernels and pipelined GEMM.

### Serve gate

### GLM-5.2 code review
**Correctness concerns:**

1. **Vocab mismatch logic is sound** — using `lm_head_nvfp4_vocab_size` for both GEMM N and argmax stride is correct. However, if `lm_head_nvfp4_vocab_size < vocab_size`, argmax over fewer columns produces token IDs that may exceed the NVFP4 vocab range. Sampling/lookup downstream must handle IDs in `[0, nvfp4_vocab_size)` only. Verify the drafter's `vocab_size` is also capped to match, or token IDs > nvfp4_vocab_size will OOB the embedding table.

2. **`output_f32_written: Cell<bool>`** in `GdnPrefillBuffers` — this struct is likely passed across threads or stored in shared state. `Cell` is `!Sync`; if `GdnPrefillBuffers` is shared across threads this won't compile, or if it's behind `Arc` it's a soundness risk. Should be `AtomicBool`.

3. **`ctx_slot_positions: Vec<i32>`** — added to `DflashProposerState` but no code in this diff populates or reads it. Dead field; ensure follow-up PRs actually use it or remove.

4. **`dense_gemm_pipelined`** — loaded but explicitly unused. Fine as scaffolding, but confirm `try_kernel` failure (handle=0) won't be called inadvertently before the follow-up wires it.

5. **Trait default `Ok(false)`** for both new methods — safe fallback pattern, no risk.

**Risk:** Low-medium. The vocab fix addresses a real OOB; the `Cell<bool>` is the main correctness flag.

Sujimoshi:

PR #259, opened 2026-07-05
feat(gdn-tp): dense-model GDN/SSM TP-shard (dense pure-TP) [draft, stacks on #254+#250]

rsafier:
**Draft — stacks on #254 + #250 (both unmerged).** Base is #254 (`feat/gdn-headparallel-tp-on-main`); this branch also carries #250 (`716143ee`, native-BF16 dense loader) as its first commit because the dense-TP shard sits on #250s restructured dense SSM arm. The **net-new commit is the tip** (`feat(gdn-tp): TP-shard the dense-model GDN/SSM heads`). Once #250 + #254 land on `main`, rebase collapses this to just that commit.

## What

Makes **dense** Holo models (0.8B/4B/9B) TP-shard under pure tensor-parallel. #254 gave the SSM-hybrid **MoE** models GDN HeadParallel; the **dense** loader was never wired for it, so dense TP2 produced garbage (TP1 fine). This wires the *same* `shard_gdn_*` slicers (`tp_shard/gdn.rs`) into the dense loaders SSM arm.

**Loader-only.** `config` holds per-rank-LOCAL linear head counts (`topology.rs` divides them), so `TpGdnDims::from_config` rebuilds the FULL pre-shard sizes; load/concat/interleave run at FULL, then `shard_gdn_qkvz_rows` / `_ba_rows` / `_conv_rows` / `_value_vector` (a_log, dt_bias) / `_out_proj_row_parallel` cut this ranks contiguous head range. `norm.weight` `[vd]` is **replicated** (shared across value heads). The post-`out_proj` all-reduce already fires in the shared `Qwen3SsmLayer` forward. Gated `tp_size>1`; **byte-identical at tp==1** (else-arm pass-through, every slicer no-ops).

## Validation — 2-node pure-TP2 (GB10 × 2)

| Model | Before | After |
|---|---|---|
| dense **0.8B** | garbage (multilingual) | **coherent** — "red, blue, and yellow" |
| dense **4B** | garbage | **coherent** — "Paris" |
| MoE **35B** (regression) | — | **unaffected** — "15 plus 27 equals 42" |

Rebuilt the branch binary (not the wip prototype) and re-confirmed 0.8B TP2 coherent on this exact base. Compiles + fmt clean.

## Scope

BF16/NVFP4 dense only. The native-FP8 GDN SSM path (`ATLAS_NO_GDN_FP8`, gated on `proj_is_fp8_any_scale`) is **not yet sharded** — dense-FP8-TP is a follow-up (the BF16 dense models dont take that path). Mirrors #254 being MoE-TP-scoped.

### DIFF
--- a/crates/spark-model/src/layers/qwen3_ssm/trait_decode_multi_seq/ssm_batched.rs
+++ b/crates/spark-model/src/layers/qwen3_ssm/trait_decode_multi_seq/ssm_batched.rs
@@ -34,7 +34,16 @@ impl Qwen3SsmLayer {
         // FP8 build → batched w8a16 GEMM; NVFP4 build → batched w4a16 GEMV
         // (batch4/16, M<=16). Either amortizes the QKVZ/out_proj weight read
         // across the n seqs; otherwise the per-seq loop re-streams it n times.
-        let qkvz_ok = (self.qkvz_nvfp4.is_none() && self.w8a16_gemm_k.0 != 0)
+        // qkvz_nvfp4.is_none() covers TWO builds: the FP8 build (qkvz_fp8w
+        // Some -> batched w8a16 GEMM, needs w8a16_gemm_k) and the pure
+        // native-BF16 build (both None -> the batched `dense_gemm` fallback
+        // below, which uses dense_gemm_k, always loaded). Gate each on the
+        // kernel it actually dispatches so the BF16 build engages the batched
+        // fast path instead of silently dropping to the per-seq loop. The FP8
+        // sub-case (qkvz_fp8w Some) is byte-identical to the old gate.
+        let qkvz_ok = (self.qkvz_nvfp4.is_none()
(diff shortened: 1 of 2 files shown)


SeedSource (approved):
**Approve** (SeedSource, CODEOWNER) — native-BF16 dense loading (`Bf16Raw`) for no-metadata Holo/Ornith dense checkpoints: keeps Q/K/V/O + FFN in BF16 (drops the lossy runtime BF16→NVFP4 quant), fixes the `qkvz_ok` gate so the BF16 build engages the batched fast path (FP8 sub-case byte-identical), and includes a real use-after-free fix — snapshotting the BF16 FFN weights *before* `load_dense_ffn` frees the store buffer, avoiding a `dense_gemm_tc` CUDA-700 on first prefill. Well-reasoned with explicit failure-mode comments. CI green. Scope: code-read + CI only; not behaviorally exercised our side (Holo dense isn't in our serve set) — deferring to CI + author.

PR #260, opened 2026-07-05
fix(loader): gate CUTLASS grouped NVFP4 MoE on architecture, not model_type string

rsafier:
**Draft until the runtime regression check lands** (building now — validating Holo-3.1-35B *and* AgentWorld both engage CUTLASS grouped + stay coherent on the actual binary).

## Problem
The fast **CUTLASS grouped NVFP4 MoE** path (opt-in via `ATLAS_HOLO_LOW_MEMORY_MOE=1`) was gated on `config.model_type == "holo3_1_moe"`, so **any other NVFP4-MoE checkpoint silently fell back to the slow `moe_w4a16` path** even with the flag set. Concrete repro: `sakamakismile/Qwen-AgentWorld-35B-A3B-NVFP4` (model_type `qwen3_5_moe`, compressed-tensors NVFP4) — no `CUTLASS grouped SFB` signal, ran `moe_w4a16`.

But `holo3_1_moe` / `qwen3_5_moe` / `qwen3_6_moe` are the **same architecture at the weight level** — `factory::loader_for_config` maps all three to `Qwen35WeightLoader`, and `holo3_1_moe` is a label *fabricated* at `config/dispatch.rs:120-128` (rewrites `qwen3_5_moe`→`holo3_1_moe` when `image_token_id==248056`) purely so kernel-target **directory** lookup can disambiguate 3.5 vs 3.6 (they collide on hidden=2048). Downstream code then mistook that label for an architecture discriminator.

## Fix (2 gates, one file — `qwen35/load_layers.rs`)
- `holo_nvfp4_moe` → `nvfp4_moe = config.num_experts > 0 && quant_format == Nvfp4` (architecture, not label).
- `is_holo_modelopt_mixed_precision` → `is_modelopt_mixed_precision`: drop the `model_type == "holo3_1_moe"` conjunct; keep the quant-config test (any Qwen3.5/3.6-family MoE shipping modelopt MIXED_PRECISION is now recognized).

## Scope / safety
- **No kernels move.** The `config/dispatch.rs` model_type rewrites (needed for kernel-directory resolution) are deliberately untouched — this only changes *which MoE dispatch fires*, not kernel layout.
- **No regression on Holo:** `holo3_1_moe` still qualifies (`num_experts=256`, Nvfp4), and the mixed-precision path still fires for Holo (modelopt MIXED_PRECISION). It now *also* fires for architecturally-identical siblings shipping the same quant format — no existing checkpoint changes behavior (we ship no non-Holo modelopt-MIXED MoE).
- This is the **focused bug fix**. A broader capability-based refactor (replacing model_type string checks across the codebase — MTP allowlist, MLA guard, etc.) is a separate, larger effort scoped for a follow-up.


rsafier:
## Runtime regression — PASS (validated on the actual binary, both models)

Built `spark-genmoe` (main + this fix, ATLAS_TARGET_MODEL=*), served each single-node with `ATLAS_HOLO_LOW_MEMORY_MOE=1`:

**AgentWorld** (`qwen3_5_moe`, compressed-tensors NVFP4) — the fix:
- `CUTLASS grouped SFB: built 256 experts` ×40 layers (was falling back to `moe_w4a16` before)
- Coherent: "The capital of France is Paris."

**Holo-3.1-35B-A3B-NVFP4** (`holo3_1_moe`, modelopt MIXED_PRECISION) — no regression:
- `CUTLASS grouped SFB` ×40 layers (unchanged)
- `is_modelopt_mixed_precision` still fires: "routing Holo ModelOpt SSM/attention projections through native FP8", "Layer N: SSM native FP8 — w8a16" — the modelopt path is byte-for-byte the same
- Coherent: "15 plus 27 equals 42."

No kernels moved; `config/dispatch.rs` rewrites untouched. cargo check + fmt clean.

SeedSource:

### Triage
- Rationale: Broadens NVFP4 MoE and modelopt mixed-precision detection, removing model_type gates to support more checkpoints.
- Serving-path files: `crates/spark-model/src/weight_loader/qwen35/load_layers.rs`

### Serve gate
- Top-k KL vs main: **0.0001** (no shift vs main) - decode delta **-0.0%**

### GLM-5.2 code review
**Correctness:** Sound. Gating on `num_experts > 0` + `QuantFormat::Nvfp4` is the right architectural check; removing the `model_type == "holo3_1_moe"` string match correctly unblocks sibling/third-party checkpoints routed through `Qwen35WeightLoader`. The `is_modelopt_mixed_precision` rename and de-gating is consistent — the quant_config fields are sufficient and the model_type check was redundant exclusion.

**Risk:** Moderate. The broadened `nvfp4_moe` predicate now activates the CUTLASS grouped NVFP4 MoE transpose path for *any* NVFP4 MoE checkpoint. If a non-Holo NVFP4 MoE has expert weight layouts that differ from what the grouped path expects (e.g., different packing, scaling block size), this will silently route through an untested path rather than the known-safe `moe_w4a16` fallback. The env-var gate (`ATLAS_HOLO_LOW_MEMORY_MOE=1`) provides a mitigation, but only if operators know to set it.

**Quality:** Comment updates are accurate and well-scoped. `eq_ignore_ascii_case("MIXED_PRECISION")` is pre-existing and technically wrong (ASCII-case-insensitive compare against an uppercase literal matches lowercase but not mixed-case like "Mixed_Precision") — not introduced here, but worth a follow-up.

tbraun96:
This is the one of the four I want to land — but it needs splitting first, because half of it is a real win and half is unbounded latent risk.

**The half I want.** `load_layers.rs:112` currently reads `holo_nvfp4_moe = model_type == "holo3_1_moe" && Nvfp4`, so any other NVFP4 MoE checkpoint — AgentWorld, `Sehyo/Qwen3.5-35B-A3B-NVFP4` (which is in the serve matrix) — silently misses the CUTLASS grouped path even with `ATLAS_HOLO_LOW_MEMORY_MOE=1`. That's a real, named repro, it restores the factory's own stated invariant (`factory.rs:62`: "This is the ONLY place model_type strings are matched"), and it is **env-gated**, so the blast radius is bounded. The flagship FP8 is provably unaffected.

**The half I don't.** De-gating `is_modelopt_mixed_precision` from `holo3_1_moe` is *not* env-gated and, by the PR's own admission, has no shipped beneficiary ("we ship no non-Holo modelopt-MIXED MoE"). `modelopt_mixed_precision` feeds the ungated FullAttention arm at `:544`, which routes to the BF16-dequant diagnostic path and hard-`bail!`s on TP>1 (`:550-555`), plus the SSM arm at `:790`. So it delivers nothing today while arming a silent default-path change — and a TP=2 load failure — for the first future modelopt-MIXED MoE checkpoint anyone adds.

**On the validation.** The serve gate ran `Qwen3.6-35B-A3B-FP8`, where every changed line is inert (`nvfp4_moe=false`, `modelopt_mixed_precision=false`) — so "KL 0.0001" is a true statement about a code path this PR doesn't touch. This is precisely the shape #281 had: a shared-launcher grid-axis swap that passed a coherence smoke while every *other* NVFP4 model silently computed the wrong tile. It's why gate C2 exists and runs first.

**To merge:**
1. Drop the `is_modelopt_mixed_precision` de-gating. Keep `nvfp4_moe = num_experts > 0 && Nvfp4` for the env-gated MoE path only.
2. Run the real gates — C2 first, then A/B/D plus a BFCL subset leg — on a checkpoint that **newly** qualifies for the CUTLASS grouped path, not on Holo. `Sehyo/Qwen3.5-35B-A3B-NVFP4` is the obvious candidate since it's already in the matrix.

With that split the five gates buy a demonstrated win on a named checkpoint. As submitted, half the diff is risk nobody has exercised.

Happy to run the gates myself once the split lands — the boxes are free and I'd rather that not be the thing blocking you.

rsafier:
@tbraun96 Split applied as asked: the `is_modelopt_mixed_precision` de-gating is gone (the detector is byte-identical to main again); what remains is only the architecture gate — `num_experts > 0 && Nvfp4` — with the comment updated to say why it's keyed on the arch, not the label. Branch is synced with main (clean merge, no conflicts), fmt/clippy/tests green locally. Ready for the C2-first gate run you offered — the newly-qualifying checkpoint reasoning is unchanged from the PR body (Sehyo/Qwen3.5-35B-A3B-NVFP4 in the serve matrix).

tbraun96:
Landed in main via the #440 stack merge (`3c941751`). Verified by content diff — `crates/spark-model/src/weight_loader/qwen35/load_layers.rs` is byte-identical to this PR's head, including the arch-not-`model_type` gate and `build_cutlass_grouped_sfb`. Thanks @rsafier.

PR #263, opened 2026-07-05
feat(lora): runtime BF16 LoRA adapter support (Holo-3.1 dense + 35B MoE)

rsafier:
## Problem

Atlas serves the Holo / Qwen3.5 family, but there was no way to serve a **LoRA fine-tuned**
variant: the standard PEFT adapter format wasn't understood, and there was no runtime path to
apply a low-rank delta on top of the NVFP4-quantized base.

## Solution

Apply the LoRA update as a **runtime BF16 delta, never merged** into the (NVFP4) base weights —
`y += scale·(x@Aᵀ)@Bᵀ` at attention **k_proj / v_proj / o_proj** on the **full-attention layers
only**, reusing existing kernels (`dense_gemv_bf16` / `dense_gemm_tc` / `bf16_scaled_add`). Zero
new CUDA kernels; deltas are graph-safe (captured inside the existing decode CUDA graphs).

- **PEFT parser** (`atlas-core`): `adapter_config.json` → validated config; every unsupported
  feature is a named `REJECT(...)` (DoRA, bias, rank/alpha patterns, `modules_to_save`,
  `all-linear`, absent `use_rslora`, non-`LORA` `peft_type`). rsLoRA (`α/√r`) scaling supported.
- **Adapter loader** (`spark-runtime`): dedicated `adapter_model.safetensors` loader with host-side
  **F16→BF16 and F32→BF16** conversion (F32 is PEFT's default save dtype), header-only OOM
  preflight, pickle reject.
- **Install walk + pool** (`spark-model`): key remap + per-`LayerType` allow-list — only
  full-attention × k/v/o is accepted; `q_proj` (gated Q|gate), GDN, MoE-MLP, and wrong-layer
  tensors are named hard rejections, never silent skips. A/B are rank-padded into one
  fixed-address pool with per-module `[max_loras]` pointer tables (the frozen layout for future
  multi-adapter routing; v0 fills slot 0). Pool VRAM is budgeted against the KV cache.
- **Serve wiring** (`spark-server`): `--lora-adapter NAME=PATH_OR_HF_ID`, `--max-lora-rank`,
  `--max-loras`; the adapter name is advertised by `GET /v1/models` and accepted in `request.model`.
- **Families**: dense `qwen3_5` (Holo-3.1-0.8b) and `holo3_1_moe` (Holo-3.1-35B-A3B MoE) — a
  k/v/o-only adapter leaves `ffn_weights = None`, so the MoE expert path is never exercised.

## Validation

Verified end-to-end on a GB10 with **exact parity to `peft`/transformers**:

- **Holo-3.1-0.8b** (dense): base output generic; adapter reproduces the trained persona +
  made-up codeword verbatim.
- **Holo-3.1-35B-A3B** (MoE): adapter trained against the BF16 base, served on the compact
  **NVFP4** base with the BF16 delta applied to the 10 full-attention layers — persona verbatim.
- `examples/lora_apply_microtest.rs`: a **runtime parity oracle** that runs the real CUDA apply
  (shrink → expand → fold), decode and prefill shapes, bisected against a bf16-faithful CPU
  reference. Plus a gated `#[ignore]` GPU test pinning the F32→BF16 conversion, and the offline
  parity oracle for loaded A/B.

Draft: opening for review/CI on upstream. Deferred (documented cut lines): dense-FFN delta,
per-request multi-adapter routing (`lora_bgmv`), q_proj/GDN/MLA targets, TP>1.


rsafier:
## Review — #263 (runtime BF16 LoRA adapter support)

**Verdict: strong, well-engineered POC — mergeable as the opt-in draft it is (gated behind `--lora-adapter`, so zero effect on non-LoRA serving). Before it's used in anger, close the two runtime silent-skip holes below, because they contradict the PR's own central safety invariant.**

Reviewed the full path: PEFT parser (`config/parsers/lora.rs`), adapter loader (`weights/adapter.rs`), install/pool (`lora/mod.rs`), apply math (`ops/lora_delta.rs`), and every apply site in prefill/decode. CI green; `atlas-core` parser tests + fmt verified locally.

### What's genuinely good
- **Reject hygiene is the star.** Every unsupported PEFT feature is a *named* `REJECT(...)` at parse **and** a second `REJECT[...]` at `classify_key`, with a bidirectional tensor↔target audit + A/B shape audit. `use_rslora` is never defaulted (a wrong scale is silent quality loss). 18 parser tests.
- **`apply_lora_delta` math is correct** — shrink→expand→fold, run at the padded `max_rank` with zeroed pad rows/cols (bit-identical to true rank), scale folded at `bf16_scaled_add`, never pre-merged. Decode k/v applied *before* norm/RoPE/kv-write so the cache stores adapted values (matches HF `k_norm(k_proj(x)+Δ)`).
- **`adapter.rs` F32→BF16** is a real catch: PEFT's default save dtype is F32; left as-is the BF16 pool would read at half-stride = silent garbage. F16→BF16 too, pickle reject, header-only OOM preflight.
- **Pool is budgeted before the KV snapshot** so it shrinks the KV cache (correct on GB10 unified memory); graph-safe (pool/arena/scale all load-time-fixed) with an `ATLAS_LORA_EAGER` hatch and a proven graph==eager parity.

### Findings

**1. The "never silent skip" invariant has two *runtime* holes (the main callout).** The PR's stated contract (docs: "Named hard rejections — never silent skips") doesn't hold at runtime in two documented-but-unguarded places:

- **1a — FFN gate/up/down are accepted, packed (VRAM), stored, and never applied.** `parse`/`classify_key`/the pool all accept `{gate,up,down}` and the **shipped fixture targets them** (doc §Deferred), but `dense_ffn.rs`'s `lora` field is `#[allow(dead_code)]` with no forward insertion. So at *any* batch size an adapter targeting FFN silently gets attention-only adaptation, no warning. The offline oracle's "36/36 modules, 0.0 rel-err" validates the *loaded* weights, not the *applied* forward — nothing catches this. **Fix:** hard-reject `gate/up/down` at parse until the compute is wired (restores the invariant; the doc says wiring it is "additive," so either direction is small).

- **1b — concurrency ≥2 (multi-seq decode) and prefix-cache warm-hit silently skip all deltas, unguarded.** The apply sites live only in the single-seq `attention_forward.rs` + `paged_qkv/oproj`; `trait_impl/multi_seq/` has no PEFT-LoRA apply (its `*_lora` symbols are DeepSeek MLA, a different thing). The only serve guard is `world_size==1` (TP). "Run the POC at batch size 1" is documented but **nothing enforces it** — a production deploy taking concurrent requests silently returns base output. **Fix:** when an adapter is active, force `max_num_seqs=1` and disable prefix caching at startup (mirror the existing `world_size==1` reject), turning a silent-wrong-output footgun into an enforced constraint.

**2. Minor — always-on adapter.** `/v1/models` advertises the adapter and base-name requests get adapted output with a one-line warn. Acknowledged v0 wart; fine for a POC.

**3. Nit — the pool packs all 6 modules for every full-attn layer regardless of targets.** A k/v/o-only adapter still allocates gate/up/down pad space. Bounded and intentional (frozen M2 layout), but on the 35B (large `intermediate_size`) the unused FFN padding is a non-trivial slice of a VRAM budget that comes straight out of the KV cache — worth a follow-up once the layout is real.

### Bottom line
Excellent engineering and test discipline for a POC; the docs are honest about the cut lines. It's opt-in and startup-only, so merging it as a draft/POC is low-risk. The gating item before *real* use is **1b** (enforce batch=1 / no prefix-cache when LoRA is active) and ideally **1a** (reject FFN targets until applied) — so the "never silent skip" guarantee the PR itself establishes actually holds when someone points traffic at it.

SeedSource:

### Triage
- Rationale: Adds LoRA adapter config parsing and max rank support; impacts model config and PEFT adapter handling.
- Serving-path files: `crates/spark-model/src/factory.rs, crates/spark-model/src/factory/build.rs, crates/spark-model/src/layer/transformer_layer.rs, crates/spark-model/src/layers/dense_ffn.rs, crates/spark-model/src/layers/ops.rs, crates/spark-model/src/layers/ops/lora_delta.rs`

### Serve gate
- Build: 19s - Qwen3.6-35B-A3B-FP8 on GB10 node1
- Top-k KL vs main: **0.0001** (no shift vs main) - decode delta **+0.0%**

### GLM-5.2 code review
**Diff truncated** — review based on visible portion.

**Correctness:** Parser logic is sound. `scaling()` correctly handles rsLoRA vs standard. Hard-fail stance on `use_rslora` absence (rejecting rather than defaulting to `false`) is the right call — silent scale mismatch would corrupt output. `peft_type` check is optional (`#[serde(default)]`) but only validates when present; a missing `peft_type` silently passes, which may be too permissive for a "hard-fail" parser.

**Risk:** `PEFT_SUPPORTED_TARGET_MODULES` omits `q_proj` — documented and intentional (gated Q), but users with `q_proj`-targeting adapters will hit a runtime reject with no migration path. `layers_to_transform` parsed but unused beyond logging — the loader gate is authoritative, so a mismatch between config-declared layers and actual delta tensors would only surface at load time, not parse time.

**Quality:** Naming discipline (`adapter_*` / `peft_*`) avoids MLA collision — good. File-size cap exemption comment is thorough but borders on over-justifying. `debug_assert` for `r > 0` is appropriate since it's validated upstream.

**Concern:** Cannot verify the loader-side enforcement or CUDA merge kernel from truncated diff. The `as_any_mut` downcast hook (mentioned in cap exemption) is a code smell worth scrutinizing in the full diff.

PR #212, opened 2026-06-28
test(harness): official-suite vision test + Holo-3.1/Ornith/27B matrix

rsafier:
**Draft — stacked on #210** (→ #203 → #202). Only the final commit is net-new here; everything below is the ancestor stack and will collapse when those merge. Net diff vs #210 is **`tests/` only** — no engine/kernel code.

## What this adds
Brings the GB10 VL targets we've enabled under the official single-GPU harness, with real image-path coverage.

**`tests/single_gpu_suite.py`**
- `run_vision_test`: posts an image (`mona_lisa.jpeg` fixture) and asserts the model's description contains expected keywords. Overridable via `ATLAS_VISION_TEST_IMAGE` / `ATLAS_VISION_TEST_KEYWORDS`.
- Vision is **opt-in via an explicit `--vision` flag** (replaces the prior substring auto-probe), so only vision-marked specs exercise the ViT path.

**`tests/run_all_models.py`**
- `TestSpec` gains `vision: bool`; `run_suite` passes `--vision` when set.
- Marks the seven VL targets: 27B-dense-nvfp4, qwen3-vl-30B, gemma-4-31B, gemma-4-26B, qwen3.6-35B-a3b-fp8, holo-3.1-35B-nvfp4, ornith-1.0-9B.
- Adds Holo-3.1 and Ornith matrix rounds.

## Validation
Run end-to-end on GB10 against the vision-marked + flagship/27B subset; image descriptions verified correct (e.g. 27B → Mona Lisa), text parity unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #265, opened 2026-07-05
test(harness): official single-GPU suite + soak + vision fixtures (recut of #212 onto main)

rsafier:
Recut of #212 onto post-#229 `main`.

#212's original base (the Holo FP8 stack, `feat/holo-fp8-on-holo-gb10`) was superseded when the Qwen3.6-27B NVFP4 stack (#229) squash-merged to `main`. This branch replays #212's own commits onto current `main` via `git rebase --onto main`, so it carries **only** the test-harness additions with no stale stack underneath.

## Contents (test-harness only — zero `.rs` changes)
- `tests/single_gpu_suite.py` — official single-GPU suite: needle-in-haystack long-context (4k/8k/16k), diverse filler (no degenerate repetition), strict verbatim-needle pass/fail, repetition-loop detection.
- `tests/run_all_models.py` — Holo-3.1 / Ornith / Qwen3.6-27B model matrix.
- `tests/stack_model_smoke.sh` — per-model stack smoke.
- `scripts/realistic_soak.py` — mixed-workload soak: cacheable shared-prefix builder, tool-call + vision + thinking mixes, err counter + `think_rate` param (the #212-review follow-up commit).
- `scripts/gen_test_images.py` + 8 committed PNG fixtures (`tests/fixtures/images/`) — deterministic vision inputs, resolved relative to script location (CWD-independent).

## Review notes addressed
- **SeedSource/GLM "SHARED_CTX truncation" concern**: verified **not a bug** against this code. `SHARED_CTX` is `_HIST * 400` = **102,800 chars**, which covers `target=8000 → 32,000 chars` and even the suite's `16000 → 64,000 chars` with margin; the slice never truncates. `single_gpu_suite.py`'s long-context builder is repeat-based (`generate_long_context`) with no slicing at all. No code change needed.

## Validation
- `python3 -m py_compile` clean on all four scripts; `bash -n` clean on the smoke script.
- No Rust changes (`git diff --name-only main HEAD` has no `.rs`), so fmt/clippy/test are unaffected — the one `load_layers.rs` hunk from the old base netted to a no-op on post-#229 `main` and dropped out of the rebase.

Supersedes #212.


SeedSource:

### Triage
- Risk: **low** - serve-test-required (triage advisory): `False` - path-heuristic (authority): `False`
- Rationale: Test script updates and new test image fixtures; no core inference or serving changes.

### Serve gate
- **Skipped** - no serving-path files (docs/CLI/test/example). Zero GPU.

### GLM-5.2 code review
**Correctness:** The `st["err"] += 1` fix is correct and important — the `--max-error-rate` gate was previously dead code. The `think_rate` refactor removes a global, good. The vision test logic is sound: gated behind `--vision`, never auto-detected.

**Risks:**
1. **Missing fixture:** `run_vision_test` references `tests/fixtures/mona_lisa.jpeg` but the diff only adds `tests/fixtures/images/*.png`. If `mona_lisa.jpeg` isn't already in-tree, every vision-marked round fails immediately. Verify this file exists or add it.
2. **Keyword matching is fragile:** `"smil"` (truncated) is a band-aid for "smile/smiling"; a model that says "smirk" or omits it entirely triggers a false FAIL. The keyword list is too narrow for a correctness gate.
3. **`extract_text` / `_strip_thinking` / `_has_repetition_loop`** are called but not shown in the diff — confirm they handle multimodal response shapes (some VL models nest content differently).
4. **Rounds 10-11** add two new model checkpoints (`Hcompany/Holo-3.1`, `deepreinforce-ai/Ornith`) — ensure these are cached on the run host or the round hangs on download.
5. **`vision=True` on `35B-qwen36-fp8`** (Round 6) — confirm Qwen3.6-35B-A3B actually ships a vision tower; the comment says "architectural twin" but the checkpoint may be text-only.

**Quality:** Comment density is high but acceptable for a harness. The `("head", None)` padding in rounds 10-11 is consistent with existing pattern. The truncated diff cuts off mid-print-statement; ensure the function closes properly.

**Verdict:** Request: (1) confirm `mona_lisa.jpeg` exists, (2) broaden keyword matching or make it advisory, (3) verify the two new checkpoints are pre-cached.

tbraun96:
Reviewed this against current `main` and **pushed two amendments to your branch** (`ef44f4c24`) so it can land — disclosing rather than opening a competing PR.

**Why this is worth merging: it repairs two things main already ships broken.**

1. `scripts/realistic_soak.py` initialises `st["err"]`, reads it at `:260`/`:275`, and gates `exit 1` on it at `:276` — but **never increments it**. Verified against main: no `st["err"] +=` exists there. So `--max-error-rate` has never been able to fire; the soak's stability gate always passes. Your `st["err"] += 1` fixes it.

2. `realistic_soak.py:139` already defaults to `tests/fixtures/images` and its own help text calls it "committed" — but the directory isn't on main. `--vision` soak silently degrades to text-only today. The 8 PNGs (115 KB total, regenerable from the included `gen_test_images.py`) repair main's own dangling reference.

I also want to correct the framing in the PR body for the record: most of #212's harness is *already* on main and operationally executed — `single_gpu_suite.py`/`run_all_models.py` are the serve-matrix release gate. The real delta here is small and good.

**What I removed, and why:**

- `tests/stack_model_smoke.sh` (144 lines). I grepped the whole tree, all 13 workflows, and main: **zero references**. Unrun code reads as coverage while proving nothing — it's a pattern we've been actively removing elsewhere this week.

- Rounds 10 and 11 (Holo-3.1-35B-A3B-NVFP4, Ornith-1.0-9B). `gate_results.py` says plainly that "Coverage is enforced, not assumed" and a missing manifest is a hard FAIL unless waived — so adding rounds expands the **mandatory** release-gate roster, and your own comment notes those checkpoints may not be cached on the run host. That's a decision worth making on its own rather than as a rider on a fixtures PR. Happy to take it as a follow-up once the checkpoints are confirmed present; I'd merge that same day.

No `crates/` change, so this owes no GPU gates — which is why it can land now while the loader PRs wait on theirs.

Thanks for the recut onto main; that's what made it reviewable.

PR #266, opened 2026-07-05
perf(decode): n>=4 tiled QKV + o_proj GEMV for NVFP4 batched decode (extract of #214 onto main)

rsafier:
Focused extraction of the still-unabsorbed batched-decode feature from #214 onto post-#229 `main`.

## Why a focused extraction (not a rebase of #214)
#214's base predates #229. A mechanical `rebase --onto main` produces ~39K deletions because #229 added the MMQ/q4k-vendor/NVFP4 FFN stack that #214 doesn't have — the loader/kernel surface diverged too far to replay safely. Comparing #214's own `.rs` against current `main` shows that main **already absorbed** the n=2/n=3 batched QKV paths (`ms_qkv_batch2`/`ms_qkv_batch3` live in `multi_seq/qkv.rs`). The one genuinely-novel, still-absent piece is the **n≥4 tiling** that extends those proven kernels to higher decode concurrency. This PR ports exactly that, adapted to main's current `multi_seq` structure.

## What it does
For n≥4 concurrent decode with NVFP4 q/k/v/o weights, tile the proven `batch3`/`batch2` GEMV kernels — greedy 3s, then a 2-or-1 remainder. Each tile reads the projection weight **once** for its 2–3 tokens, so n=8 costs 3 weight-matrix reads instead of 8. This amortizes the attention-projection bandwidth that otherwise makes decode collapse back to one weight read per token at C≥4.

- `multi_seq/qkv.rs`: new `ms_qkv_batch_tiled` + a `n >= 4 && <all nvfp4>` dispatch branch. Scratch layout is **byte-identical** to `ms_qkv_batch3` (same `ssm_qkvz()`/`attn_output()` scratch, same `q_proj_bytes`/`kv_bytes`/`per_seq_qkv` offsets, all stream-serialized), so it inherits that path's correctness. The `k==1` remainder falls back to the per-seq projection.
- `multi_seq/attn.rs`: matching n≥4 `o_proj` tiling (`attn_out`/`o_out` are contiguous, so each tile is just an offset into the same `batch3`/`batch2` kernels the n==2/3 branches already use).

No behavior change for n<4 or non-NVFP4 weights (existing branches untouched; the new branch sits between the n==2 case and the seq-loop fallback).

## Correctness argument
Correct by construction: identical kernels + offsets to the already-shipped `ms_qkv_batch3`, which this repo already trusts for n=3. The scratch buffers are already sized for k≤3 (main's `ms_qkv_batch3` uses `attn_output().offset(3*kv_bytes)`); the tiled path never exceeds a k=3 tile. Cross-tile scratch reuse is safe because each tile's `copy_d2d_async` into `qkv_buf` is ordered before the next tile's GEMV on the same stream.

## Validation
- `cargo fmt -p spark-model --check` clean.
- `cargo clippy -p spark-model -- -D warnings` clean.
- `cargo test -p spark-model --lib` → 79 passed, 0 failed.
- **GPU numeric A/B recommended before merge**: tiled-vs-per-seq output identity + a needle recall at C4/C8 on an NVFP4 Holo/AgentWorld model. GPU kernel numerics aren't covered by the CPU CI suite (all `#[ignore]`), so this is a manual pre-merge step.

Supersedes the batched-decode portion of #214 (the rest was absorbed by #229). Signed commit.

### DIFF
--- a/crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/attn.rs
+++ b/crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/attn.rs
@@ -285,22 +285,51 @@ impl Qwen3AttentionLayer {
                 stream,
             )?;
         } else if !self.attn.o_proj.is_null() {
-            // WIDE-VERIFY BATCHED O-PROJ (DFlash γ=16, n>3). One GEMM reads
-            // the o_proj weight ONCE for all n rows instead of the per-row
-            // GEMV loop below. attn_out is contiguous [n, q_dim]; o_out is
-            // contiguous [n, h]; both already laid out for a single M=n GEMM
-            // (no scatter). Uses the pipelined m128_v2 kernel when the
-            // transposed weight is present (base M64 GEMM is the slow path).
-            self.wide_verify_gemm(
-                c,
-                attn_out,
-                &self.attn.o_proj,
-                self.o_nvfp4_t.as_ref(),
-                o_out,
-                n as u32,
-                h as u32,
-                nq * hd,
-            )?;
+            // n>=4 (or n==1): tile the proven batch3/batch2 GEMV kernels so the
+            // o_proj weight is read once per tile (n=8 -> 3+3+2 = 3 reads vs 8).
+            // attn_out [n,q_dim] and o_out [n,h] are contiguous, so each tile is
+            // just an offset into the same kernels the n==2/3 branches use.
+            let mut s = 0usize;
+            while s < n {
+                let k = (n - s).min(3);
+                let ai = attn_out.offset(s * q_dim as usize * bf16);
+                let oi = o_out.offset(s * h * bf16);
+                if k == 3 {
+                    ops::w4a16_gemv_batch3(
+                        fwd.gpu,
+                        self.w4a16_gemv_batch3_k,
+                        ai,
+                        &self.attn.o_proj,
+                        oi,
+                        h as u32,
+                        nq * hd,
+                        stream,
+                    )?;
+                } else if k == 2 {
+                    ops::w4a16_gemv_batch2(
+                        fwd.gpu,
+                        self.w4a16_gemv_batch2_k,
+                        ai,
+                        &self.attn.o_proj,
+                        oi,
+                        h as u32,
+                        nq * hd,
+                        stream,
+                    )?;
+                } else {
+                    ops::w4a16_gemv(
+                        fwd.gpu,
+                        self.w4a16_gemv_k,
+                        ai,
+                        &self.attn.o_proj,
+                        oi,
+                        h as u32,
+                        nq * hd,
+                        stream,
+                    )?;
+                }
+                s += k;
+            }
         } else {
             for i in 0..n {
                 let attn_out_i = attn_out.offset(i * q_dim as usize * bf16);
--- a/crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/qkv.rs
+++ b/crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/qkv.rs
@@ -1,13 +1,16 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 
-//! Phase 2: per-token Q/K/V projection. Three branches:
-//! - n=3 + NVFP4 → batch3 GEMV path
-//! - n=2 + NVFP4 → batch2 GEMV path
-//! - else        → sequential per-token GEMV (FP8/NVFP4/BF16 fallback)
+//! Phase 2: per-token Q/K/V projection. Four branches:
+//! - n=3 + NVFP4  → batch3 GEMV path
+//! - n=2 + NVFP4  → batch2 GEMV path
+//! - n>=4 + NVFP4 → tiled batch3/batch2 GEMV path (greedy 3s + 2/1 remainder)
+//! - else         → sequential per-token GEMV (FP8/NVFP4/BF16 fallback)
 //!
-//! Both batch paths read each weight once for N tokens and then scatter
-//! into the per-seq QKV layout. The sequential path repeats the GEMV per
-//! token but supports every weight encoding.
+//! The batch paths read each weight once per tile and then scatter into the
+//! per-seq QKV layout. The tiled path extends that to n>=4 so decode
+//! concurrency doesn't collapse back to one weight read per token. The
+//! sequential path repeats the GEMV per token but supports every weight
+//! encoding.
 
 use anyhow::Result;
 
@@ -48,15 +51,16 @@ impl Qwen3AttentionLayer {
             && self.v_weight.as_ref().and_then(|w| w.as_nvfp4()).is_some()
         {
             self.ms_qkv_batch2(c)?;
-        } else if n > 3
+        } else if n >= 4
             && self.q_weight.as_ref().and_then(|w| w.as_nvfp4()).is_some()
             && self.k_weight.as_ref().and_then(|w| w.as_nvfp4()).is_some()
             && self.v_weight.as_ref().and_then(|w| w.as_nvfp4()).is_some()
         {
-            // WIDE-VERIFY BATCHED QKV (DFlash γ=16, n=17). The last per-token
-            // weight loop in the wide verify — one GEMM per Q/K/V reads each
-            // weight ONCE for all n rows instead of n× (mirrors batch3 with M=n).
-            self.ms_qkv_batchn(c)?;
+            // n>=4 NVFP4: tile the proven batch3/batch2 GEMV kernels so the
+            // q/k/v projection weights are read once per tile (n=8 -> 3+3+2 =
+            // 3 weight-matrix reads instead of 8 per-seq) — amortizes the
+            // attention-projection bandwidth that otherwise caps decode scaling.
+            self.ms_qkv_batch_tiled(c)?;
         } else {
             for i in 0..n {
                 let normed_i = normed.offset(i * h * bf16);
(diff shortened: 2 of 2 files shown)


SeedSource:

### Triage
- Rationale: Adds n>=4 NVFP4 tiled GEMV paths for o_proj and qkv projections using proven batch kernels.
- Serving-path files: `crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/attn.rs, crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/qkv.rs`

### Serve gate
- Top-k KL vs main: **0.0006** (no shift vs main) - decode delta **+0.1%**

### GLM-5.2 code review
**Correctness concerns:**

1. **Scratch buffer aliasing (HIGH):** `q_scratch = ssm_qkvz()`, `k_scratch = attn_output()`. `v_scratch = k_scratch.offset(k * kv_bytes)`. If `attn_output` buffer isn't sized for `k*kv_bytes + k*kv_bytes` (k+v scratch), this OOB-writes. The batch3/batch2 paths presumably validated this, but tiling reuses the same buffers per tile — fine if sized for max tile (3), but no assertion guards it.

2. **o_proj branch condition:** `else if !self.attn.o_proj.is_null()` — this catches n>=4 AND n==1 (falls through from the n==2/3 branches above). For n==1 this tiles into a single k==1 GEMV call, which is functionally correct but a pointless wrapper vs. the existing `else` fallback. Verify the `else` branch is now truly dead for NVFP4 o_proj or that n==1 still hits it.

3. **D2D copies not batched:** 3 separate `copy_d2d_async` per token in the scatter loop (9 for k==3 tile). These are on the same stream so serialized; minor perf but not a correctness issue.

4. **No synchronization between tiles:** Tiles reuse the same scratch buffers with no stream sync between iterations. Correct only if all kernels are enqueued on the same stream (appears so), but worth confirming no cross-stream dependencies exist.

**Risk:** Medium. Buffer sizing is the critical unverified assumption.

**Quality:** Tiling logic is clean; the k==1 remainder fallback duplicating the sequential path is acceptable but could call into a shared helper to avoid drift.

rsafier:
## GPU A/B — tiled-GEMV (this PR) vs wide-GEMM (#305), on current `main`

Fulfilling the "GPU numeric A/B recommended before merge" item. Since #305 (DFlash `commit_ctx`) landed on `main`, the `n≥4` o_proj/QKV branch this PR targets is now occupied by #305's `wide_verify_gemm` (single M=n GEMM) — so this is a **direct** head-to-head. Both variants built on the **identical current-`main` base**, differing only in the `n≥4` kernel.

**Model:** `AEON-7/Ornith-1.0-35B-AEON-…-NVFP4` (uniform NVFP4 attn) · GB10/sm_121 · temp 0 · 256-token decode-dominant · bf16 KV · `max-num-seqs 16`. Concurrency = simultaneous identical requests (batched decode). `per-seq` = server-reported `response_token/s`; `agg` = wall-clock aggregate.

| Concurrency | wide-GEMM (#305) per-seq | **tiled (this PR) per-seq** | speedup | wide agg | **tiled agg** |
|---|---|---|---|---|---|
| C=1  | 91.2 | 90.6 | 1.00× *(n<4 → shared path)* | 89.2 | 88.6 |
| C=4  | 15.8 | **26.7** | **1.69×** | 59.9 | **98.6** |
| C=8  | 9.9  | **15.1** | **1.53×** | 78.3 | **119.0** |
| C=16 | 5.9  | **8.8**  | **1.49×** | 91.2 | **134.7** |

**Result: the tiled batch3/batch2 GEMV beats the wide GEMM by ~1.5–1.7× at every batched size** — opposite to the "one weight read wins" intuition. The wide GEMM's M-tile padding (94% waste at M=4) outweighs its single-read advantage; note wide's C=4 aggregate (59.9) is actually *below* its C=1 (89.2) — it degrades under moderate batch, which the tiling fixes. Even at C=16 (≈ #305's γ=16 target) tiled wins.

### Resolution in this sync
This branch is now rebased onto current `main` with the `n≥4` conflict resolved **toward tiled** — i.e. it supersedes `wide_verify_gemm` for that branch (the function is removed; the `n==2/3` batch kernels are untouched). Justified by the table above.

### Caveats / remaining before merge
- **Single model / prompt / greedy / one run.** Directionally very strong (and since this is a MoE+SSM model where attention is only part of decode, the raw attention-kernel win is likely *larger* than the end-to-end 1.5×), but a multi-model sweep would firm up the exact factor.
- **Numeric identity:** the outstanding **tiled-vs-per-seq output-identity** check (this PR's own correctness argument) is still the must-do — temp-0 output diverges from the wide kernel (expected: different accumulation order; both coherent/correct), so it must be pinned against the `n==1` per-seq reference, not against wide.
- **DFlash interaction:** `wide_verify_gemm` was introduced by #305 for the DFlash γ=16 verify step, which flows through this same `n≥4` branch. Removing it routes DFlash verify through the tiled path too. That's functionally fine (tiled handles any n≥4) and faster here, but a `--dflash`-on run should confirm no verify-path regression before this lands, since #305's kernel was tuned for that specific call site.

rsafier:
## DFlash verify-path check — no regression from removing `wide_verify_gemm`

Closing the third caveat from the A/B above. `wide_verify_gemm` was introduced by #305 for the DFlash γ=16 verify step; this branch removes it and routes verify (an n=γ+1=16 batch) through the tiled path. Ran both binaries with `--dflash` (Ornith-1.0-35B-NVFP4 target + `z-lab/Qwen3.6-35B-A3B-DFlash` drafter, full DFlash correctness flags, γ=16):

| binary | verify step | decode | accept | crash |
|---|---|---|---|---|
| wide-GEMM (#305) | ~87.0 ms | 58.5 tok/s | 0–13% | none |
| **tiled (this PR)** | 86.8 ms | 59.8 tok/s | 0–7% | none |

- **Verify-step timing identical within noise (~87 ms).** #305's wide GEMM shows **no advantage even in its own DFlash-verify context** — the verify step is dominated by drafter-propose + SSM/MoE forward, so the o_proj/QKV kernel washes out there (unlike concurrent batched decode, where tiled is 1.5×).
- Accept rate is kernel-independent (identical math, ~5–13% — that low rate is the known 35B-A3B/drafter divergence, unrelated to this PR).
- Both coherent, **no illegal-address / crash** through the tiled verify batch.

**Conclusion:** removing `wide_verify_gemm` and using tiled everywhere is safe — faster for concurrent decode, equal for DFlash verify, no downside. This PR is synced onto current `main` with that resolution. Remaining pre-merge item is the tiled-vs-`n==1`-per-seq numeric-identity check.

rsafier:
## High-acceptance verify test — tiled still wins

The DFlash run above had low accept (~5%, adaptive suspended), so the verify path barely ran. To stress a **productive** n≥4 verify, ran **MTP forced past its throughput gate** (`ATLAS_MTP_GATE_FORCE=1`, `--num-drafts 3` → K=4, verify batch n=4) on `unsloth/Qwen3.6-35B-A3B-NVFP4` (the config that gets high spec-accept). Both binaries, identical flags:

| MTP K=4 (verify n=4, productive) | decode | mean accepted / 3 |
|---|---|---|
| wide-GEMM (#305) | 62.3 tok/s | 1.72 (~57%) |
| **tiled (this PR)** | 73.2–73.5 tok/s | 1.93 (~64%) |

**Tiled is ~1.18× faster end-to-end even with the verify running every step at high acceptance**, and accept is at parity (the 1.72↔1.93 spread is temp-0 numeric drift between the two accumulation kernels over a single run, not a systematic edge). Both coherent, no crash.

**Verify-path summary across regimes:**
- Concurrent batched decode (C=4/8/16): tiled **1.49–1.69×**.
- DFlash verify (n=16, low accept, verify-dominated-by-forward): tiled ≈ wide (no regression).
- MTP verify (n=4, ~60% accept, productive): tiled **~1.18×**.

Net: tiled is faster or equal in every regime tested — concurrent decode, low-accept DFlash verify, and high-accept MTP verify. Removing `wide_verify_gemm` in favor of tiled carries no downside on the evidence here. Remaining pre-merge item stays the tiled-vs-`n==1`-per-seq numeric-identity check.

rsafier:
## Second model (Holo-3.1) — the win is model-dependent

Ran the same n≥4 concurrent-decode sweep on **`Hcompany/Holo-3.1-35B-A3B-NVFP4`** (qwen3_5_moe). *(MTP-force is N/A here — Holo ships no MTP head: 0 MTP weight keys, `mtp_num_hidden_layers: None`.)*

| C | wide per-seq | tiled per-seq | speedup |
|---|---|---|---|
| 1  | 78.5 | 78.4 | 1.00× |
| 4  | 26.7 | 29.4 | **1.10×** |
| 8  | 17.7 | 18.5 | 1.05× |
| 16 | 10.4 | 10.5 | 1.01× |

**On Holo the tiled advantage is marginal (1.0–1.10×)** — not the 1.49–1.69× seen on Ornith. The cause is in the wide-GEMM baseline: on Ornith (qwen3_6_moe) wide degraded at moderate batch (C=4 aggregate *below* C=1) and tiling recovered it; on Holo (qwen3_5_moe) wide already scales cleanly (C=4 agg 104.6 vs C=1 76.6), leaving little for tiling to gain.

### Revised takeaway
The tiled kernel win is **model/shape-dependent, not universal**: large on Ornith, near-parity on Holo, and **never a regression** across everything tested (concurrent decode on 2 models, low-accept DFlash verify, high-accept MTP verify). So tiled is a safe drop-in — it helps some models substantially and costs nothing elsewhere — but the honest headline is "up to ~1.7×, ≥ parity everywhere," not a flat 1.5×. Whether that justifies carrying the tiled path over #305's wide GEMM is a maintainer call; the data says there's no downside to doing so.

rsafier:
## Tiebreaker: the win is the BF16→FP4 requant path, NOT model/shape

Ran the A/B on **`AEON-7/Qwen3.6-35B-A3B-heretic-NVFP4`** — same qwen3_6 arch and **byte-identical attention shape as Ornith** (16 heads / 2 KV / head_dim 256 / hidden 2048), but **native NVFP4-packed** attention on disk instead of Ornith's BF16 (which gets runtime-requantized to FP4 under `ATLAS_HOLO_FP4_PROJ_DECODE=1`).

| qwen3_6, identical shape | attn source | wide C=4 per-seq | tiled C=4 per-seq | speedup |
|---|---|---|---|---|
| Ornith-35B | BF16 → FP4 **requant** | **15.8** (pathological) | 26.7 | **1.69×** |
| heretic-35B | **native NVFP4** | 27.6 (healthy) | 29.5 | 1.07× |

Full heretic sweep: C=4 1.07×, C=8 1.06×, C=16 1.01× — i.e. **near-parity, like Holo**, not Ornith's 1.5–1.7×.

**Conclusion:** the earlier "model/shape-dependent" framing was wrong. Same shape, opposite result. The discriminator is the **attention weight source**:
- **Native NVFP4/FP8 attention (heretic, Holo, unsloth, etc.):** wide-GEMM is already fast → tiled ≈ parity.
- **BF16 attention runtime-requantized to FP4 (Ornith):** wide-GEMM is pathological (its fast pipelined `m128_v2` needs a transposed weight copy the requant path apparently doesn't build, so it falls to the slow M64 GEMM). Tiling — which uses the transpose-free batch3/batch2 GEMV — sidesteps that and wins ~1.7×.

**Implication for the PR:** the tiled kernel is only a meaningful win for **runtime-requantized-BF16 attention checkpoints** (a niche: e.g. the AEON "Ultimate" BF16 repacks). For the mainstream native-NVFP4/FP8 releases it's parity. That narrows the value proposition considerably — and suggests the *alternative* fix worth weighing is building the transposed NVFP4 copy during the BF16→FP4 requant (so `wide_verify_gemm`'s fast path works too), rather than carrying a second kernel. Either way, tiled remains ≥ parity everywhere, so it's a safe merge — but it's not the broad speedup the Ornith-only numbers suggested.

rsafier:
## Follow-up: keeping Ornith's attention in native BF16 doesn't help either

Since the Ornith win traced to the BF16→FP4 requant (not shape), the obvious question was: just **don't requant** — drop `ATLAS_HOLO_FP4_PROJ_DECODE` and serve Ornith's attention in its native BF16. Result: no better.

Ornith per-seq, all attention paths (Qwen3.6-35B-A3B, C = concurrent decode):

| Ornith attention path | C=4 | C=8 |
|---|---|---|
| native BF16 (no fp4_proj_decode) | **15.8** | 10.1 |
| FP4-requant → wide-GEMM (#305) | 15.8 | 9.9 |
| **FP4-requant → tiled (this PR)** | **26.7** | 15.1 |
| *heretic, native NVFP4 (reference)* | *27.6* | *17.7* |

**Native BF16 lands at exactly 15.8 — same as the pathological wide path.** BF16 attention has no fast batched kernel at n≥4 either; it falls to the per-token **seq-loop** (n full-weight GEMVs at 2 bytes/weight), which is just as slow as the transpose-less wide-GEMM *and* forfeits the FP4 bandwidth savings. So dropping the requant is a net loss, not a fix.

### Complete picture for the n≥4 attention path
- **Native NVFP4/FP8 attention** (heretic, Holo, unsloth, mainstream releases): wide already fast → tiled ≈ parity (1.0–1.1×).
- **BF16-repack attention runtime-requantized to FP4** (Ornith-class "Ultimate" repacks): only three options at n≥4 — BF16 seq-loop (15.8), requant+wide (15.8), or **requant+tiled (26.7)**. Tiling is the only one that's fast.

**Bottom line for merge:** tiled is ≥ parity everywhere tested (2 native-quant models, DFlash verify, MTP verify) and the **only** fast option for BF16-repack checkpoints at concurrency. It's a safe drop-in with a real (if niche) benefit. The equally-valid alternative — emit the transposed NVFP4 copy during the BF16→FP4 requant so `wide_verify_gemm`'s fast path fires (matching heretic's native 27.6) — would remove the need for a second kernel; worth weighing, but out of scope for this PR.

eisbaw:
GB10 datapoint (DGX Spark, sm_121a, CUDA 13.2, built from source), PR head as-is (conflicts with main):

`spark serve RedHatAI/Qwen3.6-35B-A3B-NVFP4 --port 8888 --lm-head-dtype nvfp4`

Aggregate decode at C=1/4/8/16/32 = 90.97 / 118.90 / 140.32 / 138.81 / 137.66 tok/s (256-tok non-stream, distinct sequences). Same-session main: 93.09 / 122.57 / 148.98 / 151.06 / 150.36. On this MoE the batched-decode cost sits in expert FFN + lm_head rather than the QKV/o_proj this tiles, and the old base drags absolute numbers — a dense-model (27B) A/B after a rebase would show this PR in its best light. Coherent at all levels.

tbraun96:
Closing this — main landed the idea, in a better form, and the measurements that justified it were against a baseline that no longer exists. Detail, because a bare "superseded" isn't reviewable:

### Main already routes small-M through a single-weight-read GEMV

`crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/qkv.rs:512`:

```rust
if m <= 4 && self.w4a16_gemv_batch4_k.0 != 0 {
    return ops::w4a16_gemv_batchm(...)   // ONE weight read for all m rows
```

and `:525` for `(5..=8)`. Landed in `22adb23ef` (#366).

This PR tiles into ≤3-row GEMVs instead (`multi_seq/qkv.rs:370`, `attn.rs:294` — `let k = (n - s).min(3);`), so it reads the projection weight **ceil(n/3)** times: 2 at n=4, 3 at n=8, **11 at n=32**. Main reads it once for every m≤8 and hands m>8 to the tile GEMM.

That cutoff is not arbitrary. Routing m=9..32 into a GEMV was measured at **−14.4% at C=16 and −29.4% at C=32** — the fallback was the faster path. The `m <= 8` guard at `qkv.rs:510` is that result written into the code. This PR routes *all* n≥4 into GEMV, which is the refuted direction.

### The 1.49–1.69× was real, against a baseline that's gone

At this branch's base `f1098db71`, `wide_verify_gemm` had **zero** batched-GEMV arms — pure tile GEMM, 94% padding at M=4. Tiling beat that comfortably. #366 then captured the same "one weight read" insight without the per-tile cost. Your own later comments walked the claim down to 1.0–1.10× and only on BF16→FP4 requant checkpoints — and #389 (`dd2ac46d5`, GPU-transpose at cold load) is the alternative you proposed there instead of carrying a second kernel. That landed.

Independent numbers from @eisbaw (2026-08-06), PR head vs same-session main at C=1/4/8/16/32:

| | C=1 | C=4 | C=8 | C=16 | C=32 |
|---|---|---|---|---|---|
| PR | 90.97 | 118.90 | 140.32 | 138.81 | **137.66** |
| main | 93.09 | 122.57 | 148.98 | 151.06 | **150.36** |

−8.5% at C=32, widening with concurrency — the signature of `ceil(n/3)` weight reads.

### Two structural problems

- The branch is rooted on a **reverted** commit: `b45af9881` is *"This reverts commit f1098db71"*, and `f1098db71` is this branch's merge-base.
- It deletes `wide_verify_gemm` outright. Main calls it from five sites and layers the fused-QKV twin, the `k64_n64` narrow-N twin and the small-M ladder on top of it (`qkv.rs:674`: *"n > 8 is REQUIRED, not an optimisation"*).

### One thing worth flagging for your other PRs

This branch conflicts with main, and a conflicting PR makes CI **silently skip** — so the green checks here were never real. Same is true of #332. Worth rebasing anything you want looked at.

The instinct was right and it's now in main via #366. Thanks for it.

PR #275, opened 2026-07-06
feat(deepseek-v4): decode-time compressed-KV append (CSA/HCA) + prefill→decode seed

SeedSource:
# DeepSeek-V4-Flash: decode-time compressed-KV append (CSA/HCA)

Completes the V4-Flash compressed-KV (DeepSeek Sparse Attention) port on GB10/SM121:
prefill persists compressed blocks (inc-1), decode reads them (inc-2), and **decode now
appends new compressed blocks every `ratio` steps** (inc-3) — so generated context is
retained past the 128-token raw sliding window instead of being lost at the window cliff.

## What this does

At each window boundary during decode, the compressor input (`normed` — the layer-input
RMSNorm output, the exact tensor prefill feeds `wkv`/`wgate`) is captured into a small
per-layer BF16 ring, then the **prefill compress pipeline is re-run over the ring** to
produce one FP8 pool block: `dense_gemm×2 → csa_compress → rms_norm → rope@W·ratio →
bf16→fp8 → pool[W]`. CSA (ratio 4) uses the 2×ratio overlap window (prev-window Ca +
current-window Cb); HCA (ratio 128) uses the single window. No new kernels — reuses the
prefill kernels; BF16 ring quantized to FP8 only at the pool write (matches prefill's
pipeline bit-for-bit).

Validated against the reference `inference/model.py` `Compressor.forward`:
- re-projecting the ring per boundary ≡ the reference's per-token `wkv` accumulate (wkv linear);
- rope position `W·ratio`, ape-per-window-position, and the Ca/Cb overlap all match;
- the boundary token **does** attend its own just-built block (reference read set
  `[0, (start_pos+1)//ratio)`), so the pool is published before the read.

**Prefill→decode seed (the key correctness piece):** the reference seeds compressor state
at `start_pos==0` (`kv_state[:ratio] = last full window` → next window's Ca; trailing
partial window → current-window slots). Prefill now mirrors this by seeding `prev_win` +
the ring's leading slots from `normed`, ungated so HCA layers (whose prefill compress is
skipped when prompt < one window) are covered too. Without the seed the first decode block's
overlap-Ca is missing → a wrong block attended for the rest of the sequence.

## Measured (fp8 EP=2, eager, 550-token greedy gate, transistor prompt)

| build | change | coherence onset |
|---|---|---|
| b8 | inc-2 (frozen pool, no decode append) — control | ~gen 100 |
| b9 | inc-3, **no** prefill seed (first Ca masked) | ~gen 40 (regressed) |
| **b10** | **inc-3 + prefill seed** | **~gen 150** |
| b11 | b10 + eps-Sinkhorn "fix" (dropped hc_pre's extra col-proj) | ~gen 90 (regressed — reverted) |

inc-3 (b10) beats the frozen-pool control and survives the 128-token cliff: appends fire
clean across the run (1551 appends, 0 NaN, fp8-faithful cos 0.9996), CSA seeded
(`prev_valid=true` first window), HCA window-0 built from the seeded ring.

## Known limitation (documented, not introduced here)

Coherence still degrades into a repetition loop at ~gen 150. This is a **pre-existing base
degradation**, not an inc-3 artifact (the frozen-pool control degrades *earlier*, ~100).
Audited and cleared both known suspects:
- **item-5 CSA indexer**: provably inert < `index_topk=512` (reference indexer returns a
  selection *gather mask*, not attention weights; below 512 it selects all blocks). Can't
  affect ~150.
- **mHC eps-Sinkhorn**: A/B-falsified — matching the reference's eps-ending Sinkhorn
  (dropping `hc_pre`'s extra no-eps column projection) *regressed* onset ~150→~90, so the
  extra projection compensates for another mHC deviation and is kept (A/B note left at the
  code site). The prefill mHC `norm=inf`/`NaN` diagnostic is unchanged by it and is
  likewise pre-existing (present in the frozen-pool control).

The residual ~150 ceiling is tracked in a separate lane; the mechanism is not yet isolated.

## Before merge

Debug instrumentation has been stripped (commit `chore(deepseek-v4): strip decode-path
debug instrumentation before merge`): the `V4ARM` per-arm diagnostic + `diag_layer_idx`
kernel-param threading in `mla_paged_decode_fp8.cu` (+ op wrapper + call site), the
`v4_*_dump` host-readback functions and their call sites, and the `ATLAS_V4_POOL_DIFF`
fidelity block. All were env- or layer-gated (no default-runtime effect). The pervasive
`diag_norm`/`diag_norm_f32` norm-logging helpers are intentionally kept (shared across the
whole attention path, out of scope). `cargo check` + `clippy` (deny-all) + `fmt` clean.

Follow-up (not blocking): a teacher-forced Rust golden for strict per-block compressed-KV
correctness.

## Task-level eval readout (pre-PR gate, 2026-07-05)

Ran the standard futures-reasoning core-8 packet against b10 (Atlas fp8-KV, EP=2, eager, greedy temp0, max_tokens
4096, diag stripped, isolated n3/n4). Full run: `model-evals/runs/20260705__deepseek-v4-flash__atlas-fp8kv-b10__gb10-ep2/`.

**Verdict: FAIL — 15/40** (bar 35/40). This quantifies the residual ~150-token ceiling at the task level: the model's
*priors are disciplined* (every top-line call was the safe one — correct verdicts on 4/8, correct event-risk call), but
the ceiling prevents any complete/structured output (**0/8 produced the required 7 sections**) and on data-heavy Qs the
degradation **corrupts the provided numbers**:

| Q | critical? | score | note |
|---|---|---|---|
| q1 inventory | | 2 | inventory correct; plan loops |
| q2 ORB trap | | 3 | correct "not valid" + reason, then loops |
| q3 risk sizing | ✓ | 2 | correct "1 contract" number, but 26-contract mis-math + P/L drift + 3204-tok loop |
| q4 mean-reversion | | 3 | correct structure-based "no short", thin |
| q5 news-event | ✓ | 3 | **PASS** — "no trade before release" |
| q6 cross-market | | 1 | immediate loop, no analysis |
| q7 missing-data | ✓ | 0 | **AUTO-FAIL** — fabricates "above VWAP/RTH" from unknown levels; no NO-TRADE |
| q8 revenge-trade | ✓ | 1 | misreads −$1,600→−$1,000, skips breach math |

Critical gate: event-risk PASS; risk-sizing reasoning broken; missing-data + revenge FAIL. Wall: ~8.9 min total,
mean 67 s/Q; eager C=1 tok/s 2.2–10.7 (cleanest sustained decode q3 = 10.7 tok/s). Serve's content-loop watchdog
(rollback/re-steer) halted 7/8 at 49–276 tok, so raw token counts *understate* how fast coherence dies.

**Takeaway for merge:** inc-3 is functionally correct and a net win, but the served model is **not usable for the
futures-research role** until the ~150-token base degradation (a model/quant property, separate from this compressed-KV
port) is resolved. The port ships; the ceiling is documented as the gating limitation with a graded readout, not just
token counts. Re-run the frozen packet once the ceiling lifts.

### Short-form supplementary annotation (2026-07-05)

To test whether the disciplined priors are usable *inside* the ceiling, re-ran the same 8 scenarios with an answer-first
≤100-token contract (verdict + one justification line, no 7-section protocol, max_tokens 512). SUPPLEMENTARY — does NOT
change the frozen 15/40 FAIL. Run: `model-evals/runs/20260705__…-b10-shortform__…/`.

**Priors usable short-form = PARTIALLY.** Qualitative/no-trade discipline lands cleanly before degradation (q2/q4/q5
correct), and critically **q7 missing-data flips from the long-form fabrication (auto-fail) to a correct clean refusal** —
fitting the task inside the ceiling recovers the safe prior. The two *numeric* critical gates (q3 risk-sizing, q8 revenge)
still fail short-form (confuses contracts vs stop distance; fabricates a "$5,000 margin").

**Provenance check (07):** the numeric failure is **on our serve, not the model.** The same base model
(`deepseek-ai/deepseek-v4-flash` via NVIDIA NIM) reasons both gates **correctly** — q3: remaining room $1,100 ÷ ~$575/ct
→ ~1 contract; q8: $500 risk → $2,100 > $2,000 → "not valid" (clean). Our b10 gets both wrong. So the numeric-value
corruption is introduced by **our stack (Atlas engine + NVFP4 requant)** — an earlier draft's "independent model reasoning
defect" claim was falsified and retracted. Precision confound noted: the NIM PASS localises the fault to our side but
cannot yet isolate the mechanism — **separate lane, under investigation**. An
in-window wedge (07 lane doc) **empirically clears the compressed-KV port this PR ships**: q3 fails with a 75-token
fully-in-window prompt (zero compression dependency), so the numeric fault is outside inc-1/2/3 (the compressed-KV port
this PR ships); the mechanism is under investigation. Eager C=1 tok/s 2.5–4.2 (short gens).

---


SeedSource:
## Merge gate — codeowner review (SeedSource)

**Gate legs run:**
- `webserver_ok`: **PASS 10/10** (GLM-5.2 via NIM → OpenCode wrote a stdlib server, `/ping` 10/10 returned exactly `pong`).
- CI: all required checks green (`cargo test --workspace`, `clippy --tests`, `fmt --check`, `cargo test --features metal`, `≤500 LoC`, `cargo deny`, SPDX, typos, mdBook). Branch `CLEAN`/up-to-date.
- Feature: the decode-time compressed-KV append (CSA/HCA) + prefill→decode seed is validated in the PR body (append fidelity cos 0.9996, coherence onset > frozen-pool control).

**KL/tok-s leg — waived, with reason:** this PR is DeepSeek-V4-Flash-only and its decode/prefill dispatch is `model_type`-gated (`is_v4_flash`), so it is unreachable from the flagship Qwen3.6-FP8 serve path. The KL-vs-`main` regression check on the flagship is therefore ≈0 by construction, not by measurement — it can add no signal a model-gated V4 change could fail. Documenting the waiver as a decision rather than an omission.

**Known limitation (unchanged, documented):** the ~150-token coherence ceiling and the q3/q8 numeric fault are separate from this compressed-KV port (the port is empirically clear of them). The numeric fault has since been root-caused to the sliding-layer RoPE θ and is fixed in a stacked follow-up.

Merging squash as codeowner.

PR #276, opened 2026-07-06
fix(deepseek-v4): main θ=10000 rope on sliding_attention layers (was θ=160000)

SeedSource:
# fix(deepseek-v4): main θ=10000 rope on sliding_attention layers (was θ=160000)

Follow-up to **#275** (the V4-Flash compressed-KV port, now merged) — the separate rope fix promised there.
This is the root cause of the V4-Flash numeric-reasoning corruption; the compressed-KV port was empirically clear of it.

## The bug

DeepSeek-V4 uses **two** RoPE bases, per layer type:
- `sliding_attention` layers (`compress_ratios[L]==0`) → **main** rope, `rope_theta=10000`, **no YaRN**;
- `compressed_sparse_attention` / `heavily_compressed_attention` → **compress** rope, `compress_rope_theta=160000` + YaRN.

Reference `modular_deepseek_v4.py` config: *"yarn is applied ONLY to layers with a compressor (CSA/HCA); … main = pure."*
Attention selects `rope_layer_type = "main" if layer_type=="sliding_attention" else "compress"`.

Atlas builds a **single** shared `yarn_inv_freq` (θ=160000 + YaRN, `ensure_yarn_inv_freq`) and applies it to the Q/K
rope **and** the eq.26 output de-rotation on **every** layer. So the ratio-0 sliding layers (indices 0, 1, 42 in this
checkpoint) rotate with θ=160000 where they must use plain θ=10000.

The error is **layer-type-dependent and position-growing** (a base-θ difference, not YaRN interpolation) — which is why
it was invisible to short-context probes:

| decode position | max per-dim angle error (θ=160000 vs θ=10000) | \|Δcos\| |
|---|---|---|
| 150 | 14.5 rad | — |
| 240 | 23.2 rad | — |
| 330 | 31.9 rad | up to 1.85 (of 2.0 max) |

## The fix

Add `main_inv_freq` (plain θ=10000, mscale=1) to `MlaWeights`; select it over `yarn_inv_freq` when
`mla.compressor.is_none()` (== sliding layer) at all four raw-arm rope sites — **rope-in and the eq.26 de-rotation, in
both prefill (`cache_skip_v4.rs`) and single-token decode (`attention_forward_v4.rs`)**. The de-rotation must match the
rope-in inv_freq or the output is scrambled. CSA/HCA layers are unchanged (still θ=160000 + YaRN). Non-V4 (mistral)
loader sets `main_inv_freq = yarn_inv_freq`. 6 files, +77/−8.

## Measured (fp8-KV EP=2, greedy, RoCE-verified; futures risk-sizing gates vs NVIDIA NIM reference)

| gate | before (θ=160000 all layers) | after (this fix) |
|---|---|---|
| q3 (max size) | "10 points / $10,000" — **wrong** | **"1 contract … take the trade"** — correct (matches NIM) |
| q8 (revenge) | fabricates a "$5,000 margin" not in the scenario | recalls the real $1,600 / $2,000; **verdict still "valid" (should be not-valid)** |
| coherence onset (280-tok non-arithmetic probe) | ~150 tok | **~150 tok (unchanged)** |

**Scope of the claim:** this fixes the sliding-layer **rope** fault — q3 flips to the correct answer, q8 stops
fabricating numbers. It is **not** a full task fix: q8's multi-step breach reasoning still errs, and the ~150-token
coherence ceiling is **unmoved** by this change (measured above) — a separate, still-open issue, not the rope.

## Coverage gap (batched / MTP)

This PR splits the rope only on the **single-token decode + prefill** paths. The **batched decode**
(`trait_impl/multi_seq/mla.rs`) and **MTP verify** paths still apply `yarn_inv_freq` (θ=160000) to sliding layers.

**Observable consequence:** single-token (`--max-batch-size 1`) output on layers 0/1/42 is now correct; **batched
decode and MTP-verify output will differ on those layers until they get the same `compressor.is_none()` split.** A
follow-up will cover them; flagged so the partial coverage is explicit, not silent.

`cargo check` + `clippy` (deny-all) + `fmt` clean.

### DIFF
--- a/crates/spark-model/src/layers/qwen3_attention/decode/attention_forward_v4.rs
+++ b/crates/spark-model/src/layers/qwen3_attention/decode/attention_forward_v4.rs
@@ -511,8 +511,18 @@ impl Qwen3AttentionLayer {
                 1,
                 mla_rope,
                 mla_rope,
-                mla.yarn_inv_freq,
-                super::super::helpers::yarn_rope_mscale(ctx.config),
+                // Sliding layers (compressor==None) = reference "main" rope:
+                // plain θ=10000, mscale=1 (no yarn). CSA/HCA keep θ=160000 yarn.
+                if mla.compressor.is_none() {
+                    mla.main_inv_freq
+                } else {
+                    mla.yarn_inv_freq
+                },
+                if mla.compressor.is_none() {
+                    1.0f32
+                } else {
+                    super::super::helpers::yarn_rope_mscale(ctx.config)
+                },
                 stream,
             )
         })?;
@@ -677,8 +687,19 @@ impl Qwen3AttentionLayer {
                 0, // no KV heads — de-rotate the query/output heads only
                 mla_rope,
                 mla_rope,
-                mla.yarn_inv_freq,
-                super::super::helpers::yarn_rope_mscale(ctx.config),
+                // MUST match the Q/K rope inv_freq for this layer type (rope-in
+                // == de-rotate-out), else the output is scrambled. Sliding =
+                // main θ=10000; CSA/HCA = θ=160000 yarn.
+                if mla.compressor.is_none() {
+                    mla.main_inv_freq
+                } else {
+                    mla.yarn_inv_freq
+                },
+                if mla.compressor.is_none() {
+                    1.0f32
+                } else {
+                    super::super::helpers::yarn_rope_mscale(ctx.config)
+                },
                 stream,
             )?;
             ops::mla_q_rope_writeback_batched(
--- a/crates/spark-model/src/layers/qwen3_attention/prefill/cache_skip_v4.rs
+++ b/crates/spark-model/src/layers/qwen3_attention/prefill/cache_skip_v4.rs
@@ -332,8 +332,18 @@ impl Qwen3AttentionLayer {
             nkv,
             rope,
             rope,
-            mla.yarn_inv_freq,
-            super::super::helpers::yarn_rope_mscale(ctx.config),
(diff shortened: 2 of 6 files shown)


SeedSource:
## Merge gate — codeowner review (SeedSource)

Follow-up to #275; root-causes and fixes the V4-Flash numeric fault.

**Gate:**
- CI: required checks (`cargo test --workspace`, `clippy --tests`, `fmt --check`, `cargo test --features metal`, `≤500 LoC`, `cargo deny`, SPDX, typos, mdBook) — merging on green.
- `webserver_ok` path validated (GLM-5.2/NIM → server `/ping` 10/10 `pong`) at #275; unaffected here.
- Correctness: the rope θ selection is verified end-to-end on a live fp8-KV EP=2 serve — q3 flips from "10 points/$10,000" to the correct "1 contract" (matches the NIM DeepSeek-V4-Flash reference), with the before/after and the angle-delta magnitudes in the PR body.

**KL/tok-s leg — waived, same reason as #275:** V4-Flash-only, `model_type`-gated (`is_v4_flash`); unreachable from the Qwen3.6-FP8 flagship path, so KL-vs-`main` is ≈0 by construction. Documented as a decision.

**Scope honesty:** fixes the sliding-layer rope numeric fault (q3 correct, q8 de-fabricated). NOT a full task fix — q8's breach reasoning still errs and the ~150-token coherence ceiling is measured unchanged (separate open issue). Batched/MTP rope paths not yet split — observable consequence noted in the body.

Merging squash as codeowner.

PR #281, opened 2026-07-09
Nemotron-H-Puzzle support and significant Mamba optimizations

AzeezIsh:
## Summary

Bring-up of [`nvidia/NVIDIA-Nemotron-Labs-3-Puzzle-75B-A9B-NVFP4`](https://huggingface.co/nvidia/NVIDIA-Nemotron-Labs-3-Puzzle-75B-A9B-NVFP4) on Atlas (GB10):

1. **Config/dispatch** for `nemotron_h_puzzle`: derive `num_hidden_layers` from `layers_block_type` (JSON null → 88), map block types to layer schedule, and populate per-layer `moe_intermediate_size` / `num_experts_per_tok` from `block_configs` (channel pruning 1280–2688, top-k 4–22).
2. **Runtime MoE**: store layer-local inter + top-k on `NemotronMoeLayer`; loaders use `moe_intermediate_size_for(layer)`.
3. **Kernels**: reuse Super 120B NVFP4 PTX via `model_types` alias (`nemotron_h_puzzle`, H=4096); register `kernels/gb10/nemotron-labs-3-puzzle-75b-a9b/`.
4. **Coherence fix**: `mamba2_ssm_decode/prefill` only summed 4 warps (assumed `state_size==128`). Puzzle uses **`ssm_state_size=96`**, so `smem_warp[3]` was uninitialized garbage → token loops. Now sum `ceil(state_size/32)` active warps.

## Validation (dgx2 / spark-43fa)

- Serves on `:8080`, ~35 tok/s, TTFT ~450–630 ms
- Completions: `2+2` → `4`, France → `Paris`, Tokyo / 60 km/h haiku suite **3/3**
- `scripts/dev/coherence_test.py`: **9/10** (soft multi-turn phrasing fail)
- `scripts/test_coherence.py`: **54 pass / 8 fail** (tools weak under bare_json; soft greedy flake)
- `ATLAS_NEMO_DUMP`: 0 NaN across 88 layers; logits top-1 mass **0.82**; two-run logit cosine **0.999998**

## Local CI (pre-PR)

| Check | Result |
|-------|--------|
| `cargo fmt --all -- --check` | pass |
| `ATLAS_SKIP_BUILD=1 cargo clippy --workspace --tests` | pass |
| `cargo test --workspace` | pass (incl. `test_parse_nemotron_h_puzzle_config`) |
| `typos` | pass |
| File-size (changed `.rs`) | all ≤500 or pre-existing allowlist (`config.rs`, `tests.rs`) |
| SPDX headers | present on all changed sources |

License-eye Docker image is x86-only here (exec format error); headers verified manually.

## Test plan

- [x] Unit: puzzle config parse (heterogeneous MoE dims)
- [x] Local CI: fmt / clippy / test / typos
- [x] GPU serve on GB10 + factual coherence probes
- [x] NEMO dump residual/logit health
- [x] dgx1 second-box reproduction (coherence 10/10 + 54/8/3, logit/KL health, local CI 1664/0, ~33 tok/s)
- [x] Head-to-head vs vLLM 0.23.1 same model+box (Atlas ~1.7x decode; vLLM lower TTFT)
- [ ] GitHub Actions CI green on this PR
- [~] Full HF dequant KL via `bench/nemotron_layer_diff.py` - not feasible for 75B on a 121 GB box; covered by API logprob health above

---

## Validation (dgx1 / spark-256a) - second-box reproduction 2026-07-14

Independently reproduced the dgx2 bring-up on the second GB10 box (121 GB unified memory) from a clean kernel rebuild of this branch (`05ed12c5`). Everything reproduces; no regressions vs dgx2.

**Build**: `ATLAS_TARGET_HW=gb10 ATLAS_TARGET_MODEL=nemotron-labs-3-puzzle-75b-a9b ATLAS_TARGET_QUANT='*'` after `cargo clean -p atlas-kernels`. Kernel audit at startup: 146 PTX modules embedded, target `(sm_121, nemotron-labs-3-puzzle-75b-a9b, nvfp4)`, quant compat kernel=nvfp4/model=nvfp4 OK. The 28 MISSING kernel lookups are all for architectures this model does not use (MLA `mla_*`, `hyper_connection`, `w4a16` alternates, compressed-attn) and fall back harmlessly; the puzzle path (`mamba2_ssm`, `nemotron_moe_prefill`, `moe`, `moe_expert_gemv`, `inferspark_prefill_paged_nvfp4_batched`, `paged_decode_fp8`, `fp8_gemm_t_blockscaled`) all resolve to `used`.

**Serve** (recipe below): loads in ~60 s, config confirmed 88 layers / 8 attention / 40 SSM / 512 experts, 0 NaN.

### Coherence

| Suite | dgx1 result | dgx2 baseline |
|-------|-------------|---------------|
| `scripts/dev/coherence_test.py` | **10 / 10** | 9 / 10 |
| `scripts/test_coherence.py` | **54 pass / 8 fail / 3 skip** | 54 / 8 |

The 8 failures are all the known bare_json tool-calling weakness (tool reliability 1/10, `tool_choice required/specific`, Anthropic `tool_use`); factual / streaming / thinking / multi-turn / unicode / API-format checks all pass. dgx1 scored 10/10 on the dev suite (the dgx2 soft multi-turn phrasing miss passed here).

### Logit / KL health (API logprobs, greedy, 4 prompts)

- NaN/inf in any logprob: **none**
- mean top-1 probability mass: **0.795** (dgx2 NEMO-dump reported 0.82)
- two-run greedy token agreement: **32 / 32 = 1.000** (fully deterministic)
- two-run top-20 logit cosine: **0.9977**; mean symmetric KL: **6.4e-3** (run-to-run float-reduction noise only)

Note: the full HF-dequant layer-diff (`bench/nemotron_layer_diff.py`) is not runnable for the 75B here (an FP reference needs ~150 GB, exceeds the 121 GB box, and that script's layer PATTERN is hardcoded for the 52-layer Nano). The API logprob health above covers the same signals (NaN-free, top-1 mass, run-to-run determinism/KL).

### Local CI (this box)

`cargo fmt --all -- --check` pass; `cargo test --workspace --locked` = **1664 passed / 0 failed**; `test_parse_nemotron_h_puzzle_config` ok.

### Performance (bench/bench-quick.py, 5 runs, 128 max tokens, thinking default)

| Prompt | Input tok | TTFT | Decode |
|--------|-----------|------|--------|
| short  | ~23   | 420 ms  | 33.4 tok/s |
| medium | ~220  | 1129 ms | 33.1 tok/s |
| long   | ~1017 | 3384 ms | 32.7 tok/s |

~33 tok/s sustained decode, in line with the dgx2 ~35 tok/s.

### vs vLLM (apples-to-apples, same model + box)

vLLM 0.23.1 **does** run this model (sparkrun-eugr-vllm image; `NemotronHPuzzleForCausalLM` registered, heterogeneous `block_configs` handled, `quantization=modelopt_mixed`, NVFP4 MoE via auto-selected `FLASHINFER_CUTLASS`). Output is coherent - the SM121 NVFP4 garbage seen with older vLLM does not reproduce here. Both engines measured with one identical client-side streaming harness (thinking off, temp 0, 256 max tokens, 5 runs) on the same GB10 box:

| Metric (medium / long prompt) | Atlas | vLLM 0.23.1 |
|-------------------------------|-------|-------------|
| Decode | **~33-34 tok/s** | ~20 tok/s |
| TTFT @ ~1k-token prompt | 3433 ms | **418 ms** |

Real tradeoff, both directions honest: **Atlas is ~1.7x faster on decode throughput**, while **vLLM has ~4-8x lower TTFT** (its FlashInfer prefill + torch.compile + CUDA graphs win first-token latency; Atlas's hybrid-Mamba2 prefill is the weaker leg here). vLLM startup was ~6 min (278 s weight load + graph capture) vs Atlas ~60 s. vLLM brought up transiently for the measurement, then torn down.

### vs llama.cpp

llama.cpp **cannot run this model at all**: no GGUF exists, `convert_hf_to_gguf.py` has no Nemotron-H converter class, the `nemotron_h_puzzle` arch is not registered (only `nemotron_h` / `nemotron_h_moe`), and the checkpoint is NVFP4-only (no BF16 to convert). Same-box reference on `Qwen3.6-27B-Q4_K_M` (llama.cpp build b9521, `-ngl 99`) for scale: prefill 78.6 tok/s, decode 11.5 tok/s (different, smaller, dense model - not a direct comparison).

### Deployment (tested on dgx1, GB10 121 GB unified)

```bash
# unified memory: keep --gpu-memory-utilization <= 0.70 to leave host headroom
target/release/spark serve nvidia/NVIDIA-Nemotron-Labs-3-Puzzle-75B-A9B-NVFP4 \
  --port 8888 --gpu-ordinal 0 \
  --gpu-memory-utilization 0.70 \
  --max-seq-len 16384 \
  --kv-cache-dtype fp8
```


SeedSource:

### Triage
- Rationale: Adds per-layer MoE config for Puzzle model; affects config parsing and nvfp4 MoE kernels.
- Serving-path files: `crates/atlas-core/src/config/dispatch.rs, crates/spark-model/src/factory.rs, crates/spark-model/src/layers/nemotron_moe.rs, crates/spark-model/src/layers/nemotron_moe/decode_helpers.rs, crates/spark-model/src/weight_loader/nemotron.rs, crates/spark-model/src/weight_map/nemotron.rs`

### Serve gate

### GLM-5.2 code review
**Correctness concerns:**
1. **`skip_deserializing`/`skip_serializing` on new Vec fields:** These fields are never populated via serde. They're only filled by `apply_nemotron_puzzle_config`. If a config is round-tripped or cloned before dispatch logic runs, these are silently empty. Acceptable given the dispatch path, but fragile.
2. **`block_configs` indexing assumes 1:1 with `layers_block_type`:** The code uses `enumerate().take(n)` on `block_configs` and indexes directly into per-layer arrays. If `block_configs` is a *template* applied with a repeat pattern (common in HF hybrid configs), this breaks. The test only covers the trivial case where lengths match. Verify against the real 75B checkpoint's JSON structure.
3. **`raw_mut.clone()` passed to `apply_nemotron_puzzle_config`:** The JSON rewrites (inserting `num_hidden_layers`, scalar fallbacks) are now in `raw_mut`, but `finalize_config` also receives `raw_mut` instead of the original `raw`. If `finalize_config` has assertions about original JSON fields, this could cause subtle divergences.
4. **No validation that `moe_intermediate_sizes.len() == num_hidden_layers`:** If `block_configs` is shorter than `layers_block_type`, non-MoE entries silently get 0 (falls back to scalar). Length mismatch should be an error, not silent fallback.

**Risk:** Medium. Heterogeneous MoE sizing affects buffer allocation, expert weight strides, and routing. If downstream CUDA kernels don't call `moe_intermediate_size_for(layer)` and instead use the scalar `max`, you'll over-allocate but still compute correctly. The reverse (using scalar when per-layer is smaller) would be a silent correctness bug. Audit all kernel call sites.

**Quality:** The `max_moe_intermediate_size()` helper is good for buffer sizing. Test coverage is minimal (4 layers, no mismatched lengths, no missing `block_configs`). Add a test where `block_configs` is absent or shorter than `layers_block_type`.

AzeezIsh:
Reproduced the bring-up on the second GB10 box (dgx1 / spark-256a, 121 GB unified) from a clean kernel rebuild of this branch. Summary in the updated PR description; highlights:

- Kernel audit clean: 146 modules, target (sm_121, nemotron-labs-3-puzzle-75b-a9b, nvfp4), quant compat OK. The 28 MISSING lookups are all unused arches (MLA, hyper_connection, w4a16 alternates); the puzzle path resolves fully.
- coherence_test.py 10/10 (dgx2 was 9/10); test_coherence.py 54/8/3, an exact match to dgx2 (the 8 fails are the known bare_json tool-calling weakness).
- Logit health: 0 NaN, top-1 mass 0.795, greedy token agreement 32/32, two-run top-20 logit cosine 0.9977, sym-KL 6.4e-3.
- Local CI green here: fmt pass, cargo test --workspace 1664/0, test_parse_nemotron_h_puzzle_config ok.
- Throughput ~33 tok/s decode, TTFT 420 ms (short) to 3.4 s (1k in). In line with dgx2 ~35 tok/s.

On llama.cpp: it cannot run this model at all (no GGUF, no Nemotron-H converter in convert_hf_to_gguf.py, nemotron_h_puzzle arch unregistered, NVFP4-only checkpoint). Same-box reference on Qwen3.6-27B-Q4_K_M was 11.5 tok/s decode / 78.6 tok/s prefill, so Atlas serves the larger 75B-A9B hybrid at roughly 2.9x that decode rate (different models, but a useful floor).

The state=96 warp-reduce fix is the load-bearing change: without it smem_warp[3] is garbage and decode loops. Looks good to me on both boxes.

AzeezIsh:
Follow-up: ran vLLM head-to-head since llama.cpp cannot load this arch. vLLM 0.23.1 (sparkrun-eugr-vllm image) DOES run it: NemotronHPuzzleForCausalLM is registered, it handles the heterogeneous block_configs, loads as modelopt_mixed, and auto-selects the FLASHINFER_CUTLASS NVFP4 MoE backend. Output is coherent (2+2=4, Paris, clean sentences) - the SM121 NVFP4 garbage we saw with the older v21-era vLLM does not reproduce on this build.

Measured both with one identical client-side streaming harness (thinking off, temp 0, 256 max tokens, 5 runs) on the same dgx1 box:

- Decode (medium/long prompts): Atlas ~33-34 tok/s vs vLLM ~20 tok/s -> Atlas ~1.7x
- TTFT at ~1k-token prompt: Atlas 3433 ms vs vLLM 418 ms -> vLLM ~8x lower
- Startup: Atlas ~60 s vs vLLM ~6 min (278 s weight load + torch.compile + graph capture)

So it is a real two-sided tradeoff, not a clean sweep: Atlas wins decode throughput, vLLM wins first-token latency (its FlashInfer prefill + torch.compile + CUDA graphs are strong on prefill; our hybrid-Mamba2 prefill is the weaker leg and the obvious next lever). Full table is in the updated PR description. vLLM was brought up only for the measurement and torn down after.

AzeezIsh:
Follow-up on the prefill/TTFT gap flagged in the vLLM head-to-head (Atlas 3433 ms vs vLLM 418 ms at a 1k-token prompt). Profiled it on dgx1 and got TTFT down to 798 ms (4.3x) with decode unchanged. NOTE: the patch is NOT pushed to this branch yet - posting the findings here so they are not lost, happy to land it as a separate perf PR stacked on this one.

Root cause is that the fast prefill paths are wired for the qwen models and were never wired for nemotron:

- `NemotronMamba2Layer::set_prefill_weights` exists (docstring: "est. 3-4x TTFT improvement for SSM layers") and had ZERO callers, so all 40 SSM layers ran the base `w4a16_gemm` instead of `w4a16_gemm_t` (FP8 MMA, N128/K32, cp.async).
- `NemotronMoeLayer::prepare_prefill_weights` also had ZERO callers, and its shared-expert transpose was gated behind the same `moe_latent_size == 0` guard as the 512-routed-expert transpose (the latter genuinely would not fit; the shared expert is only 2 small matrices per layer).
- `mamba2_ssm_prefill_persistent` had complete Rust plumbing (dispatch, launcher, shared-mem calc) but the .cu kernel was never written, so `try_kernel` returned 0 and it silently fell back to the global-memory recurrence forever.
- Separately: `dense_gemm_bf16` is a SCALAR 16x16 kernel with no MMA, and it was 34% of all prefill GPU time (603 ms). The nemotron MoE layer used it for gate / fc1_latent / fc2_latent. Swapping to the existing `dense_gemm_bf16_pipelined` (mma + cp.async, documented "~40x on large-M, cosine=1.0") took it to 23.6 ms, a 26x drop.

Measured, each step coherence-gated:

| step | TTFT @1k |
|---|---|
| baseline | 3433 ms |
| write missing `mamba2_ssm_prefill_persistent` kernel | 3163 ms |
| coalesce `w4a16_gemm` B-tile load | 2499 ms |
| coalesce MoE grouped-GEMM B-tile load | 2418 ms |
| wire SSM `set_prefill_weights` | 1780 ms |
| scalar -> pipelined tensor-core dense GEMM | 1215 ms |
| un-gate shared-expert transposed weights | 963 ms |
| rewrite persistent SSM scan (thread-per-head_dim, 4 threads/row) | 798 ms |

Gates: decode 33.4-33.5 tok/s (baseline 33.4-34.3, no regression); `test_coherence.py` 54 pass / 8 fail / 3 skip, an exact match to the pre-change baseline; `coherence_test.py` 9/10 where the one failure is the known stochastic multi-turn test (it omits temperature so runs at temp=1.0 with a 32-token cap; deterministic recall with thinking off is 5/5).

vs vLLM now: short prompt 230 ms (vLLM 305 - we win), 1k prompt 803 ms (vLLM 418 - still 1.9x). Not parity yet. nsys says what is left: MoE grouped GEMM 266 ms running at ~58% of its bandwidth floor (needs cp.async double-buffering), attention QKV/O still on the base `w4a16_gemm` (~30 ms), and more SSM parallelism. Realistic floor with the current hand-written kernels is ~650 ms; closing the last gap probably needs CUTLASS-class MoE kernels, which is what vLLM leans on.

AzeezIsh:
TTFT round 2. The prefill gap vs vLLM is now down to 1.46x at a 1k prompt, and we BEAT vLLM by 2x on short prompts. Still NOT pushed to this branch - posting so the findings and the dead ends are on record.

TTFT @1k: 3433 -> 606 ms (5.7x). Decode 34.3-34.5 tok/s (baseline 33.4, so it went UP). test_coherence.py 61 pass / 5 fail / 3 skip vs the 54/8/3 baseline (better; the 5 are the known bare_json tool-calling weakness).

Client-side, same harness both engines:

| prompt | Atlas before | Atlas now | vLLM |
|---|---|---|---|
| short | 425 ms | 158 ms | 305 ms  <- we win 2x |
| medium | 1145 ms | 267 ms | 190 ms |
| 1k | 3433 ms | 610 ms | 418 ms |

Root cause, and it is the same shape throughout: a whole family of fast prefill paths was wired for the qwen models and never for nemotron.

- `NemotronMamba2Layer::set_prefill_weights` - existed, docstring "est 3-4x", ZERO callers.
- `NemotronMoeLayer::prepare_prefill_weights` - ZERO callers, and its cheap shared-expert transpose was gated behind the same memory guard as the 512-routed-expert transpose that genuinely would OOM.
- `Qwen3AttentionLayer::set_prefill_weights` - never called for the 8 nemotron attention layers.
- `mamba2_ssm_prefill_persistent` - Rust dispatch, launcher and shared-mem calc all present; the .cu was NEVER WRITTEN, so try_kernel returned 0 and it silently fell back forever.
- `dense_gemm_bf16` is a SCALAR 16x16 kernel with no MMA at all, and it was 34% of ALL prefill GPU time (603 ms). The pipelined tensor-core drop-in was already in the tree: 603 -> 23.6 ms.
- The NVFP4 dequant LUT was `__constant__` indexed by a DIVERGENT nibble. Constant memory only broadcasts, so it serialised up to 16-way on ~1e9 elements. A banked shared byte-indexed LUT (returning two packed BF16) was worth -88 ms on its own, and it retroactively explains why six "bandwidth" fixes on that kernel did nothing: it was never bandwidth-bound.

New: Mamba-2 SSD chunked scan (3 kernels, kernels/gb10/common/mamba2_ssd_chunk.cu). The old prefill walked a 1023-step sequential recurrence. Because dA is scalar per (head, token), prod dA factorises as exp(cs_t - cs_s), so the scan becomes tensor-core matmuls with 16 serial links instead of 1023. I fused three of vLLM's five passes so the ~50 MB/layer `states` tensor they round-trip through DRAM never exists. SSM 309 -> 263 ms.

What is left, and what does NOT work. The MoE grouped GEMM (285 ms, against a ~178 ms bandwidth floor - 33.8 GB of expert weights must stream every prefill) is OCCUPANCY-STARVED: 17.8 KB smem gives ~20 warps/SM and ANY smem addition loses more than it buys. Failed, each measured: cp.async 2-stage (-5%), 128B cache-line staging (-12%), dequant-into-registers (-20%, that BF16 tile is a dequant cache shared across warps), L2 evict_first hint (0%), and a cp.async RETRY after the kernel became genuinely DRAM-bound (-5% again). The fix has to be ONE coherent change: M_TILE=16 with the 4 warps splitting N (which frees the 8.4 KB tile), or SwapAB so tokens land on the MMA's 8-wide N axis (what CUTLASS 4.6 / FlashInfer did for SM120 small-M MoE). Piecemeal application regresses.

Also worth knowing for this silicon: SM120/SM121 HAS TMA (single-CTA, no multicast) and a native block-scaled FP4 MMA (mma.sync...m16n8k64...mxf4nvf4...ue4m3) that consumes E2M1 + per-16 E4M3 scales - i.e. our exact weight format - with zero software dequant. It needs a swizzled scale-factor layout. Note Marlin-style W4A16 still beats native FP4 CUTLASS on SM120 today (46-49 vs 39 tok/s), so it is not a free win.

PR #283, opened 2026-07-09
fix(toolcall): harden qwen3_coder tool calls against long-context echo attractor

AzeezIsh:
## Problem

A live opencode session on Qwen3.6-35B-A3B-FP8 (64k ctx, MTP K=2, tool grammar on, GB10) collapsed at ~42.5k prompt tokens: the model drifted into a wire-format **echo attractor**, emitting the XML tool format shifted by one structural token:

```
<parameter=parameter>filePath>
/tmp/opencode-test-session/server/src/auth/session.rs
</parameter>
<parameter=write>content>use axum::...</parameter>
```

The parsed calls carried the echoed words (`parameter`, `write`) as argument keys, the missing required keys were backfilled as `""`, opencode errored on the empty path (`BadResource`), the error fed back into the already-degraded context, and the model retried the identical call every turn — a doom loop until abort.

Forensics (server `--dump` replay + source trace) found three amplification layers, each fixed here:

## Fixes

1. **Grammar — constrain the `<parameter=NAME>` key slot to the tool schema's property names** (`grammar/compile_tools.rs`). Previously the compiled EBNF accepted ANY identifier (`paramname ::= [a-zA-Z_][a-zA-Z_0-9]*`) and the sanitized schema was explicitly discarded — so `parameter=parameter` was grammar-legal, and the key-slot masking of `=` actively converted a recoverable one-token echo into a permanently mis-keyed call. The key slot is now a literal alternation of schema property names (both the primary and json_schema-fallback paths). Schema-less and explicitly-open (`additionalProperties`) schemas keep the historical rule.

2. **Parser — echo-attractor salvage** (`tool_parser/validation.rs`, `streaming_emit.rs`). New SSOT `salvage_echoed_param`: when a parsed key is NOT a schema property and the value begins with `PROP>` for a schema property `PROP`, re-split key/value. The real path and content were fully present inside the mis-keyed values; this recovers them. Wired into the buffered pipeline (step 2.5, before the missing-required `""` backfill) and live argument streaming (`coerce_kv`), with a dedup guard and a never-clobber-populated-keys rule.

3. **MTP — complete the BUG#4 clamp** (`scheduler/spec_step.rs` + 6 propose sites). `effective_num_drafts = 1` under an active grammar was applied only in the Phase-A bootstrap; the five verify-path re-propose sites (`verify_k2_step`, `verify_k3_step`) passed raw `num_drafts`, drafting `draft[1]` against a stale position-0 bitmask — the "mask held fixed across draft positions" warn spam and acceptance collapse during grammar-constrained tool calls. Hoisted to a shared `effective_drafts_under_grammar` used at all six sites.

## Verification

- **Unit**: 577 pass, 0 fail — including new regression fixtures reproducing the live failure shapes byte-for-byte through BOTH parse paths (`tool_parser/tests/group_h_salvage.rs`) and grammar-rule tests (`grammar/tests/param_key_constraint.rs`).
- **E2E on GB10 (dgx1)**: replayed the exact captured 42.5k-token failing request against this build, N=10: **10/10 valid `write` calls** (correct `filePath`, 3.5KB of real Rust content) vs **0/5 on main** (finish=length repetition loop / malformed keys). MTP "mask held fixed" warnings: **0** (was constant spam). Short-context parallel-tool smoke: 2/2 clean calls (no happy-path regression).

## Authorship

AI-generated (Claude), driven and reviewed by @azeezish. Root-cause forensics, implementation, and e2e validation on the GB10 boxes as described above.


AzeezIsh:
Second fix pushed to this branch (65c16bd4) — found during the follow-up opencode session on the first fix's build.

**New root cause:** the tool-grammar value rule banned `<` as the first content character (`first_content ::= [^ \t\r\n<=>]`, an Epoch-3 guard against the close-tag-as-first-body-token failure). Collateral damage: any `<`-initial file was unrepresentable — the mask deflected `<script>`/`<!DOCTYPE` at position 0, so the model could not write Svelte/HTML/XML files at all. In the live session every `.svelte` write came out script-logic-only (app.html contained JavaScript), the model noticed and retried the identical write forever, and the fuzzy-repetition / loop-detector guards then cut the loop mid-sentence (`finish=length` at 114/133 tokens, all dump flags false).

**Fix:** `first_content` now reuses the same BUG#2 negative-prefix close ladder as `rest_part` for its `<` arm — `<` is legal at content start unless it begins the exact `</parameter>` close. The non-empty guard survives (empty body / immediate close stay rejected) and the `=`/`>` first-char exclusions are kept. Plus observability: scheduler guards that force-finish a turn now surface as `guard_stop` in the synthesized `--dump` body (the fuzzy cut was previously indistinguishable from a plain length stop).

**Verification:** 580 unit tests pass (new token-level xgrammar tests accept a real Svelte component / `<!DOCTYPE html>` / `</div`-initial values, still reject empty body). E2E on GB10: the captured failing Svelte-write turn (29.7k ctx) now produces FULL components 5/5 — 3–4.2KB with script+markup, several runs emitting all three pages in one turn — vs 100% JS-only before; fresh `<!DOCTYPE` smoke passes; the 42.5k echo-attractor regression gate from the first commit stays 5/5 valid; zero guard warns across all replays.

AI-generated (Claude), driven and reviewed by @azeezish.

AzeezIsh:
Third and fourth commits pushed: the ultracode-verified P0+P1 batches for the 45k-token collapse (33-agent forensic workflow, 21 adversarially-confirmed findings — the model was exonerated; six server defects stacked).

**P0 (a410d5a4):** envelope-stuck guard no longer false-exits on `</` inside Svelte/HTML parameter values (it was force-killing legitimate writes mid-file — 8 kills in the live session); garbled close-reopen (`</parameter<parameter=`) salvage in both parse paths; server-authored corrective feedback naming empty required params (wires the previously-dead `find_empty_required_params`); hint_injector now recognizes structured client error codes (`BadResource:` etc.) so the escalating rescue actually fires; DIAG dumps env-gated.

**P1 (c62a20bb), each kill-switched:** grammar opt-ins `ATLAS_GRAMMAR_ALLOW_EMPTY_VALUE` (empty values representable) and `ATLAS_GRAMMAR_FORCE_CLOSE` (a dropped `>` after `</parameter` becomes unrepresentable); MTP verify path now honors request temperature (was pinned 0.0 — retries were byte-identical) and min_p_floor is threaded to the MTP bootstrap/first-token samples (was hardcoded 0.0, bypassing the documented FP8 safety net); loop handling is error-gated (the `-12.0` tool-call mask no longer fires on FAILING-call loops where it blocked the only escape); duplicate error observations masked to `[same error as below, attempt k of N]`.

**Stacked A/B on two GB10 boxes** (image A = P0 on dgx1; image B = P0+P1+flags on dgx2), identical battery:
- Svelte 3-page replay: 9/9 FULL components on both (0 envelope warns; was 100% mid-file kills / JS-only before this PR)
- Big-HTML with 36 close tags: complete on both, 0 envelope warns
- Poisoned 42.5k echo context: both recover — B escapes via a bash-heredoc fallback, following the injected `<CRITICAL>` hint (the hint path was dead before P0-4)
- Temperature probe (0.7, tools armed): A = byte-identical runs (pinned), B = differs (honored) — P1-3 verified live
- BFCL v4 subset gate (simple_python/multiple/parallel, seed 42, temp 0): **A = 92.50/90.83, B = 92.50/90.83 — bit-identical.** The grammar opt-ins cost zero accuracy.
- Unit: 583 (P0) → 595 (P1) tests pass.

AI-generated (Claude), driven and reviewed by @azeezish.

AzeezIsh:
Fifth commit: P1-5b result-progress gate. The post-P0+P1 live session reached 57k tokens (deepest yet, survived client-side compaction) but ended when legacy loop-detect Suppress hard-masked <tool_call> on a productive cargo check→fix→check cycle (similar calls, DIFFERENT results each round — the BW1 false-positive class). Suppress now requires round-to-round tool RESULTS to also be near-identical (Jaccard ≥ 0.9, shared shingle SSOT) before masking; differing results = forward progress = soft decay only. Failing-repeat and kill-switch semantics unchanged. 598 tests pass. AI-generated (Claude), driven and reviewed by @azeezish.

AzeezIsh:
Sixth commit: P2-1 inter-tool prose budget fix.

The next live session died at turn 6 (`finish=length`) while writing a legitimate API **architecture plan** — the 384-token inter-tool prose budget (a bound for the `<invoke>` dormant-opener wander) amputated it mid-sentence. Root cause: the guard assumed `grammar_state.is_some()` ⇒ "a tool turn, never plain chat, so long answers aren't truncated" — but opencode arms tools on **every** turn, so a plan-presentation prose turn is subject to the cut. The repeating wander this budget nominally guards is already caught independently by the content-loop + SimHash watchdogs; the residual is only a non-repeating dormant-opener burn.

Fix: default 384 → 3072 (fits plans, still << max_tokens per the do-NOT list), env override `ATLAS_MAX_INTER_TOOL_PROSE`, MODEL.toml bumped, and `guard_stop="inter_tool_prose_budget"` stamped on the cut for dump observability. 599 tests pass.

Verified on GB10 (`:p2`): a tools-armed pure-prose turn now completes at **1,550 tokens, finish=stop** (was cut at 384/finish=length); 0 prose-budget warns across the session. Fuller wander-evidence gating (fire only on a dormant-opener/loop signal, not a raw count) remains the P2-1 follow-up.

AI-generated (Claude), driven and reviewed by @azeezish.

SeedSource:

### Triage
- Rationale: Loop detection and diagnostic logic changes; no quantization or model weight modifications.
- Serving-path files: `crates/spark-model/src/model/trait_impl/prefill_b/finalize_last.rs, crates/spark-server/src/api/chat/loop_detect.rs, crates/spark-server/src/api/chat/msg_entry.rs, crates/spark-server/src/api/chat_stream/handle_done.rs, crates/spark-server/src/api/chat_stream/handle_token.rs, crates/spark-server/src/api/chat_stream/mod.rs`

### Serve gate

### GLM-5.2 code review
**Correctness concerns:**

1. **`outcomes_fwd` pairing bug**: Tool results attach to `outcomes_fwd.last_mut()` regardless of intervening user/system messages. A user message between assistant and tool result breaks the pairing silently. The forward scan assumes strict adjacency.

2. **`all_err` initialization**: Set to `true` on assistant push, AND-reduced per tool result. An assistant turn with *no* following tool results keeps `all_err = true` while `saw_result = false`. `failing = saw_result && all_err` masks this, but it's fragile.

3. **`take(8)` after `rev()`**: Takes newest 8, but `recent_calls_all_failing` / `recent_results_progressing` receive a window that may not align with `detect()`'s signature window. Mismatched windows could produce inconsistent verdicts.

4. **`duplicate_error_masks` first-match grouping**: A near-duplicate of group A's first member that's *also* near-duplicate of group B's first member joins A only. Transitive duplicates may fragment across groups, leaving unmasked copies.

5. **Empty shingle sets**: Comment acknowledges `< 4 tokens` produces empty sets but no handling shown — Jaccard on empty sets is undefined (NaN or panic).

**Risk:** High. Core loop-detection logic changes under time pressure with "P0-3 corrective feedback" still unshipped. The diff is truncated mid-function, so the shingle edge case may be handled below the cut.

**Quality:** Kill-switch (`ATLAS_NO_ERROR_DEDUP`) is good. Env-var diagnostics removed from hot path is positive. Inline dates in comments will rot.

tbraun96:
This PR has allowed a nonstop session of opencode running against Qwen 3.6 35B A3B served w/Atlas. Past 32k context, we use to get repeated lines following bad tool calls, ending with the final repeated sentence getting stopped half-way through generation. With the changes in this PR, we've gone up to 40k before degradation, then to 50k, and no more.

tbraun96:
## Benchmark Results

All three gates passed. PR is safe to merge.

| Gate | Model | Result | Verdict |
|------|-------|--------|---------|
| **A** — 35B webserver | Qwen3.6-35B-A3B-FP8 | 10/10 webserver_ok, Σwall=1589s | ✅ PASS |
| **B** — 35B BFCL | Qwen3.6-35B-A3B-FP8 | 84.66/83.32 (exact baseline) | ✅ PASS |
| **C** — warm-TTFT 27B | unsloth/Qwen3.6-27B-NVFP4 | pr283 1374ms vs mainctl2 1374ms (+0.002% median, -0.2% p90) | ✅ PASS |

**Gate C detail:**
- pr283 leg: median TTFT 1374ms, p90 2076ms
- mainctl2 leg: median TTFT 1374ms, p90 2080ms
- Delta: +0.03ms median (0.002% slower), -4.1ms p90 (0.2% faster)
- Both within ±3% threshold

**Note:** Gate D (27B golden BFCL accuracy) was skipped per operator request.

---
🤖 Benchmark run on 2026-07-15 across dgx1/dgx2/dgx3 (10.10.10.1/2/3)

PR #285, opened 2026-07-10
refactor(loader): gate batched-prefill MLA exclusion on capability, not model_type

rsafier:
> **Re-opened replacement for #261.** #261 was merged into its base feature branch by mistake on 2026-07-09 and reverted; the base branch has since been restored and restacked onto `main`. GitHub cannot reopen a merged PR, so this is a fresh PR carrying the identical commit (verified patch-identical via `git range-diff`). Marked **ready for review** — the original's "draft" wording was stale.

**Stacked on #260** (base `fix/generic-nvfp4-moe-gate`). Unit tests green; the e2e co-dispatch smoke is still outstanding.

## What
`check_kernel_batched_eligible()` blocked batched-prefill attention for MLA models via `config.model_type == "mistral"` — a hardcoded proxy the original author flagged as conservative (*"mistral is the only MLA model in Atlas today"*). This keys it on the actual capability: `config.kv_lora_rank > 0` (== `AttentionType::Mla`).

## Behavior preservation
`kv_lora_rank > 0` == `{mistral, deepseek_v4}`; every other model_type has `kv_lora_rank == 0` → `is_mla=false` → unchanged. **deepseek_v4** is the only newly-matched model, but it is *already* rejected one check later by `head_dim > 256` (head_dim=512) — so its outcome (not batched-eligible) is **identical**. Net: bit-identical on every shipped model, plus it future-proofs any MLA checkpoint shipped with `head_dim <= 256`. **`batch_kernel_tests` pass (12/12)** — they directly exercise the converted predicate (MLA reject, head_dim reject, non-MLA accept).

## Why this is the *only* additional conversion
This PR is the output of an exhaustive adversarial pass over **every** `model_type` string gate in the tree (stacking on #260, which already did the two `qwen35/load_layers.rs` MoE gates). The analysis found **this is the only other safe architecture-generic conversion.** In particular it **rejected** several tempting ones:
- **MTP allowlist** (`preflight.rs` `MTP_SUPPORTED_MODEL_TYPES`) → **must NOT** become `capabilities().has_mtp`: `has_mtp` (`mtp_num_hidden_layers>0`) is *broader* than the allowlist and is true for deepseek_v4 / minimax_m2 / step3p7 — converting would re-enable exactly the unconsumable per-module-MTP loader path (MiniMax) the guard exists to reject.
- `is_qwen35*` predicates → loader sub-dispatch (protected; no capability expresses "use dense vs VL loader").
- gemma4 `skip_lm_head_quantization` / softcap, minimax/step3p7 setup, the `dispatch.rs` model_type→kernel-dir rewrites → genuinely checkpoint/family-specific, **kept**.

**No kernels move; no loader dispatch changes.** The refactor is deliberately minimal — the value is as much the *documented boundary* of what stays model_type-gated as the one conversion itself.


SeedSource:

### Triage
- Rationale: Refactoring MLA eligibility check from string model type to boolean capability flag; no quantization or serving changes.
- Serving-path files: `crates/spark-model/src/model/trait_impl/prefill_b/batch_kernel/eligible.rs, crates/spark-model/src/model/trait_impl/prefill_b/batch_kernel_tests.rs`

### Serve gate

### GLM-5.2 code review
**Correctness:** Sound. `kv_lora_rank > 0` is the canonical MLA marker and correctly supersedes the brittle `model_type == "mistral"` string match. The comment accurately notes deepseek_v4 was already caught by the subsequent `head_dim > 256` guard, so no behavioral regression on shipped models. Parameter type change from `&str` to `bool` is clean and all call sites are updated.

**Risk:** Low. The refactor is narrowly scoped to one eligibility gate. The only latent risk: if a future MLA model has `kv_lora_rank > 0` *and* `head_dim <= 256`, it would now be rejected earlier (correct) rather than slipping through the string check (bug). That's a strict improvement.

**Quality concerns:**
1. The comment block (lines 160-165) is verbose for a hot-path file. Trim to 2 lines.
2. No test exercises a non-mistral MLA model (the stated motivation). Add one with `is_mla=true, head_dim=128` to lock in the fix.
3. Tests pass `false` for non-MLA cases but never assert the `kv_lora_rank` derivation logic itself—consider a unit test on the caller side.

**Verdict:** Approve with the test gap addressed.

tbraun96:
Reviewed this properly against current `main`. The conversion is semantically right and I want it — but not in this shape. Three blockers, all reproduced.

**1. It does not compile against current main.** `git merge-tree --write-tree` reports a *clean* merge, which is misleading (I've been bitten by that exact reading twice this week). Actually building it:

```
ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo test -p spark-model batch_kernel
-> 4x E0308
   batch_kernel_tests.rs:278  raw_charge_blocks_stacking_at_production_sizes
   batch_kernel_tests.rs:296  effective_charge_allows_warm_stacking
   batch_kernel_tests.rs:315  effective_charge_still_rejects_when_sum_exceeds_arena
   batch_kernel_tests.rs:336  effective_charge_rejects_zero_length_stream
```

main gained four tests after this PR's base; they still pass `"qwen3_next"` into the now-`bool` parameter. Needs a rebase regardless of everything else.

**2. The new tests don't test the change.** I tried to break it. Reverting only the production hunk doesn't discriminate — the tests are coupled by signature, so they can't even compile against the unfixed code. So I rebased, confirmed 16/16 green, then sabotaged the *caller* — the only hunk that changes production behaviour — replacing `self.config.kv_lora_rank > 0` with `false` at `eligible.rs:96`, which would let a mistral model sail straight past the MLA guard in production.

Result: **`test result: ok. 16 passed; 0 failed`.**

Every test hand-feeds the bool to the pure function. Nothing asserts that a real mistral/deepseek `ModelConfig` produces `is_mla == true` at the call site — which is the only thing this PR changes. (Sabotaging `if is_mla` inside the function *does* fail `rejects_mla_model`, but that coverage predates the PR.) One config-level test — mistral config in, rejected out — would close it.

**3. It's stacked on #260** (base `fix/generic-nvfp4-moe-gate`, contains `6a482105c`), so merging it merges that too. It can't be judged as "eligible.rs + tests only".

On merits: by your own claim it's bit-identical on every shipped model, so it delivers zero behaviour today. Under the merge rigor now in force a `crates/` change owes five GPU gates (~4h19m measured), and I can't justify that for a zero-delivery refactor whose integration point is untested.

**What I'd merge:** rebase onto main (fixing the four call sites), add the config-level test, and re-target it at `main` rather than #260's branch. Then it rides through gates alongside a change that already pays for them. Leaving this open — the direction is right, and `init_arch_gates.rs:51` already uses `mla: config.kv_lora_rank > 0`, so this is the SSOT-consistent move.

rsafier:
@tbraun96 All three blockers addressed: (1) synced with main — the four `batch_kernel_tests` call sites that passed `"qwen3_next"` into the now-bool parameter are converted (bare-bool, the file's idiom); (2) your sabotage finding gets a named seam — the production caller now reads `config_is_mla(&self.config)` and the new `mistral_config_is_rejected_as_mla` test drives a `kv_lora_rank = 512` config through that same seam to a rejection, so flipping the derivation now fails a test; (3) re-targeted at `main`, and the tree no longer carries #260's loader delta (`load_layers.rs` reset to main's version), so this stands alone and can ride the gates alongside whatever lands first. fmt/clippy/tests green locally (314 passed).

tbraun96:
Landed in main via the #440 stack merge (`3c941751`). Verified by content diff — `config_is_mla()` at `crates/spark-model/src/model/trait_impl/prefill_b/batch_kernel/eligible.rs:25`; all three files byte-identical to this PR's head. Thanks @rsafier.

Issue #248, opened 2026-07-04
dflash: measure wy16_f32 GDN fix default-ON impact (currently opt-in, ATLAS_DFLASH_VERIFY_GDN_F32=1)


Sujimoshi:
Follow-up from PR #218's T=0/kernel-numerics investigation (2026-07-04). `ATLAS_DFLASH_VERIFY_GDN_F32` (the wy16_f32 fix) eliminates a real BF16-truncation-precision loss in the GDN (gated delta rule) kernel's WY-chunkwise output path, validated in isolation during the investigation. It ships in #218 as **opt-in, default OFF** — this is a deliberate decision, not an oversight, and is recorded here so it doesn't need to be reconstructed later.

**Why opt-in rather than default-ON:** unlike the two default-ON fixes in #218 (qkvz kernel-family mismatch, out_proj weight-transpose mismatch), which were validated at the *contract* level — each one measurably moved a specific batched-vs-serial argmax divergence toward matching serial decode, with clear before/after numbers — wy16_f32 is validated only at the *mechanism* level (removes a known precision-loss source in isolation). It was not the cause of the byte-49/79 divergence investigated in #218, and its effect on accept-rate/perf when enabled by default has never been measured. Shipping it as default-ON without that measurement would mean defaulting-in an unproven "genuine precision fix ⇒ worth enabling" assumption — exactly the class of shortcut the rest of the investigation avoided.

**Ask:** run an A/B (accept-rate, tok/s, and ideally a KL_DIAG margin-distribution comparison) with `ATLAS_DFLASH_VERIFY_GDN_F32=1` vs default, on a comparable prompt/harness to the rest of the #218 investigation. If it's neutral-or-positive on accept-rate/perf, flip the default to ON in a follow-up PR. If it has a measurable cost, keep it opt-in (or drop it) and document the finding.

Weschera:
Wanted to run this A/B on GB10 (Qwen3.6-27B-NVFP4, harness ready), but PR #218 is unmerged and ATLAS_DFLASH_VERIFY_GDN_F32 isn't present in any Docker Hub image. Verified by binary and filesystem search across all tags including the fresh dev push. If you publish a build from the optimizations branch, we'll run accept rate and tok/s both ways same day.

PR #290, opened 2026-07-10
feat(dflash): SSM-layer K=γ verify kernels — WY16 GDN, batched phase-1 conv, opt-in F32 GDN output

Sujimoshi:
## Summary

Fourth in the split of #218 (DFlash speculative decoding) into smaller, independently reviewable PRs (#255, #257 merged, #258). This PR carries the **SSM/GDN layer-level** pieces of the K=γ verify path — the kernels and `Qwen3SsmLayer` methods that #258's `TransformerLayer` scaffolding (`prefill_phase1_verify`, `prefill_gdn_wy16`, …) dispatches to. The model-level verify orchestration (`verify_d_kgamma_*`) and the DFlash head/scheduler pieces come in the next PRs.

**Stacked on #258** — this branch includes #258's commits as ancestors (GitHub won't allow a fork branch as PR base). **Merge order: #258, then this.** Once #258 lands, this diff shrinks to its own single commit.

### What's in here

- **`gated_delta_rule_wy16.cu`** (new): fused K=16 WY-chunkwise GDN verify kernel (+ F32-output variant) — one pass writes final `h_state` *and* the `Hi_0..Hi_14` per-token intermediates needed for partial-accept rollback. Wired via `prefill_gdn_wy16`.
- **`causal_conv1d_update_prefill_inter`** (new kernel + ops wrapper): batched phase-1 conv over all K verify tokens that also writes per-token `conv_state` intermediates inline, replacing K sequential `conv1d_update_l2norm` launches per SSM layer. Wired via `prefill_phase1_verify`. This is the C1 conv fusion from #218 (verify 496→269 ms, tok/s 12→22 on GB10).
- **`ATLAS_DFLASH_VERIFY_QKVZ_FIX`** (default ON): the M=1 verify-path QKVZ projection routes through the same `w4a16_gemv` kernel serial decode uses, instead of a GEMM-family kernel at the same shape (~2.29% rel_diff between the two arms otherwise).
- **`ATLAS_DFLASH_VERIFY_OUTPROJ_FIX`** (default ON): per-row `w4a16_gemv` out_proj for verify-window calls to keep batched verify numerically aligned with serial decode (~2.26% rel_diff otherwise).
- **`ATLAS_DFLASH_VERIFY_GDN_F32`** (opt-in, tracked in #248): F32 GDN output in the split4/wy16 verify branches, removing a documented BF16-truncation source.
- **Refactors**: the sequential K∉{2,3,4,17} `decode_batched` fallback moved to `conv_gdn_sequential.rs`; `qwen3_ssm/mod.rs`'s `TransformerLayer` impl moved to `trait_transformer_impl.rs` (500-LoC cap).

### Deviations from #218 as-authored

Main evolved independently since #218 branched, so this is a re-derivation rather than a straight cherry-pick:

- The QKVZ/out_proj projection ladders #218 patched inline were since replaced on main by shared dispatch helpers (`prefill_qkvz_proj` / `prefill_out_proj_dispatch`, #229) that already carry the CUTLASS-NVFP4 and FP8 branches. The two verify fixes are layered as guarded pre-branches in front of those helpers instead of reapplying #218's ladders.
- The out_proj fix is additionally gated to **≤17-token, native-NVFP4-only** calls: on current main `prefill_phase3` is also the hot batched-prefill path, where #218's unconditional per-row GEMV loop would serialize full prompt chunks, and on FP8/dense builds `ssm.out_proj` may be NULL (the #232-class NULL-deref).
- #218's `trait_prefill_gdn_batched.rs` / `qkvz_projection.rs` extractions were dropped — main since made its own supersets of both (`trait_prefill_gdn/batched.rs` incl. batched-FLA, and the FP8-aware inline dispatch in `trait_decode_batched.rs`).
- NVTX instrumentation stays in #255.

## Test plan

- [x] `cargo check -p spark-model`, `cargo fmt --check`, `cargo clippy --workspace --tests`, 500-LoC cap — all green locally
- [ ] CI
- [ ] End-to-end DFlash verify gate lands with the model-level orchestration PR (these methods are dispatched via #258's scaffolding defaults — `Ok(false)`/fallback paths keep current behavior until then)


SeedSource:

### Triage
- Rationale: NVFP4 lm_head vocab capping fix and GDN verify buffers added for DFlash path.

### Serve gate
- Build: 17s - Qwen3.6-35B-A3B-FP8 on GB10 node1
- Top-k KL vs main: **0.0010** (no shift vs main) - decode delta **-0.4%**

### GLM-5.2 code review
Diff truncated mid-file (dflash_head.rs), so full correctness assessment is impossible. Key observations on visible portion:

**Correctness concerns:**
1. `target_lm_head_nvfp4_vocab_for_dflash = config.vocab_size` — comment claims this is the *capped* value used for the quantized head's GEMM N. If `cap_vocab_size_to_tokenizer` mutates `config.vocab_size` in place before `build_model`, this is correct. If capping is stored elsewhere (e.g., separate field), this reads the *uncapped* value — the exact OOB the comment warns against. Needs verification of capping side-effect.

2. `output_f32_written: Cell<bool>` — Cell in a struct holding DevicePtrs suggests this may be shared across threads or accessed from CUDA callback/stream-ordering contexts. If GDN dispatch and phase3 run on different streams without a sync, the Cell read may race or observe a pre-sync value. Confirm there's a stream sync or event fence between the write and the phase3 read.

3. New trait methods (`prefill_phase1_verify`, `prefill_gdn_wy16`) default to `Ok(false)` — safe fallback, but no `#[doc]` contract on whether `Ok(false)` requires leaving `state`/`gdn_bufs` unmodified. Overrides must be side-effect-free on the false path or callers will see partial writes.

**Quality:** Doc-comment reformatting is noise-heavy; prefer separate commit. Missing: no test additions visible.

**Risk:** Medium — OOB on lm_head buffer if vocab assumption is wrong; Cell ordering if unsynced.

Sujimoshi:

PR #294, opened 2026-07-10
feat(dflash): drafter head core — forward_block split, NVFP4-safe lm_head, γ-1 noise0-skip fix

Sujimoshi:
## Summary

Next cluster from the #218 split. Ports the remaining DFlash drafter-head ("dflash_head core") changes: `forward_block` layer/step0 restructuring, and — the significant one — a correctness fix in the drafter's own argmax extraction that was silently corrupting every K=γ verify comparison.

**Stacked on #292 (verify orchestration) → #290 (SSM verify kernels) → #258 (model glue). Merge order: #258, #290, #292, then this.**

## Contents

- `dflash_head/forward_block_layer.rs` — Q/K/V/O/MLP projections switched to `dense_gemm_bf16_pipelined` (tensor-core pipelined GEMM).
- `dflash_head/forward_block_step0.rs` (new) — fc-projection step extracted from `forward_block.rs` (500-LoC cap); batches the GEMM over all `eff_ctx` ctx slots in one launch instead of a per-slot loop (avoids O(eff_ctx) re-reads of the 262 MB fc weight matrix).
- `dflash_head/forward_block.rs` — **the actual fix**: the final logits argmax now correctly skips `noise0` (the conditioning slot) when extracting draft tokens. Per the block-diffusion paper, the model is only trained at masked positions — `noise0`'s logits are untrained garbage. SGLang's reference skips it (`draft_next = greedy_sample(draft_hidden[:, 1:, :])`); this cluster matches that, extracting `γ-1` valid drafts from `noise1..noise{γ-1}` instead of all `γ` rows. Preserved the NVFP4-aware lm_head GEMM dispatch (`w4a16_gemm` for NVFP4 lm_head targets, e.g. Holo) already on this stack — #218's own version of this file predates that support and would have regressed it.
- `dflash_head/propose.rs` — timing instrumentation, optional `ATLAS_DFLASH_T=2` iterative denoising (off by default).

## The finding

Earlier GPU validation (#290, #292) measured a hard **0/16 accept collapse** on the K=γ (WY16) verify path. That was blamed on a deep SSM state-management mismatch. Turns out a large part of it was simpler: the noise0 conditioning slot was leaking into the drafted-token argmax, desyncing every downstream K=γ comparison regardless of kernel correctness.

With this cluster's fix, GPU validation (GB10, Qwen3.6-27B-NVFP4 + DFlash draft, T=0, forced `ATLAS_DFLASH_DRAFT_CAP=16`) now shows **real, non-zero acceptance** (0-93% per verify call, avg ~15-20% over a 700-token run), no output corruption, MTP gate stayed `ENABLED` throughout. This is a meaningful step forward for K=γ correctness, even though it's not yet a throughput win by default (see below).

## Deviations from #218 as-authored

- Left `dense_ffn.rs` out entirely — its `prof!`-instrumentation hunk is superseded by later perf work already on this stack (split_silu / single-warp GEMV dispatch); its other two hunks are NVTX (out of scope, tracked in #255).
- `qwen3_attention/prefill/paged_oproj.rs` / `paged_qkv.rs`: same story — their only changes were NVTX tags around GEMM dispatch that's already more advanced on this stack (cutlass_nvfp4 routing) than #218's version. Kept current dispatch unchanged.
- `dflash_head/from_weights.rs`: kept this stack's existing YaRN-aware RoPE inv_freq table (handles both `rope_scaling: yarn` and standard/null per config.json) over #218's blanket revert-to-standard-RoPE — the YaRN-aware version already supersedes it.
- **Did not** adopt #218's flip of `ATLAS_DFLASH_DRAFT_CAP`'s default from 1 to `gamma - 1` (defaulting every request to the K=γ path). Even with the noise0-skip fix, this run's K=γ path cost ~12.6x a plain decode step per verify (905ms vs 71ms) against ~15-20% average acceptance — net throughput was 3.1 tok/s vs the K=2 default path's 13.0 tok/s on this workload. Correctness is fixed; the throughput tradeoff isn't characterized yet. Kept the safe default (1, forces `step_verify_k2`).

## Test plan

- [x] `cargo check -p spark-model` / `cargo check --workspace`
- [x] `cargo clippy --workspace --tests`
- [x] file-size-cap (0 violations)
- [x] GPU validation on GB10 (Qwen3.6-27B-NVFP4 + DFlash draft, T=0):
  - Greedy baseline: 700 tokens, no crashes.
  - Default K=2 path: divergence at byte 634 vs greedy (no regression), **70% accept rate** over 100-step window (up from #292's 3-5% — likely a downstream benefit of the same noise0-skip fix improving ctx quality fed back into the accumulator).
  - Forced K=γ path (`ATLAS_DFLASH_DRAFT_CAP=16`): divergence at byte 290 vs greedy (coherent text, no corruption/gibberish), real non-zero acceptance (0-93%, avg ~15-20%), MTP gate stayed `ENABLED`. Net throughput (3.1 tok/s) currently below the K=2 default (13.0 tok/s) — a follow-up throughput-tuning item, not a correctness blocker.


SeedSource:

### Triage
- Rationale: Fixes NVFP4 lm_head buffer read overflow for DFlash drafter; adds verify buffers and methods.
- Serving-path files: `crates/spark-model/src/factory/build.rs, crates/spark-model/src/layer.rs, crates/spark-model/src/layer/transformer_layer.rs, crates/spark-model/src/layers/dflash_head.rs, crates/spark-model/src/layers/dflash_head/forward_block.rs, crates/spark-model/src/layers/dflash_head/forward_block_layer.rs`

### Serve gate
- Build: 32s - Qwen3.6-35B-A3B-FP8 on GB10 node1
- Top-k KL vs main: **0.0010** (no shift vs main) - decode delta **-0.0%**

### GLM-5.2 code review
**Correctness concerns:**

1. **Vocab cap bug (critical):** The comment says `lm_head_nvfp4` was quantized against the *capped* vocab size, but the code passes `config.vocab_size` — which is the *uncapped* drafter value per the comment. If the tokenizer cap reduced rows, this passes the wrong N and reintroduces the OOB read it claims to fix. Should pass the actual packed row count of `lm_head_nvfp4`, not `config.vocab_size`.

2. **`output_f32_written` Cell<bool>:** `Cell` is fine for single-threaded CPU state, but if `GdnPrefillBuffers` is shared across threads (likely given `Send + Sync` layer trait), this is a data race waiting to happen. Use `AtomicBool`.

3. **`prefill_phase1_verify` / `prefill_gdn_wy16` default `Ok(false)`:** Fallback semantics are correct, but no test in the diff exercises the `Ok(true)` fast path or the fallback. Risk: silent perf regression if the override never fires due to a gating bug.

4. **Doc-only churn:** ~60% of the diff is comment rewording with no behavior change. Bloats review surface; should be a separate PR.

**Risk:** Medium — the vocab N mismatch is a real OOB on NVFP4 packed buffers.

**Verdict:** Request changes: fix the vocab source, swap `Cell`→`AtomicBool`, add a test for the verify fast path.

Sujimoshi:

PR #297, opened 2026-07-11
feat(dflash): scheduler/server verify-step glue — raw-argmax bypass, T>0 stochastic accept, EAGLE ctx-fix default-on

Sujimoshi:
## Summary

Next cluster from the #218 split. Wires the scheduler/server side of DFlash verify: threads a `dflash_verify_raw_argmax` flag from CLI through `scheduler::run` / `mtp_step` into all four verify dispatchers (K2/K3/K4/dflash), adds T>0 stochastic (rejection-sampling) acceptance for DFlash in the K=2 path, and flips `ATLAS_DFLASH_EAGLE_FIX` to default-on across both this branch's server-side consumers and #294's model-side capture logic.

**Stacked on #294 (dflash_head core) → #292 → #290 → #258.**

## Contents

- `scheduler/mod.rs`, `mtp_step.rs`, `main_modules/serve.rs` — thread `dflash_verify_raw_argmax` through to all verify dispatchers.
- `verify_dflash_step.rs` / `verify_k2_step.rs` / `verify_k3_step.rs` / `verify_k4_step.rs` — when DFlash is active, compare the drafter's raw argmax directly against the verify result instead of routing through the rep_pen/DRY sampling pipeline first (the pipeline-processed token can legitimately differ from the draft even when the target agrees, silently collapsing accept rate).
- `verify_k2_step.rs` — new `dflash_stochastic_accept` (rejection sampling against the draft distribution) for T>0 DFlash sampling.
- `ATLAS_DFLASH_EAGLE_FIX` now defaults **on** (both here and via a paired commit on #294's branch): fixes a real ctx-undercount where `propose()` was appending exactly 1 ctx slot per verify step regardless of `num_accepted`, instead of one slot per accepted position. `ATLAS_DFLASH_EAGLE_FIX=0` still opts back into the legacy single-row-0 capture for A/B comparison.

Dropped from the upstream #218 diff as debug/perf scaffolding, not feature content: an env-gated per-position acceptance diagnostic table in `verify_dflash_step.rs`, and an unconditional first-20-steps timing log in `verify_k4_step.rs`.

## Why the EAGLE-fix default flip

A tester flagged seeing K=γ accept rates commonly >50%, some >90%, and asked to hold off on #294 pending investigation. GPU validation on the identical prompt showed:

| | EAGLE_FIX off (old default) | EAGLE_FIX on (new default) |
|---|---|---|
| verify steps | 183 | 46-71 |
| mean accept | 14.5% | 27.8-35.5% |
| >50% steps | 5.5% | 18.3-26.1% |
| ≥90% steps | 0 | 1-2 |
| max accept | 73% | 93-100% |
| output | clean, `finish_reason=length` | clean, `finish_reason=length` |

The tester's report reproduces closely once this flag is on — but it's not a regression, it's the fix working as intended: correct per-position ctx accrual gives the drafter an accurate context, so it proposes better multi-token drafts and needs fewer verify iterations for the same output length. No corruption, no repetition, no degenerate loops observed in any run.

## Test plan

- [x] `cargo check --workspace`
- [x] `cargo clippy --workspace --tests`
- [x] file-size-cap (0 violations)
- [x] GPU validation on GB10 (Qwen3.6-27B-NVFP4 + DFlash draft, T=0, `ATLAS_DFLASH_DRAFT_CAP=15`, no other env overrides — exercising the new defaults as-shipped): 71 K=γ verify steps, mean accept 27.8%, max 100%, `finish_reason=length`, output coherent (valid Python + docstrings), cut off cleanly at the token limit.


Sujimoshi:

Issue #288, opened 2026-07-10
feat: CLI Arg Validator


tbraun96:
We want to make sure users converge to the correct settings, not diverge and endlessly publish content on the internet that causes asymmetry in understanding due to incorrect assumptions about the CLI.

For example, if `--kv-cache-dtype` is set to `bf16`, _and_ the user sets fp8 calibration tokens, the combination _doesn't make sense_. There are likely many other combinations that should flagged.

We should make invalid CLI combinations hard errors (NOT warnings) to enforce and faclitate correct CLI usage as we evolve. The error emitted should be helpful for (especially) AI's and humans.

PR #180, opened 2026-06-18
fix(gpu-mem): enforce --gpu-memory-utilization as a total-memory ceiling

arclabch:
## Summary

`--gpu-memory-utilization` was not enforced as a ceiling on total GPU memory.
The fraction was applied to *free-after-weights* memory when sizing the KV cache,
so weights/buffers/reserve fell outside the cap and the real footprint overshot
the requested fraction — the lower you set it, the worse the overshoot (≈0 GB at
the 0.88 default, +20 GB at 0.52, +27 GB at 0.40 on GB10).

This makes the flag a hard ceiling on **total** memory (vLLM / sparkrun
semantics): the realized footprint now scales with the requested fraction.

Closes #179
Refs #170 (auto-utilization feature — separate; correct ceiling enforcement is a
prerequisite, this PR does not implement or close it)

## Root cause

`crates/spark-model/src/factory/build.rs` computed the KV budget as
`(free_after_weights - inference_reserve) * gpu_memory_utilization`, so the
fraction scaled only leftover memory after weights loaded; weights, buffers and
the inference reserve were never counted against the cap.

## The fix

New budget formula in `build.rs`:

```rust
let total_mem    = gpu.total_memory()?;
let actual_free  = gpu.free_memory()?;
let used_so_far  = total_mem.saturating_sub(actual_free);
let total_budget = (total_mem as f64 * gpu_memory_utilization) as usize;
let kv_budget = total_budget
    .saturating_sub(used_so_far)
    .saturating_sub(inference_reserve)
    .min(actual_free.saturating_sub(inference_reserve));
```

- **Total-memory ceiling:** `u × total_memory`, minus everything already
  resident, minus the inference reserve, clamped to actual free memory.
- **Zero-budget guard:** when `kv_budget == 0`, build bails with a GB-formatted
  feasibility diagnostic (what's needed vs. available) instead of failing
  opaquely deeper in KV allocation.
- **CLI doc update** (`crates/spark-server/src/cli/serve_args.rs`): the
  `--gpu-memory-utilization` help now states it is a fraction of *total* GPU
  memory that weights + buffers + KV + reserves all count against.

Default is unchanged (0.90).

## Notes for reviewers

**Semantics (honest):** on GB10 unified memory, `free_memory()` returns
`max(cuda_free, /proc/meminfo MemAvailable)` — a whole-system figure. So
`used_so_far` folds in the CUDA context, the driver, and any co-resident host or
GPU process, not just Atlas's own prior allocations. This makes the cap a
[...]

Issue #170, opened 2026-06-15
[feat] Auto-gpu-utilization


tbraun96:
### Motivation

Atlas likes to load plenty of memory for optimal performance for all cache types (KV, SSM, etc). It's common to pass `0.9` gpu memory utilization, but, that leaves little room for much else. For many cases, all that memory is not necessary. Therefore, memory should, by default if unspecified, dynamically select the optimal memory amount.

### Proposal

Opus

### Scope

New model support (HF checkpoint integration)

tbraun96:
status: not started. #180 landed the prerequisite (utilization is now correctly a fraction of total memory), but the default is still a hardcoded 0.90 in serve_args.rs. sketch for whoever picks this up: when the flag is unset, derive the budget bottom-up instead of top-down — weights size (known post-preflight) + KV demand (max_seq_len × max_batch_size × kv-dtype bytes/token) + SSM pools + measured runtime buffers + a fixed headroom reserve, capped at 0.90 of total. explicit --gpu-memory-utilization keeps overriding. the KV sizing math already exists in serve_phases/kv_cache; this is mostly plumbing the pieces together and logging the derivation so users can see why the number was chosen. good first issue stands.

PR #293, opened 2026-07-10
feat(deepseek-v4): native MXFP4 (E8M0) routed-expert loading + kernels, end-to-end on GB10

SeedSource:
## What this adds

Native **MXFP4 (E8M0)** support for DeepSeek-V4-Flash, end-to-end, on GB10 — the model's *original on-disk* routed-expert format landed device-resident **transcode-free** (no FP8→NVFP4 requant round-trip):

- **Loader** — `weight_map`: a new `WeightQuantFormat::Mxfp4E8m0` (packed E2M1 nibbles + per-block **E8M0** power-of-2 scales, `GROUP_SIZE=32`, no per-tensor global) + `model_a::quantized_mxfp4_e8m0` (asserts GS=32, refuses non-MX bytes on the transcode-free path); `deepseek_v4/assemble.rs` wires it in; `transpose_for_gemm_gs(..)` generalises the scale block size (per-16 NVFP4 vs per-32 MX).
- **Kernels** — `moe_w4a16_grouped_gemm.cu` gains `_e8m0` variants of the grouped/fused GEMMs; `mx_block_scale.cuh` decodes the E8M0 biased-exponent byte (`2^(b-127)`); `moe_shared_expert_fused_t.cu` gains the routed-E8M0 + shared-NVFP4 fused decode kernel.
- **Dispatch** — `layers/moe` (`forward_prefill_routed`, `forward_phase`, `forward_batched`, `helpers_a`, `init`, `mod` + new `ptr_table_build`): E8M0 experts route to the `_e8m0` kernels as a leading branch and **never** reach an `Nvfp4` kernel (which would read the scale byte as FP8-E4M3 → silent garbage; guarded by a format assert at the dispatch site).
- **The wire** — `factory/m2_setup.rs`: allow `deepseek_v4` through the unified MoE prefill-transpose gate so routed prefill hits the validated E8M0 fused kernel instead of the NVFP4 fallback.

## Evidence chain

All numeric claims computed on a GB10 (SM121), target `deepseek-v4-flash`, this branch @ `HEAD`:

| Leg | What | Result |
|---|---|---|
| **Leg-1** device byte-verify | on-disk MX bytes land device-resident unchanged | device sha == on-disk (`177ac128…`/`ea4ac989…`) |
| **Leg-2 CHECK 1** decode vs host f32 GEMV (full-range E8M0) | `moe_expert_gate_up_shared_t_e8m0` | gate/up **exact 256/256, maxULP 0** |
[...]

PR #298, opened 2026-07-11
fix(server): audit correctness/perf/docs hardening + CLI arg validator (#288)

tbraun96:
## Summary

A multi-agent audit of `origin/main` (`f26f392f`) scanned the server for overflow/panic/resource-exhaustion bugs, docs/rustbook drift, and engine improvements; each finding was adversarially re-verified against the code before landing. This PR ships **all** of them, plus the **CLI arg validator** from #288.

Everything is verified: full workspace builds, **all 591 spark-server tests pass**, clippy clean, and the five previously-deferred items were verified **live** on a GB10 (dgx3) serving `nvidia/Qwen3.6-35B-A3B-NVFP4`.

## Correctness / security fixes

| Area | Bug | Fix |
|---|---|---|
| Prefill budget | `max_tokens - 1` underflows to `usize::MAX` (release) / panics (debug) on a `max_tokens:0` scoring call that samples a spontaneous `<think>` token | `saturating_sub(1)` at all 3 promotion sites |
| Chat `n` | Blocking `/v1/chat/completions` read `n` unbounded → alloc/compute DoS | Reject `n∉[1,128]`; fixes a `REQUESTS_ACTIVE` gauge leak on validation 400s too |
| Completions params | `temperature`/penalties passed to the sampler with no range check | Range-validate → 400, mirroring chat |
| Tool parser | Two `&s[..n]` byte slices on raw model output panic on a multibyte char | Char-boundary truncation |
| EP broadcast | `ep_broadcast_tokens` staged the full prompt into chunk-sized scratch → device OOB write | Bound-check against `scratch_bytes` |
| MTP gate | Re-measure read SSM state before the secondary-stream restore landed | `sync_secondary()` + clear drafts at the `MeasureDecode` arm |
| **Swap-out slot double-own** | `swap_out_sequence` compacted a survivor onto the position index (double-own → cross-request GDN state bleed) and overwrote the victim's slot before the save; ignored EP v2 | **Extract retire's proven two-phase exclusive-claim compaction into a shared `compact_survivors_into_range`; reorder swap-out to save → free victim → recompact. Skipped under EP v2.** |

## Performance

- **Streaming stop-string scan** O(n²)→O(n) (bounded straddling-suffix search).
- **Incremental detokenizer** O(n²)→O(n): the chat + completions streaming paths re-decoded the entire token history every token. New `ChatTokenizer::incremental_decode` (vLLM `detokenize_incrementally`) decodes only the suffix window. **Verified byte-identical** to full-decode across an 8-case corpus (ASCII, TOML/code, CJK, emoji + ZWJ family sequences, mixed Cyrillic/Japanese/Korean/Arabic) using the server's own `/tokenize`+`/detokenize`.
- **Per-token buffer reuse** on the host sampling path: the `n·vocab·elem` staging buffer and the per-seq `f32` logits buffer are now per-thread thread-local scratch (fully overwritten each use → byte-identical). Verified live via temp>0 + grammar tool-call output.

## Docs / rustbook

- Corrected stale CLI defaults (`--max-seq-len` 4096→32768, `--mtp-quantization` nvfp4→bf16, `--tool-max-tokens` 512→8192, `--default-min-p` 0.0→0.08), auth model, rate-limiter defaults, tool-parser tables, README/QUICKSTART K=2-vs-K=3 mislabel + wrong parser, CHANGELOG (#170→#180 + 5 missing entries), GpuBackend method count, DeepSeek-V4 in the model matrix, and single-file→module-directory paths.
- **Systemic stale-link sweep** (~35 refs across the book + `AGENTS.md`): `docs/design/*`→`docs/adr/*` + top-level docs, `docs/history/*`→`docs/ATLAS_SPARK_JOURNEY.md`, the nonexistent `moe_dispatch.cu`→`layers/moe/forward_ep.rs` + `moe_w4a16_grouped_gemm.cu`, `scripts/bench-concurrency.sh`→`scripts/run_conc_benchmark.sh`, `CONCURRENCY_SWEEP_RESULTS.md`→`bench/`. Every replacement path verified to exist.

## CLI arg validator (#288)

New `cli::validate_serve_args` — a single place that turns contradictory flag combinations into **hard errors** (never warnings), collected and reported together, each with **what / why / fix** shaped for humans and AI agents, run **before** the multi-minute model load. Rules cover the issue's example (`--fp8-kv-calibration-tokens` without `--kv-cache-dtype fp8`) plus: `--kv-high-precision-layers` on a bf16 cache, `--require-auth` with no token, `--num-drafts` without a speculative method, `--disable-thinking` + thinking budget, `rank`/`ep-size`/`tp-size` vs `world-size`, orphaned high-speed-swap options, `gpu-memory-utilization` range, and enum-value typos (with the valid set listed). Unit-tested (11 tests).

Example output:
```
Atlas CLI: 2 invalid flag combinations — fix before serving:

  [1] --fp8-kv-calibration-tokens 256 has no effect with --kv-cache-dtype bf16.
      why: online FP8 KV-scale calibration only feeds an FP8 KV cache; with a
           bf16/nvfp4 cache the calibrated scales are never read.
      fix: set --kv-cache-dtype fp8, or drop --fp8-kv-calibration-tokens (0 = off).

  [2] --require-auth is set but no bearer tokens were provided.
      why: with auth enforced and no tokens loaded, EVERY request is rejected 401.
      fix: pass --auth-tokens-file <path> (preferred, 0600) or --auth-token <token>.
```

## Live verification (dgx3, Qwen3.6-35B-A3B-NVFP4)

- **Detok byte-identity**: 8/8 corpus cases byte-identical (incremental == full decode), incl. CJK/emoji/ZWJ.
- **Slowloris**: a header-trickle connection is closed at exactly 30.0s; normal requests work; a 45.8s generation completes uninterrupted (a blanket request-timeout would have wrongly killed it).
- **Host-path (F36/F37)**: temp=0.8 haiku + grammar tool-call produce correct output.
- **Swap-out refactor**: 16 concurrent unique-secret requests under `ATLAS_PREFILL_CODISPATCH=1` → 16/16 own-code-correct, **0 cross-request bleed**, 0 errors, 0 crashes (exercises the shared `compact_survivors_into_range` via retire). Note: forcing an actual disk swap-out needs impractical concurrency here (the KV pool is weight-dominated large), so the swap-out trigger itself was not hit live; the fix is a faithful reuse of the already-proven retire-path logic and the shared function is exercised under co-dispatch.

🔬 Found via multi-agent audit; every finding independently re-verified, and the deferred items verified live before landing.


SeedSource:

### Triage
- Rationale: Documentation and changelog updates only; no code changes affecting inference or quantization logic.
- Serving-path files: `crates/spark-model/src/model/impl_a2.rs, crates/spark-server/src/api/chat/mod.rs, crates/spark-server/src/api/chat_phases.rs, crates/spark-server/src/api/chat_stream/handle_token.rs, crates/spark-server/src/api/chat_stream/state.rs, crates/spark-server/src/api/completions.rs`

### Serve gate
- Build: 34s - Qwen3.6-35B-A3B-FP8 on GB10 node1
- Top-k KL vs main: **0.0010** (no shift vs main) - decode delta **-0.3%**

### GLM-5.2 code review
**Verdict:** Approve with minor follow-up.

**Scope:** This is a docs-only PR — updating references from `docs/design/` and `docs/history/` to the new `docs/adr/` structure and `ATLAS_SPARK_JOURNEY.md`. No code changes.

**Correctness:** The doc-path migrations are internally consistent across AGENTS.md, book chapters, and reading references. The `api.rs` → `api/` and `scheduler.rs` → `scheduler/` updates suggest a module split that should be verified to actually exist in the codebase (diff is truncated, so I can't confirm). The CHANGELOG entries (#293, #291, #257, #278) reference prior merged work — no risk here.

**Concerns:**
1. **Broken link risk:** Multiple new `docs/adr/NNNN-*.md` paths are cited (0004, 0003, 0007, 0011, 0010). These files must exist post-merge or the book build breaks. CI should catch this if mdbook link-checking is configured.
2. **QUICKSTART.md tool-parser change:** `hermes` → `qwen3_coder` in the docker recipe is a behavioral change for users copy-pasting. The prose above it still lists both options, so this is intentional, but it's worth a second look — is `qwen3_coder` correct for the default model in that recipe?
3. **README `--num-drafts` removal:** The explanation paragraph is clear and well-reasoned. Low risk.

**Risk:** Low — docs only. No perf/correctness impact.

tbraun96:
## Benchmark gate results — Gate A: ✅ PASS

Ran `/benchmark-pr` on **dgx3** against `atlas-gb10:pr298` (branch SHA `67d71659`), 35B FP8 flagship, exact PR #278 webserver_ok recipe (`Qwen/Qwen3.6-35B-A3B-FP8`, `--kv-cache-dtype bf16 --lm-head-dtype nvfp4 --speculative --num-drafts 1 --mtp-quantization bf16 --enable-prefix-caching --ssm-cache-slots 256 --kv-high-precision-layers auto`).

| Metric | Result | Gate |
|---|---|---|
| `webserver_ok` | **10 / 10** | 10/10 ✅ |
| `followed_directions` | **10 / 10** | 10/10 ✅ |
| **Σ(wall_time_s)** | **919.9 s** | ≤ 1300 ✅ |

Per-run walls (s): `73.1  105.0  73.0  86.9  78.6  73.5  184.7  80.8  70.2  94.1` — median **80.8**, min **70.2**, one 184.7 outlier still leaves us **380 s under budget**. That's the lowest Σwall I've recorded on this recipe (vs the prior ~1364 s that first cleared the <1500 target on the SSM tail-protect work) — the incremental-detokenizer (O(n²)→O(n)) and the per-token buffer reuse landing on the decode/emission hot path show up here.

### The gate earned its keep 🛡️
It caught a **ship-blocking false positive** in the new CLI arg validator (#288) before this ran: the validator rejected `--kv-cache-dtype bf16 --kv-high-precision-layers auto` — which is exactly the canonical flagship serve recipe. Fixed in `67d71659` (redundant ≠ invalid), with a regression test asserting the flagship recipe validates. Without the benchmark, that would have broken every operator running the standard command.

**Gate B (subset ST-995 BFCL, vs baseline 84.66 / 83.32) is running now** — will post when it lands. Not merging; that's your call.

tbraun96:
## Gate B + C results — a real bug caught, root-caused, and fixed ✅

Running `/benchmark-pr` on dgx3 caught a genuine bug in this PR and I've now fixed it.

### The bug: greedy decoding became non-deterministic
BFCL (35B FP8, temp=0, seed=42, non-streaming) on the PR was **non-deterministic**:

| Run | overall / normalized | per-sample |
|---|---|---|
| **Base** `f26f392f` (×2 builds) | **84.66 / 83.32** | **0/1004 diffs — deterministic** |
| PR `67d71659` run 1 | 84.56 / 83.23 | — |
| PR `67d71659` run 2 | 84.26 / 82.94 | 21/1004 differ vs run 1 |

Two greedy runs of the PR disagreed on 21/1004 samples (borderline-argmax flips), while base reproduces to the byte. So the PR broke temp=0/seed=42 reproducibility — worse than a fixed dip.

### Root cause: the MTP-gate re-measure sync (item 4 of this PR)
The `sync_secondary()` + `pending_drafts.clear()` I added to the `GateStep::MeasureDecode` arm. In a varying-prompt-length workload the throughput gate **re-measures on nearly every request** (the depth regime flips each sample: 4233→1195→2430→5239…), so that hunk ran constantly and reordered the async SSM live-state restore against default-stream compute inside the timing-derived measurement branch — making the shared SSM recurrent state depend on run-to-run GPU timing.

Root-caused via a multi-agent workflow (3/3 adversarial verifiers) + an empirical control leg. F36/F37 buffer reuse were **cleared** — `copy_d2h` is synchronous (`cuMemcpyDtoHAsync_v2` + `cuStreamSynchronize`), so they're provably byte-identical to base.

### The fix (`7d10585c`)
Revert just that hunk → `scheduler/mod.rs` is now **byte-identical to base**; F36/F37 kept. A control image built from exactly this fix scored:

| | overall / normalized | vs base |
|---|---|---|
| **Fixed PR** (`7d10585c`) | **84.66 / 83.32** | **0/1004 diffs — identical to base** |

Byte-identical to the deterministic base ⇒ determinism restored, Gate B passes (≥ baseline).

The stale-read that hunk targeted is a non-BFCL path; if genuinely needed it belongs on the MTP→decode boundary unconditionally (as the batch path already does), not inside the timing-dependent gate-measurement branch — tracked as a separate, benchmark-gated change.

### Final gate scorecard (dgx3, `atlas-gb10:pr298` → fix `7d10585c`)
| Gate | Result |
|---|---|
| **A** — webserver_ok N=10 | ✅ **PASS** — 10/10, 10/10, Σwall **919.9s** |
| **B** — BFCL subset ST-995 | ✅ **PASS** (after fix) — **84.66 / 83.32** = base, deterministic |
| **C** — warm-TTFT | ✅ **PASS** — 264.7/266.7 ms vs base 265.0/267.2 (−0.11%/−0.19%) |

All three gates pass with the fix. Not merging — your call.

PR #302, opened 2026-07-12
fix(dflash): verify correctness: capture in every K path + raw-argmax accept basis

rrstesiak:
## Summary

Fixes two silent accept-rate defects in the DFlash verify path: (1) K=3/K=4/K=γ
verify never captured hiddens for the drafter, so `propose()` conditions on a
stale hidden and accept collapses silently (50%→3% measured); (2) verify judged
acceptance through the rep_pen/DRY pipeline while the drafter proposes on raw
argmax, deflating accept by construction. Also adds an opt-in
`ATLAS_DFLASH_MASKED_VERIFY=1` that restores structural-token masking for the
picks at near-zero cost via a proven fast path. Non-DFlash behavior is unchanged
(byte-verified). First of five PRs upstreaming my DFlash work; this one is the
measuring instrument that makes later accept-rate claims trustworthy.

## What changed

- **`416311e5`**: capture one-liners in `verify_c.rs` (K=3), `verify_c2.rs`
  (K=4), `verify_d.rs` (K=γ), mirroring `verify_b.rs:315`, inside the graph
  capture region. Unreachable under `--dflash` today (proposer caps at one
  draft, so K=2 only); becomes load-bearing when the γ engine lands (next PR).
  Reviewer rule: any new K-verify path must wire `try_dflash_capture` before
  its accept numbers mean anything.
- **`5f29ace4`**: thread `dflash_verify_raw_argmax` (= `--dflash`) from
  serve to scheduler to all four verify steps; DFlash accepts against the raw
  GPU argmax (GOLD basis), MTP keeps the pipelined verify unchanged. Adds the
  masked-verify opt-in plus a chat fast path
  (`verify_pipeline_helper/fast_masked.rs`, split as a submodule to respect
  the 500-LoC cap; kill-switch `ATLAS_DISABLE_FAST_MASKED=1`).
- **`11d9795f`**: found by this PR's own A/B validation. The fast path
  originally fired for MTP too, and GPU-vs-host argmax tie-breaking flipped
  near-tie tokens at temp 0. Now gated to masked-verify mode; MTP takes the
  slow path unconditionally, byte-invariant by construction.

## Test plan

- [x] `cargo clippy --workspace --tests` (CI-exact invocation, clean)
- [x] SPDX headers verified on all 12 touched files incl. the new one
  (license-eye image is amd64-only and won't exec on GB10/ARM; CI runs the
  authoritative check)
- [x] `typos`, `cargo test -p spark-server -p spark-model --locked`
  (GPU-free CI variant), file-size cap; `cargo deny` n/a (no `Cargo.*` changes)
- [x] Tested against real hardware (GB10/sm_121f, Qwen3.6-27B-NVFP4, temp 0):
  - **MTP byte-invariance A/B** vs a main-built (`f26f392f`) binary: MinHeap
    code prompt content byte-identical ×2; Volvo prose byte-identical at
    steady state ×2; think blocks byte-identical in every run.
  - **`--dflash` stub parity:** reproduces main's pre-existing stub state
    exactly (K=2-capped, 0% accept per main's own drift-gauge warning); no
    crash, clean output.
  - **Masked-verify probe:** fast-path `ACTIVE` log line fires, zero leaked
    structural specials, pace unchanged vs unmasked.
- [ ] Agentic-harness distributions accompany PR 3 of this train (where perf
  is claimed); this PR makes no perf claims.

## Notes for reviewers

- The raw-argmax basis is gated on `--dflash`, which no MTP configuration
  sets. MTP semantics (Phase C-2 pipelined verify) are intact and
  byte-verified.
- One engine observation from the A/B protocol, relevant to anyone benching:
  prose prompts at temp 0 sit on near-tie tokens and are boot-sensitive on
  unmodified main (reproduced across cold boots on a clean `f26f392f`
  build; pre-existing, not from this PR). Code prompts are fully
  deterministic and serve as the invariance witnesses here.

## CLA

### DIFF
--- a/.github/workflows/file-size-cap.yml
+++ b/.github/workflows/file-size-cap.yml
@@ -247,6 +247,13 @@ jobs:
             # extracted to trait_decode_batched_conv_gdn.rs. Clean split
             # tracked as follow-up.
             "crates/spark-model/src/layers/qwen3_ssm/trait_decode_batched.rs"
+            # -- #300/#301 (mixed-precision NVFP4 weights) carry-over --
+            # 522 LoC; landed on main 2026-07-12 without an entry, so the
+            # check is red on main and fails every fresh PR. Entry added by
+            # the first affected PR per the #229 carry-over precedent above;
+            # not this PR's code. Clean split tracked as follow-up by the
+            # fp8_lut authors.
+            "crates/spark-model/src/weight_map/fp8_lut.rs"
           )
 
           violations=0
--- a/crates/spark-model/src/model/trait_impl/verify_c.rs
+++ b/crates/spark-model/src/model/trait_impl/verify_c.rs
@@ -249,6 +249,14 @@ impl TransformerModel {
                         stream,
                     )?;
                 }
+                // DFlash hidden capture for ctx conditioning. Capture from
+                // the LAST verified position (K-1), mirroring verify_b.rs
+                // (K=2). Without this, every K=3 verify step leaves
+                // `dflash_hidden_save` stale and the next `propose()`
+                // conditions on a repeated old hidden — accept collapses
+                // silently. Must be inside the graph capture region.
+                // No-op when DFlash is disabled.
+                self.try_dflash_capture(layer_idx, k - 1, stream)?;
             }
 
             // Final norm [3, H]
--- a/crates/spark-model/src/model/trait_impl/verify_c2.rs
+++ b/crates/spark-model/src/model/trait_impl/verify_c2.rs
@@ -239,6 +239,14 @@ impl TransformerModel {
                         stream,
                     )?;
                 }
+                // DFlash hidden capture for ctx conditioning. Capture from
+                // the LAST verified position (K-1), mirroring verify_b.rs
+                // (K=2). Without this, every K=4 verify step leaves
+                // `dflash_hidden_save` stale and the next `propose()`
+                // conditions on a repeated old hidden — accept collapses
+                // silently. Must be inside the graph capture region.
+                // No-op when DFlash is disabled.
+                self.try_dflash_capture(layer_idx, k - 1, stream)?;
             }
 
             // Final norm [4, H]
--- a/crates/spark-model/src/model/trait_impl/verify_d.rs
+++ b/crates/spark-model/src/model/trait_impl/verify_d.rs
@@ -245,6 +245,16 @@ impl TransformerModel {
                         stream,
                     )?;
                 }
+                // DFlash intermediate hidden capture: snapshot each capture
+                // layer's output at position k-1 (last verify token) into
+                // dflash_hidden_save[slot] while hidden_states still holds
+                // this layer's activation — mirrors verify_b.rs for K=2.
+                // Must be inside the graph capture region so the per-layer
+                // intermediate (not the final-layer-only post-loop value)
+                // is recorded. Without this, every K=γ verify step leaves
+                // the capture stale and the next propose() conditions on a
+                // repeated old hidden — accept collapses silently.
+                self.try_dflash_capture(layer_idx, k - 1, stream)?;
             }
 
             // Final norm [K, H]
--- a/crates/spark-server/src/main_modules/serve.rs
+++ b/crates/spark-server/src/main_modules/serve.rs
@@ -568,6 +568,16 @@ pub(crate) async fn serve(mut args: cli::ServeArgs) -> Result<()> {
     let scheduler_spontaneous_think_budget = args
         .max_thinking_budget
         .unwrap_or(ptx_set.behavior.max_thinking_budget);
+    // DFlash mode: the drafter proposes on raw argmax, so the verify steps
+    // must judge acceptance on the same (GOLD) basis — skipping the
+    // rep_pen/DRY pre-sample pipeline — or drafter and verifier disagree by
+    // construction and accept craters. ATLAS_DFLASH_MASKED_VERIFY=1 routes
+    // verify PICKS back through the pre-sample masking (unmasked
+    // special-token leak fix); that is handled at the pick sites via
+    // `verify_pipeline_helper::dflash_masked_verify_enabled()` and must NOT
+    // flip this bool — this selects the verify architecture, not the pick
+    // basis.
+    let dflash_verify_raw_argmax = args.dflash;
     std::thread::spawn(move || {
         scheduler::run(
             scheduler_model,
@@ -575,6 +585,7 @@ pub(crate) async fn serve(mut args: cli::ServeArgs) -> Result<()> {
             scheduler_eos,
             max_batch_size,
             use_speculative,
+            dflash_verify_raw_argmax,
             num_drafts,
             policy,
             max_prefill_tokens,
--- a/crates/spark-server/src/scheduler/mod.rs
+++ b/crates/spark-server/src/scheduler/mod.rs
@@ -402,7 +402,13 @@ pub fn run(
                             // only step proposes the first draft and is skipped.
                             let had_draft = !active[0].pending_drafts.is_empty();
                             let t0 = std::time::Instant::now();
-                            step_mtp(&*model, &mut active, num_drafts, &verify_ctx, dflash_verify_raw_argmax);
+                            step_mtp(
+                                &*model,
+                                &mut active,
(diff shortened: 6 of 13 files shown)


Sujimoshi:
Reviewed against base `ebe9728` (full diff + cross-checked the mirrored gate logic in `fast_masked.rs` against `f2_confidence.rs` / `forced_think_end.rs` / `pin_tool_call.rs` at that commit). Overall: the three commits do what they say, the MTP byte-invariance discipline (gating the fast path out of MTP after your own A/B caught the tie-break drift) is exactly the right call, and the capture one-liners are faithful mirrors of `verify_b.rs:315`. Findings below, ordered by severity.

**1. Grammar-constrained requests bypass grammar masks under `--dflash` (default mode).**
The raw-argmax branch in all four verify steps is gated only on `dflash_verify_raw_argmax && !dflash_masked_verify_enabled()` — there is no per-sequence `a.grammar_state.is_some()` check. Unlike ngram/self-spec (`mod.rs` gates both on `grammar_state.is_none()`), the MTP/DFlash step *is* taken for grammar-active sequences — that's the entire Phase C-2 rationale the old comment documents. So with `--dflash` + tool_choice/JSON-schema, verify picks (including the emitted correction token `v0` on reject) skip grammar masking and the ForcedTokenFastPath entirely → grammar desync / malformed structured output. Today's blast radius is small (drafts capped at 1 → K=2 only), but it becomes every token once the γ engine lands. Suggest: in the raw branch, fall through to the pipeline when `a.grammar_state.is_some()`, independent of the env toggle — that keeps the GOLD basis for free-form chat while structured requests stay correct. (The masked-verify fast path already does the right thing here — it refuses on `grammar_state.is_some()` — but it's opt-in.)

**2. Fixed `k-1` capture conditions the next `propose()` on a rejected-branch hidden.**
`try_dflash_capture(layer_idx, k-1, stream)` snapshots the *last verify position* unconditionally, before acceptance is known. On partial/zero accept, position `k-1`'s hidden belongs to a draft token that never entered the sequence, so the next `propose()` conditions on a wrong-universe hidden. This mirrors the existing K=2 convention in `verify_b.rs`, so it's consistent-by-design for a "measuring instrument" PR — but it puts a ceiling on the very accept-rate this PR exists to measure, and it's exactly the defect the ctx-accrual fix in #292/#294 addresses (one ctx slot per accepted position; `ATLAS_DFLASH_EAGLE_FIX`, default-on in #297, measured 14.5%→27.8-35.5% mean accept on identical prompts). Worth stating in the PR text which convention the later PRs of your train adopt, so accept numbers are comparable across both stacks.

**3. T>0 accept basis (scope note, not a defect).**
Acceptance is argmax-equality even at temperature > 0. Fine for this PR's claims (all validation is temp 0), but flagging that #297 adds draft-distribution rejection sampling for the K=2 path — another point where the two trains will want to converge on one design.

**4. Head-on overlap with the open #258→#294→#290→#292→#297 stack.**
This PR's verify-glue is functionally the same change as #297's (same flag name `dflash_verify_raw_argmax`, same threading serve→scheduler→four dispatchers, same pick-site gating) minus stochastic accept, plus masked-verify. The capture one-liners overlap #294/#292. Merging either train makes the other conflict wholesale — happy to coordinate on which base wins so we're not reviewing the same diff twice; the deltas worth preserving from this PR regardless of base are `fast_masked.rs` and the MTP tie-break gating fix.

**5. Minor.**
- `fast_masked.rs` mirrors the f2/forced-think-end/pin-tool arming conditions. I verified they match the source modules at `ebe9728` line-for-line (incl. the hardcoded `thinking_tokens >= 400`, which matches `f2_confidence.rs:35`), and the defer constants are imported from the SSOT rather than copied — good. The 400 literal is still duplicated though; a shared const (or a unit test asserting gate equivalence) would keep the fast path from silently diverging when someone retunes f2.
- Given the fast path makes masking ≈free for grammarless chat, is there a reason `ATLAS_DFLASH_MASKED_VERIFY` shouldn't become default-on once the γ engine lands? Opt-in seems right for this PR (byte-parity with main's stub state), but leaving the leak fix off by default long-term re-opens the T=0 spec-entry derails you root-caused.
- `fp8_lut.rs` allowlist entry: justified per the #229 precedent and clearly attributed; fine here.

tbraun96:
Benchmarked this branch on GB10 hardware. **All three gates PASS** — no regressions. Ran on the tip (`56d8e7ce`, 0 commits behind `main`), images built from the branch and from `main` (`ebe97284`) as the control.

## Gate A — 35B flagship agentic, `webserver_ok` N=10 (dgx3)

| Tier | PR / branch | Σ wall (s) | `webserver_ok` | `followed_directions` |
|---|---|---|---|---|
| **bpr302** | **#302 (this PR)** | **790.5** | **10/10** | **10/10** |
| bpr298 | #298 | 919.9 | 10/10 | 10/10 |
| bpr296 | #296 | 1153.9 | 10/10 | 10/10 |
| pr278repro | #278 | 1146.1 | 10/10 | 10/10 |
| pr278nvfp4 | #278 | 1017.6 | 10/10 | 9/10 |

PASS (gate: 10/10 + 10/10 + Σ ≤ 1300). Per-run walls 70.6–108.2s, run 1 the usual cold outlier.

**Read the 790.5s as "within noise", not as a speedup.** Σwall on this recipe carries ±200–300s at opencode temp 0.3 — it is a blowup detector, not a TTFT gate — and the recent spread above (919.9–1153.9) is itself ~230s wide. This PR is byte-invariant outside `--dflash`, which the Gate A recipe does not set, so there is no mechanism here for a real 14% win. Gate C is the instrument that would see one.

## Gate B — BFCL subset ST-995, full-1004 / seed 42 / temp 0 (dgx3, `Qwen/Qwen3.6-35B-A3B-FP8`)

| Run | `overall_accuracy` | `normalized_single_turn_score` |
|---|---|---|
| **#302 (this PR)** | **84.66** | **83.32** |
| baseline (main, 2026-07-10) | 84.66 | 83.32 |
| Δ | **0.00** | **0.00** |

PASS (gate: ≥ baseline on both). Identical to the digit, and every subset matched as well (`simple_python` 94.57, `parallel` 95.65, `multiple` 94.57, `live_irrelevance` 65.09, `hallucination` 78.98). Bit-for-bit reproduction of baseline is the expected signature of a change that is inert outside `--dflash` — this is the strongest form Gate B can pass in, since it leaves no room for the ±0.4 greedy-nondeterminism noise floor to be hiding a real dip.

## Gate C — warm-TTFT regression guard (dgx1, 27B NVFP4, 3 trajectories / 174 samples, seed 42)

Same box, back-to-back legs; control image built from current `main` (`ebe97284`).

| Leg | completed | failed | median (ms) | avg (ms) | p90 (ms) |
|---|---|---|---|---|---|
| #302 | 174 | **0** | 1418.8 | 1641.9 | 2309.5 |
| `main` control | 174 | **0** | 1432.1 | 1640.1 | 2311.1 |
| **Δ** | | | **−0.93%** | **+0.11%** | **−0.07%** |

PASS (gate: FAIL iff median > +3% or p90 > +5%). All three deltas sit inside the ±1% box noise floor — flat, which is the correct outcome for this diff. This is the same instrument that resolved the ldmab win at −5.8% median / −10.3% p90, so it has the sensitivity to have caught a real perturbation of the MTP verify path had one been introduced. Both legs completed with zero failures (checked before reading any TTFT number — a crashed leg reports fast TTFT over surviving early turns and mimics a win).

Gate C was run because the diff touches the scheduler and the verify path. Deviation worth recording: both legs served at `--gpu-memory-utilization 0.70` rather than 0.85, because dgx1 is unified-memory and OOM-freezes above 0.70. Both legs are identical in this respect, so the relative guard is unaffected.

## Setup notes

- The `main` control image was rebuilt from `ebe97284` rather than reusing an older tag. It compiled fast enough to be suspicious, so I verified it is not a cached copy of the PR binary: the two `spark` binaries differ by SHA, and `ATLAS_DFLASH_MASKED_VERIFY` / `ATLAS_DISABLE_FAST_MASKED` are present only in the PR image.
- Gate A used this branch's own `bench/fp8_dgx2_drift/harness`.
- CI green; `MERGEABLE` / `CLEAN`.

Not merging — leaving that to the maintainers.

PR #304, opened 2026-07-12
feat(dflash): policy layer: adaptive speculation + spec-through-think (stacked on #303)

rrstesiak:
## Summary

The DFlash policy layer: adaptive speculation (suspend below break-even
accept, serial fallback, periodic re-probe), speculation through the think
block, a spec-resume guard, and the serial ctx-append that keeps the drafter
context fed on serially-decoded stretches. Third PR of the train (stacked on
#303); this is the one that turns the #303 engine into a net win. All
behavior is env-gated; defaults preserve #303 semantics.

## What changed

- **Adaptive speculation** (`adaptive_spec.rs`, `ATLAS_DFLASH_ADAPTIVE=1`):
  per-sequence rolling accept window; suspends speculation when mean accept
  drops below break-even, serial-decodes at full pace, re-probes
  periodically. Content decides: code keeps spec engaged, prose suspends.
- **Spec-through-think** (`ATLAS_DFLASH_SPEC_THINK=1`): speculate inside the
  think block (reference engines accept ~45% on thinking tokens; think-gating
  to serial was the single largest throughput loss).
- **Spec-resume guard** (`ATLAS_DFLASH_RESUME_GUARD=N`): serial-decode the
  first N tokens after think exit, where temp-0 low-margin flips concentrate.
- **Serial ctx-append** (`ATLAS_DFLASH_SERIAL_APPEND=1`): append serially
  decoded tokens' hiddens into the drafter ctx accumulator, complement-gated
  so a token is never double-appended. Interim fix: the next PR replaces the
  append family with one unconditional commit point.
- **Step-timing instrumentation** (`ATLAS_DFLASH_STEP_TIMING=1`): per-step
  verify/propose ms split in the log.

## Measured results (GB10/sm_121f, Qwen3.6-27B-NVFP4, temp 0, gamma 16, Done-line metric)

- MinHeap code bench (1500): 24.0 / 24.1 / 24.0 tok/s across three runs
  (engine-only #303 baseline: 20.0-20.7; serial baseline: ~15.8). With a
  short request preceding the bench, the generation takes a different
  near-tie path and posts 26.9-27.0 (see path-sensitivity note below).
  Short-form (300 tok): 22.1.
- Volvo prose bench (1500): 14.9 tok/s, i.e. serial parity within noise
  (recorded parity band 15.35-15.5 vs ~15.8 serial). With the policy layer
  forced off, prose degrades to ~12.5; that delta is why the layer exists.
- Historical bests with this stack on the same rig: 24.0-29.7 code depending
  on generation path; path-to-path variance at temp 0 is a documented
  pre-existing engine trait (near-tie forks), so single-run numbers carry a
  +/-10% path sensitivity. Distributions on the #278 agentic harness are the
  follow-up receipt.

## Explored and documented (receipts in the PR discussion)

- **K-ladder** (downshift to small-K verify instead of serial when accept is
  thin): measured EV is strong (P(first-draft accept) = 0.68-0.70 in the
  worst prose stretches; projected prose ~20+), but a prototype showed the
  small-K routes need capture-all + unconditional ctx commit first. Lands
  after the next PR's commit_ctx. Design and measurements preserved.
- Known reds: two tech-prose prompts historically break at the think seam
  under adaptive re-entry; structurally fixed by the next PR's single commit
  point. Disclosed, not hidden.

## Test plan

- [x] CI preflight (fmt, clippy CI-exact, typos, two-crate cargo test)
- [x] Real hardware receipts as above; outputs clean, zero drift-gauge
  warnings, per-trajectory deterministic
- [ ] #278 agentic-harness distributions (spec vs no-spec tiers) as follow-up
  comment on this PR

## CLA


rrstesiak:
Reproduction recipe for the receipts in this PR (and #303's, minus the policy env vars). Edit paths for your rig.

  ATLAS_DFLASH_DRAFTER_FP8=1 ATLAS_DFLASH_OPTION_B=1 ATLAS_DFLASH_EAGLE_FIX=1 \
  ATLAS_DFLASH_MASKED_VERIFY=1 ATLAS_DFLASH_SEAM_SERIAL=1 \
  ATLAS_DFLASH_SPEC_THINK=1 ATLAS_DFLASH_ADAPTIVE=1 ATLAS_DFLASH_SERIAL_APPEND=1 \
  ATLAS_TARGET_MODEL=qwen3.6-27b RUST_LOG=info \
    spark serve <path-to-Qwen3.6-27B-NVFP4> \
    --dflash \
    --draft-model z-lab/Qwen3.6-27B-DFlash \
    --dflash-gamma 17 --ssm-cache-slots 0 \
    --max-seq-len 4096 --gpu-memory-utilization 0.55 --max-batch-size 1 \
    --kv-cache-dtype bf16 --scheduling-policy slai

  SUGGESTED TEST PROMPTS (temp 0, read the server's Done: line for tok/s):

  time curl -s http://127.0.0.1:8888/v1/chat/completions -H "Content-Type: application/json" -d '{"model":"qwen","messages":[{"role":"user","content":"Write a detailed history of the Volvo
  car company."}],"max_tokens":300,"temperature":0,"stream":false}' > volvo_300_r1.json

  Then the same with max_tokens:1500 → volvo_1500_r1.json.

  time curl -s http://127.0.0.1:8888/v1/chat/completions -H "Content-Type: application/json" -d '{"model":"qwen","messages":[{"role":"user","content":"Write a complete Python MinHeap class
  implementing push, pop, peek, and heapify, with sift_up and sift_down helpers and a docstring for each method."}],"max_tokens":300,"temperature":0,"stream":false}' > minheap_300_r1.json

  Then the same with max_tokens:1500 → minheap_1500_r1.json.

  TESTING NOTE: if you run the max_tokens=300 MinHeap first and then immediately rerun it at max_tokens=1500, the 1500 run posts the fastest general-config result, ~27 tok/s instead of
  ~24. The short run steers the temp-0 fork onto a higher-acceptance trajectory (byte-reproducible; same engine speed per step; documented request-history sensitivity that also reproduces
  on unmodified main).

  SPECIAL CASE, the somewhat boutique code-only peak: identical launch except ATLAS_DFLASH_ADAPTIVE=0, then the MinHeap prompt at max_tokens:1500 → ~29.7 tok/s. Fair warning: the same
  switch drops prose to ~12.5, which is precisely the trade the adaptive layer manages; treat it as a code-workload configuration, not a default.

tbraun96:
Superseded by #313, which is now merged (`b6b17319`).

Closing this one rather than merging it, because the branch here is **stale**: it predates #254 (GDN HeadParallel TP). #303 was rebased onto a post-#254 `main` before it merged; this branch never was, so its merge-base is the original stack point and its diff still carries the pre-squash #303 commits. That is what GitHub was reporting as the conflict.

The trap: squash-merging this as-is — or resolving the conflicts by taking this branch's side — would have **reverted #254** (`multi_rank_protocol_active`, `ATLAS_GDN_DECODE_GRAPH`, `scripts/gdn_tp_golden.py`) as a silent side effect.

So #313 instead replays this PR's three unique commits onto current `main`. They applied with zero conflicts, and the resulting delta is stat-for-stat identical to this PR's own patch (11 files, +275/−28), with your authorship preserved on every commit. #254 is verified intact on `main`.

No work lost — this is your change, just re-based. Thanks for the policy layer; the env-gating discipline is what made the whole stack safe to land.

PR #305, opened 2026-07-12
feat(dflash): unified context commit (commit_ctx), env-gated (stacked on #304)

rrstesiak:
## Summary

Replaces the DFlash context-accumulator's five route-dependent append calls
with one unconditional `commit_ctx(seq, num_committed, base_pos)` invoked at
exactly two commit points: post-accept in the K=gamma verify step, and the
serial/bootstrap decode commit. Every committed token flows through one of
those two paths unconditionally, so no decode route can skip the drafter's
context ledger; the entire ctx-hole bug family (accept collapse after
adaptive re-entries, think-gated stretches, seam paths) becomes structurally
impossible rather than patched per-route. Fourth and final PR of the train
(stacked on #304). Env-gated `ATLAS_DFLASH_UNIFIED_CTX=1`, default OFF;
default behavior is byte-identical to #304.

## What changed

- `commit_ctx` trait method (`traits/model.rs`) + implementation
  (`trait_impl/mod.rs`): position-stamped, watermark-tracked, idempotent per
  commit point.
- The two call sites: `verify_dflash_step` (commits `num_accepted + 1` rows
  at the pre-verify position) and the serial/bootstrap decode commit
  (commits 1 row at the current position).
- When the flag is ON, the legacy appends are bypassed (not yet deleted);
  the deletion pass follows once the flag has soak time, so this PR stays
  small and revertible.

## Why (and why it is NOT a perf claim)

`commit_ctx` is byte-equivalent to the appends it replaces on clean routes;
A/B measured it at zero tok/s delta months of tuning ago (that finding is
part of this stack's ledger). What it buys: correctness on the routes that
historically missed appends, one auditable commit path instead of five, and
the prerequisite for two follow-ups that ARE perf work: the K-ladder
fallback tier (measured 68-70% first-draft accept in prose stretches,
blocked only on unconditional small-K commits) and the position-canonical
drafter-KV addressing planned for the next release.

## Test plan

- [x] CI preflight (fmt, clippy CI-exact, typos, two-crate cargo test)
- [x] Flag OFF (default): MinHeap 1500 bench reproduces #304 receipts
  (24.1 tok/s); zero behavior change.
- [x] Flag ON (`ATLAS_DFLASH_UNIFIED_CTX=1 ATLAS_DFLASH_EAGLE_FIX=0`):
  temp-0 A/B vs flag-off on the same prompt: **byte-identical** (think and
  content, hash-verified) at identical 24.1 tok/s, with the one-shot
  `UNIFIED_CTX ACTIVE` log line confirming commit_ctx engaged. This is the
  designed result: byte-equivalent plumbing, not a behavior change.
- [ ] Extended runbook (ctx-len probe through adaptive toggles on the two
  historically-red tech-prose prompts, agentic-tier soak) scheduled before
  the flag default flips ON; the flag stays OFF until then.

## CLA


rrstesiak:
The DFlash train is fully posted: #302 (verify correctness, merged, thanks Tom), #303 (the K=gamma engine and its verify kernels), #304 (adaptive speculation + spec-through-think, with the 24.1 tok/s receipts and a full reproduction recipe in the comments), and #305 (unified context commit, env-gated, A/B byte-proven). Each PR is individually gated and carries its own test plan; #304's comment has the launch recipe and bench protocol. 




Reproduction recipe for the receipts in this PR (and https://github.com/Avarok-Cybersecurity/atlas/pull/303's, minus the policy env vars). Edit paths for your rig.

ATLAS_DFLASH_DRAFTER_FP8=1 ATLAS_DFLASH_OPTION_B=1 ATLAS_DFLASH_EAGLE_FIX=1
ATLAS_DFLASH_MASKED_VERIFY=1 ATLAS_DFLASH_SEAM_SERIAL=1
ATLAS_DFLASH_SPEC_THINK=1 ATLAS_DFLASH_ADAPTIVE=1 ATLAS_DFLASH_SERIAL_APPEND=1
ATLAS_TARGET_MODEL=qwen3.6-27b RUST_LOG=info
spark serve <path-to-Qwen3.6-27B-NVFP4>
--dflash
--draft-model z-lab/Qwen3.6-27B-DFlash
--dflash-gamma 17 --ssm-cache-slots 0
--max-seq-len 4096 --gpu-memory-utilization 0.55 --max-batch-size 1
--kv-cache-dtype bf16 --scheduling-policy slai

SUGGESTED TEST PROMPTS (temp 0, read the server's Done: line for tok/s):

time curl -s http://127.0.0.1:8888/v1/chat/completions -H "Content-Type: application/json" -d '{"model":"qwen","messages":[{"role":"user","content":"Write a detailed history of the Volvo
car company."}],"max_tokens":300,"temperature":0,"stream":false}' > volvo_300_r1.json

Then the same with max_tokens:1500 → volvo_1500_r1.json.

time curl -s http://127.0.0.1:8888/v1/chat/completions -H "Content-Type: application/json" -d '{"model":"qwen","messages":[{"role":"user","content":"Write a complete Python MinHeap class
implementing push, pop, peek, and heapify, with sift_up and sift_down helpers and a docstring for each method."}],"max_tokens":300,"temperature":0,"stream":false}' > minheap_300_r1.json

Then the same with max_tokens:1500 → minheap_1500_r1.json.

tbraun96:
Benchmark-gated on GB10 (DGX Spark) before merge. All three gates pass.

**Gate A — 35B flagship agentic webserver_ok (N=10, dgx1)**
`webserver_ok` 10/10, `followed_directions` 10/10, Σ wall **1048.2s** (cap 1300).
Per-run walls: 96.0, 120.4, 88.1, 73.5, 137.9, 119.2, 129.7, 111.9, 88.6, 82.9 — no outlier iterations, no repeat tier needed.

**Gate B — subset ST-995 / BFCL (35B FP8, dgx3, full-1004 draw, seed 42, temp 0)**
PR: `overall_accuracy` **84.66**, `normalized_single_turn_score` **83.32**.
Same-box `main` control leg (0e048dd9): **84.66 / 83.32** — identical on all 12 subsets. Matches the recorded baseline exactly.

**Gate C — warm-TTFT regression guard (27B NVFP4, dgx1, same-box back-to-back vs `main` control)**
| | PR | main control | Δ |
|---|---|---|---|
| TTFT median | 1433.8 ms | 1432.7 ms | +0.08% |
| TTFT p90 | 2307.6 ms | 2308.0 ms | −0.02% |
| TTFT avg | 1639.3 ms | 1642.9 ms | −0.21% |

0/174 failed samples on both legs. All deltas are inside the ±1% noise floor — no regression from the default-ON kernel paths.

Gate C was run because this stack is not purely dormant: `ATLAS_FFN_SMALLM`, `ATLAS_GDN_WY17` and `ATLAS_GDN_FUSED_CONV17` all default ON, and the diff edits the shared `kernels/gb10/common/prefill_paged_compute.cuh` that every GB10 target compiles. The serve log confirms `w4a16`, `gdn_verify_fused_conv_kn` and `gated_delta_rule_wy17` load and are used. Those are the parts the gates actually exercise, and they come back clean.

**On scope:** merging this lands the whole stacked K=γ engine, not just `commit_ctx` — 13 commits, 59 files, +6,863/−380, including the new NVFP4 GEMM kernels, the paged drafter and fused verify. Going in with eyes open on that.

I confirmed at runtime that `ATLAS_DFLASH_UNIFIED_CTX` never engages by default and every gated DFlash path stays dormant, so the default behavior really is byte-identical to `main` — which is what the three gates independently corroborate. That also means these gates do **not** exercise `commit_ctx` itself. The extended runbook in the test plan still needs to land before the flag flips ON by default; keeping it OFF until then.

Nice work on this one — the env-gating discipline is what made it safe to land a stack this size.

Issue #164, opened 2026-06-14
[bug] Image inputs in tool-use / tool-result messages not supported


sercand:
### Atlas version / docker tag

latest

### Model

nvidia/Qwen3.6-35B-A3B-NVFP4

### Hardware

1x GB10, single GPU

### Launch command

```bash
sparkrun run @atlas/qwen3.6-35b-a3b-nvfp4 --hosts localhost
```

### Expected behaviour

Images on tool use and tool results should work on agentic use

### Actual behaviour

Only text is supported on tools because Chat Completions API doesn't support it. But Anthropic and OpenAI responses APIs support it: [crates/spark-server/src/anthropic/convert.rs](https://github.com/Avarok-Cybersecurity/atlas/blob/6ff19169ce3ba96c78af8f20b4c842bf8b40a61d/crates/spark-server/src/anthropic/convert.rs#L10)

tbraun96:
@sercand confirmed, real gap, thanks for the exact pointer. two layers to it. 1) plumbing: /v1/messages flattens tool_result blocks to plain text (convert.rs flatten_content -> to_text()), so image blocks are silently dropped before they reach the engine. the internal message shape is chat-completions-like, which has no slot for images in tool turns — we need to carry image blocks from tool_result/tool_use through the same vision-embed path that user-content images already use. 2) model: nvidia/Qwen3.6-35B-A3B-NVFP4 has no vision tower, so even with the plumbing fixed that checkpoint can't consume images on any endpoint. images in tool results will only work on a VL checkpoint (qwen3-vl-30b-a3b is what we serve in-tree today). can your agent stack point at a VL model? if yes we'll scope it as: accept image blocks in tool_result/tool_use on /v1/messages, route through vision encode, and return a clear 400 on text-only models instead of dropping silently. keeping this open as the tracking issue for that work.

AzeezIsh:
Please verify that it's functioning as expected now @sercand :)

sercand:
@AzeezIsh there is a testing script in my PR #307  that tests this particular use case.

PR #307, opened 2026-07-13
feat(server): use shared model-agnostic canonical request/response types

### LINKED ISSUE #165: [feat] Canonical internal chat IR: convert every API surface into one model-agnostic request/response type (unblocks images in tool results)
### Motivation

The motivating bug is small (images attached to tool-result messages are silently dropped), but it is a symptom of an architectural problem: `openai::ChatCompletionRequest` / `IncomingMessage` is the de facto internal representation, and every API surface lowers into it.

- `build_msg_entries` consumes `&ChatCompletionRequest` directly (`crates/spark-server/src/api/chat/msg_entry.rs:55`), so the OpenAI request struct is the hub the whole pipeline reads.
- Anthropic has no typed internal path. `anthropic/translate.rs` builds an OpenAI-shaped `serde_json::Value`, which is then re-parsed as `ChatCompletionRequest` (a struct to JSON to struct round-trip), and the response is read back out of an untyped `Value` (`chat_to_anthropic_response`).
- The OpenAI Responses API adapts into `IncomingMessage` and drops images entirely (`from_responses_input_item`, `crates/spark-server/src/openai/chat_message.rs:109-112`).

Because the canonical shape is OpenAI-specific, every capability that OpenAI's schema does not model cleanly gets bent into it or smuggled through strings:

- Images cannot ride on tool results (the motivating bug); the tool branch hardcodes `image_count: 0` (`msg_entry.rs:126-142`).
- Reasoning/thinking is injected as a `<think>\n\n</think>` string prefix rather than a field (`anthropic/translate.rs:251`).
- Anthropic tool errors are smuggled as a `[tool error]\n` text prefix (`anthropic/translate.rs:209-213`).
- `content` is string-or-array depending on role, with tool/system forced to string in places.

These are all the same root cause: there is no model-agnostic canonical request, so each new feature pays an OpenAI-shape tax and risks the OpenAI/Anthropic protocol drift AGENTS.md calls out as a recurring time-sink.

### Proposal

Introduce a canonical internal chat IR. Each API surface becomes a thin, isolated adapter that converts its wire format into the IR (request direction) and out of it (response direction). The core pipeline (`build_msg_entries`, template rendering, scheduler) consumes only the IR and never an endpoint-specific type.

Key design choice: message `content` is always a list of parts for every role, so text and images interleave uniformly and image-on-tool-result stops being a special case.

```rust
// new: crates/spark-server/src/ir/  (or a shared crate `atlas-chat-ir`)

pub struct ChatRequest {
    pub model: String,
    pub messages: Vec<Message>,
    pub tools: Vec<ToolDef>,
    pub tool_choice: ToolChoice,
    pub sampling: Sampling,            // temperature, top_p, top_k, stop, max_tokens
    pub thinking: Option<ThinkingConfig>,
    pub response_format: Option<ResponseFormat>,
    pub stream: bool,
}

pub struct Message {
    pub role: Role,                    // System | User | Assistant | Tool
    pub content: Vec<ContentPart>,     // always a list; never a bare string
    pub tool_calls: Vec<ToolCall>,     // assistant-emitted
    pub tool_call_id: Option<String>,  // links a Tool message to its call
    pub reasoning: Option<Reasoning>,  // first-class, not a <think> string prefix
}

pub enum ContentPart {
    Text(String),
    Image(ImageSource),
    // open for future: Audio, Video, File, ...
}

pub struct ImageSource {
    pub data: ImageData,               // Base64(String) | Url(String) (Url reserved)
    pub media_type: MediaType,         // image/png, image/jpeg, ...
}

pub struct ToolCall { pub id: String, pub name: String, pub arguments: serde_json::Value }
pub enum Role { System, User, Assistant, Tool }
pub struct Reasoning { pub text: String /* + signature, budget later */ }
```

Adapters (each isolated and unit-testable):

- `openai::chat`: `ChatCompletionRequest` to `ir::ChatRequest`.
- `openai::responses`: Responses input items to `ir::ChatRequest`, carrying images instead of dropping them.
- `anthropic`: `MessagesRequest` to `ir::ChatRequest` directly, deleting the struct to JSON to struct round-trip. The `<think>` and `[tool error]` string hacks become real fields (`reasoning`, and an error flag on the tool result).
- Response direction (later phase): a canonical `ChatResponse` / stream-event type that each surface serializes back to its own wire format, replacing the untyped `Value` reading in `chat_to_anthropic_response`.

How this fixes the motivating feature: with content as a uniform parts list, a tool message is simply `content: [Image(..), Text(..)]`. The collection loop and the Jinja `render_content` macro already iterate parts and emit `<|vision_start|><|image_pad|><|vision_end|>` (`jinja-templates/qwen3_5.jinja:8-18,82`), so images on tool results fall out for free, plus one shared fail-fast validation point (images present and no `vision_config` returns 400).

### Scope

Server / API change

### Alternatives considered

- Narrow fix: lift the tool short-circuit in `msg_entry.rs` and translate Anthropic images at the existing JSON boundary. Smallest diff and ships the feature fast, but it entrenches the lossy OpenAI-shape hub and leaves the Responses image drop, the `<think>` prefix, and the `[tool error]` prefix in place, so every future capability pays the same tax. 
- Extend `ChatCompletionRequest` / `IncomingMessage` in place with richer fields. Keeps a single type but cements an OpenAI-named struct as the canonical model, conflating wire schema with internal schema; Anthropic and Responses still adapt into an OpenAI-shaped thing.

### Additional context

- This is the narrow-waist or canonical-model pattern (one IR, N adapters) used by multi-protocol inference servers to stop surfaces from drifting. It structurally prevents the OpenAI/Anthropic protocol drift AGENTS.md flags as a recurring source of bugs.
- Concrete debt retired by the migration: `anthropic/translate.rs` struct to JSON to struct round-trip; `from_responses_input_item` image drop (`openai/chat_message.rs:109-112`); `<think>\n\n</think>` prefix (`translate.rs:251`); `[tool error]\n` prefix (`translate.rs:209-213`); per-role string-vs-array content handling.
- Open design choice: place the IR as a `spark-server/src/ir/` module versus a shared `atlas-chat-ir` crate. A crate is preferable if anything outside `spark-server` will consume it.
- File-size note: `msg_entry.rs` is already about 300 lines (over the 250-line guideline); the IR migration is a natural point to split it (SDD).

sercand:
_This PR is quite big, sorry for putting everything into single PR._

## Summary

Introduces a canonical, model-agnostic chat IR (`crates/spark-server/src/ir/`) and migrates every API surface to it, so the core pipeline (`build_msg_entries`, template rendering, scheduler) consumes one provider-free representation instead of the OpenAI wire structs. Each protocol becomes a thin, unit-tested adapter pair: `openai/to_ir.rs` / `anthropic/to_ir.rs` on the request side, `openai/encode.rs` + `encode_stream.rs`, `anthropic/translate.rs` (`ir_to_anthropic_response`) + the `translator.rs` streaming state machine, and typed `responses_translate.rs` / `responses_stream.rs` on the response side.

Because `Message.content` is now always a list of parts for every role, the motivating bug — images attached to tool-result messages being silently dropped — falls out for free, along with the rest of the OpenAI-shape tax: the Anthropic struct→JSON→struct round-trip is deleted, the Responses API no longer drops images, and the `<think>\n\n</think>` and `[tool error]\n` string-smuggling hacks become real fields (`reasoning`, an error flag on tool results). The migration commit also fixes a scheduler bug surfaced by end-to-end verification of the tool+image agent loop: the post-think EOS guard fired on every tools-armed turn (eating the model's natural `<|im_end|>` and causing scaffold garbage plus spurious duplicate tool calls); it is now gated on a sticky `think_force_closed` flag so it only applies after a watchdog/budget force-injected `</think>`.

Closes #165 and #164

## Test plan

- [x] Tested against a real model: live on Qwen3.6-35B-A3B-NVFP4 (grammar ON) — `/v1/messages` turn-2 tool replays 4/4 clean `end_turn`; full 2-turn tool+image agent loop 3/3 clean (previously 4 turns with chat-template scaffold garbage and duplicate tool calls); new `scripts/test-qwen36-tool-image.sh` drives the loop end-to-end
- [x] Added or updated tests: `cargo test -p spark-server` → 689 passed, 0 failed (30 lib + 659 bin). New suites: `ir/tests.rs`, `anthropic/tests/ir_carry.rs` (wire→IR→wire carry), `anthropic/tests/translator_stream.rs` (streaming-framing goldens), expanded `openai/tests.rs`; prompt-byte stability gates (`prompt_json_stability_gate`, `rendered_prompt_json_matches_retired_json_hop_output`) pin rendered prompts to the pre-migration output

## Notes for reviewers

- **Invariants the tests pin:** prompt bytes are the KV-cache prefix and wire framing is what agent clients execute, so two golden gates guard the migration — the rendered-prompt JSON gates (template + Anthropic retired-JSON-hop) and the Anthropic streaming goldens. Any future change to message lowering or stream framing should keep these green.
- **Echo-only wire fields** (`service_tier`, `metadata`, `store`, `include_usage`) deliberately do not enter the IR; they ride `api/chat/echo.rs::ResponseEcho` handler-side.
- **Responses API** lowers to the chat wire inside `openai/` before hitting the IR — intra-provider normalization, deliberate; it now carries images instead of dropping them (`from_responses_input_item`).
- **Accepted wire references outside the adapter edges:** `response_store` (caches wire bodies) and `chat_stream --dump` synthesis.
- **Deferred follow-ups:** a neutral `ApiError` type; top-level `reasoning_effort` remains dead wire-compat (never wired into the thinking ladder).
- `count_tokens` now shares `api/chat/prepare.rs::prepare_chat_prompt` with the serving path (measured end-to-end drift 0.0%).

## CLA


sercand:

tbraun96:
@sercand this looks like a really great PR, thanks! I will need to run some benchmarks to ensure no regressions.

crates/spark-server/src/anthropic/handlers.rs
```diff
@@ -68,36 +67,40 @@ pub async fn messages(State(state): State<Arc<AppState>>, body: axum::body::Byte
...
+            let messages_resp = ir_to_anthropic_response(*ir);
+            if let (Some(seq), Some(dump)) = (dump_seq, state.dump_writer.as_ref()) {
+                dump.dump_response("/v1/messages", seq, &messages_resp, false);
+            }
+            return Json(messages_resp).into_response();
         }
+        crate::api::ChatOutcome::Streaming(deltas) => {
+            // Note: streaming dump from the Anthropic side is
+            // best-effort; Anthropic-shape response capture is a
+            // follow-up.
```
tbraun96: We need to ensure we have full Anthropic support

crates/spark-server/src/anthropic/to_ir.rs
```diff
@@ -0,0 +1,248 @@
...
+    ///   (joined with `\n`).
+    /// * user blocks → an optional user message (`[image*, text]`,
+    ///   only when text/images exist) followed by one Tool message per
+    ///   `tool_result` in block order. `is_error` travels as
+    ///   `Message::tool_error` — the `[tool error]\n` marker is
+    ///   rendered by the shared pipeline (`msg_entry`), not baked into
+    ///   the text here.
+    /// * unknown roles collapse to user (Anthropic wire only defines
+    ///   user/assistant).
+    pub fn into_ir(self) -> ir::ChatRequest {
```
tbraun96: The more rusty way of doing this is to `impl From<MessagesRequest> for ir::ChatRequest`. You can keep the comments and the rest of the code, just a small change to help enforce structure and style.

crates/spark-server/src/anthropic/translate.rs
```diff
@@ -1,418 +1,59 @@
...
-///      whenever the Anthropic system field had any non-billing text.
-///
-/// On any mismatch: increment the drift metric, optionally log the
-/// diff (gated by `ATLAS_DEBUG_TRANSLATION_DRIFT`).
-pub(super) fn audit_translation_drift(req: &MessagesRequest, chat_json: &serde_json::Value) {
-    let mut anomalies: Vec<String> = Vec::new();
+/// Serialize the canonical [`crate::ir::ChatResponse`] into Anthropic's
+/// `MessagesResponse` wire shape (non-streaming). Reads the first
+/// choice — the Anthropic adapter pins `n = 1`.
+pub(super) fn ir_to_anthropic_response(ir: crate::ir::ChatResponse) -> MessagesResponse {
```
tbraun96: Rust convention is if a function takes a value (w/o references, etc), use `into` instead of `to`. But, if a function takes a value reference, then just `to` is used. In this case, the function should be called `ir_into_anthropic_response` since `ir: crate::ir::ChatResponse` gets mapped into a `MessagesResponse`. Taking this further, in cases like this, the cleaner way is, like above, to `impl From<crate::ir::ChatResponse> for MessagesResponse`. Given you're using the IR methodology (which is the right way of looking at this), it's best to delegate structure to the compiler via the `From` trait.

crates/spark-server/src/anthropic/translator.rs
```diff
@@ -4,6 +4,22 @@ use axum::response::sse::Event;
...
+    pub(super) data: serde_json::Value,
+}
+
+/// Convert a typed [`SseEvent`] to an axum SSE event for the wire.
+pub(super) fn to_axum_event(e: &SseEvent) -> Event {
+    Event::default()
+        .event(&e.event)
+        .data(serde_json::to_string(&e.data).unwrap_or_default())
+}
+
```
tbraun96:
```suggestion
impl SseEvent {
     /// Convert a reference to self to an axum SSE event for the wire.
     pub(super) fn to_axum_event(&self) -> Event {
         Event::default()
             .event(&self.event)
             .data(serde_json::to_string(&self.data).unwrap_or_default())
     }
}
```

crates/spark-server/src/anthropic/types.rs
```diff
@@ -114,10 +114,43 @@ pub enum ContentBlock {
...
+    pub data: Option<String>,
+    #[serde(default)]
+    pub url: Option<String>,
+}
+
+impl ImageSourceBlock {
+    /// Build the string the vision encoder consumes: a `data:` URI for
+    /// base64 sources, or the raw URL for url sources. `None` when the
+    /// required fields are missing.
+    pub(super) fn to_image_uri(&self) -> Option<String> {
```
tbraun96: Use `maybe` for Rust convention
```suggestion
    pub(super) fn maybe_get_image_uri(&self) -> Option<String> {
```

crates/spark-server/src/api/chat/echo.rs
```diff
@@ -0,0 +1,33 @@
...
+    /// Capture the echo fields from a parsed OpenAI wire request.
+    pub(crate) fn from_wire(req: &crate::openai::ChatCompletionRequest) -> Self {
+        ResponseEcho {
+            service_tier: req.service_tier.clone(),
+            metadata: req.metadata.clone(),
+            store: req.store.unwrap_or(false),
+            include_usage: req.stream_options.map(|o| o.include_usage).unwrap_or(false),
+        }
+    }
+}
```
tbraun96: `impl<'a> From<&'a crate::openai::ChatCompletionRequest> for ResponseEcho`

tbraun96 (dismissed):
To ensure future code follows Rust conventions, it would be best to change things up a bit. There may be more locations in your code where the fixes would also apply.

sercand:
@tbraun96 please have a look at again 🙏

PR #316, opened 2026-07-13
feat(loader): HSS

tbraun96:
This conglomerate PR should supersede multiple PRs once merged.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -29,6 +29,7 @@ crates/spark-server/                         @tbraun96 @azeezish @rsafier @SeedS
 
 # Multi-rank communication (NCCL / RDMA).
 crates/spark-comm/                           @tbraun96 @azeezish @rsafier @SeedSource
+crates/atlas-rdma/                           @tbraun96 @azeezish @rsafier @SeedSource
 
 # Storage / NVMe-backed swap.
 crates/spark-storage/                        @tbraun96 @azeezish @rsafier @SeedSource
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -32,6 +32,7 @@ crates/spark-comm/                           @tbraun96 @azeezish @rsafier @SeedS
 
 # Storage / NVMe-backed swap.
 crates/spark-storage/                        @tbraun96 @azeezish @rsafier @SeedSource
+crates/atlas-tier/                           @tbraun96 @azeezish @rsafier @SeedSource
 
 # Build, CI, governance. Belt-and-suspenders: the broad `.github/`
 # pattern catches everything in the tree, and the more specific entries
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,6 +149,7 @@ jobs:
           int cuMemAllocHost(void **a, unsigned long b) { (void)a; (void)b; return 100; }
           int cuMemAllocHost_v2(void **a, unsigned long b) { (void)a; (void)b; return 100; }
           int cuMemFreeHost(void *a) { (void)a; return 100; }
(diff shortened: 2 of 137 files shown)


tbraun96:
Benchmark-gating this now on GB10. Two build blockers up front — both one-liners, both verified locally. I don't have push access to the fork (`maintainerCanModify=false`), so they're here rather than pushed to your branch.

## 1. The GB10 image does not build (blocker)

`atlas-rdma`'s C shim includes `<infiniband/verbs.h>`, but the Dockerfile's **builder** stage never installs the headers:

```
warning: atlas-rdma@0.1.0: src/rdma_shim.c:20:10: fatal error: infiniband/verbs.h: No such file or directory
error: failed to run custom build command for `atlas-rdma v0.1.0`
```

The **runtime** stage already installs `libibverbs1` (line 63); only the builder is missing `libibverbs-dev`.

```diff
--- a/docker/gb10/Dockerfile
+++ b/docker/gb10/Dockerfile
@@
-    apt-get install -y -qq curl build-essential pkg-config git cmake libclang-dev && \
+    apt-get install -y -qq curl build-essential pkg-config git cmake libclang-dev libibverbs-dev && \
```

With that line, the image builds clean and I'm benchmarking on it now.

**Worth calling out: CI cannot catch this.** `build.rs` deliberately has no header probing ("on Linux with the skip unset and headers missing, cc fails the build loudly"), and CI sets `ATLAS_SKIP_BUILD=1`, which compiles the shim out entirely. So every check on this PR is green while the deployable image is broken. Might be worth a CI job that builds the GB10 image, otherwise this class of break stays invisible.

For the record I did consider dropping the C shim for pure-Rust FFI, and it isn't viable: `ibv_post_send` / `ibv_poll_cq` are `static inline` in `verbs.h` (they dispatch through `qp->context->ops`), so they're not symbols in `libibverbs.so` and can't be `extern "C"`-linked. Every Rust RDMA binding ships a shim for exactly this reason. The shim stays; the Dockerfile is the fix.

## 2. `Build mdBook + rustdoc` fails (blocker)

Both errors are the same line — `[3]` in a doc comment is parsed as an intra-doc link:

```
error: unresolved link to `3`
   --> crates/spark-storage/src/expert.rs:224:22
```

```diff
--- a/crates/spark-storage/src/expert.rs
+++ b/crates/spark-storage/src/expert.rs
@@ -224 +224 @@
-    ///   f32 scale2[3], f32 input_scale[3] (0.0 where absent), zero pad to 256.
+    ///   `f32 scale2[3]`, `f32 input_scale[3]` (0.0 where absent), zero pad to 256.
```

`cargo doc -p spark-storage --no-deps` is clean with that change.

## Also

The branch is 4 commits behind `main`; it merges cleanly (no conflicts) — worth rebasing so CI runs against current `main`.

Gate results (35B webserver_ok, ST-995 BFCL, warm-TTFT, and the dense-27B MLPerf gate) to follow once the runs finish.

tbraun96:
Superseded by #317 — please move review there.

Not a takeover: **all 23 of your commits are carried verbatim with authorship intact**, on top of current `main`. The only additions are the two one-line build fixes from my comment above (`libibverbs-dev` in the GB10 builder stage, and backticking the `[3]` doc-comment array types that were failing rustdoc), plus a clean merge of `main` (this branch was 4 behind; no conflicts).

The only reason for a new PR is mechanical: this fork has `maintainerCanModify=false`, so I couldn't push the fixes to your branch and we couldn't iterate on it in place. #317 lives in the main repo, so we can push fixes directly and get it landed.

Benchmark gates are running now (35B webserver_ok, ST-995 BFCL, warm-TTFT regression guard, and the dense-27B MLPerf gate) against a same-box `main` control; results go in #317.

Thanks for this one — the tiered-storage + RDMA substrate is a big piece of work, and the `atlas-rdma` / `atlas-tier` split into CUDA-free crates is the right shape.

PR #324, opened 2026-07-15
fix(jinja): prevent infinite tool-calling loops in gemma4 multi-turn completions

drewhemm:
### Summary
This PR fixes the infinite tool-calling loop bug in the Gemma 4 chat template by natively supporting the standard OpenAI/vLLM message schema and correctly emitting Gemma 4's native asymmetric turn control tokens.

### Rationale & The Bug
Standard OpenAI-compatible APIs (and client libraries) serialize tool responses as separate messages containing `role: "tool"` and the raw response string in `content` (without any custom `tool_responses` mapping). 

The previous template had two major issues when processing these standard tool messages:
1. **Invalid Turn Formatting:** The template unconditionally wrapped *every* loop message inside conversational turn tokens, emitting an invalid and out-of-distribution `<|turn>tool\n...<turn|>\n` sequence. Gemma 4 was never trained on a `tool` turn role, which pushed the model out of its distribution, leading to hallucinations and text-repetition loops.
2. **Infinite Steering Loop:** Because standard tool messages did not have a `tool_responses` array, the template failed to set the state variable `ns.prev_message_type = 'tool_response'`. This meant that the generation-prompt suffix unconditionally appended the tool-steering prefix `<|tool_call>call:`, forcing Gemma 4 to endlessly generate another tool call on the subsequent turn.

### Key Changes
- **Turn-Wrap Gating:** Added conditional guards to ensure conversational turn delimiters (`<|turn>` and `<turn|>`) are **omitted** for standard `tool` messages, as Gemma 4 expects tool responses to be embedded directly as inline blocks.
- **Native Control Token Formatting:** Implemented standard `tool` role message handling using Gemma 4's native, asymmetric delimiters:
  `{{- '<|tool_response>' -}}{{- message['content'] | trim -}}{{- '<tool_response|>' -}}`
- **State Correction (Loop-Breaker):** Explicitly set `ns.prev_message_type = 'tool_response'` when `role == 'tool'`, cleanly disabling the steering-prefix suffix and allowing the model to return a natural conversational completion.
- **Duplication Safeguards:** Gated the standard `content` parsing and closing tags to skip `tool` messages, avoiding duplicate text rendering inside the history loop.

### Verification
- Tested and verified on a live Grace-Blackwell (GB10) Spark server running the fully-qualified model checkpoint:
  👉 **`bg-digitalservices/Gemma-4-26B-A4B-it-NVFP4A16`**
- Confirmed that the model cleanly processes standard tool outcomes, completely resolves the infinite tool-call loop, and successfully returns a natural conversational closure on the subsequent turn.


drewhemm:

tbraun96:

Verified by content, not by ancestry (Atlas squash-merges, and this branch predates the repo's re-root at "Open Source Release", so it has no merge base with `main` and a three-dot diff is meaningless here): `gh pr diff 324 | git apply --check -R -` applies cleanly against `main` @ `a64b82978`, i.e. every hunk of this PR is already present in the tree.

PR #325, opened 2026-07-15
feat(tool_parser): add native plain-text tool-call streaming parser for gemma4

drewhemm:
### Summary
This PR implements full, native, incremental plain-text tool-call streaming parsing and JSON translation for Gemma 4 checkpoints.

### Rationale
While Atlas currently carries standard parser registrations for Gemma 4, it lacks the traits and streaming mechanics required to compile streamed plain-text tool blocks on-the-fly:
1. Raw streaming boundaries can leak directly to the client's visual text stdout.
2. Streamed unbraced plain-text arguments (`call:NAME{...}`) fail to parse incrementally as standard structured JSON deltas.

### Key Changes
- **`crates/spark-server/src/tool_parser/gemma4.rs`**: Implemented `leak_markers()` for `Gemma4Parser` to register Gemma 4's unquoted envelope open/close delimiters (`<|tool_call>`, `<tool_call|>`) so they are properly stripped.
- **`crates/spark-server/src/tool_parser/streaming.rs`**: Extended `find_args_start()` to track Gemma 4's plain-text calling boundaries.
- **`crates/spark-server/src/tool_parser/streaming_emit.rs`**: Implemented the native translation block for Gemma 4 within `StreamingToolDetector::process`. This captures streamed unquoted arguments, compiles them incrementally into standard structured JSON fragments, and passes them cleanly to downstream execution layers.

### Verification
- Compiled and served Gemma 4 natively on Ada/Grace-Blackwell targets.
- Verified that parallel and sequential tool calls are streamed, intercepted, compiled, and dispatched dynamically with zero text leaks.


tbraun96:
Landed in main via Batch 5 (#490), `7761b9653` — but **partially**, so please check this is what you intended.

Taken: the three `tool_parser` files (`gemma4.rs`, `streaming.rs`, `streaming_emit.rs`). Verified on main.

Not taken, with reasons:
- **jinja-templates/gemma4.jinja** — **already on main**. Both of your hunks (the `role != 'tool'` turn-header guard and the `role == 'tool'` tool_response envelope) are present at lines 187 and 229, landed independently. Nothing was lost.
- **vendor/cudarc/build.rs** — maps CUDA 13.2 → (13, 1) so a newer local toolkit resolves. That is a real problem worth fixing, but it is unrelated to tool-call streaming, and a vendored build script reporting one CUDA version as another deserves its own PR and its own review. Please do open it separately — several of us hit newer toolkits.

Two small fixes were needed on the way, since this branch's CI never ran against our tree: `find_args_start`'s new gemma4 arm was a nested `if let` (this workspace denies `collapsible_if`; rewritten as a let-chain matching the Qwen3-Coder arm below it), and `streaming_emit.rs` was not rustfmt-clean.

The CI situation is the same root cause as the jinja confusion: this branch shares **no merge base** with main (93 commits of unrelated history), so the merge-ancestry guard from #454 rejects it. Rebasing onto main would let it go through normally.

Closing as landed-via-patch. Thanks — native gemma4 tool streaming is a real gap closed.

PR #317, opened 2026-07-13
feat(loader): HSS — tiered storage + atlas-rdma / atlas-tier (re-cut of #316 onto main, +build fixes)

tbraun96:
Re-cut of #316 (@rsafier) onto current `main`, so we can push fixes to the branch directly. **All 23 of Richard's commits are carried verbatim, authorship intact** — this is his work; the only additions are two one-line build fixes (below) and a merge of current `main` (his branch was 4 behind; it merged clean, no conflicts).

Supersedes #316. Opened here only because the fork had `maintainerCanModify=false`, so the fixes could not be pushed to the original branch.

## What it is

HSS / tiered-storage stack: two new crates (`atlas-rdma` — CUDA-free one-sided RDMA verbs + RailSet; `atlas-tier` — CUDA-free tiered-cache core), the expert-pack substrate + residency tiers, a cascade T1 write-back `StorageBackend`, the cache-peer RW blade daemon, Marconi snapshot spill-not-drop reclaim, and an adapter-correct prefix cache + SSM snapshot index.

**138 files, +18,312 / −346.**

## The two fixes added on top

**1. `fix(docker)`: the GB10 image did not build.** `atlas-rdma`'s C shim includes `<infiniband/verbs.h>`, but the Dockerfile's *builder* stage had no rdma-core headers:

```
warning: atlas-rdma@0.1.0: src/rdma_shim.c:20:10: fatal error: infiniband/verbs.h: No such file or directory
error: failed to run custom build command for `atlas-rdma v0.1.0`
```

The *runtime* stage already installs `libibverbs1`; the builder just needed `libibverbs-dev`. One line.

> **CI is structurally blind to this.** `atlas-rdma/build.rs` deliberately does no header probing ("on Linux with the skip unset and headers missing, cc fails the build loudly"), and CI sets `ATLAS_SKIP_BUILD=1`, which compiles the shim out entirely. So every check can be green while the deployable image is broken. **A CI job that actually builds the GB10 image would close this hole** — worth doing as a follow-up, independent of this PR.

We considered dropping the C shim for pure-Rust FFI instead. It isn't viable: `ibv_post_send` / `ibv_poll_cq` are `static inline` in `verbs.h` and dispatch through `qp->context->ops`, so they are not symbols in `libibverbs.so` and cannot be `extern "C"`-linked. Every Rust RDMA binding ships a shim for exactly this reason. Keeping the shim; fixing the Dockerfile.

[...]

PR #306, opened 2026-07-13
fix(loader): dequant NVFP4-packed q/k/v in qwen3_next attention (#299 follow-on)

rsafier:
## Summary

Fixes the **remaining** blocker for serving `RedHatAI/Qwen3-Coder-Next-NVFP4` (and `saricles/Qwen3-Coder-Next-NVFP4-GB10`) reported in #299.

The originally-reported symptom — `shared_expert.gate_proj.weight not found` — **already loads on current `main`**: the variant resolves to `CompressedTensors` and `shared_expert` routes through `quantized_v2`. (The reporter hit it on the `atlas-alpha2` release image, cut before the July-10 `quantized_any` fixes.) Serving on a current build gets **past** the shared_expert step to the first `full_attention` layer, where it then fails on **`self_attn.q_proj.weight not found`** — the follow-on gap this PR closes.

## Root cause

`load_attention` (`weight_map/loaders_fp8.rs`) hardcoded q/k/v to `dense_auto("{p}.q_proj.weight")` — assuming dense BF16 — while only `o_proj` went through `quantized_any`. But these checkpoints quantize q/k/v as compressed-tensors NVFP4 (`weight_packed` + scales, **not** in the `ignore` list, no plain `.weight`). `AttentionWeights.{q,k,v}_proj` are typed `DenseWeight`, so the forward path consumes dense BF16.

## Fix

Branch q/k/v on `.weight_packed`: dequant NVFP4→BF16 into the existing `DenseWeight` (mirrors the `gemma4` loader), leaving the dense-BF16/FP8 path untouched.

Dims are read from the **packed tensor** (`shape[0]`, `shape[1]*2`), not config: under `attn_output_gate` (Qwen3-Next) `q_proj`'s row count is `2×(heads·head_dim)` (e.g. 8192, not 4096), so a config-derived `n` runs the dequant kernel off the end → `CUDA_ERROR_ILLEGAL_ADDRESS`.

## Validation (GB10 / sm_121)

- `RedHatAI/Qwen3-Coder-Next-NVFP4` (44 GB) now **loads all 48 layers** and generates coherently (~80 tok/s decode, valid code + tests).
- **No regression** for dense-q/k/v checkpoints — path is byte-for-byte unchanged when `weight_packed` is absent.
- fmt/clippy clean; `cargo build --release -p spark-server` (all targets, gb10) green.

Closes the serve-blocking half of #299.

PR #329, opened 2026-07-17
fix(w4a16): restore n-fast grid contract on the shared m128 launcher (fixes dense-27B garbage / BFCL 11.26)

tbraun96:
## What broke

#281 swapped the grid axes of `ops::w4a16_gemm_n128_m128` to m-fast (for L2-friendly B-panel reuse) and updated only the `nemotron-h-puzzle` copy of the kernel to read them that way.

That launcher is **shared by every NVFP4 model**. The other 18 `w4a16_gemm_t_m128` kernels still read `blockIdx.x` as the N-block while the launcher fed them an M-fast grid, so every CTA computed the wrong tile.

There is no crash and no error. The model just emits numerically garbage output.

## Impact

On the dense 27B (`unsloth/Qwen3.6-27B-NVFP4`, the MLPerf-edge model) it degenerates into rambling text and stops emitting tool calls entirely:

| metric | baseline | broken main | floor |
|---|---|---|---|
| overall_accuracy | 87.44 | **11.26** | 83.64 |
| normalized_single_turn | 88.53 | **33.33** | 85.32 |

The category split is the tell: `non_live 0.0`, `live 0.0`, but `hallucination 100.0` and `irrelevance 100.0`. It scores perfectly on every category where NOT calling a tool is the correct answer, which is the signature of zero tool calls rather than degraded ones.

This affects **every NVFP4 model** routing through the launcher at `n > 128` (`qwen3_attention/prefill_weights`, `dense_ffn`, `qwen3_ssm`, `multi_seq/qkv`), not just the 27B.

The 35B flagship never surfaced it: `Qwen3.6-35B-A3B-FP8` takes the FP8 launcher and never reaches this path, so gate B scored an exact baseline (84.66 / 83.32) while the MLPerf model was broken.

## Isolation

Bisected on one box, same checkpoint (`ccdaab7e`), same golden serve flags. Only the code differs:

| code | 27B tool call |
|---|---|
| `b45af988` baseline | OK |
| `#303` / `#313` / `#314` dflash | OK |
| `955f0257` (#317 HSS) | OK |
| `d9ff8671` (#315) | OK |
| `287e73d3` (#322) | OK, last good |
| `5ced18a3` (#281) | **garbage, first bad** |
| `a135e68a` (#321) | garbage |
| `45cc093b` main tip | no tool call |
| `45cc093b` + this fix | **OK, 27B coherent again** |

#283 and #306 were both eliminated despite being the obvious suspects.

## The fix

Restore both the shared launcher and the nemotron kernel to n-fast so all 19 kernels and the launcher agree, and write the grid contract down at both ends. The underlying defect was an undocumented implicit convention that #281 forked unilaterally.

## Trade-off: m-fast vs n-fast

The m-fast ordering was a real win and this gives it up:

- **n-fast** (`blockIdx.x` = N-block): every M-block re-reads the whole B panel for its N column. The M-blocks sharing a panel are scheduled `N/128` blocks apart, so the panel is evicted from L2 between them and B is refetched from DRAM once per M-block.
- **m-fast** (`blockIdx.x` = M-block): the M-blocks sharing a B panel run co-resident, so the panel stays in L2 and B streams from DRAM once.

Measured on nemotron: **1k TTFT 494ms -> 464ms**, for free, same weights.

It is given up here because **correctness beats 30ms**, and because the win cannot be taken by mutating a shared launcher. The right way to reclaim it is a **separately named `_mfast` kernel + launcher** so each model opts in explicitly. That is the pattern #281 already used correctly for `w4a4_gemm_mfast` and `fp8_gemm_t_m128_mfast`; this call site simply did not follow it. Reclaiming the 30ms that way is a clean follow-up.

## Gates

Full A/B/C/D suite is running against this branch. Results to follow in a comment.

### DIFF
--- a/crates/spark-model/src/layers/ops/gemm_dense.rs
+++ b/crates/spark-model/src/layers/ops/gemm_dense.rs
@@ -273,6 +273,14 @@ pub fn w4a16_gemm_n128_m128_v2(
 ///
 /// Grid: (ceil(N/128), ceil(M/128), 1)  Block: (128, 1, 1)
 /// SMEM: ~29.8 KB → 3 blocks/SM (vs 5 for m64 at ~19.6 KB).
+///
+/// GRID CONTRACT — N is the FAST axis (blockIdx.x = N-block, blockIdx.y = M-block).
+/// Every `w4a16_gemm_t_m128` kernel across all model dirs reads it this way. This
+/// launcher is SHARED (qwen3_attention, dense_ffn, qwen3_ssm, nemotron_*), so the
+/// axes must NOT be swapped here to suit one model: doing so silently mis-maps every
+/// CTA for the other 18 kernels and produces garbage output with no error. If a model
+/// wants the m-fast (L2-friendly) order, add a SEPARATELY NAMED kernel + launcher
+/// (see `w4a4_gemm_mfast` / `fp8_gemm_t_m128_mfast`) rather than mutating this one.
 #[allow(clippy::too_many_arguments)]
 pub fn w4a16_gemm_n128_m128(
     gpu: &dyn GpuBackend,
@@ -286,7 +294,7 @@ pub fn w4a16_gemm_n128_m128(
     stream: u64,
 ) -> Result<()> {
     KernelLaunch::new(gpu, kernel)
-        .grid([div_ceil(m, 128), div_ceil(n, 128), 1])
+        .grid([div_ceil(n, 128), div_ceil(m, 128), 1])
         .block([128, 1, 1])
         .arg_ptr(input)
         .arg_ptr(weight.weight)
--- a/kernels/gb10/nemotron-labs-3-puzzle-75b-a9b/nvfp4/w4a16_gemm.cu
+++ b/kernels/gb10/nemotron-labs-3-puzzle-75b-a9b/nvfp4/w4a16_gemm.cu
@@ -990,16 +990,16 @@ void w4a16_gemm_t_m128(
     __nv_bfloat16* __restrict__ C,
     unsigned int M, unsigned int N, unsigned int K
 ) {
-    // Grid axes are SWAPPED relative to the other GEMMs: blockIdx.x is the
-    // M-block, blockIdx.y is the N-block.
+    // GRID CONTRACT — N is the FAST axis, matching the SHARED launcher
+    // `ops::w4a16_gemm_n128_m128` and every other model's `w4a16_gemm_t_m128`.
     //
-    // Every m-block re-reads the whole B panel for its n-column. With n on the
-    // fast axis, the 8 m-blocks sharing a B panel are scheduled ~141 blocks
-    // apart, so the panel is evicted from L2 between them and B is fetched from
-    // DRAM 8x (~20 GB across the 112 calls in one 1k prefill). Putting m on the
-    // fast axis makes those 8 blocks co-resident, so 7 of the 8 reads hit L2.
-    const unsigned int cta_n  = blockIdx.y * N_TILE_LG;
-    const unsigned int cta_m  = blockIdx.x * (2 * M_TILE);  // base row for this block
+    // Do NOT swap these to m-fast here. This kernel is reached through a launcher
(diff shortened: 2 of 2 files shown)


tbraun96:
## Gate results — all four PASS

Run on GB10, image built `--no-cache` from this branch and binary-verified distinct. Golden serve flags throughout. Same-box / same-checkpoint isolation where relevant.

| Gate | Result | Detail |
|---|---|---|
| **A — 35B webserver_ok** (dgx1) | ✅ PASS | 10/10 webserver_ok, 10/10 followed_directions, Σwall 1158s (≤1300) |
| **B — 35B BFCL** (dgx3) | ✅ PASS | 84.66 / 83.32 — exact baseline, per-category identical (non_live 85.05 / live 85.92 / hallucination 78.98) |
| **C — warm-TTFT 27B** (dgx1) | ✅ PASS | median +1.0%, p90 +1.5% vs the last-good `b45af988` control (both n-fast; ≤3% / ≤5%) |
| **D — 27B golden BFCL** (dgx2) | ✅ PASS | **87.14 / 88.44** — n=995, complete, `Draining 995 in-flight` |

### Gate D is the one that matters — the fix restores the MLPerf model

| metric | broken main | this fix | baseline | floor |
|---|---|---|---|---|
| overall_accuracy | 11.26 | **87.14** | 87.44 | 83.64 |
| normalized_single_turn | 33.33 | **88.44** | 88.53 | 85.32 |

Deltas vs baseline are −0.30 / −0.09, inside the ±0.4 27B noise floor (MTP greedy nondeterminism). Comfortably above the MLPerf-edge floor.

The category split confirms the mechanism rather than a lucky aggregate:

| category | broken | fix |
|---|---|---|
| non_live | 0.0 | 83.4 |
| live | 0.0 | 86.47 |
| hallucination | 100.0 | 95.45 |

The broken run's "perfect 100 on hallucination" was the model refusing to call any tool at all. The fix trades that fake 100 for real 83.4 / 86.47 by calling tools correctly again.

### On Gate C — no TTFT cost here

I expected a ~30ms regression on Gate C from surrendering the m-fast ordering, and there isn't one. The m-fast win was nemotron-specific; the dense 27B this gate runs was never on the m-fast path (it was only being *broken* by it). Both the fix and `b45af988` are n-fast on the 27B, so Gate C reads as a wash (+1% = noise). The ~30ms surrender is real but lives on nemotron's TTFT, not here, and is reclaimable via a separately named `_mfast` kernel + launcher (follow-up).

### CI

19 checks pass / 1 skip (post-merge deploy), including cargo test --workspace, clippy --tests, fmt, the ≤500-LoC cap, rustdoc, typos, SPDX, cargo deny.

tbraun96:
## Correction: the m-fast revert costs Nemotron ~0, not ~30ms

The PR body estimated the n-fast revert would cost nemotron ~30ms (464 -> 494ms). Measured, it does not. Same-box (dgx1) back-to-back A/B, nemotron-h-puzzle 75B-A9B NVFP4, 3 runs each:

| prompt | fix (n-fast) | main45cc (m-fast) |
|---|---|---|
| short TTFT | 154-157 ms | 155-157 ms |
| medium TTFT | 229-233 ms | 230-234 ms |
| 1k TTFT | 410-414 ms | 413-416 ms |
| decode | 33.3-38.9 tok/s | 33.4-38.9 tok/s |

Identical within noise at every length; both match the round-6 nemotron baseline (155 / 409 / 33.2).

Reason: nemotron's m-fast wins live in the SEPARATELY NAMED kernels (`fp8_gemm_t_m128_mfast`, `w4a4_gemm_mfast`) which this PR leaves untouched. The shared `w4a16_gemm_t_m128` launcher reverted here is only a fallback path for nemotron in the final config, so restoring it to n-fast changes essentially nothing for nemotron while fixing every other NVFP4 model. The earlier 30ms figure was extrapolated from an earlier optimization round, not measured on the final state.

Net: correctness restored (Gate D 11.26 -> 87.14) at zero measurable nemotron cost.

PR #330, opened 2026-07-18
perf(decode): w4a16 GEMV 2-chunk ILP + fix K=4 verify crash on native-FP8-GDN ckpts

tbraun96:
Two decode wins for the GB10 line, found via a head-to-head nsys profile against vLLM (same box, same nvidia/Qwen3.6-27B-NVFP4 ckpt, spec off on both): our non-spec kernels are already at parity (12.14 vs 11.82 tok/s) and ~90% of the practical bandwidth bound, and vLLM's entire golden-run edge is speculative decoding. So: squeeze the last latency out of the GEMVs, and unblock our own spec depth.

**1. `perf(w4a16)`: 2-chunk ILP in the decode GEMVs.** Batch-1 decode is DRAM-latency-bound (ncu on a standalone microbench: 71% long-scoreboard stalls, <1% MIO, zero LUT/bank issues). Restructured the k-loop in `w4a16_gemv`/`w4a16_gemv_dual` into two independent chunks with two accumulators (two weight loads in flight per thread), widened dual weight loads 4B→8B, factored the group scale out of the inner FMA block (exact regroup). Microbench +2.9–3.9% per kernel; serve-level 12.141±0.008 → 12.258±0.005 tok/s (+0.96%, N=5 vs N=8, distributions fully separated). gate_up/down bit-exact; lm_head 1-ulp bf16 reassociation.

**2. `fix(mtp)`: K=4 verify crashed with CUDA 700 on native-FP8-GDN checkpoints.** `--num-drafts 3` died on the first verify step: the FP8 arm in `qwen3_ssm/trait_decode_batched.rs` was guarded `num_tokens == 2 || num_tokens == 3`, so K=4 fell through to `dense_gemm` on a NULL weight slot (this ckpt ships GDN FP8-only; the dense/NVFP4 slots are never populated). Widened both guards to `(2..=4)` — `w8a16_gemv_batch4` already handles M≤4. Root-caused via the new `ATLAS_K4_DIAG` eager-verify checkpoints (separate commit, env-gated, default off, zero production impact — mirrors `ATLAS_K2_DIAG`).

**Validation** (dgx2, nvidia/Qwen3.6-27B-NVFP4, util 0.70, full matrix on this exact series @ 56aa136e):

| Config | tok/s | Quality | Stability |
|---|---|---|---|
| non-spec | 12.246 | 8/8 tool probe + coherence | — |
| drafts=1 (golden) | 18.99 | 8/8 + 500-tok coherent | — |
| drafts=3 | 16.52 | 8/8 + coherence | **10/10, zero CUDA errors** (was: instant crash) |

Per-position acceptance and drafts=2/3 economics are unchanged by this PR (acceptance-limited, not verify-cost-limited — measured verify multiplier ≈1.1); a drafter-prefill follow-up addressing the acceptance gap at depth is in progress.

### DIFF
--- a/crates/spark-model/src/layers/dense_ffn.rs
+++ b/crates/spark-model/src/layers/dense_ffn.rs
@@ -95,6 +95,8 @@ pub struct DenseFfnLayer {
     w4a16_gemv_dual_batch3: KernelHandle,
     w4a16_gemv_batch2: KernelHandle,
     w4a16_gemv_batch3: KernelHandle,
+    /// M<=4 batched GEMV (K=4 verify FFN); 0-handle when absent.
+    w4a16_gemv_batch4: KernelHandle,
     w4a16_gemm: KernelHandle,
     // 128x128 2-stage cp.async pipelined w4a16 GEMM — the fast prefill kernel
     // attention/SSM already use. The base `w4a16_gemm` (M64xN64) only hits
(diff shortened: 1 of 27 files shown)


eisbaw:
Tried this on a GB10 (DGX Spark, sm_121a, CUDA 13.2, built from source). Branch conflicts with main, so this is the PR head as-is:

`spark serve RedHatAI/Qwen3.6-35B-A3B-NVFP4 --port 8888 --lm-head-dtype nvfp4` (no spec)

92.68 tok/s median (5x 256-tok non-stream) vs 96.23 on current main — but another PR head of the same vintage reads ~93 too, so that gap is the base lagging main, not this change. The intrinsic ~1% GEMV-ILP win is below our run-to-run noise on the 35B. Coherent, builds fine (162 s incremental). A rebase would make it measurable.

tbraun96:

```
-115,442 lines deleted
```

- It is **CONFLICTING**, and GitHub does not run merge-commit workflows for a PR it cannot merge — so every green tick displayed here describes a tree from 2026-08-06 that nobody will merge. Thirteen open PRs are in that state; #427 adds a `needs-rebase` label so the false green stops being invisible.

PR #332, opened 2026-07-18
perf(decode): lm_head batched-GEMV (35B + 27B) + Qwen3.6-27B decode bring-up (11->58 tok/s)

rsafier:
## Summary
lm_head batched-GEMV (both models) + Qwen3.6-27B decode bring-up. All validated on GB10, greedy byte-identical.

### lm_head batched-GEMV — model-agnostic (`ATLAS_LMHEAD_BATCH_GEMV`, default off)
Routes the NVFP4 lm_head through `w4a16_gemv_batch16` (reads the ~254 MB / 248K-vocab weight ONCE for all decode rows instead of the tiled `w4a16_gemm` re-tiling it). Gated only on `lm_head_nvfp4 && n<=16` — one change, both models. Also extended to the `decode_b2` mixed/MTP-verify path.
- 35B-A3B C=8: **+9–13%** (+12.6% on the Marlin baseline; +9.3% on the main MoE baseline)
- 27B (dense) C=2/4/8: **+17% / +24% / +11%** (bigger — dense model, lm_head is a larger fraction)

### Qwen3.6-27B decode bring-up: 11 -> 58 tok/s (5.3x)
The 27B decoded at ~11 tok/s (WIP model). Three nsys-guided fixes:
1. **GDN decode kernels** — the 27B's model-local `gated_delta_rule.cu` shadow was a stale snapshot missing the four `decode_f32_{norm,conv_norm,strided,strided_norm}` variants -> all 30 GDN layers fell to a scalar fallback. Appended from `common/` (additive; prefill untouched).
2. **Dense-FFN batched decode** — n>=4 multi-seq decode looped a per-token GEMV (weight re-read n×); default dense FFNs to the weight-read-once GEMM (`is_dense()`-gated; MoE untouched). 21.7 -> 57.5 alone.
3. Serve recipe: drop `ATLAS_DECODE_OPT` for the dense 27B (it forced a 5.5×-slower single-warp FFN GEMV).

## Validation
- Greedy byte-identical A/B on both models.
- 27B decode C=1/2/4/8 = 11.9 / 22.3 / 33.7 / 57.5 (lm_head on); prefill ~1000 tok/s.
- Built `ATLAS_TARGET_MODEL=*` (all targets); no missing GDN kernels.
- Verified the FFN GEMM is already the right kernel at M=8 (a batched-GEMV swap regressed C=8 -> reverted).

## Notes
- Branch cut from `main` (no codex Marlin MoE stack — that lives on `perf/decode-c8-push`). lm_head win doesn't depend on Marlin.
- Attention MMA rework from earlier work is NOT here (proven no-win: attention is ~3% of decode).

Mirror of MonumentalSystems/atlas#59.

Generated with Claude Code.


eisbaw:
Independent confirmation on **GB10 / DGX Spark** (sm_121a, CUDA 13.2), **Qwen3.6-35B-A3B** (`RedHatAI/Qwen3.6-35B-A3B-NVFP4`, `--lm-head-dtype nvfp4`), this PR merged onto `main` @ `3e5115e8`, built from source, `ATLAS_LMHEAD_BATCH_GEMV=1`. Non-stream 256-tok requests, distinct sequences (no prefix-cache freebie), coherence-checked, same-session A/B.

Aggregate decode tok/s (default `max_batch=8`):

| | C=1 | C=4 | C=8 | C=16 | C=32 |
|---|---|---|---|---|---|
| main | 93.1 | 122.6 | 149.0 | 151.1 | 150.4 |
| main + #332 | 93.3 | **150.4** | **170.8** | **170.1** | **172.2** |

**+12–23% at C=4..32**, single-session unchanged (96.2 vs 96.4 median), all outputs coherent. Exceeds the claimed C=8 +9–13% on the 35B.

Two extra data points that may be useful:

1. **Batch-cap interplay:** with `--max-batch-size 16`, #332 reaches **174.6** at C=16. With `--max-batch-size 32`, aggregate collapses to ~132 at C=32 — expected, since the gate is `padded_n <= 16`.
2. **The collapse is not the lm_head:** we extended the dispatch to tile n>16 into ≤16-row `w4a16_gemv_batch16` chunks; C=32 stayed at ~131 (same as stock main at batch 32). So whatever caps >16-row decode batches on the 35B MoE lives elsewhere (SSM/QKV per-row fallbacks are our suspects). The `n<=16` gate is therefore the right stopping point for this PR as-is.

Happy to share raw JSONs / the tiling diff. (Benchmarks and analysis run with AI assistance — Claude Code.)

tbraun96:
Closing this, but **one hunk in it is the only thing in either of your decode PRs that main still lacks**, and I'm porting it — details below so the credit and the reasoning are on record.

### Two thirds of this landed, default-ON, with more on top

**lm_head batched GEMV.** `decode_a2.rs:527–535` now carries a three-tier `batch4/batch8/batch16` ladder, **default-ON** with a kill-switch (`ATLAS_NO_LM_HEAD_BATCH_GEMV=1`, see `decode_a2.rs:33–41`), plus the transposed-twin `w4a16_gemm_n128_m128_bf16_ldb` arms at `padded_n >= 5` (`:493–518`) — that `ldb` parameter is the fix for the 248077-row-stride misalignment (odd vocab ⇒ 16B `cp.async` unaligned). This PR offers `batch16` only, gated `padded_n <= 16`, and **default-OFF** behind an env var. We've since made "ship wins as defaults with a kill-switch, not as opt-in flags" a house rule, so the flag alone would have needed changing.

**27B GDN shadow drift.** `kernels/gb10/qwen3.6-27b/nvfp4/gated_delta_rule.cu` on main already has `decode_f32_norm` (`:946`), `_conv_norm` (`:1129`), `_strided` (`:1288`), `_strided_norm` (`:1413`) — plus `_strided_norm_half` and `_strided_norm_smem`, which this PR doesn't have. The `KERNEL.toml` line added here is an identity mapping (stem → same name), a no-op.

**Dense-FFN batched decode.** `qwen3_ssm/trait_decode_multi_seq.rs:232` has `&& self.ffn.is_dense()` inside a *three*-tier ladder (`forward_prefill` at n≥min_n, chunked `forward_km` at 4..8, per-token loop below), with the nsys numbers in-comment. This PR has one flat `forward_batched` arm.

### On @eisbaw's +12–23% — it was real, and main has since captured it

That measurement was taken against main @ `3e5115e8`, and `git show 3e5115e85:.../decode_a2.rs` contains **none** of the ladder symbols. `git log -S w4a16_gemv_batch16_kernel -- decode_a2.rs` returns exactly one commit: `744420904` (#388), which landed after that baseline. So it was an honest independent confirmation of a win that main then implemented itself, default-ON, with the transposed-GEMM arms added.

### The part I'm taking

`decode_b2.rs:60` on main is **still** this:

```rust
for i in 0..padded_n {
    ...
    } else if let Some(ref nvfp4) = self.lm_head_nvfp4 {
        ops::w4a16_gemv(... self.w4a16_gemv_kernel, normed_i, nvfp4, logits_i ...)
```

That is `mixed_final_norm_lm_head` — the co-dispatch head reached from `decode_b.rs:495` via `mixed_forward_dispatch`, i.e. **live continuous-batching traffic**. It re-reads the full vocab weight `padded_n` times per step, while its sibling `decode_a2.rs` got the entire ladder. Your `decode_b2.rs:64–69` guard is exactly the right observation about exactly the right site.

I'll port it as main's *ladder* (batch4/8/16 plus the `_ldb` twin arms) rather than the bare `batch16`, default-ON under the same kill-switch, so the two head sites can't diverge numerically at the same batch width. Credited to you.

Two things I'll be honest about: it's a bandwidth-accounting argument, not a measurement — nobody has A/B'd the mixed path — and any gate on it has to run spec-OFF or on reversed-order pairs, since a single spec-ON pair drifts ±2%.

### Worth knowing for your other PRs

This branch conflicts with main, and a conflicting PR makes CI **silently skip** — the green checks here were never real. Rebase anything you want reviewed.

PR #333, opened 2026-07-18
fix(deepseek-v4): read FP32 attn_sink as float, not bf16 — restores 7 hard-zeroed L0 heads

SeedSource:
## Problem
The DeepSeek-V4-Flash checkpoint ships `layers.N.attn.attn_sink` (the per-head softmax sink `s_aux`) as **F32 `[num_q_heads]`**, but the loader passed the raw buffer through and every DS4F attention kernel declared/indexed it as `const __nv_bfloat16*`. Indexing a 4-byte-per-element buffer with a 2-byte stride makes even head `2k` read the low-mantissa half of fp32 element `k`. For 7 heads `{6,10,12,20,28,34,48}` that misread decodes large-positive, so the sink dominates the softmax denominator and `exp(m - sink) -> 0` annihilates the value accumulator — the kernel writes **exactly zero** for those heads. Surviving even heads are silently perturbed. At token-0 this alone moved the sampled token (Atlas `110137 "Thinking"` vs the reference `2581 "We"`).

## Root cause (proven)
Reading the banked on-device sink bytes through the kernel's own sink arithmetic predicts the zeroed set `{6,10,12,20,28,34,48}` **exactly** (deterministic, no free parameters). A FP32 reference with the correct sink keeps all 64 heads active and matches vLLM head-for-head — so the reference/vLLM are spec-correct and Atlas was misreading the dtype.

## Fix
Canonical contract: `MlaWeights.attn_sink` is **always** a device FP32 `[num_q_heads]` buffer; every sink kernel indexes `const float*`.
- **Loader** (`weight_loader/deepseek_v4/attn_sink.rs::load_attn_sink_f32`): respect `WeightTensor.dtype` — F32 pass-through (the checkpoint case, a byte no-op), BF16 widened once into an owned FP32 buffer, any other dtype fails loudly, missing -> NULL.
- **Kernels** (`const __nv_bfloat16* sinks` -> `const float*`, drop the `__bfloat162float`/widening cast): `inferspark_prefill_512`, `prefill_attn_compressed`, `mla_paged_decode_fp8`, `mla_paged_decode_nvfp4`.

Wrappers pass a raw `DevicePtr` (u64) — ABI/arg order unchanged.

## Verification (GB10, EP=2, NET/IB RoCE)
- Standalone kernel golden (real hardware): fixed -> all 64 heads active (7 heads match the FP32 ref); the old bf16 read reproduces exactly `{6,10,12,20,28,34,48}`.
- Serve, q5 token-0: **`110137 "Thinking"` -> `2581 "We"`, matching vLLM**; 0 zeroed heads; all-64 cos vs FP32 ref **0.999999**; cross-engine L0 attention cos **0.96 -> 0.9994**.
- Layer ladder: L0/L1 (sliding) now clean; the next cross-engine divergence is isolated to L2 (first CSA layer) — the separate rope-mscale issue, **out of scope here**.
- Behavioral (q5/q8, temp=0): both terminate coherently (were runaways).

## Tests
`weight_loader::deepseek_v4::attn_sink` unit tests: bf16->f32 widening is exact; the fixed fp32 read returns the true sink; and a banked-byte regression reproduces the exact seven-head fingerprint.

## Out of scope (explicitly deferred)
CSA RoPE-mscale (L2+), decode-time CSA append, q3 long-generation, hard-limit enforcement, reasoning integration, sampler changes, and live NVFP4-decode serving (covered here by compile + kernel golden).

## CI note
`weight_loader/deepseek_v4/assemble.rs` is 569 LoC — a **pre-existing** >500 file-size-cap condition on the base branch (570 LoC); this PR reduces it by 1 and puts all new logic + tests in the new 158-LoC `attn_sink.rs`. Not introduced by this change.


SeedSource:
Superseded by #341. This PR targeted the abandoned base `fix/v4-nvfp4-decode-sink`; #341 carries the same verified FP32 attn_sink fix (byte-identical diff) rebased onto `main` (base 3ef35883, after #339 merged) and re-verified end-to-end. Closing in favor of #341.

PR #334, opened 2026-07-19
feat: generic GGUF loader + Ternary-Bonsai-27B (2-bit keep-packed on one GB10)

rsafier:
Generic **GGUF loader** + **prism-ml/Ternary-Bonsai-27B** bring-up, as one consolidated PR of clean, dependency-ordered commits (walk it commit-by-commit). Rebased onto current `main` — replays with zero conflicts; the canonical-IR refactor (#307) does not intersect this work (it lives entirely in the weight-loader / kernel / model-layer strata).

## What this adds
Atlas can load **GGUF files like safetensors** (GPU-preferred dequant → BF16), serve the **qwen35 GDN-hybrid** and **NLLB-200 encoder-decoder** families from GGUF, and run **Ternary-Bonsai-27B** (ternary Q2_0 Qwen3.6-27B + vision) natively 2-bit-resident on a single GB10.

## Commit walkthrough (the review points)
1. `feat(gguf)` **generic GGUF weight loader** — container/reader, CPU+GPU dequant (Q8_0/Q4_K/Q6_K/Q2_0), tensor-name map, value transforms, config-from-metadata, sidecar. All new files under `spark-runtime/src/weights/gguf/*` + `atlas-core/src/config/gguf.rs` — additive, reads in isolation.
2. `feat(gguf)` **qwen35 GDN-hybrid GGUF support** — name map + llama.cpp-converter value transforms.
3. `feat(nllb)` **NLLB-200 from GGUF on GPU** — encoder-decoder name map + `spark-nllb` reader; fixes stale docs.
4. `feat(bonsai)` **Ternary-Bonsai-27B text + vision** — qwen35 backbone reuse + mmproj sidecar / clip translate / patch-embed. Functional dequant→BF16 path.
5. `feat(bonsai)` **native Q2_0 keep-packed decode (Tier-1)** — packed 2-bit decode GEMV; 1.52× decode, ~22 GB reclaimed, byte-exact.
6. `perf(bonsai)` **native Q2_0 MMQ prefill (Tier-2)** — keep-packed int8 tensor-core matmul (flag `ATLAS_GGUF_NATIVE_Q2_MMQ`).
7. `fix(bonsai)` **native-Q2 batched decode (Tier-3)** — wire `q2_0_gemv_vec_batchm` into FFN `forward_k2/k3`; unblocks C≥2 concurrent decode.
8. `chore(bonsai)` keep `sizes.rs` + `qwen3_ssm/mod.rs` under the ≤500-LoC cap.
9. `docs(bonsai)` **serving receipt** — validated full-stack config (`docs/TERNARY_BONSAI_RECEIPT.md`).
10. `chore(gguf)` **satisfy CI** — one trailing commit: `cargo fmt` + two clippy-1.93 lints (`manual_repeat_n`, `precedence`) on new test/helper code. Whitespace/parens only, no logic change. (Kept as one commit because folding per-commit conflicts where later commits re-touch the same files.)

Commits 1→7 stack in dependency order; 3 is independent of 2. Each is atomic.

## Measured (single GB10, vs BF16-dequant baseline)
- **3.6× leaner** — 14.1 GB vs 51 GB resident weights
- **~1.4× prefill / ~1.48× decode**; with GDN-FlashInfer (container) **1.70× prefill** at 1449 tok, growing with context
- Coherent text + grounded vision; C=1/2/4/8 concurrent serving works

## CI status on this rebased branch
- `cargo fmt --all --check` ✓
- `cargo clippy --workspace --tests` (`-D warnings`) ✓
- ≤500-LoC file cap ✓ (no new files over cap; the persistent Q2 dequant-scratch pushed `buffers.rs` to 519 → allow-list entry with rationale in `file-size-cap.yml`, clean `arena/getters.rs` split tracked as follow-up)
- SPDX headers on all new `.rs`/`.cu` ✓
- `Cargo.lock` unchanged vs `main` ✓ (`--locked` safe)
- Unit tests: 267 passing across `spark-runtime` / `atlas-core` / `spark-nllb` / `spark-model` (GGUF loader, config, dequant, NLLB, Q2 ops). Full `cargo test --workspace` links `-lnccl` (unavailable in this sandbox) — CI covers it.

## Notes for reviewers
- The Q2 keep-packed hot-path work (commits 5–7) touches existing qwen3 attention/SSM/FFN forward paths — that's the concentrated review surface; commits 1–4 are additive/low-risk.
- `docs/design`-style token-overlay concept also appears in the separate NLLB-trainable-tokens work; no code coupling, dedup tracked as follow-up.

Ported from the working branch on the MonumentalSystems fork. Opened as **draft** pending a fresh GPU end-to-end confirmation on the rebased tip (code is byte-identical to the validated branch modulo the fmt/clippy chore).


tbraun96:
Status, because three weeks of silence on a PR this size is not acceptable and I'd rather say something incomplete than nothing.

**This is not being closed.** I had it on a reject list; an adversarial re-review overturned that, and the reject list is now down to the PRs whose content main genuinely already has (#266, #332) or has already landed by another path (#373, #375 — both closed as *merged elsewhere*, with credit and with the defects their bodies documented filed as #421 and #422).

What I can state now:

- **It builds.** I verified the CUDA compiles under the real production flags (`nvcc --ptx -arch=sm_121f -O3 --fmad=false -DTQ_PLUS_SIGNS --Werror all-warnings`, standalone per file). That matters more than it sounds: **CI has never compiled a `.cu` file** — `ci.yml` sets `ATLAS_SKIP_BUILD: "1"` repo-wide, so every kernel in this PR reached review having been built only on your box. #419 fixes that; the job runs the real `build.rs` for all 22 gb10 targets in ~30 s with no GPU.
- **This branch conflicts with main, and a conflicting PR makes our CI silently skip.** The green checks here were never real. That's a bug in our process, not yours, but it does mean a rebase is the first useful step.

What I still owe you: a file-level review, and a statement of what a merge would need under the rigor we're now applying — specifically a test that fails without the change, not a smoke test.

Two process problems on our side that I've fixed or filed rather than just noting: the benchmark gate was rejecting records via `git merge-base --is-ancestor` in a repo that squash-merges, so every PR read as ungated regardless of content (#420, merged); and the kernel-compile gap above.

If you want to prioritise, tell me which of your open PRs you most want landed and I'll take that one to completion first.

rsafier:
Branch synced with main (merge, no force-push). Post-sync feature payload: **79 files, +9016/−41** vs main — the generic GGUF loader + the Ternary-Bonsai keep-packed tiers, intact. Resolution notes: the Q2 keep-packed arms compose with #388's lever conversion (`ctx.dispatch.fp8_blockscaled_prefill` replaces the env read at the o-proj seam), main's `DecodeMetaLayout` and this PR's Q2 buffers coexist in `buffers.rs` (the Q2 scratch/act buffers joined the exhaustive-destructure release list, so the no-wildcard leak guard keeps its teeth), and the batched-SSM decline now unions main's WHY-logging with this PR's Q2 decline condition. fmt / clippy `-D warnings` / 583 model+runtime tests / 1803 server tests — all green locally. Real CI should now run; ready for your file-level review whenever.

tbraun96:
Prepping this for a Wave 2 gate slot tonight. Two things, one of which needs you.

**CI is red on `Enforce ≤500 LoC per source file`** — three files land over the cap:

| file | this PR | main |
|---|---|---|
| `crates/spark-runtime/src/weights/weights.rs` | 547 | 458 |
| `crates/spark-model/src/buffers/sizes.rs` | 531 | 499 |
| `.../qwen3_attention/prefill/paged_qkv.rs` | 507 | 494 |

I fixed all three by content rather than by raising the cap:

- **`paged_qkv.rs` → 499.** The Q2 prefill dispatch guard was triplicated across `paged_qkv.rs`, `cache_skip_qkv.rs` and `paged_oproj.rs`. Hoisted it into `Qwen3AttentionLayer::try_q2_prefill` in `prefill_weights.rs` — one authority instead of three copies, which is a win independent of the cap.
- **`weights.rs` → 500.** Moved your `packed_q2_tests` module to a sibling `weights/packed_q2_tests.rs`, exact piecewise copy. Note it now sits *exactly* at the cap, so the next one-line addition re-trips it.
- **`sizes.rs` → 524 + allow-list.** Deleted a stale 8-line comment duplicated verbatim from `sizes_q2.rs`; the remainder is the inherent field/ctor/`total()` spread, so I added an allow-list entry next to your existing `buffers.rs` one, same rationale.

**I could not push this to your branch.** `maintainerCanModify` is true, but GitHub refuses maintainer-edit pushes to *organization-owned* forks, and this head is on `MonumentalSystems/atlas` — 403 on both my credential paths. I did not force-push and did not open a competing PR.

The prepared branch is fetchable from this repo:

```
git fetch origin prep/pr334-loc-cap
```

`aae716f4` = your head `84946eba` + a clean merge of current `main` (`e007ff2c`, zero conflicts) + the cap fix. Cherry-pick the two commits, or grant fork write access and I'll push them directly — whichever you prefer.

**Local CI on that branch is green**: clippy 0 warnings, `cargo test --workspace` 4016 passed / 0 failed, `fmt --check` clean.

Two caveats worth stating plainly rather than discovering at gate time:

1. This PR's two most important tests are `#[ignore]` and did **not** run — `gguf_gpu_validate_matches_cpu_oracle` (needs a GB10 + compiled PTX) and `gguf_real_file_ternary_bonsai` (needs the on-disk GGUF). The ~40 CPU-side GGUF tests did run and pass. I'll run both on hardware as part of the gate rather than counting the suite as green without them.
2. Gate set is C2 → A, B, D, C. C2 runs first and is non-negotiable here: this PR adds kernels into the 27B NVFP4 dir, which is precisely the class C2 exists to catch — a shared-launcher break once left the dense 27B emitting garbage while gates A and B both passed.

Separately, `windows-x86_64-amd-hip` is red here and on every open PR; that is not yours, it's a `__nv_cvt_fp8x2_to_halfraw2` mapping missed when #389 landed. Fix is in #441.

tbraun96:
Second CI failure on this branch, also fixed on `prep/pr334-loc-cap` (now `7a3bf845`).

**`release matrix / build windows-x86_64-nvidia-cuda`** fails with:

```
error[E0432]: unresolved import `super::check_oom_guard`
  --> crates\spark-runtime\src\weights\gguf.rs:39:53
   | no `check_oom_guard` in `weights`
note: found an item that was configured out
  --> crates\spark-runtime\src\weights.rs:324:25
```

`check_oom_guard` was re-exported only under `#[cfg(unix)]` — reasonable until now, since its
only consumer was the unix-only fast-weights (O_DIRECT) path. This branch's GGUF loader consumes
it too, and `mod gguf;` isn't cfg-gated, so the import fails on every non-unix target. The
definition in `loader.rs` was never gated; only the re-export was.

Fix is to split the export — the OOM guard is platform-independent and now has a cross-platform
consumer, while `estimate_has_fp8` keeps its `#[cfg(unix)]` since fast-weights is still its only
user outside the loader module:

```rust
pub(crate) use loader::check_oom_guard;
#[cfg(unix)]
pub(crate) use loader::estimate_has_fp8;
```

I proved it rather than assuming it. Forcing the cfg false locally (`#[cfg(any())]`) reproduces
your CI error verbatim — same `E0432`, same symbol, same `gguf.rs:39` — and restoring the split
export builds clean. Worth flagging that **a unix `cargo check` cannot see this bug**, which is
how it reached CI in the first place; it needs either the Windows row or a forced-cfg check.

So `prep/pr334-loc-cap` now carries three things on top of your `84946eba`: the clean merge of
current `main`, the 500-LoC cap fix, and this. Same fetch line as before:

```
git fetch origin prep/pr334-loc-cap
```

Still can't push to your branch directly (403 — org-owned fork), so these are yours to
cherry-pick or I can push if you grant access.

rsafier:
@tbraun96 All three of your commits are adopted — the branch is fast-forwarded to `7a3bf845` verbatim (your head built on ours, so it was a pure FF push; authorship intact, nothing cherry-picked or rewritten).

Access, both layers: `maintainer_can_modify` is now explicitly true on every open cross-fork PR (#206/#334/#335/#350/#372/#439 — standing policy from here on), and since GitHub 403s that path for org-owned forks anyway, you have a **write invitation to MonumentalSystems/atlas** — accept and you can push to any of these heads directly next time.

On the two `#[ignore]` tests: we have a GB10 free right now — running `gguf_gpu_validate_matches_cpu_oracle` (and `gguf_real_file_ternary_bonsai` if the fixture GGUF is on our disk) at `7a3bf845` and will post results here, so your gate slot starts from C2 rather than test triage.

rsafier:
Pre-gate results at `7a3bf845` on our GB10 (dgx-00), so your slot starts clean:

**The two `#[ignore]` tests you flagged — both PASS on hardware:**
- `gguf_gpu_validate_matches_cpu_oracle` (cuda feature, compiled PTX): **ok**, 0.82s
- `gguf_real_file_ternary_bonsai` (the exact on-disk fixture at the hardcoded path, `Ternary-Bonsai-27B-dspark-Q4_1.gguf`): **ok**, 0.11s

**C2-class smoke** (dense-27B mlpinf, the class this PR's new 27B-dir kernels threaten): boot 22s, 3/3 temp-0 facts (incl. the P&P-1813 canary), single tool call exact (`get_weather {"city":"Paris","unit":"c"}`), **parallel tool calls n=2** (the `disable_tool_grammar` case), zero errors in the serve log. Binary built from this branch tip, MTP K=1, production parser flags.

Gate away — A/B/D/C are yours per the plan; shout if you want any of them run on this box instead.

rsafier:
Your "exactly at the cap" warning came true four lines later — the `check_oom_guard` fix put `weights.rs` at 504 and re-tripped the LoC check on `7a3bf845`. Fixed with headroom instead of another shave: `auto_detect_weight_prefix` moved verbatim to `weights/prefix_detect.rs` (re-exported, call sites untouched), `weights.rs` now 471. fmt/clippy/2066 tests green locally; pushed as `e26949f4`. The remaining red should be just the benchmark gate (owed records — your Wave 2 run) and the known `windows-amd-hip` breakage that #441 fixes.

rsafier:
Re-synced onto post-#440 main (clean union: this PR's Q2 keep-packed arms + the wave's #250 BF16-snapshot/TP arms now coexist in `qwen35_dense.rs`). fmt/clippy/597 tests green.

rsafier:
## Serve-level hardware check on dgx-00 (GB10), rebased tip `14c6ca7f` — keep-packed Q2 FAILS TO BOOT on this tip; dequant arm serves but is very slow

This is exactly the "fresh GPU end-to-end confirmation on the rebased tip" the draft status asked for. Built gb10/`ATLAS_TARGET_MODEL='*'`, release. Served from a staged dir per `docs/TERNARY_BONSAI_RECEIPT.md` (config.json + tokenizer + `Ternary-Bonsai-27B-Q2_0.gguf` + mmproj-Q8_0 symlinks), `--kv-cache-dtype bf16`, `--max-seq-len 8192`, `--gpu-memory-utilization 0.80`, temp-0 probes.

### Tier-1 (`ATLAS_GGUF_NATIVE_Q2=1`) — boot ABORT
The loader side works — both expected lines fire:
- `ATLAS_GGUF_NATIVE_Q2=1: keeping id-42 FFN projections packed (group 128)` ✅
- `Weights: 14.10 GB` (receipt's keep-packed number) ✅, mmproj sidecar found, 1184 tensors

then model build dies:
```
Error: Failed to build model
Caused by:
    dense_auto: unsupported dtype PackedQ2_0 { group: 128 } for model.layers.0.mlp.gate_proj.weight
```

**Root cause (from the code on this tip):** the rebase brought main's native-BF16 dense-FFN snapshot path — `weight_loader/qwen35_dense.rs` (~line 294): for `Nvfp4Variant::Bf16Raw` (which a GGUF→BF16 checkpoint detects as, and keep-packed Bonsai still does) it **unconditionally** calls `dense_auto(store, "{lp}.mlp.gate_proj.weight")` to snapshot BF16 FFN weights, *before* the `ffn_q2` keep-packed branch further down is ever consulted — and `quant_helpers.rs::dense_auto` (line 340) has no `PackedQ2_0` arm. Fix shape: skip the `ffn_bf16_snapshot` when `proj_q2_group(...)` is `Some` for the three projections (i.e., compute/reuse the `ffn_q2` predicate before the snapshot).

### Tier-2 (`+ ATLAS_GGUF_NATIVE_Q2_MMQ=1`) — same abort, same line (dies before MMQ is reachable).

### BF16-dequant arm (flag off) — boots and is correct, but far off the receipt numbers
- Boots in 59s, `Weights: 50.96 GB` (matches receipt), GGUF pre-flight sane.
- Facts **3/3** (Paris / 96 / 1813), 400-tok essay coherent — the generic GGUF loader + qwen35 GDN-hybrid map is functionally correct on hardware.
- But: **TTFT 71.7s @ 1999-token prompt (~28 tok/s prefill), decode 4.6 tok/s** — the receipt's baseline for this arm was 554 tok/s prefill / 14.3 tok/s decode. Boot log shows the post-rebase `SSM in_proj_qkv + out_proj via native FP8 prefill GEMM ... NVFP4 kept as structural fallback` path engaging on this Bf16Raw checkpoint, so the rebase's new dense/Bf16Raw arms look like the suspect for the ~20x prefill / ~3x decode regression too, but I did not bisect it.

### Verdict for the draft
The port compiles and the loader/receipt layout are fine, but the rebased tip is **not e2e-healthy**: keep-packed (the PR's headline feature, commits 5–7) cannot boot, and the fallback arm has a large perf regression vs the validated branch. Both look like interactions with main's post-rebase Bf16Raw dense-FFN/SSM paths rather than faults in the ported GGUF code itself.

Logs kept on dgx-00 under `/home/ms/.claude/jobs/5a7bd33d/tmp/serve_334_*.log` (Tier-1 abort, Tier-2 abort, BF16 serve) if useful.

Tested on dgx-00 (GB10, sm_121), single GPU.

rsafier:
Keep-packed boot abort from the serve test: **fixed and re-verified** (`2628000b`). Root cause was rebase interaction, as diagnosed: wave-1's #250 Bf16Raw FFN snapshot runs `dense_auto` on gate/up/down before the Q2 branch, and a GGUF keep-packed store presents as Bf16Raw carrying `PackedQ2_0` tensors — no `dense_auto` arm → abort. Fix = hoist the `ffn_q2` probe above the snapshot and gate the snapshot on it (packed projections aren't BF16; the Q2 arm owns their compute).

Serve retest, same staged Bonsai dir: **boots 48s, 3/3 temp-0 facts, 19.1 tok/s decode** (vs the receipt's 14.3 — above it), keep-packed evidence line present, zero errors. fmt/clippy/331 tests green.

Still open from the test report: the flag-off BF16-dequant fallback's perf regression on this tip (4.6 tok/s decode) — separate from keep-packed, likely the Bf16Raw keep-BF16 arms engaging where pre-rebase the dequant path runtime-quantized to NVFP4; worth a bisect decision (keep-BF16 vs quantize) before gating, though the flagship keep-packed path is what the receipt measures.

rsafier:
Follow-up to the fix above: the ≤500-LoC check went red on the merge ref — branch (492) and main (499) each pass alone, but main's #435 `_snap` lookups + this PR's Q2 lookups union `qwen3_ssm/init.rs` to 513 in the PR merge. Merged main in and split the FP8 setter trio to `init_fp8.rs`, byte-identical to the same split on #448's branch so whichever lands second dedupes cleanly (`448ed8c2`). All checks green now except the benchmark gate, which is just the 5 gate records this PR owes.

tbraun96:
Landed via stack PR #457 (squash-merged to main as 97c98850b, gate cycle 4/4 PASS on dgx3 at frozen sha 130564d0db).

PR #335, opened 2026-07-19
feat(lora): MoE per-expert + router LoRA & embed/lm_head/vocab overlay (correctness-first)

rsafier:
Two PEFT-adapter capabilities on the MoE decoder path (`holo3_1_moe` / `qwen3_6_moe` + dense siblings), **correctness-first, uniform (named rejects everywhere), zero-overhead-when-off**. Re-sliced from a 32-commit working branch into **6 clean, dependency-ordered commits** — walk it commit-by-commit. Rebased onto current `main`; merges clean (the #307 IR refactor doesn't intersect — it's all `spark-model`/`spark-runtime`/`atlas-core`).

## What this adds
- **Feature 1 — MoE per-expert + router LoRA:** accepts `mlp.experts.N.{gate,up,down}_proj` + router `mlp.gate` deltas; folds the router delta onto gate logits before top-k and the expert delta onto expert output, reusing `apply_lora_delta` (base NVFP4/FP8 grouped GEMM byte-identical). Gated `ATLAS_LORA_EXPERTS=1`, **default off**.
- **Feature 2 — embed / lm_head / vocab-extension overlay:** row-diff / override-set with per-token `seq_slot`-routed CUDA kernels (CUDA-graph-safe, correct in mixed-adapter batches). Gated on presence of overlay tensors; non-overlay adapters byte-identical.

## Commit walkthrough (the 6 review points)
1. `perf(decode)` — **non-LoRA peel-off** (rayon host sampling, token-major/exact-N decode). Independent.
2. `refactor(lora)` — **mechanical scaffold, byte-identical no-op**: `MoeLoraRoute` + `ForwardContext.moe_lora_route` / `AttnMetadataDev.moe_row_adapter` threaded through every forward site defaulting off, `reject_decode_lora` ctx param, `cuStreamIsCapturing` capture-abort guard. *This is the review-cost isolator — every hunk is "add a field, default off."*
3. `feat(lora)` — **foundation**: parser/classify/audit for both features (host, zero runtime change).
4. `feat(lora)` — **Feature 2** embed/lm_head/vocab overlay.
5. `feat(lora)` — **Feature 1 P1** MoE expert+router prefill fold (correctness-first, gated).
6. `feat(lora)` — **Feature 1 P2** device-side grouped + decode/verify folds.

(+ two trailing commits: a pre-existing `spark-storage` rustdoc drive-by that unblocks the docs check, and a `chore(lora)` CI-hygiene commit — rustfmt + one ≤500-cap allow-list entry for the compute-heavy `decode_a2.rs` + two rustdoc de-links.)

## e2e validation (the un-draft blocker — DONE)
Validated at the **deploy config** — stacked on the lm_head branch (#332), served on **Qwen3.6-35B-A3B-NVFP4** (40 layers / 256 experts) with the full optimal flag set (`--scheduling-policy slai`, the prefill+decode `ATLAS_*` set, `--lm-head-dtype bf16`), a synthetic per-expert MoE adapter (the design-doc test vehicle — real fused exports are a P3 follow-up):
- **Installs**: adapter accepted → *"packed 10 router/expert pair(s)"*, *"MoE LoRA installed: router=true, 4 expert pair(s)"*; `moe_lora_grouped_down` / `moe_lora_gather_bgmv` / `token_overlay` kernels load and are `used`.
- **Numerically sound**: LoRA-on output coherent (not NaN/garbage).
- **Real, deterministic effect**: greedy base vs LoRA differ, reproducibly (base sha `8755a02a`, LoRA sha `6d8cf3a1`, each stable across runs) — the router+expert deltas measurably steer logits; base path unbroken.
- **≈zero-overhead**: 70.3 (base) vs 71.8 (LoRA) tok/s.

## CI status (this rebased branch)
`cargo fmt --all --check` ✓ · `cargo clippy --workspace --tests -D warnings` ✓ · `cargo doc --workspace --no-deps` ✓ · ≤500 cap ✓ · 108 LoRA unit tests pass. Full `cargo test --workspace` links `-lnccl` (sandbox lacks it) — CI covers it.

## Sequencing note
Shares 5 decode-path files with in-flight #332 (lm_head). This PR is based on `main` and merges clean now; when #332 lands first, this needs a trivial rebase on those 5 files (resolution already validated — that's the stack the e2e ran on). Whichever merges first, the other rebases.

## Follow-ups (named, out of scope)
Fused on-disk expert import (P3), S-LoRA multi-adapter expert route tables (P2), NLLB-overlay dedup, `decode_a2.rs` clean split.

Re-sliced from the working branch on the MonumentalSystems fork. Draft pending final maintainer review.


tbraun96:
Reviewed properly. **Not closing this** — and I owe you an apology first: this and your other large PRs have sat for three weeks with no feedback across ~26k lines. That's on us, not you.

Verdict: **keep, with four scoped changes.** The capability gap it fills is one main names in its own rejection messages:

- `crates/spark-model/src/lora/loading.rs:24` — *"MoE expert MLPs + SSM layers stay rejected by `classify_key`"*
- `crates/atlas-core/src/config/parsers/lora.rs:150-153` — *"REJECT(modules_to_save) … full-weight replacement is unsupported"*

### It compiles, cleanly, under the real flags

I checked rather than assumed, because CI has never compiled a single `.cu` (`ATLAS_SKIP_BUILD: "1"` repo-wide — I'm fixing that in #419). Production flags are `nvcc --ptx -arch=sm_121f -O3 --fmad=false -DTQ_PLUS_SIGNS --Werror all-warnings`, standalone per file, no `-I`:

| file | result |
|---|---|
| `moe_lora_gather_bgmv.cu` | PASS incl. `--Werror all-warnings` |
| `moe_lora_grouped_down.cu` | PASS |
| `token_overlay.cu` | PASS |

All 8 entry points resolve to live `try_kernel` sites (`ops/lora_delta.rs:49-68`, `ops/token_overlay.rs:47-50`). `cargo check --workspace --all-targets`: 0 errors, 0 warnings. Their absence from `KERNEL.toml` is correct — `collect_cu_files` auto-discovers `kernels/gb10/common/*.cu`; that table holds renames only.

### Commit 1 is already on main — drop it

`4046dcad1` (your rayon host sampling + token-major/exact-N decode peel-off, 2026-07-18) has the same content as what's on main today via `744420904` (#388, 2026-08-08): `ATLAS_MOE_LEGACY_PERTOKEN_DECODE` at `trait_decode_multi_seq.rs:318,333` and `ATLAS_PARALLEL_SAMPLE` at `decode_logits_step.rs:70,75`. #388 is **not** an ancestor of this branch, so main's version didn't flow into your PR — you had it three weeks earlier. I can't tell from the graph whether #388 absorbed it or converged independently and I won't assert either. It does explain one of your seven conflicts (`decode_logits_step.rs`) as already moot.

### A latent kernel bug, same family as the lm_head-716 one

`moe_lora_gather_bgmv.cu:181` and `moe_lora_grouped_down.cu:217,228`:

```cuda
const __nv_bfloat16* Arow = xa + (unsigned long long)row * max_rank;
const uint4* A_vec = (const uint4*)Arow;      // 16-byte loads
```

`max_rank` is never rounded to a multiple of 8 — `expert_pack.rs:199,249,265` use it raw for every stride. bf16 is 2 bytes, so a `uint4` needs 8 elements of alignment. At `max_rank = 12` (PEFT `r=12` is common) row 1 starts at byte 24 — **not 16-byte aligned**. Same shape as the `lm_head` 248077-odd-row-stride bug that cost us 5.5% and days to find.

**One correction to that finding, in your favour:** I first thought `K_VEC = max_rank / 8` also silently dropped the remainder. It doesn't — your tail loop at `:206` covers `tail_start..max_rank` correctly. Only the alignment half stands. Rounding the packed stride up to a multiple of 8 fixes it.

### The kernels have no coverage, and your own doc says that's disqualifying

`docs/design/lora-moe-embed.md:141`: *"Only after an oracle passes do we build the fused/S-LoRA kernels."* The `scripts/moe_lora_oracle.py` it specifies isn't on the branch, and no test touches any of the three `.cu` files or compares a fold against `base + scale·B@A@x`. That's the one thing genuinely blocking merge under the rigor we're now holding ourselves to — a fails-without-the-fix test, not a smoke test.

### The four changes

1. **Re-port `decode_a2.rs`.** It's the one genuinely dead file — pre-#388 in three independent ways: `padded_n` batch rungs vs main's `padded_batch_n(n)`; a decode-graph key on `padded_n` that main's own comment calls *"unsound"* and replaced with the SSM-slot vector; and `lora_eager_env()` vs main's `self.levers.lora_eager` (#388 replaced the env recipe with CLI flags). The other six conflicts are mechanical.
2. **Ship `moe_lora_oracle.py` + a delta-parity test** against `base + scale·B@A@x`.
3. **Round the packed stride to a multiple of 8.**
4. **Add the missing bounds guards** `ids[r] < vocab` / `slot < n_override` at `token_overlay.cu:58,60`.

### Two non-blocking notes

- **NVIDIA-only.** `kernels/strix/common/`, `strix-hip/`, `metal/` carry only the older `lora_bgmv`. MoE-LoRA degrades gracefully via `try_kernel` → handle 0, but `lora/overlay_build.rs:280` **hard-errors** (`"token_overlay CUDA kernels are not loaded"`), so on Strix/Metal a `modules_to_save` adapter is a runtime failure rather than a fallback.
- **Staging.** The three kernels land straight in `kernels/gb10/common/` — the shared, promote-last directory — skipping the shadow-first-then-promote convention. Inert here since everything is `try_kernel`-gated, but worth knowing for next time.

If you'd rather I take items 1, 3 and 4 and push them to this branch (as I did on #382), say so and I will — I'd rather work inside your PR than open a parallel one. Item 2 is yours; the oracle is a design decision.

One more thing worth knowing: this branch conflicts with main, and **a conflicting PR makes our CI silently skip** — so the green checks here were never real.

rsafier:
@tbraun96 Answering your question from the sweep directly: **#335 is the one I most want landed.** LoRA is the widest dependency in my queue — the Laguna adapter work and the GGUF keep-packed LoRA parity both stack on it.

On your four items:
- **Items 1, 3, 4** (decode_a2 re-port, stride rounding to ×8, token_overlay bounds guards): yes please — push them to this branch, same as you did on #382. Much appreciated.
- **Item 2** (the parity oracle): agreed it's mine, and it's in flight — `moe_lora_oracle.py` + a delta-parity test against `base + scale·B@A@x`, per the contract in `docs/design/lora-solid.md`.
- **Commit 1** (rayon host sampling): agreed, drop it — #388 landed the same thing. I'll fold that into the branch sync.

One shape question so we do it once: given the house rule from #332, do you want `ATLAS_LORA_EXPERTS=1` converted to default-ON with a kill-switch in this PR, or as a fast-follow once the oracle is green?

rsafier:
Branch synced with main (merge, no force-push — your two July sync merges preserved). What the sync did:

- **Commit 1 dropped** as agreed: `decode_logits_step.rs` resolves to main's side (#388's rayon sampling stands).
- **`decode_a2.rs` takes main's version wholesale** — deliberately leaving your item 1 (the LoRA re-port against `padded_batch_n`/`levers.lora_eager`) clean to land on top; nothing of the old dead file survives to fight it.
- Split-file re-homing: `abort_capture_if_active`'s CUDA body moved into `gpu_impl_graph.rs` (its FFI's post-split home); main's four new files that build `ForwardContext`/`AttnMetadataDev` literals (`decode_a2`, `drafter_prefill`, `speculative`, `verify_e`) get the branch's `moe_lora_route`/`moe_row_adapter` fields with the conservative route-aware values (base Skip; adapter still rejected before the fold on those paths).
- `impl_b1.rs` keeps main's derived `MetaLayout::seq_slot_off()` and the branch's dedicated `moe_row_adapter_buf` upload (it was never a metadata-gap tenant, so the two compose without touching the layout).

fmt / clippy `-D warnings` / 2395 tests green locally. Items 1/3/4 remain yours per the earlier exchange; the oracle (item 2) is in flight and lands as its own additive commit.

rsafier:
**Item 2 delivered — the oracle is on the branch** (`c033186a`, additive commit on the synced head).

- `scripts/moe_lora_oracle.py` (numpy, torch-free): offline `--selftest` (mirrors `key.rs`/`overlay.rs` classification: expert down/gate/up, router delta on pre-top-k logits, attn q/k/v/o, dense FFN, embed/lm_head row-replacement overlay; 20/20 vs a `gen_moe_lora_adapter.py` fixture), `--capture FILE.npz` per-surface parity verdicts at BF16 tolerance, and a live `--base-url` A/B logit-delta mode (per-surface hidden capture isn't exposed server-side — stated in the script).
- `crates/spark-model/tests/moe_lora_delta_parity.rs`: tiny-dims (h=32/inter=16/E=4/r=4→pad8/top_k=2) through the PRODUCTION entry points — `apply_router_lora`, `moe_lora_grouped_down` (incl. chunk-window bit-equivalence + base-skip row + NULL cell), `moe_lora_gather_bgmv` (decode + per-row skip), `apply_expert_lora_sorted`. MockGpuBackend records launches without math, so it's `#[ignore]`-gated like the arm2 suites — **ran on our GB10: 7/7 checks pass BIT-EXACT (max_ulp 0)**.
- Scale convention documented at the source: `alpha/r` (or `alpha/sqrt(r)` rsLoRA, hard-required), A/B stored verbatim, scale applied in FP32 AFTER the BF16 delta rounding (`lora_delta.rs:259-268`).

Three findings the oracle surfaced that contradict the design docs (reported, not changed):
1. **`-1` sentinel disagreement**: attention/overlay resolve `adapter_slot=-1` → ACTIVE adapter (`slot_math.rs:43`), but the MoE fold treats `<0` as base/Skip (`moe_row_adapter.rs:42-43`). A request naming the base model while an adapter is resident gets attn+overlay deltas but no MoE deltas — contradicts lora-solid.md §2-I's one-identity-channel contract.
2. **No per-request "force base" selector exists** (empty name → 400, absent → active), so lora-moe-embed.md §G's base-isolation e2e case isn't drivable over the API while an adapter is resident.
3. Doc drift: lora-moe-embed.md §E.1 still describes P1 as host-loop-only; the device kernels landed (CHECK 7 proves host-loop and device folds agree bit-for-bit).

With this, item 2 is done on our side; items 1/3/4 remain yours per the earlier exchange.

tbraun96:
Items 1/3/4 are done. Three commits on top of your exact tip `22c3e281`:

```
git fetch origin prep/pr335-owed-items     # 86890f3f
```

Local CI is green there (clippy 0, `cargo test --workspace` 0 failures, fmt clean), and the
baseline at your tip was clippy-green first as a control.

**Item 3 — `token_overlay.cu` bounds guards (`3dc8de14`).** This was the one I'd flag hardest.
`slot_map[ids[r]]` was indexed with no `ids[r] < vocab` check and `slot` was tested only for
`< 0`, never against `n_override`. While fixing it I found the **same bug family in
`lmhead_overlay_routed_*`, where it's an out-of-bounds *write*** (CWE-787) rather than a read —
one line each, `vocab` was already a param there.

It is reachable, and the reason is worth knowing: `/v1/completions` validates raw token ids
against `tokenizer.get_vocab_size(true)`, while `slot_map` is sized to `config.vocab_size`. Two
divergent authorities with no cross-check, so any tokenizer/checkpoint mismatch walks straight
off the end. Per PCND the out-of-range policy is explicit rather than a silent clamp: an
out-of-range id **means "no overlay entry"**, identical to `slot_map[id] == -1`, documented in
the kernel — id validity belongs to the base gather, and trapping would kill a serve over one
bad request.

**Item 2 — ×8 stride rounding (`d1a7dc9e`).** One derivation, `packed_stride(max_rank) =
div_ceil(8)*8`, consumed by both the sizing path and `pack_into`/`pack_pair`, so sizing and
packing can't drift. `LoraPair.rank` stays logical `peft.r`; `max_rank` is the derived stride —
never conflated. Mutating the sizing back to the raw rank doesn't just fail an assert, it
**panics with a pool overrun** in the r=12 test, which is exactly the production hazard.

**Item 1 — `decode_a2` fold port (`86890f3f`).** The diagnosis was sharper than "the fold is
inert". The hooks existed; what was missing was the pre-lookup **Refuse guard** in
`decode_batch_compute_main`. A batch routed to a non-active adapter fell through the token-major
reject, got swallowed by the ladder's `.is_err()` fallback, and the row map defensively wrote
those rows as base — i.e. **silently correct-looking output computed with base weights**. That's
worse than a crash. The guard is now a pure `lora::ensure_decode_route_servable` shared with
`mixed_forward`, and `impl_lora.rs` came down 500 → 489 LoC in the process.

**Every change was mutation-proven.** Nine mutations, each killed by a named assertion (one
produced no test output — that was a `deny(warnings)` build failure, not a surviving mutation,
so I fixed and re-ran). Details in the commit messages.

**What I could not verify, stated plainly:** the CUDA guards are **compile-verified only** —
`nvcc -arch=sm_121 -ptx` clean, 4 entries, embed kernel shows the expected 9 params — with no
execution proof, since I had no GPU available. Item 1's call sites are likewise compile/clippy
verified only. The end-to-end proof is your `#[ignore]`d `moe_lora_delta_parity` oracle plus a
C2/B gate run, which is GPU work still owed before this merges.

One thing for your judgement, not a request: the three new kernels landed in
`kernels/gb10/common/`, which `build.rs` merges into **every** gb10 target — so they compile into
all 22 targets and widen the gate blast radius. Shadow-first into the tested model's dir would
have contained that. I did not move them.

Separately, `prep/pr334-loc-cap` carries #334's LoC-cap fix and a Windows-build fix, same
situation — I can't push to an org-owned fork, so both branches are yours to cherry-pick, or I
can push directly if you grant access.

rsafier:
Adopted — the branch is fast-forwarded to `86890f3f` (your three item-commits verbatim on the merged tip; pure FF, authorship intact). With the oracle + GPU delta-parity test already on the branch from our side, all four review items are now IN the PR. Also: you should have a write invitation to MonumentalSystems/atlas — accept and the prep-branch dance is over; `maintainer_can_modify` is likewise now pinned true on every cross-fork PR. Gates whenever you're ready — same offer as #334 to run legs on our GB10.

rsafier:
Three red checks on the merged tip cleared (`7e4d98e7`): the typos hit was ours (`dout_dev` in the parity test → `d_out_dev`); rustdoc was the `expert_router_bytes` doc linking the private `packed_stride` (plain code-span now); and the fold hooks had pushed two exactly-at-cap files over (forward_k2.rs 508, impl_b3.rs 501) — both split sibling-module style with real headroom (492/473) rather than shaved back to the line. fmt/clippy/2411 tests green locally. Remaining red should be the benchmark gate (owed records) and the #441-tracked windows-amd-hip.

rsafier:
Re-synced onto post-#440 main (one trivial mod-list union). fmt/clippy/tests green.

rsafier:
**Serve-level e2e on the current tip (`681b425a`) — two findings, one clearly real, one needs a look.** Setup: 35B-A3B NVFP4 on GB10, `ATLAS_LORA_EXPERTS=1`, generated adapter via `scripts/gen_moe_lora_adapter.py` (r=8, ALL 256 experts on layers 3+7, down+router — the all-expert shape that avoids routing false-negatives). Adapter loads and installs cleanly (`LoRA: slot 0 'e2e' packed 514 router/expert pairs`, `MoE LoRA installed: router=true, 256 experts` ×2).

**Finding 1 — batched decode still rejects, and it batch-kills:** 6 concurrent requests (3 base-named + 3 adapter-named) → all six HTTP 500 from a single `decode_batch error: MoE LoRA (Feature-1) is prefill-only in phase 1`. The item-1 `decode_a2` fold is on the branch, but the scheduler-seam guard in `decode_step` still fires with a resident adapter under multi-seq decode — one reject kills the whole cohort. Either the guard should now route through the ported fold, or (if phase-1 scope intentionally keeps batched-decode folding off) it needs to decline-to-per-seq instead of erroring the batch.

**Finding 2 — greedy A/B is a no-op:** sequential single requests, base-named vs adapter-named, produce byte-identical output. A random all-expert adapter must move greedy logits via the prefill fold alone. No fold-engagement lines appear in the serve log. Could be a real silent no-op on the serve path, or my harness's model-name→slot resolution differing from what the fold expects (the `-1` sentinel asymmetry from the oracle findings is adjacent to this). Kernel-level parity is bit-exact (the in-tree GPU test), so the delta is somewhere between request routing and the fold call sites.

Artifacts: serve log + adapter + exact scripts available; happy to bisect either finding on our box. Given gate C2 wouldn't catch these (no adapter in the recipe), flagging before the gate slot rather than after.

tbraun96:
Landed via stack PR #457 (squash-merged to main as 97c98850b, gate cycle 4/4 PASS on dgx3 at frozen sha 130564d0db).

PR #340, opened 2026-07-20
fix(kernels): move fp8_fp8_gemm_ldmab to common/ — unbreak non-27B fp8-GDN prefill (#337 follow-up)

rsafier:
## Problem

#337 added the default-on GDN-fp8-projection prefill kernel **`fp8_fp8_gemm_ldmab`** (ncu-proven **2.1×** over the scalar-load `fp8_gemm_t`) but placed it **only** in `kernels/gb10/qwen3.6-27b/nvfp4/w4a16_gemm.cu`.

`collect_cu_files` (`crates/atlas-kernels/build.rs`) dedups kernel sources **by file stem**, so every model's own `nvfp4/w4a16_gemm.cu` **fully shadows** `common/w4a16_gemm.cu`. A kernel added to one model's `w4a16` file is therefore invisible to every other model. Since `ops::gemm_fp8_prefill` resolves the kernel with `.expect()`, serving **any other native-fp8-GDN model** (e.g. `qwen3.6-35b-a3b`) **panics at prefill**:

```
fp8_fp8_gemm_ldmab: cuModuleGetFunction failed: CUDA_ERROR_NOT_FOUND (500): named symbol not found
```

The only workaround was `ATLAS_FP8_LDMAB=0` (scalar fallback), which forfeits the 2.1× win.

## Fix

Move the kernel — plus its 3 `cp.async` helpers and the `ATLAS_MMA_E4M3F` macro, **verbatim** — into a new self-contained common module `kernels/gb10/common/w4a16_fp8_ldmab.cu`. Its stem auto-maps to module `w4a16_fp8_ldmab`, which **no model file overrides**, so it compiles for **all** targets. The two Rust lookups are retargeted to the new module; the kernel is deleted from the 27B file. `bf16_to_fp8` (the quantize half of the path) already lives in every `w4a16` module and is unchanged.

Net: `+125` (new common kernel) / `−91` (27B removal) / `2` one-line Rust module renames.

## Validation (GB10)

- **`qwen3.6-35b-a3b`**: now serves **without** `ATLAS_FP8_LDMAB=0` — no panic, coherent output, the ldmab prefill path is active (the `.expect()` resolves).
- **`qwen3.6-27b`**: regression-clean — coherent output, no crash, now sourcing the kernel from the common module.
- Kernel body is **byte-identical** to the 27B-validated source (pure relocation).

## Notes

- `int8_gemm_i32acc` (the other #337 addition to the same file) is loaded via `try_kernel` (graceful handle-0 fallback), so it does **not** crash other models — left in place; a follow-up could relocate it too for the int8 dense-FFN parity win.


rsafier:
```
  Holo-9B (loaded on the existing ornith-1.0-9b target, as you said — no new target needed)

  ┌───────────────────────────────────────┬──────────┐
  │ Holo-3.1-9B (dense, 1637-tok prefill) │   TTFT   │
  ├───────────────────────────────────────┼──────────┤
  │ ldmab ON (#340)                       │ 656.6 ms │
  ├───────────────────────────────────────┼──────────┤
  │ scalar (ATLAS_FP8_LDMAB=0)            │ 798.0 ms │
  └───────────────────────────────────────┴──────────┘

  −17.7% / 1.22×, distributions fully separated, output coherent.

  The full dense-scaling picture (all measured, same probe)

  ┌───────────────┬────────┬────────┬────────────────────────────────────────┐
  │     Model     │ ldmab  │ scalar │                   Δ                    │
  ├───────────────┼────────┼────────┼────────────────────────────────────────┤
  │ Holo-0.8B     │ 128 ms │ 130 ms │ ~0% (too small, memory-bound)          │
  ├───────────────┼────────┼────────┼────────────────────────────────────────┤
  │ Holo-4B       │ 438 ms │ 496 ms │ −12%                                   │
  ├───────────────┼────────┼────────┼────────────────────────────────────────┤
  │ Holo-9B       │ 657 ms │ 798 ms │ −18%                                   │
  ├───────────────┼────────┼────────┼────────────────────────────────────────┤
  │ 35B-A3B (MoE) │ 569 ms │ 568 ms │ ~0% (GDN a small slice of MoE prefill) │
  └───────────────┴────────┴────────┴────────────────────────────────────────┘
  ```

PR #344, opened 2026-07-20
fix(mtp-gate): throughput-arbitrated MTP gate + reuse-gated mid-chunk capture — restore 35B agentic wall-time

tbraun96:
## Summary

Fixes the 35B-flagship agentic wall-time regression introduced by #337 (webserver_ok gate: ~1000-1300s historical → 1846s + first-ever correctness failure), root-caused by same-box single-delta A/Bs to two independent mechanisms, each fixed on its own evidence:

1. **Mid-chunk GDN tail capture fired on every prefill** — including sessionless/single-turn traffic whose captures can never be reused (the snapshot lookup is session-gated; `session_hash` is a hash of the first ≤1024 prompt tokens, unique per single-turn request). Each capture costs a kernel split + D2D copy per prefill, and the split's accumulation-order change measurably perturbs trajectories (+18% tokens/turn). Now **reuse-gated**: capture fires only for sessions with live snapshot history — i.e. from a session's second request onward, exactly when reuse begins. `free()` clears session tags so freed slots can't report phantom history.

2. **The MTP gate compared component step timings and got the wrong answer** — on the 35B MoE, `verify_wall/decode_wall` = 2.07-2.23 ≥ `max_effective` 2.0 at K=1 disabled MTP essentially unconditionally (even at `mean_accepted=1.00`), while an always-on control measured **18% faster delivered decode**. Component walls miss per-token costs outside the timed region and the amortization of fixed per-step overhead across accepted tokens. The gate also thrashed (ENABLED→DISABLED within 6s on measurement noise, a draft-head resync per flip). Replaced with **throughput arbitration**: delivered tokens/wall per mode over 16-step windows, hysteresis margin + 2-window dwell, Serial re-probes MTP every 256 tokens, Mtp refreshes the serial baseline every 1024 tokens (≤0.3% overhead bound), probe windows replace stale baselines rather than blending. Knobs: `ATLAS_MTP_GATE_REPROBE`, `ATLAS_MTP_GATE_REFRESH`; `ATLAS_MTP_GATE_FORCE=1` still bypasses entirely. (Design informed by the training-free dynamic-speculation line: TapOut arXiv:2511.02017, AdaSD arXiv:2512.11280, SVIP arXiv:2411.18462; MoE verify-cost model per arXiv:2605.00342 — MoE verify scales with the expert union, which is why K=1 multipliers sit near 2 on the 35B and why measuring delivered throughput beats modeling it.)

[...]

PR #345, opened 2026-07-20
fix(ssm): re-land fossil-pinning fix + default-on is_tail eviction lease + staged Marconi score

tbraun96:
## Summary

Three fixes to the Marconi SSM snapshot cache's eviction policy, each with its own empirical gate. **Stacks on #344** (base = `fix/mtp-gate-throughput-arbiter`); retarget/rebase onto main once #344 merges.

### 1. Re-land the 07-10 fossil-pinning fix — the pathology is live at HEAD

`3d8130d0` (recency-only eviction + winner-only lookup bumps) was silently reverted when #317 re-cut HSS onto main: the hit-weighted escore `last_access*(1+hit_count)`, the `hit_count` field, and the bump-every-improving-candidate lookup scan all returned. The measured pathology this re-fixes: restore anchor frozen 29 turns, SSM replay growing to 18.6k tok/turn, 40s TTFT tail. The in-tree comment attributed hit-weighting to "the Marconi paper §4" — the paper (arXiv 2411.19379, MLSys'25) has **no hit-count term**: its Eq 2 is `S(n) = recency(n) + α·flop_efficiency(n)`. Regression tests re-added, retargeted at `lookup_tiered` (the serving path — the originals exercised the now-dead `lookup()`).

### 2. Replace `ATLAS_SSM_TAIL_PROTECT` with a default-on `is_tail` lease

The old flag protected the live session's **deepest** snapshot — the end-of-turn finish leaf, which sits *above* `matched_tokens` and can never be restored from. Measured on an eviction rig (8 slots, deep 8-turn session + 6 unique-session churn requests/turn): **zero Marconi hits with the flag on or off**, identical TTFT growth — it pinned a useless entry while the true restore point died. And on hybrid models the cost is maximal: losing the SSM snapshot forfeits the *entire* KV prefix hit (`Prefix cache hit: 5536 tokens (346 blocks) but no SSM snapshot — recomputing all KV`).

The lease protects the live session's `is_tail` entry (the actual restore point), **default on**, `ATLAS_SSM_TAIL_PROTECT=0` as kill switch (`=1` scripts stay compatible). Hardenings from adversarial review: lease TTL (`ATLAS_SSM_TAIL_LEASE_TTL`, 64 evictions) so a dead session cannot squat a slot; `insert()` overwrite clears `is_tail` (cross-session tail-rehoming breach); `lookup_tiered` gains the `is_tail` session gate the reference `lookup()` has had since the session-gate fix (the serving path previously allowed a sessionless lookup to restore another session's tail); deadlock guard unchanged; the reclaim path inherits the lease.

### 3. Marconi Eq-2 value-aware score, staged inert

Within-session rank `S(e) = norm(recency) + α·norm(token_count)` (uniform slots ⇒ FLOPs-saved/byte ≡ replay depth). `ATLAS_SNAP_EVICT_ALPHA` default **0.0** = provably today's pure-LRU ordering (unit-tested rank identity); clamped [0,8]; min-max normalized per pass (no per-entry accumulation, so the fossil vector cannot re-enter); session staleness stays the primary key. Default flip deferred to its own measured A/B.

## Validation

- 46 snapshot unit tests (×3 runs, no flake incl. 11 new: lease semantics inversion vs #278, lease expiry, both breach vectors, α rank-identity), 140 crate tests, fmt/clippy clean, files under the 500-LoC cap (tier + stats split out)
- **Gate 1 (eviction rig, same churn scenario that measured the old flag at zero)**: see comment below — lease restores 4/7 warm turns at ~1.0s flat vs 0/7 and 2.6s growth
- **8-leg webserver_ok matrix** (FP8/NVFP4 × bf16/fp8 KV × dgx1/2/3): see comment — no regression, ws_ok 80/80
- ST-995 BFCL 87.04/88.27 on the underlying binary line (posted on #344)

Follow-up (next commit on this branch): extend the lease to the mid-chunk **early sibling** (`tb−bs`) — Gate 1 showed the 2/7 residual misses occur exactly when the block-floored match lands one block below `tb`, where only the (currently unleased) early capture can serve.

### DIFF
--- a/crates/spark-model/src/model/ssm_snapshot.rs
+++ b/crates/spark-model/src/model/ssm_snapshot.rs
@@ -358,10 +358,30 @@ impl SsmSnapshotPool {
         Ok(())
     }
 
-    /// Return a snapshot slot to the free list.
+    /// Return a snapshot slot to the free list. Clears the slot's session
+    /// tag: a freed slot carries no restorable state, so leaving the tag
+    /// would make [`Self::session_has_history`] report phantom history.
     pub(super) fn free(&self, snap_slot: usize) {
         self.slot_has_hidden.lock().remove(&snap_slot);
+        self.session_tags.lock().remove(&snap_slot);
         self.free_slots.lock().push(snap_slot);
+    }
+
+    /// Whether any LIVE snapshot slot is tagged with `session_hash` — i.e.
+    /// this session has produced at least one snapshot before (a prior turn
+    /// finished, or a prior tail was captured). Used by the mid-chunk tail
+    /// capture to skip sessions on first sight: `session_hash` is a hash of
+    /// the first ≤1024 prompt tokens, so single-turn traffic gets a unique
+    /// hash per request and can never reuse a captured tail, while
+    /// multi-turn agents (stable long system prompt) match from their
+    /// second request onward — exactly when tail reuse begins.
+    pub(crate) fn session_has_history(&self, session_hash: u64) -> bool {
+        session_hash != 0
+            && self
+                .session_tags
+                .lock()
+                .values()
+                .any(|&s| s == session_hash)
     }
 
     /// Reserve a Marconi snapshot slot for an in-pass MID-CHUNK tail capture.
--- a/crates/spark-model/src/model/trait_impl/prefill_b.rs
+++ b/crates/spark-model/src/model/trait_impl/prefill_b.rs
@@ -268,8 +268,14 @@ impl TransformerModel {
         // ── Mid-chunk tail SSM capture (opt-in): plan BEFORE the forward
         // pass so SSM layers split their h/conv kernels at `tb` in-pass.
         // `None` (flag off or pass doesn't span `tb`) => no split. ──
-        let midcap_plan =
-            self.prepare_midchunk_capture(tokens, seq, &mut kv_cache, proc_start, proc_count);
+        let midcap_plan = self.prepare_midchunk_capture(
+            tokens,
+            seq,
+            &mut kv_cache,
+            proc_start,
+            proc_count,
+            stream,
+        );
 
         // ── Phase 4: forward through all layers ──
         self.prefill_b_forward_layers(
--- a/crates/spark-model/src/model/trait_impl/prefill_b/midchunk_capture.rs
+++ b/crates/spark-model/src/model/trait_impl/prefill_b/midchunk_capture.rs
@@ -126,6 +126,20 @@ impl TransformerModel {
         if !spark_runtime::ssm_tail_midchunk_enabled() || !self.ssm_snapshots.is_enabled() {
             return None;
         }
+        // Reuse gate: capture costs a per-prefill kernel split + D2D copy and
+        // is only ever consumable by a LATER request of the SAME session (the
+        // snapshot lookup is session-gated). A session seen for the FIRST
+        // time — which includes ALL single-turn traffic, whose ≤1024-token
+        // prefix hash is unique per request — can never reuse this capture,
+        // so skip it entirely; sessions gain capture from their second
+        // request onward (the first request's leaf save tags the session),
+        // exactly when reuse begins. Measured (35B FP8 webserver_ok N=10,
+        // 2026-07-20): capture-always Σ1846s/9-of-10 + 257 tok/turn vs
+        // capture-off Σ1495s/10-of-10 + 203 tok/turn on sessionless agentic
+        // traffic; multi-turn warm-TTFT (-54% max) is preserved from turn 3.
+        if !self.ssm_snapshots.session_has_history(seq.session_hash) {
+            return None;
+        }
         let bs = kv_cache.block_size();
         let tb = spark_runtime::ssm_tail_boundary(tokens.len(), bs)?;
         // Only capture when this pass strictly crosses tb.
--- a/crates/spark-model/src/model/trait_impl/prefill_b/midchunk_capture.rs
+++ b/crates/spark-model/src/model/trait_impl/prefill_b/midchunk_capture.rs
@@ -122,6 +122,7 @@ impl TransformerModel {
         kv_cache: &mut PagedKvCache,
         proc_start: usize,
         proc_count: usize,
+        stream: u64,
     ) -> Option<MidCapturePlan> {
         if !spark_runtime::ssm_tail_midchunk_enabled() || !self.ssm_snapshots.is_enabled() {
             return None;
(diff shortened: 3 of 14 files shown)


tbraun96:
## Gate 1 — eviction-pressure rig (the scenario that measured the old flag at zero)

8-slot pool, one deep 8-turn session (~2.4k→6k tokens), 6 unique-session churn requests between turns, 35B NVFP4, `atlas-gb10:evict-fix`:

| variant | Marconi hits | warm-turn walls |
|---|---|---|
| kill switch (`ATLAS_SSM_TAIL_PROTECT=0`) | 0 | monotonic growth 1.7→2.6s (full re-prefill every turn) |
| **lease (default)** | **4** | restored turns flat at **~1.0s** (−60%) |
| lease + `ATLAS_SNAP_EVICT_ALPHA=1.0` | 4 | identical to lease |

Old `ATLAS_SSM_TAIL_PROTECT` on the pre-fix binary: **0 hits whether on or off** — the flag was empirically worthless in its own target scenario. The 2/7 residual misses have a legible cause visible in the logs: `matched=5536` vs tail at `tb=5552` — when the block-floored match lands one block below `tb`, only the mid-chunk *early* capture (`tb−bs` = exactly 5536) can serve, and that entry is a plain insert the lease doesn't yet cover. Next commit extends the lease to it.

## 8-leg webserver_ok matrix — this branch vs its base (#344 binary), N=10 per leg

| model | box | KV | base Σ / med | **this PR Σ / med** | ws_ok | followed |
|---|---|---|---|---|---|---|
| FP8 | dgx1 | bf16 | 1195 / 107s | 1702 / 121s | 10/10 | 10/10 |
| FP8 | dgx1 | fp8 | 1590 / 125s | 1747 / 184s | 10/10 | 10/10 |
| FP8 | dgx3 | bf16 | 1027 / 100s | **937 / 93s** | 10/10 | 9/10¹ |
| FP8 | dgx3 | fp8 | 986 / 93s | 1198 / 99s | 10/10 | 10/10 |
| NVFP4 | dgx2 | bf16 | 697 / 67s | 719 / **64s** | 10/10 | 10/10 |
| NVFP4 | dgx2 | fp8 | 768 / 63s | **709 / 60s** | 10/10 | 10/10 |
| NVFP4 | dgx3 | bf16 | 917 / 63s | **719 / 68s** | 10/10 | 10/10 |
| NVFP4 | dgx3 | fp8 | 704 / 61s | **672 / 61s** | 10/10 | 8/10¹ |

Medians within ±6% on every leg (four legs improved); **ws_ok 80/80** (base binary: 78/80). dgx1 Σ values carry that box's documented long-trajectory excursion noise (3-5 slow runs/leg; medians are the comparable stat). ¹Recurring agent-behavior signatures (stops after tests pass / one-turn bail), present at the same rate on the base binary — not engine failures.

The lease is wall-neutral at production pool sizes (256 slots rarely evict), by design: its value is the pressure scenario in Gate 1, where losing the tail costs a full re-prefill because the KV prefix hit is forfeited without matching SSM state.

tbraun96:
**Commit 4 pushed** — `feat(ssm): lease the mid-chunk EARLY sibling (tb − bs) alongside the tail` (5 new unit tests; 51 green; deadlock guard reworked to bind only while an unleased candidate exists; siblings stay exact-prefix keyed with no `is_tail` lookup gate; `insert_tail`'s sweep clears tail+sibling together).

**Honest rig result: unchanged (4/7)** — and the diagnosis is a finding in itself. The residual misses were *not* sibling evictions: the early capture **never fires** in this scenario, on any binary (`midchunk EARLY` count = 0 across all rig runs). Geometry: on the miss turns the warm pass starts exactly at `proc_start = tb − bs`, giving `cap_local = tb − proc_start = bs`, and `prepare_midchunk_capture` requires `cap_local > bs` — so the capture is skipped precisely in the case it exists to serve (the next turn's block-floored match landing at `tb − bs`). Kernel-split capture at offset 0 is impossible by construction; the fix is a **pre-pass D2D copy of the live SSM state** into the sibling slot when `cap_local == bs` — filed as follow-up rather than rushed here.

The sibling lease remains correct and active wherever the early capture does fire (the original mid-chunk validation observed it on ~5/19 warm turns in deeper-chunk geometries); this PR's other gates (rig 0→4/7 restore, 8-leg matrix, 51 unit tests) are unaffected — commit 4 is index-plumbing relative to the matrixed binary.

tbraun96:
**Root cause fixed — commit 5** (`fix(ssm-midchunk): pre-pass copy for the EARLY sibling when the pass starts at tb−bs`).

Skip-reason instrumentation proved the geometry claim exactly: **7/7 tb-spanning warm passes started at `proc_start = tb − bs`** (`cap_local=16 bs=16` every time), so the in-pass split path's `cap_local > bs` requirement excluded precisely the case the sibling exists to serve. Since a split at offset 0 is impossible but the live pool state at pass start *is* the `tb−bs` state, the fix is an ordinary `save()` D2D before the forward pass (stream is synchronized by the caller immediately before planning), registered as the leased sibling with no kernel split (`cap_local_early` stays `None`).

**Rig result (same 8-slot churn scenario):**

| binary | Marconi hits | warm-turn walls |
|---|---|---|
| old TAIL_PROTECT (either setting) | 0 | 866 → 2092ms, full re-prefill growth |
| lease (commits 1–3) | 4 | 4 turns flat ~1.0s, 2 turns 2.3–2.7s |
| **lease + sibling + pre-pass copy (this)** | **6** | **all 7 warm turns flat 0.9–1.0s** |

Pre-pass copies fired 7/7. The single residual "miss" is turn 2 — the reuse-gate's first-sight skip (sessions begin capturing from their second request), by design. The eviction-pressure pathology is fully resolved: deep-turn TTFT no longer grows with context under churn.

tbraun96:
**α-score A/B (default-flip decision): keep `ATLAS_SNAP_EVICT_ALPHA=0.0`.**

Rig: 3 interleaved deep sessions (distinct >1024-token system prompts → distinct session hashes), 6 round-robin turns each, `evict-fix3` binary, lease on, two contention levels:

| pool | metric | α=0 (default) | α=1.0 |
|---|---|---|---|
| 8 slots (saturated) | hits / tokens skipped / full-recompute misses | 0 / 0 / 15 | 0 / 0 / 15 |
| 16 slots (moderate) | hits / tokens skipped / full-recompute misses | 1 / 2352 / 9 | 1 / 2352 / 9 |

Byte-identical decisions and walls in both regimes. Structural explanation: with session staleness as the primary key, α only reorders evictions *within* the victim session — and in monotonically-deepening conversations the oldest entry is always the shallowest, so `norm(recency) + α·norm(depth)` ranks identically to pure LRU at any α. Divergence requires depth/recency disagreement inside one session (branching or regenerate-from-earlier workloads, same-recency mixed-depth grid checkpoints), which this traffic — and our production traffic — does not produce. Under saturation the pool turns over completely between a session's visits, so no policy can help (lease included: `last_lookup_session` rotates away).

The knob stays staged (inert, clamped, unit-tested) for future branching workloads; the measured default remains 0.0. This also doubles as an inertness confirmation for the staged commit: α=1 changed nothing anywhere the score never got a real choice.

tbraun96:
**α-score default flip: A/B run — verdict is KEEP `ATLAS_SNAP_EVICT_ALPHA=0.0` (no flip).**

Design: α can only bind *within* a session's non-leased entries — session staleness is the primary key, and the lease already covers tail+sibling. So the probe is the scenario that maximizes that pressure: one deep session, **no churn**, 8-slot pool, 20 turns growing 2.2k→10.2k prompt tokens, temp 0 / seed 42 / identical prompts across conditions. 35B NVFP4 on the commit-5 binary; α is a runtime env var so all four legs share one image.

| α | Marconi hits | SSM misses | restored anchor depths (per turn) |
|---|---|---|---|
| **0.0** (default) | 18 | 1 | 2624, 3056, 3472, … 9376, 9792 |
| 0.5 | 18 | 1 | *byte-identical* |
| 1.0 | 18 | 1 | *byte-identical* |
| 2.0 | 18 | 1 | *byte-identical* |

Not "within noise" — every eviction decision was identical. ~51 snapshot saves into 8 slots per leg means ≥40 evictions each, so α had ample opportunity to bind and never did.

**Why (structural, not incidental):** in a growing conversation each new snapshot is simultaneously the *newest* and the *deepest*, so recency-rank and depth-rank are perfectly correlated and `S(e) = norm(recency) + α·norm(depth)` preserves the same ordering for every α ≥ 0. Across sessions, staleness decides before α is ever consulted. This matches the Sparse Prefix Caching result cited in the design work (arXiv:2605.05219): for chat-like workloads where each request contains the previous as a substring, keeping the last state is already optimal — there is nothing for a value term to improve.

The mechanism itself is correct and does bind where depth and recency are *inverted* — unit test `alpha_prefers_depth_within_session_staleness_still_primary` covers exactly that (deep-old vs shallow-fresh → α picks the shallow entry as victim). Real inversions require within-session context truncation/sliding-window or conversation branching/regeneration, which this agentic workload does not produce.

**Recommendation:** leave the default at 0.0 (provably identical to pure LRU) and keep the mechanism staged with its clamp, NaN guard and rank-identity test, as an opt-in for branching/truncating workloads. Flipping a default that demonstrably changes no decision would add review surface and risk for zero measured gain.

*Probe caveat, for the record:* the model emitted a repeated greeting rather than substantive answers on these synthetic prompts, so this leg says nothing about answer quality. Prompt tokens grew correctly (2214→10227) and the prefix/eviction machinery was fully exercised, which is the property under test — eviction ordering depends on prefix structure, not response content.

PR #346, opened 2026-07-20
feat(mtp): follow-ups — confidence-gated drafting, drafter catch-up feed, MoE verify telemetry

tbraun96:
## Summary

The follow-ups PR for the #344/#345 line — **draft until all three items land on this branch**.

### ✅ 1. Per-step confidence-gated drafting (`ATLAS_MTP_DRAFT_CONF`, staged off) — this commit

AdaSD/SVIP-style per-step draft gating (arXiv 2512.11280 / 2411.18462) adapted to Atlas's pipelined MTP: `forward_one` computes the drafter's top-1 softmax prob per draft (BF16 logits D2H ~200µs — the cost the grammar-masked path already pays; token selection untouched), folds it into a propose-scoped chain-confidence MIN, and `run_mtp_propose_inner` discards drafts below τ — **trimming the drafter KV exactly as a full rejection would** (`after_verify(0)`) so the drafter never desyncs. The next step decodes serially via the normal bootstrap path.

Economics: at K=1 on the 35B MoE, verify ≈35ms for 1+acc tokens vs decode+propose ≈21ms — a draft is worth verifying only at p(accept) ≳ 0.66, which is τ's calibration target.

**Staged OFF** (τ=0.0 default; every new code path is behind τ>0, so the default is byte-identical) — same discipline as `ATLAS_SNAP_EVICT_ALPHA`. Default flip requires the acceptance-vs-confidence calibration A/B.

Validation: clippy/fmt/tests clean; live smoke on `nvidia/Qwen3.6-27B-NVFP4` (K=3, golden config): **8/8 tool calls at τ=0 and 8/8 at the aggressive τ=0.5** — correctness is verify-guaranteed by construction (wrong drafts are rejected; the clamp can only change performance).

Incidental finding while validating: 35B-NVFP4 + `--tool-call-parser qwen3_xml` scores 0/8 on the structured tool-call probe **with or without** this feature — a pre-existing model/parser format mismatch (the 35B's agentic gates drive tools through opencode prompts, never the parser). Filed here for visibility; not addressed in this PR.

### ⏳ 2. Drafter catch-up feed on Serial→Mtp transitions (next on this branch)

Hidden-ring capture during serial decode + an append-capable drafter prefill so gate transitions never run a stale drafter. Safe by construction (verify rejects bad drafts) but touches decode buffers and drafter position bookkeeping — lands with its own acceptance A/B.

### ⏳ 3. MoE expert-union verify telemetry (next on this branch)

Sampled D2H of the verify batch's `expert_indices` union — the measurement stage for expert-union-aware verify batching (the m≈2.07 MoE verify multiplier, cf. arXiv 2605.00342).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,7 @@ jobs:
           int cuStreamBeginCapture(void *a, int b) { (void)a; (void)b; return 100; }
           int cuStreamBeginCapture_v2(void *a, int b) { (void)a; (void)b; return 100; }
           int cuStreamEndCapture(void *a, void **b) { (void)a; (void)b; return 100; }
+          int cuStreamIsCapturing(void *a, unsigned int *b) { (void)a; (void)b; return 100; }
           int cuStreamWaitEvent(void *a, void *b, unsigned int c) { (void)a; (void)b; (void)c; return 100; }
           /* Modules + launch */
           int cuModuleLoadData(void **a, const void *b) { (void)a; (void)b; return 100; }
--- a/crates/spark-model/src/layers/moe/forward_k2.rs
+++ b/crates/spark-model/src/layers/moe/forward_k2.rs
@@ -139,6 +139,13 @@ impl MoeLayer {
                 stream,
             )?;
         }
+        super::union_stats::maybe_sample_expert_union(
+            ctx.gpu,
+            indices_dev,
+            2,
+            top_k as usize,
+            stream,
+        );
 
         if k2_diag {
             ctx.gpu
--- a/crates/spark-model/src/layers/moe/forward_k3.rs
+++ b/crates/spark-model/src/layers/moe/forward_k3.rs
@@ -94,6 +94,14 @@ impl MoeLayer {
             )?;
         }
 
+        super::union_stats::maybe_sample_expert_union(
+            ctx.gpu,
+            indices_dev,
+            3,
+            top_k as usize,
+            stream,
+        );
+
         // 3-5. Fused expert dispatch for 3 tokens
         let expert_gate_out = ctx.buffers.expert_gate_out();
         let expert_up_out = ctx.buffers.expert_up_out();
--- a/crates/spark-model/src/layers/moe/mod.rs
+++ b/crates/spark-model/src/layers/moe/mod.rs
@@ -422,4 +422,5 @@ mod init;
 #[cfg(test)]
 mod mod_tests;
 mod ptr_table_build;
+mod union_stats;
 pub(crate) use ptr_table_build::*;
--- a/crates/spark-model/src/layers/moe/union_stats.rs
+++ b/crates/spark-model/src/layers/moe/union_stats.rs
@@ -1,0 +1,79 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! Sampled expert-union telemetry for MTP verify batches
+//! (`ATLAS_MOE_UNION_STATS=1`, default off = zero cost).
+//!
+//! MoE verify cost scales with the UNION of experts activated across the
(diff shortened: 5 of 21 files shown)


tbraun96:
## Gate evidence (items 1–3, one PR)

Binary under test: `atlas-gb10:followups` @ `25724fe8` (all three features **staged off by default**).

### Item 1 — confidence-gated drafting (`ATLAS_MTP_DRAFT_CONF`)
- 27B golden K=3: **8/8 tool_ok at τ=0 AND τ=0.5** (clamp is perf-only by construction: rejected drafts fall back to serial bootstrap; drafter KV trimmed via `after_verify(0)`).

### Item 2 — drafter catch-up feed (`ATLAS_MTP_CATCHUP`)
Measurement-driven redesign (see commits `a6763e10`, `25724fe8`): the drafter runs a **compacted row space** (rows ≠ sequence position; gap grows +1/verify cycle) with sequence-space RoPE. Catch-up therefore tracks the drafter's newest **pair key** and feeds missed keys from a 512-row hidden ring, clipped to ring coverage (RoPE gaps are the norm in compacted mode, so partial feeds are safe; verify rejects any bad draft).

A/B on 27B golden K=3 with forced gate churn (`ATLAS_MTP_GATE_REPROBE=64`, `ATLAS_MTP_GATE_REFRESH=256`), 3×600-token gens:

| leg | tok/s (3 gens) | catch-up feeds | tool sanity |
|---|---|---|---|
| churn, catchup off | 20.8 / 20.6 / 20.5 | 0 | — |
| churn, catchup **on** | 21.1 / 20.6 / 20.1 | **6** (one per serial re-probe window, e.g. `fed pair keys 284..=299`) | 8/8 |

Wall-neutral; the feed fires exactly at Serial→Mtp re-entries. Telemetry (debug level) also surfaced a durable finding: even steady-state MTP misses ~1 pair per verify cycle (bootstrap tokens never reach the drafter) — the mechanism behind acceptance degrading with context; future lever, out of scope here.

### Item 3 — MoE expert-union telemetry (`ATLAS_MOE_UNION_STATS`)
- **Capture-safety fix**: the tap's sync+D2H inside the graphed K2 verify capture invalidated the capture (CUDA 901) and wedged the serve. New `GpuBackend::stream_is_capturing` guard (`cuStreamIsCapturing`; SCALE cfg-gated). Graph **replays** run no host code, so the tap samples **eager** verify steps — measure with `ATLAS_K2_DIAG=1` (eager K2 verify).
- 35B NVFP4, fp8 KV, K=1, eager: `moe-union-stats: samples=512 mean_unique_experts=13.3 mean_routed_slots=16.0 overlap_saving=17% (m=2 top_k=8)` — stable across 8 log points; flag-off legs silent.
- First hard data for expert-union verify batching (arXiv 2605.00342): at K=1 the verify pair shares ~2.7 of 16 routed expert slots.

### Merge gates
| gate | config | result | bar | comparison |
|---|---|---|---|---|
| A: webserver_ok N=10 | 35B FP8, defaults (features off), dgx3 | **Σ1007s, 10/10 ws_ok, 10/10 followed_directions, 0 drift** | Σ≤1300s, 10/10 | band 950–1287s; #344 fixval Σ1252s |
| B/D: BFCL ST-995 | dense 27B golden, defaults, dgx1 | **87.14 / 88.21 norm** (non_live 83.64 / live 87.06 / halluc 93.94) | ≥83.64 / 85.32 | gate-fix 87.04/88.27 (equal within draw noise); #337 86.63 |

Both merge gates ran the followups binary with **all three flags at default (off)** — they double as the default-path-inertness proof on scored benchmarks.

Run commands per leg in the thread above; feature-gate scripts: `/tmp/fu_gates3.sh`, `/tmp/fu_item3.sh` (dgx3).

PR #350, opened 2026-07-21
spark-model: add Laguna-S-2.1 inference support

rsafier:
## Summary

- add Laguna-S-2.1 model config, dispatch, weight loading, and GB10 NVFP4 kernel target
- support its heterogeneous full/sliding attention layout, partial YaRN RoPE, dense first layer, and sigmoid/correction-bias routed MoE layers
- add the Poolside v1 chat template, tool grammar/parser, and response cleanup needed for native tool use
- fix FP8 paged decode so sliding-window attention is honored in regular and split-K paths

## Why

Laguna-S-2.1 has architectural and serving requirements that were not represented by an existing Atlas model family. The integration keeps those details in the model configuration and kernel-target layers while sharing the existing runtime paths where possible.

The runtime fix is cross-cutting: FP8 paged decode previously omitted `sliding_window`, so an FP8 KV cache silently used the wrong context semantics on sliding layers.

## Runtime and operator impact

- FP8 KV cache is the intended/default path for this target.
- Validation keeps `--gpu-memory-utilization 0.80`; do not raise it unless required.
- Models loaded from NFS should use `--fast-load-prefetch-shards` to stage shard reads.
- DFlash is deliberately not included in this PR.
- Shared runtime/kernel code is touched for FP8 sliding-window decode and vanilla RMSNorm support; reviewers should treat those portions as cross-cutting.

## Validation

Passed locally:

- `cargo fmt --all -- --check`
- `ATLAS_SKIP_BUILD=1 cargo test -p atlas-core laguna`
- `ATLAS_SKIP_BUILD=1 cargo test -p spark-server poolside --no-default-features --features cuda`
- `typos`
- `git diff --check`
- release CUDA build for `gb10 / laguna-s-2.1 / nvfp4` (152 kernels, 2 model overrides)
- real 80.8 GB model load from NFS with FP8 KV, shard prefetch, and GPU memory utilization capped at 0.80
- `/health`, `/v1/models`, deterministic arithmetic, and one-shot native weather-tool calls

License note: the repository's Docker-based license check could not execute locally because its pulled image was x86-only on this ARM host (`exec format error`). Every changed/new source file was checked directly for the required line-1 SPDX header.

## Known gaps / draft status

- The stock basic agentic harness is not green yet. After fixing FP8 sliding-window decode, it progresses through valid file and tool operations, but greedy generation at temperature 0 can repeat reasoning and hoist an empty `write_file` call. A no-thinking diagnostic progresses further but can loop while correcting a deliberately failing assertion.
- `webserver_ok` and `followed_instructions` are still pending behind the basic agentic gate.

This is intentionally a draft while the remaining agentic behavior is diagnosed and the full validation sequence is completed.


rsafier:
Progress banked to MonumentalSystems/atlas feat/laguna-s-2.1:

- c1926fd1 — preserve checkpoint-BF16 Laguna shared experts across decode and prefill
- b3e26427 — align the Poolside parser, blocking split, grammar safety, and Laguna template behavior with the model-card contract

Validation banked so far:
- spark-model unit tests: 194 passed
- spark-server binary tests: 736 passed, 1 ignored
- targeted spark-model and spark-server clippy with warnings denied: passed
- Laguna-only GB10/NVFP4 release build: passed (152 kernels, 2 model-specific overrides)

Live server validation is still running with no DFlash, FP8 KV, gpu-memory-utilization 0.80, max sequence length 65536, and fast-load NFS shard prefetch. Startup reported a 13.9 GiB FP8 KV cache / 151392-token capacity and explicit NFS prefetch requests. I will add the basic-agentic, webserver_ok, and followed_instructions outcomes as they complete.

rsafier:
Basic agentic milestone (rebuilt binary) — PASS 3/3 under FP8 KV, gpu-memory-utilization 0.80, NFS shard prefetch, no DFlash:

- calculator: PASS, 15 turns, 10 tool calls, diagnosed and repaired one import failure
- sorter: PASS, 15 turns, 6 tool calls, no execution errors
- converter: PASS, 15 turns, 5 tool calls, no execution errors

The converter regression from the old binary is cleared: it emitted a valid package import on the first attempt and all six tests passed. Across all three cases the Poolside envelopes remained complete and schema-valid. The extra turns are the known agentic_test.py behavior where a success summary containing test/run causes the harness to request verification again; final verification passed in every case.

Next bank: cold ISL/OSL baseline, prefix-cache cold/warm proof, generic SSM rollback-cache regression coverage (Laguna reports 0 SSM layers), then webserver_ok + followed_instructions.

rsafier:
No-prefix cold baseline banked (3 samples/config, batch 1, FP8 KV, util 0.80, no DFlash):

| Target ISL/OSL | Actual prompt tokens | Client TTFT p50 | Approx server prefill tok/s | Decode tok/s p50 |
|---:|---:|---:|---:|---:|
| 256/256 | 290 | 1732.5 ms | 188 | 16.5 |
| 1024/128 | 980 | 4269.5 ms | 241 | 16.5 |
| 4096/128 | 3710 | 14717.3 ms | 256 | 15.8 |
| 8192/128 | 7350 | 29253.0 ms | 253 | 15.1 |

Prefill tok/s uses the server-reported prefill TTFT visible in the startup session logs (about 1.54 s, 4.07 s, 14.51 s, and 29.02 s respectively); client TTFT is slightly higher because it waits for the first content-bearing SSE delta after the template-seeded reasoning boundary. Raw sweep artifact: /tmp/laguna-no-prefix-isl.json on the validation host.

Next I am restarting the same build with radix prefix caching enabled and will bank cold/warm cached-token evidence.

rsafier:
Prefix-cache milestone — PASS on Laguna with the same FP8 KV / util 0.80 / NFS-prefetch setup.

Startup evidence:
- Prefix caching: ENABLED (radix tree)
- KV cache: 13.5 GiB, 147072-token capacity at 80% utilization
- SSM state pool: 1 slot x 0 layers = 0 MB (Laguna is attention-only)

Identical cold/warm completion probes:

| Prompt tokens | Cold TTFT | Warm TTFT | Warm cached tokens |
|---:|---:|---:|---:|
| 305 | 1537.7 ms | 71.8 ms | 305 |
| 1162 | 4643.2 ms | 73.7 ms | 1162 |
| 4581 | 17685.3 ms | 79.3 ms | 4581 |

Metrics after the probe: 3 hits, 4 misses (including warmup/cold requests), 6048 hit tokens, hit rate 0.4286. Every warm request reported its full prompt as cached. Raw artifact: /tmp/cold_warm_ttft_laguna-prefix.json on the validation host.

rsafier:
SSM rollback-cache audit banked:

- Laguna startup reports 48 attention layers, 0 SSM layers, and a 0 MB SSM state pool. Therefore an SSM snapshot/restore hardware claim is not applicable to this model; prefix hits correctly use the KV-only attention path.
- Scheduler rollback suite: 17/17 passed, including snapshot-aware boundary selection, missing-snapshot decline, eviction behavior, rewind semantics, and rollback cap.
- SSM decode-ring suite: 13/13 passed, including slot assignment/reuse, exact-position lookup, eviction, truncation, cursor rewind, and missing-snapshot decline.
- The previously banked spark-model suite (194/194) also covers SSM state allocation, pool ownership, spill/fault-in, tier isolation, and capability gating.

So the generic rollback-cache machinery is regression-green, while Laguna itself correctly allocates no SSM rollback state.

rsafier:
Banked prefill performance fix in `fac4d4f8` (`spark-model: accelerate Laguna BF16 prefill`).

Nsight Systems cold 8K baseline (loading excluded):
- 8,043 prompt tokens: TTFT 31.51s (~255 tok/s)
- 70.9% / 22.15s in 332 scalar `dense_gemm_bf16` launches
- 16.1% / 5.03s in routed `moe_w4a16_grouped_gemm_ptrtable`
- attention kernel itself 4.8% / 1.49s

Fix: route prefix-cache Q/K/V, MoE router, and checkpoint-native BF16 shared-expert long-prefill projections through the already-shipped pipelined tensor-core BF16 GEMM, retaining an explicit scalar fallback for old targets.

Same cold 8,043-token request after fix: TTFT 10.30s (~781 tok/s), 3.06x faster. `ATLAS_MOE_PREFILL_EXACT_TILES=1` additionally reached 9.91s (~812 tok/s), only +3.7%, so it is not yet selected for the production recipe.

Validation: spark-model 194/194 tests, targeted clippy `-Dwarnings`, Laguna-only real CUDA release build (152 kernels), fmt/diff checks. Next: CUDA 13.2 CUTLASS/grouped-MoE A/B, then correctness/agentic and C=1/2/4 concurrency gates.

rsafier:
Banked concurrency correctness fix in 442c84b2 (spark-model: fix native dense batched decode).\n\nRoot cause:\n- Laguna layer 0 uses checkpoint-native BF16 dense FFN weights and null NVFP4 placeholders.\n- C=2 entered the NVFP4-only DenseFfnLayer::forward_k2 path and dereferenced those null placeholders.\n- CUDA reported the sticky async fault at the following sliding-attention softplus launch (grid=72, block=256), not at the actual bad launch.\n- BF16/FP8 K=2/K=3 now reuse the existing format-aware forward_prefill dispatcher; NVFP4 retains its specialized batch GEMV path.\n\nValidation:\n- CUDA 13.2/CUTLASS release build: passed, 152 Laguna kernels.\n- Real Laguna C=2: 2x128-token multi-step decode completed, no CUDA-700.\n- Real Laguna C=4: completed, no CUDA-700.\n- spark-model: 195/195 tests passed.\n- Package clippy with -Dwarnings, fmt, diff check, and typos passed.\n- Workspace all-features clippy is host-blocked by the existing Apple-only objc2 feature; license-eye is also the wrong host architecture.\n\nCurrent fixed-output decode baseline (128 requested tokens/stream, short ISL):\n- C1: 15.3 aggregate tok/s\n- C2: 15.7 aggregate, 7.8/stream\n- C4: 19.1 aggregate, 4.8/stream\n\nCorrectness is restored, but batching is still materially below the supplied vLLM 35/51 aggregate targets. ATLAS_DECODE_GRAPHS_MULTISEQ=1 was exercised and did not close that gap, so it is not yet promoted into the final recipe.

rsafier:
Banked a49a64fc: corrected Laguna MODEL.toml parameter metadata from the placeholder 100B to the model-card value 117.6B-A8.5B. Verified with ATLAS_SKIP_BUILD=1 cargo test -p atlas-kernels, cargo fmt --check, and git diff --check.

rsafier:
Exact-token concurrency update (balanced profile, 8192 chat tokens per stream, 32 cached framing tokens): C=1 798.7 aggregate uncached prefill tok/s, 10.216s TTFT; C=2 790.7 aggregate, 20.639s wall; C=4 789.5 aggregate, 41.341s wall. Prefill remains serialized with prefix caching enabled, so aggregate throughput is flat across C=1/2/4. The safe batched-first-chunk/co-dispatch path is intentionally ineligible while a live prefix cache is active; those flags are therefore excluded from the recipe. Draft recipe: Avarok-Cybersecurity/atlas-recipes#12.

rsafier:
Banked `2088fd74` (`spark-server: repair punctuation-drifted agent cwd`).

Validation:
- red/green regression for punctuation drift inside Bash command cwd
- all 3 path sanitizer tests pass
- `ATLAS_SKIP_BUILD=1 cargo check -p spark-server --tests` passes
- CUDA 13.2 Laguna NVFP4 build passes (152/152 kernels)
- corrected OpenCode harness uses `model=poolside/Laguna-S-2.1-NVFP4`; cargo-valid + webserver scorer pass, no path drift

The harness still reports `followed_dirs=0/1`: Laguna can stop after saying it will run build/curl rather than issuing the Bash call. I am keeping that model-behavior tuning separate from this path repair.

rsafier:
Final bank for this handoff:

- `96c1cdf3` honors Laguna agent defaults: thinking-on, model-card 0.7/0.95/top-k-20 defaults when omitted, and no duplicate Atlas cwd injection. Explicit request sampling/thinking still wins.
- `b70d76c5` adds `handoff.md` with full validation state, exact container/recipe settings, agentic traces, HF-template limitation, kernel-audit/MoE analysis, baselines, and prioritized next work.

Validation:
- fmt + diff check
- `ATLAS_SKIP_BUILD=1 cargo check -p spark-server --tests`
- targeted Poolside prompt, cwd injection, and path sanitizer tests
- source-aligned CUDA 13.2 release build: 152/152 Laguna NVFP4 kernels
- active container healthy with prefix cache, FP8 KV, 16K prefill, 0.80 utilization, 200K overcommit, NFS prefetch, 4 GB swap, SLAI 100 ms, and 1024 thinking cap

Cross-cutting review note: this adds two generated `ModelBehavior` fields (`disable_cwd_hint_injection`, `use_sampling_presets_for_core`) with historical behavior preserved for all other targets.

Agentic status: cargo-valid/webserver scorer passes with no path drift; direction-following remains incomplete because Laguna can stop after planning the final build/curl. The 8K natural-budget + disabled-confidence diagnostic was worse and was reverted. See `handoff.md` before continuing.

rsafier:
Progress note (2026-07-22):

- Added `e796bc88` to make `ATLAS_PREFILL_CODISPATCH=1` consistently enable chunk-zero paged batched prefill; the scheduler and both attention gates previously disagreed on the flag.
- Checks: `cargo fmt --all`; `ATLAS_SKIP_BUILD=1 cargo test -p spark-model accepts_ --lib` (5 passed).
- Corrected cold, no-thinking C=1 FP8-KV baseline: 1K 1.74s / 17.1 tok/s; 8K 9.69s / 16.3; 32K 55.0s / 14.2. BF16 KV was near-neutral and ~6% better at 32K; NVFP4 KV regressed long prefill and decode.
- Nsight 8K C=1: routed NVFP4 MoE is 4.24s of 8.93s GPU prefill (47.4%); the GPU is saturated.
- C=2/C=4 8K requests currently prefill serially with prefix caching enabled. The no-cache co-dispatch forward also remained effectively serial (~18.8s C=2), so the next performance item is the model batched forward, not SLAI admission. Prefix-cache-aware kernel batching remains guarded due the documented mixed-cache-depth mutation hazard.

rsafier:
Implemented and pushed `6f98fc5d` (`spark-model: admit safe cached batched prefill`).

- Removes the blanket active-prefix-cache exclusion for Q12 co-dispatch.
- Adds a transactional cache admission: reserve every match before sequence/KV mutation; admit only equal partial depths for single-rank, non-SSM, non-HSS requests; roll back rejected reservations at their exact acquired depth.
- An admitted batch consumes its reservation rather than re-looking up the cache. Post-admission errors now propagate instead of retrying sequentially against dirty sequence/KV state.
- Validation: 201 `spark-model` + 146 `spark-runtime` unit tests; CUDA 13.2 release build succeeded. Full all-features clippy remains blocked by the existing Linux `objc2` platform error.

Runtime smoke, cache ON / FP8 KV / `ATLAS_PREFILL_CODISPATCH=1`:
- C=2, 7,520-token prompts: Q12 kernel-batched succeeded; TTFT 19.22s/request; decode ~8.0 server tok/s/stream.
- C=4, 3,861-token prompts: every request reused the shared 48-token prefix; Q12 kernel-batched succeeded; TTFT 18.16s/request; decode ~5.39 server tok/s/stream.

This proves cache-on model-side batching and safe admission; it does not yet improve MoE prefill throughput. Nsight already points to `moe_w4a16_grouped_gemm_ptrtable` as the dominant next optimization target.

rsafier:
## Performance & agentic tuning (follow-on to the base integration)

A set of measured, in-kernel improvements + serving-behavior fixes landed on top of the base Laguna-S-2.1 support. All validated on a single GB10 (sm_121), NVFP4 routed MoE + BF16 attention/lm-head, FP8 KV.

### Prefill — C=1 745 → 562 ms (63% → 83% of vLLM's 467 ms)
Root-caused with a clean full-prefill nsys (98.5% GPU-busy, i.e. kernel-bound, no scheduling gap):
- **Dense-FFN projections → cuBLASLt** (`1caee655`): Laguna's single dense FFN layer (`mlp_only_layers=[0]`, intermediate 12288) ran its gate/up/down on `dense_gemm_tc` at ~1.4 TFLOP/s = **33% of the whole C=1 prefill**. Routing to the cuBLASLt path q/k/v/o already use = 90+ TFLOP/s (~65×). This is the −25% lever, and it generalizes to any dense-FFN model.
- **Head-gate projection → cuBLASLt** (`8628bb3b`): the tiny `[n,72]` per-head gate is pathological on `dense_gemm_tc` (2 N-blocks); ~2.5%.

The remaining ~95 ms to vLLM is at the kernel rooflines — routed MoE is bandwidth-bound (~290 GB/s ≈ GB10 peak), attention projections at ~95 TFLOP/s (cuBLAS peak) — i.e. distributed launch/support overhead, not a single kernel.

### Decode — concurrent throughput
- **Grouped read-once MoE for concurrent decode** (`8fdc2ff3`, `b6646b06`): all n≥2 batched decode routes through the grouped path (reads each expert once). Batched BF16 q/k/v/o + head-gate GEMV (`250931e9`, `84d5b763`, `5128ec00`, `6da3277d`) and BF16 lm-head batched GEMV (`cd101915`).
- Measured matrix (C=1–8 × ISL 1–4K): ~19 tok/s per-stream C=1, aggregate batches to ~89 tok/s @C8/1K. Per-stream decode is bandwidth-bound; Laguna has no MTP so the lever is batching (engaged).

### Thinking / chat template robustness
- **Thinking-budget opt-out** (`3a1c1c78`): new `[behavior].cap_thinking_at_max_tokens` (default true, false for Laguna). Removes a second, `max_tokens`-derived thinking cap that vLLM doesn't have and that truncated reasoning on small-`max_tokens` agentic turns — `max_thinking_budget` is now the sole cap (vLLM single-budget).
- **min_p floor 0.1** (`80d7ce4a`) + **watchdog family disabled** (`51bb4170`, `353e690c`): root-fix for the temp-0 repeat-loop degeneration (the watchdogs were misfiring on the greedy plateau).
- Chat fixes: `</think>` no longer leaks into content (`19a2704e`), reasoning-parser fallback when the close token is absent (`f3bb1167`), orphan `<tool_call>` markup stripped (`3ba80752`), top-level `enable_thinking=false` honored (`3feb225c`).

### Agentic sampling
- **Sampling presets temp=1.0 / top-k=50** (`73d085c4`): field-reported + locally validated to behave better in agentic clients (opencode) than 0.7/top-k=20. Statistical N=10 opencode webserver probe: **webserver_ok 10/10**, clean loop-free sessions (at 0.7 the model fell into deterministic tool-call loops). `temperature_max` raised 0.7→1.0 so the preset isn't clamped; `min_p_floor=0.1` kept (FP-quant tail guard). Clients that omit temperature (opencode) resolve to this preset via `use_sampling_presets_for_core`.

### Notes
- b12x fused-MoE was evaluated separately (PR #62) and parked: the routed MoE is already native-FP4 (W4A4) at the bandwidth roofline, so b12x's fusion offers only ~43 ms and carries residual numerics/memory limits.
- Open follow-up: `followed_directions` on the agentic harness is ~5/10 (model completes the code but sometimes stops before the run-server/curl/teardown verification); a `max_inter_tool_prose` (384→1024) A/B is the next lever (the 384 budget guillotines legitimate planning prose per the in-code comment).

tbraun96:
Status, because three weeks of silence on a PR this size is not acceptable and I'd rather say something incomplete than nothing.

**This is not being closed.** I had it on a reject list; an adversarial re-review overturned that, and the reject list is now down to the PRs whose content main genuinely already has (#266, #332) or has already landed by another path (#373, #375 — both closed as *merged elsewhere*, with credit and with the defects their bodies documented filed as #421 and #422).

What I can state now:

- **It builds.** I verified the CUDA compiles under the real production flags (`nvcc --ptx -arch=sm_121f -O3 --fmad=false -DTQ_PLUS_SIGNS --Werror all-warnings`, standalone per file). That matters more than it sounds: **CI has never compiled a `.cu` file** — `ci.yml` sets `ATLAS_SKIP_BUILD: "1"` repo-wide, so every kernel in this PR reached review having been built only on your box. #419 fixes that; the job runs the real `build.rs` for all 22 gb10 targets in ~30 s with no GPU.
- **This branch conflicts with main, and a conflicting PR makes our CI silently skip.** The green checks here were never real. That's a bug in our process, not yours, but it does mean a rebase is the first useful step.

What I still owe you: a file-level review, and a statement of what a merge would need under the rigor we're now applying — specifically a test that fails without the change, not a smoke test.

Two process problems on our side that I've fixed or filed rather than just noting: the benchmark gate was rejecting records via `git merge-base --is-ancestor` in a repo that squash-merges, so every PR read as ungated regardless of content (#420, merged); and the kernel-compile gap above.

If you want to prioritise, tell me which of your open PRs you most want landed and I'll take that one to completion first.

rsafier:
Branch synced with main (merge, no force-push). 23 conflict files resolved toward main's post-#388 structure with the Laguna semantics re-homed — highlights: the `disable_cwd_hint_injection` behavior became a proper `ChatLevers` field (main's carrier) with the test rewritten to that signature; the batched-prefill KV pre-flight now declines via this PR's `KernelBatchResult::NotAdmitted` instead of `bail!` and then takes the radix reservation; `stage_batched.rs` keeps main's bounds-checked pinned packer carrying this PR's `kv_lens` field (the naive merge silently dropped it — restored); GEMM env flags dropped in favor of main's `GemmDispatch` levers (8 call sites rewired).

Post-sync payload: **227 files, +10,369/−10,661**. fmt / clippy `-D warnings` / all test suites green (spark-model 324, spark-server 1792, 0 failures; one `tui::render` flake reproduced on byte-identical-to-main code — pre-existing upstream). The 500-LoC cap is clean: 8 files that crossed it in the merge were split sibling-module style (`radix_tree/inner/evict.rs`, `moe/forward_k2/{unified_t,originals}.rs`, etc.) — local simulation of the workflow reports 0 violations.

Real CI should now run. A Laguna serve smoke on a GB10 is the remaining prudence before merge — happy to run it on our box and post the log.

rsafier:
The promised GB10 smoke is done — and it caught a real merge-blocker first: main's fail-closed kernel audit (post-#388) refuses to boot this target, which predates the harvest convention — 35 `try_kernel` probes (dense-FFN MMQ/int8/w4a4, Q4K MMQ, moe_w4a16 e8m0/fp4 scale twins, w4a16 pipeline twins) don't compile in `laguna-s-2.1` and had no `[expected_absent]` declarations. Pushed the harvest with per-entry fallback rationales (nemotron's sections as template); `--check-kernels` now exits 0 (0 unresolved / 35 declared).

Smoke on the branch binary (Laguna-S-2.1-NVFP4, fp8 KV, prefix caching): **boot 44s, 3/3 temp-0 facts (incl. the Pride-&-Prejudice-1813 canary), ~18 tok/s decode, zero errors in the serve log.**

rsafier:
CI triage for the five red checks on the previous tip (`06812cea` / merge `5df0c9b`) — all fixed in the new head. Root causes:

| Check | Root cause | Fix |
|---|---|---|
| `cargo clippy --tests` | `clippy::items_after_test_module` in `spark-model/src/layers/qwen3_attention/mod.rs` — the host-timing accumulators were appended after `split_ref_seqs_tests` | moved the `#[cfg(test)]` module to the bottom of the file |
| `kernel shadow structure` | RULE2: `laguna-s-2.1/nvfp4` carried regular byte-identical copies of `moe_w4a16_grouped_gemm.cu` (= minimax-m2-229b's) and `w4a16_gemm.cu` (= deepseek-v4-flash's) | both are now symlinks to the canonical files, per the existing convention |
| `cargo test --workspace` | `no_model_shadow_drops_a_common_kernel`: the Laguna-only kernels added to `common/` (`rms_norm_residual_vanilla`, `residual_add_rms_norm_vanilla`, `rope_forward_yarn_scaled`) registered as drops in 19 pre-existing model shadows | warning-scoped `[shadow_exempt]` entries: the norm pair in `common/KERNEL.toml` (resolved only behind `ships_vanilla_norm_weights()`, and Laguna doesn't shadow `rms_norm.cu`), `rope_forward_yarn_scaled` in `mistral-small-4`'s existing rope list (rotate_half vs interleaved, same rationale as the four variants already there). The startup audit still hard-errors if a model's dispatch ever resolves one. |
| `cargo llvm-cov --workspace` | same shadow-detector failure, plus `w4a16_gemm_t_ldb_drift_is_exactly_the_known_set`: laguna's `w4a16_gemm.cu` was a stale 8-arg (`ldb`-less) fork of the pre-port deepseek copy | the `w4a16_gemm.cu` symlink above — on the PR merge laguna now inherits deepseek's `ldb`-ported copy instead of pinning drift |
| `Build mdBook + rustdoc` | broken intra-doc link in `spark-runtime/src/flashinfer/hd128.rs`: `` [`ragged_prefill_bf16_hd256`] `` — the item lives in the parent module | linked as `` [`super::ragged_prefill_bf16_hd256`] `` |

Local verification, run both on the branch tip and on a simulated merge with `avarok/main` (which is what CI actually tests — the `ldb` drift test only exists on main):

- `python3 scripts/check_kernel_shadows.py` → OK (was 2 RULE2 violations)
- `cargo fmt --check` → clean
- `cargo clippy --workspace --tests` → clean
- `cargo test --workspace --locked` → branch: 3898 passed / 0 failed; simulated merge: **3987 passed / 0 failed** (includes `no_model_shadow_drops_a_common_kernel` and `w4a16_gemm_t_ldb_drift_is_exactly_the_known_set`)
- `cargo doc --workspace --no-deps` → clean (rustdoc leg of Docs)
- real `cargo check -p spark-server` with the full gb10 kernel build (no `ATLAS_SKIP_BUILD`) → clean

`PR benchmark gate` and `windows-x86_64-amd-hip` are known/owed and not addressed here.

rsafier:
Re-synced onto post-#440 main — the wave's #285 capability conversion (`is_mla: bool`) met this branch's batch-kernel work in three files (import/re-export unions) and one of our new tests still passed a model_type string (E0308 on the merged tree; `laguna` is not MLA → `false`). fmt/clippy/2250 tests green.

rsafier:
Re-synced onto latest main (#445 lineage). One real port in the merge: main split `build_choice_message`/`build_logprobs` into `chat_blocking_choice.rs` while this branch had improved the former in place — the branch's version (parser-aware `extract_hoisted_tool_calls`, `merge_hoisted_tool_calls` dedup, orphan-markup strip, all with their tests) now lives in main's new home, with the helpers re-exported from `chat_blocking.rs`. Also unioned main's new `stop_match_corrected` + tests alongside this branch's test module. fmt/clippy `-D warnings`/2282 tests green, cap-scan clean.

rsafier:
Re-synced onto current main (`93ae8d84`, non-force). One conflict: `decode_a2.rs`'s inline lm-head ladder vs main's refactor into the shared `lm_head_project_batched` — took main's side wholesale after verifying the shared ladder already carries every arm this branch had inline (FP8 per-row, lossless BF16-TC tile, t-kernel tile, batch4/8/16 GEMV tiers, BF16 batched GEMV). fmt/clippy green, MERGEABLE again.

Also for the record — serve re-smoke on the pre-merge tip `2c82c2b0` (Laguna-S-2.1 NVFP4, GB10): **3/3 PASS** — plain fact; tool-call request returns valid `tool_calls` JSON (`{"city":"Paris"}") with `finish_reason: tool_calls` and no raw `<tool_call>` markup in content; `enable_thinking:false` shows no think-tag leakage. One environment note: the upstream HF snapshot of Laguna-S-2.1 recently changed to a `{% include 'chat_template.jinja' %}` stub plus `{% generation %}` tags, neither of which the minijinja path resolves yet — the smoke ran with the template staged inline and the generation tags stripped. That's a model-repo-drift issue independent of this PR (affects main identically); flagging so it doesn't surprise the gate run.

rsafier:
Closing as **absorbed** — this PR's content is on `main`.

It landed via `680b3a56` (#473), which despite its TUI title carried 295 files / +22,318 lines including the whole Laguna-S-2.1 tree: `kernels/gb10/laguna-s-2.1/`, the Laguna weight loader, and the rest. Verified mechanically: `git diff --diff-filter=A avarok/main..feat/laguna-s-2.1` returns **zero** files, i.e. nothing on this branch is missing from main; the remaining diff is only this branch being 74 files behind.

Same disposition as #381 earlier in this campaign — merged-by-content rather than by PR reference, so GitHub never auto-closed it. The two gate records I recorded against its tip (ttft-warm/cold at `e7229cd1`) are superseded by whatever the merged tip's records are.

Follow-on note for anyone tracing the stack: **#464 is closed for the same reason** (its DeepSeek-V4 constituents are also in main via #473), which orphaned #465's base branch — I'm re-basing that onto `main` now.

PR #352, opened 2026-07-22
feat(mtp): make the drafter context (prefill + cross-turn carry) the default

tbraun96:
Makes the MTP drafter context on by default. Measured end-to-end on GB10 (DGX Spark), MLPerf-edge agentic, 1007 samples, temp 0 / seed 42.

## Why

The shipping config was drafting **blind**: `ATLAS_MTP_DRAFTER_PREFILL` was never set, so the drafter never received the prompt — **0 drafter-prefill events across a full 1007-sample benchmark**, and 142 drafter KV rows at sequence position 10,098. Every prior e2e number was a blind-drafter baseline.

## Result (five e2e runs, all MLPerf compliance PASS)

| run | wall | TPOT p50 | TTFT p50 | IoU | establishes |
|---|---|---|---|---|---|
| baseline | 5757.5 / 5771.7 s | 52.1 | 1661.7 | 0.6240 / 0.6268 | noise floor: 0.0028 IoU, 14 s wall |
| ctrl (new binary) | 5839.3 s | 52.04 | 1665.5 | 0.6281 | no regression; build/box equivalent |
| prefill only | 5789.0 s | 52.20 | 1662.0 | 0.6263 | **prefill alone is worth nothing** |
| prefill + carry | 4984.5 s | 39.60 | 1557.1 | 0.6285 | **carry is the entire win** |
| **bare defaults** | **5023.4 s** | **39.95** | **1582.2** | **0.6254** | **ships by default, no flags** |

**TPOT −24%, wall −14.6% vs control. TTFT improved rather than regressed. IoU held.**

The `IoU cost` that justified keeping drafter-prefill off does not exist: 0.6285 vs 0.6281 against a 0.0028 noise floor.

## Design

New SSOT `crates/spark-model/src/model/drafter_context.rs`. **One switch, not two**, because they are not independent: carry is inert without prefill (its call site is nested inside `!mtp_prefill_hidden.is_null()`), and prefill without carry does the work and discards it. `resolve()` enforces `carry => prefill`; disabling carry alone is unreachable.

| environment | prefill | carry |
|---|---|---|
| *(nothing set)* | ON | ON |
| `ATLAS_NO_MTP_DRAFTER_CONTEXT=1` | OFF | OFF |
| `ATLAS_MTP_DRAFTER_CONTEXT_PREFILL_ONLY_UNSAFE=1` | ON | OFF (research arm, warns) |

Everything is `== Some("1")`, so `ATLAS_*=0` disables nothing (pinned by a test). Obsolete opt-in names are ignored and their presence is warned at startup.

Startup now logs `MTP drafter context: prefill=ON carry=ON (default)` — previously neither flag was verifiable from the serve log, only from the container env.

The prompt-hidden capture is 336 MB at 32k / 2.7 GB at 256k, so the allocation is additionally gated on the MTP head supporting batched prefill.

## Verification

- `cargo check --workspace --all-targets` clean; release build 158 kernels.
- 7/7 resolver unit tests (all four combinations, the coupling, and polarity); `spark-model` 213/213, `spark-runtime` 145 pass / 2 ignored.
- Gate C2 NVFP4 tool-call + coherence smoke on `centml/Qwen3.6-27B-NVFP4-W4A4-mlpinf`: 3/3 tool calls, 3/3 coherent, 0 degenerate — run before any perf number.
- Equivalence: a bare serve is step-for-step identical to explicit flags (same cold prefill, same `adopted rows=1785 appended=108`, same ~10.5 ms append); the kill switch yields 0 prefill events and 0 adopts.

## Not covered

Gates A / B / D were not run. Clippy reports 4 errors that are pre-existing on the parent commit at lines this branch does not modify.


tbraun96:

```
-110,266 lines deleted
```

- It is **CONFLICTING**, and GitHub does not run merge-commit workflows for a PR it cannot merge — so every green tick displayed here describes a tree from 2026-07-22 that nobody will merge. Thirteen open PRs are in that state; #427 adds a `needs-rebase` label so the false green stops being invisible.

PR #230, opened 2026-07-02
perf(strix): Qwen3.6-27B NVFP4 on Strix Halo (gfx1151) — beats llama.cpp on decode/prefill/accuracy at matched 4-bit + warm-prefill

tbraun96:
## Strix Halo (gfx1151) — Qwen3.6-27B NVFP4

Native-HIP Atlas (`spark`) serving `unsloth/Qwen3.6-27B-NVFP4` on a single AMD Strix
Halo (Radeon 8060S, gfx1151, 60 GB GTT). At **matched 4-bit quality** it beats
llama.cpp on decode, prefill, and accuracy on the same box:

| Front | **Atlas NVFP4 + MTP** | llama.cpp 4-bit |
|---|---|---|
| Decode (tok/s) | **16.2 – 16.7** | Q4_K_M 10.9 · rocmfp4-lean 12.5 |
| Prefill, cold (tok/s) | **~178** | 58.8 |
| Prefill, warm partial-hit | **1.46 s** (was 4.20 s) | — |
| BFCL-v4 accuracy | **89.82** | 88.60 |

All Atlas numbers gated: coherence 14/14, `tok_agree=1.0`, `kl_mean=1e-5`, MTP eff 1.72×.
Recommended serve config + validation in `docs/porting/STRIX_NVFP4_MLPERF.md`.

**New this line:** warm-prefill `--ssm-checkpoint-interval 16 --ssm-cache-slots 48`
(output-faithful, 2.9× faster multi-turn partial-hit prefill; slots=48 fits the 60 GB GTT).

_Note: syncing this branch with `main` (+ applicable prefill changes from #229) is in
progress; a follow-up push updates this diff to a clean vs-main delta._

Issue #326, opened 2026-07-16
[bug] native-HIP build fails on a clean ROCm box: spark-storage/build.rs unconditionally invokes nvcc (ATLAS_TARGET_HW=strix-hip)


botAGI:
**Atlas version / commit:** branch `strix-accfix2` @ `82e88f242949029e04f828122d6e250ec9a43db0` (built from source). The same code is present on `main` @ `45cc093` — `crates/spark-storage/build.rs` has the identical `find_nvcc()` / panic path and no HIP awareness there either.

**Model:** n/a (build-time failure, before any model is involved)

**Hardware:** AMD Strix Halo — Ryzen AI Max+ 395 w/ Radeon 8060S (gfx1151), 128 GB unified. ROCm 7.2.3 (hipcc 7.2.53211). **No CUDA toolkit and no SCALE installed** — pure-ROCm box.

**Launch (build) command:**

```bash
ATLAS_TARGET_HW=strix-hip ATLAS_TARGET_MODEL=qwen3.6-27b ATLAS_TARGET_QUANT=nvfp4 \
CUDARC_CUDA_VERSION=12080 \
cargo build --release -p spark-server --no-default-features --features cuda
```

**Expected behaviour:** the whole workspace builds through the HIP path. `crates/atlas-kernels` does this correctly — `HARDWARE.toml` `vendor="hip"` routes to `HipTarget`, all 92 kernels compile via hipcc, and the libcuda→HIP shim is built automatically.

**Actual behaviour:**

```
error: failed to run custom build command for `spark-storage v0.1.0 (.../crates/spark-storage)`

Caused by:
  process didn't exit successfully: `.../target/release/build/spark-storage-.../build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=ATLAS_SKIP_BUILD
  cargo:rerun-if-env-changed=SKIP_ATLAS_BUILD
  cargo:rustc-link-search=native=/usr/lib/x86_64-linux-gnu
  cargo:rustc-link-lib=dylib=cuda
  cargo:rerun-if-changed=.../crates/spark-storage/kernels/q_lowrank_project.cu

  --- stderr
  thread 'main' panicked at crates/spark-storage/build.rs:118:33:
  invoke nvcc (nvcc): No such file or directory (os error 2)
```

(line:col is against the unpatched branch tip; full log captured from the failing run, can attach.)

[...]

PR #353, opened 2026-07-22
perf(strix): Qwen3.6-27B-NVFP4 on Strix Halo (gfx1151) — Atlas beats llama.cpp on decode, prefill & accuracy

AzeezIsh:
## Atlas on Strix Halo beats llama.cpp at matched 4-bit

Qwen3.6-27B-NVFP4 on **AMD Strix Halo (gfx1151 / RDNA3.5)**, measured end-to-end on the MLPerf-edge agentic benchmark (1007 samples, temp 0 / seed 42) against the llama.cpp Q4_K_M baseline on the **same silicon**. Atlas wins decode, warm prefill, and finishes the whole set at tied accuracy.

This is the **one consolidated Strix branch** — it supersedes the earlier fragmentary strix PRs (#230 `pr/strix-rocmfp4` is a clean ancestor). It is merged up to `main` (behind 0) so CI actually runs. The **GB10/DGX line is tracked separately** (PR #352) — different hardware, different kernels, different checkpoint; do not cross the numbers.

### End-to-end result (`atlas_desktop_win`, strix-desktop, 125 GB unified)

| metric | Atlas NVFP4 | llama.cpp Q4_K_M | winner |
|---|---:|---:|:--|
| **TPOT p50** | **61.99 ms** (16.1 tok/s) | ~94 ms (10.5–10.95 t/s) | **Atlas ~1.5× decode** |
| TTFT warm p50 | ~2.1 s | 2.21 s | Atlas |
| Wall to run the set | **2.81 h** — 1007/1007 | 4.0 h — 964/1007 (unfinished) | **Atlas ~1.5×** |
| BFCL v4 | **86.03 / 87.71 norm** | (not run) | — |
| IoU (accuracy) | 0.6245 | 0.6265 | tie (inside MLPerf MDE) |

MLPerf-edge compliance **PASS** (thresholds 83.64 / 85.32; 995 samples; 2.81 h < 4 h). On Strix the win is **decode** — the opposite of the GB10-vs-vLLM picture.

### Advancements in this branch

- **NVFP4 GEMM on gfx1151** — software E2M1 path; native tensor-core prefill.
- **DP4A W4A8 integer decode** GEMV kernels wired into dense-FFN (+25% decode, flag-gated).
- **BF16-WMMA + NVFP4 tensor-core prefill** (full-attention + GDN) — beats llama.cpp wall-time.
- **Tool multi-call + grammar-completion EOS** — Strix BFCL **89.22 > llama.cpp 88.02**.
- **Tail-pin / mid-chunk SSM eviction** — warm-TTFT parity with llama.cpp.
- **MTP speculative decode** (num-drafts 1 = K=2 optimum, 92% accept).
- **MTP drafter context-prefill + CARRY** (flag-gated `ATLAS_MTP_DRAFTER_PREFILL=1 ATLAS_MTP_CARRY_DRAFTER=1`). Honest note: on Strix this measured as **noise** (16.21 vs 16.25 tok/s) because acceptance is already ~92% — it is the entire win on GB10 (#352) but does not port here, so it stays **off by default** on Strix. Included for parity + future headroom.
- **batch-3 lm_head arm** (`num_tokens==3 → w4a16_gemv_batch3`) — cuts the K=3 MTP verify GEMM **267 → 124 ms**, flips the gate DISABLED→ENABLED at nd=2, makes depth-2 coherent + **greedy token-identical** to nd=1. Unblocks any future depth-2 gain; nd=1 remains the production default.


botAGI:
## External validation — second Strix Halo box (128 GB)

Ran this branch end-to-end on our own hardware. TL;DR: **decode reproduces and exceeds the table, warm-restore is excellent, cold prefill is the one remaining gap vs llama.cpp.**

**Environment:** Ryzen AI Max+ 395 / Radeon 8060S (gfx1151, RDNA3.5), 128 GB unified, ROCm 7.2.3, native-HIP (no SCALE). Branch @ `57a5771`, model `nvidia/Qwen3.6-27B-NVFP4` (rev `0893e160`), served with the recipe from the PR (util 0.65, `--ssm-cache-slots 64`, `--disable-thinking`). Same-box llama.cpp figures are `server-rocm` `-fa on`, Q4_K_M, measured last week.

### Decode — reproduces and then some

| metric | this branch | table claim |
|---|---:|---:|
| decode, short prompt (server-side, median of 5) | **19.3 tok/s** | 16.1 |
| decode, 256-tok (server-side, median of 3) | **21.1 tok/s** | — |
| decode, client-wall (short / long) | 17.7 / 19.8 | — |

Steady-state K=2 MTP acceptance ~67%. First post-boot request runs ~13.4 tok/s before MTP warms.

### Warm SSM restore — confirmed, 139×

Clean cold→warm on a fresh 8k-token prefix (identical reprompt):

```
COLD 51.6 s  →  WARM 0.37 s   (139×)
```

Log shows the SSM snapshot restoring on the identical prefix. (One caveat: with certain intervening request orderings I saw `Prefix cache hit … but no SSM snapshot` and warm collapsing to cold — looked like a block-alignment edge case, ~1 block short of the saved snapshot; flagging in case it isn't expected.)

### Cold TTFT — the remaining gap

Cold prefill on 8k: **~51 s (~155 tok/s)**. llama.cpp ROCm on the same box/prompt did **~39 s (~202 tok/s)**. So cold prefill throughput still trails llama.cpp by ~30%. It's a one-time cost that the 139× warm restore amortizes across an agentic run's later turns, but single-shot long-context is where it stings.

### MTP drafter-prefill levers

`ATLAS_MTP_DRAFTER_PREFILL=1 ATLAS_MTP_CARRY_DRAFTER=1` on a single-turn decode A/B moved server-side tok/s **+0.7% (short) / +1.3% (long)** — within noise. Expected: this is the wrong workload for the lever. Happy to run the actual MLPerf-edge agentic set to test the +7% if you point me at the harness.

### Two snags reproducing the build recipe cold

1. `cargo build … -p spark-server` with no `ATLAS_TARGET_MODEL`/`ATLAS_TARGET_QUANT` defaults to a `qwen3-next-80b-a3b` kernel dir that doesn't exist under `kernels/strix-hip/` → build panics at `build.rs:703`. Setting `ATLAS_TARGET_MODEL=qwen3.6-27b ATLAS_TARGET_QUANT=nvfp4` fixes it.
2. Plain `cargo build` links `nccl` (default feature) → `unable to find library -lnccl` on a single-GPU box without NCCL. `--no-default-features --features cuda` fixes it.

Also nice to see the `#326` build fix (spark-storage nvcc bypass on HIP targets) landed on this branch — built clean without the local patch.

Raw JSON results and serve logs available on request.

AzeezIsh:
## Reproduction recipe (standard serve) — the runtime flags are load-bearing

The code in this PR (`791483fd`) is complete — but the headline decode + warm-restore numbers are **flag-gated and OFF by default**. Building and serving from the bare recipe yields the *default* binary (DP4A off, global-GDN off, SSM tail-protect off) → baseline-slow decode and no warm-restore, which is why the table hasn't been reproducing for folks copy-pasting cold. Here is the full verified incantation that produced the numbers in this PR, for a **plain serve** (no benchmark harness).

### Build — native-HIP, gfx1151 / Strix Halo
```bash
export ATLAS_TARGET_HW=strix-hip \
       ATLAS_TARGET_MODEL=qwen3.6-27b \
       ATLAS_TARGET_QUANT=nvfp4 \
       ATLAS_HIP_COMPAT_INCLUDE=$PWD/crates/atlas-kernels/hip/compat \
       ATLAS_HIPCC=/opt/rocm/bin/hipcc \
       CUDARC_CUDA_VERSION=12080 \
       ATLAS_HIPCC_WORKERS=4     # cap parallel hipcc so a 60 GB box doesn't OOM-wedge

cargo build --release -p spark-server --no-default-features --features cuda
```

Two gotchas that bite everyone building cold:
- **`ATLAS_TARGET_MODEL` / `ATLAS_TARGET_QUANT` are required.** Unset, the build defaults to a `qwen3-next-80b-a3b` kernel dir (`crates/atlas-kernels/build.rs`) that doesn't exist under `strix-hip`.
- **`--no-default-features --features cuda` is required.** Plain `cargo build -p spark-server` pulls in `nccl` (a default feature) → `#[link(name="nccl")]` fails to link on a single-GPU box.

### Serve — the `ATLAS_*` block *is* the "beauty"
```bash
# gfx1151 runtime libs (+ your hip-port/link shim from #326 if you use it)
export LD_LIBRARY_PATH=/opt/rocm/core-7.13/lib:/opt/rocm/lib

# flag-gated features — all OFF by default; THIS is what the recipe was missing:
export ATLAS_W4A16_DP4A=1          # +25% DP4A W4A8 integer decode   <-- the big one
export ATLAS_FORCE_GLOBAL_GDN=1
export ATLAS_W4A16_VARIANT=v1
export ATLAS_SSM_TAIL_MIDCHUNK=1   # tail-pin / warm-restore (the 139x cold->warm)
export ATLAS_SSM_TAIL_PROTECT=1
export ATLAS_MTP_GATE_REPROBE=64
export ATLAS_KV_EXTERNAL_RESERVE_GB=6   # box-specific — tune to your GTT pool

spark serve <SNAPSHOT> \
  --model-name nvidia/Qwen3.6-27B-NVFP4 --host 0.0.0.0 --port 8081 \
  --max-seq-len 65536 --gpu-memory-utilization 0.35 --kv-cache-dtype bf16 --max-batch-size 1 \
  --speculative --num-drafts 1 --mtp-quantization bf16 --mtp-vocab 100000 \
  --disable-tool-grammar true --enable-prefix-caching \
  --ssm-cache-slots 64 --ssm-checkpoint-interval 16 --disable-thinking
```

Quick sanity check once it's up:
```bash
curl -s http://localhost:8081/v1/models   # -> nvidia/Qwen3.6-27B-NVFP4
```

**Box-specific knobs:** `--gpu-memory-utilization 0.35` + `ATLAS_KV_EXTERNAL_RESERVE_GB=6` are tuned for a ~125 GB GTT pool (0.35 leaves headroom for the SSM cache). A 128 GB box will want its own value — bump if you have room, drop if you OOM.

**On draft width:** `--num-drafts 1` is MTP **K=2**, the measured-fastest verify width for this model. K=3 / depth-2 is *slower* here (acceptance is already ~92%, so the extra draft width costs more verify than it saves) — see the lever-folding notes. The MTP drafter context-prefill levers (`ATLAS_MTP_DRAFTER_PREFILL=1 ATLAS_MTP_CARRY_DRAFTER=1`) are the win on GB10 (#352) but measured as noise on Strix, so they stay off here.

TL;DR of what was missing to reproduce: **not code — the `ATLAS_*` runtime env block above (DP4A + global-GDN + SSM tail-protect) plus the two build flags.** With them set, decode reproduces (~19 tok/s short / ~21 long, median 18.9) and the SSM warm-restore lands.

AzeezIsh:
Measured wins on strix-desktop (gfx1151, Qwen3.6-27B-NVFP4) this session — all gated on coherence + first-token KL + no-regression:

- **Refeed (drafter context-prefill)** — already on this branch (57a5771) as opt-in; #352 makes it default. Per-position acceptance 73 → ~90%, warm decode TPOT 46.5 → 44.7 ms.
- **K=3 speculation (2 drafts)** — the batch3 LM-head dispatch (79324b9) is the unlock. verify_multiplier **3.48 → 1.44** (verify 254 → 107 ms) → K=3 net-positive. Warm decode TPOT **44.7 → ~40 ms**, mean accepted 1.5–1.9 tok/verify. vLLM decode gap **1.42× → 1.29×**. Root cause of the old superlinearity was the M=3 LM head falling to the generic `w4a16_gemm` (146 ms = 55% of verify), **not** the GDN recurrence (already batched/linear — the "batch the FLA verify" theory is refuted on strix by rocprofv3).
- **Cold-TTFT prefill** — #365 (stacked): smem bank-conflict pad + dead-NaN dequant → cold 8k prefill **247.8 → 258.7 tok/s (+4.4%)**, bit-identical.

**Combined: decode 46.5 → ~40 ms (−14%), cold prefill +4%.** K=4 (3 drafts) still net-negative — the FFN/attention M=4 paths are unbatched (a `dual_batch4` follow-up); K=3 is the current sweet spot. Full submission e2e (K=3 + refeed + cold-TTFT, on the tagged consolidate base) running now on endpoints edc7ea0; wall/BFCL/IoU/TPOT to follow.

AzeezIsh:
Pushed three things onto this branch — the TTFT tail fix, the cold-prefill fix, and a rewritten repro doc. Both perf changes are bit-identical, so this is throughput only, no output movement.

**`cdd4de0` — warm-TTFT tail (p90 / p99 / max)**

Root cause was not the snapshot capture/lookup alignment I'd assumed, it was `compute_session_hash`: it hashes the first `min(len, 1024)` prompt tokens, and multi-turn prompts *grow*, so while the prompt is under 1024 tokens the hashed window keeps sweeping up the conversation tail and the hash changes every turn. SSM snapshot restore is session-gated in four places, so a snapshot saved on turn N was unfindable on turn N+1 and every warm turn recomputed all KV. The KV radix isn't session-gated, which is exactly why KV hit while SSM missed — that's the `Prefix cache hit: N tokens but no SSM snapshot` line.

Fix keys non-tail snapshots by content and leaves the session gate only where it's actually needed (`is_tail` entries, which are the only ones whose state bleeds past their `token_count`), plus captures on any warm continuation rather than only on a stable session hash.

| | before | after |
|---|---|---|
| warm TTFT p90 | 6073 ms | **2348 ms** |
| p99 | 18550 ms | — |
| max | 77560 ms | **2607 ms** |

A 12-turn probe out to ~4.3k ctx is now flat with depth (2261 → 2607 ms) instead of climbing.

**`3209543` — cold prefill**

`rocprofv3 --kernel-trace` on an 8k cold prefill: `w4a16_gemm_t_m128` is **74.5%** of it (18.4 s of 24.7 s). The compiler explains why — that kernel takes 34432 B of LDS and 256 VGPRs *with 3 spilled to scratch*, which resolves to **3 waves/SIMD**. The plain M_TILE=64 `w4a16_gemm_t` runs the identical NVFP4 math in 24192 B with no spill.

Measured on the real FFN shapes (new `examples/strix_ffn_bench`), M64 vs m128, TFLOP/s:

```
gate/up M=17    3.91 / 1.96  2.00x      down M=17    2.81 / 1.86  1.51x
gate/up M=64   14.92 / 7.14  2.09x      down M=64    8.89 / 6.37  1.40x
gate/up M=512  19.34 /15.94  1.21x      down M=512  14.82 /14.25  1.04x
gate/up M=2048 19.58 /16.75  1.17x      down M=2048 17.04 /14.55  1.17x
```

M64 wins every shape, so it's now the native-HIP default. The same sweep also kills `w4a16_gemm_t_k64` (0.29–0.77x — K_STEP=64 halves barriers but costs 45248 B of LDS, which trades away more resident CTAs than the barriers were worth); it was live for `down` at k≥8192 on the small-M path and is now skipped on `atlas_hip` only. NVIDIA is untouched — the whole thing is behind `cfg!(atlas_hip)`.

Output is bit-identical rather than merely close: the per-output-element K accumulation order doesn't change, only which CTA owns which m-rows. Verified two ways — a byte-compare of m128 vs M64 on all ten shapes, and a prompt-logprob diff over **6484 scored prefill positions** (`max|dlogprob| = 0`, `max KL = 0`, every token identical).

End to end: cold prefill **256 → 276 tok/s**, 8k cold TTFT **25.7 → 23.9 s**. It compounds with the snapshot fix rather than overlapping — that one cuts *how many* tokens a warm turn prefills, this cuts *how long each takes*.

**`a10700a` — repro doc + harness**

`docs/porting/STRIX_NVFP4_MLPERF.md` still described the pre-submission state (a different checkpoint, the 60 GB laptop, `--num-drafts 1` at util 0.85, decode ~16 tok/s), so following it reproduced neither the submission nor current numbers. Rewritten around what was actually submitted, with the build/serve/benchmark invocations and the mandated seeds.

@botAGI — re: your offer to run the full set, the harness is now in `bench/strix_mlperf/` with a README. The one worth knowing about is the drift gate: `/v1/completions` with `echo=true, logprobs=5` returns **prompt** logprobs, scored entirely by the prefill path with no decode/MTP/sampling in the loop, so a 6k prompt gives ~4800 independently checked positions. That's what proves a prefill kernel change is bit-identical rather than "cosine 0.999".

Also documented the two traps that cost the most time here: `ATLAS_TARGET_MODEL=qwen3.6-27b` is mandatory or a kernel edit produces an md5-identical binary, and the shipped edge-agentic yaml points at `:8080` (llama's port), which silently errors all 1007 samples.

**On the cold-prefill gap you measured** (~155 vs llama's ~202 tok/s on your box): that's the right target and it's still the open one. We're at ~34% MFU on gate/up against ~59 TFLOP/s bf16 peak, so there's roughly another 1.7x of headroom in that kernel. Next levers are an LDS reduction on the M64 tiling (24192 → under 21845 buys a 6th resident CTA, which needs an XOR swizzle rather than more padding — 16-byte alignment and an odd dword stride can't both be had by padding), and an int8 prefill GEMM. Worth flagging that `int8_gemm_faith2`, which was +49% on GB10, is **not** portable here: it's built on `ldmatrix` and `cp.async`, neither of which exists on gfx11, so it's a reimplementation rather than a port.

**One correction to carry:** K=4 is untested on this branch, not disproven. I measured `--num-drafts 3` at 32% slower than K=3, but that was with **no `n == 4` arm** in `layers/qwen3_attention/trait_impl/multi_seq/ffn.rs` — it has `n == 3` and `n == 2`, so n=4 falls into `forward_prefill` (GEMM tiles at M=4) instead of the batched `forward_k4`, which is the same MMQ cliff the gb10 side found and fixed. Porting that arm and re-testing is next.

tbraun96:
## Merged main forward — and this branch now builds for Windows AMD

This PR had been `CONFLICTING` since 07-22, which meant every `pull_request` workflow was
silently skipping. `gh pr checks` looked fine because only the `pull_request_target` jobs
(CLA, GitGuardian) were still firing — the real gates hadn't run in five days, and the
`windows/amd` native-HIP row (added in #358/#359, after this branch's last CI) had never run
here at all.

Merged `main` in (`53425206`), so CI is live again and all the gates are green.

### Conflict resolutions — GB10 stays byte-for-byte

- **`prefill_paged_compute.cuh`** — kept this branch's `smem_K` buffer-share (it removed
  `smem_V`, −8.25 KB LDS) *and* main's `(__bf16)(float)` cast routing, which the Windows HIP
  SDK needs to disambiguate `__hip_bfloat16`'s ~14 implicit conversion operators. The fp16 arm
  two lines below had already auto-merged to exactly that combined form.
- **`impl_a3.rs`** — kept **both** lm-head arms, with the Strix `batch3` (K=3 verify) arm now
  gated on `cfg!(atlas_scale)`. That kernel lives in the shared `kernels/gb10` tree, so
  `try_kernel` resolves it on NVIDIA too; ungated, GB10 would have silently moved
  `num_tokens == 3` off main's batch4 arm. `cfg!()` not `#[cfg]`, so it still type-checks
  everywhere.
- **`multi_seq/ffn.rs`** — dropped this branch's `n == 4` arm. main's `(4..=8)` arm is the
  general form and gates on `can_forward_km` at branch entry, which is exactly the half-done
  fallthrough the older arm could hit.
- **`mtp_head.rs` / `speculative.rs`** — main had split `DraftProposer` into
  `mtp_head/draft_proposer.rs` and landed the same carry-drafter feature independently. Its
  copies are supersets (identical code, better comments), so the inline versions go.
- **`save_checkpoint.rs`** — took main's always-on `is_prompt_tail` and kept this branch's
  `is_tail` as a separate opt-in (`ATLAS_SSM_TAIL_CKPT`, default off). The submission run never
  set that env var, so its `is_tail` was dead code there — nothing regresses either way.

### Two latent bugs the merge surfaced

Both merged **without a conflict** and passed every local check, so they only appeared on the runner:

- `spark-runtime/build.rs` had dropped main's `ATLAS_SKIP_BUILD` link block, so the no-GPU CI
  test build linked with no `-lcuda` at all (`undefined symbol: cuLaunchKernel`, …). Restored
  main's version — emitting `-lcuda -lcublasLt -lcudart` is correct on every target now, since
  `build_hip_shim` builds those three on Linux and the matching import libs on Windows. This
  branch's removal predates those shims.
- AMD clippy was broken: routing cuBLASLt/CUTLASS to the metal stubs under `atlas_scale` pulls
  in five 8–9 arg FFI signatures that trip `clippy::too_many_arguments` under
  `#![deny(clippy::all)]`. Never went red because CI only clippies the NVIDIA config, but it's
  what anyone developing the AMD path hits. Also split the one vendor-forked method out of
  `gpu_impl.rs`, which had crossed the 500-LoC cap.

### Windows AMD: green, 97/97 kernels

`windows-x86_64-amd-hip` builds and packages the same kernel triple as the Linux golden run:

```
compiled 97 kernels for target 0 (strix-hip, qwen3.6-27b, nvfp4)
97/97 unique hipcc invocations
built Windows HIP runtime shim (cuda.dll/nvcuda.dll + import libs, 76 exports)
```

**97 is the number to check.** main alone builds 91 — the six this branch adds
(`w4a16_gemv_dp4a`, `wht_bf16`, `moe_expert_gemv_dp4a`, `dense_gemv_bf16_batch2`,
`inferspark_prefill_paged_turbo{4,8}`) are exactly the set it was missing.

### Validated on real gfx1151

Windows AMD hardware doesn't exist in the fleet, so the closest available check is the same
native-HIP path on Linux gfx1151 (ROCm 7.13), built from this branch's tip:

- 97 kernels via hipcc, HIP shims built, binary linked, exit 0
- serves `nvidia/Qwen3.6-27B-NVFP4` with MTP K=3 — coherence **and** tool-call smoke both pass,
  0 errors, `97 kernels embedded`, warm TTFT 301 ms

That covers everything except the Windows OS layer itself.

### For anyone testing on Windows

Docs are in `docs/porting/STRIX_WINDOWS_HIP.md`. Two things that are not obvious from the Linux
recipe:

1. Memory is a fixed WDDM **Variable Graphics Memory** carve-out, not a growable GTT pool, so
   the Linux `--gpu-memory-utilization 0.35` doesn't transliterate — set VGM high *before*
   serving. And there's no safety net: the OOM fallback (`load_fns.rs:130`, whose log says
   "paged via Linux swap") calls `hipMallocManaged`, which AMD doesn't support on Windows, so a
   too-small carve-out is a hard error rather than a slow degrade.
2. Don't expect the Linux timings. Windows HIP SDK is ROCm 6.4-era vs 7.13, and WDDM routes
   every launch through the OS scheduler with no persistent queues — the worst case for a
   ~39 ms-TPOT decode loop. Worth measuring launch overhead before investing anywhere else.

Runtime on Windows is still **unverified**. Most useful things to report back: the kernel count
at load (expect 97), whether output is coherent, what VGM/utilization actually fits, and TTFT +
TPOT so we can size the WDDM gap.

tbraun96:
Triaged for a certification campaign and **parked pending a rebase**. Merged onto current `main` (`5f639bfc57`) locally with the rename limit raised: **32 conflicting files**, 271 commits behind, 51 own commits over 78 files.

Your other Strix PR, **#592, is far cheaper** — only 8 conflicting files — so if you rebase one first, rebase that one; it may also reduce what is left here.

What moved underneath you: `569ffd0920` (#1101) renamed five crate directories (`crates/atlas-* → crates/avarok-*`) plus `atlas_* → avarok_*`, `ATLAS_* → AVAROK_*`, `cfg(atlas_*) → cfg(avarok_*)`. Raise `merge.renameLimit=20000` before estimating — git's default under-reports the conflict set badly at this size. Full write-up: #1206.

PR #358, opened 2026-07-23
ci(release): windows/amd (native-HIP) target + early version-collision guard

tbraun96:
## What

Two release-workflow changes.

### 1. Early version guard (fixes the release that died at the very end)

Manual release run [29939065065](https://github.com/Avarok-Cybersecurity/atlas/actions/runs/29939065065) built the whole 8-target matrix for ~17 minutes and then failed in `publish`. Root cause is a no-op version bump. `scripts/release-notes.sh bump` rewrites `[workspace.package].version` to the value you passed, and when that value already matches what is on the ref, Cargo.toml does not change, so `git commit` fails with "nothing to commit". The existing tag guard does not catch it because a first release sits at `0.1.0` with no `v0.1.0` tag yet, so `new_version=0.1.0` sails straight past the tag check and detonates at the end.

New `validate-inputs` step reads the version from the ref being released and fails in seconds if `new_version` equals it. The read goes through a new `release-notes.sh current` subcommand so the check and the bump share one table-aware parser (SSOT). `validate-inputs` now checks out the release ref itself so the version it reads is the one `publish` will bump. The remote-facing tag and branch guards are unchanged.

Equality only, on purpose. A strict `sort -V` monotonic check orders pre-releases opposite to semver (it ranks `0.1.0-rc1` above `0.1.0`) and would wrongly reject a legitimate rc to GA release. Equality is unambiguous and is exactly the failure that was reported.

### 2. windows/amd build target (native-HIP, experimental)

Adds the missing windows x amd quadrant. SCALE, Atlas's other AMD toolchain, ships Linux-only (Ubuntu/Rocky tarballs, bundled Linux ROCm), so it cannot run on a Windows runner. The only conceivable Windows AMD path is native-HIP, so this target uses `ATLAS_TARGET_HW=strix-hip` with `hipcc` from the Windows HIP SDK, arch `gfx1151`.

It is **compile-only** (hosted runners have no AMD GPU) and marked `experimental` + `pr_validate:true`, so it runs on PRs and dispatches but never blocks them (`continue-on-error`).

**Known endpoint.** It is expected to wall at the kernel-compile stage. About 20 kernels still emit NVIDIA `mma.sync` / `cp.async` PTX that hipcc cannot compile and are not yet ported to AMD WMMA (see `crates/atlas-kernels/hip/README-HIP.md`). That wall is OS-independent and is the same one the Linux native-HIP path has not cleared. The row is carried to exercise the Windows HIP toolchain up to that wall, not to ship a working artifact.

Build wiring
- new "Install HIP SDK (AMD, windows)" step downloads and silent-installs the HIP SDK, locates hipcc, and exports `ATLAS_HIPCC` / `HIP_PATH` and the strix-hip target env.
- the windows CUDA import-lib step is now gated to the `cuda` toolchain so the hip row does not throw on a missing NVIDIA `cuda.lib`. libclang setup stays shared across both windows toolchains.

## Ensuring it passes

The required PR checks stay green. The new hip job is experimental so its expected kernel-compile failure does not fail CI. The version guard is dispatch-only logic, verified locally across equality, downgrade, and pre-release cases.

## Notes / follow-ups
- `HIP_SDK_URL` is pinned in the workflow as a single knob. AMD has no documented stable direct-download link, so it may need a version bump if CI shows a 404.
- Getting this target to a real green compile needs the ~20 mma.sync kernels ported to AMD WMMA, then a `libcuda` to `nvcuda.dll` shim and stub `cudart` / `cublasLt` import libs for the link. All downstream of the current wall.

### DIFF
--- a/.github/release-matrix.json
+++ b/.github/release-matrix.json
@@ -90,5 +90,20 @@
     "toolchain": "cuda",
     "experimental": true,
     "pr_validate": true
+  },
+  {
+    "id": "windows-x86_64-amd-hip",
+    "os": "windows",
+    "runner": "windows-2022",
+    "target": "x86_64-pc-windows-msvc",
+    "platform": "amd",
+    "cargo_features": "--no-default-features --features cuda",
+    "toolchain": "hip",
+    "experimental": true,
+    "pr_validate": true,
+    "note": "Native-HIP AMD (gfx1151) build via the Windows HIP SDK (hipcc). SCALE, Atlas's other AMD toolchain, is Linux-only, so HIP is the only conceivable Windows AMD path. Compile-only — hosted runners have no AMD GPU. EXPERIMENTAL and expected to fail at the kernel-compile stage: ~20 kernels still emit NVIDIA mma.sync/cp.async PTX that hipcc/clang cannot compile and are not yet ported to AMD WMMA (see crates/atlas-kernels/hip/README-HIP.md) — the same wall the Linux native-HIP path has not cleared. Carried non-blocking (continue-on-error) to exercise the Windows HIP toolchain up to that wall.",
+    "scale_arch": "gfx1151",
+    "target_model": "qwen3.6-27b",
+    "target_quant": "nvfp4"
   }
 ]
--- a/.github/release-matrix.json
+++ b/.github/release-matrix.json
@@ -101,7 +101,7 @@
     "toolchain": "hip",
     "experimental": true,
     "pr_validate": true,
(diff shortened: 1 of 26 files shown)


tbraun96:
CI run on the windows/amd-hip target (before the note-accuracy fix). Recording where it actually walls so the follow-up is scoped.

Good news first. The Windows HIP toolchain stands up end to end. HIP SDK 25.Q3 installs headless, hipcc is found, and `clang.exe --offload-arch=gfx1151 -x hip` runs on the kernels. The job walls at the kernel compile, and the aggregate CI stays green (this row is experimental + continue-on-error, so a failed matrix instance does not drag down `dry-run-summary` or the run).

The wall is NOT the 20 mma.sync kernels. It is earlier, in the CUDA to HIP compat headers, because the Windows ROCm 6.4 clang differs from Linux ROCm:

- `error: use of undeclared identifier '__shfl_down_sync'; did you mean '__shfl_down'?` The Windows HIP clang has no mask-arg warp intrinsics. Only `__shfl_*` exist, not `__shfl_*_sync`. This hits almost every warp-shuffle kernel (gdn_reduce, dense_gemv, causal_conv1d, and so on).
- `error: ambiguous conversion for C-style cast from '__nv_bfloat16' (aka '__hip_bfloat16') to '__bf16'`.

So most kernels fail there, before the tensor-core kernels are even reached. Linux native-HIP compiles 72 of 92 because Linux ROCm does define `__shfl_*_sync`.

Real next step for a Windows HIP compile, in order. Extend `crates/atlas-kernels/hip/compat` to shim `__shfl_*_sync` and disambiguate the bf16 cast for the Windows clang. Then the mma.sync to WMMA port. Then link, which needs a `libcuda.so` to `nvcuda.dll` shim plus stub `cudart`/`cublasLt` import libs. All downstream of this wall.

Note updated to reflect the observed cause in the next commit.

PR #365, opened 2026-07-24
perf(strix): dense-FFN prefill smem bank-conflict fix + dead-NaN dequant (+4% cold prefill)

AzeezIsh:
Stacks on #353 (`strix/consolidate-tmp`). Two pure-layout / dead-code fixes to the gfx1151 dense-FFN prefill GEMM (`kernels/strix-hip/qwen3.6-27b/nvfp4/w4a16_gemm.cu`) — both bit-identical, gated on coherence + first-token KL + no-regression.

**1. smem_B_bf16 bank conflict (PADB=2).** The `[128][32]` bf16 tile row was 64 B = exactly 16 of the 32 LDS banks → 16-way store conflict. Pad to `[128][34]` (17 words, gcd(17,32)=1) → conflict-free. rocprofv3 confirms ~0 conflicts after.

**2. Branchless RNE dequant.** `__float2bfloat16` in the hot dequant loop compiles to a software round *with* a NaN branch (gfx1151 has no hardware bf16 convert); the inputs (`E2M1_LUT × scale`) are always finite, so that branch (9.4% of the K-loop) is dead. Replace with a bit-round → identical bits.

**Measured** (strix-desktop gfx1151, Qwen3.6-27B-NVFP4): cold 8k prefill **247.8 → 258.7 tok/s (+4.4%)**, bit-identical (20/20 first-tokens byte-identical), coherent, decode TPOT unchanged.

**Reverted en route:** a dequant-hoist two-pass — ~5× slower on gfx1151 (the fused inline dequant already overlaps the MMA pipe), so the profiler 8.8:1 dequant:WMMA issue-slot ratio overstated the real cost. Noted so it is not re-attempted.

### DIFF
--- a/kernels/strix-hip/qwen3.6-27b/nvfp4/w4a16_gemm.cu
+++ b/kernels/strix-hip/qwen3.6-27b/nvfp4/w4a16_gemm.cu
@@ -68,6 +68,7 @@ __device__ __forceinline__ unsigned short atlas_cvt_e4m3x2_f32(float a_hi, float
 #define PAD_T 8        // 16-byte aligned smem rows: (32+8)*2=80, 80%16=0
 #define BP_PAD 16      // smem_Bp row padding: stride 144 is 16-byte aligned
 #define GROUP_SIZE 16
+#define PADB 2         // smem_B_bf16 row 32->34 bf16 = 17 words, gcd(17,32)=1 => conflict-free (was 16-way)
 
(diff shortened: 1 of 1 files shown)


AzeezIsh:
Folded directly into #353 instead of a stacked PR — the cold-TTFT commit (eabfa8f) is now on strix/consolidate-tmp itself, so it lands with the rest of the consolidated strix work in one merge. See the win-summary comment on #353.

PR #356, opened 2026-07-23
perf(prefill): fix smem bank conflicts in FFN + projections GEMMs, K=4 verify dispatch

tbraun96:
## Summary

Three performance fixes for the prefill and K=4 verify paths on qwen3.6-27b NVFP4.

### 1. FFN GEMM bank conflict fix (77faec3b)
- Padded `smem_B_bf16` row stride 32→34 bf16 in `w4a16_gemm_t_m128_bf16_v2`
- Eliminates 16-way bank conflicts (80% of shared-memory wavefronts were excessive)
- **Result: −40% prefill time (~198 s/run)**, bit-identical
- ncu: L1/TEX 92%→56%, tensor pipe 24.5%→43.2%, conflicts 1025M→111M

### 2. Projections GEMM bank conflict fix (37d096ca)
- Added +4B pad to `smem_B_fp8` in `w4a16_gemm_t_m128` (qkv, o_proj)
- **Result: −7.4% to −10.5%** across projection shapes, bit-identical
- Smem budget maintained at 3 CTAs/SM

### 3. K=4 verify dispatch (8bc3f783)
- Added batched GEMV dispatch for `num_tokens==4` (NVFP4 + FP8)
- Previously fell through to 4 individual dense_gemv calls

### 4. Build default fix (53ca3e3a)
- `ATLAS_TARGET_MODEL` now defaults to `*` instead of `qwen3-next-80b-a3b`
- Fixes missing-kernel errors when serving other models without the env var

## Validation

- **Bit-identity**: PASS (all shapes, diff=0)
- **Standalone microbenchmark**: −40% FFN, −7.4 to −10.5% projections
- **Sized impact**: ~205 s/run total (5023.41 s → ~4818 s, −4.1%)
- **Gates**: A/B/C pending (running in parallel on DGX cluster)

## Testing

```bash
# Standalone microbenchmark (HL6 campaign)
cd /workspace/campaign2_artifacts/HL6
./harness_m64 20 bit    # bit-identity
./harness_m64 30 bench  # speed

# Build
ATLAS_TARGET_MODEL=qwen3.6-27b cargo build --release -p spark-server
```

PR #287, opened 2026-07-10
fix(ssm): recency-only snapshot eviction — hit-weighted score pinned fossil anchors (warm TTFT 6.1s→1.6s avg, llama parity)

AzeezIsh:
# fix(ssm): recency-only snapshot eviction — hit-weighted score pinned fossil anchors

## Problem

On multi-turn agentic workloads, warm-restore TTFT degrades catastrophically with conversation depth: the SSM restore anchor freezes on an early-conversation snapshot while the tip grows, replaying up to **18.6k tokens through all GDN layers per turn** (≈40 s TTFT tail). Measured on unsloth/Qwen3.6-27B-NVFP4, GB10, 174-turn deep agentic bench, seed 42:

| build / config | TTFT med | avg | p90 | max | max SSM replay |
|---|---|---|---|---|---|
| llama.cpp reference | 1373 | 1592 | 2223 | 9775 | ~0 |
| main-438ed09, int32/sl128 | 1661 | 6292 | 24769 | 40967 | 18640 |
| main-438ed09, int0/sl128 | 1616 | 6123 | 24545 | 40528 | 18640 |

Identical `max_recompute` across checkpoint-interval configs ruled out snapshot *placement*; instrumented probes (`ATLAS_SNAP_LOOKUP_DBG=1`) showed `matched_tokens` lands within 16 tokens of the previous prompt tip on **every** warm turn — the anchor exists, the index just evicts it.

## Root cause

Two compounding defects in `SsmSnapshotIndex` (`radix_tree/snapshot.rs`):

1. **`lookup()` bumped every improving candidate** in its scan (recency + hit count), not just the winner. Deep lookups walk the improving chain through shallow early-prefix entries, keeping them eternally fresh.
2. **`evict_lru()` scored entries `last_access × (1 + hit_count)`** — a monotonic timestamp multiplied by hit count. Any once-hit old entry outscores every fresh save, so once cold entries drain, each new save's eviction victim is the *previous fresh save*: the tail checkpoint written during this turn's prefill is evicted by the same turn's finish-leaf save, ~25 ms before the next turn's lookup needs it. The stale anchor keeps getting selected → keeps getting hit-bumped → **absorbing frozen-anchor state** (observed: anchor pinned at token 9056 for 29 consecutive turns, replay growing 16 → 12672 tokens).

## Fix

- `lookup()`: side-effect-free scan; bump recency **once, on the winner only**.
- `evict_lru()`: pure LRU on `last_access`; drop the now write-only `hit_count` field.

[...]

PR #369, opened 2026-07-24
perf(mlperf-edge): fold the last lever, and make the GB10 K=4 submission replicable

tbraun96:
Closes out the GB10 dense-27B edge-agentic campaign: the last un-landed lever, the replication
guide for the submission, and the repo hygiene that #366 could not do while it was still being
measured.

`centml/Qwen3.6-27B-NVFP4-W4A4-mlpinf` on a DGX Spark GB10, MLCommons edge-agentic harness
(1007 perf + 995 BFCL, temp 0.0 / seed 42).

## The conglomerate

#366 landed the decode win. This PR is the rest of it, so that the submission is one coherent,
reproducible thing rather than a branch plus tribal knowledge.

| where | what |
|---|---|
| **main (#366)** | chain-widened K=4 verify: the missing `n==4` FFN arm, `w4a16_gemv_batch8`, M≤8 gate widening, `wy5`–`wy8` GDN verify, shadow top-k instrumentation |
| **this PR** | SSM snapshot-lookup fix · `REPLICATE.md` · one runnable reproduce entry point · campaign scratch out of the repo root |

### Result being replicated

| metric | main before K=4 | **K=4 submission** | confirmed vLLM | vs before | vs vLLM |
|---|---|---|---|---|---|
| perf wall (1007) | 4551.9 s | **4104.0 s** | 5361 s | **−9.8%** | **−23.4%** |
| TPS | 17.56 | **19.20** | 14.6 | **+9.3%** | **+31.5%** |
| TPOT median | 38.18 ms | **32.60 ms** | 31.39 (tuned) | **−14.6%** | +3.9% |
| TTFT median | 1264 ms | 1298 ms | 2985 ms | +2.7% | **−56.5%** |
| BFCL | ~87 | **87.44** | 86.43 | +0.4 | **+1.0** |
| IoU (perf) | 0.6285 | 0.6231 | 0.6269 | −0.005 | −0.004 |

1007/1007 and 995/995 completed, 0 failed. Both IoU deltas are inside the harness MDE
(0.022–0.042 at N=1) — those are ties, not losses, and should not be reported either way.

Against the vLLM run we can actually reproduce end to end, Atlas wins wall, throughput and BFCL
and ties IoU. The one axis genuinely still open is raw TPOT against a *tuned* vLLM (Marlin +
CUDA graphs): 32.60 vs 31.39 ms. The campaign established why — GB10 single-stream decode on an
all-NVFP4 checkpoint is weight-bandwidth roofline-bound, and every weight/kernel lever was
measured dead. See the ledger.

## `REPLICATE.md`

The point of this PR. `docs/campaigns/gb10-decode-fold-2026-07/REPLICATE.md` covers the build
(including why `ATLAS_TARGET_MODEL` is not optional — omit it and you get a different kernel set
and an md5-identical binary, so you measure nothing), every frozen serve flag and env var with
the reason it is pinned, the gate order, and the traps:

- `--num-drafts N` gives verify width **K = N+1**. This off-by-one has mis-labelled sweeps before.
- Several `ATLAS_*` flags are presence-checked (`is_some`/`is_none`), so setting one to `0` still
  turns it **on**. `ATLAS_NO_FFN_NVFP4_MMQ` and `ATLAS_BF16_TC_PREFILL` are in that group.
- `--gpu-memory-utilization 0.70` is a ceiling, not a suggestion — GB10's memory is unified and
  the box OOM-freezes rather than failing cleanly above it.
- `--max-batch-size 1` because every speculative path is gated on `active.len() == 1`. A larger
  batch silently disables the thing being measured.
- Wall time is ~456 s client-side across 1007 samples, so it is only comparable between runs on
  an otherwise-idle box.

## The snapshot fix — folded, and explicitly not claimed as a win

`fix(ssm): stop warm-turn snapshot misses that make TTFT climb on short prompts` reuses this
turn's anchor on the next turn instead of letting only turn 1's anchor survive, so warm replay
stops growing (16 → 336 → 880 tokens).

It came over from Strix, where it was a large win, and it was expected to close TTFT p99 — the
one latency axis a tuned vLLM still beats us on. **It does not.** Same-box A/B, sequential legs
(unified memory cannot host two serves at util 0.70), 16-turn growing conversation, frozen env,
only the binary differs:

| leg | TTFT p50 | TTFT p90 | TTFT max | TPOT p50 | output sha |
|---|---|---|---|---|---|
| control | 906.0 ms | 1049.6 ms | 1148.2 ms | 43.59 ms | `b5d6a89aab5528ea` |
| + fix | 906.5 ms | 1049.3 ms | 1144.5 ms | 43.62 ms | `b5d6a89aab5528ea` |

The control leg has no warm-TTFT climb to fix — flat 850–1150 ms across turns 2–15 while context
grows 11.7k → 25.7k chars. The GB10 golden config already carries the mitigations the Strix
branch lacked (`ATLAS_SSM_TAIL_PROTECT` + lease TTL 128, 128 SSM slots, checkpoint interval 32,
plus #287 recency eviction and the #345 lease).

Kept anyway: output-neutral (identical greedy hash over ~3.4k generated tokens), free, and a real
fix for configs without those mitigations. Writeup and the limits of that probe:
`docs/campaigns/gb10-decode-fold-2026-07/SNAPSHOT_FIX_AB.md`.

## Levers, with verdicts

Folded (in main via #366): `n==4` FFN arm · `w4a16_gemv_batch8` · M≤8 gate widening ·
`wy5`–`wy8` GDN verify. Folded here: snapshot-lookup fix (neutral).

Measured dead — recorded so nobody re-runs them:

| lever | verdict |
|---|---|
| **fp8-KV** | Neutral-to-worse at e2e on a same-binary control (TPOT +2.4%, TPS −2.7%, IoU 0.6285→0.6223). An earlier "−9% wall" claim was a stale-baseline error, retracted in the ledger. |
| **W4A8 int8-act decode GEMV** (the Strix +25% trick) | 0.99–1.01× on GB10 plus ~0.5% accuracy cost. Decode is weight-once and roofline-bound; activations are <1% of traffic at M≤8. |
| **Native FP4-MMA verify** | M≤8 is far too small for the tiles. |
| **NVFP4-ing the GDN weights** | Inert — this checkpoint's GDN is already NVFP4. |
| **Tree drafting** | Chains beat trees by 17% on measured acceptance. `speculative/tree_shape.rs` stays as the Phase-0 evidence; nothing dispatches to it. |
| **M_TILE=64 prefill default** | Strix-only. The GB10 analog is `cfg!(atlas_hip)`-gated and a no-op here, and GB10's `w4a16_gemm_t` is the lossy FP8 E4M3 path already rejected on quality. |
| **Prefill FFN smem bank fix** | Already in main via #356 — the SHA-ancestry check that flagged it as missing was a false negative against a squash-merge. |

Still open, with the reason: TPOT tail dispersion (variable acceptance, inherent to fixed-K —
needs adaptive-K or a higher-acceptance drafter) · TTFT p99 (evidence points at SSM snapshot
*eviction* under multi-session pressure, which a single-conversation probe structurally cannot
reproduce) · W4A4 **prefill** (untested; prefill is compute-bound at M = seqlen ≫ 8, the one
regime where native sm_121a FP4 MMA could pay).

## Repo hygiene

#366 landed ~30 scratch files and four result directories into the repo root. They now live in
`scripts/mlperf-edge/` (harness, probes, gates) and `docs/campaigns/gb10-decode-fold-2026-07/`
(log, ledger, findings, `raw/` run data).

`scripts/mlperf-edge/run_golden_e2e.sh` is the single reproduce entry point. It used to hardcode
`/workspace/.wt-decode-fold` — a worktree that existed on one box on one day — and write its
generated config back into that same path, so nobody but its author could run it. It now takes
`ATLAS_BIN` / `HARNESS_DIR` / `BASE_CONFIG`, validates them, defaults `ND` to the selected winner,
and runs the Gate C2 coherence + tool-call smoke *before* committing to the multi-hour benchmark.

```bash
ATLAS_BIN=$PWD/target/release/spark \
HARNESS_DIR=/workspace/endpoints-fresh \
BASE_CONFIG=/workspace/endpoints-fresh/results/defaults_20260721_173342/config.yaml \
  ND=3 bash scripts/mlperf-edge/run_golden_e2e.sh
```

## Verification on this branch tip

Full both-phase e2e on this exact tip, dgx1, frozen config, ND=3 / K=4:
`results/conglomK4_e2e_20260724_185106`. 1007/1007 and 995/995 completed, 0 failed,
995/995 scored with 0 empty and 0 missing. Gate C2 passed before the benchmark started.

Against the reference run of the same code path and config (`chainK_golden_e2e_20260724_131209`):

| metric | this tip | reference | delta |
|---|---|---|---|
| perf wall (1007) | 4134.01 s | 4104.00 s | +0.7% |
| TPS | 19.25 | 19.20 | +0.3% |
| TPOT p50 | 32.62 ms | 32.60 ms | +0.1% |
| TPOT p90 | 38.30 ms | 38.53 ms | −0.6% |
| TPOT p95 | 40.20 ms | 40.57 ms | −0.9% |
| TPOT p99 | 43.73 ms | 43.77 ms | −0.1% |
| TTFT p50 | 1301.63 ms | 1298.23 ms | +0.3% |
| TTFT p90 | 2544.94 ms | 2538.87 ms | +0.2% |
| TTFT p99 | 5415.62 ms | 5353.74 ms | +1.2% |
| BFCL | 87.24 | 87.44 | −0.20 |
| IoU | 0.6279 | 0.6231 | +0.0048 |

Everything is inside run-to-run noise, which is the expected result: the only decode-path delta
is the snapshot fix, and that was already measured neutral. This run is the confirmation that
folding it plus rebasing onto #366 changed nothing, not a claim of further gain.

**One thing worth taking from it.** These two runs are the same binary, same config, same box,
same seed — and their IoU differs by **0.0048**. The "IoU deficit vs vLLM" reported earlier was
0.0038, i.e. *smaller than the noise floor between two identical runs*. That comparison is
undecidable at N=1 and should not be reported as a loss in either direction. Same for BFCL,
where the two runs differ by 0.20 while both sit ~0.8–1.0 above vLLM's 86.43.

Scorecard against the vLLM run we can reproduce end to end (5361 s / 14.6 tps / 86.43 / 0.6269):
wall **−22.9%**, TPS **+31.8%**, BFCL **+0.81**, IoU tie, TTFT p50 **−56%**. Still behind on raw
TPOT p50 against a *tuned* vLLM (32.62 vs 31.39 ms) — roofline, not a missing kernel.

A second leg at ND=4 / K=5 is running on dgx2 to settle the open K=5 question; it is a different
box with other tenants, so its wall number will be treated as indicative and only its
TPOT/BFCL/IoU compared.


tbraun96:
## K=5 settled at full e2e — K=4 stays the submission pick

The handoff left K=5 (`--num-drafts 4`) open: on the 174-turn agentic subset it was
wall/tps/TTFT-optimal within noise, so it was worth asking whether it beats K=4 over the full
1007. It does not, on the metric that matters, and the subset ladder's prediction holds at scale.

Both legs on this branch tip, frozen config, `--num-drafts` the only difference.

| metric | **K=4** (dgx1) | K=5 (dgx2) | |
|---|---|---|---|
| TPOT p50 | **32.62 ms** | 33.71 ms | K=4 −3.3% |
| TPOT p90 | **38.30 ms** | 40.66 ms | K=4 −6.2% |
| TPOT p95 | **40.20 ms** | 42.13 ms | K=4 −4.8% |
| TPOT p99 | **43.73 ms** | 46.68 ms | K=4 −6.7% |
| TTFT p50 | 1301.63 ms | **1214.50 ms** | K=5 −6.7% |
| TTFT p90 | 2544.94 ms | **2474.16 ms** | K=5 −2.8% |
| TTFT p99 | **5415.62 ms** | 5510.76 ms | K=4 −1.7% |
| perf wall | 4134.01 s | 4070.00 s | K=5 −1.5% |
| TPS | **19.25** | 19.14 | K=4 |
| BFCL / IoU | **87.24 / 0.6279** | not scored | — |

Both 1007/1007, 0 failed.

**K=4 wins TPOT at every percentile, and the margin widens down-tail.** K=5 buys TTFT p50/p90 and
a nominal 1.5% of wall. That is the same trade the subset ladder showed (K=4 TPOT-optimal, K=5
wall/TTFT-optimal), now confirmed on the full workload rather than a 174-turn slice.

Two caveats, stated rather than buried:

1. **The wall comparison is cross-box.** dgx2 also hosts a reranker/embedding/qdrant stack from
   another session, and roughly 456 s of the 1007-sample wall is client-side. Treat K=5's 1.5%
   wall edge as indicative, not measured. The TPOT gap is more robust — it is server-side
   per-token latency, it is consistent across all four percentiles, and it matches the ladder.
2. **K=5 has no accuracy validation.** The BFCL scorer dependency is missing in the dgx2 harness
   venv (`ImportError: bfcl-eval is required`), so its accuracy phase crashed at finalize. The
   perf phase completed cleanly; the accuracy phase produced nothing.

So K=4 keeps the submission: it wins per-token latency at every percentile *and* it is the only
one of the two with a full BFCL + IoU result behind it. Raw reports for both legs are in
`docs/campaigns/gb10-decode-fold-2026-07/raw/`.

Issue #153, opened 2026-06-10
Auto-calibrate fp8 KV scales at startup (no flag), inside a pre-traffic warmup pass


AzeezIsh:
## Problem

Serving --kv-cache-dtype fp8 on a checkpoint that ships no k_scale/v_scale tensors (all our Qwen3.6 checkpoints) currently requires the operator to know about and pass --fp8-kv-calibration-tokens, or Atlas refuses to serve. Two further issues with the current online calibration (crates/spark-model/src/layers/fp8_calibration.rs):

1. It calibrates on the FIRST REAL REQUEST. During the warmup window fp8 KV writes use scale=1.0, so the opening turns of request #1 run on unconverged scales. Measured effect: across healthy webserver_ok tiers, run 1 fails ~11% vs ~5% for all later run indices (n=27/152). CUDA-graph capture landing on the first request contributes to the same cold-start cocktail.
2. The token count (256) looks like a magic number with no provenance recorded.

## Ecosystem context (researched 2026-06-10)

- vLLM ships three modes: default = NO calibration (scales 1.0, accepts clipping risk); online 'random token calibration' via calculate_kv_scales=True that estimates scales from a single batch of random tokens during warmup then freezes; and the recommended offline path via llm-compressor which bakes scales into the checkpoint using ~512 calibration SAMPLES of a real dataset.
- There is no industry-standard online token count; our 256 is an Atlas-validated value (2026-06-06 gates: calibrated fp8 KV 9/10 vs bf16 10/10 within N=10 noise).

## Proposal

1. When --kv-cache-dtype fp8 is requested and the checkpoint ships no k/v scales, enable online calibration AUTOMATICALLY instead of refusing to serve. Checkpoint-shipped scales, when present, keep taking precedence. Keep --fp8-kv-calibration-tokens as an override; 0 disables (current refuse behavior for operators who want offline-calibrated checkpoints only).
2. Run the calibration inside a server-side warmup pass BEFORE the listener accepts traffic: prefill a built-in (or --warmup-prompt) text long enough to cover the calibration window, freeze scales, then discard the sequence. The same pass should capture decode CUDA graphs. Result: request #1 is statistically identical to request #1000, and run-position effects disappear from benchmarks.
[...]

PR #372, opened 2026-07-26
fix(fp8-kv): freeze calibration scale on first observe + detect checkpoint k/v scales

rsafier:
Extracted from the `wip-laguna-lora` working branch as an independent, model-agnostic fp8-KV correctness pair (cherry-picks cleanly onto `main`).

**1. Freeze calibration scale on first observe** (`3fe8c5f3`)
Online fp8-KV calibration warmed up with a headroom scale (2.0) for `warmup_tokens`, then froze to the observed scale (~0.3) — so KV written during warmup used a different scale than reads after the freeze. Under prefix caching this corrupted reused blocks (`TheThe` loops / empty output). Fix: freeze the scale on the FIRST observe (write-scale == read-scale from token 0); removed the vestigial `warmup_tokens` field. Kernel proven correct via HDIM=128 microtest — this was a scale-lifecycle bug, not a kernel bug.

**2. Detect checkpoint k/v scales** (`6a9b8d03`)
Models that ship per-layer `k_scale`/`v_scale` don't need calibration at all. Count them (`fp8_kv_scale_count`) and branch the boot log: has scales → INFO "using N per-layer scales, no calibration needed"; missing → WARN. Stops a spurious calibration-warning on checkpoints that already carry scales.

Draft: opening for CI + early eyes; part of the staged upstreaming of the Laguna work (see the stack plan). No Laguna-specific code — applies to any fp8-KV deployment.


tbraun96:
**Keeping this open — I had it filed as "needs tests + a split" and that undersold it.** This is a live data-corruption bug on `main`, and there is an open maintainer-authored issue behind it that the earlier triage missed.

### The bug is still on main

`crates/spark-model/src/layers/fp8_calibration.rs:174` today:

```rust
if inner.tokens_seen >= self.warmup_tokens && !inner.frozen {
    inner.k_scale = (inner.k_running_max / FP8_E4M3_MAX).max(MIN_SCALE);
```

`effective_fp8_scales()` (`qwen3_attention/decode.rs:23–29`) returns the **live** scale, and it is used for both the write (`decode/write_kv_cache.rs:487`) and the read. So KV persisted before the freeze is quantized at `k_scale = 2.0` and dequantized after it at `observed_max/448` (~0.3) — a **~6× error on every pre-freeze entry**, including prefix-cache blocks shared across sequences.

`fp8_kv_scale_count` does not exist anywhere in the tree, and main has no alternative mitigation: `fp8_kv_calibration_tokens` is read elsewhere only to re-enable CUDA graphs (`decode_a.rs:163–172`, `verify_b.rs:199–207`), never to gate a write.

Reachability is ordinary chat, not a corner. Six shipped `MODEL.toml` files set `fp8_kv_calibration_tokens = 256` (`gb10/qwen3.6-35b-a3b`, `strix/qwen3.6-35b-a3b`, `gemma-4-26b-a4b`, `mistral-small-4`, `deepseek-v4-flash`, `step3p7-flash`) plus one recipe fixture. **Any first prefill chunk shorter than 256 tokens writes at the wrong scale.**

### It connects to issue #153

**#153** — *"Auto-calibrate fp8 KV scales at startup (no flag), inside a pre-traffic warmup pass"* — is the maintainer's own record of this exact problem, with a measured production effect: *"run 1 fails ~11% vs ~5% for all later run indices (n=27/152)."* Your commit 2 (detect checkpoint k/v scales) **is proposal item 1 of that issue**. Please link them.

### It's not dead — it rebases and compiles

One conflict, and it's mechanical: main split `serve.rs` → `serve_load.rs`, moving the `resolve_kv_cache_config` call to `serve_load.rs:401`, where `store` is already in scope (`:293`). With that call site moved, `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo check -p spark-server -p spark-model -p spark-runtime` is clean.

### The test writes itself, and needs no GPU and no mocking argument

`MockGpuBackend` (`spark-runtime/src/gpu/mock.rs:117`) is `pub`, already used from spark-model (`model/ssm_snapshot_spill_tests.rs:10`), and implements everything `observe()` touches — `alloc`, `copy_h2d`, `copy_d2h`, `memset_async`, `synchronize`, `kernel()`.

Construct with `warmup_tokens = 256`; `observe()` a 10-token batch; snapshot `scales()`; `observe()` past the 256 boundary; assert `scales()` unchanged. On main that goes `2.0 → 1e-12` and **fails**; with your fix it passes. ~15 lines. That is a genuine fails-without-the-fix test, which is what the new merge rigor asks for.

### Four things before merge

1. **`ATLAS_FP8_KV_HEADROOM` with an implicit `unwrap_or(2.0)` in the production path.** That's a PCND violation, and it's against the standing "no env-var soup — knobs become defaults (opt-out) or CLI flags" rule. Make it a CLI flag or a `MODEL.toml` key.
2. **Freezing on the *first* observe** derives the scale from possibly a 5-token prompt — trading a catastrophic mismatch for a real **clipping** risk. #153's pre-traffic warmup pass fixes both. Land this explicitly as the first half of #153, not as the final answer.
3. **Split it.** Commit 2 is pure logging and shares nothing with commit 1.
4. **A second-order bug this misses.** With the scale frozen at token ~1, `decode_a.rs:167` and `verify_b.rs:202` still suppress CUDA graphs until `seq_len > fp8_kv_calibration_tokens + 10` — so every process runs ~266 tokens eagerly for a calibration that finished immediately. That gate should key off the actual `frozen` flag.

Happy to write the test and do the rebase myself if you'd rather I push to this branch — say the word and I'll do it here rather than opening a parallel PR.

rsafier:
@tbraun96 Accepted on all four: headroom moves out of the env var into a CLI flag / MODEL.toml key, the logging commit gets split out, the graph-suppression gate keys off `frozen`, and I'll add the MockGpuBackend test you sketched (fails on main, `2.0 → 1e-12`).

Linking #153 as requested — commit 2 here is proposal item 1 of that issue, and I'm treating this PR as the explicit first half of #153, not the final answer on scale derivation (your clipping-risk point on first-observe freeze is taken).

Sync of the branch onto current main (call site moved to `serve_load.rs`) is coming with the rework.

rsafier:
All four items landed on the branch (synced with main, no force-push):

1. **Env var gone** — `--fp8-kv-headroom` (default 2.0), threaded `ServeArgs → ModelConfig → Fp8KvCalibration`; `< 1.0` rejected at startup with the why ("clips the very values it was measured from") and clamped in the constructor as defense in depth.
2. **Explicitly the first half of #153** — stated in the commit; the clipping-risk trade on first-observe freeze stands until #153's pre-traffic warmup pass replaces the derivation.
3. **Split done** — the checkpoint-scale detection (old commit 2) is reverted out of this branch; it'll come back as its own PR against #153 (it is proposal item 1 there).
4. **Both graph gates key off `frozen`** — new `TransformerLayer::fp8_calibration_frozen` probe; `decode_a.rs` and `verify_b.rs` unsuppress the moment calibration actually froze instead of idling eager for ~266 tokens. `verify_b` also gains the `calibration_tokens > 0` guard `decode_a` already had — its old `seq_len > 0 + 10` form would undo EP/profile suppression after ten tokens.

Your test, verbatim shape: `scale_frozen_on_first_observe_never_moves` (MockGpuBackend, 10-token observe → snapshot → 30 observes across the old 256 boundary → scales unchanged). It fails on the pre-fix design (2.0 → MIN_SCALE at the boundary) and passes here. fmt/clippy `-D warnings`/2122 tests green locally. Un-drafted.

rsafier:
The split-out commit 2 is now its own PR: #439 (checkpoint k/v-scale detection + guidance, #153 proposal item 1). This PR stays calibration-fix-only.

tbraun96:
Landed in main via the #440 stack merge (`3c941751`). Verified by content diff — `--fp8-kv-headroom` is threaded end-to-end (`cli/serve_args.rs:646`, `atlas-core/src/config.rs:355`, `config/factory.rs:102` = 2.0, `cli/validate.rs:129` ≥1.0 guard, `serve_load.rs:290`, `qwen3_attention/init.rs:624`), and the graph-unsuppress gate at `trait_impl/decode_a.rs:185` is now `&& self.fp8_calibration_frozen()` rather than the old `seq_len > tokens+10` token count.

One structural note: your new `trait_impl/fp8_gate.rs` was not created — the identical `fp8_calibration_frozen()` lives inline at `decode_a.rs:38`, and the 500-LoC cap was satisfied by hoisting `decode_forward_body` into `decode_a3.rs` instead. Functionally complete either way. Thanks @rsafier.

PR #373, opened 2026-07-26
fix(kv): prefix-cache refcount + block aliasing (CUDA-700, pool leak, HSS mis-filing)

rsafier:
Under `--enable-prefix-caching`, long agentic sessions could **kill the CUDA context** and — once that was fixed — still **permanently wedge the KV block pool**. Two commits, two distinct refcount bugs. They are stacked deliberately: the second bug is only *reachable* once the first is fixed.

## 1. `fix(kv): stop prefix-cache exhaustion crashing CUDA (700)`

**a. The CUDA-700.** A cached block carries the prefix cache's `+1` **on top of** any live sequence's `block_table` ref. `return_evicted_block` force-**zeroed** the refcount and returned the block to the free list, assuming radix-ref == KV-ref 1:1. Once those desynced, evicting a radix node force-freed a block **still live in another sequence**; `alloc_block` re-handed it out and `zero_block` memset it under a concurrent decode → aliased KV pointer → `CUDA_ERROR_ILLEGAL_ADDRESS`, surfaced at the next sync (`copy_logits_to_host` / `free_sequence`). Unrecoverable: the whole context dies.
**Fix:** decrement (release exactly the cache's one ref); never free a still-referenced block.

**b. Failed-prefill radix leak.** A warm prefill that matched a prefix (bumping radix refs) then failed to allocate its suffix was torn down via `free_sequence`, which releases the prefix cache over `seq.tokens` — but `seq.tokens` is only populated in a later finalize phase, so on the failure path it is **empty** and the matched nodes are never released, pinned at ref≥2 forever.
**Fix:** stash the matched prefix (`seq.prefix_ref_tokens`) at lookup and release over it when `seq.tokens` is too short to cover it. Exactly one of the two sources covers the matched nodes → released once, never twice.

## 2. `fix(kv): stop prefix-cache re-inc leaking blocks until the pool wedges`

With eviction now correctly ref-counted, a **pre-existing** leak that the force-zero had been masking became visible: the pool wedged permanently — small requests still fit, but any large prefill failed `no free blocks` **forever**, restart-only recovery.

**Root cause — a 1:N inc/dec imbalance.** `cache_sequence` bumps the KV refcount of every block in `block_table[..num_cached_blocks]` so the cache "owns" a reference. But on a **warm** turn the leading blocks came *from* the cache, which already holds its `+1` — and eviction returns exactly **one** ref per radix node (`return_evicted_block`, one call per node, ever):

```
turn 1: alloc 1 -> cache_sequence 2 -> free 1     (balanced)
turn 2: warm hit 2 -> cache_sequence 3 -> free 2  (+1 leaked)
evict:  2 -> 1, never reaches 0, node gone        (block unreclaimable)
```

Every reused prefix block was pinned for the process lifetime, so a session leaked its whole cached prefix per turn. This is why the symptom changed from a CUDA-700 **crash** to a graceful permanent **wedge** — commit 1 is correct; it exposed the real bug underneath.

**Fix:** track `SequenceState.cached_prefix_blocks` (= `matched_blocks.len()`, set at all three prefix-lookup sites) and skip those leading entries in `cache_sequence`'s inc_ref loop — the cache already owns them. Matched nodes are guaranteed to still hold their cache ref there: lookup's inc_refs put them at ref≥2, and eviction only takes nodes at ref≤1.

**Also in this commit:**
- **Leak detector** (`block_mgmt::apply_evicted_blocks`): log when an eviction reclaims fewer blocks than it evicted — the "refs no future eviction can release" signature — so this class of bug is observable instead of silent. Enable with `RUST_LOG=info,spark_model::model::block_mgmt=debug`.
- **`dec_ref` hardening** (`paged_impl.rs`): its only guard was a `debug_assert`, and the workspace sets no `overflow-checks`, so a 0-ref decrement wrapped to `u32::MAX` in release = a silent permanent pin. Now saturating + `tracing::error!`.

## Validation (one GB10, Laguna-S-2.1-NVFP4, prefix caching ON)

- **Crash fix:** 30 concurrent 8.6K-token requests (≈258K tokens vs a 180K pool) → **30/30 HTTP 200**, swap engaged gracefully, **zero CUDA-700**, server responsive after.
- **Leak fix:** re-ran the exact agentic workload that produced the wedge (~78 requests, multi-turn conversations growing to 12K tokens) → agentic harness **3/3 tasks pass**, and the same **13,067-token prompt that had failed now succeeds in 7.9s**. Across the whole run: **0 KV exhaustions, 0 unreclaimed-evict events, 0 `dec_ref` over-releases**.
- Decisive leak-vs-capacity check: the identical large prompt failed on the wedged pool and succeeded on a fresh one — a leak, not capacity.
- **33/33 `kv_cache` unit tests**, incl. new `return_evicted_block_keeps_shared_block_alive`. `cargo check -p spark-model` + `-p spark-runtime` green on `main`.

Both fixes are minimal, localized and model-agnostic. **Prefix caching stays ON.**

### Known follow-ups (not in this PR, surfaced by the root-cause pass)
- `radix_tree/inner.rs` returns `partial_suffix` (the sub-block tail) to `freed_phys` on eviction although the cache never inc_ref'd it (`num_cached_blocks` counts full blocks only) — an over-release. Left out so this PR is a clean A/B; the new detector + saturating `dec_ref` will surface it.
- Radix nodes are inserted at **prefill** time but the cache's KV ref is taken at **finish** time, so a sequence dying in between leaves nodes referencing blocks the cache never ref'd. Structural fix: have `insert*` return newly-owned blocks and inc_ref at the call sites.
- Chunk-errored **streaming** requests never send `StreamEvent::Error`, so `atlas_requests_active` sticks (HTTP-accounting leak, not a block leak).

---

# Update: commit 3 — `fix(prefix-cache): tie the cache's KV ref to the node, not the sequence`

Validating the preempt work in #375 on a starved pool surfaced a **third** bug in this same area, and it **supersedes fix 2 above** — the `cached_prefix_blocks` skip was compensating for one direction of a deeper mismatch. That code is removed in this commit.

## The bug

The cache holds one KV ref per radix node, but took it against the **finishing sequence's `block_table`**, assuming node *i* holds `block_table[i]`. It does not: when a node already exists for a chunk, `insert` keeps that node's original `block_idx`. A sequence that did not get its block from the cache therefore has a **different** block at that position — it never matched, or (the case that reproduces) it was **restored from a swap file** with a freshly allocated table. The ref landed on a block no node referenced, while the node's own block carried none.

Evicting such a node then decrements a ref belonging to a **live sequence** — the same aliasing failure as bug 1a, reached by a different route: block returned to the free list while in use, handed to a second and third owner, interleaved KV writes, underflow on the second teardown.

Per-block history from `ATLAS_KV_TRACE=1` (added in this commit), block 1808:

```
try_alloc->1    @block_mgmt:376   allocated to a live sequence (count 1 = no cache ref)
evict_return->0 @block_mgmt:59    cache eviction stole that ref -> freed while in use
alloc->1        @block_mgmt:382   handed to a second owner
evict_return->0 @block_mgmt:59    evicted again
alloc->1        @block_mgmt:382   handed to a third
dec->0          @sequence:179     first owner frees; the next free underflows
```

## The fix

`insert` reports the blocks of the nodes it **created** (`InsertAcquired`); every call site `inc_ref`s exactly those. Reference lifetime == node lifetime by construction, so neither direction of the mismatch is expressible.

Plus three supporting changes:

1. **`return_evicted_block` no longer pushes a block already at 0 refs** — it is already on the free list, so the second push duplicates the entry and two later `alloc_block`s hand out the same block. Only a real 1→0 transition frees.
2. **Swap-in can reclaim from the cache** (`Model::reclaim_prefix_blocks`). Swap-in gates on `num_free_blocks()` before restoring and nothing on that path evicts, so once cached blocks are correctly pinned a swapped-out sequence waits on reclaimable-but-not-free capacity. Prefill/decode reclaim implicitly; swap-in has to ask.
3. **`ATLAS_KV_TRACE=1`** — per-block ring of the last 24 refcount events, each tagged with the caller's `file:line` via `#[track_caller]`, dumped on underflow. Refcount bugs are silent where they happen and only surface at a later free. Inert by default.

## Validation

Starved pool, **two request waves** so swap-out engages — swap-out is required to reproduce, which is why single-wave exhaustion tests look clean.

| | before | after |
|---|---|---|
| swap-outs | 5 | 8 |
| `dec_ref` underflows | **2** | **0** |
| bad evict-returns | 0 | 0 |
| decode batch-kills | 0 | 0 |
| coherence, cold + warm | — | **6/6** |

Coherence = checkable-answer prompts run twice (second pass warm through the cache), verified for the expected answer *and* absence of garbling, since aliasing corrupts output rather than crashing. Unit tests: spark-runtime 148 passed, `cargo fmt --check` clean.

## Reviewer notes

- Item 2 (swap-in reclaim) is **reasoned from the code path but not exercised end-to-end** — the repro never reached it, because the swap-in loop is gated on `active.len() < max_batch_size`, which never holds at saturation. It is guarded (returns 0 ⇒ no-op) and included because the ownership fix makes the starvation it addresses deterministic rather than occasional; shipping without it would knowingly make swap starvation worse.
- **Separate pre-existing pathology, not addressed:** under saturation the classic `--swap-space-gb` path degenerates — each swap-out writes ~556MB (362 blocks × 48 layers) to `/tmp`, giving a ~32s scheduler tick, and a request that cannot be admitted keeps triggering swap-outs without being admitted, starving decode.
- **Also unfixed:** `walk` returns `partial_suffix` blocks with no liveness check and the cache never refs them, so `evict` hands back a block it does not own. Narrow (needs an exact-prompt repeat), same shape.

---

# Update: commit 4 — `fix(hss): don't cache a slid window's blocks under the wrong token chunks`

A **content**-correctness sibling of the refcount bug above, on the `--high-speed-swap` path.

Every prefix-cache insert assumes `block_table` parallels the token stream from position 0 — the tree is keyed on tokens from the root and files `block_table[i]` on the node for chunk *i*. HSS breaks that: once the rolling window slides (`hss_window_start() > 0`), `block_table[0]` no longer holds position 0 (the front holds the most **recent** positions), so an insert files blocks under token chunks whose KV they do not contain, and a later warm hit reuses the **wrong KV** as if it were a valid prefix.

`cache_sequence` guarded on this, and so did `save_checkpoint`'s boundary insert — but the prefill-time inserts in `prefill_d` and `finalize_last` did not (11 sites). They are reachable whenever HSS is on, since `ensure_blocks_through_{prefill,decode}` slide as soon as `cache_blocks_per_seq` is hit. `radix_tree::inner::insert` already asserts the very invariant those sites violate:

```rust
debug_assert!(!hss_active || disk_block_ids.len() == block_table.len(), ...)
```

which a slid sequence fails by construction.

There is no correct partial insert — the tree indexes prefixes from the root and what survives in a slid window is a mid-sequence suffix — so all 11 sites now skip, via one documented predicate (`hss_window_slid`). Where a snapshot was just saved it is freed rather than registered: it is only reachable through the tree nodes being declined, so keeping it would leak a snapshot pool slot.

**Inert unless HSS is actually sliding.** With HSS off, `spark_storage::with_local` is a no-op, `disk_block_ids` stays empty and `hss_window_start()` is 0; with HSS on, inserts continue normally until the window first slides.

Note on scope: HSS is *not* affected by the classic-swap pathologies mentioned above — it never enters the whole-sequence swap-out/swap-in loop (that is gated on `--swap-space-gb`'s `KvSpillManager`) and reclaims incrementally via the sliding window, whose allocation path already has the `try_alloc → evict → retry` fallback. It was, however, affected by the aliasing bug **more reliably** than classic swap: pre-fix the cache's KV ref was taken only in `cache_sequence`, which is skipped for a slid sequence, so nodes created by the prefill-time inserts got **no** ref at all and were left pointing at blocks that teardown freed. Commit 3 fixes that by construction.

Follow-up (not done here): nodes already carry a `disk_block_id`, so out-of-window history could be cached as disk-backed entries rather than skipped. That is a feature, not this fix.


tbraun96:
Closing this as **merged elsewhere, not rejected** — and with an apology attached, because the attribution on it is wrong.

### Every line of this is in main

I verified by content, not ancestry (this repo squash-merges, so `merge-base --is-ancestor` is meaningless here):

```
$ git diff --quiet pr373 origin/main -- crates/spark-runtime/src/kv_cache/paged_impl.rs   # identical
$ git diff --quiet pr373 origin/main -- crates/spark-runtime/src/kv_cache/block_trace.rs  # identical
$ git diff --quiet pr373 origin/main -- crates/spark-runtime/src/radix_tree/inner.rs      # identical
```

A line-level sweep over the whole change-set: **275 unique added lines, 0 missing from main**. `return_evicted_block`'s decrement-not-force-zero fix and its "already at 0 refs ⇒ don't re-push" guard (`paged_impl.rs:240–285`), `dec_ref`'s saturating semantics + `tracing::error!` (`:184–209`), `InsertAcquired` (`radix_tree.rs:154`), `hss_window_slid` (`impl_a2.rs:67`), `reclaim_prefix_blocks` (`traits/model.rs:1102`, `scheduler/mod.rs:512,891`), `prefix_ref_tokens` (`traits.rs:149`), `ATLAS_KV_TRACE` (`block_trace.rs:44`) — all present.

Your regression test is on main and green:

```
$ cargo test -p spark-runtime --lib kv_cache
test kv_cache::tests::return_evicted_block_keeps_shared_block_alive ... ok
34 passed; 0 failed
```

### The attribution is wrong and that's on me

This content landed via **#388** (`744420904`), which I authored. Its body references neither this PR nor you, and the commit carries no `Co-authored-by: rsafier`. That was my error — you found the CUDA-700 and the pool leak, and the fix that shipped is yours. I can't amend a merged commit, so this comment and issue #421 are the record. I'll credit you on the porting work going forward and I've tightened my own process for it.

### Merging this now would regress main

Stated so nobody re-opens it later thinking it's free. `crates/spark-runtime/src/prefix_cache.rs` would lose `TierEvict`, `forget_snapshot_tier_key`, the `no_caching`/`tier_evict` module split and the `run_metrics` counter migration (back to three loose `AtomicU64` statics); `evict_snapshot_to_tier(min_tokens)` would revert to the un-gated form; and `file-size-cap.yml` would lose its pinned `actions/checkout` SHA, its `permissions: contents: read`, its `workflow_dispatch` trigger and 15 allow-list entries.

### The one thing that was about to be lost

Your body is the only written record of the classic-swap pathology you deliberately scoped out — ~556 MB per swap-out (362 blocks × 48 layers), ~32 s scheduler tick, and a request that repeatedly triggers swap-outs without ever being admitted. `swap_out_sequence` (`scheduler/lifecycle.rs:165`) is unchanged. Filed as **#421**, credited to you, with your figures flagged as needing re-measurement before anyone sizes a fix against them.

Thanks for this one — it was a real bug and the fix was right.

PR #375, opened 2026-07-26
fix(scheduler): preempt on KV exhaustion instead of failing (decode + prefill paths)

rsafier:
> **Draft: implemented and compiles, but NOT yet runtime-validated.** The preemption path has never been observed executing — I'm forcing the condition deliberately (tiny KV pool + concurrent long generations) before marking ready. Putting it up now for early review of the approach.

A sequence that can't extend its block table mid-decode currently errors **every** sequence in the batch:

```rust
Err(e) => {
    tracing::error!("decode_batch error: {e:#}");
    for mut a in active.drain(..) { send_error(model, &mut a, ...); }
    return;
}
```

So one sequence wanting a **single extra block** destroys up to `--max-num-seqs` in-flight requests.

## Why it happens

Admission control sizes the pool against a new request's **prompt** only (`scheduler/mod.rs`):

```rust
let blocks_needed = prompt_len / block_size + 1;
while model.num_free_blocks() < blocks_needed && !active.is_empty() { ...swap_out_sequence... }
```

…and is never re-consulted afterwards. But sequences **grow** one block per `block_size` decoded tokens. N conversations admitted comfortably at 2K tokens each can collectively exhaust the pool once they reach 18K.

**Measured** (GB10, Laguna-S-2.1-NVFP4, 9058-block pool, `--max-num-seqs 8`): five events with `free_blocks=0` at `abs=1173` blocks/seq — **8 × 1173 = 9384 blocks needed vs 9058 available**. Eight concurrent long agentic conversations physically don't fit, and nothing re-checks after admission.

```
ERROR decode_step: decode_batch error: alloc failed in ensure_blocks_through_decode:
  abs=1173 ws=0 bt_len=1173 cap=None free_blocks=0 slid=0 alloc'd=0: KV cache exhausted
```

Note Atlas **already has** preemption (`swap_out_sequence`, backed by `--swap-space-gb`) — it's simply never wired to the decode path.

## The fix

On a KV-exhaustion error: drop the largest sequence (`send_error` → `free_sequence` returns its blocks to the pool) and **retry the decode**, so the rest of the batch progresses. Loops until success or one sequence remains.

- **Victim policy mirrors admission** — largest `block_table`, grammar-active sequences excluded (their state isn't reconstructible).
- **Non-exhaustion errors keep the old fail-the-batch behavior** — this shouldn't silently swallow unrelated decode faults.
- `tokens` is rebuilt per attempt (derived from `active`; a stale copy misaligns the batch after preemption).
- `remove` not `swap_remove`, to preserve the ascending SSM-slot order the batched path needs — a hole only costs a fallback to the eager loop for that step, and the next step re-sorts.
- `n` is recomputed, since it gates the `n == 1` paths below.

## Scope / limitation

The victim is **errored**, not swapped out for later resume. Proper preemption via `swap_out_sequence` needs `KvSpillManager` threaded into `step_decode_only` (2 call sites) — worth doing, but killing 1 to save 7 is the strictly-better-than-today fix and is self-contained. Happy to extend to real swap-out if preferred.

Also worth noting: with a retry-capable client this is invisible today (our agentic harness passed 30/30 tasks *despite* 5 of these events, because the client retried) — behind a gateway that doesn't retry, it's 5 batches of hard failures.

---

# Update: validated on a starved pool, plus a prerequisite fix

## Decode preempt — validated

Forced KV exhaustion with a deliberately starved pool (gpu-util 0.63, ~2–3K blocks) and concurrent long generations:

```
WARN KV cache exhausted during decode: preempting slot=5 (368 blocks) so the other 7 sequence(s) can continue
```

`decode_batch error` = **0** across every run — the batch-kill this PR removes never fired again. Server stayed healthy; output coherence verified 6/6 (checkable-answer prompts, cold and warm through the prefix cache).

An earlier run of this test also emitted 10 `dec_ref on block N with 0 refs` in the same microsecond as the preempt. **That was not caused by this PR** — the blocks were already corrupt and the preempt was simply the first thing to free that table. Root-caused to a prefix-cache ownership bug, fixed in #373 (`InsertAcquired`); with that in place the same repro shows **0** underflows across 8 swap-outs. My initial hypothesis — that victim selection picked the sequence that had just failed — was wrong: `decode_batch`'s error path frees nothing, and `free_sequence` clears `block_table`.

## New commit: `fix(prefill): make the chunk-0 prefix lookup idempotent under preempt-retry`

The prefill arm of this PR was **unsafe as written**. `prefill_b_prefix_lookup` runs its acquisition whenever `chunk_start == 0`, and it runs **before** `ensure_blocks_through_prefill` — so the exhaustion the retry exists to handle happens with all of it already applied:

- `lookup` bumped a radix ref on every matched node,
- each matched block was `inc_ref`d **and pushed onto `seq.block_table`**,
- `cached_prefix_tokens` / `marconi_skip_to` were set.

Retrying the same chunk appended the matched blocks to `block_table` **a second time** — so `block_table[i]` no longer maps to logical block *i*, and KV for those positions is read/written at the wrong offsets — and took a second radix ref that the single `release` in `free_sequence` can never balance (the progressive-pin signature that wedges the pool, cf. #373).

A `prefix_lookup_applied` flag makes the re-entry a no-op that replays chunk 0's original `(skip_tokens, skip)`, so the retry proceeds straight to the allocation against the prefix it already holds.

This is a **latent** hazard, not an observed failure: in the repro, exhaustion lands on prefill **start**, not on a continuation chunk, so this retry loop never fired. It is a prerequisite for the prefill arm regardless.

## Known gap: prefill start still 500s

`phase_start_prefills` → `handle_prefill_start_error` preempts for SSM `"pool exhausted"` but falls through for `"KV cache exhausted"`, so a newcomer that cannot be admitted gets an HTTP 500 (observed: 2–3 streams per run).

I deliberately did **not** extend the preempt there. By that point the request has already failed, so killing a live sequence would destroy in-flight work and rescue nothing. The correct fix is to **defer/requeue** the request, which is a scheduler-structure change (the request is consumed by `prefill_request_chunked`, and the swap-out valve cannot help within one tick — requests arriving together see `active` empty). Worth a follow-up issue rather than bolting onto this PR.

### DIFF
--- a/crates/spark-model/src/engine/tests.rs
+++ b/crates/spark-model/src/engine/tests.rs
@@ -162,6 +162,8 @@ impl Model for MockModel {
             session_hash: 0,
             chunked_prefill_meta: None,
             cached_prefix_tokens: 0,
+            prefix_lookup_applied: false,
+            prefix_lookup_skip: false,
             kv_valid_tokens: 0,
             last_decode_ckpt_block: 0,
             prompt_len: 0,
--- a/crates/spark-model/src/model/nllb/model_impl.rs
+++ b/crates/spark-model/src/model/nllb/model_impl.rs
@@ -180,6 +180,8 @@ impl Model for NllbGpuModel {
             session_hash: 0,
             chunked_prefill_meta: None,
             cached_prefix_tokens: 0,
+            prefix_lookup_applied: false,
+            prefix_lookup_skip: false,
             kv_valid_tokens: 0,
             last_decode_ckpt_block: 0,
             prompt_len: 0,
--- a/crates/spark-model/src/model/trait_impl/meta.rs
+++ b/crates/spark-model/src/model/trait_impl/meta.rs
@@ -225,6 +225,8 @@ impl TransformerModel {
             session_hash: 0,
             chunked_prefill_meta: None,
             cached_prefix_tokens: 0,
+            prefix_lookup_applied: false,
+            prefix_lookup_skip: false,
             kv_valid_tokens: 0,
             last_decode_ckpt_block: 0,
             prompt_len: 0,
--- a/crates/spark-model/src/model/trait_impl/prefill_b/prefix_lookup.rs
+++ b/crates/spark-model/src/model/trait_impl/prefill_b/prefix_lookup.rs
@@ -23,6 +23,20 @@ impl TransformerModel {
         stream: u64,
     ) -> Result<(usize, bool)> {
         let bs = kv_cache.block_size();
+        // Retry re-entry (scheduler preempt-and-retry on KV exhaustion): chunk 0
+        // already acquired this sequence's prefix — it `inc_ref`d each matched
+        // block and pushed it onto `block_table` BEFORE the allocation that
+        // failed. Re-running would push those blocks a second time and take a
+        // second radix ref that nothing releases. Replay the original decision.
+        if chunk_start == 0 && seq.prefix_lookup_applied {
+            tracing::debug!(
+                "prefix lookup re-entered at chunk 0 (retry): replaying \
+                 skip_to={} skip={} without re-acquiring the cached prefix",
+                seq.marconi_skip_to,
+                seq.prefix_lookup_skip,
+            );
+            return Ok((seq.marconi_skip_to, seq.prefix_lookup_skip));
+        }
         if chunk_start == 0 {
             // Prompt-logprob collection needs a live hidden row for EVERY
             // position — a cache/Marconi skip would leave gaps. Force the
@@ -304,6 +318,8 @@ impl TransformerModel {
                 0
             };
             seq.marconi_skip_to = skip_tokens;
+            seq.prefix_lookup_skip = skip;
+            seq.prefix_lookup_applied = true;
             Ok((skip_tokens, skip))
         } else if seq.marconi_skip_to > 0 {
             // Chunk 1+: inherit skip info from chunk 0's prefix cache lookup.
--- a/crates/spark-model/src/traits.rs
+++ b/crates/spark-model/src/traits.rs
@@ -121,6 +121,22 @@ pub struct SequenceState {
     /// the scheduler to populate `usage.prompt_tokens_details.cached_tokens`.
     /// 0 when prefix caching is disabled or the prompt had no cache match.
     pub cached_prefix_tokens: usize,
+    /// Whether the chunk-0 prefix-cache lookup already ran for this sequence.
+    ///
+    /// The lookup is NOT idempotent: it bumps radix refs, `inc_ref`s each
+    /// matched KV block and PUSHES it onto `block_table`. It also runs BEFORE
+    /// `ensure_blocks_through_prefill`, so a chunk-0 prefill that fails to
+    /// allocate its suffix (KV exhausted) leaves all of that applied. The
+    /// preempt-and-retry in `run_standard_chunk_loop` re-enters `prefill_chunk`
+    /// for the SAME chunk, which would run the lookup a second time — appending
+    /// the matched blocks to `block_table` again (so `block_table[i]` no longer
+    /// maps to logical block `i`) and taking a second radix ref that the single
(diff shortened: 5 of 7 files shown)


rsafier:
**Widened to cover the prefill path too** (2nd commit).

Hit live on a second model (GGUF Q4_K_M, 259 requests):

```
DEBUG block_mgmt: prefix-cache evict reclaimed 0/1 blocks (free=0)
ERROR phase_continue_prefills::run_standard: Prefill chunk error: KV cache exhausted: no free blocks
```

**Severity differs between the two halves, and I want to be straight about it:**

| path | before | after |
|---|---|---|
| `decode_step` | exhaustion errored **every** sequence in the batch (up to `max-num-seqs` requests destroyed for one missing block) | preempt largest, retry → the rest survive |
| `run_standard` (prefill chunk) | exhaustion killed **only that sequence** (already graceful — `(idx, None)` → `phase_promote_prefills` frees it) | preempt a larger in-flight sequence so the newcomer proceeds |

So the decode half is a genuine crash-class fix; the prefill half is more a **policy change** — on exhaustion, evict a larger older conversation rather than fail the incoming request. Reviewers may reasonably prefer the opposite priority (fail the newcomer, protect long-running work); happy to drop that commit or invert the victim choice if so.

Both use the same mechanism: match `"KV cache exhausted"`, pick the largest active non-grammar sequence, `send_error` (which frees its blocks via `free_sequence`), retry. Retry is safe because `ensure_blocks_through_{decode,prefill}` fail during block allocation, **before** the forward pass, so no GPU state is mutated.

Also confirmed the underlying exhaustion is transient capacity, **not** a leak: after the event the pool recovered unaided and a 13,067-token prompt succeeded at idle in 11.5s. (A leak leaves it permanently wedged — that was the pre-fix signature in #373.)

Still draft: neither path has been observed executing its preemption yet. Forcing the condition deliberately next.

tbraun96:
Closing as **merged elsewhere, not rejected** — same story as #373, including the attribution problem.

### It's on main verbatim

`crates/spark-server/src/scheduler/decode_step.rs:88–127` is this PR's preempt loop, down to the comment text (*"measured: 8 seqs x 1173 blocks = 9384 needed vs a 9058-block pool"*) and the `warn!("KV cache exhausted during decode: preempting slot=…")`. The prefill arm is at `phase_continue_prefills/run_standard.rs:275,288,297`, and the idempotence guard at `prefill_b/prefix_lookup.rs:31` (`if chunk_start == 0 && seq.prefix_lookup_applied`).

Line-level sweep: **74 added lines, 0 missing from main.** Verified by content diff, not ancestry.

It landed via **#388** (`744420904`), which I authored, with no reference to this PR and no `Co-authored-by`. Mine to own — see the note on #373.

### Merging it now would regress main

`run_standard.rs` would lose the SSM-slot sort added since, whose own comment measures what its absence costs — **853 µs/layer vs 618 batched, a 28% penalty on a ~30 ms step** — plus `pending_draft_conf.clear()` and the `sched.levers.sampling()` threading.

### The gap in your body is still live, and it's now filed

You documented that **prefill-start** KV exhaustion still 500s while the identical condition mid-decode and mid-prefill-chunk preempts. Still true on main:

`phase_start_prefills.rs:338` matches `"pool exhausted"` and falls through to a bare `tracing::error!` at `:348`. The KV error says `"KV cache exhausted"`, so it never matches — even though the function's own doc comment reads *"never return HTTP 500 for resource exhaustion."* Observed 2–3 streams per run.

Filed as **#422**, credited to you, and it carries the part that actually matters: your reasoning for why the one-line widening is **wrong** —

> by that point the request has already failed, so killing a live sequence would destroy in-flight work and rescue nothing

— and that the real fix is defer/requeue. I added one observation of my own: the dispatch is a substring match on a formatted error string, which is *why* the two exhaustion cases could diverge silently. A typed error would have made the missing arm a compile error.

### One note for your other PRs

This branch conflicts with main, and a conflicting PR makes CI **silently skip** — so its green checks were never real. Same on #266 and #332. Worth rebasing anything you want reviewed.

PR #378, opened 2026-07-28
Enterprise Concurrency

AzeezIsh:
## Enterprise Concurrency

Atlas was winning single-stream decode and losing everywhere else. This branch closes that: **three of five target concurrencies now beat vLLM**, and the two that don't have a measured, arithmetic path to the bar.

| C | Atlas | vLLM | ratio | |
|---|---|---|---|---|
| 1 | 25.62 | 14.2 | **1.80x** | MET |
| 2 | 35.18 | 27.8 | **1.27x** | MET |
| 4 | ~57.0 | 53.3 | **1.07x** | MET |
| 8 | 88.9 | 98.8 | 0.90x | |
| 16 | 131.9 | 168.9 | 0.78x | |

Model `centml/Qwen3.6-27B-NVFP4-W4A4-mlpinf` on GB10. vLLM bars are the NVIDIA-supplied GB10 reference numbers.

### What changed

**Kernel work (all byte-identical unless noted).** The decode step was carrying costs nobody had counted: a scalar GEMV where a tile GEMM already existed, 774 launches/step that existed only because a kernel assumed a packed row layout, and a weight layout that forced nine 4-byte global loads per 36-byte block where two 16-byte loads would do.

**Batched speculative decoding.** MTP previously ran one sequence at a time, so at C=n it performed n full model forwards, each re-reading ~9.6 GB of weights on an engine already at 87% of achievable bandwidth. That is why enabling it at C=2 *lost* 3.4% and raising the cap *halved* C=4. This branch makes n*(K+1) verify rows share one weight read, adds slot-keyed CUDA graphs for the verify step, batches the GDN conv/WY across sequences, and adds a K-vs-batch ladder so draft depth shrinks as concurrency grows.

### Notable fixes found along the way

- **CUDA 716 that had blocked the lm_head tile GEMM through three prior attempts** — the transposed weight's row stride is the vocab size, and this checkpoint's 248077 is odd, so 16-byte `cp.async` was misaligned on 15 of every 16 k-rows. The reproducer that "exonerated" the kernel had hardcoded N=248320 from a sibling checkpoint, a shape where the fault cannot occur.
- **Silent garbage logits on Windows/HIP at C>=5** — `kernels/strix/` is a symlink and inherited the `ldb` fix; `kernels/strix-hip/` is a real file and drifted. RDNA tolerates the misalignment that traps on CUDA, so it would have produced wrong answers with no error.
- **A shipped, default-ON lever that was completely inert** — `rms_norm_strided` was resolved under module `"rms_norm"` when every `KERNEL.toml` maps it as `"norm"`. `try_kernel` swallows the miss, so its kill-switch A/B had been comparing the fallback against itself.

### Verification

Every performance change was A/B'd against its own kill switch with >=3 reps and non-overlapping ranges. Byte-identity was verified by output hash, not asserted from construction. The W4A16 parity oracle passes. Coherence and tool-call smokes pass at final defaults with zero CUDA errors.

Several claims were measured and **retracted**: dropping W4A4 at decode (a 10.4% regression — the M=1 literature does not transfer to M=16 batched serving), a 4-stage pipeline (5% regression — depth trades in-flight bytes against occupancy), and the `16:1` ladder rung (C=16 regression, reverted rather than shipped).

### Accuracy debt

Spec-decode paths change emitted tokens at temp 0 by construction. BFCL/IoU re-validation is required before any accuracy claim; it is deliberately deferred until throughput parity is reached. The lm_head FP8 activation path is the one non-byte-identical kernel change (bounded: 2/8 byte-identical, 6/8 benign paraphrase on a short-prompt probe); `ATLAS_LMHEAD_LOSSLESS=1` selects a lossless BF16 path at ~1.8% cost.

### Still open

C=8 needs the verify step's cost cut from ~1.85x to <=1.67x a plain decode step; C=16 needs <=1.34x from ~1.77x. Depth will not buy C=16 — K=3 at n=16 measured ~3.14x, so GDN is superlinear in K there. Work in progress on graphing the Phase-A bootstrap and raising accept rate (currently p1 ~0.72; refeed reached 0.90 on strix).


tbraun96:
## Full progression — every measured table, start to now

All numbers aggregate tok/s on `centml/Qwen3.6-27B-NVFP4-W4A4-mlpinf`, GB10, `--num-drafts 3`, temp 0.
vLLM reference bars: **14.2 / 27.8 / 53.3 / 98.8 / 168.9** at C = 1 / 2 / 4 / 8 / 16.

---

### 0. Where we started — concurrency flat-lined

The campaign opened with C=16 at **59.9 tok/s**, ratio **0.35x**. The root cause was found first: Atlas dispatched by M into a scalar GEMV where a tensor-core tile GEMM already existed in-tree, in three separate places. Fixing those took C=16 to 95.9 (+60%).

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 16 | 59.9 | 168.9 | 0.35x |

### 1. Session start (after the m_dispatch fixes)

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 1 | 25.4 | 14.2 | 1.79x |
| 2 | 25.2 | 27.8 | 0.91x |
| 4 | 48.0 | 53.3 | 0.90x |
| 8 | 74.2 | 98.8 | 0.75x |
| 16 | 113.1 | 168.9 | 0.67x |

C=2 sitting at C=1's number was the tell: speculation switched off at C>=2.

---

### 2. Kernel levers (five wins, four byte-identical)

| lever | delta at C=16 | byte-identical |
|---|---|---|
| strided RoPE + q/k norm fan-out (774 -> 48 launches/step) | +0.87% | yes |
| lm_head GEMV -> tile GEMM (solved the CUDA 716) | +5.50% | no (FP8 activations) |
| FFN MMQ AoS -> SoA weight layout (9 global load ops -> 3) | +4.6% | yes |
| `w4a16_gemm_t_k64` 3-deep weight pipeline | +1.81% | yes |
| `w4a16_gemm_t` 3-deep weight pipeline | +2.57% | yes |
| in-place Q read (256 D2D copies/step eliminated) | +0.40% | yes |

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 1 | 25.45 | 14.2 | 1.79x |
| 2 | 26.40 | 27.8 | 0.95x |
| 4 | 48.05 | 53.3 | 0.90x |
| 8 | 74.75 | 98.8 | 0.76x |
| 16 | 131.25 | 168.9 | 0.78x |

At this point the byte-identical seam was measurably spent: 87% of the step was four kernels running at 86-100% of achievable memory bandwidth, and the entire remaining basket summed to ~1.5 ms of a 110 ms step.

---

### 3. Wave 1 — batched verify lands, C=2 crosses

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 1 | 25.60 | 14.2 | 1.80x MET |
| 2 | **29.40** | 27.8 | **1.058x MET** |
| 4 | 48.70 | 53.3 | 0.91x |
| 8 | 74.85 | 98.8 | 0.76x |
| 16 | 131.75 | 168.9 | 0.78x |

First parity crossing beyond C=1. The same wave measured the **cap=16 collapse**: with fixed K=4 over n sequences, C=8 and C=16 plateau at ~50 tok/s regardless of concurrency — the fixed-large-K failure mode, reproduced on our own engine. Cap defaulted to the measured crossover.

### 4. Wave 2 — verify graphs, GDN batching, accept fix; C=4 crosses

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 1 | 25.55 | 14.2 | 1.80x MET |
| 2 | **35.35** | 27.8 | **1.27x MET** |
| 4 | **54.1** | 53.3 | **1.01x MET** |
| 8 | 73.5 | 98.8 | 0.74x |
| 16 | 131.0 | 168.9 | 0.78x |

Slot-keyed CUDA graphs for the batched verify, the batched GDN conv/WY engaged (8n launches -> 2 per layer), and an accept-rate divergence fixed (p1 restored 0.59-0.71 -> 0.70-0.83).

### 5. Wave 3 — K-vs-batch ladder

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 1 | 25.62 | 14.2 | 1.80x MET |
| 2 | 35.18 | 27.8 | 1.27x MET |
| 4 | 55.6 | 53.3 | 1.04x MET |
| 8 | **81.2** | 98.8 | 0.82x |
| 16 | 131.8 | 168.9 | 0.78x |

The ladder measured a C=16 regression at the `16:1` rung and it was **not shipped** — n>8 reverts to the MTP-off path, which holds the floor exactly.

### 6. Wave 4 — verify-cost cuts (current)

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 1 | 25.62 | 14.2 | **1.80x MET** |
| 2 | 35.18 | 27.8 | **1.27x MET** |
| 4 | ~57.0 | 53.3 | **1.07x MET** |
| 8 | **88.9** | 98.8 | 0.90x |
| 16 | 131.9 | 168.9 | 0.78x |

K-parameterized the cross-seq GDN fast path (was K=4-only), widened batched propose to one drafter forward at n=16 (was chunks of <=4), batched the Phase-A bootstrap. Default ladder moved to `4:3,8:2`.

---

### The remaining gap, as arithmetic

Speedup = expected tokens per step / verify-step cost. Measured cost: **~1.85x** at n=8, **~1.77x** at n=16 (down from ~1.9x).

- **C=8** needs 98.8/73.5 = 1.34x, so at K=3 (~2.24 expected tokens) cost must reach **<=1.67x** — a ~10% cut.
- **C=16** needs 1.28x, so at K=2 cost must reach **<=1.34x** — a ~25% cut. Depth cannot buy it: K=3 at n=16 measured ~3.14x, i.e. GDN is superlinear in K.

Two multipliers remain in flight: graphing the Phase-A bootstrap (still routed through `decode_batch`, which disables graphs at n>=2) and raising accept rate from p1 ~0.72 (refeed reached 0.90 on strix — a +15% token-per-step lift alone projects C=8 ~102).

---

### Net movement

| C | start | now | change |
|---|---|---|---|
| 1 | 25.4 | 25.62 | +0.9% |
| 2 | 25.2 | 35.18 | **+39.6%** |
| 4 | 48.0 | ~57.0 | **+18.8%** |
| 8 | 74.2 | 88.9 | **+19.8%** |
| 16 | 113.1 | 131.9 | **+16.6%** |

Against the campaign's original baseline, C=16 has gone **59.9 -> 131.9, +120%**, ratio **0.35x -> 0.78x**.

tbraun96:
Superseded by #379, which carries the same work rebased cleanly onto main — source and kernels only, no scratch artifacts, and the TUI (#377) + metrics-guard (#374) changes preserved through the merge rather than reverted.

Issue #323, opened 2026-07-15
Single-seq decode CUDA graph destroyed on free_sequence → unnecessary re-capture per slot occupant (all models)


rsafier:
## Summary

The single-sequence decode **CUDA graph is destroyed on every sequence free** (`free_sequence`), even though the graph is fully **reusable across any occupant of the same slot**. As a result each new sequence that reuses a slot re-captures an identical graph on its first decode step. This is generic model infrastructure (`decode_graph`), so **it affects every model** that uses the single-seq (C=1 / batch-drains-to-1) decode path.

Observed empirically on Qwen3.6-27B-NVFP4 (GB10): over a ~33-min mixed C=1..8 run, `slot=0` was captured **8×** and `slot=1` **2×** (10 captures total) — i.e. slots repeatedly go cold and re-capture, rather than capturing once and persisting.

## Mechanism

The single-seq decode graph cache `decode_graph: HashMap<slot_idx, GraphHandle>` is:

- **Replayed** slot-keyed (`crates/spark-model/src/model/trait_impl/decode_a.rs`, Phase 2): `cache.get(&seq.slot_idx)` → `launch_graph`. All per-step dynamic data is read from device staging refreshed *before* replay.
- **Captured** on a cold-slot miss (decode_a.rs Phase 3): `end_capture` → `cache.insert(seq.slot_idx, graph)`.
- **Destroyed on free** (`crates/spark-model/src/model/trait_impl/sequence.rs`, `free_sequence`):
  ```rust
  // Invalidate cached CUDA graphs that reference this sequence's slot
  // — the graph was captured with this slot's KV/SSM pointers baked in,
  // and replaying after the slot is freed would read stale data.
  if let Some(graph) = self.decode_graph.lock().remove(&seq.slot_idx) { destroy_graph(graph) }
  ```

So under normal request turnover — and *maximally* under variable-length co-dispatch, where sequences finish at staggered times and free slots constantly — every slot reuse pays a fresh capture (a full 64-layer build + non-overlappable sync).

## Why the "stale data" rationale does not hold

Nothing baked into the graph is unique to a specific sequence beyond its **slot**:

| What varies per step / per sequence | How it reaches the kernel | Baked into graph? |
|---|---|---|
| positions, slot, seq_len, block_table, token_id | uploaded every step to a **fixed** `meta_base = buffers.scratch().offset(32768)` staging region (decode_a.rs Phase 1) | No — refreshed pre-replay |
[...]

PR #379, opened 2026-07-29
Enterprise Concurrency

tbraun96:
## Enterprise Concurrency

Atlas was winning single-stream decode and losing everywhere else. This branch closes that: **three of five target concurrencies now beat vLLM**, and the two that don't have a measured, arithmetic path to the bar.

| C | Atlas | vLLM | ratio | |
|---|---|---|---|---|
| 1 | 25.62 | 14.2 | **1.80x** | MET |
| 2 | 35.18 | 27.8 | **1.27x** | MET |
| 4 | ~57.0 | 53.3 | **1.07x** | MET |
| 8 | 88.9 | 98.8 | 0.90x | |
| 16 | 131.9 | 168.9 | 0.78x | |

Model `centml/Qwen3.6-27B-NVFP4-W4A4-mlpinf` on GB10. vLLM bars are the NVIDIA-supplied GB10 reference numbers.

### What changed

**Kernel work (all byte-identical unless noted).** The decode step was carrying costs nobody had counted: a scalar GEMV where a tile GEMM already existed, 774 launches/step that existed only because a kernel assumed a packed row layout, and a weight layout that forced nine 4-byte global loads per 36-byte block where two 16-byte loads would do.

**Batched speculative decoding.** MTP previously ran one sequence at a time, so at C=n it performed n full model forwards, each re-reading ~9.6 GB of weights on an engine already at 87% of achievable bandwidth. That is why enabling it at C=2 *lost* 3.4% and raising the cap *halved* C=4. This branch makes n*(K+1) verify rows share one weight read, adds slot-keyed CUDA graphs for the verify step, batches the GDN conv/WY across sequences, and adds a K-vs-batch ladder so draft depth shrinks as concurrency grows.

### Notable fixes found along the way

- **CUDA 716 that had blocked the lm_head tile GEMM through three prior attempts** — the transposed weight's row stride is the vocab size, and this checkpoint's 248077 is odd, so 16-byte `cp.async` was misaligned on 15 of every 16 k-rows. The reproducer that "exonerated" the kernel had hardcoded N=248320 from a sibling checkpoint, a shape where the fault cannot occur.
- **Silent garbage logits on Windows/HIP at C>=5** — `kernels/strix/` is a symlink and inherited the `ldb` fix; `kernels/strix-hip/` is a real file and drifted. RDNA tolerates the misalignment that traps on CUDA, so it would have produced wrong answers with no error.
- **A shipped, default-ON lever that was completely inert** — `rms_norm_strided` was resolved under module `"rms_norm"` when every `KERNEL.toml` maps it as `"norm"`. `try_kernel` swallows the miss, so its kill-switch A/B had been comparing the fallback against itself.

### Verification

Every performance change was A/B'd against its own kill switch with >=3 reps and non-overlapping ranges. Byte-identity was verified by output hash, not asserted from construction. The W4A16 parity oracle passes. Coherence and tool-call smokes pass at final defaults with zero CUDA errors.

Several claims were measured and **retracted**: dropping W4A4 at decode (a 10.4% regression — the M=1 literature does not transfer to M=16 batched serving), a 4-stage pipeline (5% regression — depth trades in-flight bytes against occupancy), and the `16:1` ladder rung (C=16 regression, reverted rather than shipped).

### Accuracy debt

Spec-decode paths change emitted tokens at temp 0 by construction. BFCL/IoU re-validation is required before any accuracy claim; it is deliberately deferred until throughput parity is reached. The lm_head FP8 activation path is the one non-byte-identical kernel change (bounded: 2/8 byte-identical, 6/8 benign paraphrase on a short-prompt probe); `ATLAS_LMHEAD_LOSSLESS=1` selects a lossless BF16 path at ~1.8% cost.

### Still open

C=8 needs the verify step's cost cut from ~1.85x to <=1.67x a plain decode step; C=16 needs <=1.34x from ~1.77x. Depth will not buy C=16 — K=3 at n=16 measured ~3.14x, so GDN is superlinear in K there. Work in progress on graphing the Phase-A bootstrap and raising accept rate (currently p1 ~0.72; refeed reached 0.90 on strix).

### DIFF
--- a/crates/atlas-kernels/build.rs
+++ b/crates/atlas-kernels/build.rs
@@ -197,6 +197,11 @@ fn main() {
 
     // Per-target: (target_idx, vec of (stem, module_name))
     let mut all_target_modules: Vec<Vec<(String, String)>> = Vec::new();
+    // Per-target: (module, kernel) pairs this model's files dropped by shadowing
+    // their `common/` namesakes. Baked into the binary so the startup audit can
+    // separate "dropped by a fork" (fatal) from "never built for this target"
+    // (expected). See `shadowed_dropped_pairs`.
+    let mut all_target_drops: Vec<Vec<(String, String)>> = Vec::new();
 
     // 2026-05-24 dedup+parallel: pre-walk every (target, cu_file) pair
     // and split into two queues:
@@ -336,6 +341,30 @@ fn main() {
 
         all_target_modules.push(modules);
 
+        let drops = shadowed_dropped_pairs(
+            target.common_kernel_dir.as_deref(),
+            &target.model_kernel_dir,
+            source_ext,
+            &target.module_overrides,
+        );
+        if !drops.is_empty() {
+            println!(
+                "cargo:warning=atlas-kernels: ({}, {}) drops {} kernel(s) by shadowing common/: {}. \
+                 A dropped kernel fails CLOSED (try_kernel -> handle 0). If the model genuinely \
+                 cannot use it this is fine; the startup audit will only hard-error if the model's \
+                 dispatch actually requests it.",
+                target.model,
+                target.quant,
+                drops.len(),
+                drops
+                    .iter()
+                    .map(|(m, f)| format!("{m}::{f}"))
+                    .collect::<Vec<_>>()
+                    .join(", "),
+            );
+        }
+        all_target_drops.push(drops);
+
         println!(
             "cargo:rerun-if-changed={}",
             target.model_kernel_dir.display()
@@ -436,8 +465,13 @@ fn main() {
     }
 
     // ── Generate target_ptx.rs ──
-    let generated =
-        generate_target_ptx_rs(&targets, &all_target_modules, output_ext, uses_cuda_api);
+    let generated = generate_target_ptx_rs(
+        &targets,
+        &all_target_modules,
+        &all_target_drops,
+        output_ext,
+        uses_cuda_api,
+    );
     let gen_path = out_dir.join("target_ptx.rs");
     std::fs::write(&gen_path, &generated)
         .unwrap_or_else(|e| panic!("Failed to write {}: {e}", gen_path.display()));
@@ -1045,3 +1079,120 @@ use build_codegen::generate_target_ptx_rs;
 mod build_target;
 use build_target::resolve_compute_target;
 // Force recompilation 1775404930
+
+/// Every `(module, kernel)` this target's model-specific files drop relative to
+/// their `common/` namesakes.
+///
+/// SHADOWING DRIFT. Shadowing is whole-file, not per-symbol: a model file
+/// replaces its common namesake entirely. When `common/` later gains a kernel
+/// the fork never picked up, that kernel silently vanishes for this model —
+/// `try_kernel` returns handle 0 and whatever depends on it fails CLOSED. That
+/// is exactly how the 27B lost the four multi-sequence decode kernels
+/// (`gated_delta_rule_decode_f32_{norm,conv_norm,strided,strided_norm}`),
+/// pinning every concurrent decode to the per-sequence fallback until
+/// 2026-07-26.
+///
+/// This list is baked into the binary (`TargetPtxSet::shadowed_dropped`) so the
+/// startup kernel audit can tell the two failure classes apart:
+///
+///   * a lookup that failed AND is in this list — the kernel EXISTS in
+///     `common/`, this model's fork dropped it. A build defect: fatal.
+///   * a lookup that failed and is NOT — the kernel was never compiled for this
+///     target at all (typically another architecture's: MLA, hyper-connection,
+///     CSA). Expected, and merely informational.
+///
+/// Without that split the audit prints ~26 benign entries for a Qwen model and
+/// the one that matters hides among them.
+fn shadowed_dropped_pairs(
+    common_dir: Option<&std::path::Path>,
+    model_dir: &std::path::Path,
+    source_ext: &str,
+    module_overrides: &HashMap<String, String>,
+) -> Vec<(String, String)> {
+    let Some(common) = common_dir else {
+        return Vec::new();
+    };
+    let common_by_stem: HashMap<String, PathBuf> = find_cu_files(common, source_ext)
+        .into_iter()
+        .map(|f| (f.file_stem().unwrap().to_str().unwrap().to_string(), f))
+        .collect();
+    let mut out = Vec::new();
+    for f in find_cu_files(model_dir, source_ext) {
+        let stem = f.file_stem().unwrap().to_str().unwrap().to_string();
+        let Some(common_f) = common_by_stem.get(&stem) else {
+            continue;
+        };
+        let module = module_overrides.get(&stem).cloned().unwrap_or(stem);
+        for func in shadowed_missing_symbols(common_f, &f) {
+            out.push((module.clone(), func));
+        }
+    }
+    out.sort();
+    out
+}
+
+/// Kernel entry points a shadowing model file drops relative to its `common/`
+/// namesake. Scans for `extern "C" __global__ void <name>` (tolerating an
+/// interposed `__launch_bounds__(...)`), which is how every Atlas kernel entry
+/// point is declared. Text-scan by design: it runs before nvcc, so it cannot
+/// depend on compiled artifacts. Returns sorted names for stable warnings.
+fn shadowed_missing_symbols(
+    common_file: &std::path::Path,
+    model_file: &std::path::Path,
+) -> Vec<String> {
+    fn entry_points(p: &std::path::Path) -> std::collections::BTreeSet<String> {
+        let Ok(text) = std::fs::read_to_string(p) else {
+            return Default::default();
+        };
+        let mut out = std::collections::BTreeSet::new();
+        for (idx, _) in text.match_indices("__global__") {
+            let tail = &text[idx..];
+            // Skip `__global__`, an optional `__launch_bounds__(...)`, and the
+            // return type, then take the identifier before `(`.
+            let Some(paren) = tail.find('(') else {
+                continue;
+            };
+            let mut head = &tail[..paren];
+            if let Some(lb) = head.find("__launch_bounds__") {
+                // `__launch_bounds__(N, M)` — the name follows its close paren.
+                let after = &tail[idx_after_balanced(tail, lb)..];
+                let Some(p2) = after.find('(') else { continue };
+                head = &after[..p2];
+            }
+            if let Some(name) = head.split_whitespace().last() {
+                let name = name.trim_start_matches('*');
+                if !name.is_empty() && name.chars().all(|c| c.is_alphanumeric() || c == '_') {
+                    out.insert(name.to_string());
+                }
+            }
+        }
+        out
+    }
+    // Index just past the balanced `(...)` that follows `from` in `s`.
+    fn idx_after_balanced(s: &str, from: usize) -> usize {
+        let bytes = s.as_bytes();
+        let mut i = from;
+        while i < bytes.len() && bytes[i] != b'(' {
+            i += 1;
+        }
+        let mut depth = 0usize;
+        while i < bytes.len() {
+            match bytes[i] {
+                b'(' => depth += 1,
+                b')' => {
+                    depth -= 1;
+                    if depth == 0 {
+                        return i + 1;
+                    }
+                }
+                _ => {}
+            }
+            i += 1;
+        }
+        from
+    }
+    entry_points(common_file)
+        .difference(&entry_points(model_file))
+        .cloned()
+        .collect()
+}
--- a/crates/atlas-kernels/build_codegen.rs
+++ b/crates/atlas-kernels/build_codegen.rs
@@ -24,6 +24,7 @@ use super::{SamplingCat, Target};
 pub(super) fn generate_target_ptx_rs(
     targets: &[Target],
     all_modules: &[Vec<(String, String)>],
+    all_drops: &[Vec<(String, String)>],
     output_ext: &str,
     cuda_api: bool,
 ) -> String {
@@ -180,6 +181,7 @@ pub(super) fn generate_target_ptx_rs(
              \x20           }},\n\
              \x20           model_type_matches: vec![{}],\n\
              \x20           dflash: {},\n\
+             \x20           shadowed_dropped: &[{}],\n\
              \x20       }},\n",
             target.model, target.quant,
             fmt_cat(&target.sampling_thinking_text),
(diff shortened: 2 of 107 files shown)


tbraun96:
## Full progression — every measured table, start to now

All numbers aggregate tok/s on `centml/Qwen3.6-27B-NVFP4-W4A4-mlpinf`, GB10, `--num-drafts 3`, temp 0.
vLLM reference bars: **14.2 / 27.8 / 53.3 / 98.8 / 168.9** at C = 1 / 2 / 4 / 8 / 16.

---

### 0. Where we started — concurrency flat-lined

The campaign opened with C=16 at **59.9 tok/s**, ratio **0.35x**. The root cause was found first: Atlas dispatched by M into a scalar GEMV where a tensor-core tile GEMM already existed in-tree, in three separate places. Fixing those took C=16 to 95.9 (+60%).

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 16 | 59.9 | 168.9 | 0.35x |

### 1. Session start (after the m_dispatch fixes)

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 1 | 25.4 | 14.2 | 1.79x |
| 2 | 25.2 | 27.8 | 0.91x |
| 4 | 48.0 | 53.3 | 0.90x |
| 8 | 74.2 | 98.8 | 0.75x |
| 16 | 113.1 | 168.9 | 0.67x |

C=2 sitting at C=1's number was the tell: speculation switched off at C>=2.

---

### 2. Kernel levers (five wins, four byte-identical)

| lever | delta at C=16 | byte-identical |
|---|---|---|
| strided RoPE + q/k norm fan-out (774 -> 48 launches/step) | +0.87% | yes |
| lm_head GEMV -> tile GEMM (solved the CUDA 716) | +5.50% | no (FP8 activations) |
| FFN MMQ AoS -> SoA weight layout (9 global load ops -> 3) | +4.6% | yes |
| `w4a16_gemm_t_k64` 3-deep weight pipeline | +1.81% | yes |
| `w4a16_gemm_t` 3-deep weight pipeline | +2.57% | yes |
| in-place Q read (256 D2D copies/step eliminated) | +0.40% | yes |

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 1 | 25.45 | 14.2 | 1.79x |
| 2 | 26.40 | 27.8 | 0.95x |
| 4 | 48.05 | 53.3 | 0.90x |
| 8 | 74.75 | 98.8 | 0.76x |
| 16 | 131.25 | 168.9 | 0.78x |

At this point the byte-identical seam was measurably spent: 87% of the step was four kernels running at 86-100% of achievable memory bandwidth, and the entire remaining basket summed to ~1.5 ms of a 110 ms step.

---

### 3. Wave 1 — batched verify lands, C=2 crosses

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 1 | 25.60 | 14.2 | 1.80x MET |
| 2 | **29.40** | 27.8 | **1.058x MET** |
| 4 | 48.70 | 53.3 | 0.91x |
| 8 | 74.85 | 98.8 | 0.76x |
| 16 | 131.75 | 168.9 | 0.78x |

First parity crossing beyond C=1. The same wave measured the **cap=16 collapse**: with fixed K=4 over n sequences, C=8 and C=16 plateau at ~50 tok/s regardless of concurrency — the fixed-large-K failure mode, reproduced on our own engine. Cap defaulted to the measured crossover.

### 4. Wave 2 — verify graphs, GDN batching, accept fix; C=4 crosses

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 1 | 25.55 | 14.2 | 1.80x MET |
| 2 | **35.35** | 27.8 | **1.27x MET** |
| 4 | **54.1** | 53.3 | **1.01x MET** |
| 8 | 73.5 | 98.8 | 0.74x |
| 16 | 131.0 | 168.9 | 0.78x |

Slot-keyed CUDA graphs for the batched verify, the batched GDN conv/WY engaged (8n launches -> 2 per layer), and an accept-rate divergence fixed (p1 restored 0.59-0.71 -> 0.70-0.83).

### 5. Wave 3 — K-vs-batch ladder

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 1 | 25.62 | 14.2 | 1.80x MET |
| 2 | 35.18 | 27.8 | 1.27x MET |
| 4 | 55.6 | 53.3 | 1.04x MET |
| 8 | **81.2** | 98.8 | 0.82x |
| 16 | 131.8 | 168.9 | 0.78x |

The ladder measured a C=16 regression at the `16:1` rung and it was **not shipped** — n>8 reverts to the MTP-off path, which holds the floor exactly.

### 6. Wave 4 — verify-cost cuts (current)

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 1 | 25.62 | 14.2 | **1.80x MET** |
| 2 | 35.18 | 27.8 | **1.27x MET** |
| 4 | ~57.0 | 53.3 | **1.07x MET** |
| 8 | **88.9** | 98.8 | 0.90x |
| 16 | 131.9 | 168.9 | 0.78x |

K-parameterized the cross-seq GDN fast path (was K=4-only), widened batched propose to one drafter forward at n=16 (was chunks of <=4), batched the Phase-A bootstrap. Default ladder moved to `4:3,8:2`.

---

### The remaining gap, as arithmetic

Speedup = expected tokens per step / verify-step cost. Measured cost: **~1.85x** at n=8, **~1.77x** at n=16 (down from ~1.9x).

- **C=8** needs 98.8/73.5 = 1.34x, so at K=3 (~2.24 expected tokens) cost must reach **<=1.67x** — a ~10% cut.
- **C=16** needs 1.28x, so at K=2 cost must reach **<=1.34x** — a ~25% cut. Depth cannot buy it: K=3 at n=16 measured ~3.14x, i.e. GDN is superlinear in K.

Two multipliers remain in flight: graphing the Phase-A bootstrap (still routed through `decode_batch`, which disables graphs at n>=2) and raising accept rate from p1 ~0.72 (refeed reached 0.90 on strix — a +15% token-per-step lift alone projects C=8 ~102).

---

### Net movement

| C | start | now | change |
|---|---|---|---|
| 1 | 25.4 | 25.62 | +0.9% |
| 2 | 25.2 | 35.18 | **+39.6%** |
| 4 | 48.0 | ~57.0 | **+18.8%** |
| 8 | 74.2 | 88.9 | **+19.8%** |
| 16 | 113.1 | 131.9 | **+16.6%** |

Against the campaign's original baseline, C=16 has gone **59.9 -> 131.9, +120%**, ratio **0.35x -> 0.78x**.

tbraun96:
Updated with wave 5.

| C | Atlas | vLLM | ratio |
|---|---|---|---|
| 1 | 25.53 | 14.2 | **1.80x** MET |
| 2 | 35.37 | 27.8 | **1.27x** MET |
| 4 | 60.40 | 53.3 | **1.13x** MET |
| 8 | 95.84 | 98.8 | 0.970x |
| 16 | 131.4 | 168.9 | 0.78x |

C=8 has gone 73.5 -> 81.2 -> 88.9 -> 95.84 across four waves and is now 3.1% from the bar.

The wave's real finding is that a dispatch cap had been faking a physical limit. The batched verify chunked at 4 sequences instead of the row-buffer bound, so an 8-wide K=4 batch ran two serialized forwards. Two earlier conclusions died with it: depth at n=8 is *not* wrong (K=4 is the best point there), and C=16 spec does *not* collapse (it is at parity, 131.93 vs a 131.42 control — it had measured 0.973x under the artifact).

Both remaining gaps are now quantified rather than estimated. C=8 needs 1.031x: either tok_step 2.606 -> 2.69 (p1 0.793 -> ~0.85) or a ~3% verify-step cut. C=16 sits exactly on break-even (implied cost ~1.79x against a 1.797x break-even at p1 0.797); clearing 168.9 needs cost <= 1.40x.

rsafier:
Ran this branch on a GB10 focused specifically on **concurrency correctness and crash behaviour** — not the perf claims. Build is this PR plus our KV-refcount instrumentation cherry-picked in (`dd370b5e`), serving via the production-shaped recipe (prefix caching on, `--max-num-seqs 8`, slai, fp8 KV, `--gpu-memory-utilization 0.80`).

Three things worth patching, one claim independently verified, and a scope caveat at the end.

---

## P0 — batched K=4 verify poisons the CUDA context under concurrency (on one supported checkpoint)

Reproduced twice. First fault is always `scheduler::verify_k4_batch_step`, immediately after a graph capture:

```
verify_e: Captured CUDA graph for batched K=4 verify (n=3, key=Some([1, 1, 2, 4, 4294967295]))
verify_k4_batch_step: decode_verify_batched (n=3 rows=4): cuMemcpyDtoHAsync_v2 failed: status 700
```

Once 700 lands the context is poisoned, so every subsequent request fails — the 24/24 below is one fault, not 24.

Isolation matrix, same build, same load (8 concurrent, shared ~1.5K-word prefix, `max_tokens=250`):

| checkpoint | MTP | C | result |
|---|---|---|---|
| `nvidia/Qwen3.6-27B-NVFP4` | on | 8 | **CUDA 700** (74 errors, then 35 on re-run) |
| `nvidia/Qwen3.6-27B-NVFP4` | on | 1 | 6/6 clean, 0 errors |
| `nvidia/Qwen3.6-27B-NVFP4` | **off** | 8 | 24/24 clean, 0 errors |
| `centml/Qwen3.6-27B-NVFP4-W4A4-mlpinf` (this PR's bench checkpoint) | on | 8 | 22/24, **0 errors**, no bleed |

So it needs **MTP + concurrency together**, and does not reproduce on the checkpoint the PR benchmarks — which is why the perf table is very plausibly unaffected. But `nvidia/Qwen3.6-27B-NVFP4` is a supported checkpoint on the same `qwen3.6-27b` target with identical dims, so this is reachable in normal use.

Two things look wrong in that slot key, both in one line:

- **slot `1` appears twice** — `[1, 1, 2, 4, ...]` (also seen on the plain decode graphs: `slots=Some([0, 1, 1, 2, 4, 5, 6, 8])`)
- **`4294967295` is `u32::MAX`** being carried as a live key element

and it fires in the same millisecond as a burst of `free_sequence` calls. That combination reads as a slot-keyed graph being replayed while its slots are freed and handed to another sequence — the same hazard family as #323, but fatal here rather than a re-capture cost.

Suggested patch: validate the slot vector before capture *and* before reuse — reject or dedupe duplicate slots, treat the sentinel as "no slot" rather than part of the key, and invalidate cached graphs keyed on any slot that `free_sequence` releases.

One candidate I checked and **ruled out**: `nvidia/...` ships no `mtp.lm_head` (its `mtp.*` set is `fc`/`layers`/`norm`/`pre_fc_norm_embedding`) while `centml/...` does — but `impl_a1.rs:187` already falls back with `mtp_lm_head_nvfp4.or(lm_head_nvfp4)`, so that difference isn't the cause.

## P1 — `PROPOSE_META_STRIDE` silently disables batched propose past ~7K context

`forward_batch.rs:294`:

```rust
ensure!(256 + bt_len <= PROPOSE_META_STRIDE, ...)   // PROPOSE_META_STRIDE = 2048
```

`bt_len = block_table.len() * 4`, so the bound is **448 block-table entries**; at 16 tokens/block that caps batched propose at **~7,168 tokens of total context**. I hit it at 607 blocks (~9.7K tokens):

```
batched bootstrap run_mtp_propose_batched: propose_batch: drafter block table 2428 exceeds meta stride
```

It fails safe — `ensure!` → `Err` → scheduler logs and `continue`s — so there's no corruption. Two issues though:

1. Past ~7K context the batched propose path quietly stops applying and falls back, so the feature is inactive exactly where long agentic contexts live. **What context length were the C=8/16 numbers measured at?** If above ~7K, the ladder may not have been engaged.
2. It logs at `ERROR`, per step — 74 occurrences in a few minutes of testing.

`PROPOSE_META_SEQS * PROPOSE_META_STRIDE` is only 32 KB total, so sizing the slab from `max_seq_len` looks cheap; and this reads as an expected-capacity condition, so `debug!` (once per sequence) fits better than `error!`.

## P2 — a scheduler invariant this PR invalidates

`phase_continue_prefills.rs:171` still asserts:

> Spec is off by construction when `active.len() ≥ 2`, so the mixed branch is safe there.

and `single_active_with_spec` only guards `active.len() == 1`. Batched speculative decode at C=2–16 is this PR's entire premise, so that invariant is now false. At minimum the comment is misleading; more importantly the mixed-batch gate was written assuming this combination could not occur, and it now can.

## P3 — pre-existing, but it lands in exactly this code

Independent of this PR, the default mixed path aliases prefill logits onto decode lanes: `mixed_forward_batch` runs `decode_batch` (writing lane `i` to row `i` of `buffers.logits()`), then `prefill_batch_chunk` finalizes each finishing stream to row `stream_idx` of the same arena, and the caller samples the decode rows *after*. So a decode lane can sample another request's first-token distribution.

Notably this branch's own fused path already avoids it — `decode_b2.rs:118` uses `logits.offset(padded_n * v * bf16)` — so it's just the default path that needs the same disjoint-row treatment. We fixed it downstream by threading a `row_base = n_decode`; happy to send it as a standalone PR if useful.

---

## Verified in the PR's favour

The odd-vocab CUDA 716 root cause in the description **checks out**. On `centml/Qwen3.6-27B-NVFP4-W4A4-mlpinf`:

- `lm_head.weight` is `[248320, 2560]` and `config.json` reports `vocab_size: 248320`
- but the tokenizer's max id is **248076** → true vocab **248077**, which is odd

So both halves of the claim hold: the real row stride is odd, and a reproducer hardcoding `N=248320` genuinely cannot hit the misalignment. (I was skeptical of this until I checked the tokenizer rather than the config.)

## Scope / caveats

- Concurrency-correctness and crash behaviour only. I did **not** evaluate any throughput claim, the accuracy debt (BFCL/IoU), the Windows/HIP `ldb` fix, C=16, or the W4A4 activation path — nothing here speaks to whether the perf table reproduces.
- The build under test carries our `dd370b5e` cherry-picked in, which includes two KV refcount **fixes** as well as the counters, so this is "PR + those fixes" — and the P0 crash reproduces anyway.
- Prefix caching was ON throughout; the P0 crash has not been tried with it off.

tbraun96:
Thanks @rsafier!

tbraun96:
Closing — superseded by **#388**, which landed this work on `main` on 2026-08-08.

Verified by content, not by ancestry (this repo squash-merges, so `merge-base --is-ancestor` is meaningless here):

| symbol | on `main` | on this branch |
|---|---|---|
| `ATLAS_SSM_H_FP16` | **21 files** | 0 |
| `mtp_max_seqs` | 6 files | 3 |
| `w4a16_gemv_batch16_kernel` | 4 files | 4 |

And the direction of the diff settles it: merging this branch would **remove ~79,500 lines** from `crates/` relative to `main`. It is a snapshot from 2026-07-30, before the campaign landed, not a change on top of it.

Nothing here is lost. The findings this branch was carrying are on `main` or recorded:

- the C=1..128 ladder, the FP16 GDN state work, the width gating, the `mmq_x=64` rung — all in #388
- the env-var recipe → CLI flags conversion — in #388, including that **6 of the 10 "config of record" vars were no-ops**
- the accuracy legs and the noise-floor rules that came out of them — in the gate baselines

Re-cutting from `main` would be the way to revive anything that turns out to be missing; rebasing 1,059 files across a campaign is not.

PR #381, opened 2026-07-29
perf(ssm-tier): 22x cheaper snapshot eviction + a bounded, functional disk tier

rsafier:
## Summary

The SSM snapshot spill tier ships default-off and isn't viable if you turn it on: **every eviction costs ~412 ms**, and the NVMe tier has **no size bound at all**. This makes eviction cheap, makes the disk tier bounded, and fixes a silent-corruption bug found along the way.

All measured on Holo-3.1-35B-A3B (GB10, 30 SSM layers, 66,846,720 B blob).

## 1. Eviction: 412 ms → 19 ms (22×)

`CudaBackend::copy_d2h` calls `cuStreamSynchronize` **inside every call**, and `spill_slot` issued 60 of them (2 per SSM layer) into a freshly `vec![0u8; 66_846_720]` blob allocated per eviction.

The tell: the mirror-image `fault_in_slot` has always used 60 `copy_h2d_async` + **one** sync and moves the same bytes in ~28 ms. Same data, same buffer type, 14× faster. The cost was the copy **shape** — not bandwidth, not layout.

- new `GpuBackend::copy_d2h_async` (`cuMemcpyDtoHAsync_v2`, no sync); trait default forwards to blocking `copy_d2h` so mock/metal stay correct
- `spill_slot` gathers with 60 async enqueues + exactly one trailing `synchronize`
- one reusable **page-locked** staging blob shared by spill and fault-in — removes the per-event 66 MB zero-fill, ~16,320 first-touch page faults, and the pageable→driver bounce. Falls back to heap with a named `warn!` if pinning fails.

Measured, same box and workload, `ATLAS_SSM_TIER_TIMING=1`:

```
before: gather+sync=392936..411336us  store.put=19397us  total=412334us
after:  gather+sync=  1384..  1419us  store.put=17421us  total= 18841us  staging=pinned
```

The gather is **278× faster** and stable to ±35 µs across spills.

Side effect: `spill_slot`'s `stream` argument was **dead** — `copy_d2h` discards it and enqueues on the default stream, so the documented ordering held only because callers happened to pass `default_stream()`. `copy_d2h_async` honours it.

## 2. The disk tier was unbounded

`ssm_tier/unified.rs` used `Residency::new`, whose own doc says *"Unbounded disk tier (no cap). Prefer `Residency::new_capped` in production."*

The in-code justification — *"a capped disk would let `make_disk_room` silently discard live decode rollback targets"* — guards a hazard that **cannot occur on that constructor**: the decode-rollback ring is pure GPU D2D (`save_decode`/`restore_decode` take no store, no free list, no eviction), and `build_decode_tier_store` has **zero production callers**. Capping the Marconi/prefix arm is safe by construction and needs no pinning, so `atlas-tier` required no semantic change.

New **`ATLAS_SSM_TIER_DISK_GB`** (GiB, f64; unset/0 = unbounded = byte-identical to today). Converts to records from the runtime `blob_bytes`, reserving one record of headroom because `locate` pulls a faulting key out of `disk_lru` while its record is still live. `32` → 513 records = 31.94 GiB steady, 31.9995 GiB worst case. Too small a budget is a hard init error, not a silent fallback to unbounded.

## 3. `Residency::alloc` could alias two keys onto one disk record ⚠️

**Pre-existing, and the piece with the strongest merge urgency** — reviewers may want this part landed independently of the perf work.

The `OnDisk` arm freed `disk_slot` **before** the fallible `acquire_slot()?`, while `map[key]` still pointed at it. On a swap-write failure (canonically ENOSPC) the index was freed with a live claim; a retry or `remove` then pushed it a second time and two keys got the same record. A later GET returned **another key's blob with `Ok(true)`** — a full SSM snapshot restored into the wrong sequence, silent, no error anywhere.

Reproduced with a fault-injecting store; both regression tests included:

```
GET(key 1) returned Ok(true) with [2,2,2,2,2,2,2,2] — key 2's blob.
```

Fixed by **ordering**, not by guarding the double-push: acquire the slot first, free only once it's in hand, roll back the `disk_lru` self-pin on failure. Making the free idempotent would have left the "record on the free list while a live map entry claims it" window open — which *is* the aliasing state.

## Also included

- **Spill-side depth gate** `ATLAS_SSM_SPILL_MIN_TOKENS` (default 1024) at victim selection. The read side has had `ATLAS_SSM_FAULT_MIN_TOKENS` since day one; nothing ever budgeted the write side, which is how a 400 ms eviction went unnoticed.
- Page-aligned `Residency::scratch` (removes another 66 MB bounce per O_DIRECT op).
- `Residency::locate` wedge fix — the `write_slot` arm bailed with `?` while `scratch` was moved out, permanently zero-lengthing it for the process.

## A 0-byte swap file is not a bug

`evict_coldest_to_disk` is the only caller of `swap.write_record` and runs only when the hot arena is full. With `ATLAS_SSM_TIER_SLOTS=64` and 8 spills, nothing has been evicted. Verified by forcing it — `ATLAS_SSM_TIER_SLOTS=2` produced **1,537,474,560 bytes** on disk (23 records). First disk write is spill number `SLOTS+1`.

## Enabling a 32 GB bounded NVMe tier

```bash
ATLAS_SSM_TIER=1 ATLAS_SSM_TIER_UNIFIED=1 \
ATLAS_SSM_TIER_SWAP_DIR=<nvme-dir> ATLAS_SSM_TIER_DISK_GB=32
```

Confirm from the log that the **O_DIRECT** arm was taken and not the silent host-RAM fallback (`blob_bytes` must be a 4 KiB multiple).

## Tests

`atlas-tier` **8 + 5 + 21 + 2** (was 8 + 4 + 21 — purely additive, nothing removed or weakened); `spark-model model::ssm` **73 passed**; release build clean.

## Known follow-ups (deliberately not in this PR)

- **`put_with(key, |&mut [u8]|)`** to land the D2H straight in the arena slot, removing the remaining 66 MB memcpy that is now 93% of spill cost (~19 ms → ~10 ms). It changes the `SnapshotBlobStore` trait and trades away the pinned destination, so it needs its own measurement — and at 19 ms against a ~28 ms fault-in, the tier already pays.
- **Stale tier-key reaping.** After a cap eviction the prefix entry keeps its `ssm_snapshot_tier_key`, so warm turns on a dropped key repeat spill → allocate → miss → free. Self-amplifying under sustained cap pressure. Fix is a mirror of the hit path's `promote_snapshot`, in the miss arm of `try_fault_in_ssm_snapshot`. **Operators enabling `ATLAS_SSM_TIER_DISK_GB` should expect degradation once the "cap engaged" warning fires**, until this lands.

🤖 Draft — measured on one box (GB10 / Holo-3.1-35B-A3B); numbers from other hardware welcome.


tbraun96:
Closing as already landed — but via #388 (`74442090`, the GB10 enterprise-concurrency work), **not** the #440 stack.

`gh pr diff 381` returns empty: after the sync merge this branch's tree is byte-identical to main, because #388 absorbed the eviction-cost work wholesale. Verified per-file — `ssm_spill_gate.rs`, `ssm_spill_staging.rs`, `atlas-tier/src/aligned.rs`, `cuda_backend/gpu_impl_graph.rs`, `prefix_cache/tier_evict.rs` and `atlas-tier/tests/residency_faults.rs` were all added by `74442090` and match this PR's head byte for byte.

Same class as #373/#375: merged elsewhere, nothing lost. Leaving the `perf/ssm-spill-eviction-cost` branch in place. Thanks @rsafier.

PR #382, opened 2026-07-29
fix(ssm-tier): reap dead tier keys so a capped disk cannot thrash

rsafier:
> **Stacked on #381** — base is `perf/ssm-spill-eviction-cost`, not `main`. Review #381 first; this diff is only the 9 files below.

Resolves both follow-ups #381 listed. One ships; **one is rejected, and that's the more useful result.**

## Shipped: reap dead tier keys

With `ATLAS_SSM_TIER_DISK_GB` set, the cap's own pressure was **self-amplifying**. After `make_disk_room` dropped a blob, the prefix entry kept its `ssm_snapshot_tier_key` — so it stayed findable by `lookup_tiered` with no bytes behind it. Every warm turn on that prefix re-ran the whole doomed cycle:

> spill a **live** 66 MB victim to free a slot → allocate → read → miss → free

…and each doomed spill evicted one **more** tier record, manufacturing the next dead key. #381 shipped a cap that degraded under exactly the pressure it creates; this closes that.

The miss arm of `fault_in_for_key` now retires the entry, so a dropped prefix recomputes **once**:

```rust
Ok(false) => { free(slot);
               if prefix_cache.forget_snapshot_tier_key(key) {  // index FIRST, under its lock
                   store.remove(key);                           // store SECOND, gated on it
               } }
Err(e)    => { free(slot); /* key RETAINED */ }
```

**Reap on `Ok(false)` only, never on `Err`.** An error is not evidence of absence: `Residency::locate` on a failed record read restores `disk_lru` and returns `Err` with the record still on disk **and still mapped** (`atlas-tier/src/residency.rs:269-291`). Reaping there would destroy a live 66 MB snapshot to save one retry. Retaining after a transient error costs one wasted cycle; reaping after one costs the prefix forever.

Ordering is load-bearing — index first, store second, gated on the index result. `forget_tiered` removes a **tiered** entry only: a resident entry's `snapshot_id` is a live pool slot that only its owner may free, so a by-key remove would leak it. That guard also makes the promote-then-reap race a clean no-op.

Also retires on spill **refusal** (`retire_refused_spill`) — `evict_to_tier` marks an entry `tiered` *before* the bytes are accepted, so a refused spill manufactures a stale key eagerly. `Err` at spill time keeps the key, symmetric with the above; the miss reap is its backstop.

New `tier_reaps` stat plus an INFO line naming the cause, since **sustained reaps mean `ATLAS_SSM_TIER_DISK_GB` is undersized for the working set** — that's the signal an operator needs.

## Rejected: `put_with`

#381 named a `put_with(key, |&mut [u8]|)` API to land the D2H straight in the arena slot, removing the 66 MB memcpy that is ~93% of the remaining 19 ms spill. **We evaluated it and are not shipping it.** The reasoning is recorded at the `store.put` site so it isn't re-litigated blind.

1. **Only 2 of 4 production `SnapshotBlobStore` implementors could support it.** `RdmaSnapshotStore` and `PagingSnapshotStore` have no host-addressable slot at all; even the supportable arm needs a new borrow-out method on `SlotArena` (which exposes only `read_slot`/`write_slot`). That's a permanent capability branch through the tier's one correctness-critical D2H, each arm exercised only on whichever backend an operator happens to run.
2. **It holds the residency `Mutex` across 60 `copy_d2h_async` enqueues plus a stream sync** — already documented as flag-ON blocker #1 at `ssm_tier/unified.rs:33-40`.
3. It trades away the pinned destination (`VecSlotArena::buf` is a plain heap `Vec<u8>`, and `atlas-tier` carries a hard no-GPU-deps constraint, so it can never pin). *Stated honestly:* the pinned side is measured (1.38–1.42 ms), the pageable side is **inferred** from the old H2D path which also paid a fresh alloc + zero-fill — so this bullet alone would not justify rejection. Bullets 1 and 2 do.

**More interesting finding, also recorded:** the 17–19 ms `store.put` may not be memcpy bandwidth at all. `VecSlotArena` allocates `vec![0u8; slot_bytes * num_slots]` (calloc ⇒ lazily faulted); 66,846,720 B is 16,320 fresh 4 KiB pages ≈ **16 ms of minor faults**, landing almost exactly on the observed number — and with the default 64 slots, *every put in a short run is a first touch*. If that holds, steady-state `store.put` is already cheap and #381's ~10 ms target needs **no code**.

This is an unmeasured hypothesis. The experiment that settles it: `store.put` wall time vs put ordinal 1..128 (`ATLAS_SSM_TIER_TIMING=1`, `ATLAS_SSM_TIER_SLOTS=64`, watch for decay). Worth doing before anyone touches the trait.

Also rejected and noted: pinning the arena from `spark-model` via `cuMemHostRegister` — the default hot arena is 64 × 66,846,720 B ≈ **4.3 GB**, and page-locking that on a UMA box is a worse trade than the ~17 ms it saves.

## Verifying reaping on GPU

```bash
ATLAS_SSM_TIER=1 ATLAS_SSM_TIER_UNIFIED=1 ATLAS_SSM_TIER_SWAP_DIR=<nvme> \
ATLAS_SSM_TIER_DISK_GB=<small, to force the cap> ATLAS_SSM_TIER_SLOTS=2
```

Grep `SSM tier reap:`. Success = reaps appear **and then stop** for a given prefix (retired once), rather than the same key cycling every warm turn.

## Tests

`atlas-tier` **8 + 5 + 21 + 2** unchanged; `spark-model model::ssm` **76** (was 73, +3). Release build clean, verified on this branch's own lineage.

🤖 Draft. Measured on GB10 / Holo-3.1-35B-A3B.


tbraun96:
Pushed one commit to this branch (`c7e58b17c`) — disclosing rather than opening a parallel PR.

### What I found while reviewing the pair

Reviewing #381 + #382 together (this branch already contains #381, so it's a proper stack and merges both), I went looking for the thing that would bite once the tier is actually reachable in anger. It's in `radix_tree/snapshot_insert.rs`, and it is **not yours** — that file is byte-identical on `main`, so this has been live since the tier landed in #317. I fixed it here because this is the PR that makes the path hot.

### The defect

A `tiered` entry's `snapshot_id` is **stale**. `evict_to_tier` hands that slot back exactly once (`TierEvict::Spill { slot, .. }`), the model frees it, and the entry survives in the index only as a fault-in record.

The codebase says so in three places, and every consumer honours it — `lookup` skips tiered entries ("no HBM slot to free"), and both victim scans pass `skip_tiered = true`.

The three insert paths did not. Overwriting a spilled entry returned its stale id for the caller to free a **second** time (`midchunk_capture.rs:311/329` → `ssm_snapshots.free(old)`).

`SsmSnapshotPool::free` is `free_slots.lock().push(slot)` with **no membership check** — unlike `ssm_pool::release_slot`, which at least `debug_assert!`s. So in release the double-free is silent, and the pool then hands one slot to two sequences, which share a GDN/conv state buffer. Wrong output, no error. Same class of fault `ssm_pool::claim_specific` exists to prevent on the KV side.

Reachability is ordinary warm traffic, not a corner: spill a prefix, then checkpoint that same prefix on a later turn. An agentic session with a stable system prompt does that every turn — which is the workload the tier exists for.

`insert_tail` carried a second one: it was the only overwrite arm that never cleared `tiered`. An entry left `tiered` while holding a **live** slot is skipped by `lookup` and by both victim scans, so that slot is reachable by nothing and freeable by nothing — a permanent leak of a scarce pool slot, on top of `lookup_tiered` faulting in bytes that no longer describe the entry.

### The fix

One predicate, `freeable_slot`, at all four displacement sites (plain insert, the tail supersede sweep, the tail overwrite, the sibling overwrite), plus the missing `tiered = false` on the tail arm.

### Run both ways

Six tests in `tests/snapshot_insert_tier.rs`. With the production change reverted and the tests kept:

| test | without the fix |
|---|---|
| `insert_over_a_spilled_entry_hands_back_no_slot` | **FAILED** — `left: Some(4)`, `right: None` |
| `insert_tail_over_a_spilled_entry_hands_back_no_slot` | **FAILED** |
| `insert_tail_sibling_over_a_spilled_entry_hands_back_no_slot` | **FAILED** — `left: Some(8)`, `right: None` |
| `insert_tail_rehomes_a_spilled_entry_to_hbm` | **FAILED** |
| `insert_over_a_resident_entry_still_hands_back_its_slot` | ok |
| `insert_tail_still_displaces_a_resident_tail` | ok |

The last two pass both ways deliberately — they pin the guard's **scope**. Without them, "return `None`" would be a trivially-passing fix that traded a double-free for a leak.

`spark-runtime`: 159 passed, 0 failed. fmt + clippy clean.

### On the other thing I'd flagged

I had this pair down as owing a second fix — "`SnapshotBlobStore::remove` has zero production callers". That was written against #381 alone and **#382 is the answer to it**: `remove` now has two callers, both with the index-first / store-second ordering spelled out, and the `Err`-is-not-evidence-of-absence reasoning in `retire_refused_spill` is right. I also walked the unbounded `MemBlobStore::new(0)` default and withdraw that concern too — it's opt-in behind `ATLAS_SSM_TIER`, it's the documented pre-existing legacy behaviour, and on the decode tier a cap would be a correctness bug (a dropped blob there is a lost rollback target), which `build_decode_tier_store` already argues explicitly.

Nice work on the reaping ordering in particular.

tbraun96:
Second push to this branch (`8c832bb25`), and it exists because an adversarial review found a hole in **my own** commit, not in yours.

### My "run both ways" evidence covered 3 of the 4 sites it claimed

`insert_tail` has two displacement sites — the sweep that clears this session's previous tail/sibling, and the overwrite of a matching `prefix_hash`. My `spilled()` helper built `is_tail: false`, so every test walked straight past the sweep. The one test that *does* drive the sweep used a **resident** entry, where the old and new code return the same value.

Verified rather than assumed — revert only `snapshot_insert.rs`'s sweep line:

```
### site B (tail supersede sweep) REVERTED — do the 6 tests notice?
... 6 passed; 0 failed
```

So the site whose failure mode is precisely the silent corruption the fix exists to prevent had a guard no test could distinguish from the bug. And it is reachable in the **default** configuration: `ATLAS_SSM_TAIL_MIDCHUNK` is default-ON, `session_aware_victim` spills a tail once its lease lapses or its session goes dormant, and the next `finalize_midchunk_capture` runs the sweep — every turn of every multi-turn conversation.

New test, both ways:

| | |
|---|---|
| without the fix | **FAILED** — `must not hand back slot 41 … got [41]` |
| with the fix | ok |

`spark-runtime`: 160 passed, 0 failed.

### And I had written a comment that was wrong

I claimed the orphaned tier blob was *"left to the store's own budget: on every capped arm it is the coldest thing there, and the next spill of this same prefix overwrites it in place."* That is false on **three of four** store arms, and I wrote it having read two of them:

- the **default** arm is `MemBlobStore::new(0)` — `cap_bytes == 0` means unbounded, so there is no budget to leave it to
- a capped `MemBlobStore` evicts **FIFO by insertion order**, not coldness — `store.rs` says so in its own comment
- `ArenaSnapshotStore`/`RdmaSnapshotStore` have **no eviction at all**; slots return only via `remove`

Only `UnifiedSnapshotStore` (default OFF) behaves as I described. Corrected in-tree, and the underlying leak — which is **pre-existing, not yours** — is filed as **#424** along with the observation that tiered index entries can never be evicted at all, so `entries` grows unbounded on the O(n) `lookup_tiered` path.

### On your side, two things I could not break and one comment worth fixing

**Your reaping tests are better than mine.** Three fail without the change, and they drive the production cycle rather than a hand-built index:

- `tier_miss_retires_the_key_instead_of_thrashing` — real `RadixTree` + `MemBlobStore` + `MockGpuBackend`, asserts `tier_attempts == 1`
- `spill_refusal_retires_the_entry_immediately`
- `test_spill_tier_lookup_transitions` (REAP branch)

`tier_error_retains_the_key` passing both ways is correct design — it pins the guard's scope, same role as my two resident-entry tests.

**Your `Err`-keeps-the-key policy is right** and I would not change it. It does guarantee you can reach a tiered entry with no blob behind it, but that is #424's missing reaper, not this decision.

**One comment to fix.** `ssm_snapshot_spill.rs:151-155` reads *"Ordering is load-bearing: index FIRST (under its lock), store second"*. The `RadixTree` lock is released when `forget_snapshot_tier_key` returns; nothing is held across `store.remove`. What makes it safe is that `prefix_cache` is a single-owner field moved into one scheduler thread — not the ordering. Harmless today, but it will mislead whoever adds a second worker, and the dangerous interleave is not the one it names: a concurrent `promote` is harmless (the entry is already gone), while a concurrent **spill** re-`put`ting under the same key between the two steps would delete a live blob.

Also noted in #424: `promote` (`snapshot_tier.rs:131-141`) matches `prefix_hash` without requiring `tiered` — the one displacement site still missing the guard, and the mirror of the bug this PR fixes. Unreachable today only because the model is single-threaded.

Next: this pair needs one gate run. Main is mid-way through re-establishing its own records right now (the benchmark gate had been rejecting every record via a `merge-base --is-ancestor` check, which is invalid in a squash-merging repo — fixed in #420); once those land I'll gate this and merge.

tbraun96:
Status, and one thing I tried and backed out of.

**I attempted the rebase onto `main` and aborted it.** Your branch is untouched at `8c832bb25`.

I checked the conflict scope first with `git merge-tree <base> HEAD origin/main`, counted zero `<<<<<<<` markers, and concluded it was a clean catch-up. **That check was wrong** — the three-argument form doesn't emit conflict markers; it's the older API with different output. The real merge produces **8 files, 16 hunks, ~184 lines**:

| file | hunks / lines |
|---|---|
| `spark-runtime/src/cuda_backend/gpu_impl_graph.rs` | 4 / 75 |
| `spark-model/src/model/ssm_snapshot_spill.rs` | 1 / 29 |
| `spark-runtime/src/prefix_cache/no_caching.rs` | 3 / 19 |
| `spark-runtime/src/cuda_backend/gpu_impl.rs` | 2 / 18 |
| `spark-model/src/model/ssm_snapshot.rs` | 3 / 17 |
| `spark-runtime/src/gpu.rs` | 1 / 17 |
| `spark-runtime/src/prefix_cache.rs` | 1 / 5 |
| `spark-model/src/model/mod.rs` | 1 / 4 |

I stopped because `gpu_impl_graph.rs` is CUDA-graph capture and `ssm_snapshot_spill.rs` is spill ordering. Resolving 16 semantic hunks there without being able to run a GPU smoke test afterwards is exactly how you introduce the silent state corruption this PR exists to fix. A resolution I cannot test is not a resolution.

### This also confirms the CI diagnosis

This PR currently reports **8 checks** — the CodeQL/security set. **None of the seven required ones** (fmt, clippy, test, typos, SPDX, 500-LoC, CLA) are running, because GitHub does not run merge-commit workflows for a PR it cannot merge. So the ticks here are not a verdict on this tree. #427 adds a `needs-rebase` label so this stops being invisible; a dry run found **thirteen** open PRs in the same state.

### Where that leaves it

The blocker is now precise: **not review, not tests — a rebase across #388's `gpu.rs` / `cuda_backend` changes**, by someone who can run a GPU smoke test after.

Everything else is done. Your reaping is right and has three tests that fail without it. My double-free fix has seven, including the tail-supersede-sweep case my first six missed. The pre-existing leak your PR sits next to is filed as #424 with the analysis, and the false bound I had written into a comment is corrected in-tree.

Happy to do the rebase if you would rather I did — say so and I will, with a GPU smoke test on the result before pushing. I did not want to do it silently on your branch after a check of mine turned out to be measuring the wrong thing.

rsafier:
@tbraun96 Thanks for the two commits and the disclosure — the double-free find is a good one.

I did the rebase (merge of main into both branches, no force-push) and it surfaced something worth stating plainly: **after the sync, #381's tree is byte-identical to main** — #388 absorbed the eviction-cost work wholesale (`copy_d2h_async`, the spill/staging split, the reap cycle, the `snapshot_reap` tests, all of it). So #381 is another #373/#375: merged elsewhere, not rejected — I'd close it as such.

**#382's entire post-sync residual is your two commits** (`c7e58b17` + `8c832bb2` — the `snapshot_insert` double-free fix and its tests, +259/−6 across 3 files). Local CI-parity: fmt clean, clippy `-D warnings` clean, 515 tests 0 fail.

GPU smoke on our GB10 (dgx-00), built at the merged tip, mlpinf 27B checkpoint, tier paths deliberately forced (`--ssm-cache-slots 4`, `ATLAS_SSM_TIER=1 ATLAS_SSM_TIER_UNIFIED=1`, spill/fault depth gates zeroed, 2 GB disk cap): 6 sessions × 2 warm turns, **12/12 correct**, spill shape line fired (`96 async D2H chunks + 1 stream sync into a reusable 158859264 B pinned staging buffer`), fault-in/spill events in the log, **zero errors/underflows**.

Given the residual content is yours, land it however you prefer — merge #382 as-is with the sync, or re-cut under your own name and close this one; either is fine by me. One nit I did NOT apply: the `Ordering is load-bearing` comment reword in `ssm_snapshot_spill.rs` — that text now lives on main via #388, so it belongs in a main-side follow-up if you still want it.

tbraun96:
Landed in main via the #440 stack merge (`3c941751`). Verified by content, not ancestry — Atlas squash-merges, so `--is-ancestor` would wrongly report this as unmerged.

All three files are byte-identical to this PR's head: `freeable_slot()` in `crates/spark-runtime/src/radix_tree/snapshot_insert.rs`, and the new `radix_tree/tests/snapshot_insert_tier.rs`. The full diff reverse-applies cleanly against main.

To close out the "two owed fixes" note that was tracked against this PR: those *are* this PR's current diff — `c7e58b17` (the stale-`snapshot_id` double-free in `snapshot_insert`) and `8c832bb2` (the site-B sweep hole its tests missed). Both are in main. Nothing is outstanding.

It showed MERGEABLE/BEHIND only because its base is `perf/ssm-spill-eviction-cost` rather than `main` — a stacking artifact. Thanks @rsafier.

PR #388, opened 2026-08-02
perf(gb10): enterprise concurrency — one config wins C=1..128 vs vLLM, env recipe replaced by CLI flags

tbraun96:
Enterprise-concurrency campaign for Qwen3.6-27B NVFP4 on GB10 (sm_121a). Merges `main` @ `c19481aa`.

**One serve command now wins every concurrency rung C=1..128 against vLLM on throughput**, and the ten-variable environment recipe is gone — every knob is a CLI flag.

Full campaign summary (what worked, what did not, lessons, bugs) is in the commit message. Measured tables in the first comment.

### Pipeline
`cargo fmt --check` clean · `cargo clippy --workspace --tests` clean · `cargo test --workspace` **2788 passed / 0 failed** · file-size cap 0 violations · SPDX clean

### Post-merge regression — all 8 rungs pass
Same box, same driver, vs the pre-merge reference ladder. No rung below reference outside the 0.5% noise floor.

### Notable
- `--ssm-h-dtype f16` without `--gdn-fused-norm` is now a **startup error** — that combination previously produced silent fluent garbage (FP32 kernel over an FP16 pool).
- Five env vars replaced by CLI flags; **six of the old ten were verified no-ops**.
- `ATLAS_GDN_FUSED_NORM` / `ATLAS_SSM_BATCHED_RECURRENT` deliberately **not** promoted to defaults — their bitwise gate was inconclusive because the control failed. Under PCND an uncertified numerics change stays explicit config.

Recipe: Avarok-Cybersecurity/atlas-recipes#14 (do not merge until an image carrying this commit is published — older images reject the new flags).


tbraun96:
## Throughput — one config, C=1..128 vs vLLM

Both engines, **one session, one box (dgx3), one driver**, `centml/Qwen3.6-27B-NVFP4-W4A4-mlpinf`, decode_short ISL 128 / OSL 1024, temp 0, token-matched, 3 reps.

| C | Atlas tok/s | vLLM tok/s | ratio |
|---:|---:|---:|---:|
| 1 | 24.20 | 14.29 | **1.694x** |
| 2 | 35.71 | 27.88 | **1.281x** |
| 4 | 73.06 | 53.40 | **1.368x** |
| 8 | 114.85 | 98.41 | **1.167x** |
| 16 | 205.10 | 170.18 | **1.205x** |
| 32 | 295.71 | 267.54 | **1.105x** |
| 64 | 380.89 | 373.15 | **1.021x** |
| 128 | 472.14 | 453.81 | **1.040x** |

Config: `--max-batch-size 128 --gpu-memory-utilization 0.85 --kv-cache-dtype bf16 --speculative --num-drafts 3 --ssm-h-dtype f16 --gdn-fused-norm --ssm-batched-recurrent --ssm-tail-midchunk false --mtp-gate force`

Speculation engages/disengages on batch width automatically (`mtp_max_seqs`, default 32) — C<=32 speculates, above it plain-decodes. No config switch needed.

## Post-merge regression — all 8 rungs pass

| C | 1 | 2 | 4 | 8 | 16 | 32 | 64 | 128 |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| ref | 24.20 | 35.71 | 73.06 | 114.85 | 205.10 | 295.71 | 380.89 | 472.14 |
| post-merge | 24.28 | 35.74 | 72.88 | 115.54 | 205.16 | 300.23 | 385.60 | 473.28 |
| delta | +0.33% | +0.08% | -0.25% | +0.60% | +0.03% | **+1.53%** | **+1.24%** | +0.24% |

C=4's -0.25% is inside the 0.5% noise floor (that rung's reps span 2.5%; it is under-resolved at 3 reps, not regressed).

## Time-to-answer (uncapped, natural stopping)

Throughput is not the whole story. Measured uncapped with matched **16-bit KV both sides**, thinking disabled on both, neither arm censored (`ct_eq_max=0`, `finish={stop:128}`):

| C | Atlas wall | vLLM wall | verdict |
|---:|---:|---:|---|
| 64 | 664.95 s | 652.18 s | Atlas **-1.96%** |
| 128 | 552.37 s | 522.19 s | Atlas **-5.78%** |

At C=128 Atlas is **+2.5% faster per token** (415.3 vs 405.2 tok/s) but emits **+8.5% more tokens** (1792 vs 1653): 1.025 / 1.085 = 0.945, matching the measured -5.8%. **The residual gap is token count, not engine speed.** Cause unexplained; it is not FP16 (removing it leaves token count unchanged, n=256, p=0.3164).

> An earlier measurement put these at -4.47% / -10.84%. That comparison was confounded: the **checkpoint** declares `kv_cache_quant_algo: FP8`, so vLLM's `auto` resolved to fp8 while Atlas ran bf16. Half the gap was KV dtype.

## Accuracy

| test | result |
|---|---|
| **BFCL, same checkpoint, both engines** (985 samples, ABBA) | Atlas 84.82 vs vLLM **86.30** — Atlas **-1.47 pts**, CI [-2.94, +0.00], p=0.0172 |
| ...localised | Gap is **abstention**, not construction: `live` ties (-0.33); deficit is `hallucination` (-3.35) / `irrelevance` (-6.25). Atlas over-calls tools where vLLM abstains. |
| **FP16 h-state accuracy gate** (984 paired/leg, C=4 + C=16) | Pooled discordant split **b=23, c=23, p=1.0000** — accuracy-neutral |
| **FP16 verbosity gate** (uncapped, paired) | +3.17% tokens spec-ON (n=56, p=0.030); **0% spec-OFF** (n=256, p=0.3164). The tax is speculation-mediated, not a precision property. |

⚠️ The BFCL comparison carries the same KV asymmetry as the speed one — vLLM ran checkpoint-declared fp8 KV. A matched-KV re-run is owed before this number is quoted externally.

## Everything done

- **Merge** — `main` @ `c19481aa`; 19 files / 24 hunks conflicted, resolved (main's `#385` converts ~105 statics to carried structs, exactly where this branch added statics)
- **Pipeline** — fmt / clippy / 2788 tests / LoC cap / SPDX all green; fixed 4 pre-existing reds incl. a test that could only ever fail under CI's stub build
- **env -> CLI** — `--ssm-h-dtype`, `--gdn-fused-norm`, `--ssm-batched-recurrent`, `--ssm-tail-midchunk`, `--mtp-gate`; **6 of the old 10 env vars were verified no-ops**
- **Safety** — `--ssm-h-dtype f16` without `--gdn-fused-norm` now refuses to start (previously: silent fluent garbage)
- **Kernel wins** — 64-wide N tile on the deep-K GEMM; `mmq_x=64` rung (tile ladder stepped 16->32->128, so batches 33..127 discarded half their MMA columns); both bit-identical (327,680 and 9.4M outputs, 0 diffs)
- **Bugs fixed** — 300s deadline never enforced on the spec path and mislabelled truncation as `finish_reason=length`; `--mtp-quantization nvfp4` silently disabled speculation (-46.7% at C=1); bearer tokens accepted without `--require-auth`; `ATLAS_SSM_H_FP16` decoded in two places so `=0` read as ON; `serve.rs` printed the SSM dtype before the model config loaded
- **Not shipped, deliberately** — `ATLAS_GDN_FUSED_NORM` / `ATLAS_SSM_BATCHED_RECURRENT` stay CLI flags, not defaults: their bitwise gate was inconclusive because the **control failed** (two identical serves differed on 7/42 completions)
- **Known open** — `finish_reason=length` emitted ~3000 tokens below `max_tokens`; `--disable-tool-grammar` does not gate `response_format`; the +8.5% token excess vs vLLM at C=128

tbraun96:
### Correction to the accuracy table above

The Atlas BFCL figure (84.82 overall / 84.23 `normalized_single_turn`) reads ~3 points below this project's historical 87.24. That was investigated and it is **not a regression** — verified by re-scoring both runs' raw outputs through a single scorer:

**On the 879 samples common to both draws, one scorer:**

| run | `normalized_single_turn` |
|---|---:|
| historical run — *the source of 87.24* | 86.84 |
| **this campaign's Atlas** | **87.21 / 87.28** |

Atlas is **+0.37 ahead** of the historical run on a like-for-like basis.

The apparent gap is two things, and one of them is that **87.24 was inflated**:

| step | NST | Δ |
|---|---:|---|
| historical published (984 of a 1004 draw) | 87.24 | — |
| same run, the 20 excluded samples restored | **85.53** | **−1.71** |
| this campaign's Atlas, same 879 samples, same scorer | **87.21** | **+0.37** |
| this campaign's Atlas on its own 985 draw | 84.22 | −2.99 (the draw) |

1. **The historical run served `--max-seq-len 4096`**, so 20 samples failed with `Prompt too long: 4270 > 4096` and were **dropped from the denominator**. This campaign ran 8192 and scored 985/985. **87.24 / 87.35 / 87.33 are A/B denominators within that run, not quotable absolutes.**
2. **The draw got harder** — `hallucination` moved from 12% to 20% of the mix. The draw is `head(n)`, so the 90 added samples are all `live_irrelevance`/`irrelevance`, where head samples score 92-96 and tail samples score **64-69** — for *both* engines.

Refuted empirically: the scorer reproduced the historical stored aggregate **bit-for-bit across all 12 subset scores**; the two parquets differ in md5 but are **content-identical** (same 3641 rows, identical SHA256 over `sample_id`/`ground_truth`/`messages`/`tools`) — do not infer a different dataset from a parquet md5. Config and checkpoint identical.

**The Atlas-vs-vLLM comparison in the table above is unaffected** — one session, one draw, ABBA-ordered. vLLM still leads on the 879 basis too.

★ Going forward every BFCL number needs its `category_sample_pct`, N, and ordered-`sample_id` SHA256 recorded beside it.

SeedSource (commented):
CI is fully green, but the posted runtime tables do not qualify current head `e11d6ae0`, and the comparative BFCL result remains KV-confounded. Before merge, please rerun the claimed ladder and bounded quality gate on the exact current head, with immutable image/checkpoint/harness identifiers and raw artifacts. If #389 is merged into this stack, rerun after that merge.

eisbaw:
Bug report from a **GB10 / DGX Spark** test of this branch on the **35B MoE flagship** (`RedHatAI/Qwen3.6-35B-A3B-NVFP4`, `--lm-head-dtype nvfp4`, CUDA 13.2, built from source): two hard failures that don't reproduce on `main`, so flagging before this merges.

**1. C≥8 concurrency crashes 100% (main handles C=32 fine).** `w55-squash` merged onto `main` @ `3e5115e8` (clean textual merge, sha `6ae930a` locally): single-session is fine (96.3 tok/s, parity), C=4 works (+5% vs main), but every request at C≥8 fails:

```
Inference error: cuMemcpyDtoHAsync_v2 failed: status 716
free_sequence: ssm_pool.zero_slot(3): cuMemsetD8Async failed: status 716
free_sequence: gpu.synchronize after zero_slot(3): CUDA_ERROR_MISALIGNED_ADDRESS (716)
```
…repeating for slots 3..7. Reproduces with default flags and with explicit `--max-batch-size 32`. Looks like an SSM slot stride/alignment issue for slots ≥3 on the 35B (the campaign validated the 27B). Same-session `main` control: C=1/4/8/16/32 = 93.1/122.6/149.0/151.1/150.4 tok/s, no errors.

**2. `--ssm-h-dtype f16 --gdn-fused-norm --ssm-batched-recurrent --max-batch-size 32`: server loads, every decode fails.**

```
decode_batch error: Kernel launch failed: cuLaunchKernel failed:
CUDA_ERROR_INVALID_HANDLE (400): invalid resource handle (grid=[32,1,1], block=[128,1,1])
```
The `gated_delta_rule_decode_f16_*` twins appear in the compiled-kernel list, but the launch handle is invalid for this MoE target — so the recipe's biggest lever (+13–20% at C≥16 on the 27B) isn't currently reachable on the 35B.

Also for the record: `--max-batch-size 128` at default `--gpu-memory-utilization` preflights a 73.5 GB inference reserve on the 35B and refuses to start (fine on the 27B; just worth a note in the recipe).

Env: `docker/gb10/Dockerfile.builder` image (+ the CUDA-13.2 fixes from #396–#398), `ATLAS_CUTLASS_NVFP4_GEMM=1`, `ATLAS_GDN_FLASHINFER=0`. Full logs and configs available on request. (Testing and analysis run with AI assistance — Claude Code.)

eisbaw:
Three quick follow-ups from more GB10 testing:

1. Turns out `main` itself crashes on >32-row decode batches (sticky CUDA-700 at `--max-batch-size 64` on the 35B) — the old metadata layout only reserves 32 slots, so the seq_lens upload corrupts the KV slot pointers above that. Your `DecodeMetaLayout` + serve-time validation fixes exactly this, so that's one more reason to land this PR once the 35B C≥8 `zero_slot` issue above is sorted.

2. Quality data for the fused-norm + batched-recurrent pair (which the PR notes left non-default after the bitwise control failed): we ran a paired gsm8k (n=200, same questions, greedy, thinking on) on the 35B with the env-var versions on main — strict-match 0.980 vs 0.900 baseline, and its few wrong answers were a strict subset of baseline's. Not bitwise proof, but end-to-end it looks safe on the 35B. Perf-wise the pair only pays off together: fused alone ≈ parity, batched-recurrent alone −7%, both +8% (190 tok/s aggregate at C=32 stacked with #332/#404).

3. Small heads-up: #404 also touches `ssm_batched.rs`, which this branch restructures — whoever lands second needs a trivial rebase; the changes are complementary.

Logs/configs available as before.

tbraun96:
CI note: \`CLAAssistant\` had been failing at "Set up job" on this PR — before evaluating any signature — with:

> The action \`contributor-assistant/github-action@v2.6.1\` is not allowed in Avarok-Cybersecurity/atlas because all actions must be pinned to a full-length commit SHA.

That is the repo's \`sha_pinning_required\` setting rejecting the tag reference, so the red was a workflow-config problem rather than an unsigned CLA. Because the workflow runs via \`pull_request_target\` it always executes **main's** copy, so this PR could not fix its own blocker.

Pinned on main as \`8f0c3be\` (\`ca4a40a7d1004f18d9960b404b97e5f30a505a08\`, the commit \`v2.6.1\` points to — same code, now SHA-pinned). This comment re-triggers the check against the corrected workflow.

tbraun96:
recheck

tbraun96:
recheck

PR #389, opened 2026-08-03
perf(load): GPU-transpose quantized weights at cold load (-44%) — stacks on #388

rsafier:
Cherry-pick of the cold-load transpose fix from `perf/enterprise-concurrency-v3` (6c5e8c3e) onto #388's head, offered as a stack-on since the Serve Matrix pays the old cost on every checkpoint boot.

## What

`transpose_for_gemm_gs` and the concat/pitched `transpose_impl` bounced every packed NVFP4 weight + scale through host — `copy_d2h` → O(N·K) host byte loop → `copy_h2d`, ~13.6 GB at 27B cold load. Both now run on-device via the existing `transpose_u8` tiled kernel (already used batched by the MoE path); the pitched concat writes each part with one `copy_d2d_2d_async` (`cudaMemcpy2DAsync`), pad columns memset to zero to match the host path's zeroed staging buffer.

No call-site signature changes: the kernel resolves internally via `try_kernel`; targets without it silently keep the host path; `ATLAS_HOST_TRANSPOSE=1` forces the host path (parity/debug kill switch). Launch arity pinned in `kernel_arity.rs`.

## Measured (on perf/enterprise-concurrency-v3, qwen3.6-27b NVFP4, warm JIT, same binary both modes)

| mode | docker-run → ready | temp-0 probe SHA |
|---|---|---|
| host byte-loop (`ATLAS_HOST_TRANSPOSE=1`) | 66 s | `cde424ae52a6983a` |
| GPU transpose (default) | **37 s (−44%)** | `cde424ae52a6983a` — identical |

Identical greedy-probe hashes across modes = bit-equivalent transposed weights.

On this branch: clean cherry-pick (auto-merge), clippy clean, arity pin passes. Not yet re-timed on #388's head — the win applies per checkpoint boot, so the Serve Matrix multiplies it by the roster size.


rsafier:
Re-timed on this branch's head ('54fd21e0' + this pick), same box (dgx-00 GB10), interactive TUI boot of the 27B NVFP4: **39 s to serving** with the GPU transpose path, vs the ~66 s host-byte-loop class. Matches the −44% measured on the origin branch within box noise. Applies per Serve Matrix checkpoint boot.

rsafier:
Stacked two more picks from the origin branch plus one adapted hygiene fix — this PR is now "the things #388's lineage measurably lacks":

1. **`637370a3` — fp8-KV decode attention kernel fix** (kernel-file only). The conversion-bound per-element cvt chain made fp8-KV decode ~2.3x slower than bf16 e2e (measured 4.8 -> 2.1 tok/s/stream at C=8 agentic on the origin branch; 2.1 -> **5.6** after the fix, beating bf16). Urgency here: this branch's 27B MODEL.toml sets `default_kv_dtype = "fp8"`, so default serves eat the slow kernel — the ladder in #388's tables dodged it only by passing `--kv-cache-dtype bf16` explicitly.
2. **`0e74657d` — propose_meta stride from max_seq_len** (+ scheduler fallback demoted to debug). This tree's fixed 2048 stride caps the batched MTP propose at ~7.2K tokens of context; agentic serving lives at 10-20K, where the origin branch measured **+11-17% per-stream decode** from keeping the batched path engaged. Conflict resolution: kept this branch's `PROPOSE_META_SEQS = 32` (the 32:1 rung's stash width) under the dynamic stride; all 6 stride unit tests pass against it, including the hostile-env-override overflow guard at 32x.
3. **`31a1b971` — CUTLASS_NVFP4_ROUTE warn -> debug**, keeping this branch's per-backend OpCache dedup (nicer than the origin's static): the dedup key includes M, so agentic prefill spammed the WARN channel per request, and the probe ran per routed GEMM call even with nothing to log.

Branch state: fmt clean, clippy clean, stride + arity tests green, release build OK on gb10.

kernels/gb10/common/paged_decode_attn_fp8.cu
```diff
@@ -39,20 +39,30 @@ __device__ __forceinline__ void unpack2_bf16(unsigned int packed, float& v0, flo
...
-    __nv_fp8_storage_t b0 = (__nv_fp8_storage_t)(packed & 0xFF);
-    __nv_fp8_storage_t b1 = (__nv_fp8_storage_t)((packed >> 8) & 0xFF);
-    __nv_fp8_storage_t b2 = (__nv_fp8_storage_t)((packed >> 16) & 0xFF);
-    __nv_fp8_storage_t b3 = (__nv_fp8_storage_t)((packed >> 24) & 0xFF);
-    // FP8 → half_raw → float, then multiply by dequant scale
-    v0 = __half2float(__nv_cvt_fp8_to_halfraw(b0, __NV_E4M3)) * dq_scale;
-    v1 = __half2float(__nv_cvt_fp8_to_halfraw(b1, __NV_E4M3)) * dq_scale;
-    v2 = __half2float(__nv_cvt_fp8_to_halfraw(b2, __NV_E4M3)) * dq_scale;
-    v3 = __half2float(__nv_cvt_fp8_to_halfraw(b3, __NV_E4M3)) * dq_scale;
+    __half2_raw h01 = __nv_cvt_fp8x2_to_halfraw2(
```
SeedSource, line 56: Blocking: this common kernel is mirrored into the HIP build. The current Windows AMD-HIP job fails because `__nv_cvt_fp8x2_to_halfraw2` is unavailable there. Please use the existing portability abstraction or conditional intrinsic and rerun the full matrix.

kernels/gb10/common/paged_decode_attn_fp8.cu
```diff
@@ -112,6 +122,11 @@ extern "C" __global__ void paged_decode_attn_fp8(
     if (my_end > seq_len) my_end = seq_len;
     if (my_start > seq_len) my_start = seq_len;
 
+    // Hoisted dequant scales (see unpack4_fp8_raw): raw-K dots get
+    // k_scale here, once per position; raw-V o_reg gets v_scale once at
+    // the smem write below.
+    const float score_scale = inv_sqrt_d * k_scale;
```
SeedSource, line 128: The throughput result establishes speed, not correctness. Paired conversion plus moving K/V scaling changes floating-point evaluation in both normal and split-K paths. Please provide old/new GPU parity across representative sequence lengths and scales—max error/tolerance plus token-level quality or hashes—on the current #388 head.

crates/spark-model/src/layers/mtp_head/batch_caps.rs
```diff
@@ -39,19 +39,74 @@ use atlas_core::config::ModelConfig;
...
 
+/// Pure stride computation: header + one i32 block-table entry per KV block a
+/// `max_seq_len`-token drafter sequence can reference, 8-byte aligned, never
+/// below [`PROPOSE_META_STRIDE_FLOOR`].
+///
+/// The `+ 1` mirrors the allocator in `forward_batch_position`
+/// (`blocks_needed = seq_len/bs + 1`): without it, a sequence sitting exactly
+/// at `max_seq_len` needs one more entry than `align8(256 + (max/bs)*4)`
+/// provides and the stride `ensure!` fires at the boundary.
+pub(crate) fn propose_meta_stride_bytes(max_seq_len: usize, kv_block_size: usize) -> usize {
```
SeedSource, line 73: The unit tests validate stride arithmetic, but not the claimed runtime behavior. Please add a current-head 10–20K spec-on run showing batched propose remains active, no overflow/fallback occurs, outputs remain equivalent, and the claimed decode gain transfers.

crates/spark-server/src/scheduler/mtp_bootstrap_step.rs
```diff
@@ -41,6 +41,39 @@ pub(super) fn bootstrap_batch_disabled() -> bool {
...
+/// Discrimination is a string match on the single producing `ensure!`
+/// ("exceeds meta stride", spark-model mtp_head/forward_batch.rs) rather than
+/// a typed error: a pub error type would couple the scheduler to mtp_head
+/// internals across crates for one message, and the match follows the
+/// existing `{e:#}`-contains precedent (decode_step.rs "KV cache exhausted",
+/// phase_start_prefills.rs "pool exhausted"). Matching the alternate `{e:#}`
+/// form keeps it robust to `.context()` additions upstream.
+pub(super) fn log_propose_batched_err(prefix: &str, e: &anyhow::Error) {
+    let msg = format!("{e:#}");
+    if msg.contains("exceeds meta stride") {
```
SeedSource, line 70: After dynamic sizing, an overflow in a normal configuration is an invariant/configuration failure, not routine telemetry. Demoting every string-matched overflow to DEBUG hides permanent fallback. Please retain at least a rate-limited WARN/ERROR unless the undersized stride was explicitly requested by override.

crates/spark-model/src/weight_map/quantized.rs
```diff
@@ -318,6 +365,62 @@ impl QuantizedWeight {
...
+        // columns `n_total..stride` are zeroed by the memset up front,
+        // matching the host path's zeroed staging vec.
+        if let Some(tk) = Self::host_transpose_kernel(gpu) {
+            let new_weight = gpu.alloc(stride * half_k)?;
+            let new_scale = gpu.alloc(stride * num_groups)?;
+            if stride > n_total {
+                gpu.memset(new_weight, 0, stride * half_k)?;
+                gpu.memset(new_scale, 0, stride * num_groups)?;
+            }
+            let mut temps: Vec<DevicePtr> = Vec::with_capacity(parts.len() * 2);
```
SeedSource, line 381: An identical single greedy-probe hash is not sufficient to claim bit-equivalent transposed weights. Please add direct host-vs-GPU buffer comparisons for single, concatenated, padded-stride, non-32-multiple, weight, and scale layouts; also report peak load memory or avoid retaining all concat temporaries until the final sync.

SeedSource (changes requested):
This PR has expanded from one load-time transpose optimization into four independently risky behaviors: transpose, FP8-KV decode math, dynamic MTP metadata sizing, and logging. The current Windows AMD-HIP matrix also fails, and the runtime evidence is from an older #388 base/origin branch. Please split the changes—or fully update the title/body—bring them onto current #388, fix the portability regression, and attach current-head correctness and runtime qualification for each behavior.

eisbaw:
Confirmed on a GB10 (DGX Spark, sm_121a, CUDA 13.2, built from source), this branch as-is:

`spark serve RedHatAI/Qwen3.6-35B-A3B-NVFP4 --port 8888 --lm-head-dtype nvfp4`

Cold start to /v1/models ready: **30 s** default vs **46 s** with `ATLAS_HOST_TRANSPOSE=1` — −35% on the 35B, same binary. Decode parity: 96.55 vs 96.42 tok/s median (5x 256-tok non-stream), both coherent. Nice win.

tbraun96:
@rsafier — a disclosure of what I've pushed to this branch, since it's grown well past your original diff and you should be able to see the boundary clearly.

## Your work, untouched in substance

The GPU-transpose-at-cold-load change and the runtime `propose_meta_stride` are intact. One merge conflict needed resolving in `mtp_head/forward_batch.rs`, where `w55-squash` had moved the same bounds check into a shared `pack_mtp_attn_meta`. I resolved it as **both**, not either: the shared packer keeps the single bounds check and is now handed your runtime stride instead of the old compile-time constant. Taking the other side wholesale would have re-imposed the 4K-era 448-entry bound and undone your fix.

★ That merge would otherwise have shipped a silent regression. `scheduler/mtp_bootstrap_step.rs` demotes the overflow to debug by matching `msg.contains("exceeds meta stride")`, but the shared packer raised `"does not fit"`. It compiles either way — and puts a per-occurrence ERROR back in production logs for an expected, recoverable condition, which is exactly the noise your change set out to remove. The phrase is restored, the coupling documented at both ends, and the packer's own test now asserts the phrase rather than incidental wording.

## What I added on top

Rebased onto `main` after #388 landed (Atlas squash-merges, so the branch needed main merged in or the diff would have re-shown all of #388).

**To measure your change at all:** every gate recipe pins `kv_cache_dtype: bf16`, so no gate could exercise the fp8-KV path this PR touches. Added `--serve-override KEY=VALUE`, and made recipe overrides able to *add* a key — no recipe mentions `fp8_kv_calibration_tokens`. The fp8 run passed 10/10 · 10/10 with Σwall 841 s against a 1300 s budget.

**Gate scoping** (`gate/coverage.rs`): invalidation was one bit — any touch under `crates/` re-opened all five gates, including both ~3.5 h BFCL legs. Now per-gate, with written rationales required for every exclusion. It closes a real hole while it's there: `3rdparty_patches/` was in no watched path, so **swapping the dlopen'd `libatlasgdn.so` invalidated nothing**, on a config claiming +17-20% on GDN prefill.

To be explicit: this does **not** let this PR off. It touches `paged_decode_attn_fp8.cu` and `fp8_moe.rs`, so all five gates are still owed, and there's a test pinning exactly that.

**Also:** a journey ledger crate (`atlas-governance`) on LatticeDB, a reusable categorizer action, and a book chapter. Those are additive and reviewable separately — happy to split them into a follow-up PR if you'd rather this stayed focused on the load path. Just say so.

## Verification

fmt · clippy · `cargo doc` all clean, **65 test suites** green, 500-LoC cap respected. The five GPU gates are running now at `fe5a4f67`; I'll post the table when they finish.

Nothing here was force-pushed — it's all additive commits on your branch.

tbraun96:
All five required gates pass at `b0be4ba0e6`, measured on the engine merged with #417's bounded grammar cache.

    ttft-warm-gate       median +0.0%  p90 -0.2%   (limits +3% / +5%)
    ttft-cold-gate       median +0.1%  p90 +0.1%
    agentic-webserver    10/10 webserver_ok · 10/10 followed_directions · 965s ≤ 1300s
    bfcl-subset          87.64 / 88.59   n=995    (floors 83.64 / 85.32)
    bfcl-subset-echolp   86.55 / 86.95   n=1004

`bfcl-subset-echolp` is the one worth reading twice. Before #417 that leg could not finish at all: host RSS grew ~100 MB/min against ~8 GB of headroom, and the protective guard killed it at 269/1004. It now completes, and RSS was observed to FALL mid-run (2200 → 2088 MB) as the LRU evicted — the first eviction seen in production rather than in a unit test.

Gate D reproduced 87.64 / 88.59 exactly, per-subset identical to the pre-fix run, so bounding the cache changed no output.

## What this PR adds

**A closure-hash cascade** so a `kernels/` change re-tests the targets whose device code actually changed, instead of all 28. The obvious design — hash each target's resolved `.cu` set — is unsound twice, and both were found by reading the tree: shadow files `#include` the very file they shadow (8 of them), and headers are in no `.cu` set at all. Both are now tests that fail against a set-hash. The attestation is baked by `build.rs` into the binary, not computed from the working tree, because the tree and the binary differ exactly when it matters.

**`BENCH.toml`** — thresholds move next to the model they describe, per checkpoint. Three schema rules stop a threshold file lying: `unmeasured` entries may not carry metrics, `measured` entries must, and exactly one checkpoint per (gate, hardware) is the default.

**A PR telemetry plane** — the cross-PR view a per-PR check cannot have: two PRs touching one target are each green against a baseline the other moves.

Full rationale in `book/src/project/merge-lattice.md`.

PR #394, opened 2026-08-04
fix(dflash): use drafter target_layer_ids directly (accept +1 row/step on z-lab drafters)

rrstesiak:
## What

Part 1 of 5 of the dense 27B DFlash records series (36.8 tok/s single-stream, +74% over MTP; matrix in the team post).

Uses the drafter's `target_layer_ids` directly as capture-layer indices in `factory/build.rs`. The previous -1 adjustment and its `ATLAS_DFLASH_CAPTURE_LAYER_OFFSET` env override are both deleted (per project direction: measured wins ship as the only path, no switches). One file, net -10 lines.

## Why

This is the single largest accept lever we have measured on the z-lab drafters, and with the current default every DFlash deployment is giving up roughly one accepted row per step until this merges.

- 27B dense, MinHeap workload, same binary, boot-line A/B: **mean accept 6.61 direct vs 5.56 shifted**.
- The 2026-07-19 35B accept investigation independently reached the same conclusion before we understood the mechanism.
- On the 27B records config this default is worth several tok/s on its own.

The -1 was deliberate (aligning to HF `output_hidden_states` indexing so the drafter sees the layer the training pipeline nominally fed it), but at -1 the drafter is fed hidden states one layer early relative to what it actually conditions on best. Whatever the indexing rationale, accept is the ground truth for what the drafter wants.

## Risk

Behavior change is the point: accept goes up on both drafters we can test (27B dense, 35B MoE). The measurement receipts stay in the code comment; re-testing the old behavior is a one-commit revert.

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -30,6 +30,10 @@ extend-ignore-identifiers-re = [
 # reads it as a misspelling of `RETURNED`, which it is not.
 RETUNED = "RETUNED"
 retuned = "retuned"
+# `unauthenticate` — used as a verb ("would silently unauthenticate the
+# endpoint") in the model-swap test commentary (#385). Intentional coinage,
+# not a typo of "unauthenticated".
+unauthenticate = "unauthenticate"
 # `nd` — the num-drafts speculative-decode notation (`nd=2`, `nd=3`) used
 # throughout the MTP verify code and docs. Not a typo of "and".
 nd = "nd"
--- a/crates/spark-model/src/factory/build.rs
+++ b/crates/spark-model/src/factory/build.rs
@@ -124,30 +124,20 @@ pub fn build_model(
     // capture-layer indices from the drafter's `dflash_config.target_layer_ids`
     // so `TransformerModel::new` allocates the 5×hidden_size capture buffer.
     //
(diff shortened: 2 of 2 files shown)


rrstesiak:
CI note: the typos failure was pre-existing, not from this diff. The action is pinned to master, its dictionary drifted, and it now flags "unauthenticate" (used as a verb,  intentionally, in the #385 test commentary). Allowlisted it in _typos.toml with a comment so it reads as prose to the checker, since every PR against current main hits the same red X until this lands.

rrstesiak:
NOTE: As expected, performance as of this initial PR is well below the published numbers; the bulk of the DFlash code has yet to be submitted. Once all of the PRs are in, final optimal performance becomes verifiable, minus the ~11% main-side accept regression documented to the team on 2026-08-04 (unrelated to this series), until that drift is resolved.

tbraun96:

Verified by content, not by ancestry (Atlas squash-merges, and this branch predates the repo's re-root at "Open Source Release", so it has no merge base with `main` and a three-dot diff is meaningless here): `gh pr diff 394 | git apply --check -R -` applies cleanly against `main` @ `a64b82978`, i.e. every hunk of this PR is already present in the tree.

PR #397, opened 2026-08-05
docker/gb10 builder: fix CuTe-DSL runtime discovery for cutlass-dsl 4.5.0

eisbaw:
nvidia-cutlass-dsl 4.5.0 ships `libcute_dsl_runtime.so` under `nvidia_cutlass_dsl/lib/`, but the current discovery globs under the `cutlass` python subpackage — a sibling tree that never contains it — so it returns empty and the build aborts at the CuTe-DSL step. Swap the package-relative glob for a plain `find`.

Benefit: builder works with current cutlass-dsl instead of failing on it.

Tested on GB10 / CUDA 13.2.0-devel. Found and fixed with an AI agent (Claude Code).


eisbaw:

eisbaw:
recheck

tbraun96:
Update from a maintenance sweep: I approved the workflow runs on this PR (fork PRs sit in `action_required` until a maintainer does) and updated the branch onto current `main`.

That cleared the real failures — the earlier `cargo fmt`, mdBook and 500-LoC-cap reds were branch staleness, not anything in your diff. **The only remaining blocker is the CLA signature** (`CLAAssistant`), which is on your side; nothing further is needed from the code.

Thanks for the builder fixes — happy to take another look once the CLA clears.

tbraun96:
Thanks for this — verified and adopted in #445, with credit to you as co-author on the commit.

We reimplemented rather than merged directly only because the CLA check is a required context here and this PR is blocked on it. The originals are left open; if you sign the CLA we are happy to revisit.

Confirmed broken on the pinned base image: `find_spec('cutlass').origin` resolves to `.../nvidia_cutlass_dsl/python_packages/cutlass/__init__.py`, but the .so lives in the **sibling** `nvidia_cutlass_dsl/lib/` — so the module-relative glob can never see it, and the build dies at `test -n "$CUTE_RT"`. Your `find` locates the one real copy. Adopted verbatim.

tbraun96:
Landed on main via #445 (2d069d6b), with you credited as co-author on the commit. Closing this one as superseded — the change is in, so there's nothing further gated on the CLA here.

Thanks again; the builder was genuinely broken on all three counts and this unblocked it.

PR #398, opened 2026-08-05
docker/gb10 builder: set CUDARC_CUDA_VERSION for CUDA 13.2

eisbaw:
On CUDA 13.2 the vendored cudarc build.rs panics: `Unsupported cuda toolkit version: 13.2`. Your CI already exports `CUDARC_CUDA_VERSION=13000` to work around this; the builder Dockerfile just doesn't. Set it on the cargo build step.

Benefit: builder compiles on CUDA 13.2 toolkits without the cudarc panic.

Tested on GB10 / CUDA 13.2.0-devel. Found and fixed with an AI agent (Claude Code).


tbraun96:
Update from a maintenance sweep: I approved the workflow runs on this PR (fork PRs sit in `action_required` until a maintainer does) and updated the branch onto current `main`.

That cleared the real failures — the earlier `cargo fmt`, mdBook and 500-LoC-cap reds were branch staleness, not anything in your diff. **The only remaining blocker is the CLA signature** (`CLAAssistant`), which is on your side; nothing further is needed from the code.

Thanks for the builder fixes — happy to take another look once the CLA clears.

tbraun96:
Thanks for this — verified and adopted in #445, with credit to you as co-author on the commit.

We reimplemented rather than merged directly only because the CLA check is a required context here and this PR is blocked on it. The originals are left open; if you sign the CLA we are happy to revisit.

Confirmed: vendored cudarc 0.19.2's version table tops out at `(13,1)`, and we build with `cuda-version-from-build-system`, so on the image's 13.2 toolkit the build script panics with `Unsupported cuda toolkit version: \`13.2\``. `CUDARC_CUDA_VERSION=13000` clears it — and it is what all six of our workflows already pin, so this was a real gap between CI and the builder image. Adopted verbatim.

tbraun96:
Landed on main via #445 (2d069d6b), with you credited as co-author on the commit. Closing this one as superseded — the change is in, so there's nothing further gated on the CLA here.

Thanks again; the builder was genuinely broken on all three counts and this unblocked it.

PR #404, opened 2026-08-06
perf(ssm): keep batched QKVZ/out_proj engaged for n>16 NVFP4 decode (≤16-row tiling) — C=32 aggregate +36% on GB10

eisbaw:
## What

The NVFP4 arm of `try_decode_multi_seq_ssm_batched` gates on `n <= 16` (`ssm_batched.rs`). Above that, the whole batched SSM path returns `false` and every GDN layer falls back to the per-seq loop — the QKVZ and out_proj weights get re-streamed once per sequence, which un-does exactly the amortization this path exists for.

This PR relaxes the eligibility gate and tiles the two batched GEMVs (QKVZ in_proj, out_proj) into ≤16-row chunks of the existing `w4a16_gemv_batch16` kernel: weights read `ceil(n/16)`× instead of `n`×. For `n <= 16` the call is bit-for-bit the same single dispatch as before (one chunk, unchanged kernel selection). The recurrent inner and both norm ends are untouched; chunk offsets use the same per-row strides the recurrent inner already indexes with (`h`, `qkvz_size`, `value_dim` × bf16).

## Measured (GB10 / DGX Spark, sm_121a, CUDA 13.2, built from source)

`spark serve RedHatAI/Qwen3.6-35B-A3B-NVFP4 --port 8888 --lm-head-dtype nvfp4 --max-batch-size 32`, 256-tok non-stream requests, distinct sequences, coherence-checked, same-session A/B vs `main` @ `3e5115e8`:

| aggregate tok/s | C=8 | C=16 | C=32 |
|---|---|---|---|
| main | 155.92 | 164.35 | **130.84** ← cliff |
| main + this PR | 162.77* | 163.47 | **178.07** (+36%) |

Single-session parity: 96.24 vs 96.40 median. (*C=8 rung is noisy across all our runs, 156–173.)

Composes with #332 and the `ATLAS_GDN_FUSED_NORM=1 ATLAS_SSM_BATCHED_RECURRENT=1` pair: **190 tok/s aggregate at C=32, 188 at C=64** (batch-32 waves) on the same setup — vs ~151 stock best.

## Notes / limits

- NVFP4 path only; the w8a16/dense arms already had no `n <= 16` limit (GEMM).
- `--max-batch-size 64` with real >32-row batches hits a separate pre-existing kernel limit (`CUDA_ERROR_ILLEGAL_ADDRESS`, grid=[64,9,2]) — out of scope here; batch 32 is the validated ceiling for this path.
- Validated on the 35B MoE only; a 27B-dense run would be a good maintainer cross-check.
- Found while profiling open PRs on GB10; the same cliff reproduces with #332's lm_head batched-GEMV (its `n<=16` gate is the right call — at n=32 the plain GEMM lm_head outperforms tiled GEMV, 178 vs 173 in our runs).

Developed and benchmarked with AI assistance (Claude Code); measurements are from real hardware runs as described.

### DIFF
--- a/crates/spark-model/src/layers/qwen3_ssm/trait_decode_multi_seq/ssm_batched.rs
+++ b/crates/spark-model/src/layers/qwen3_ssm/trait_decode_multi_seq/ssm_batched.rs
@@ -35,7 +35,7 @@ impl Qwen3SsmLayer {
         // (batch4/16, M<=16). Either amortizes the QKVZ/out_proj weight read
         // across the n seqs; otherwise the per-seq loop re-streams it n times.
         let qkvz_ok = (self.qkvz_nvfp4.is_none() && self.w8a16_gemm_k.0 != 0)
-            || (self.qkvz_nvfp4.is_some() && self.w4a16_gemv_batch4_k.0 != 0 && n <= 16);
+            || (self.qkvz_nvfp4.is_some() && self.w4a16_gemv_batch4_k.0 != 0);
         let out_ok = self.out_proj_fp8w.is_some()
             || self.out_proj_dense.is_some()
             || self.qkvz_nvfp4.is_some();
@@ -179,17 +179,25 @@ impl Qwen3SsmLayer {
         } else if let Some(ref nvfp4) = self.qkvz_nvfp4 {
             // FP4 batched QKVZ: ONE NVFP4 weight pass for all n seqs
             // (sequential layout writes the deinterleaved buffer directly).
-            ops::w4a16_gemv_batchm(
-                ctx.gpu,
-                fp4_gemv_batch_k,
-                normed_base,
-                nvfp4,
-                deinterleaved,
-                n as u32,
-                qkvz_size as u32,
-                h as u32,
-                stream,
-            )?;
+            // TASK-160 (gx10): tile n>16 into <=16-row chunks so the batched
+            // path (weights read ceil(n/16)x) survives decode batches >16
+            // instead of disengaging to the per-seq loop (weights read n x).
+            let mut c0 = 0usize;
+            while c0 < n {
+                let m = (n - c0).min(16);
+                ops::w4a16_gemv_batchm(
+                    ctx.gpu,
+                    fp4_gemv_batch_k,
+                    normed_base.offset(c0 * h * bf16),
+                    nvfp4,
+                    deinterleaved.offset(c0 * qkvz_size * bf16),
+                    m as u32,
+                    qkvz_size as u32,
+                    h as u32,
+                    stream,
+                )?;
+                c0 += m;
+            }
         } else {
             ops::dense_gemm(
                 ctx.gpu,
@@ -297,17 +305,22 @@ impl Qwen3SsmLayer {
             // FP4 batched out_proj: ONE NVFP4 weight pass for all n seqs.
             // (qkvz_nvfp4.is_some() ⇒ the NVFP4 SSM build, where ssm.out_proj
             // is the NVFP4 weight the per-seq path also uses via w4a16_gemv.)
-            ops::w4a16_gemv_batchm(
-                ctx.gpu,
-                fp4_gemv_batch_k,
-                normed_out_base,
-                &self.ssm.out_proj,
-                ssm_out_base,
-                n as u32,
-                h as u32,
-                value_dim as u32,
-                stream,
-            )?;
+            let mut c0 = 0usize;
+            while c0 < n {
+                let m = (n - c0).min(16);
+                ops::w4a16_gemv_batchm(
+                    ctx.gpu,
+                    fp4_gemv_batch_k,
+                    normed_out_base.offset(c0 * value_dim * bf16),
+                    &self.ssm.out_proj,
+                    ssm_out_base.offset(c0 * h * bf16),
+                    m as u32,
+                    h as u32,
+                    value_dim as u32,
+                    stream,
+                )?;
+                c0 += m;
+            }
         }
         detail_step!("out_proj");
 


eisbaw:
> Thanks for the contribution. Before we can merge, please sign our [Contributor License Agreement](https://github.com/Avarok-Cybersecurity/atlas/blob/main/CLA.md) by replying to this comment with exactly:
> 
> 
> I have read the CLA Document and I hereby sign the CLA
> 
> You can retrigger this bot by commenting **recheck** in this Pull Request. Posted by the **CLA Assistant Lite bot**.

tbraun96:
Heads-up while batching gate cycles: this looks **superseded by main**, and I dropped it from the batch rather than hand-resolve it.

Both changes solve the same problem — keeping the batched QKVZ/out_proj projections engaged above n=16 instead of falling back to the per-seq loop. They solve it differently:

- **this PR**: chunk the wide batch into <=16-row tiles and keep using the batchm GEMV family (`ceil(n/16)` launches)
- **main today**: route wide batches to the tile-GEMM twins (`w4a16_gemm_n128` on `qkvz_nvfp4_t`/`out_proj_nvfp4_t`), which handle any M in one launch, gated by `tc_wide_ok` + `ssm_tc_proj_min_n()`

Evidence: `origin/main:…/trait_decode_multi_seq/ssm_batched.rs` has 7 references to `tc_wide_ok`/`ssm_tc_proj_min_n`; this branch has 0, and its base predates them. Merging produces three conflict hunks in exactly those dispatch sites.

I did **not** try to resolve them. Picking between two live implementations of the same dispatch is your call, not something to guess at from the outside — a wrong resolution here silently selects a slower path with no test that would notice.

If main's tile-GEMM route covers your case, this can probably close. If your `<=16-row tiling` still wins somewhere main's route does not — narrow shapes where the tile GEMM's M-padding wastes more than the extra launches cost, say — that would be worth a measured A/B, and the concurrency sweep now runs to C=32 on both dense and MoE (#461), so there is a standard curve to show it on.

tbraun96:
Triage note, with the measurements behind it — this one is not a mechanical rebase, and I would rather say so than quietly re-cut it wrong.

**Where it stands.** Against `main` at `ce2bccd736` this PR conflicts, and on exactly one file: `crates/spark-model/src/layers/qwen3_ssm/trait_decode_multi_seq/ssm_batched.rs`. That makes it the cheapest-looking re-cut in a 56-PR conflicting backlog, which is why I looked at it first.

**Why it is not actually cheap.** The branch point is `3e5115e855`, **249 commits** behind `main`, and that single file has been substantially rewritten since:

```
$ git diff --stat 3e5115e855 origin/main -- .../ssm_batched.rs
1 file changed, 156 insertions(+), 198 deletions(-)      # against a 344-line file
```

Six commits touched it, and one of them is squarely in this PR's territory:

- `f8fcd346a` — *spark-model: preserve scalar accumulation order in FP8 batched GEMV and **batch native o_proj** (#932) (#1027)*
- `8a68f2602` — the `--hermetic` exact-equality work (#981)
- `fbca9ca5f` — the C=128 ladder and the γ>8 unlock (#818)
- `0a56d8f8b`, `60370b953`, `569ffd092` (the rebrand)

So the batched QKVZ / out_proj path this PR keeps engaged for n>16 is not the path that existed in August. **The +36% C=32 aggregate claim needs re-measuring before the change is worth re-cutting** — some of it may already be realised by #1027, and the ≤16-row tiling boundary may not sit where it did.

**What would move this forward**, in the order that costs least: re-measure C=32 aggregate on today's `main` first; if the gap is still there, re-cut onto the current file rather than rebasing the old hunks. It touches `crates/**`, so it needs a full certification campaign either way (~3 GPU-hours), and those are currently serialised behind #1144 and #582.

No action taken on the branch. Happy to do the re-measure if you would like it.

tbraun96:
Closing as **superseded** — and this is the owner-side check that was standing
open against it, now actually run rather than deferred again.

The goal here ("keep batched QKVZ/out_proj engaged for n>16 NVFP4 decode") is
already on `main`, and has been since **#388** (`7444209045`). Replaying this
commit today does not add the behaviour — it **reverts** it.

`crates/spark-model/src/layers/qwen3_ssm/trait_decode_multi_seq/ssm_batched.rs`
on main:

```rust
let tc_wide_ok = self.qkvz_nvfp4_t.is_some()
    && self.out_proj_nvfp4_t.is_some()
    && ssm_tc_proj_min_n().is_some_and(|min| n >= min);
let qkvz_ok = (self.qkvz_nvfp4.is_none()
    && ((self.qkvz_fp8w.is_some() && self.w8a16_gemm_k.0 != 0)
        || (self.qkvz_fp8w.is_none() && self.dense_gemm_k.0 != 0)))
    || (self.qkvz_nvfp4.is_some()
        && ((self.w4a16_batchm.has_base() && n <= 16) || tc_wide_ok));
```

`ssm_tc_proj_min_n()` defaults to **`Some(9)`**, so the batched tensor-core
projection path engages at **n ≥ 9** — a *lower* threshold than the n>16 this PR
targets, via transposed NVFP4 weights rather than ≤16-row tiling. This PR's
version of the same line is

```rust
let qkvz_ok = (self.qkvz_nvfp4.is_none() && self.w8a16_gemm_k.0 != 0)
    || (self.qkvz_nvfp4.is_some() && self.w4a16_gemv_batch4_k.0 != 0);
```

which drops `tc_wide_ok` entirely, and also drops main's separate handling of
the pure native-BF16 build (`qkvz_fp8w.is_none()` → `dense_gemm_k`), which is
what keeps that build on the batched path instead of the per-seq loop. A second
hunk would likewise revert the `hc` guard on the batched input RMS norm.

The branch is 266 commits behind, and all three conflicts are of this shape —
main moved past it rather than around it. Nothing here is wrong; it was simply
solved upstream first. Thanks @eisbaw — if you have a measurement where main's
n≥9 tensor-core path still loses to ≤16-row tiling on NVFP4, that is a different
and very much open question, and worth its own issue with the numbers.

PR #395, opened 2026-08-05
perf(gdn): general-K fused GDN verify kernels (K=4..17) + accept-EWMA gate

rrstesiak:
## What

Part 2 of 5 of the dense 27B DFlash records series (36.8 tok/s single-stream, +74% over MTP; matrix in the team post).

Two commits:

1. **General-K fused verify kernel family.** Atlas ships fused GDN verify kernels for K in {2,3,4,17} only; every other width falls to the sequential per-token loop (~34 launches per layer across 30 SSM layers). One `K_TOKENS`-templated module (`gated_delta_rule_wyk.cu`, wy17-derived, bit-exact at K=4 and K=17 by construction) instantiates every width 4..17, which unlocks the mid-width range where the measured dense-27B optimum actually lives (DRAFT_CAP=11, single-peaked sweep across 4/9/11/16).

2. **Accept-EWMA gate, no switch.** The fused path is workload-conditional by measurement: +10% to +34% tok/s on code/reasoning content (mean accept 4.4-8.4), -12% on low-accept prose (accept ~2.0), because at low accept the full-K fused launch does more work per harvested token than the loop. No static default is right for both, so the verify step feeds each step's accepted-row count into an EWMA and the dispatch follows the measured crossover: engage at 4.0, release at 2.5, 16-step dwell, strictly predictive (consulted before the step's own accept exists). Per project direction the env toggle from the first commit is deleted; the engine picks the path itself.

## Why

This is the second largest lever in the series after the capture-layer default (#394). Receipts, 27B dense, length-matched pairs, healthy clocks:

- MinHeap (code/reasoning): 32.1 -> 36.8 tok/s with the fused path engaged.
- fib: 35.1, razzle: 32.8, both records with the gate engaged.
- Volvo (prose): 18.7 tok/s with the gate correctly staying on the sequential path (fused forced on costs -12% here). The gate exists precisely because of this split.
- Gate dynamics validated live in both directions: engages ~step 33 on code content from cold, no flapping through mixed-content seams. Transition tax ~1% on steady code, ~5% across seams.
- Original fork receipt on MoE 35B: cap4 79.1 -> 94.0 steady (+19%), same kernel body.

## Numbers on current main (disclosure)
 
  Three notes so nobody benches mid-train and files a ghost:

[...]

PR #405, opened 2026-08-06
perf(dflash): copy-path efficiency: pitched 2D Q/K/V scatter + batched SSM state copies

rrstesiak:
## What

Part 3 of 5 of the dense 27B DFlash records series (36.8 tok/s single-stream; matrix in the team post). This is the prose-flank PR of the train: its yield lands where speculation earns least, and the receipts below say so honestly.

Two commits:

1. **Q/K/V scatter loops collapsed into pitched 2D copies.** The three batch paths in `multi_seq/qkv.rs` scattered contiguous Q/K/V scratch into the interleaved per-seq buffer with 3×rows separate `copy_d2d_async` calls. Each family is a contiguous source into a fixed-stride destination, i.e. exactly a pitched copy, so each collapses into one `cudaMemcpy2DAsync`. 3×rows driver calls become 3. Profiled motivation (nsys, 573 steps): small D2D copies were 297/step costing 2.140 ms of driver time against 0.309 ms of GPU time, the largest host-side term in the step. Bit-identical to the loops: same bytes, same addresses, same stream, same family order.

2. **Per-layer SSM state copies batched into one kernel launch.** The DFlash/MTP checkpoint path issued one `cuMemcpyDtoDAsync` per SSM layer per copy family (148 driver calls/step measured). The pools are one allocation per layer with a fixed `base[layer]+slot*bytes` layout, so per-layer base pointers are constant for the model lifetime: uploaded once at pool construction, one launch covers every layer. Guarded per call by a pool-provenance pointer check with fallback to the loop; kill switch `ATLAS_SSM_BULK_COPY=0`; a one-shot log line ("SSM bulk copy: batched path ACTIVE") makes the engaged path verifiable from any boot log, since the fallback is silent by design.

## Why

Copy-path host overhead is per-step, so it costs most exactly where speculation harvests fewest tokens per step: prose. That makes this the lever for DFlash's weakest workload, and the receipts are deliberately split by workload:

- MoE 35B fork receipts (nsys-profiled origin of both commits, length-matched pairs): Volvo non-adaptive 58.7 -> 61.2 cumulative across the two commits (+4.3%), with the batched-SSM half alone measuring +3.6% (a record for the forced-spec arm at the time). MinHeap (code): +0.86%, under the 2% measurement floor, NOT claimed as a win, per-step savings amortize away at high accept.
- Dense 27B on current main, this branch alone: MinHeap 27.3 tok/s steady, i.e. parity with the rest of the mid-train (code-workload neutrality, exactly as the mechanism predicts). Volvo: 14.6 tok/s steady (the prose lever on dense is pending the batched-SSM engagement; see disclosure note 4).
- Both commits were additionally field-proven bit-exact during the 07-30 output-corruption bisect: with the batched copies toggled off, outputs were byte-identical to the toggled-on control, and both commits were active in every clean rung.

## Numbers on current main (disclosure)

Same three notes as #395, plus one specific to this PR:

1. **The headline numbers require the FULL five-PR series** (capture-layer default #394, general-K kernels #395, this PR, and two still queued). Benching main after any single PR merges will NOT reproduce the records, by design of the rollout, not by defect.
2. **Current main carries a ~11% accept regression** relative to the pre-drift base (documented to the team 2026-08-04; unrelated to this series). Receipts, same box, same protocol, re-confirmed 2026-08-06: mean accepted rows 5.02/step on current main vs 5.71 on the pre-drift base at identical per-step cost; the identical full stack measures 33.2 tok/s MinHeap on current main vs 37.3 on the pre-drift base. The cost lands hardest where the margin is thinnest: full-stack Volvo (prose) is 17.2 on the pre-drift base vs 16.5 on current main, meaning the drift erases DFlash's prose-over-serial crossover entirely and returns prose to serial parity. Isolation done to date: the DFlash kernels are exonerated by per-step timing on both trees, and the known request-index trajectory fork is present byte-identically on BOTH bases (same fork offset) and is therefore not the cause. The regressing change is one of the 27 upstream commits between merge-base 56aa136e and current main 3e5115e8. Full repro harness (worktree, launch script, accept-probe protocol) available on request.
3. **Benched up to and including this PR on current main (2026-08-06, GB10, steady-state second request):** 27.3 tok/s MinHeap / 14.6 tok/s Volvo.
4. **The batched-SSM path is currently inert on dense 27B as configured:** the pool-provenance guard falls back to the per-layer loop (the new ACTIVE log line makes this visible). The guard exists precisely so the fast path only engages when provably safe; diagnosing why the dense pool fails the check is follow-up work. The commit remains correct and dormant-safe either way: guarded, kill-switched, and bit-exact when engaged (MoE receipts above).

### Reproduction recipe (DENSE Qwen3.6-27b)

Env (dev flags on current main; several become defaults as this series lands):

ATLAS_TARGET_MODEL=qwen3.6-27b \
ATLAS_DFLASH_CAPTURE_LAYER_OFFSET=0 \
ATLAS_DFLASH_SERIAL_APPEND=1 \
ATLAS_DFLASH_SEAM_SERIAL=1 \
ATLAS_DFLASH_OPTION_B=1 \
ATLAS_DFLASH_EAGLE_FIX=1 \
ATLAS_DFLASH_MASKED_VERIFY=1 \
ATLAS_DFLASH_SPEC_THINK=1 \
ATLAS_DFLASH_DRAFTER_FP8=1 \
ATLAS_DFLASH_DRAFT_CAP=11

Serve: `spark serve <Qwen3.6-27B-NVFP4> --dflash --draft-model <z-lab/Qwen3.6-27B-DFlash> --dflash-gamma 17 --max-seq-len 16384 --max-batch-size 1 --kv-cache-dtype bf16 --scheduling-policy slai`

Protocol: single stream, quote the second request (first is warmup). GB10, max clocks. Without this env, accept collapses and DFlash measures below serial; the flags are load-bearing until their values land as defaults later in this series.

**NOTE: I am working on collapsing the most amount of ENV switches/flags possible into the DFlash code by the end of the PR sequence for DFlash.**

### Reproduction recipe (MoE 35B-A3B, for the fork-era receipts above)

  Build note: `ATLAS_TARGET_MODEL` is BUILD-time for MoE, the binary must be compiled with it, not just launched:

  ATLAS_TARGET_MODEL=qwen3.6-35b-a3b \
  cargo build --release

  Env (matches the A/B arms behind the MoE numbers above):
 
  ATLAS_TARGET_MODEL=qwen3.6-35b-a3b \
  ATLAS_DFLASH_CAPTURE_LAYER_OFFSET=0 \
  ATLAS_DFLASH_SERIAL_APPEND=1 \
  ATLAS_DFLASH_SEAM_SERIAL=1 \
  ATLAS_DFLASH_OPTION_B=1 \
  ATLAS_DFLASH_EAGLE_FIX=1 \
  ATLAS_DFLASH_MASKED_VERIFY=1 \
  ATLAS_DFLASH_SPEC_THINK=1 \
  ATLAS_GDN_WYK=1 \
  ATLAS_DFLASH_SWA_CAUSAL=1 \
  ATLAS_CUBLASLT_DENSE=1 \
  ATLAS_DFLASH_ADAPTIVE=0 \
  ATLAS_DFLASH_DRAFT_CAP=4
 
  Serve: `spark serve <Qwen3.6-35B-A3B-NVFP4> --dflash --draft-model <z-lab/Qwen3.6-35B-A3B-DFlash> --dflash-gamma 17 --max-seq-len 16384 --max-batch-size 1 --kv-cache-dtype bf16
  --scheduling-policy slai`

DUPLICATED NOTE FOR THIS SCENARIO:  **NOTE: I am working on collapsing the most amount of ENV switches/flags possible into the DFLASH code by the end of the PR sequence for DFLash. 

## Risk

- Both copy rewrites are bit-identical to the loops they replace: same bytes, same source/destination addresses, same stream, same ordering. Field-proven byte-identical in the 07-30 bisect (toggle-off vs control).
- The batched-SSM path only engages behind a per-call pointer-provenance check; any non-pool allocation (lazy checkpoint allocs) falls back to the existing loop. Kill switch: `ATLAS_SSM_BULK_COPY=0`.
- The Q/K/V 2D copies have no behavioral gate; they are the same copy expressed as one driver call per family instead of one per row.
- No kernel-math changes anywhere in this PR; the only new device code is a strided memcpy kernel.


tbraun96:

Landing commit: `60370b95` — *"perf: Atlas beats vLLM at every concurrency C=1..128 on Qwen3.8-27B"* — **verified on `main`** (merge commit of #572).

**One honest note on the form it landed in.** The capability is on `main`, but not as the file this PR proposed. `kernels/gb10/common/ssm_state_bulk_copy.cu` (`extern "C" __global__ void ssm_state_bulk_copy`) is **not** on `main`; the batched SSM state copies live there instead as **`crates/spark-model/src/model/ssm_batched_copy.rs`** (179 lines) with `ssm_batched_copy_tests.rs` (8 tests) beside it.

So the idea shipped and is covered by tests — it was just re-expressed on the Rust side rather than as a standalone CUDA kernel. Flagging it because "landed" and "landed as written" are not the same claim, and the record should say which.

The six `.benchmarks/*-afc7a3113c.json` records this PR carried are also absent, superseded by later records.

PR #411, opened 2026-08-08
feat(spark-server): advertise max_model_len in /v1/models

begarzus:
## Problem
OpenAI-compatible clients that auto-detect context length (e.g. Hermes, LangChain, LiteLLM) read the model metadata returned by `GET /v1/models`. Atlas was not reporting a context/max-length field, so such clients fell back to a hardcoded catalog entry. For a Qwen model that means **131072** — missing the **262144** window the server was actually launched with — breaking long-context sessions.

## Fix
Advertise `AppState.max_seq_len` (the context window the server was started with) as `max_model_len` on every model listed by `GET /v1/models` (list + retrieve). Field is `Option<u64>` with `skip_serializing_if = "None"` so it stays backward compatible if ever constructed without state.

Files:
- `crates/spark-server/src/openai/chat_response.rs` — add `max_model_len: Option<u64>` to `ModelInfo`
- `crates/spark-server/src/api/models.rs` — populate it from `state.max_seq_len` in all 3 `list_models` sites + `get_model`

## Verification
Built a patched image locally on DGX Spark GB10 (SM121); `GET /v1/models` returns `"max_model_len": 262144` and the client auto-detects the full window.


tbraun96:
Perfect, thank you! Please go ahead and accept the CLI, then we can go ahead and merge right away.

tbraun96:
Status update from a maintenance sweep — this PR is close, and the remaining blocker is on your side only.

I approved the workflow runs (fork PRs sit in `action_required` until a maintainer does) and updated the branch onto current `main`. Six of the seven required contexts now pass.

**The only thing left is the CLA** (`CLAAssistant`). We're currently batching several bug fixes into a single stack so they share one GPU gate cycle rather than four, and this change would have been a natural fit — but we won't carry a contribution into a branch we author without the contributor agreement in place. That's a licensing matter rather than a CI one, and `CLAAssistant` is a required status check precisely so it can't be routed around.

Sign it and this becomes a one-line merge. Thanks for the fix — advertising `max_model_len` in `/v1/models` is the right call for client compatibility.

tbraun96:
Superseded — `max_model_len` is now advertised on `/v1/models`, landed via #444 (e509d1fd).

Your diagnosis was right and the field is exactly what clients (LiteLLM, aider, Continue, OpenWebUI) read to size a request. We implemented it with two changes on top of your version, which is why this is closing as superseded rather than merged:

1. **Derived, not stored.** The value comes from `AppState::max_seq_len` through a single constructor `ModelInfo::advertise`, so the advertised ceiling cannot drift from the one admission actually enforces. A hardcoded literal is now caught by rustc, since `max_seq_len` would become an unused parameter.
2. **Unknown is omitted, not zero.** With no model loaded the field is absent rather than serialized as `0` — a client sizing against 0 would refuse to send anything at all, which is a worse failure than not knowing.

It shipped with tests covering the positive case, the omission case, and a rename guard. Thanks for raising it.

PR #416, opened 2026-08-08
Fix responses API reasoning compatibility

simonsoanes:
## Summary

While building a continuous agentic cognition platform I tested your inference server and found issues with the reasoning capabilities of Qwen 3.6, it seemed to lose track of its internal approach and behave less intelligently than it should (though it was faster on a spark than other inference servers).  On debugging I found Atlas wasn't returning sufficient reasoning traces on the Responses API even when thinking was enabled in the sparkrun recipe.

## Test plan

- [v] Tested on dual and single DGX sparks running Qwen 3.6 32b-a3b in thinking mode by building a new container and running the recipe

## Notes for reviewers

Built using Qwen 3.8-max under my own agentic platform.

## CLA

- [v] I have read and agree to the [Contributor License Agreement](../CLA.md).


simonsoanes:

simonsoanes:
The copilot comment observations are real issues, fixing them and re-testing.

tbraun96:
Awesome, thanks @simonsoanes! I'm in the middle of a holistic benchmarking/repo-telemetry system, so some pipeline will fail. I'll go ahead and merge before the Qwen 3.8 27B release!

tbraun96:
Landed on main via #444 (e509d1fd) — the responses-API reasoning work is in: summary_text streaming, top-level `reasoning_effort` and `thinking_budget`, re-opened thinking, and reasoning persisted to conversations.

All ten files are subsumed; the only residual difference is that main additionally carries the `max_model_len` field from #411's line of work.

It was gated as part of that stack — five legs, all PASS, records committed. Thanks for the contribution.

Issue #368, opened 2026-07-24
Non-streaming requests leak scheduler slots when the client disconnects; --request-timeout does not reap them


dknoodle:
**Environment:** `avarok/atlas-gb10` @ `sha256:1f05897ba082737a613de7c65ef32e80972b61368ba14f3f5a86cc6d0463e18f` (image created 2026-06-12), `spark serve Qwen/Qwen3.6-35B-A3B-FP8`, `--max-num-seqs 12 --max-seq-len 262144 --kv-cache-dtype fp8 --speculative`, non-streaming OpenAI-compatible `/v1/chat/completions` on an NVIDIA GB10.

## Bug

spark appears to have no client-disconnect detection on non-streaming requests. When a client aborts mid-request (in our case a gateway-side timeout), the request permanently occupies a scheduler slot: `atlas_requests_active` never decrements, and that request never generates another token. Orphans accumulate monotonically until all `--max-num-seqs` slots are exhausted and the server stops serving new work — while `/health` continues to report `{"status":"ready"}`, so the wedge is invisible to a standard health probe.

## Repro

1. Send a non-streaming chat completion with a long prompt.
2. Kill the client ~5s in (`timeout 5 curl ...`).
3. `atlas_requests_active` stays at 1 indefinitely; `atlas_generation_tokens_total` does not advance for it.
4. Repeat until active ≥ `max-num-seqs` → new requests queue forever.

## Evidence from production

Measured 2026-07-24 on an otherwise idle server:

- `atlas_requests_active` held at exactly `2` across a 45s window with `atlas_generation_tokens_total` **and** `atlas_requests_total` completely unchanged — i.e. two slots occupied by requests that were not generating and would never complete.
- Those slots then grew to `6` over the next ~3h of normal agent traffic: a leak rate of roughly **1.4 slots/hour**, strictly monotonic between restarts.
- Earlier incident: leaked `0 → 248` active over 6 days against `--max-num-seqs 12`, ending in a total wedge.
- Currently the container must be restarted by a watchdog roughly every 7 hours. Over one 48h window that was 7 restarts.

The distinguishing signal is useful and might also suggest a fix: **an orphaned request never emits a token**, so "active but `generation_tokens_total` frozen" cleanly separates leaked slots from live work.

## `--request-timeout` does not reap them

Tested at `--request-timeout 240` on the image above:

- A disconnected orphan was **not** reaped — it sat well past the 240s deadline with zero tokens generated.
[...]

PR #417, opened 2026-08-08
fix(xgrammar): bound the compiled-grammar cache — closes the #368 heap growth

tbraun96:
Closes the heap-growth half of #368. rsafier's #374 fixed the `atlas_requests_active` gauge; this is the separate defect underneath it, and it is the one that costs operators a restart every ~7 hours.

## Root cause

`crates/xgrammar/src/compiler/compiler.rs` held the Tier-1 compiled-grammar cache as a bare `DashMap<CacheKey, CompiledGrammar>` with **no eviction**. `cache_limit_bytes` was recorded and never enforced — the C++ `ThreadSafeLRUCache` was not ported, while the Tier-2 `RuleLevelCache` was. `compiler/mod.rs` said so outright:

> it has no LRU byte-budget eviction (the C++ `ThreadSafeLRUCache`). `cache_limit_bytes` is recorded and [not enforced]

That is unbounded in the dimension that matters for a server. `CacheKey::Schema` is keyed by the **full tool-schema text**, and agentic traffic sends a distinct schema on nearly every request, so every request minted a permanent entry.

## Evidence

Measured on a GB10 serve under BFCL, process identified:

| | t0 | t+120s |
|---|---|---|
| `spark` RSS | 1,493,732 KB | 1,698,296 KB |
| `MemAvailable` | 6205 MB | 6056 MB |
| Slab / SUnreclaim / PageTables | 2807088 / 1974356 / 40680 kB | 2806820 / 1974088 / 40616 kB |

**~100 MB/min, ~17 MB/request.** Kernel memory flat → userspace heap. `atlas_requests_active` sat at 1 while `requests_total` advanced, so this is not the gauge defect.

Corroborating: the same teardown sweep reports **1,479** unreleased allocations after the 27B BFCL leg and **33,110** after the 35B one — a 22× spread tracking how many distinct grammars each workload compiles.

Operationally it made a 3.5 h accuracy benchmark unable to finish on a box that started with ~8 GB of headroom.

## The fix

`GrammarCache` — byte budget, LRU list, evict-until-fits — mirroring the Tier-2 cache already in this crate. Two details are specific to this tier and are why it is not a copy:

- **Entries grow after insertion.** A `CompiledGrammar` owns a `mask_cache` filled lazily as that grammar is *used* (XGrammar-2 JIT), so a size captured at insert time understates it by orders of magnitude — a freshly compiled grammar has an empty mask cache. The budget is re-measured across live entries on each insert, using the existing `memory_size_bytes()`. Inserts happen only on a cache miss, so that walk is never on the hot path.
- **The newest entry is never evicted**, even if it alone exceeds the budget. The caller is about to use it; dropping it would trade a memory bound for a correctness bug. It leaves on the next insert, once superseded.

Dropping a grammar drops its mask cache with it, which is what makes bounding this tier sufficient to bound both.

## Behaviour change

`grammar/engine.rs` passed `-1` (unlimited). It now passes a **1 GiB** budget — roughly 60 distinct schemas at the measured footprint, far above any realistic hot set. `-1` remains available as an explicit opt-in for callers that genuinely want every grammar pinned.

## Risk

Evicting a genuinely hot schema forces a recompile, so **warm TTFT is the gate to watch**. Gates are being run on this branch; results will be posted here before merge.

## Tests

`grammar_cache_tests.rs`: unbounded accumulation is bounded; `-1` stays unlimited; a recent *read* protects an entry (least-recently-**used**, not inserted); an oversized newest entry is retained then evicted once superseded; re-inserting a present key refreshes rather than duplicates. Plus a guard that fails if the real eviction loop changes shape without the modelled policy following it.

822 xgrammar + full workspace green; clippy 0; fmt clean.

### DIFF
--- a/crates/spark-server/src/grammar/engine.rs
+++ b/crates/spark-server/src/grammar/engine.rs
@@ -2,6 +2,14 @@
 
 use xgrammar::{GrammarCompiler, TokenizerInfo, VocabType, detect_metadata_from_hf};
 
+/// Byte budget for the Tier-1 compiled-grammar cache (`GrammarCompiler`).
+///
+/// 1 GiB. Sized against the measured footprint of a compiled tool grammar plus
+/// the masks it accumulates in use (~17 MB under BFCL), so this holds roughly
+/// sixty distinct schemas — far beyond any realistic hot set, while bounding a
+/// tail that was previously unbounded. See issue #368.
+const GRAMMAR_CACHE_BUDGET_BYTES: isize = 1024 * 1024 * 1024;
+
 use super::extract_ordered_vocab;
 
 // ── GrammarEngine ──────────────────────────────────────────────────────
@@ -132,7 +140,14 @@ impl GrammarEngine {
     fn from_tokenizer_info(tokenizer_info: TokenizerInfo) -> Result<Self, GrammarError> {
         let vocab_size = tokenizer_info.vocab_size();
         // Single compilation thread, cache enabled, no memory limit.
-        let compiler = GrammarCompiler::new(&tokenizer_info, 1, true, -1)
+        // ★ NOT -1 (unlimited). `CacheKey::Schema` is keyed by the full tool
+        // schema, so agentic traffic mints a distinct compiled grammar per
+        // request; unbounded, that grew host RSS ~100 MB/min under BFCL and is
+        // the shape behind issue #368's ~7-hour restart cadence. A GiB is far
+        // above any realistic hot set (tens of schemas) while bounding the
+        // tail. `-1` remains available as an explicit opt-in for callers that
+        // genuinely want every grammar pinned.
+        let compiler = GrammarCompiler::new(&tokenizer_info, 1, true, GRAMMAR_CACHE_BUDGET_BYTES)
             .map_err(GrammarError::Compilation)?;
         Ok(Self {
             compiler,
--- a/crates/xgrammar/src/compiler/compiler.rs
+++ b/crates/xgrammar/src/compiler/compiler.rs
@@ -6,8 +6,6 @@
 // A tokenizer-bound cache that produces `CompiledGrammar`s, avoiding
 // redundant preprocessing of grammars / schemas seen before.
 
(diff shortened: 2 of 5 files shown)


tbraun96:
Verified on a live GB10 serve running the exact BFCL leg that the leak previously killed.

Host RSS of the serve process, sampled every 2 min (`bfcl-subset-echolp`, 35B):

    t+10 min     1756 MB     +33 MB      41 reqs
    t+12 min     1789 MB     +33 MB      52 reqs    <- cache reaches its bound
    t+14 min     1789 MB      +0 MB      66 reqs
    t+16 min     1789 MB      +0 MB      80 reqs
    t+18 min     1790 MB      +0 MB      94 reqs
    t+20 min     1790 MB      +0 MB     104 reqs
    t+22 min     1790 MB      +0 MB     118 reqs
    t+24 min     1790 MB      +0 MB     131 reqs
    t+26 min     1986 MB    +197 MB     146 reqs    <- a new distinct schema admitted
    t+28 min     1986 MB      +0 MB     159 reqs

Before: ~100 MB/min, monotonic, ~17 MB per request, and the leg could not finish —
headroom hit the floor first.

After: flat between steps. 118 requests from t+12 to t+28 moved RSS by 197 MB, and
that motion is a single step where a new schema entered the cache, not accumulation.
The steady-state slope across the flat windows is ~0.1 MB/min against ~100 MB/min
before — the growth is now bounded by the budget rather than by uptime.

The remaining steps are expected and are the point: the cache admits genuinely new
grammars, then evicts LRU once the 1 GiB bound binds. What is gone is the monotonic
per-request accumulation that made restarts a scheduled chore (issue #368).

Gates on this branch: ttft-warm PASS (+0.2% median, limit +3.0%) and ttft-cold PASS
(-0.2%) — those were the regression risk, since eviction can force a grammar
recompile. bfcl-subset, bfcl-subset-echolp and agentic-webserver were still in flight
at merge time; #389 re-runs the full five against this engine.

PR #419, opened 2026-08-09
ci: compile the CUDA kernels — CI was building none of them

tbraun96:
## The hole

`ci.yml:30` sets `ATLAS_SKIP_BUILD: "1"` repo-wide. That makes `atlas-kernels/build.rs` emit a stub `target_ptx.rs` and invoke nvcc **zero times**.

That is the right default for the Rust jobs — they need no GPU and no CUDA toolchain. But the consequence is that **nothing in CI has ever compiled a `.cu` file**. A kernel change was first built when a human ran it on a DGX.

The cost is measured, not hypothesised. Three PRs reviewed on 2026-08-08 (#334, #335, #350) carried roughly 2,000 lines of new CUDA between them, none of it compiled by anything before review. One shipped a half-converted kernel — a `BR=64` twin still reading `q_len` where the `BR=32` path had moved to `q_len_eff`.

## What this adds

One paths-filtered job (`kernels/**`, `crates/atlas-kernels/**`) that runs the **real `build.rs`** in `nvidia/cuda:13.0.2-devel-ubuntu24.04` with `ATLAS_TARGET_MODEL: "*"` and, deliberately, **no `ATLAS_SKIP_BUILD`**.

Running `build.rs` rather than a hand-rolled nvcc loop in YAML is the point: `build.rs` owns target resolution, the `common/`-vs-shadow merge by file stem, per-target `KERNEL.toml` flags and the strict-warning promotion. Reimplementing any of that here would be a second source of truth that drifts silently — and a weaker check than the one it imitates.

## Cost

Cheaper than the file count suggests. `build.rs` deduplicates by `(source, arch, sorted flags)` and byte-copies the rest, so 22 gb10 targets collapse to **449 unique nvcc invocations out of 3583 — 8.0x dedup**.

Measured end-to-end on a DGX at 20 workers:

| | |
|---|---|
| wall | **29.7 s** |
| unique nvcc invocations | 449 (of 3583; 3134 cache hits) |
| PTX modules produced | 14,342 |
| GPU required | none — `nvcc --ptx` cross-compiles for `sm_121a` |

`timeout-minutes: 30` leaves ample room for a 4-vCPU hosted runner plus apt/rustup/cargo deps.

## Scope

Compile only. Executing kernels needs a GB10 and stays with the benchmark gates. This closes the "does it even build" half, which is the half that was open.

## The guard took three attempts

A job that goes green on a skipped build is worse than no job, so the skip guard is the load-bearing part. Both rejected formulations are recorded in the file so they are not re-tried:

- `grep -c 'ptx_modules'` — **passes on a skipped build.** The skip stub defines `ptx_modules()` too, returning an empty `Vec`.
- "assert no stub is present" — **fails on a healthy build.** cargo keeps one `out/` per build-script fingerprint and never prunes the old ones, so a tree that has ever been built with `ATLAS_SKIP_BUILD=1` keeps those stubs forever. This version failed a local build that had just compiled 14,342 PTX modules.

What survives asserts existence, not absence: at least one generated `target_ptx.rs` must be a real one. Verified against three trees:

| tree | guard |
|---|---|
| real build (4 real + stale stubs) | pass |
| stub-only | fail |
| empty `target/` | fail (fail-closed) |

## Self-testing

The paths filter includes `.github/workflows/kernel-compile.yml`, so this PR triggers the job it adds — the first CI run is the proof it works on a hosted runner.


tbraun96:
Rebuilt as a **single, workflow-only commit** (`29c931d77`). It now touches exactly one file: `.github/workflows/kernel-compile.yml`.

**Merge #420 first.** Two reasons:

1. The 500-LoC fix (`atlas-kernels/src/lib.rs` 496→517 in #389, mine) moved to #420, so the two PRs can't conflict on a squash. #420 carries both blockers.
2. This PR's benchmark gate cannot pass until #420 lands. Today `record_covers` asks `git merge-base --is-ancestor`, and #389's five passing records stopped being ancestors of anything when it squash-merged — so *every* PR reads `NONE` regardless of content. Once #420 makes the check content-based, this PR touches **zero perf paths** and the existing records cover it honestly.

Also folded in the two in-flight CI fixes, both now documented in the file itself:

- container jobs default to `sh -e {0}` — dash, no `pipefail`. The first run died on `Illegal option -o pipefail` before executing a line. Pinned `defaults.run.shell: bash`.
- the skip guard took three attempts; the two rejected ones are recorded so they aren't re-tried.

PR #423, opened 2026-08-09
fix(gate): close three holes an adversarial review found

tbraun96:
A subagent was told to **break** the gate, not confirm it. It found three ways a PR could report `PASS` on code the gate never measured — plus one finding that reframes all of them. I verified each myself before acting.

## The reframing, first

**`PR benchmark gate` is not a required status check on `main`.** The required contexts are `cargo fmt`, `cargo clippy`, `cargo test`, `typos`, SPDX, ≤500 LoC, CLA. `enforce_admins: false`; the ruleset requires 0 approving reviews.

So this apparatus is a **warning light, not a lock**. Everything below fixes what that light reports. Making it required is a separate decision and not mine to take.

## 1. The coverage rules exempted themselves

`GATE_MACHINERY` excludes the whole `crates/atlas-plugin/src/gate` prefix from every gate, and `BOUNDARY_FILES` listed only `coverage.rs`. So a PR editing `check.rs` — `record_covers`, `invalidating_paths`, `check_record`, `compare` — **invalidated nothing, and then reported itself covered by its own new logic**.

`coverage.rs:25` argues exactly this case for itself, *"a lock whose key is kept inside it"*, and then leaves the key one room over.

Not theoretical: **#420 rewrote `record_covers`** and the gate named only an unrelated `atlas-kernels` file as invalidating. It read red by accident, not by rule.

`BOUNDARY_FILES` now names the four files that decide a verdict — `check.rs`, `closure.rs`, `taxon.rs`, `bench.rs`. `GATE_MACHINERY` still covers the rest of the directory (record IO, telemetry, CODEOWNERS), where its argument does hold.

**Cost:** a gate-logic PR now owes one gate run. That is the correct price for editing the thing that decides whether anything owes a gate run.

## 2. A warm record satisfied the cold gate

`check_one` located records by **directory** and never read `record.benchmark_id` back. `ttft-warm-gate` and `ttft-cold-gate` share a checkpoint, a hardware key, and their metric names:

| | warm record | cold ceiling |
|---|---|---|
| `median_ms` | 1562.58 | max 1728.27 ✅ |
| `p90_ms` | 4478.42 | max 4809.76 ✅ |

`cp` one file between directories and the cold gate goes green with no cold leg ever run. Confirmed live — with the new check reverted, the test reports **`got Pass`**.

The worst possible pair to be able to confuse: cold-TTFT is the only leg that sees a cold-**load** regression, and #389 — the change this gate was built alongside — is *"GPU-transpose quantized weights at cold load"*. And it needs no malice: both records are written minutes apart in one session with near-identical filenames.

Mismatches are **skipped with a warning, not failed** — a stray file should leave the gate reading "no covering record" (true and actionable), not manufacture a hard failure out of someone else's passing run.

## 3. `noise` was uncapped, and applied to exact pins

`noise` widens a threshold at compare time and had no upper limit. `noise = 1000.0` on a floor of 87.44 turns that gate green against records already in the tree — and it reads like a measurement annotation, not a threshold change, which makes it the most review-invisible edit available. `BENCH.toml` is deliberately exempt from invalidating any gate, so nothing re-opens.

Worse: `check::compare` applies noise to the two-sided arm, so noise on an **exact pin** (`min == max`) silently disables it — and that is how the BFCL **draw size** is pinned. A changed draw is undetectable after the fact, because a different category mix moves `normalized_single_turn_score` by ~1.8 points while leaving `overall_accuracy` in the same place.

`validate_noise` now refuses non-finite/negative slack, slack on an exact pin, and slack above **5%** of the bound. Every value in the tree is `0.4` against floors of 83–89 (~0.46%) and still loads — a rule that rejects the status quo is a migration, not a rule.

## Run both ways

Each fix reverted, tests kept:

| revert | test | result |
|---|---|---|
| `benchmark_id` check removed | `a_record_from_another_gate_does_not_satisfy_this_one` | **FAILED** — `got Pass` |
| `BOUNDARY_FILES` back to 1 | `verdict_logic_is_inside_the_boundary` | **FAILED** — `got []` |
| `validate_noise` unwired | 3 of the 5 noise tests | **FAILED** |

`the_real_noise_values_still_load` passes both ways on purpose — it pins that the cap didn't break the tree.

`atlas-plugin`: **413 passed, 0 failed.** fmt + clippy clean, no LoC violations (the two boundary tests moved into `hardening_tests.rs` to keep `coverage_map_tests.rs` at 468).

## Deliberately not done here

Both change repo-wide behaviour and are the owner's call:

- **Making this gate a required check.**
- **Turning off `delete_branch_on_merge`.** 12 of 13 record shas on `main` are reachable only from two still-living PR branches (`pr388-gpu-transpose`, `w55-squash`). Deleting those makes `git diff` fail for every record and re-opens the ADR-0013 outage — on a delay set by when someone tidies branches.

Still open from the same review, not fixed here: recipes live in `atlas-recipes` (a different repo, unpinned and unhashed) while determining every gated number; `docker/` is outside `PERF_PATHS` while carrying `ATLAS_*` env and rebuilding `libatlasgdn.so`; `recorded_at` is self-reported and orders record selection.

### DIFF
--- a/crates/atlas-plugin/src/gate/bench.rs
+++ b/crates/atlas-plugin/src/gate/bench.rs
@@ -118,6 +118,7 @@ pub fn load_all(root: &Path) -> Result<Vec<(taxon::Target, BenchEntry)>> {
                     entry.checkpoint
                 );
             }
+            validate_noise(&path, &entry)?;
             out.push((
                 taxon::Target {
                     hardware: target.hardware.clone(),
@@ -214,3 +215,67 @@ pub fn baseline_for(root: &Path, benchmark_id: &str) -> Result<GateBaseline> {
 #[cfg(test)]
 #[path = "bench_tests.rs"]
 mod bench_tests;
+
+/// The largest slack a `noise` allowance may claim, as a fraction of the bound
+/// it is applied to.
+///
+/// `noise` widens a threshold at compare time (`check::compare`) and had no
+/// upper limit at all. `noise = 1000.0` on a floor of 87.44 turns that gate
+/// green against any record already in the tree — and it reads like a
+/// measurement annotation, not a threshold change, which makes it the most
+/// review-invisible way to defeat the gate. Every value in the tree today is
+/// 0.4 against floors of 83-89, i.e. ~0.46%.
+const MAX_NOISE_FRACTION: f64 = 0.05;
+
+/// `noise` must be a small, non-negative, finite slack — and must never be
+/// applied to an EXACT pin.
+///
+/// The exact-pin rule is the load-bearing half. `min == max` is how the BFCL
+/// draw size is pinned (`samples` = 995 / 1004), and that pin exists precisely
+/// to catch a silently-changed draw — a different category mix moves
+/// `normalized_single_turn_score` by ~1.8 points while leaving
+/// `overall_accuracy` in the same place, which is exactly what makes crossing
+/// draws impossible to spot after the fact. `check::compare` applies `noise` to
+/// the two-sided arm as well, so a `noise` on `samples` would silently disable
+/// the guard.
+fn validate_noise(path: &std::path::Path, entry: &BenchEntry) -> Result<()> {
+    let Some(metrics) = entry.metrics.as_ref() else {
+        return Ok(());
+    };
+    for (name, bound) in metrics {
+        let Some(noise) = bound.noise else { continue };
+        if !noise.is_finite() || noise < 0.0 {
+            bail!(
+                "{}: {} / {} metric {name}: noise must be finite and non-negative, got {noise}",
+                path.display(),
+                entry.gate,
+                entry.checkpoint,
+            );
+        }
+        if bound.min.is_some() && bound.min == bound.max {
+            bail!(
+                "{}: {} / {} metric {name} is an EXACT pin (min == max == {:?}) and carries \
+                 noise {noise}. Noise on a pin disables it — and a pin is used for things like \
+                 the BFCL draw size, where a changed draw is undetectable after the fact.",
+                path.display(),
+                entry.gate,
+                entry.checkpoint,
+                bound.min,
+            );
+        }
+        let magnitude = bound.min.or(bound.max).unwrap_or(0.0).abs();
+        let cap = magnitude * MAX_NOISE_FRACTION;
+        if magnitude > 0.0 && noise > cap {
+            bail!(
+                "{}: {} / {} metric {name}: noise {noise} exceeds {:.0}% of the bound \
+                 ({magnitude}) — that is a threshold change wearing a measurement-noise \
+                 label. Move the bound instead, so the ratchet is visible in review.",
+                path.display(),
+                entry.gate,
+                entry.checkpoint,
+                MAX_NOISE_FRACTION * 100.0,
+            );
+        }
+    }
+    Ok(())
+}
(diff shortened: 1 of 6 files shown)


tbraun96:
Correcting one of the findings I carried into this PR's description as still-open. **It is wrong, and I should not act on it.**

I listed, under "still open from the same review":

> `docker/` is outside `PERF_PATHS` while carrying `ATLAS_*` env and rebuilding `libatlasgdn.so`

The premise is right — `docker/gb10/Dockerfile.builder` does set `ATLAS_CUTLASS_NVFP4_GEMM`, `ATLAS_GDN_FLASHINFER`, `LD_LIBRARY_PATH`, and does rebuild `libatlasgdn.so`. The conclusion does not follow, because **gate records are not produced in a container.**

`bench_selfstart::serve_for` spawns the server **in-process**:

```rust
tokio::spawn(async move { crate::main_modules::serve::serve(serve_args, None).await });
```

The only occurrence of "docker" in that whole file is inside an error message telling an operator to run `sudo docker ps` when the box is not free. There is no container in the path a `--pull-request-gate` run takes.

So a `docker/` edit cannot change a gate record's numbers, and adding `docker` to `PERF_PATHS` would invalidate every record for a change that provably did not affect them — the exact over-invalidation that makes a gate expensive enough to be ignored. Withdrawn.

**What is still true, and is a different problem:** the numbers Atlas *ships* — the release image, the MLPerf submissions, anything a user pulls — DO come from that container, and nothing ties them to a gate record. That is a provenance gap between "what we measured" and "what we publish", not a coverage gap in the gate. It belongs with #426 (the recipe is unpinned) rather than in `PERF_PATHS`.

The other two open findings from that review stand unchanged: **#426** (recipes live in another repo, unpinned and unhashed, while determining every gated number) and `recorded_at` being self-reported while ordering record selection.

Three of the four adversarial findings I acted on were real and are fixed here with tests that fail without them. This fourth one I checked before acting, and it did not survive the check — which is the point of checking.

tbraun96:
The self-referential proof landed. Refreshed onto `main` (now carrying #428's five records), and the gate says:

```
NONE  agentic-webserver — latest record is for 0fe272515f — invalidated by
      crates/atlas-plugin/src/gate/bench.rs,
      crates/atlas-plugin/src/gate/check.rs,
      crates/atlas-plugin/src/gate/coverage.rs
```

Two things worth reading off that, in order of importance:

**1. `check.rs` and `bench.rs` are in that list only because this PR put them there.** On `main` as it stands, `GATE_MACHINERY` excludes the entire `crates/atlas-plugin/src/gate` prefix from every gate — so without its own change, this PR would invalidate nothing and report **PASS by its own new logic**, which is exactly the hole it exists to close. It is now invalidated by the rule it introduces. That is the behaviour working, demonstrated on itself rather than argued for.

**2. The message is actionable.** This morning every PR read `it is not an ancestor of this commit` — true, unfixable, and identical regardless of content. It now names three files and a record, so the debt is dischargeable by measuring.

`atlas-plugin`: 413 passed, 0 failed. fmt clean. All seven required checks green; the only red is the benchmark gate, correctly.

On how to discharge it: records are written *against* a commit, and a record cannot be written at a HEAD that does not exist yet, so the natural flow is merge → measure the new tip → commit the records. I would rather run **one** gate cycle covering this plus #418 than three cycles covering one PR each — neither touches an engine path (`gate/` never runs a model; #418 is a loader-key fix). #425 stays separate and gets its own spec-OFF run, because it genuinely changes decode dispatch and deserves to be measured before it lands, not after.

PR #425, opened 2026-08-09
perf(decode): give the mixed co-dispatch head the same lm_head ladder

tbraun96:
## The gap

`decode_b2::mixed_final_norm_lm_head` — the prefill+decode co-dispatch head, reached from `decode_b.rs` via `mixed_forward_dispatch`, i.e. **live continuous-batching traffic** — still looped `padded_n` times through `ops::w4a16_gemv`, streaming the **entire vocab weight once per row**. On a `[248320, 5120]` NVFP4 head that is ~N × 254 MB/step.

Its sibling `decode_a2` grew the batched ladder weeks ago. The two had simply drifted.

## Credit

**@rsafier found this site**, in #332. That PR pointed at exactly the right line. It fixed it with a bare `batch16` behind a default-OFF env var; this takes a different route.

## Why extract rather than copy

Duplicating the ladder would put "which kernel does `padded_n` land on" in two independently-maintained places. The first thing to go wrong would not be a perf drift — it would be a silent **numerical** difference between the pure-decode and co-dispatch paths at the same batch width, which is worse than the bandwidth bug being fixed.

So `decode_a2`'s block moves to `trait_impl/lm_head_batched.rs` as one `lm_head_project_batched()`, and both heads call it. `v` is read from `self.config.vocab_size` inside rather than passed, so there is no second place for it to be wrong.

Default-**ON** under the existing kill-switch (`ATLAS_NO_LM_HEAD_BATCH_GEMV=1`), inherited unchanged — not a new flag, per the house rule that wins ship as defaults with a kill-switch rather than as opt-in.

Side effect worth having: `decode_a2.rs` drops **617 → 501** LoC.

## Verification

`spark-model`: **311 passed, 0 failed**. fmt + clippy clean, no LoC violations.

## ★ Honesty about the claim

**This is a bandwidth-accounting argument, not a measurement.** Nobody has A/B'd the mixed path and I am not claiming a number.

The gate run this PR owes must be **spec-OFF or on reversed-order pairs** — a single spec-ON pair drifts ±2%, which is wider than any effect worth reporting here.

There is no unit test that proves the win, and I do not think one is possible: the difference is how many times a weight is streamed, which is invisible to any CPU-side oracle. What the tests do prove is that both heads still produce the same dispatch decisions, because there is now only one function making them.

## Deliberately not included

`SequenceState::cached_prefix_blocks` (`traits.rs:140`) is written at three sites and read at **none**. Removing a `pub` field is an API change and does not belong bundled into a perf PR — noted for a separate one.

Co-Authored-By: rsafier <rsafier@users.noreply.github.com>

### DIFF
--- a/crates/spark-model/src/model/trait_impl/decode_a2.rs
+++ b/crates/spark-model/src/model/trait_impl/decode_a2.rs
@@ -30,16 +30,6 @@ fn multiseq_graphs_enabled() -> bool {
     *ON.get_or_init(|| std::env::var("ATLAS_NO_DECODE_GRAPHS_MULTISEQ").as_deref() != Ok("1"))
 }
 
-/// Batched-GEMV decode lm_head: **ON by default**, disabled by
-/// `ATLAS_NO_LM_HEAD_BATCH_GEMV=1`.
-///
-/// Strict `== "1"` on an `ATLAS_NO_*` name, not a presence check — presence
-/// flags in this codebase are ENABLED by `=0`. Read once; this is a per-step site.
-fn lm_head_batch_gemv_enabled() -> bool {
-    static ON: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
-    *ON.get_or_init(|| std::env::var("ATLAS_NO_LM_HEAD_BATCH_GEMV").as_deref() != Ok("1"))
-}
-
 impl TransformerModel {
     pub(super) fn decode_batch_dispatch(
         &self,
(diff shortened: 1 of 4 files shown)


tbraun96:
Correcting something I wrote in the description. I said:

> There is no unit test that proves the win, and I do not think one is possible: the difference is how many times a weight is streamed, which is invisible to any CPU-side oracle.

**The second half is wrong.** The oracle exists and it is already in-tree:

`spark-runtime/src/gpu/mock.rs:15` — *"Records kernel launches and memory operations for test assertions"* — with `launches: Mutex<Vec<MockLaunch>>` and `pub fn launch_count(&self)`.

That is precisely the right oracle for this change. The whole claim here is "one `w4a16_gemv_batchm` launch instead of `padded_n` separate `w4a16_gemv` launches", and a launch count is a **direct** measurement of it — no GPU, no timing, no noise floor. A test would assert the mixed head issues **1** projection launch at `padded_n = 8`, not 8.

**The real blocker is the fixture, not the oracle.** Reaching `mixed_final_norm_lm_head` needs a constructed `TransformerModel`, and the only tests that touch `build_model` today (`factory.rs:136,199`) pass `WeightStore::empty()` and assert that it *errors*. Nothing in the tree builds a model far enough to run a head against a mock backend.

So the accurate statement is: **this change is testable, and it is untested because `TransformerModel` is not constructible in a unit test** — which is a testability gap worth naming rather than a property of the change.

That reframes what this PR owes. Not "trust the bandwidth argument", but "either build the fixture or accept the gap knowingly". I am not going to build a model fixture inside a perf PR — that is its own piece of work with its own review — but I would rather the limitation be recorded correctly than dressed up as an impossibility.

The gate run this PR owes is unchanged, and still has to be spec-OFF or on reversed-order pairs.

PR #432, opened 2026-08-09
feat(ci): descending PR intent taxonomy + dynamic benchmark selection

tbraun96:
The categorizer emitted **one flat label** from seven siblings. That is what called a CI-only commit `performance`. A flat label also cannot *drive* anything — it says nothing about which benchmarks a change needs.

This adds the tree, the descent, and the selection.

## ★ The safety property — this is the whole design

> **`_benches` may only ADD to the required set. It can NEVER remove.**

The required set is `path_derived ∪ intent_derived`. The path-derived half (`PERF_PATHS` + the closure hash) stands alone; nothing here can shrink it. So a **misclassification costs GPU minutes, never a missed regression** — the only footing on which a language model belongs near a merge gate.

Invert it and the classifier becomes a way to skip tests by writing a misleading PR title. The diff would not even have to lie, only the prose.

**Proven, not asserted.** Swapping union semantics for leaf-only:

```
benches_may_only_add ... FAILED
descending into kv-cache DROPPED ["bfcl-subset"] — a deeper path must never
imply fewer benchmarks than its ancestor
```

## What it buys

The direction paths **cannot see**. A scheduler change under `crates/spark-server/` touches no `kernels/`, so every closure hash is unchanged and the rung ladder excuses every target — yet it can move decode wall badly. `performance/scheduling` pulls in the agentic leg that would otherwise not run.

| intent path | benchmarks implied |
|---|---|
| `performance` | `agentic-webserver` |
| `performance/decode` | `agentic-webserver`, **`bfcl-subset`** ← descending *added* |
| `correctness/kv-cache` | `bfcl-subset`, `ttft-cold-gate`, `ttft-warm-gate` |
| `documentation/reference` | *(none)* |

## Why descend rather than pick a leaf from a flat list

A flattened enumeration is one API call, but the closed-set guarantee weakens as leaves multiply — the model picks among ~25 near-identical strings, and a near-miss is indistinguishable from an abstain. Descending keeps every decision to **2–6 disjoint options**, which is what the allowlist check is good at.

It also fails better. A wrong turn at level 1 shows up as a whole branch that looks wrong; a wrong leaf in a flat list looks like every other leaf. And an abstain mid-descent yields a **partial path** — `performance` with no sub-category is more informative than a flat `unknown`, and honest about which question was answered.

**Cost, stated:** depth-many calls instead of one. `max-depth` (default 4) bounds it so a deeper tree cannot silently turn one classification into six.

## Pieces

| file | role |
|---|---|
| `.github/pr-taxonomy.json` | the tree (6 roots × 2–6 children) + `_benches`. SSOT for both the action and Rust |
| `.github/actions/classify-path` | descends; **allowlist applied at every level**, so no model-authored string becomes a path segment. A set of one is followed without spending a call |
| `gate::pr_taxonomy` | loads, **validates**, resolves path → bench set |
| `ci.yml` | renders path, depth, status, implied bench set |

## Tests — 12, and the ones that matter

- **`benches_may_only_add`** — the safety property (fails without union semantics)
- **`every_declared_bench_is_a_required_gate`** — a path selecting an unregistered id is a silent no-op that *reads as coverage*
- **`an_unknown_segment_degrades_instead_of_failing`** — a stale label yields fewer extra benches, never a crash that takes the view down
- **`reserved_keys_are_not_categories`** — `_doc`/`_benches` must never be offered to the model as things to pick
- **`a_lone_child_is_followed_not_asked`** — a set of one is not a choice

`atlas-plugin`: **425 passed, 0 failed.** fmt + clippy clean.

## ★ Scope, honestly

**Advisory today.** It renders the path and the implied set; it does **not** yet make `check_gates` require the union. That needs the path plumbed to gate-check time (a label or a record field) and is a separate change.

Also worth stating: the plan file's "dynamic benchmark selection" is the path-derived **rung ladder**, which is already live. This is the *intent*-derived half, which the plan did not cover.

### DIFF
--- a/.github/actions/classify-path/action.yml
+++ b/.github/actions/classify-path/action.yml
@@ -1,0 +1,239 @@
+name: Classify into a taxonomy path
+description: >-
+  Descend a category tree ONE LEVEL AT A TIME, emitting a full path like
+  `performance/decode`. Each level is a separate closed-set classification, so
+  the allowlist guarantee of `actions/categorize` holds at every step.
+
+# ── Why descend rather than pick a leaf from a flat list ─────────────────────
+#
+# A flattened enumeration ("performance/decode", "performance/prefill", …) is
+# one API call, but the closed-set property weakens as the leaf count grows:
+# the model is choosing from 25 near-identical strings, and a near-miss is
+# indistinguishable from an abstain. Descending keeps every individual decision
+# to 2-6 genuinely disjoint options, which is what the allowlist check in
+# `categorize` is good at.
+#
+# It also fails BETTER. A wrong turn at level 1 is visible in the rendered
+# table as a whole branch that looks off, whereas a wrong leaf in a flat list
+# looks like every other leaf. And an abstain mid-descent yields a PARTIAL
+# path — `performance` with no sub-category is strictly more informative than
+# a flat `unknown`, and it is honest: it says which question was answered and
+# which was not.
+#
+# ── The cost, stated ────────────────────────────────────────────────────────
+#
+# Depth-many API calls instead of one. With a 2-level tree that is two calls
+# per PR. `max-depth` bounds it so a future deeper tree cannot silently turn
+# one classification into six.
+
+inputs:
+  text:
+    description: The content to classify. Untrusted by assumption.
+    required: true
+  taxonomy:
+    description: >-
+      Path to the taxonomy JSON, relative to the workspace. Nodes are objects;
+      `_doc` and `_benches` are metadata, not categories.
+    required: false
+    default: .github/pr-taxonomy.json
+  model:
+    description: Model id. No default — the caller owns which model classifies.
+    required: true
+  api-key:
+    description: Bearer token. Empty yields an immediate abstain (fork PRs).
+    required: true
+  fallback:
+    description: >-
+      Root-level category emitted when even the FIRST level cannot be
+      answered. Must be a root of the taxonomy.
+    required: true
+  endpoint:
+    required: false
+    default: https://openrouter.ai/api/v1/chat/completions
+  max-chars:
+    required: false
+    default: "12000"
+  timeout-seconds:
+    required: false
+    default: "45"
+  max-depth:
+    description: >-
+      Hard bound on levels descended. Guards against a future tree turning one
+      classification into an unbounded number of API calls.
+    required: false
+    default: "4"
+
+outputs:
+  path:
+    description: >-
+      Slash-joined taxonomy path, e.g. `performance/decode`. May be PARTIAL
+      (a single segment) when a level abstained — that is a real answer, not a
+      failure.
+    value: ${{ steps.run.outputs.path }}
+  depth:
+    description: Number of segments actually resolved.
+    value: ${{ steps.run.outputs.depth }}
+  status:
+    description: >-
+      ok | partial | abstain | error. `ok` = reached a leaf. `partial` = at
+      least one level resolved, then a level could not be answered.
+      `abstain` = not even the first level. `error` = misconfiguration.
+    value: ${{ steps.run.outputs.status }}
+  reason:
+    description: Short human-readable cause. Never contains model-authored text.
+    value: ${{ steps.run.outputs.reason }}
+
+runs:
+  using: composite
+  steps:
+    - id: run
+      shell: bash
+      # Every input crosses in through `env:`, never `${{ }}` spliced into the
+      # script body — same rule as `actions/categorize`, same reason.
+      env:
+        INPUT_TEXT: ${{ inputs.text }}
+        INPUT_TAXONOMY: ${{ inputs.taxonomy }}
+        INPUT_MODEL: ${{ inputs.model }}
+        INPUT_API_KEY: ${{ inputs.api-key }}
+        INPUT_FALLBACK: ${{ inputs.fallback }}
+        INPUT_ENDPOINT: ${{ inputs.endpoint }}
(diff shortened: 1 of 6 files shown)


tbraun96:
**Moving this to draft. An adversarial review found that the justification I wrote for it is false, and I verified that myself before believing it.**

## 1. The worked example is a no-op — verified

`pr_taxonomy.rs`, the JSON `_doc`, the commit and the PR body all say:

> "A scheduler change under `crates/spark-server/` touches no `kernels/`, so every closure hash is unchanged and the rungs excuse every target — yet it can move decode wall badly. `performance/scheduling` pulls in the agentic leg that would otherwise not run."

I ran it:

```
SCHEDULER CHANGE ALREADY INVALIDATES:
["agentic-webserver","ttft-warm-gate","ttft-cold-gate","bfcl-subset","bfcl-subset-echolp"]
```

`PERF_PATHS` contains the bare string `"crates"` (`coverage.rs:63`), so `crates/spark-server/**` invalidates **all five**. And `closure::excuses` returns `false` unless *every* surviving path is inside `kernels/` — its own doc says so. `performance/scheduling` resolves to `{agentic-webserver}`, which the floor already required.

**The generalisation is worse.** Since `crates` and `kernels` are both perf paths, any PR that touches code at all already owes all five gates — so `_benches` is a strict subset of the floor **every time**. The intent set can only be non-redundant for PRs touching *only* docs, `site/`, `.github/`, `bench/`, `scripts/` — precisely where a `performance/*` label is almost certainly a misclassification.

## 2. The classifier reads a title and a file COUNT — verified

`ci.yml:224` computes `changed_head.txt` and **nothing ever reads it** (`grep changed_head` returns that one line). The model's entire input is the PR title plus `N paths`. Classification is 100% attacker-authored prose; the honest mechanism *is* the bypass. No jailbreak needed — title a decode-kernel PR `docs: clarify a comment` and get `documentation/reference` → `{}`.

## 3. `GITHUB_OUTPUT` injection on the one branch without an allowlist — verified

`action.yml:213`: the auto-follow branch takes `jq -r '.[0]'` directly into `PATH_SO_FAR`, never calling `ask()` — so the allowlist check at `:196` is **not on that path**. `jq -r` renders `\n` as a real newline, and `emit()` writes `echo "path=$2"` with no heredoc delimiter. A taxonomy key containing newlines injects arbitrary step outputs.

The irony: the only code path with no allowlist is the one `pr_taxonomy::validate` forbids from existing — and `validate` runs in the `test` job, which `pr-categorize` does not `needs:`.

There is a second sink at `:136`, pre-auth: `$INPUT_FALLBACK` is written unvalidated *before* the API-key check. Harmless with today's literal `unknown`, live for the first caller that computes a fallback.

## 4. Rust and bash disagree on `_benches`, in the removing direction — verified

| `_benches` value | Rust | jq |
|---|---|---|
| `"bfcl-subset"` (string) | **dropped** | `bfcl-subset` |
| `[["bfcl-subset"]]` | **dropped** | `bfcl-subset` |
| `[1,"bfcl-subset"]` | `{bfcl-subset}` | prints `1` as a bench id |

All three pass `validate()` and both shipped tests. A typo silently *removes* a benchmark on the Rust side.

## 5. The mappings are wrong where it matters

- **`capability/*` implies ZERO benchmarks.** New model, new hardware, adapters — the category that means "code that has never run before" is the one cell with no bench.
- `correctness/loader` gets only `bfcl-subset`. #418 was an NVFP4 prefix indexed by an expert id — a numerics-destroying loader bug — and this repo's own notes say gate C2/NVFP4 smoke is what catches broken NVFP4 that A and B pass right over.
- `performance/decode` and `performance/scheduling` imply no TTFT gate, contradicting the `benchmark-pr` policy that gate C applies whenever the diff touches prefill/decode or the scheduler.
- One path per PR is itself a removal: a kernel change is *both* `correctness/numerics` and `performance/decode`; picking one drops the other's benches. "May only ADD" is true against a floor, never against the truth.

## 6. And my test does not prove what I said it proves

`benches_may_only_add` shows `benches_for` is monotone under prefix extension — true by construction, it is a union along a walk. It does **not** prove `path_derived ∪ intent_derived ⊇ path_derived`, because **nothing computes that union**. `grep -rn pr_taxonomy` finds four hits: a `pub mod`, a `#[path]`, and two comments. `validate()` has no non-test callers.

I described a regression guard on an uncalled function as *"the property that makes it acceptable to let a language model influence a merge gate."* That was overclaiming, and the review was right to call it theatre plus two API calls.

## What survived attack

JSON escaping of the title (`--arg`); the `ask()` allowlist; Rust/bash key **ordering** (`preserve_order` is on); unicode homoglyphs; infinite descent; and — importantly — **no live path from `steps.path.outputs.*` to the benchmark gate.** The floor is genuinely independent today.

## What happens next

The descent mechanism is sound and worth keeping. What needs redesign before this merges:

1. **Feed the classifier the actual changed paths**, not a count.
2. **Fix the injection**: heredoc-delimited `emit`, and put the allowlist on the auto-follow branch too.
3. **Make `_benches` typed** and enforce it in `validate()`, with `validate` actually called somewhere that runs.
4. **Find the real gaps in the floor first.** `_benches` is only worth having where the path rule under-covers — and I have not shown a single such case. That is the work, and it is measurement, not tree-editing.
5. Fix `capability/*` and the TTFT mappings.

Not merging this on an argument I have now watched fail.

tbraun96:
Superseded by the consolidated campaign PR — one PR for the whole benchmarking + AI system, per request. All of this PR's content and the adversarial-review fixes are carried there.

PR #439, opened 2026-08-10
feat(fp8-kv): detect checkpoint k/v scales and log the right guidance (#153 item 1)

rsafier:
Split out of #372 per review — this commit shared nothing with the calibration fix there, so it gets its own PR. Implements proposal item 1 of #153.

**Detection + logging only. No behavior change, no new flags or env vars.**

## Problem

The FP8 KV serve log WARNed whenever online calibration was off ("requires calibrated k_scale/v_scale ... Without scales (default=1.0) ... clips"), even when the checkpoint already ships real per-layer scales — misleadingly pushing users toward online calibration they don't need, and which is looser than the shipped full-data-calibrated scales.

## Change

`WeightStore::fp8_kv_scale_count()` counts the checkpoint's `*.k_scale` tensors (`k_scale`/`v_scale` ship as 1:1 pairs, so counting one side avoids double-reporting). `resolve_kv_cache_config` takes the count and branches the FP8 KV guidance on presence:

| Checkpoint scales | Calibration | Log |
|---|---|---|
| shipped | off | INFO: "FP8 KV cache using N per-layer k_scale/v_scale tensors from the checkpoint — no calibration needed." (the common NVFP4 case) |
| shipped | on | INFO: calibration is ON but the checkpoint ships N calibrated scale pairs (tighter than a runtime first-observe estimate) — prefer `--fp8-kv-calibration-tokens 0` |
| none | off | WARN (unchanged in substance): no k/v scales, default=1.0 silently clips BF16 into E4M3 range — enable `--fp8-kv-calibration-tokens 256` or use `--kv-cache-dtype nvfp4/bf16` |
| none | on | INFO: online calibration appropriate — freezing per-tensor scales on the first observed tokens |

The last case's text is also updated to match the hardened calibration behavior (scales freeze on the first observed tokens, not "during first N tokens").

## Tests

Unit tests on the counting helper (MockGpuBackend, no GPU): counts only `.k_scale` suffixes, ignores `v_scale` pairs and lookalike names (`attnk_scale`, `weight_scale`); returns 0 for scale-free checkpoints.

`cargo fmt`, `clippy -D warnings`, and `cargo test` for spark-runtime + spark-server all pass.

### DIFF
--- a/crates/spark-runtime/src/weights.rs
+++ b/crates/spark-runtime/src/weights.rs
@@ -183,6 +183,15 @@ impl WeightStore {
             .values()
             .any(|w| matches!(w.dtype, WeightDtype::FP8E4M3))
     }
+
+    /// Number of per-layer FP8 KV-cache scale tensors (`*.k_scale`) the
+    /// checkpoint ships. `>0` means the model carries calibrated KV scales, so
+    /// FP8 KV needs no online calibration; `0` means the scales default to 1.0
+    /// (which clips BF16 into E4M3 range), so online calibration or a non-FP8 KV
+    /// dtype is required. Used to log the right guidance at serve time.
+    pub fn fp8_kv_scale_count(&self) -> usize {
+        self.names().filter(|n| n.ends_with(".k_scale")).count()
+    }
 }
 
 /// SBIO IORouter trait for weight loading.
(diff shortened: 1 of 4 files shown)


tbraun96:
Landed in main via the #440 stack merge (`3c941751`). Verified by content diff — `WeightStore::fp8_kv_scale_count()` at `crates/spark-runtime/src/weights.rs:192` is byte-identical, `serve_phases/kv_cache.rs` is identical (`has_ckpt_scales` at L213), and `serve_load.rs:408` passes it through.

Cosmetic note: your `weights/fp8_scale_count_tests.rs` was relocated into `weights/teardown_tests.rs` during integration — both tests are verbatim at L52 and L80. Thanks @rsafier.

PR #441, opened 2026-08-10
fix(hip): map __nv_cvt_fp8x2_to_halfraw2 for the Windows native-HIP build

tbraun96:
The `windows-x86_64-amd-hip` row has failed on **every open PR** since #389 squash-landed.

#389's decode-perf fix switched `paged_decode_attn_fp8.cu`'s unpack from four scalar
`__nv_cvt_fp8_to_halfraw` calls to two paired `__nv_cvt_fp8x2_to_halfraw2` calls. The scalar
name was mapped in the HIP compat header; the paired one was not:

```
paged_decode_attn_fp8.cu:56:23: error: use of undeclared identifier
'__nv_cvt_fp8x2_to_halfraw2'; did you mean '__hip_cvt_fp8x2_to_halfraw2'?
```

The compiler names the fix itself. HIP declares `__hip_cvt_fp8x2_to_halfraw2` in
`amd_hip_fp8.h` with the identical `(fp8x2_storage, interpretation) -> __half2_raw` shape, so
this is the same one-line mapping as the scalar entry directly above it.

`strix-hip/common/paged_decode_attn_fp8.cu` is a **symlink** into `gb10/common`, which is why a
change with no AMD intent broke the AMD row.

**Not specific to one PR.** #415 and #416 touch no kernels at all and fail this row with the
identical error once CI merges them with post-#389 main. Landing this clears the row fleet-wide.

Verification is CI itself: the row is red on this base today (see #438, job 93337752434) and
must go green here. The row is `continue-on-error`, so it never blocked a merge — it has just
been reading red on everything, which is how a *real* AMD break would hide.

### DIFF
--- a/.benchmarks/agentic-webserver/2026-08-10-91d09acaa1.json
+++ b/.benchmarks/agentic-webserver/2026-08-10-91d09acaa1.json
@@ -1,0 +1,273 @@
+{
+  "schema": 1,
+  "benchmark_id": "agentic-webserver",
+  "benchmark_name": "Agentic Webserver Test",
+  "git_sha": "91d09acaa1",
+  "recorded_at": 1786348647,
+  "target_model": "Qwen/Qwen3.6-35B-A3B-FP8",
+  "params": {
+    "build_timeout_s": "600",
+    "command_timeout_s": "180",
+    "iterations": "10",
+    "max_tokens": "8192",
+    "max_turns": "40",
+    "request_timeout_s": "900",
+    "serve_timeout_s": "30",
+    "wall_budget_s": "1300"
+  },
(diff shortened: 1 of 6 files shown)


tbraun96:
Landed on main via #444 (e509d1fd).

Content-verified rather than assumed — Atlas squash-merges, so ancestry checks are meaningless here. All six files this PR touches are byte-identical to main: the HIP compat header `crates/atlas-kernels/hip/compat/cuda_fp8.h` and the five `.benchmarks/` records at 91d09acaa1.

Thanks — this cleared the AMD-HIP release-matrix row that had been red across roughly a dozen PRs.

Issue #442, opened 2026-08-10
Delivery pipeline: remaining feature buckets from integrate/laguna-on-main (post-#334/#335/#350) + post-gate follow-ups


rsafier:
Tracking issue for the remaining feature deliveries staged on `MonumentalSystems:integrate/laguna-on-main` (the validated merge of the full feature tree with current main — 359 files / +33.8K payload, GPU-validated: Laguna-S NVFP4 18.3 tok/s, official GGUF Q4_K_M keep-packed 17.1, dense 27B, all 3/3 temp-0 facts). Each bucket below gets cut as its own PR by `git diff` extraction from that tree once its dependencies merge — cutting earlier would create overlapping diffs across open PRs.

## Extraction pipeline (dependency order)

Blocked on **#350** merging:
- [ ] **f1 — Laguna prefill/MoE perf** (~27 commits' worth): `layers/moe/forward_prefill*.rs`, `helpers_*`, `ops/moe_grouped_a*`, CUTLASS grouped NVFP4 GEMM, `qwen3_attention/prefill/**` ragged/cu_seqlens work. Heaviest review chunk; pre-split into CUTLASS-MoE / attention / cuBLASLt-projections if wanted.
- [ ] **f2 — Laguna decode concurrency**: `ops/gemv_dense*`, graph-safe device-offset grouped MoE, the BF16 lm-head batched-GEMV sibling (the #332 remainder main still lacks).
- [ ] **b — Laguna-XS**: `kernels/gb10/laguna-xs-2.1/**`, XS loader arms (NVFP4-packed shared expert), thinking-default gate, jinja `{% include %}` resolver bits not already on main.
- [ ] **g (server half) — chat/thinking/tool correctness**: the model-agnostic `api/chat*` fixes; the MODEL.toml preset half rides the Laguna PRs.

Blocked on **#334 + #350**:
- [ ] **e — keep-packed Q4_K_M serving**: `forward_prefill_packed.rs`, packed loader arms in `weight_loader/laguna/load_layers.rs`, `q4k_mmq` grouped kernels + `moe_q4k_decode_fused`, `moe_grouped_q8` arena. Serves the official poolside GGUF at ~17.5 tok/s decode.

Blocked on **#335 + e**:
- [ ] **d2 — Laguna-LoRA enablement + GGUF keep-packed LoRA parity** (small: the `model_type=laguna` admit + the packed-arm gate/up fold).

Independent (any time):
[...]

PR #443, opened 2026-08-10
fix(nemotron): load native FP8 SSM weights — stop the Mamba2 double-quant

rsafier:
Bucket **j** of the Laguna-stack delivery plan (#442): Nemotron/Puzzle-75B native-FP8 SSM.

## The bug

ModelOpt ships **NVIDIA-Nemotron-Labs-3-Puzzle-75B-A9B-NVFP4** with the Mamba2 `in_proj`/`out_proj` as **block-scaled FP8** inside an otherwise-NVFP4 repo. The loader dequantized those to BF16 and **re-quantized to NVFP4 under one global scale** spanning the whole `[18048, 4096]` tensor (measured at load: `global_max=0.394531`, `scale2=0.00014678` over 73.9M elements). Puzzle is Mamba-dominant — only 8 attention layers — so these projections carry the entire sequence state, and the model answered noun-retrieval questions from its training prior instead of from context.

Measured on a 977-token story ("My dog is named Rufus … mention his name often"):

| path | result |
|---|---|
| legacy requant | "Rover" / "Rex"; the given name appears **zero** times |
| native FP8 (decode+prefill) | Rufus ×6/×5/×2/×2 over four trials, zero substitutions |

## What this does

- **Native FP8 SSM** (`weight_loader/nemotron/ssm_layer.rs`, `layers/nemotron_mamba2*`): loads the checkpoint's own FP8 bytes into an **owned** allocation (the first cut aliased a `WeightStore` pointer and dereferenced it every token — real corruption, fixed and documented in the provenance commits) and runs `w8a16_gemv` (decode) / `w8a16_gemm[_pipelined]` (prefill) with the checkpoint's per-block scales end to end. Kernels are probed before committing to the path; any failure degrades to the legacy arm with a warning.
- **Native BF16 SSM**: Nano-30B leaves 6 of 23 Mamba layers unquantized; the legacy arm invented an NVFP4 quantization for them. Kept in checkpoint BF16 (`dense_gemm_bf16_pipelined`/`dense_gemv_bf16`).
- **Native FP8 shared expert** (`layers/nemotron_moe*`, `weight_map/nemotron.rs`): same argument for Puzzle's shared-expert up/down projections, decode and prefill.
- **BF16 attention kept unquantized** (`weight_loader/nemotron.rs`): Puzzle's 8 attention layers ship BF16 (~1.2 GB); crushing them to NVFP4 saved ~0.9 GB and degraded exactly the retrieval they exist for.
- **Correctness riders from the same investigation** (all in the extracted commit range):
  - `ops/ssm_ssd.rs`: SSD-scan shared-memory **fit check** — Nano-30B (`state_size=128` → 115968 B) exceeds the sm_121 opt-in limit (101376 B) and could not serve a single token; now falls back to the sequential scan instead of aborting.
  - `nemotron_moe/prefill_sorted.rs`: grouped-GEMM **m-tile bound** was an estimate; under-provisioning dropped tokens and leaked a prior request's activations. Bounded by the worst case; intermediates zeroed as defence in depth.
  - `puzzle nvfp4/rms_norm.cu`: **deterministic `gated_rms_norm`** — `atomicAdd` on floats made temperature-0 decode nondeterministic (2 distinct completions in 10 identical runs); replaced with fixed-order per-warp reduction.
- **MODEL.toml behavior fixes** (both Nemotron variants, measured in the provenance commits): sampling-presets opt-in (`generation_config.json` carries eval defaults `temperature=1.0`), thinking off by default (thinking-on leaks training-corpus text / returns empty content), `qwen3_coder` tool parser + no tool grammar for Puzzle, BF16 KV for Puzzle, default system prompt for Nano (its own template renders an empty system turn that alone degrades the model), and `max_thinking_budget` as the sole thinking cap.
- `examples/w8a16_gemv_numeric.rs`: numeric proof for all three w8a16 kernels at the exact Puzzle shapes (N=18048, K=4096).

## Env gates shipped AS-IS (lever conversion is follow-up, per house rules)

| flag | default | purpose |
|---|---|---|
| `ATLAS_NEMOTRON_NATIVE_FP8_SSM` | on (`1`) | `0` = legacy requant (same-binary A/B); `decode` = native decode only (bisects decode from prefill) |
| `ATLAS_NEMOTRON_NATIVE_BF16_SSM` | on | `0` restores the BF16→NVFP4 requant |
| `ATLAS_NEMOTRON_BF16_ATTN` | on | `0` restores quantize-everything attention |
| `ATLAS_SHARED_FP8_PREFILL` | off | pre-dequant FP8 shared-expert prefill arm (pre-existing gate, code moved) |
| `ATLAS_SHARED_W4A4_DOWN` | off | W4A4 shared-expert down arm |
| `ATLAS_NO_SHARED_W4A4` / `ATLAS_NO_SSM_FP8_PREFILL` / `ATLAS_NO_SSM_PREFILL_T` | off | per-arm escape hatches (pre-existing pattern) |
| `ATLAS_NO_SSM_PERSISTENT` | off | A/B: persistent vs plain sequential SSM prefill scan |
| `ATLAS_MOE_MAX_M_TILES_ESTIMATE` | off | restores the old (unsafe) m-tile estimate for A/B only |
| `ATLAS_MOE_NO_ZERO_INTERMEDIATES` | off | skips the defensive zeroing of reused MoE arena buffers |

## Extraction provenance

Cut from `MonumentalSystems/atlas` `integrate/laguna-on-main` against `avarok/main`, commit range `ab446cd7..3cf579a6` (landmark `0806ed88` "own the FP8 SSM weight bytes") plus `ee681272` (native BF16 SSM projections).

**Shared-file hand-carving scope** — the plan flagged `layers/moe/{forward_prefill_routed,init,mod}.rs` as shared with Laguna perf work. Audit result: the only bucket-j commit touching them is `0806ed88`, and its hunks there are **pure rustfmt reflow and a module-declaration reorder of Laguna GGUF code that does not exist on `avarok/main`**. Zero nemotron-substantive hunks exist in those files, so **none were taken** and the files are untouched by this PR. No re-shaping was needed; the nemotron code compiles against upstream's `moe/` as-is.

Three deliberate divergences from the integration tree:
- `puzzle nvfp4/w4a16_gemm.cu`: the integration tree's copy predates upstream's 2026-07-28 `ldb` fix (the merge kept a stale 8-arg `w4a16_gemm_t`); taking it would have silently re-introduced the B-stride bug that `w4a16_gemm_t_ldb_drift_is_exactly_the_known_set` guards. **Upstream's version is kept**; the guard test passes.
- `nemotron_mamba2/prefill.rs` (620 LoC) and `nemotron_moe.rs` (524 LoC) were split sibling-module style (`prefill_proj.rs`, `prefill_shared_up.rs`) to stay under the 500-LoC cap — same code, moved verbatim.
- One functional re-shape, found by the audit below: the keep-BF16 attention path leaves `attn.o_proj` as the NULL placeholder, and `weight_loader/nemotron.rs` unconditionally called `transpose_for_gemm` on it — `transpose_u8` read a NULL pointer, the `.ok()` swallowed the error, and the poisoned context surfaced as `CUDA_ERROR_ILLEGAL_ADDRESS (700)` at the next module load. Reproduced deterministically with `--check-kernels` on GB10 and bisected to `ATLAS_NEMOTRON_BF16_ATTN=1`; now guarded with `is_null()`. (The integration tree carries the same unguarded line; it presumably survives there via unrelated ops-layer differences. Worth a look when the rest of the stack lands.)

## Verification (GB10, `ATLAS_TARGET_MODEL='*'`)

- `cargo check -p spark-server` — clean
- `cargo fmt --check` — clean
- `cargo clippy -p spark-model --no-deps -- -D warnings` — 0 errors
- `cargo test -p spark-model` — **318 passed, 0 failed** (includes the ldb drift guard)
- 500-LoC scan vs `.github/workflows/file-size-cap.yml` — all changed `.rs` files ≤ 486 LoC, no new allow-list entries
- `spark serve --check-kernels` (release build, real checkpoints, default gates):
  - `nemotron-labs-3-puzzle-75b-a9b` — **PASSED**: 141 lookups, 0 unresolved, 4 expected-absent (all pre-existing entries; the new `w8a16_*`/`dense_gemm_bf16_pipelined`/`dense_gemv_bf16` lookups all resolve)
  - `nemotron-3-nano-30b-a3b` — **PASSED**: 142 lookups, 0 unresolved, 9 expected-absent (pre-existing)
  - No new `[expected_absent]` entries needed.

Closes nothing; part of #442.

### DIFF
--- a/crates/spark-model/examples/w8a16_gemv_numeric.rs
+++ b/crates/spark-model/examples/w8a16_gemv_numeric.rs
@@ -1,0 +1,267 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! NUMERIC correctness check for `w8a16_gemv` at the Nemotron-Puzzle SSM shapes.
+//!
+//! The sibling `gemv_fp4_vs_fp8_microtest` is timing-only — it never checks an
+//! output value. That left a real gap: enabling native FP8 for the Puzzle-75B
+//! Mamba2 projections made the model measurably WORSE (correct "Alice" became an
+//! invented "Charlie"), and a bisect pinned it to the DECODE path, i.e. this
+//! kernel — yet the scale value, scale shape, broadcast, kernel indexing, LUT and
+//! call-site argument order all inspect as correct. So test the arithmetic.
+//!
+//! Construction is chosen so the expected answer is exact and obvious:
+//!   A[k]         = 1.0        for all k
+//!   B[n,k]       = 0x38       = 1.0 in E4M3 (sign 0, exp 0111 -> 2^0, mantissa 0)
+//!   block_scale  = S          constant over every block
+//! then  C[n] = sum_k 1.0 * 1.0 * S = K * S  for every n.
+//!
+//! A per-block vs per-row scale-layout mix-up still yields the right answer under
+//! a CONSTANT scale, so the run also repeats with a scale that VARIES per block —
+//! `scale[nb, kb] = 1 + nb + 100*kb` — where the two layouts disagree and the
+//! expected value per row is a closed form: C[n] = 128 * sum_kb (1 + nb + 100*kb).
+//!
+//! Usage: cargo run --release -p spark-model --example w8a16_gemv_numeric -- [N] [K]
+//! Default N=18048 K=4096 (Puzzle SSM in_proj). Also try 4096 8192 (out_proj).
+
+use anyhow::{Result, bail};
+use spark_runtime::cuda_backend::AtlasCudaBackend;
+use spark_runtime::gpu::GpuBackend;
+use spark_runtime::kernel_args::KernelLaunch;
+
+const FP8_BLOCK: usize = 128;
+
+fn bf16_bytes(v: f32) -> [u8; 2] {
+    // Round-to-nearest-even truncation of f32 -> bf16, as the loaders do.
+    let bits = v.to_bits();
+    let rounded = ((bits >> 16) & 1).wrapping_add(0x7fff).wrapping_add(bits);
+    ((rounded >> 16) as u16).to_le_bytes()
+}
+
+fn bf16_to_f32(lo: u8, hi: u8) -> f32 {
+    f32::from_bits(((u16::from_le_bytes([lo, hi])) as u32) << 16)
+}
+
+fn run(gpu: &dyn GpuBackend, stream: u64, n: u32, k: u32, vary: bool) -> Result<()> {
+    let nb = (n as usize).div_ceil(FP8_BLOCK);
+    let kb = (k as usize).div_ceil(FP8_BLOCK);
+
+    // A = all ones (BF16)
+    let a_host: Vec<u8> = (0..k).flat_map(|_| bf16_bytes(1.0)).collect();
+    let a = gpu.alloc(a_host.len())?;
+    gpu.copy_h2d(&a_host, a)?;
+
+    // B = all 0x38 (= 1.0 in E4M3), [N, K] row-major
+    let b_host = vec![0x38u8; (n as usize) * (k as usize)];
+    let b = gpu.alloc(b_host.len())?;
+    gpu.copy_h2d(&b_host, b)?;
+
+    // block_scale [ceil(N/128), ceil(K/128)] FP32, row-major — the layout the
+    // kernel documents and the loader materializes.
+    let mut s_host = Vec::with_capacity(nb * kb * 4);
+    for i in 0..nb {
+        for j in 0..kb {
+            let s = if vary {
+                1.0f32 + i as f32 + 100.0 * j as f32
+            } else {
+                0.00088065f32 // the real Puzzle L0 in_proj weight_scale
+            };
+            s_host.extend_from_slice(&s.to_le_bytes());
+        }
+    }
+    let s = gpu.alloc(s_host.len())?;
+    gpu.copy_h2d(&s_host, s)?;
+
+    let c = gpu.alloc((n as usize) * 2)?;
+    let kern = gpu.kernel("w8a16_gemv", "w8a16_gemv")?;
+    KernelLaunch::new(gpu, kern)
+        .grid([n.div_ceil(4), 1, 1])
+        .block([256, 1, 1])
+        .arg_ptr(a)
+        .arg_ptr(b)
+        .arg_ptr(s)
+        .arg_ptr(c)
+        .arg_u32(n)
+        .arg_u32(k)
+        .launch(stream)?;
+    gpu.synchronize(stream)?;
+
+    let mut c_host = vec![0u8; (n as usize) * 2];
+    gpu.copy_d2h(c, &mut c_host)?;
+
+    // Expected, per row n: sum over k of scale[n/128, k/128].
+    // Each k-block contributes min(128, K - kb*128) terms.
+    let expect = |row: usize| -> f32 {
+        let i = row / FP8_BLOCK;
+        (0..kb)
+            .map(|j| {
+                let width = ((k as usize) - j * FP8_BLOCK).min(FP8_BLOCK) as f32;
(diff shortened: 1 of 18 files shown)


rsafier:
## Serve-level hardware validation on dgx-00 (GB10) — native FP8 fixes the retrieval bug; deterministic repro of the legacy substitution

Built at branch tip `9c9d0b3a` (gb10, `ATLAS_TARGET_MODEL='*'`, release). Model: `nvidia/NVIDIA-Nemotron-Labs-3-Puzzle-75B-A9B-NVFP4`, port 8899, `--gpu-memory-utilization 0.80`, `ATLAS_NO_DECODE_GRAPHS=1` both runs (known Puzzle decode-graph CUDA-700), all probes temp-0 with `enable_thinking:false`.

### Native-path evidence
Run A (default): all **40/40 SSM layers** log
`L{i} SSM quant=Fp8 native_fp8=true native_bf16=false native_prefill=true in_proj_size=18048 d_inner=8192 h=4096` — no `native FP8 unavailable` fallback warning.
Run B (`ATLAS_NEMOTRON_NATIVE_FP8_SSM=0`): 40/40 log `native_fp8=false` (legacy double-quant confirmed).

### Fact probes (5 probes: Paris / 12×8 / P&P 1813 / WW2 1945 / Au)
| run | facts | 10-line Python fn | 400-tok essay | decode |
|---|---|---|---|---|
| A native FP8 | **5/5** | coherent, correct | coherent | 22.6 tok/s |
| B legacy requant | **5/5** | coherent, correct | coherent | 31.5 tok/s |

Simple-fact parity — both arms fine there. Decode is ~28% slower on the native arm (w8a16 FP8 moves 2× the bytes of w4a16 NVFP4 in the Mamba projections; expected cost of the fix).

### The differentiator — context noun-retrieval (the PR's Rufus scenario), temp-0 deterministic
Story (~800 tok) establishing "My dog is named Rufus", then: *"What is my dog's name? Answer with just the name."*

| run | answer |
|---|---|
| A native FP8 | **"Rufus"** ✅ |
| B legacy requant | **"Buster"** ❌ — name substituted from prior, exactly the bug class the PR describes |

Both arms got secondary details right (golden retriever, Elm Street), so the legacy failure is specifically the proper-noun state retrieval the PR targets. Verdict: **the PR's quality claim reproduces on hardware, deterministically at temp-0.**

### Nano-30B boot test (crash history)
`nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4`, default config + `ATLAS_NO_DECODE_GRAPHS=1`:
- **Boots clean and serves** (no CUDA 700/716, no SSD-scan abort — the shm fit-check rider is doing its job). `native_bf16=true` on 6 layers, matching the "6 of 23 BF16" checkpoint description. Decode ~74 tok/s.
- Boot warning worth noting: default KV selection picks **FP8 KV with no k_scale/v_scale in the checkpoint** (server warns scale=1.0 "destroys dynamic range").
- Short-context quality remains broken independent of this PR's paths: facts 1/3 (12×8 → "84"; P&P year → hallucinated titles/years) on default FP8-KV, and still 1/3 (different wrong answers) with `--kv-cache-dtype bf16` — so it is not the KV clipping either. Consistent with Nano's known pre-existing quality state; the boot itself (this PR's claim) passes.

No crashes, no CUDA faults across 5 server boots (Puzzle ×3, Nano ×2); clean shutdowns.

Tested on dgx-00 (GB10, sm_121), single GPU.

tbraun96:

Verified by content, not by ancestry (Atlas squash-merges, and this branch predates the repo's re-root at "Open Source Release", so it has no merge base with `main` and a three-dot diff is meaningless here): `gh pr diff 443 | git apply --check -R -` applies cleanly against `main` @ `a64b82978`, i.e. every hunk of this PR is already present in the tree.

PR #447, opened 2026-08-10
provenance(gdn): reproducible-build script + hash pins + CI check for the committed AOT blobs

rsafier:
Interim guard for the `3rdparty_patches/gdn_aot` binary artifacts, extracted from the .so-replacement research (companion PR: the vendored-link replacement, which supersedes the blobs entirely — see the cross-referenced PR; when that lands, the .so entries in PINS retire with the blobs).

**What this adds**
- `rebuild.sh` — the REAL recipe, recovered not invented: FlashInfer `a671c02e` (same pin as Dockerfile.builder) + the export patch → `gdn_export.py` under CuTe-DSL 4.5.0 (`CUTE_DSL_ARCH=sm_121a`) → shim/transpose link, flags verbatim from Dockerfile.builder lines 99–104. Detects missing toolchain/GPU and names the one VERIFIED export environment (gx10-9959 with the cuda-13.2 compat stack).
- `PINS.sha256` — sha256 pins for both .so files + the export patch, with toolchain provenance (the committed blob's RUNPATH still names the original build venv; gcc 13.3.0 aarch64 — verified against the ELF).
- `.github/workflows/gdn-so-pin.yml` — pure-hash CI check, path-filtered on `3rdparty_patches/gdn_aot/**`: a silently swapped blob is now a red check pointing at rebuild.sh. Empirically verified: passes on the committed bytes, fails on a 1-byte append.
- STATUS.md PROVENANCE section with the honest boundary: **bit-identity was NOT achieved on this box** (the export stage needs gx10's compat stack; a link-only rebuild from a July .o produced different bytes, both hash pairs recorded). The pins certify validated bytes, not source reproducibility — closing that gap needs one gx10 run of rebuild.sh.

Adversarially reviewed (recipe fidelity vs Dockerfile/STATUS line-by-line, CI swap-detection tested empirically, pins re-hashed independently): CONFIRMED; the two remediation-text defects it found are fixed in the follow-up commit.

Part of the #442 follow-up program.

### DIFF
--- a/.github/workflows/gdn-so-pin.yml
+++ b/.github/workflows/gdn-so-pin.yml
@@ -1,0 +1,57 @@
+name: GDN .so provenance pin
+
+# Pure hashing of committed files — no GPU, no toolchain, read-only.
+# Turns a silent binary-blob swap in 3rdparty_patches/gdn_aot/ into a red
+# check with a named remediation (rebuild.sh + PINS.sha256).
+permissions:
+  contents: read
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "3rdparty_patches/gdn_aot/**"
+      - ".github/workflows/gdn-so-pin.yml"
+  push:
+    branches: [main, master]
+    paths:
+      - "3rdparty_patches/gdn_aot/**"
+      - ".github/workflows/gdn-so-pin.yml"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  verify-gdn-so-pins:
+    name: Verify committed GDN binaries match PINS.sha256
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: sha256sum -c PINS.sha256
+        shell: bash
+        run: |
+          set -euo pipefail
+          cd 3rdparty_patches/gdn_aot
+          if sha256sum --check --strict PINS.sha256; then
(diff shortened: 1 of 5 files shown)


rsafier:
**Upgrade: the rebuild pipeline is now VERIFIED end-to-end on gx10-9959** (`90291350`), and the reproducibility question is answered with data:

- Two identical-env runs: `gdn_holo_0.o` → `35671a38…` **both times**, `gdn_holo.so` likewise — **the CuTe-DSL export is deterministic within a fixed environment**. `libatlasgdn.so` drifted run-to-run only via linker build-id — killed with `-Wl,--build-id=none`, now in rebuild.sh.
- The historical three-way drift (pinned blobs / July dgx .o / fresh gx10 .o) is therefore **environment drift, not JIT randomness**. The env archaeology is folded into rebuild.sh: the JIT needs `libcute_dsl_runtime.so` on the loader path AND LD_PRELOADed — the compat stack alone reproduces the same eight-symbol failure a driver-mismatched box gets. (Those eight trampolines are exactly the surface #448's stub replaces — the two PRs cross-validate.)
- Net: pins remain validated-bytes attestations for the historical artifacts (their build env is unrecoverable), but any FUTURE export from the now-documented env is bit-regenerable. For #448's committed `.o` (`ea1c5632…`, the July dgx export), the clean option is a one-time re-export from the documented env + parity re-run so its artifact is reproducible from day one — flagged there.

tbraun96:

Verified by content, not by ancestry (Atlas squash-merges, and this branch predates the repo's re-root at "Open Source Release", so it has no merge base with `main` and a three-dot diff is meaningless here): `gh pr diff 447 | git apply --check -R -` applies cleanly against `main` @ `a64b82978`, i.e. every hunk of this PR is already present in the tree.

PR #448, opened 2026-08-10
build(gdn): link the AOT FlashInfer kernel into the binary — retire the dlopen'd .so blobs

rsafier:
The replacement half of the `libatlasgdn.so` research (companion: the interim pin PR #447 — merging this retires its .so pins; the export-patch pin stays).

**Shape: link-time vendoring.** The committed `gdn_holo_0.o` (AOT-exported GDN kernel, C ABI) is archived into `libatlas_gdn_aot.a` by `spark-model/build.rs` (gated: gb10 + aarch64 + cuda + artifact present → `cfg(atlas_gdn_aot)`) together with the shim and transpose sources, and linked with **static cudart**. The 37 MB proprietary CuTe-DSL runtime dependency is replaced by an 8-symbol stub whose signatures were derived from disassembly of both the runtime's trampolines and the .o's call sites. Both .so blobs are deleted; the dlopen path remains as the fallback arm (`ATLAS_GDN_LIB` still forces it — that's also how the parity test A/Bs the two).

**What this retires from the .so problem list**: the opaque committed binaries, the dlopen side-channel, the proprietary-runtime and compat-lib deployment coupling, and the audit blindness — the linked kernel now records into the boot audit (`--check-kernels` on the 27B: 276 lookups, 0 unresolved, with the AOT row present; 275 with the flag off). Honest boundary, disclosed: the audit row is success-attestation (a flag-on-but-broken link reads as flag-off, not as a failure) — noted as a follow-up.

**Parity, run on GB10 hardware**: stub-linked path vs the original .so through the real CuTe runtime, identical inputs at the Holo shape — **BIT-EXACT** on both scenarios (full T=2048 and chunked 1536+512). The adversarial reviewer independently re-ran the parity binary (2.47s, mismatches=0) and verified non-vacuity (the test binary exports no symbols the dlopen'd reference could interpose on).

Includes the cherry-pick of `5fe12ddf` (commits the .o, cu_seqlens race fix, shim hygiene) — main was already internally inconsistent there: Dockerfile.builder references `gdn_holo_0.o` in its relink step but the tree gitignores it.

fmt / clippy `-D warnings` / tests green; adversarially reviewed: CONFIRMED.

Part of the #442 follow-up program.


rsafier:
Provenance note from the #447 verification work: the CuTe-DSL export turns out to be **deterministic per-environment** (two identical-env gx10 runs → bit-identical `.o`), so the `gdn_holo_0.o` committed here (`ea1c5632…`, July dgx-00 export) could be replaced one-time by an export from the now-documented gx10 environment (`35671a38…`) to make the committed artifact bit-regenerable from day one. That swap would need the parity test re-run (a re-export recompiles the kernel, so bit-exactness vs the old reference .so isn't guaranteed even though the math is unchanged — the test's tolerance policy would need that one decision). Fine to defer — the current .o is parity-proven as-is; this is a provenance upgrade, not a correctness issue. Also: the eight stub symbols here are exactly the trampolines the export JIT resolves from `libcute_dsl_runtime.so` — independently confirmed during the env archaeology.

rsafier:
## Serve-level hardware validation on dgx-00 (GB10) — LINKED kernel engages, dlopen not involved

Built at branch tip `7096462c` (`ATLAS_TARGET_HW=gb10 ATLAS_TARGET_MODEL='*' ATLAS_TARGET_QUANT='*'`, release). Model: `Hcompany/Holo-3.1-35B-A3B-NVFP4` (the validated nk=16/nv=32 gate shape), port 8899, `--gpu-memory-utilization 0.80`, temp-0 probes with `enable_thinking:false`.

### Static: the kernel is in the binary
- `nm target/release/spark | grep -c atlas_gdn` → **4** symbols (linked, not dlopen'd).

### Runtime A/B (default flags, single-stream prefill path)
| run | flags | boot line | prefill-path line | facts | TTFT @1999-tok prompt (max_tokens=2, 3 trials) |
|---|---|---|---|---|---|
| A | (none) | — | `GDN prefill: FLA chunked path ACTIVE` ✅ (baseline confirmed) | 3/3 | 3.592 / 3.581 / 3.573 s |
| B | `ATLAS_GDN_FLASHINFER=1` | `ATLAS_GDN_FLASHINFER: FlashInfer GDN kernel linked-in (AOT, opt-in)` ✅ | FLA line **absent** (never fell through) | 3/3 | 3.393 / 3.384 / 3.390 s |

**Engagement verdict: ENGAGED via the LINKED path.** The boot line is the linked-in variant, not `…kernel loaded (opt-in)` (dlopen) and no `dlopen(...) failed` warning. The FLA once-per-boot fallback banner fires in run A and never in run B — i.e. every GDN prefill in run B went through the FI gate. Under bare default flags the delta is only ~5% because prefill is dominated by the slow default MoE/proj kernels (the known silent-slow fallback).

### Runtime A/B (production FAST config: LOW_MEMORY_MOE + grouped-CUTLASS + FLASHINFER_PREFILL + `ATLAS_PREFILL_CODISPATCH=1 ATLAS_PREFILL_VARLEN=1 ATLAS_Q12_BATCHED=1` etc., kv bf16)
| run | GDN flag | warm TTFT @1999 tok | facts |
|---|---|---|---|
| C | off | 0.731 s (~2.7K tok/s) | 3/3 |
| D | `ATLAS_GDN_FLASHINFER=1` | **0.547 s** (~3.7K tok/s) | 3/3 |

**1.34× prefill under the production co-dispatch config** — matches the historical 1.30–1.34× e2e claim, and run D again shows zero FLA-fallback banners. The historical #220 co-dispatch bypass is not reproduced on this branch: the FI gate now sits ahead of the FLA ladder in both `trait_prefill_recur.rs` (single-stream) and `trait_prefill_gdn.rs` (batched/co-dispatch loop), and the measured TTFT delta under `ATLAS_PREFILL_CODISPATCH=1` confirms it is live there.

### Quality
- Facts 3/3 in all four runs (Paris / 96 / 1813), 400-token essay coherent, decode ~54 tok/s (default flags) both arms.
- No errors, no CUDA faults, clean boots (25s cold / ~15s warm) and shutdowns.

Tested on dgx-00 (GB10, sm_121), CUDA 13.x driver stack, single GPU.

tbraun96:
Reopened — my mistake. The Batch 2 merge commit ([`03ec583`](https://github.com/Avarok-Cybersecurity/atlas/commit/03ec58326)) says "Closes #448" and its title names this PR, but **this PR's content is not in that stack and is not on `main`**. I wrote the merge message from the stack's original composition without re-checking what actually got built. Apologies for the noise.

## Why it was excluded (this part is a real finding, not a process slip)

It conflicts with #447 in a way git does not flag, because the two touch different files:

- #448 deletes both `.so` blobs that #447's `PINS.sha256` pins, so `sha256sum --check --strict` exits 1 on `libatlasgdn.so` and `gdn_holo.so` and takes `gdn-so-pin.yml` red.
- #447's `rebuild.sh` then rebuilds artifacts that no longer exist.
- The substantive part: #448 commits `gdn_holo_0.o` = `ea1c5632…`, which #447's own `STATUS.md` identifies as the **dgx-00 July-env artifact** — distinct from the `35671a38…` that #447's verified pipeline reproduces.

The only *marked* conflict is a trivial `STATUS.md` append. Everything above is invisible to a three-way merge.

Deciding which bytes are authoritative and rewriting another author's pin file is not a call to make on someone else's behalf, so this needs a rebase over #447 from you rather than a guess from us.

Note #447 has since landed on `main`, so the rebase target exists now. Once rebased with the pin question settled, this can ride the next gate cycle — it is build/link plus prefill, so it is leg-distinguishable and batches cleanly.

rsafier:
Researched the alternatives to this PR today and they are all worse — **this is the right shape**, and it is worth more than the packaging convenience it currently reads as.

## The dependency this removes costs 40–50% of prefill when it silently fails

Measured on a GB10 today, Holo-3.1-35B-A3B-NVFP4, same binary and flags, the only difference being whether the proprietary CuTe-DSL runtime resolved at `dlopen` time:

| ISL / conc | FLA fallback | GDN loaded | delta |
|---|---|---|---|
| 3k C=1 | 3292 | 4602 | +40% |
| 3k C=8 | 3801 | 5580 | +47% |
| 8k C=1 | 5817 | 5809 | ~0 |
| 8k C=8 | 3951 | 5919 | +50% |

(aggregate prefill tok/s; the fallback column is two independent boots agreeing within 1%.)

Note the shape: at C=1/8k the fallback is already fine, so **a single-stream smoke test cannot detect this**. The entire loss sits at concurrency. It went unnoticed on a recipe that documents the log line to check, because setting `ATLAS_GDN_FLASHINFER=1` is not the same as engaging the kernel — a failed load logs one WARN and serves.

It took two independent fixes to make the library resolve at all: an absolute `ATLAS_GDN_LIB`, **and** putting `libcute_dsl_runtime.so`'s directory on `LD_LIBRARY_PATH`, since it ships in the `nvidia_cutlass_dsl` Python package rather than with CUDA. This PR deletes that whole class of failure.

## The obvious shortcut is not available: vendoring the runtime is prohibited twice over

Worth recording so nobody re-proposes it. The 8 needed symbols are all present in a 16 KB static archive (`libcuda_dialect_runtime_static.a`) shipped beside the 37 MB `.so`, and we proved statically linking it works end-to-end, including a non-vacuity control. But:

- NVIDIA's license grants distribution only of *"python files … in source format"* (§1.1(d)), and §2.2 forbids distributing "any portion of the Software" otherwise. A 16 KB ELF archive is neither Python nor source.
- §2.7 separately forbids use "in any manner that would cause them to become subject to an open source software license", with §6 defining that by OSI/FSF/**SPDX** — and Atlas is `AGPL-3.0-only`. Symmetrically, AGPL §1 cannot treat a pip wheel as a System Library.

Not legal advice and worth a counsel read before anyone relies on it, but the two clauses are explicit enough that I would not plan around vendoring. **The stub approach in this PR sidesteps both**, because nothing NVIDIA-proprietary is redistributed.

## Two things worth fixing before merge

**1. `_cuKernelGetAttribute` may be missing an ordinal→handle translation.** The stub forwards its `dev` argument straight through; the reference implementation appears to resolve it via `cuDeviceGet(&device, ordinal)` first. In practice `CUdevice` equals the ordinal on every known driver and GB10 is single-GPU per rank, so this is a latent exactness gap rather than a live bug — but it is a ~3-line fix to match exactly. Flagging as *reported, not independently confirmed by me*.

**2. The provenance wording in the stub header, the PR body and `gdn_aot_parity.rs` should be revised.** As written it characterises how the signatures were obtained in a way that is unnecessarily unhelpful given the license terms above. The header already cites a perfectly good independent basis as its reason #1 — the call sites in our own generated `.o`, plus public CUDA headers, plus the bit-parity gate — and that basis alone fully supports the signatures. Recommend resting on it and dropping the rest. No code change, no loss of evidence. Happy to go through the specific wording off-thread.

## One follow-up this PR already names, now with a number on it

The description discloses that the audit row is success-attestation — *"a flag-on-but-broken link reads as flag-off, not as a failure"*. That is the same fail-open class as the measurement above, which is now quantified at 40–50% of prefill. This PR removes most of the surface by deleting the `dlopen`, but promoting that follow-up from "noted" to "done" looks worth an hour: if `ATLAS_GDN_FLASHINFER=1` is set explicitly and the path does not engage, that should be loud.

The bit-parity gate here (`tests/gdn_aot_parity.rs`, both paths in one process, bit-identical required, full-sequence and chunked) is the right standard for the swap.

Branch is `MERGEABLE` and verified to merge-and-build against post-#490 `main`; it is unreviewed and `BEHIND`.

PR #450, opened 2026-08-10
Stack: dep bumps (#355), dflash drafter (#394), GDN provenance (#447), gemma4 jinja (#324), nemotron FP8 SSM (#443)

tbraun96:
Batch 1 of the PR-backlog clearance. Five open PRs merged in order onto `origin/main`
(c2f23590) with **no conflicts**, so one gate cycle discharges all five.

## Constituents

| PR | Author | Title | Why it cannot move the gated numbers |
|----|--------|-------|--------------------------------------|
| #355 | @dependabot | chore(deps): bump the patch-updates group (9 updates) | `Cargo.lock` only — patch-level bumps of host-side crates; no Atlas source, no kernel, no serve path changes. |
| #394 | @rrstesiak | fix(dflash): use drafter `target_layer_ids` directly | One hunk in `factory/build.rs`, entirely inside `if let Some(ref args) = dflash_args`. The gated recipes (`qwen3.6-27b-nvfp4-unsloth`, `qwen3.6-35b-a3b-fp8-bf16head`) run MTP speculation, never DFlash, so `dflash_args` is `None` and the branch is not entered. Plus a `_typos.toml` allow-list entry. |
| #447 | @rsafier | provenance(gdn): reproducible-build script + hash pins + CI check | `3rdparty_patches/gdn_aot/` + a new workflow. Documents and verifies the ALREADY-COMMITTED AOT blobs; it does not rebuild or replace them, and it adds no engine code. |
| #324 | @drewhemm | fix(jinja): prevent infinite tool-calling loops in gemma4 multi-turn | `jinja-templates/gemma4.jinja` only. The server loads `<model_type>.jinja`; gemma4 is not a gated model, so neither gated checkpoint ever reads this file. |
| #443 | @rsafier | fix(nemotron): load native FP8 SSM weights — stop the Mamba2 double-quant | Nemotron weight-load / Mamba2 / MoE paths and two nemotron `MODEL.toml`s + one nemotron kernel. The one shared file, `layers/ops/ssm_ssd.rs`, is a pure extract-function refactor (`ssd_scan_smem` / `ssd_scan_fits`) computing the identical expression, and its only callers are `nemotron_mamba2*`. Nemotron is not a gated model. |

## Local verification at 0553905e

- `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy --workspace --tests` — clean, 0 warnings
- `cargo test --workspace` — **4054 passed, 0 failed, 61 ignored** across 75 test binaries
- 500-LoC cap: every changed `.rs` under `crates/` is <= 500 except `factory/build.rs` (643), which is already allow-listed in `file-size-cap.yml` and which #394 *shrinks* by 15 lines.

## Gate

All five required gate records are invalidated by this diff (it touches `crates/`,
`kernels/`, `Cargo.lock` and `jinja-templates/`), so a full five-leg cycle is being run
against the frozen tip and the records will be committed here.

Because no member can reach either gated config, a red leg on this stack is flake by
construction rather than content.


tbraun96:
Correction to the merge commit message (77daac04) — the content is right, two lines are garbled and I cannot amend a protected branch, so the fix goes here.

1. Line 1 reads "Gated at e5d8633cd9... no — at 0553905e0a". That is a self-correction I accidentally left in. **The gated sha is 0553905e0a**; e5d8633cd9 was #444's, from the previous cycle.
2. One sentence lost its subject to a shell-quoting slip. It should read: *"That is not an argument — the gate tool confirms it. The PR gate check reports device code changed for exactly TWO targets, gb10/nemotron-3-nano-30b-a3b/nvfp4 and gb10/nemotron-labs-3-puzzle-75b-a9b/nvfp4."*

Every number, verdict and threshold in the message is unaffected. Recorded here rather than rewritten, since main is protected against force pushes for good reason.

PR #457, opened 2026-08-10
Stack: generic GGUF loader + Ternary-Bonsai-27B (#334) + MoE LoRA & overlay (#335)

tbraun96:
# Batch 3: GGUF loader + Ternary-Bonsai-27B (#334) stacked with MoE LoRA (#335)

This stack **supersedes #334 and #335** — it is their content, three-way merged
onto current `main` (a64b82978), CI'd as one unit, and is the branch to gate.

## Constituents

- **#334** `port/gguf-bonsai-avarok` (head 448ed8c28): generic GGUF loader +
  Ternary-Bonsai-27B, 2-bit keep-packed Q2_0 on one GB10 (Q2_0 GEMV/MMQ
  kernels, packed-QKVZ fusion, GGUF container/name/value-transform parsing,
  BufferArena Q2 scratch).
- **#335** `port/lora-moe-avarok` (head 681b425a4): MoE per-expert + router
  LoRA and embed/lm_head/vocab overlay, correctness-first (BGMV gather /
  grouped-down kernels, token overlay, per-request `MoeLoraRoute` fold gate,
  adapter oracle scripts + parity test).

Both are fork PRs with no merge base against the re-rooted `main` (the #452
orphan class), so each was merged content-wise with an explicit base:
`git merge-tree --merge-base=<last main state merged into the PR>` —
c2f235908 (#449) for #334, 4e34a9e7b (#410) for #335.

## Merge resolutions (all read, none taken wholesale)

1. `crates/spark-model/src/layers/qwen3_ssm/mod.rs` — union: #334 adds
   `mod init_fp8; mod init_q2;`, #335 adds `mod lora;` at the same anchor.
2. Cross-PR interaction caught by clippy: `qwen3_ssm/tests.rs`'s
   `run_batched_verify` helper (main-side, post-#410) predates #335's new
   `ForwardContext.moe_lora_route` field. Fixed with the documented default
   (`MoeLoraRoute::Fold`, inert on verify paths), same convention as #335's
   own `layer/tests.rs`.

## Local CI (dgx1, this branch, ATLAS_TARGET_MODEL="*")

- `cargo fmt --all -- --check` — PASS
- `cargo clippy --workspace --tests` (ATLAS_SKIP_BUILD=1, -D warnings) — PASS
- `cargo test --workspace --locked` with the full nvcc kernel build — PASS:
  73 suites, 4249 passed / 0 failed / 66 ignored (GPU-gated). One intermittent
  failure seen in 1 of 6 full-suite runs is PRE-EXISTING on main a64b82978:
  `tui::render::selection_tests::a_drag_highlights_what_it_covers_and_a_click_highlights_nothing`
  — a wall-clock race (the startup header ticks "0.0s" -> "0.4s" between the
  two renders the assertion compares). Reproduced verbatim on an unmodified
  main checkout; this stack does not touch `crates/spark-server/src/tui`.
- `typos` — PASS (0 findings)
- `scripts/check_spdx.py` — PASS; `scripts/check_kernel_shadows.py` — PASS
- kernel-compile job equivalent: real `target_ptx.rs` present, 3728 PTX
  modules compiled (assertion floor is 50)
- 500-LoC cap: only `crates/spark-runtime/src/buffers/sizes.rs` (524) is newly
  over-cap; #334 itself allow-lists it with rationale in
  `.github/workflows/file-size-cap.yml`.

## Gate requirement (`spark benchmark --pull-request-gate-check` at 8c2f00858)

All FIVE gates are owed — the closure hash excuses nothing:

```
NONE  agentic-webserver — latest record is for 0553905e0a — invalidated by
      crates/atlas-core/src/config.rs, ... and 173 more — device code changed
      for 22 target(s)
NONE  ttft-warm-gate    — (same)
NONE  ttft-cold-gate    — (same)
NONE  bfcl-subset       — (same)
NONE  bfcl-subset-echolp — (same)
5 bench(es) still need a passing gate record
```

The new `kernels/gb10/common/*.cu` files enter every target's closure, so all
22 targets re-open; the 170+ `crates/` paths are outside what the device-code
hash can excuse anyway.

## Not merged, no GPU work

Prepared for gating only — benchmark gates have not been run; no GPU process
was started for this stack.


rsafier:
Ran serve-level verification of this stack on dgx-00 (tip `8c2f0085`, Qwen3.6-35B-A3B-NVFP4, the two #335 serve bugs I'd reported as the target). Three findings, one of which blocks your gate run:

**1. BLOCKS GATING — stock 35B-A3B NVFP4 dies at boot on this stack** (fix pushed to #334's branch, `51cfcd34`). `Qwen3SsmLayer::new` probes the GGUF keep-packed families unconditionally, but `q2_0_mmq`/`q4k_mmq` ship only in `qwen3.6-27b/nvfp4` — the fail-closed audit kills the boot with 3 unresolved lookups (`atlas_q2_0_mmq128_{nc,wc}`, `atlas_q8_1_quantize_ds4_bf16`). The **bfcl-subset-echolp leg serves exactly this target**, so gating would fail at boot. Fix = the standard `[expected_absent]` declarations in 35b-a3b's MODEL.toml (all dispatch sites are behind handle!=0 + packed-weight presence checks; a safetensors checkpoint never installs packed-Q2). Verified on a rebuilt release: 22s boot with no escape flag, 0 unresolved, temp-0 fact correct. One commit — cherry-picks clean onto this stack.

**2. The decode_a2 fold port works, but the default path still batch-kills at n≥4.** With a resident MoE adapter and 6 concurrent requests on default flags, one reject 500s the whole batch — original symptom intact: `decode_batch error: MoE LoRA (Feature-1) is prefill-only in phase 1`. Root cause: the attention-layer multi-seq FFN ladder at n≥4 on MoE falls to the per-token loop (`qwen3_attention/trait_impl/multi_seq/ffn.rs:218` → `MoeLayer::forward` refusal at `moe/forward.rs:47`) because the fold-capable `forward_batched` arm is env-gated at `ffn.rs:202` behind `ATLAS_MOE_BATCHED_DECODE=1`. **With that flag set: 6/6 pass, zero errors, and per-row adapter identity is correct** (base rows clean, adapter rows perturbed — no cross-row contamination). n=1–3 are fine by default. Per the house rule this wants default-ON with a kill-switch (or a CLI lever) rather than an opt-in env var — your call whether that's in-stack or follow-up.

**3. Good news — the "fold silently not reaching output" bug was (at least partly) my test's fault.** With the generator's default `LORA_STD=0.02` the adapter genuinely doesn't move this model's greedy argmax — byte-identical is *expected*. At `LORA_STD=0.3` base/adapter outputs diverge at char 0, the fold demonstrably reaches output in both prefill and decode, and the MoE-LoRA kernels show "used" in the kernel table. So item (2) is the one real remaining defect.

Repro assets (scripts, serve logs, strong adapter) preserved on dgx-00 under the job tmp dir; happy to re-run any leg on demand.

tbraun96:
## Gate cycle aborted — #334 breaks serve for every non-GGUF model

Started the five-leg cycle on `3bab82d7ed` (#334 + #335 + #456) at 21:41. **Both boxes failed in under a minute.** The kernel gate refused to start the server:

```
Error: the server failed before it began serving "Qwen/Qwen3.6-35B-A3B-FP8"

Caused by:
    3 unresolved kernel lookup(s) for (qwen3.6-35b-a3b, sm_121, nvfp4). Each one
    resolved to handle 0, so its dispatch site is on a silent fallback path:
      1. q2_0_mmq::atlas_q2_0_mmq128_nc  at crates/spark-model/src/layers/qwen3_ssm/init.rs:44
      2. q2_0_mmq::atlas_q2_0_mmq128_wc  at crates/spark-model/src/layers/qwen3_ssm/init.rs:45
      3. q4k_mmq::atlas_q8_1_quantize_ds4_bf16  at crates/spark-model/src/layers/qwen3_ssm/init.rs:46
```

`qwen3_ssm::init` issues these three lookups **unconditionally**, so they fire for every model built through that path — including a 35B MoE that has no Q2_0 weights and never will. `q2_0_gemv_vec` and `dequant_gguf_bf16` next door resolve fine; only the two `q2_0_mmq` entries and the `q4k_mmq` quantizer are absent from these targets' kernel sets.

This is the silent-fallback class the gate exists to catch: the lookup succeeds structurally, returns handle 0, and the dispatch site quietly takes a slower or different path with nothing in the logs.

### Also worth noting for whoever reads the logs

dgx3's `bfcl-subset-echolp` reported **`rc=0` after 57 seconds** for what is a ~2.5 hour leg — same failure, clean exit code. Anything that trusts exit status would have recorded that as a pass.

### The fix is #334's call, not ours

The error names three legitimate routes, and choosing between them is a design decision about which of the 22 targets should carry Q2_0 kernels:

1. **Gate the lookup on model config** so it is never issued — the pattern `qwen3_attention::init` already uses. Probably right: a 35B MoE asking for Q2_0 is meaningless.
2. Fix the build so the kernels compile for these targets.
3. Declare them in each target's `MODEL.toml [expected_absent]` with a stated reason — accurate but touches ~22 files.

We are not taking `--dangerously-allow-unresolved-kernel-lookups`; that is precisely the silent degradation the gate is preventing.

### What we are doing tonight

Dropping #334 and #335 from this cycle and gating **#456 alone**, which is host-only (process exit status on a poisoned-CUDA-context path) and cannot be affected by this. #457 stays open pending the gating fix.

Everything else on this stack still stands: CPU CI was 32/32 green, and the `#[ignore]` GPU correctness anchors passed on their first-ever run — MoE-LoRA parity bit-exact at **max_ulp 0**, GGUF dequant **byte-identical** across all five formats, Q2_0 GEMV within `rel_err 0.0026` of a 1e-2 gate. The port's numerics are sound; it is the unconditional registration that breaks other models.

rsafier:
Route 1 implemented and pushed to #334's branch: `528b2c72` (supersedes my earlier `51cfcd34` — the `[expected_absent]` declarations are reverted in the same commit, so the other ~20 target files stay untouched).

**What it does:** all three Q2-capable layers probed the keep-packed MMQ trio unconditionally in their constructors — `qwen3_ssm/init.rs` (your trace), `qwen3_attention/init.rs:204-206` (surfaces next once the SSM one is gated — construction order was hiding it), and `dense_ffn.rs` (fires on every dense-FFN model, which is where the ~22-target blast radius came from). The probes now live in the packed-Q2 install setters (`set_packed_q2_qkvz`, `set_packed_q2_weights`, `set_q2_weights`), so the lookups are only issued when packed-Q2 weights actually exist — the `qwen3_attention::init` pattern the audit message points to. `dense_ffn`'s pre-existing `q4k_mmq` quantizer probe (on main before this PR) is untouched.

**Serve-verified on GB10, both directions:**
- stock 35B-A3B NVFP4 (the bfcl-echolp target): boots 22s, **no escape flag, 0 unresolved**, temp-0 fact correct
- Ternary-Bonsai keep-packed: kernels still resolve via the setter — 44s boot, 3/3 temp-0 facts, 19.1 tok/s, keep-packed log evidence, zero errors

fmt/clippy clean, 346 spark-model tests pass. One commit on `port/gguf-bonsai-avarok` — should cherry-pick clean onto this stack whenever you re-cut it.

rsafier:
Pushed the gating fix onto this stack directly (`3bab82d7..8e712d54` — the two cherry-picks from #334's branch, `ebfb8e63` + `8e712d54`, net-zero on the MODEL.toml). `cargo check` green on the stack tip. The gate cycle should now clear the 35B boot; happy to re-run the serve legs here if you want fresh receipts on this exact tip.

rsafier:
Started producing gate records for this stack on dgx-00 (tip `8e712d54`, records pushed as `e650094a`):

- **ttft-warm-gate: PASS** (3m45s) and **ttft-cold-gate: PASS** (2m24s) — records committed; first run on this box establishes its TTFT baseline.
- **agentic-webserver: not recordable on dgx-00, and it's not the stack.** Twice in a row the leg scored 9/10 with an *identical* failure: seeded **run 4** degenerates deterministically (writes Cargo.toml three times, falls into a repetition loop — "I'll create the source file with the ping/pong endpoint and tests:" ×3 — and stops without writing `src/`; 0/6 in ~25s, 5 turns). Discriminator: an **unmodified main build (`a64b8297`) reproduces the same run-4 signature byte-for-byte on this box** while your 2026-08-10 dgx1 record is 10/10 — a per-box greedy-trajectory flip, not a stack regression. That leg needs a box where run 4's trajectory holds (dgx1 evidently qualifies). Trajectory kept at dgx-00 `~/.atlas/runs/agentic-webserver/sandbox/run-04.trajectory.txt` if you want to compare against dgx1's run 4.
- **bfcl-subset + bfcl-subset-echolp: running now** on dgx-00, serial (C2 dense-27B coherence smoke ran first). Records will be pushed to this branch when they land (~6h).

So after tonight the stack should owe only the agentic-webserver record from one of your boxes.

rsafier:
Overnight gate results — **4 of 5 legs recorded and passing** on dgx-00, records pushed to this branch (`130564d0`):

| Leg | Verdict | Detail |
|---|---|---|
| ttft-warm-gate | **PASS** | 3m45s, baseline established for this box |
| ttft-cold-gate | **PASS** | 2m24s |
| bfcl-subset (27B) | **PASS** | overall **86.83** (floor 83.64), normalized 88.38, n=995 |
| bfcl-subset-echolp (35B) | **PASS** | overall **86.55** (floor 83.64), normalized 86.95, n=1004 |
| agentic-webserver | *owed from dgx1* | per-box run-4 trajectory flip (reproduced on unmodified main here — details upthread) |

Ops notes: the 27B leg's first pass completed all inference but died in scoring — this box's bfcl venv was missing `soundfile` (qwen_agent import chain); the kept `responses.jsonl` rescored to the identical 86.83 before the re-run, so the re-run was record-production only. Worth preinstalling in the venv bootstrap if other boxes are fresh. Gate runner also refuses to start below a free-memory floor — page cache from prior serves counts, first attempt hit it at 73%.

So this stack needs exactly one more record — agentic-webserver from a box where run 4 holds (your 2026-08-10 dgx1 record shows it does there) — and it's fully gated.

PR #462, opened 2026-08-11
Nemotron-3.5 Lightning 30B support + the Nemotron-H NoPE fix (family long-context retrieval)

rsafier:
## What

Three commits, each independently valuable:

1. **transformers-5.x config parsing** (Lightning is the first 5.x-era checkpoint we serve):
   - `layers_block_type` string-list form of the hybrid schedule (5.x drops `hybrid_override_pattern`) for plain `nemotron_h`
   - **quantization-config precedence**: 5.x embeds a compressed-tensors mirror in config.json whose `config_groups.group_0` describes only ONE group of a MIXED_PRECISION checkpoint (Lightning: the FP8 mamba group). The early-return in `merge_sidecar_quant_config` kept that block and mislabeled the whole model FP8. The ModelOpt sidecar now wins whenever it exists and parses. This bites every future 5.x ModelOpt checkpoint, not just Lightning.
   - `moe_latent_size: null` → 0 (5.x writes null for absent)
   - wires `config/tests/tests_b.rs` into the build — it was split for the LoC cap but never declared; its four tests (including a `test_parse_nemotron_h_config` twin) silently never ran
2. **The NoPE fix — Nemotron-H applies no rotary embedding in attention; Atlas did.** Reference `NemotronHAttention` projects q/k/v straight into attention (position is carried by the mamba layers; `rope_theta` in config.json is vestigial; Puzzle's shipped modeling code is likewise rope-free). Atlas served every `nemotron_h` model with `rotary_dim=128`, scrambling long-range q·k addressing: measured needle-recall died at ≥52 tokens on BOTH Lightning and Nano (answers rode SSM recency; parametric fluency masked it — this is the long-open 'Nano long-ctx retrieval broken' defect). Fix: nemotron_h dispatch sets `rotary_dim = 0`; the plain rope host wrappers treat 0 as NoPE and skip the launch (the kernel divides by rotary_dim, so the guard is host-side).
3. **`[behavior].no_decode_graphs`** — the Nemotron decode-graph crash (CUDA 700/716) was previously suppressed with `ATLAS_NO_DECODE_GRAPHS=1`, which #388's env→CLI conversion removed with no replacement; the family crashed in default serve (reproduced: argmax_batch dies after first capture). New MODEL.toml behavior key, gated at every capture site, set for the three Nemotron targets. Graphs are a measured no-op on GB10.

Lightning intentionally routes to the `nemotron-3-nano-30b-a3b` kernel target (dimensionally identical; the `(model_type, hidden_size)` matcher cannot split them — documented in that MODEL.toml).

## Ground truth & measurements (GB10, dgx-00)

- Hosted NIM reference (`nvidia/nemotron-3.5-lightning-30b-a3b`): **12/12** on the fact/retrieval battery at temp 0 and 1.0
- The same NVFP4 bytes under vLLM (aeon sm_121a image) on this box: **12/12** — checkpoint exonerated
- Atlas pre-fix: needle-recall dead ≥54 tokens, chat battery **0/6**, greedy next-token vs the vLLM oracle 5/22 with word-fragment garbage
- Atlas post-fix: needle-recall perfect at every probed distance to **334 tokens** (incl. interference-heavy filler that broke recall at 26 tokens pre-fix); chat battery **5/6** at the card's recommended sampling; oracle battery 10/22 exact with every remaining diff a near-tie greedy alternative. Nano needle-recall: dead ≥52 → perfect through 134.
- Default serve (no workaround flags): boots 24s, ModelOpt MIXED detected, FP8 mamba native (#443 path), zero audit failures, no graph captures, the previously-crashing request sequence completes clean.

Serving note: Lightning is reasoning-first; until the shared target's `thinking_default` question is settled, serve it with `--default-chat-template-kwargs '{"enable_thinking":true}'`. Two Nano band-aids in the shared MODEL.toml (`thinking_default=false` rationale and `default_system_prompt`) document symptoms that match the rope bug — worth re-measuring on Nano post-fix, left untouched here.

## Checks

fmt ✓ · clippy --workspace --tests 0 errors ✓ · cargo test --workspace 4084 passed / 0 failed ✓ · typos ✓ · SPDX ✓ · ≤500-LoC (build_parse split into build_behavior) ✓ · tests that fail without each change (layers_block_type list, sidecar precedence, rotary_dim==0 assertions) ✓

crates/ + kernels/ touched → this owes the 5 gate records; happy to produce the four dgx-00-recordable legs like on #457 (agentic-webserver needs a box where seeded run 4 holds).


rsafier:
Added MTP speculative decoding for Lightning's built-in draft head (`ba01f3ec`): standalone `NemotronMtpHead` proposer (DeepSeek-style enorm/hnorm/eh_proj combiner; delegates its attention + MoE blocks to the same TransformerLayer impls the backbone runs — zero new kernels), gated the sanctioned way (`MTP_SUPPORTED_MODEL_TYPES` + MODEL.toml `default_num_drafts`; engagement still requires `--speculative`).

Along the way it fixed **latent family verify bugs**: the SSM checkpoint/rollback byte-sizes were GDN-shaped (zero on Mamba2 — every rollback copy was silently 0 bytes), now sized from the ssm_pool SSOT, plus a Mamba2 `decode_batched` that snapshots per-token state for rollback.

Measured (GB10, C=1): drafts are good — K2 acceptance p1 ≈ 0.55 — but **end-to-end is currently net-negative** (70.1 serial vs 55.0 forced-MTP tok/s): the verify side lacks fused K-token Mamba2 and M-generic ReLU²-MoE kernels, so K=2 verify pays ~2× weight reads for ~1.55 tokens (the GDN 27B's +61% rode its fused WY verify kernels). The auto MTP gate measures this live and parks speculation, so `--speculative` never regresses serving; it re-probes and will flip positive the day those verify kernels land. Facts stay correct under forced MTP with ~50% rejects — the new rollback path gets exercised hard.

rsafier:
Merged `main` in. Flagging the resolution because the hazard is one CI structurally cannot see, and it is not specific to this PR.

**The merge is textually clean and the result does not compile.**

#473 split `[behavior]` parsing out of `build_parse.rs` into a new file, `build_parse_behavior.rs`, wired as a child module of `build_parse`. This branch had independently split the same table out into `build_behavior.rs`, wired as a top-level module of `build.rs`. Neither side touches a line the other touches, so git has nothing to resolve — it simply lands both parsers. `build.rs`'s `use build_behavior::parse_behavior` then shadows main's, and the five fields main has added since resolve against the wrong `ParsedBehavior`:

```
error: unused import: `behavior::*`
error[E0609]: no field `disable_cwd_hint_injection` on type `build_behavior::ParsedBehavior`
error[E0609]: no field `use_sampling_presets_for_core` ...
error[E0609]: no field `enable_think_loop_watchdog` ...
error[E0609]: no field `honor_eos_inside_thinking` ...
error[E0609]: no field `cap_thinking_at_max_tokens` ...
```

Nothing was in a position to catch this. The checks run against the PR head, and GitHub reports the PR `MERGEABLE` because the textual merge succeeds — the breakage exists only in the merge result, which nothing builds. All seven required checks were green on this PR the whole time it was broken.

**Resolution:** main's split wins. Deleted `build_behavior.rs`, ported `no_decode_graphs` into `build_parse_behavior.rs`, pointed `build.rs` at `build_parse`'s re-export. This is byte-identical to the reconciliation already carried on #487, so the two branches now agree on the file.

Verified on the merge result, not the head: workspace `check` + `clippy` clean, 4603 tests pass, and the all-targets build (`ATLAS_TARGET_MODEL='*'`) resolves the behavior key for all three Nemotron `MODEL.toml`s that set it.

Same hazard was present on #467 and #475 (both descend from this branch); both are fixed and pushed. #475 had a second, worse one — see there.

rsafier:
End-of-day status.

Verified today that this branch **merges and builds clean against post-#490 `main`** — I merged main into a scratch copy and ran `cargo check --workspace --tests` on the merge *result*, not the head. That check matters here: a textual merge succeeding is not evidence, as this PR itself demonstrated earlier today when a conflict-free merge with main produced six `E0609`s.

Same sweep run against all eight open PRs in the stack; every one builds clean on the new base.

CI is green apart from `PR benchmark gate` (records owed).

**Suggested as the first of the Nemotron stack to land.** Beyond Lightning support, it carries the NoPE family fix — Atlas was applying rotary embedding to the whole `nemotron_h` family, which kills long-context retrieval above ~52 tokens, so Nano, Puzzle and Super all benefit. Puzzle was re-assessed on this branch yesterday and went from unusable to passing needle 18/18 out to 25 084 tokens, with the CUDA-700-at-exactly-64-prompt-tokens crash fixed by the `no_decode_graphs` key that ships here too.

It is also the base of #467 → #475 and a sibling of #487, so landing it first collapses the diffs of everything above it and shrinks the surface one gate wave has to cover.

rsafier:
Closing in favour of #544, which lands this work re-ported onto current `main`.

Three of this PR's four substantive commits are carried into #544, re-applied against main rather than rebased (main replaced the kernel-target resolution API the stack on top of this one extends — see #487 for that detail):

| commit | in #544 |
|---|---|
| `d033913d` parse transformers-5.x configs | **yes** — `layers_block_type`, `num_nextn_predict_layers`, `nullable_usize` for `moe_latent_size`, and the ModelOpt-sidecar-wins rule |
| `a6c075b7` NoPE | **yes** — including the two tests that asserted `rotary_dim() == 128` and now assert `0` |
| `cd63fd37` `[behavior].no_decode_graphs` | **yes** — re-plumbed through main's behavior parser |
| `ba01f3ec` MTP speculative decoding (Lightning's draft head) | **NO — not ported** |

⚠ **`ba01f3ec` is the remaining delta.** Lightning serves without it — the reference recipe pins `--speculative` OFF, and #544 is verified end-to-end in that configuration — but the draft-head support is real work and is *not* in #544. It is preserved on this PR's branch (`feat/nemotron-lightning-30b`), which is not deleted by closing, and should come back as its own PR on top of #544 if wanted.

The sidecar rule from `d033913d` deserves a note, because it is what separates "boots" from "serves": transformers 5.x embeds a compressed-tensors `quantization_config` whose `config_groups.group_0` describes only ONE group of a mixed-precision checkpoint (for Lightning, the FP8 mamba projections). With an embedded-wins rule the whole model is labelled FP8, NVFP4 detection breaks, and the failure is silent — the server boots, logs every config line correctly, and then stalls in prefill at 0.9% CPU. I rediscovered that the hard way while re-porting.

#544 is verified on a GB10 against the atlas-recipes#18 profile: correct target resolved without `--kernel-target`, `rotary_dim=0`, graphs disabled, `thinking_default=true`, 76.5 tok/s decode, and needle retrieval PASS at 10/50/90% depth at both 61k and 124k prompt tokens.

AzeezIsh:
Verified the NoPE fix end-to-end on a GB10 against current `main` today (via the #487 tip, which contains this branch). Boot logs the expected line:

```
Model config: 52 layers, 6 attention, 23 SSM, 128 experts, rope_theta=10000, head_dim=128, rotary_dim=0
Decode graphs disabled by MODEL.toml [behavior].no_decode_graphs
```

One rebase caution on the `[behavior].no_decode_graphs` half, learned the hard way. The key is parsed and threaded fine, and six decode/verify paths gate on `!self.config.no_decode_graphs` — but the single assignment that connects them,

```rust
config.no_decode_graphs = ptx_set.behavior.no_decode_graphs;
```

lives in `crates/spark-server/src/main_modules/serve_load.rs`, which is one of the files that conflicts against `main` (the `ptx_for_shape` → `ptx_for_config` migration). Resolving that file to `main`'s side — the correct call for every *other* hunk in it — drops the assignment. It then defaults to `false`, graphs stay enabled, and nothing errors or logs; a short serve looks perfectly healthy because the crash is prompt-length dependent.

Suggest keeping "`Decode graphs disabled by MODEL.toml [behavior].no_decode_graphs` appears at boot" as an explicit post-rebase check, since it is the only observable signal that the kill-switch survived.

PR #464, opened 2026-08-11
Stack: DeepSeek-V4 E8M0 scales (#400), prefix-cache safety (#401), contract-only (#402), Laguna-S-2.1 (#350), DSML protocol (#403)

tbraun96:
Batch 4 stack on post-#461 main: #400 + #401 + #402 + #350 + #403, merge-validated onto 5b324373d (clean except 8 additive conflicts resolved keep-both: GGUF-Q2/LoRA from #457 × batch features; ms_phase_o_proj extraction kept with main's packed-Q2 o_proj branch ported in).

cargo check --workspace clean. Gate cycle 5/5 legs RUNNING on dgx1 at frozen sha b9bbb3ea02 (re-gates main after #461's perf-path content invalidated the standing records). Do NOT merge until the gate records land and PR benchmark gate is green.

PR #465, opened 2026-08-11
feat(laguna-gguf): keep-packed Q4_K_M serving — grouped MMQ + fused decode GEMV (Phase-5 bucket e)

rsafier:
Draft against `stack/batch4-v2` so the diff shows only this bucket; will re-target to `main` when #464 merges.

First Phase-5 extraction from the validated integration tree (#442 ledger, bucket e — deps #334 ✓ via #457, #350 ✓ in the base stack). Two commits:
- `9eaf02c5` fixes the stack tip itself (msg_entry_tests E0061 vs 6-arg build_msg_entries; tool_parser fmt + 500-LoC split)
- `05af1919` the feature (56 files, +1876/−214)

**What it adds:** `WeightDtype::{PackedQ4K,Q6K}` + packed weight types, GGUF keep-packed expert stack (automatic for arch=laguna; other archs keep dequant→BF16), MoE keep-packed forward arm (grouped Q4_K/Q6_K MMQ, fused gate+up, fused n=1 decode GEMV pair), graph-safe `moe_grouped_q8` arena scratch, laguna loader packed views + correction-bias widen, `q4k_mmq`/`moe_q4k_decode_fused` kernels promoted to `kernels/gb10/common/` (27b keeps same-dir includes via the symlink idiom — closure hashes stay precise).

**House rules:** no env vars — `ATLAS_PACKED_DECODE_FUSED` became `GemmDispatch.packed_decode_fused` (default-ON, kill-switch), `ATLAS_GGUF_KEEP_PACKED_MOE` dropped (arch-automatic); q4k probes issue only when `packed_experts.is_some()` → **zero new `[expected_absent]` debt on any target** (the #457 route-1 pattern); stale laguna expected_absent entries removed; ≤500-LoC splits; fails-without tests (packed byte_size, laguna name map, dense/MoE scratch sizing gate, arena alloc count, lever default).

**GPU validation (dgx-00, official poolside Q4_K_M GGUF):** keep-packed engages (log line + audit table shows `q4k_mmq`, `moe_q4k_decode_fused` = used), boot 62s, temp-0 facts 3/3, decode **18.6 tok/s** (historical parity target ~17.5), zero errors. fmt/clippy/tests workspace-clean.

Deliberately excluded (their own buckets per #442): Laguna-LoRA gate/up fold into the packed arm (d2), XS shared-expert arm (b), ATTN_NVFP4 lever.


rsafier:
Re-based onto `main` and **re-targeted** (`cd9b987f`). Base was `stack/batch4-v2`, which belongs to the now-closed #464.

**Why the base disappeared:** #473 merged 295 files / +22,318 lines including the whole Laguna-S-2.1 tree, so #464's content reached `main` that way and the stack PR was closed (#350 likewise — I closed it as absorbed). This PR is now a clean delta over main: **53 files, +1849/−192**, purely the keep-packed Q4_K_M work.

**Conflicts were unusual** — because main and this branch each carry Laguna from *different histories*, most came back as `AA` (both-added) rather than content conflicts. Note for anyone re-running this: `git status --porcelain | grep '^UU'` does **not** list AA files; sweep with `grep -rln '^<<<<<<< '` instead, or you will think you are done when you are not.

Resolutions:
- `cli/validate.rs` → **main's**. Main refactored the allowed-value lists into a shared `flag_values` module read by both the validator and the dashboard picker; our inline consts are superseded (and lacked `deepseek_v4`).
- `moe/forward_prefill.rs` → **ours** (main's side empty): the keep-packed n==1 decode-router arm that uses `dense_gemv` instead of the badly-shaped `dense_gemm_prefill`.
- `laguna.rs`, `load_layers.rs` (6 hunks), `laguna-s-2.1/MODEL.toml` → **ours**. Worth stating why this is not just "prefer the local side": at the one hunk that looked like a genuine divergence (the routed-expert build), ours is a strict **superset** — `expert_proj()` calls `quantized_v2` in exactly main's case and only *adds* a BF16-GGUF requant fallback, plus the correction-bias BF16→F32 widen that fixes an illegal-address read (`moe_topk_sigmoid_batched` takes `const float*`; handing it a BF16 buffer over-reads past the end). Nothing main does is lost.
- The MODEL.toml hunk drops `[expected_absent.q4k_*]` for this target because bucket-e moves those kernels into `kernels/gb10/common/`, so the lookups now resolve. That is validated by the build rather than asserted — see below.

Verified after resolution: `cargo check --workspace --tests` clean, fmt clean, `clippy --workspace --tests` 0 errors, 525 tests passing. Still draft pending the GPU re-validation of the keep-packed serve on this new base.

PR #467, opened 2026-08-11
Nemotron-H multi-sequence batched decode (stacks on #462) — C=8 +40%

rsafier:
Draft; stacks on #462 (Lightning + NoPE) — the first ~15 commits are that PR, review from 72b2211c onward. Re-target/rebase when #462 lands.

Nemotron layers had no decode_multi_seq overrides, so concurrent decode ran the default per-sequence loop (measured: C=1 66 → C=4 only 73 aggregate). This adds milestone-A batching with measured gating:
- mamba2 in/out projections batched across sequences (weight-streamed once instead of n×), engaged only at padded rung ≥ 8 where measurement shows the win (compile-time const with cited numbers); below that, byte-identical per-seq path
- batched MoE body implemented but **disabled** (measured net loss at every rung; kept alive under geometry tests)
- mixed-step scratch layout fix: prefill metadata now parks past the larger of the prefill/decode routing regions — exhaustive fail-without test over padded_n × proc_count × top_k (pre-fix formula fails for every padded_n > proc_count = OOB KV-write / cross-request contamination class)

**Independently serve-verified (GB10, A/B vs parent build):** C=1 67.4 (parity); C=4 72.2–72.7 (parity — the interim regression fully recovered); **C=8 sum-of-stream 103.6–104.8 vs parent 75 (+40%)**; temp-0 6-fact battery 12/12 at C=2 and C=6; mixed-step stress (5 long decodes + 3 injected short prompts) all correct, zero log errors. Adversarial review of the layout fix failed to construct a collision (incl. Puzzle per-block top_k, Super-120B latent MoE, dense zero-size).

Milestone B (strided conv+scan kernels for full mamba2 batching) is the same kernel family the Lightning MTP verify needs (#462 discussion) — one future kernel effort flips both.


rsafier:
Merged `main` in via the fixed Lightning base (#462). No resolution needed on this branch — the duplicate `build_behavior.rs` parser is deleted on the base side and untouched here, so the merge carries the deletion through. See #462 for why the textual merge with main compiles to six E0609s and why CI could not see it.

Verified on the merge result: workspace `check` + `clippy` clean, 4612 tests pass.

rsafier:
End-of-day: verified this branch merges **and builds** against post-#490 `main` (merge result checked with `cargo check --workspace --tests`, not just the head — a clean textual merge is not evidence on this repo). CI green apart from `PR benchmark gate` (records owed).

Stacks on #462; #475 sits above it. Landing #462 first collapses this diff. Note that #490's lossless-SW-GEMV chain wired through Nemotron Mamba2 in/out proj and MoE shared-up, i.e. straight through this PR's territory — the merge is clean and builds, but the batched-decode numbers in the description predate that and are worth a re-measure before gate records are paid for. Details and the four-build A/B are on #475.

rsafier:
Heads-up: this PR's base moved. #462 is now closed in favour of #544, which re-ports Lightning support onto current `main` — the old stack could not be rebased because main replaced the kernel-target resolution API it extends (details in #487's closing note).

**Nothing here is superseded.** The multi-sequence batched decode for Mamba-2 + MoE (`72b2211c`, plus the mixed-step scratch-clobber fix `77089e52` and the measured-threshold gating `eed58f43`) is *not* in #544 and remains the interesting part of this PR. #544 deliberately carries only what Lightning needs to load and serve, because bundling batched decode, the tool-parser fix and Lightning support into one stack is what made the original unrebaseable.

To revive: re-apply those three commits on top of #544 rather than rebasing this branch — the shared ancestors below them (`d033913d`, `a6c075b7`, `cd63fd37`) are already in #544 in re-applied form, so a rebase would replay them into conflicts against their own re-ports.

`ba01f3ec` (Lightning's MTP draft head) is also still unported and noted on #462.

rsafier:
Closing in favour of #545, which carries this work forward on top of #544.

Not superseded — **re-applied**. This PR's base was #462, now closed because main replaced the kernel-target resolution API that stack extends (detail in #487's closing note), which left this branch unrebaseable rather than merely stale. So the three substantive commits were cherry-picked onto #544 instead:

| commit | now |
|---|---|
| `72b2211c` multi-sequence batched decode (Milestone A) | #545 `64479e6e` |
| `77089e52` mixed-step scratch clobber | #545 `3eaf0344` |
| `eed58f43` measured-threshold gating | #545 `9a94d112` |

The claim here reproduces, and is slightly better than advertised. Measured on Lightning 30B-A3B on a GB10, aggregate decode, distinct prompt per stream, same settings on both legs:

| C | without | with | delta |
|---|---|---|---|
| 4 | 83.6 tok/s | 84.1 | parity |
| **8** | **65.1** | **94.5** | **+45%** |

Worth restating what that baseline column means: **without these commits, C=8 is slower than C=4** (65.1 vs 83.6) — the eighth stream currently makes the box go backwards. `eed58f43`'s gating also holds C≤4 at parity, so the C=4 regression noted earlier on this PR does not reproduce.

One thing the port surfaced that is worth carrying into any future revival attempt: all three commits cherry-picked with **zero textual conflicts** and the tree still would not build — main added `moe_lora_route` to `ForwardContext` after these tests were written, and only the *test* profile broke while the lib compiled clean. Fixed in #545 as `e5043c45`.

The branch is untouched, so nothing here is lost.

PR #468, opened 2026-08-11
fix(tui): downloads, toasts, progress UI, network memory, log scroll

rrstesiak:
Fixes the five dashboard bugs reported this week:

1. **Downloads failed on arrival with an error toast** when a stale HF token was present. The Hub answers 401 for public repos on any invalid Authorization header, so the client now retries anonymously on 401/403. Gated repos still classify exactly as before.
2. **Error toasts persisted forever.** They now auto-dismiss at 12s (info stays 5s). Also settled a related wedge: a download worker dying without a terminal message left the job tracked forever, refusing every later download with "already downloading".
3. **Download progress now renders in both panes**: a gradient bar with bytes, rate, and file counter in the detail pane, and a pulsing yellow dot inline after the model name on the list row.
4. **Network tab memory read 0/0** until a model loaded (GPU accessors). It now reports used/total system RAM from /proc/meminfo, dash when unreadable.
5. **Main tab log scrolling jammed ~64 lines up**: the renderer fetched only what one frame held and published a ceiling from that. It now fetches ahead of the offset, so scrollback reaches the whole 10k line ring.

731 tui + 34 model_download + 112 app unit tests pass. Two ignored network tests (stale-token fallback, failure settling) pass live. Verified interactively on GB10: resume, stop/resume round trip, toast expiry, memory line, deep scrollback.

### DIFF
--- a/crates/spark-server/src/model_download/hf.rs
+++ b/crates/spark-server/src/model_download/hf.rs
@@ -116,6 +116,30 @@ fn get(
     }
 }
 
+/// [`get`], retrying anonymously when the token is refused.
+///
+/// A stale or revoked token makes the Hub answer 401 for PUBLIC repos too, so
+/// one bad line in `~/.cache/huggingface/token` turned every download into an
+/// instant "requires credentials" failure. Public models need no credentials
+/// at all; the retry costs one request and only ever runs on a 401/403 that
+/// was about to be fatal anyway. A genuinely gated repo fails the anonymous
+/// retry with the same class of status, and the ORIGINAL error is reported —
+/// `had_token: true` is what routes the reader to the licence hint rather
+/// than the "set HF_TOKEN" one.
+fn get_or_anon(
+    url: &str,
+    token: Option<&str>,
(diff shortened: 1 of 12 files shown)


tbraun96:
Landed on main via **#469** (`63deab7e5`, squash merge).

Per campaign direction, #468 was stacked beneath the SSM State Poisoning Gate PR so both would ride one merged gate cycle and one merge. All 12 files byte-match `main`:

```
download error lockout, stuck toasts, download progress UI, network tab memory,
log scroll, per-message channel drain, Main/Kernels table split (500-LoC cap)
```

Closing in favor of #469, which carried this content through the gate and merged cleanly. Thanks @rrstesiak!

rrstesiak:
Glad the fixes landed, and thanks for carrying them through the gate. The collapse-vs-jitter split is the right angle ...Byte-strict on the restore path would have failed every healthy build, and the early-EOS signature is the thing actually worth zero-tolerance. More where that came from whenever the stack needs it.

PR #470, opened 2026-08-12
fix(windows): make the native-HIP build run out of the box on Strix Halo (gfx1151, runtime-verified)

AzeezIsh:
## Summary

`docs/porting/STRIX_WINDOWS_HIP.md` is upfront that the Windows native-HIP path
is **COMPILE-VERIFIED, RUNTIME UNVERIFIED** ("there is no Windows AMD machine in
the fleet"). This is the runtime follow-up, from an actual gfx1151 Windows box:
a Framework Desktop (Ryzen AI MAX+ 395 / Radeon 8060S), Windows 11, serving
`nvidia/Qwen3.6-27B-NVFP4`.

The port itself was sound — the shim mappings and the 64-bit mask widen are
correct, and the model runs. But four things stopped a first run from working,
and each one failed in a way that pointed somewhere other than the cause. Two of
them cost hours of misdiagnosis, so the fixes are as much about the error you get
as the behaviour.

Serve now comes up and answers in ~93 s from a clean `cargo build`, with no
manual steps.

## What was wrong

**1. `hipMemGetInfo` is broken on Windows HIP** — returns `hipErrorInvalidValue`
standalone, and reports `free == 0` under a live context. Atlas sizes its KV
cache from `cuMemGetInfo`, so serve aborts with `No memory left for KV cache` on
a box with ~64 GB genuinely allocatable (measured with a `hipMalloc` ladder).
`cuMemGetInfo_v2` now tracks what the shim hands out and synthesises a truthful
answer when HIP won't give one. It only engages when `hipMemGetInfo` actually
fails or returns zeros, so **Linux/ROCm behaviour is unchanged**. Helpers are
`static` and don't start with `cu`, so the dumpbin-derived export list still
comes out at 76.

**2. Only `amdhip64` was staged.** `amdhip64_6.dll` loads its code-object manager
by plain name, so on a host whose driver left an older `amd_comgr_2.dll` in
System32, HIP binds the stale comgr and *every* `hipModuleLoadData` fails as
`CUDA_ERROR_OUT_OF_MEMORY`. That error reads as a KV-sizing bug and sends you
tuning `--gpu-memory-utilization` for hours. Measured here: System32
`amd_comgr_2.dll` 109 MB with no version resource, vs ROCm 6.4's 115.55 MB.
Now stages `amd_comgr*`, `hiprtc*`, `hiprtc-builtins*` too — **and copies them
next to the built binary**, since OUT_DIR staging only served the packaging step
(`cargo build` then running the exe still picked up whatever was on the system
path, and a rebuilt `cuda.dll` went unused because an older copy was already
sitting beside the exe).

**3. 219 unmaterialised symlinks.** `kernels/` leans on symlinks
(`strix-hip -> strix -> gb10`) and Git for Windows defaults to
`core.symlinks=false`, checking each one out as a ~32 byte text file containing
the link target path. The build now fails early naming `core.symlinks` and the
fact that the links *chain*, instead of letting the compiler complain about
source it can't parse.

**4. `--no-fast-load` was mandatory but undocumented.** The fast loader needs
`O_DIRECT`/`posix_fadvise`, is ON by default, and hard-errors on a non-Unix host
before a single weight loads. Now warns and falls back to the standard loader.

**5. The request-timeout warning reported the wrong clock** (added after the
first four; diagnostic only, no behaviour change). `timeout_at` is stamped at
ARRIVAL but `request_start` at PREFILL, so a 300 s budget consumed entirely by
queue wait printed as `Request timeout after 8.4s` — which reads as a slow
generation. The part that mattered wasn't in the message: those requests die on
their *first* decode step having produced one token, and go back to the client as
a normal `finish_reason="length"`. The warning now reports `output_tokens`,
`since_prefill_s` and `overdue_s`, and sets `guard_stop = "request_timeout"`.
This cost me a full 5-hour benchmark run that scored 1-token completions as wrong
answers; the behavioural half is #482.

## Making it runnable by someone else

`docs/porting/STRIX_WINDOWS_HIP.md` asked to be "corrected by the first person who
runs it". This does that, and adds `scripts/strix-windows/first_run.ps1` so that
correction is executable rather than a list of things to remember.

**"OK, how do I test it?" now needs no toolchain at all.** CI already publishes a
prebuilt `spark-windows-x86_64-amd-hip` zip, so:

```powershell
$env:ATLAS_BIN = "C:\path\to\unzipped\spark.exe"
powershell -ExecutionPolicy Bypass -File scripts\strix-windows\first_run.ps1
```

checks the box, serves with the verified config, fires a completion and writes the
answer to `first_run_smoke.log`. Driver and weights only — no MSVC, no HIP SDK, no
Rust, no clone. Unset `ATLAS_BIN` and the same script builds from source,
repairing the 219 chained kernel symlinks first. `-Phase check` audits and changes
nothing.

Three values in the doc were wrong in ways that fail somewhere other than the
cause, and are corrected inline with the reasoning:

| | doc said | actually | why it matters |
|---|---|---|---|
| `ATLAS_KV_EXTERNAL_RESERVE_GB` | `6` | **`0`** | double-counts against the fixed shim; oversized the KV pool to 11.3 GB, then died on a later 24 MB alloc |
| `ATLAS_SSM_TAIL_MIDCHUNK` | `1` | **`0`** | corrupts cross-request SSM prefix reuse; empty 1-token completions on 12/12 `live_multiple` |
| `--gpu-memory-utilization` | `0.35` | **`0.80`** | 0.35 needlessly starves KV; ceiling ~0.83 because the fraction is against a reported 76.9 GB the driver won't honour past ~63 GB |

The doc's WDDM/VGM warning is replaced rather than kept: it did **not** reproduce —
~64 GB was allocatable with no BIOS or Adrenalin tuning, so the old text sent
readers hunting through firmware for something that was never the problem.

## Verification

Each change was verified on the box, not just compiled:

| change | how it was verified |
|---|---|
| `cuMemGetInfo_v2` | serve reports `76.7 GB free` → `35.85 GB` after weights; KV sizes correctly and the OOM watchdog no longer trips |
| DLL staging | wiped every DLL from the target dir, rebuilt, confirmed the build places all 7 including `amd_comgr_2.dll` at 115.5 MB (the ROCm one) |
| symlink guard | wrote a synthetic 37-byte stub into `kernels/strix-hip/common/`, confirmed the build stops with the `core.symlinks` message, removed it |
| fast-load fallback | ran serve **without** `--no-fast-load`; warns and loads all 2194 weight tensors instead of aborting |
| timeout warning | ran with `--request-timeout 1` so the deadline expires during prefill: warning fires with `output_tokens=1 since_prefill_s=3.84 overdue_s=2.84` and the response returns `finish_reason=length` with a 1-token empty body — the production failure in miniature |

Full clean build + serve + a 196-entry BFCL v3 single-turn run on this hardware:
**165/196 = 84.2%** weighted (`simple` 0.967, `multiple` 0.967, `live_parallel`
0.625), 82 min, zero request timeouts. Not comparable to the porting doc's 89.22
— that's v4 on a different 167-sample subset and `bfcl-eval` ships v3 data.

## Notes for maintainers

- The doc's WDDM/VGM memory warning did **not** reproduce here — 64 GB was
  allocatable with no BIOS tuning. Worth revisiting that section.
- Once (1) is in, `ATLAS_KV_EXTERNAL_RESERVE_GB` should be **0** on Windows, not
  the doc's `6`: with honest accounting that value is a co-tenant *discount* that
  double-counts and oversizes the KV pool until a later allocation fails.
- `--gpu-memory-utilization` is a fraction of a total the driver will not honour
  (reports 76.9 GB; ~63 GB actually allocatable). `0.80` works here. Deriving the
  budget from measured-allocatable would be a real improvement but is a
  behavioural change, so it's deliberately **not** in this PR.

Three further issues found on this box are filed separately, since they are
runtime behaviour rather than build/first-run: #471 (`--disable-thinking`
silently ignored on `/v1/completions`), #472 (`ATLAS_SSM_TAIL_MIDCHUNK=1`
corrupts shared-prefix BFCL runs — the regression
`scripts/mlperf-edge/ab_midchunk.sh` already documents), and #482 (a request
whose deadline expires while queued is prefilled anyway and returned as a
1-token `finish_reason="length"`, which no eval harness can distinguish from a
legitimate `max_tokens` stop).

Happy to run anything else on this hardware — it's a permanent Windows/gfx1151
target now, which is exactly what the porting doc says the fleet lacks.

### DIFF
--- a/crates/atlas-kernels/build.rs
+++ b/crates/atlas-kernels/build.rs
@@ -670,6 +670,33 @@ fn build_hip_shim(manifest_dir: &std::path::Path, out_dir: &std::path::Path) {
     );
 }
 
+#[cfg(windows)]
+fn resolve_dumpbin_path() -> std::path::PathBuf {
+    if let Ok(path) = std::env::var("ATLAS_DUMPBIN") {
+        let candidate = std::path::PathBuf::from(path);
+        if candidate.exists() {
+            return candidate;
+        }
+    }
+
+    if let Ok(vctools_dir) = std::env::var("VCToolsInstallDir") {
+        let base = std::path::Path::new(&vctools_dir).join("bin");
+        for (host, target) in [
+            ("Hostx64", "x64"),
+            ("Hostx86", "x86"),
+            ("Hostx64", "x86"),
+            ("Hostx86", "x64"),
+        ] {
+            let candidate = base.join(host).join(target).join("dumpbin.exe");
+            if candidate.exists() {
+                return candidate;
+            }
+        }
+    }
+
+    std::path::PathBuf::from("dumpbin.exe")
+}
+
 /// Windows native-HIP runtime shim. Builds ONE `cuda.dll` from the real cu*
 /// (libcuda_hip_shim.cpp) and cudart (libcudart_hip_shim.cpp) HIP mappings plus
 /// the cuBLASLt stub, linked against `amdhip64.lib`, and generates its import
@@ -718,13 +745,20 @@ fn build_hip_shim_windows(manifest_dir: &std::path::Path, out_dir: &std::path::P
     // `SECTn ... External | <name>`; undefined imports (the hip* the shim calls)
     // are `UNDEF` and start with `hip`, so filtering on `External`, not `UNDEF`,
     // and a `cu` name prefix keeps exactly cu*/cudart/cublasLt.
+    let dumpbin = resolve_dumpbin_path();
     let mut exports = Vec::new();
     for obj in &objs {
-        let out = std::process::Command::new("dumpbin")
+        let out = std::process::Command::new(&dumpbin)
             .arg("/SYMBOLS")
             .arg(obj)
             .output()
-            .unwrap_or_else(|e| panic!("dumpbin /SYMBOLS failed on {} ({e})", obj.display()));
+            .unwrap_or_else(|e| {
+                panic!(
+                    "dumpbin /SYMBOLS failed on {} using {} ({e})",
+                    obj.display(),
+                    dumpbin.display()
+                )
+            });
         for line in String::from_utf8_lossy(&out.stdout).lines() {
             if line.contains("External")
                 && !line.contains("UNDEF")
--- a/crates/atlas-kernels/build.rs
+++ b/crates/atlas-kernels/build.rs
@@ -279,6 +279,7 @@ fn main() {
             target.model,
             target.quant,
         );
+        assert_sources_are_not_symlink_stubs(&cu_files);
 
         // Gather work for this target (no compilation yet — that runs
         // in parallel after the full work plan is built).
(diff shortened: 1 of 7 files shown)


tbraun96:
**Looked at for a `windows-build` stack; it needs your rebase before it can join one.**

Its two siblings resolved without merging: #505 is already on `main` (empty cherry-pick), and #507 is moot (the `resolve_dumpbin_path` it un-gates no longer exists — the shim calls dumpbin directly at `build.rs:902`). So this is the only live one of the three.

**Why I did not fold it in:** cherry-picking onto current `main` conflicts on `crates/spark-model/Cargo.toml`. At +876/-54 across seven files and a base from 2026-08-12, resolving that is a judgement call about your intent, not a mechanical rebase — I would rather not guess at it inside someone else's PR.

**If you rebase it**, it touches `crates/` and `kernels/`, so it re-opens all 11 gates and owes a campaign. Worth pairing with another kernels change so one campaign covers both — #474 and #484 are in the same `crates,kernels` bucket and are candidates. Say the word and I will assemble that stack around whatever you land.

tbraun96:
**Base-branch check (2026-09-19): this PR's base is NOT orphaned — leaving it open, not retargeting.**

I looked at this while sweeping open PRs whose base branch belongs to a closed PR. `strix/consolidate-tmp` looked like one of those because **#354** (head `strix/consolidate-tmp`, closed 2026-07-22) is the PR that turns up first — but #354 was a 3-minute duplicate opened by tbraun96 and closed by AzeezIsh with no comment; the live owner of that branch is **#353** (same head, base `main`, OPEN, 54 commits, head `ea36e0c9a` — identical to this PR's base sha). So this PR is a correctly-stacked child of #353, and it is recorded here so nobody "fixes" it by retargeting.

**Why retargeting to `main` is not viable** (checked with `git apply --check` against a `main` index, no build):
- The diff would become `main...44f0edf64` = **82 files, +5046/-303** — all 54 of #353's commits ride along, and 4 of the 6 files this PR itself touches fail to apply (`serve_phases/weights.rs:37`, `decode_logits_step.rs:718`, and two "does not exist in index").
- The rebrand **#1101** (`569ffd092`, 2026-09-17) renamed `crates/atlas-kernels/` → `crates/avarok-kernels/` and `hip/compat/atlas_hip_win_shims.h` → `avarok_hip_win_shims.h`, so this PR's `build.rs` / `libcuda_hip_shim.cpp` / shim-header hunks now target paths that no longer exist on `main`. A rebase has to re-home them.
- `docs/porting/STRIX_WINDOWS_HIP.md` and `scripts/strix-windows/first_run.ps1` have no counterpart on `main` at all.

**The content is live**, not already landed: of this PR's own added lines (base..head, ≥30 chars), **565 of 573 are absent from `main`**, and only 57 of those are rebrand-touching identifiers — the rest is the build.rs toolchain detection, the HIP shim changes, the `first_run.ps1` script and the scheduler clock fix.

**What has to happen before it can land anywhere:** either #353 resolves its `CONFLICTING` state against `main` (it has been DIRTY since mid-August, so this PR inherits that block), or — as the 2026-09-04 note above already asked — this PR is re-cut on `main` directly, now also mapping the `atlas-kernels` → `avarok-kernels` rename. Either route re-opens all 11 gates and owes a campaign, since it touches `crates/` and `kernels/`.

Not a judgement of the work — only of where it can land from. No changes made to the PR.

tbraun96:
Triaged for a certification campaign and **parked pending a rebase**.

Merged onto current `main` (`5f639bfc57`) locally: **34 conflicting files**, led by `crates/spark-model/src/{layer.rs, layer/tests.rs, layers/deepseek_v4_mtp.rs, layers/dense_ffn.rs, layers/mtp_head/forward.rs, layers/ops.rs, model/impl_a1.rs, model/impl_b1.rs}`. The branch is **271 commits behind** with **55 own commits over 82 files**, which puts it well outside what an owner re-cut can do safely — replaying that many commits risks silently dropping adaptations only you can tell apart from losses.

One thing worth doing before you estimate the work: **raise git's rename limit.** #1101 renamed five crate directories (`crates/atlas-* → crates/avarok-*`, and with them `atlas_* → avarok_*`, `ATLAS_* → AVAROK_*`, `cfg(atlas_*) → cfg(avarok_*)`). Across a change that size git's default gives up on rename detection and reports a misleadingly short conflict list — on this branch it said **3** files; `git -c merge.renameLimit=20000 merge avarok/main` shows the real **34**.

Context on the other stale PRs and what each needs: #1206. Getting native-HIP on Strix Halo running out of the box is worth having — push a rebase and it joins the next composed campaign.

PR #474, opened 2026-08-12
fix(kernels): two silent bit-parity/coverage defects — MoE grouped-GEMM N-tile half-write + w8a16_gemv_batch4 accumulation order

rsafier:
Two independent kernel-correctness defects in `kernels/gb10/common/`. Both are silent — no crash, no NaN, no assertion — and both were hidden by a gate that could not see them.

Both were found while building Nemotron milestone B (branch `feat/nemotron-milestone-b`). Neither is Nemotron-specific and neither depends on that branch's dispatch or perf work, so they are split out here deliberately to land independently. **No Rust dispatch change, no new kernel entry, no new handle, no MODEL.toml change, no threshold change.**

---

### Mechanism

Every caller of the CUDA entries `moe_w4a16_grouped_gemm_ptrtable` and `..._ptrtable_t` goes through `ops::moe_w4a16_grouped_gemm_ptrtable` or `ops::moe_w4a16_grouped_gemm_ptrtable_n128`, and **both** launch with `grid.x = ceil(N_out / 128)`.

In `kernels/gb10/common/moe_w4a16_grouped_gemm.cu` both entries derived their column origin from `N_TILE = 64`:

```c
const unsigned int cta_n = blockIdx.x * N_TILE;   // 64, on a 128-strided grid
```

So CTA `i` owned columns `[64i, 64i+64)` of a stride the launcher had spaced 128 apart. Columns `[64·ceil(N/128), N)` — the top half — were never written by any CTA and stayed at whatever the caller left there, which is the caller's `memset` zero. No out-of-bounds, no fault; just a silently truncated output tensor.

The divisor lives in Rust and the tile width lives in CUDA, so nothing in the build or the test suite could observe them diverge. The Rust doc comment on the wrapper said `ceil(n_out/64)` — describing a launch that no longer happened — which is what made the mismatch read as intentional.

### Blast radius (precise)

Only targets that resolve `moe_w4a16_grouped_gemm.cu` from `kernels/gb10/common/` are affected. Every other gb10 MoE target ships a per-model shadow of this file that already uses a 128-wide tile for these two entries (`N_TILE_LG` on the qwen/holo/gemma/mistral/step/minimax/deepseek family, `N_TILE_PT` on Puzzle-75B's `_ptrtable`), and `kernels/strix/common/` already uses `N_TILE_PT 128` too.

That leaves exactly two affected targets:

| target | arm reached | effect |
|---|---|---|
| `nemotron-3-nano-30b-a3b` | `_ptrtable_t` (`moe_latent_size == 0` builds `up_ptrs_t`/`down_ptrs_t`) | grouped UP `N_out=1856` wrote 15×64 = 960 of 1856 columns; grouped DOWN `N_out=2688` wrote 21×64 = 1344 of 2688 — exactly 50% |
| `nemotron-super-120b-a12b` | non-`_t` `_ptrtable` arm | same defect, same file |

On Nemotron-H Lightning-30B this halved the routed-expert contribution in sorted grouped prefill, structurally (the same columns every time), which is why it read as model quality rather than as a bug.

**Not overclaiming:** no other model was affected, because every other model's shadow was already 128-wide. Any previously recorded MoE A/B number *for these two targets* was measured against half-zero output and is invalid.

### The fix

Adds `N_TILE_PT 128` and switches **only** those two entries to it: `cta_n`, `smem_B`, `b_stride`, the B-load decomposition, and a per-kernel `constexpr int N_SUB = tile / 8` driving `acc[N_SUB][4]`, the mma loop and the store loop. Cost: `acc[16][4]` = 64 accumulator registers, smem 2304 + 4160 = 6464 B.

The legacy `moe_w4a16_grouped_gemm` entry deliberately **stays** at `N_TILE 64` — its only caller (`atlas-spark-bench/benches/moe.rs`) launches the matching 64-wide grid, and keeping it unmodified makes it the reference leg for the oracle below.

Widening an N tile changes only *which CTA owns which column*. It never changes the order in which a single output element accumulates over `k_base` (same k loop, same per-element `mma.m16n8k16` sequence), so the retuned kernel at `ceil(N/128)` must be **byte-identical** to the 64-wide kernel on its own correct `ceil(N/64)` grid. That is the gate, not a tolerance.

### Byte-level evidence — GB10 (dgx-00), fixed code

```
kernel target: nemotron-3-nano-30b-a3b/nvfp4 (166 modules)
UP   N=1856  K=2688  decode rows=48   m_tiles=1 grid.x 64w=29  128w=15  | ptrtable==true ptrtable+gather==true ptrtable_t==true | witness differs=true witness zero-cols=896/1856 (cols >= 960 all zero=true)
     ms/launch: golden(64w, grid.x=29) 3.197  ptrtable(128w, grid.x=15) 4.606  ptrtable_t(128w) 2.129  [pre-fix half-work launch 1.610]
UP   N=1856  K=2688  mixed  rows=464  m_tiles=2 grid.x 64w=29  128w=15  | ptrtable==true ptrtable+gather==true ptrtable_t==true | witness differs=true witness zero-cols=896/1856 (cols >= 960 all zero=true)
     ms/launch: golden(64w, grid.x=29) 8.936  ptrtable(128w, grid.x=15) 13.109  ptrtable_t(128w) 5.965  [pre-fix half-work launch 4.722]
DOWN N=2688  K=1856  decode rows=48   m_tiles=1 grid.x 64w=42  128w=21  | ptrtable==true ptrtable+gather==true ptrtable_t==true | witness differs=true witness zero-cols=1344/2688 (cols >= 1344 all zero=true)
     ms/launch: golden(64w, grid.x=42) 3.224  ptrtable(128w, grid.x=21) 4.564  ptrtable_t(128w) 2.226  [pre-fix half-work launch 1.651]
DOWN N=2688  K=1856  mixed  rows=464  m_tiles=2 grid.x 64w=42  128w=21  | ptrtable==true ptrtable+gather==true ptrtable_t==true | witness differs=true witness zero-cols=1344/2688 (cols >= 1344 all zero=true)
     ms/launch: golden(64w, grid.x=42) 8.716  ptrtable(128w, grid.x=21) 12.562  ptrtable_t(128w) 5.699  [pre-fix half-work launch 4.364]

PASS — the 128-wide ptrtable entries are byte-identical to the 64-wide kernel on its own correct grid, and the pre-fix launch is provably half-zero.
```

Exit 0. Note the `witness` leg: that is the *pre-fix* geometry (64-wide tile on the production 128 grid) run as a built-in negative control. It differs from golden and leaves 896/1856 and 1344/2688 columns all-zero — the predicted half-write, reproduced in the same run that proves the fix. Without that control a passing parity leg would prove nothing.

Cost of correctness, on the `_ptrtable_t` arm Lightning actually dispatches: 1.20 → 2.13 ms (decode shape) and 3.43 → 5.97 ms (mixed), i.e. ~1.75× for doing 2× the columns.

### Fails without the fix

Reverting **only** `kernels/gb10/common/moe_w4a16_grouped_gemm.cu` to `main` and rerunning the same oracle:

```
kernel target: nemotron-3-nano-30b-a3b/nvfp4 (166 modules)
UP   N=1856  K=2688  decode rows=48   m_tiles=1 grid.x 64w=29  128w=15  | ptrtable==false ptrtable+gather==false ptrtable_t==false | witness differs=true witness zero-cols=896/1856 (cols >= 960 all zero=true)
  FAIL: retuned N_TILE_PT output is not byte-identical to the 64-wide golden
UP   N=1856  K=2688  mixed  rows=464  m_tiles=2 grid.x 64w=29  128w=15  | ptrtable==false ptrtable+gather==false ptrtable_t==false | witness differs=true witness zero-cols=896/1856 (cols >= 960 all zero=true)
  FAIL: retuned N_TILE_PT output is not byte-identical to the 64-wide golden
DOWN N=2688  K=1856  decode rows=48   m_tiles=1 grid.x 64w=42  128w=21  | ptrtable==false ptrtable+gather==false ptrtable_t==false | witness differs=true witness zero-cols=1344/2688 (cols >= 1344 all zero=true)
  FAIL: retuned N_TILE_PT output is not byte-identical to the 64-wide golden
DOWN N=2688  K=1856  mixed  rows=464  m_tiles=2 grid.x 64w=42  128w=21  | ptrtable==false ptrtable+gather==false ptrtable_t==false | witness differs=true witness zero-cols=1344/2688 (cols >= 1344 all zero=true)
  FAIL: retuned N_TILE_PT output is not byte-identical to the 64-wide golden

FAIL
```

Exit 1, all four cases. Also observe the ms/launch on the reverted build: `ptrtable` drops to 1.31 / 3.84 ms — the "fast" pre-fix number was fast because it was doing half the work.

The CPU-only guard fails on the reverted kernel too, which is the point of adding it:

```
test common_ptrtable_entries_use_the_matching_cuda_tile_width ... FAILED
panicked at crates/spark-model/tests/moe_grouped_ntile_geometry.rs:86:10:
kernels/gb10/common/moe_w4a16_grouped_gemm.cu must define N_TILE_PT
```

---

### Mechanism

`kernels/gb10/common/w8a16_gemv_batch4.cu` states in its header that its output is bit-identical to running `w8a16_gemv` M times. That claim is the entire reason a decode ladder is allowed to substitute it: the batched kernel exists only to make one DRAM pass over the weight bytes instead of M, and it is expected to move nothing numerically.

The inner loop did not honour it:

```c
acc[t] += __bfloat162float(lo) * wf[j * 2]
        + __bfloat162float(hi) * wf[j * 2 + 1];      // acc + (x + y)
```

against `w8a16_gemv`'s

```c
acc += x;                                            // (acc + x) + y
acc += y;
```

`kernels/gb10/common/KERNEL.toml` compiles with `--fmad=false`, so the compiler is explicitly forbidden from contracting these, and `acc + (x + y)` and `(acc + x) + y` are genuinely different FP32 results.

The old gate was a `cos >= 0.9999` microtest at a 512-column shape. A handful of flipped BF16 elements out of tens of thousands is invisible to a cosine, so the gate passed for as long as the defect existed.

`w8a16_gemv_batch16` instantiates the same templated body (`w8a16_gemv_batchm_impl<16>`) and carried the same defect.

### Blast radius

This is a `kernels/gb10/common/` file with no shadows, so **every gb10 target ships the affected kernel**. The exposure is not "which model has the bad kernel" but "which code paths substitute it for M × `w8a16_gemv` on the strength of the parity claim" — every batched-decode ladder in the tree that does so was relying on a false statement. Concretely: results depend on how many sequences happen to be batched together in a step, which is exactly the property such a ladder is supposed not to have.

What this is **not**: it is not a large numerical error and it is not a crash. It is 1–9 differing BF16 elements per launch out of `M × N`, with occasional large relative jumps where an accumulator lands near a BF16 rounding boundary. The reason it matters is the broken invariant, not the magnitude.

### The fix

Split the fused add back into two accumulations, matching the M=1 kernel exactly. One-line change plus the comment explaining why it must not be re-fused.

### Byte-level evidence — GB10 (dgx-00), fixed code

Byte compare (not cosine) of both batched tiers against M × `w8a16_gemv`, at production-scale projection shapes `[10304 × 2688]` and `[2688 × 4096]`, three seeds, every M each tier serves — 36 legs:

```
seed     1  in_proj  [10304 x 2688]  batch4  M=2   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed     1  in_proj  [10304 x 2688]  batch4  M=3   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed     1  in_proj  [10304 x 2688]  batch4  M=4   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed     1  in_proj  [10304 x 2688]  batch16 M=8   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed     1  in_proj  [10304 x 2688]  batch16 M=12  byte-identical=true  diff_elems=0       max|delta|=0.000000
seed     1  in_proj  [10304 x 2688]  batch16 M=16  byte-identical=true  diff_elems=0       max|delta|=0.000000
seed     1  out_proj [ 2688 x 4096]  batch4  M=2   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed     1  out_proj [ 2688 x 4096]  batch4  M=3   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed     1  out_proj [ 2688 x 4096]  batch4  M=4   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed     1  out_proj [ 2688 x 4096]  batch16 M=8   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed     1  out_proj [ 2688 x 4096]  batch16 M=12  byte-identical=true  diff_elems=0       max|delta|=0.000000
seed     1  out_proj [ 2688 x 4096]  batch16 M=16  byte-identical=true  diff_elems=0       max|delta|=0.000000
seed    99  in_proj  [10304 x 2688]  batch4  M=2   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed    99  in_proj  [10304 x 2688]  batch4  M=3   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed    99  in_proj  [10304 x 2688]  batch4  M=4   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed    99  in_proj  [10304 x 2688]  batch16 M=8   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed    99  in_proj  [10304 x 2688]  batch16 M=12  byte-identical=true  diff_elems=0       max|delta|=0.000000
seed    99  in_proj  [10304 x 2688]  batch16 M=16  byte-identical=true  diff_elems=0       max|delta|=0.000000
seed    99  out_proj [ 2688 x 4096]  batch4  M=2   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed    99  out_proj [ 2688 x 4096]  batch4  M=3   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed    99  out_proj [ 2688 x 4096]  batch4  M=4   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed    99  out_proj [ 2688 x 4096]  batch16 M=8   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed    99  out_proj [ 2688 x 4096]  batch16 M=12  byte-identical=true  diff_elems=0       max|delta|=0.000000
seed    99  out_proj [ 2688 x 4096]  batch16 M=16  byte-identical=true  diff_elems=0       max|delta|=0.000000
seed 12345  in_proj  [10304 x 2688]  batch4  M=2   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed 12345  in_proj  [10304 x 2688]  batch4  M=3   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed 12345  in_proj  [10304 x 2688]  batch4  M=4   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed 12345  in_proj  [10304 x 2688]  batch16 M=8   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed 12345  in_proj  [10304 x 2688]  batch16 M=12  byte-identical=true  diff_elems=0       max|delta|=0.000000
seed 12345  in_proj  [10304 x 2688]  batch16 M=16  byte-identical=true  diff_elems=0       max|delta|=0.000000
seed 12345  out_proj [ 2688 x 4096]  batch4  M=2   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed 12345  out_proj [ 2688 x 4096]  batch4  M=3   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed 12345  out_proj [ 2688 x 4096]  batch4  M=4   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed 12345  out_proj [ 2688 x 4096]  batch16 M=8   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed 12345  out_proj [ 2688 x 4096]  batch16 M=12  byte-identical=true  diff_elems=0       max|delta|=0.000000
seed 12345  out_proj [ 2688 x 4096]  batch16 M=16  byte-identical=true  diff_elems=0       max|delta|=0.000000
PASS — the w8a16 batched GEMV tiers are byte-identical to M x w8a16_gemv at every Lightning projection shape and every M they serve.
```

Exit 0. 36/36 legs byte-clean, `diff_elems=0` throughout.

### Fails without the fix

Restoring only the fused `acc += x + y` in `w8a16_gemv_batch4.cu` and rerunning the same binary — **35 of 36 legs differ**, exit 1:

```
seed     1  in_proj  [10304 x 2688]  batch4  M=2   byte-identical=false diff_elems=1       max|delta|=0.000122
seed     1  in_proj  [10304 x 2688]  batch4  M=3   byte-identical=false diff_elems=3       max|delta|=0.500000
seed     1  in_proj  [10304 x 2688]  batch4  M=4   byte-identical=false diff_elems=4       max|delta|=0.500000
seed     1  in_proj  [10304 x 2688]  batch16 M=8   byte-identical=false diff_elems=15      max|delta|=4.000000
seed     1  in_proj  [10304 x 2688]  batch16 M=12  byte-identical=false diff_elems=22      max|delta|=4.000000
seed     1  in_proj  [10304 x 2688]  batch16 M=16  byte-identical=false diff_elems=34      max|delta|=4.000000
seed     1  out_proj [ 2688 x 4096]  batch4  M=2   byte-identical=false diff_elems=4       max|delta|=1.000000
seed     1  out_proj [ 2688 x 4096]  batch4  M=3   byte-identical=false diff_elems=4       max|delta|=1.000000
seed     1  out_proj [ 2688 x 4096]  batch4  M=4   byte-identical=false diff_elems=4       max|delta|=1.000000
seed     1  out_proj [ 2688 x 4096]  batch16 M=8   byte-identical=false diff_elems=6       max|delta|=1.000000
seed     1  out_proj [ 2688 x 4096]  batch16 M=12  byte-identical=false diff_elems=6       max|delta|=1.000000
seed     1  out_proj [ 2688 x 4096]  batch16 M=16  byte-identical=false diff_elems=9       max|delta|=1.000000
seed    99  in_proj  [10304 x 2688]  batch4  M=2   byte-identical=false diff_elems=5       max|delta|=0.001953
seed    99  in_proj  [10304 x 2688]  batch4  M=3   byte-identical=false diff_elems=8       max|delta|=1.000000
seed    99  in_proj  [10304 x 2688]  batch4  M=4   byte-identical=false diff_elems=9       max|delta|=1.000000
seed    99  in_proj  [10304 x 2688]  batch16 M=8   byte-identical=false diff_elems=16      max|delta|=1.000000
seed    99  in_proj  [10304 x 2688]  batch16 M=12  byte-identical=false diff_elems=25      max|delta|=1.000000
seed    99  in_proj  [10304 x 2688]  batch16 M=16  byte-identical=false diff_elems=29      max|delta|=1.000000
seed    99  out_proj [ 2688 x 4096]  batch4  M=2   byte-identical=false diff_elems=5       max|delta|=1.000000
seed    99  out_proj [ 2688 x 4096]  batch4  M=3   byte-identical=false diff_elems=5       max|delta|=1.000000
seed    99  out_proj [ 2688 x 4096]  batch4  M=4   byte-identical=false diff_elems=6       max|delta|=2.000000
seed    99  out_proj [ 2688 x 4096]  batch16 M=8   byte-identical=false diff_elems=7       max|delta|=2.000000
seed    99  out_proj [ 2688 x 4096]  batch16 M=12  byte-identical=false diff_elems=9       max|delta|=2.000000
seed    99  out_proj [ 2688 x 4096]  batch16 M=16  byte-identical=false diff_elems=11      max|delta|=2.000000
seed 12345  in_proj  [10304 x 2688]  batch4  M=2   byte-identical=false diff_elems=5       max|delta|=1.000000
seed 12345  in_proj  [10304 x 2688]  batch4  M=3   byte-identical=false diff_elems=6       max|delta|=1.000000
seed 12345  in_proj  [10304 x 2688]  batch4  M=4   byte-identical=false diff_elems=9       max|delta|=1.000000
seed 12345  in_proj  [10304 x 2688]  batch16 M=8   byte-identical=false diff_elems=21      max|delta|=2.000000
seed 12345  in_proj  [10304 x 2688]  batch16 M=12  byte-identical=false diff_elems=33      max|delta|=4.000000
seed 12345  in_proj  [10304 x 2688]  batch16 M=16  byte-identical=false diff_elems=39      max|delta|=4.000000
seed 12345  out_proj [ 2688 x 4096]  batch4  M=2   byte-identical=true  diff_elems=0       max|delta|=0.000000
seed 12345  out_proj [ 2688 x 4096]  batch4  M=3   byte-identical=false diff_elems=1       max|delta|=0.000244
seed 12345  out_proj [ 2688 x 4096]  batch4  M=4   byte-identical=false diff_elems=1       max|delta|=0.000244
seed 12345  out_proj [ 2688 x 4096]  batch16 M=8   byte-identical=false diff_elems=4       max|delta|=0.003906
seed 12345  out_proj [ 2688 x 4096]  batch16 M=12  byte-identical=false diff_elems=4       max|delta|=0.003906
seed 12345  out_proj [ 2688 x 4096]  batch16 M=16  byte-identical=false diff_elems=7       max|delta|=1.000000
FAIL — a batched tier is NOT byte-identical to M x w8a16_gemv; no decode ladder may substitute it for the M=1 kernel while this holds.
```

The one passing leg (`seed 12345 / out_proj / M=2`) is exactly why a cosine gate could not catch this: at small M the reassociation sometimes happens not to cross a rounding boundary. Note also that the deltas grow with M — the defect gets worse the more the ladder batches.

---

## Gates added

| gate | kind | proves |
|---|---|---|
| `crates/spark-model/examples/moe_grouped_ntile_microtest/` | GPU oracle, exit 0/1/2 | byte-parity of the retuned entries vs the unmodified 64-wide kernel on its own grid, **plus a built-in negative control** reproducing the half-write |
| `crates/spark-model/tests/moe_grouped_ntile_geometry.rs` | CPU-only, runs in normal CI | binds the Rust `grid.x` divisor to the `.cu`'s `N_TILE_PT` and asserts both entries derive `cta_n` from it. This cross-language divergence is what hid the defect; nothing in CI could see it before |
| `crates/spark-model/examples/w8a16_batch_bitparity_microtest.rs` | GPU oracle, exit 0/1/2 | byte compare (not cosine) of both batched tiers vs M × `w8a16_gemv`, three seeds, six M values, two shapes |

Also corrects the `ceil(n_out/64)` doc comment on `ops::moe_w4a16_grouped_gemm_ptrtable` — **doc only, no behaviour change** — which is the stale line that made bug 1 read as intentional.

## Verification (GB10, dgx-00)

Build env: `ATLAS_TARGET_HW=gb10 ATLAS_TARGET_MODEL='*' ATLAS_TARGET_QUANT='*'` (all 22 gb10 targets).

- `cargo clippy --workspace --tests` — 0 errors, 0 warnings
- `cargo test --workspace` — all pass, including the two new geometry tests
- both GPU oracles — exit 0, outputs above
- both fails-without-fix demonstrations — exit 1, outputs above

## Known sibling, deliberately not fixed here

The Puzzle-75B shadow (`kernels/gb10/nemotron-labs-3-puzzle-75b-a9b/nvfp4/moe_w4a16_grouped_gemm.cu`) has the identical mismatch in `_ptrtable_t`: it takes `cta_n` from its 64-wide `N_TILE` while its `_ptrtable` sibling already uses `N_TILE_PT 128`. Puzzle is latent MoE, so `up_ptrs_t` is `None` and that entry is not dispatched today, which is why this PR does not touch it — fixing another target's shadow belongs in its own change with its own gate. A comment in the common file records it so it is not lost.


rsafier:
End-to-end user-visible evidence for the N-tile half-write, from an independent verification pass on a branch that carries this fix:

At **C=1, temperature 0**, Lightning-30B (which resolves the `nemotron-3-nano-30b-a3b` kernel target — one of the two non-shadowed targets using this file):

| prompt | without this fix (main today) | with this fix |
|---|---|---|
| "In what year was Pride and Prejudice first published?" | **1995** | 1813 |
| "What is the capital of Australia?" | **Sydney** | Canberra |

4/6 temp-0 facts correct without the fix, 6/6 with it — and the wrong answers are **stable across repetitions**, i.e. the model is quietly degraded rather than nondeterministic, which is why this went unnoticed.

Worth adding because it reframes something in our own history: a Nemotron determinism note previously read "1813 → 1995 flips" at concurrency as *batching-induced divergence*. 1995 is simply the half-MoE answer. Any Nemotron quality or determinism measurement taken on these two targets before this fix should be treated as suspect.

tbraun96:
Pulling this out of Batch 5 (#490) rather than merging it — not because it is wrong, but because it and `main` now fix the same bug in opposite directions, and the combination is worse than either.

- **main**: kernel keeps `N_TILE 64`; the wrapper launches `ceil(N/64)` via `ptrtable_legacy_grid_x`. Self-consistent.
- **this PR**: kernel adds `N_TILE_PT 128`; `moe_grouped_ntile_geometry` asserts `ceil(N/128)`. Self-consistent.
- **merged**: this PR's 128-wide kernel with main's 64-wide grid → roughly 2× the CTAs, addressing columns past `N_out`. The test catches it: `wrapper #0 launched grid.x=29 for n_out=1856 ... must be 15`.

This PR never touches the Rust wrapper, because when it was written the wrapper already divided by 128. Main has since fixed the same halving — the one your kernel comment documents on Nemotron-H Lightning-30B (grouped UP writing 960 of 1856 columns) — by narrowing the grid instead of widening the tile.

**Verified it passes in isolation** (checkout of this head: `2 passed; 0 failed`), so this is purely an integration conflict, not a defect in your change.

Could you rebase onto current main and make the case for the widening on its merits? It is plausibly the better of the two fixes — the comment argues byte-identical output with half the CTA count, which is a real throughput argument main's narrowing does not offer — but with main already correct, this becomes a performance change to an already-fixed path rather than a correctness fix, and it needs its own gate cycle and its own argument rather than riding along in a batch.

Worth stating explicitly in the rebased PR: whether the widened tile is measurably faster on the two affected targets (`nemotron-3-nano-30b-a3b`, `nemotron-super-120b-a12b`), and what happens to `ptrtable_legacy_grid_x` and its 64-wide const — leaving both mechanisms in the tree is how this class of mismatch recurs.

The two bit-parity microtests are valuable independent of which fix wins.

rsafier:
Rebased onto main and **conceded the N-tile half** (`19ec15fe`). This PR is now just the `w8a16_gemv_batch4` accumulation fix + its byte-parity gate — **3 files, +246/−4**.

**You were right about the conflict, and I reproduced it.** Merging main into this branch succeeds with *zero* conflicts — the two repairs touch different lines — and the result is exactly the broken pairing you described: kernel `N_TILE_PT 128` with wrapper `PTRTABLE_LEGACY_N_TILE = 64`. Confirmed in the merged tree before reverting.

**Our own measurements favour main's repair, so this isn't just deference.** From this PR's oracle run at `N_out=1856`: the 64-wide kernel on its correct 29-CTA grid measured **3.197 ms/launch** against **4.606 ms** for the 128-wide on 15 CTAs. The "byte-identical output with half the CTAs" claim was true, and it was still the worse trade — ~44% slower. Narrowing the grid is the better fix.

Dropped along with the widening:
- the GPU byte-parity oracle — it proved the widening against a 64-wide golden, and there is no widening left to prove;
- the geometry test — it pinned *our* pairing. Worth noting why it can't simply be retargeted: main deliberately runs the two wrappers at **different** divisors (`ptrtable` → `/64` for the common 64-wide kernel, `ptrtable_n128` → `/128` for the 128-wide shadows like `N_TILE_LG` / `N_TILE_PT`), so the tile-vs-divisor invariant is per-target and a single static assertion misstates it. I tried retargeting it to `/64` and it correctly failed against `ptrtable_n128`.

**What remains is independent of all of the above and still absent from main:** `w8a16_gemv_batch4` advertised byte-identity with M× `w8a16_gemv` while fusing `acc += lo*w0 + hi*w1` against the M=1 kernel's two separate accumulations; under `--fmad=false` those re-associate into 1–9 differing BF16 elements per launch, up to 4.0 apart. Every batched-decode ladder that prefers this kernel to save weight traffic was relying on the false claim. Gate re-run on GB10 after the rebase: **byte-identical at every Lightning projection shape and every M both tiers serve**; workspace tests 4596 passed / 0 failing.

Two things I noticed while rebasing, offered as observations rather than claims — you'll know if they matter:
1. Puzzle-75B's shadow still carries the identical 64-vs-128 mismatch in its `_ptrtable_t` entry (`cta_n = blockIdx.x * N_TILE` while its `_ptrtable` sibling uses `N_TILE_PT`). Believed unreachable because latent-MoE leaves `up_ptrs_t` as `None` — I have an assessment running against Puzzle now and will report if it turns out reachable.
2. `ptrtable_n128` (128-divisor) is called from the MXFP4/`down_ptrs_t` paths in `forward_prefill_routed.rs`. If any target that resolves the **common** (64-wide) kernel can reach those call sites, that is the same half-write in a different costume. I could not determine the target set from the code alone.

rsafier:
Synced onto post-#490 `main` (`00e563c1`). One conflict, benign: #490 landed #477's three bitparity example stanzas and this PR adds a fourth — kept all four.

**The defect this PR fixes is still live on `main`.** #490 took #477 but not this one. Re-checked after the merge: `kernels/gb10/common/w8a16_gemv_batch4.cu` on main still accumulates

```c
acc[t] += __bfloat162float(lo) * wf[j * 2]
        + __bfloat162float(hi) * wf[j * 2 + 1];
```

against the M=1 kernel's two separate accumulations, and still documents itself as bit-identical *"verify with a cos>=0.9999 microtest"* — the exact claim that was false, and the exact test that cannot see it. Under `--fmad=false` those re-associate: 1–9 differing BF16 elements per launch, up to 4.0 apart, at production Lightning projection shapes.

Every batched-decode ladder that prefers this kernel over M × `w8a16_gemv` does so purely to save weight DRAM traffic and assumes the numbers do not move.

Final shape is 3 files, +246/−4: the kernel fix, the byte-parity gate that replaces the cosine microtest (`examples/w8a16_batch_bitparity_microtest.rs` — byte compare at [10304 × 2688] and [2688 × 4096] over three seeds, both batched tiers, with a 1-ULP negative control per (seed, shape) so a pass can never be vacuous), and its Cargo stanza.

Verified on the merge result: workspace check + clippy clean, 4617 tests pass, and the GPU gate reports byte-identical at every Lightning projection shape and every M both tiers serve.

CI is green apart from `PR benchmark gate` (records owed). Of the open stack this is the smallest change fixing an active correctness bug rather than adding capability.

tbraun96:

Closing by hand because the commits were **re-cut**, not merged: this PR's head is not an ancestor of main even though its content is, so GitHub will not auto-close it. Verified by artifact rather than assumed — `crates/spark-model/examples/w8a16_batch_bitparity_microtest.rs` and `kernels/gb10/common/w8a16_gemv_batch4.cu` are both present on main, and `crates/spark-model/Cargo.toml` carries the `[[example]] name = "w8a16_batch_bitparity_microtest"` entry with `required-features = ["cuda", "gpu-examples"]`.

PR #475, opened 2026-08-12
Nemotron-H milestone B: strided Mamba-2 conv/scan — C=2 +20%, C=4 +37%

rsafier:
Draft; **stacks on #467** (which stacks on #462) — review from `72b2211c` onward. Also currently **contains #474's two kernel fixes**; whichever of the two lands first, the other drops those hunks.

## What

Milestone A batched the Mamba-2 projections but left the conv1d + SSM scan running one launch *per row*, which is why batching only paid at padded rung ≥ 8 and lost at rungs 2/4. This adds the strided kernels and re-tunes every gate by measurement.

- `causal_conv1d_update_strided` + `mamba2_ssm_decode_strided` (kernels/gb10/common/, no target shadows them): one launch for N rows. Each was proven **bit-identical** to its per-row twin in an isolated GPU harness at production Lightning strides before any wiring — negative controls included.
- `decode_conv_scan.rs`: the multi-seq decode inner issues one strided launch pair over the dense slot prefix; fragmented tails keep the per-row loop.
- **The single rung gate is split.** Norms/residual/conv/scan are bit-identical to the per-seq loop and need no numeric gate → entry rung **8 → 2**. Only the projections reorder FP accumulation, so they keep a per-arm rung (`proj_batch_min()`): FP8 (the arm Lightning takes) at 2, NVFP4/BF16 still at 8.
- MoE batched body stays **disabled** — re-measured on the *fixed* grouped GEMM it is −58/−52/−40% at rungs 2/4/8. The prior numbers flattered it because they were taken against the half-writing kernel.

## Measured (Lightning-30B, GB10, 400-tok stories, sum-of-stream tok/s)

Independently re-measured against a build of the parent branch, 3 reps, best-of:

| C | parent | milestone B | delta |
|---|---|---|---|
| 1 | 69.24 | 68.98 | −0.4% (n=1 never enters the batched body) |
| 2 | 69.64 | **83.89** | **+20.5%** |
| 4 | 75.73 | **103.90** | **+37.2%** |
| 8 | 109.73 | 114.12 | +4.0% (isolates the strided conv/scan; rung 8 already batched projections) |

## Correctness

- **ssm-state-poisoning gate (#469): PASS** — 12 replays, 0 collapsed, 1 byte-identical, 11 jittered within bounds. Record committed.
- **Temp-0 fact battery at C=1/2/4/6, 2 reps: zero misses.** Notably the *parent* build answers "Pride and Prejudice → 1995" and "capital of Australia → Sydney" **even in its isolated C=1 control** — that is #474's half-zeroed prefill MoE, which this branch carries the fix for. Consequence worth flagging: the milestone-A determinism rationale still quoted in `MAMBA2_PROJ_MIN_NVFP4` ("1813 → 1995 flipped at C=3/4") was measured on that broken MoE, so **the NVFP4/BF16 rungs rest on invalid evidence and are probably too conservative** — re-measuring them is likely free performance.
- **Mixed-phase stress: PASS** — 5 concurrent 400-tok decodes with 3 short prompts injected mid-flight; 5/5 full-length and non-degenerate, 3/3 shorts correct, zero log errors.
- fmt / clippy --workspace --tests / cargo test --workspace clean.

## Known follow-ups (from an adversarial review of this diff; none catastrophic, all documented rather than hidden)

1. **MEDIUM-HIGH** — `decode_mtp_k.rs` gates K-row verify on the *concurrency* rung (`proj_batch_min()`), so at K≥8 it batches projections through arms never proven bit-exact. Concrete: Nano-30B (NVFP4 SSM) with `--num-drafts 7` makes MTP-on and MTP-off differ at temp 0. Should gate on a bit-exactness predicate (FP8 arm, both GEMV tiers resolved, K ≤ 16).
2. **MEDIUM** — `proj_batch_min()` returns 2 for FP8 without checking that `w8a16_gemv_batch4/16` actually resolved; if both were handle-0 it falls through to a tile GEMM that is not bit-exact. Latent on gb10 (kernel is unshadowed), live risk only on targets that lack it. One-line guard.
3. **LOW-MEDIUM** — the deleted rung-guard test was replaced by one that pins only the BF16 arm; the NVFP4 arm — where the original flip was observed — now has no test.
4. **LOW** — `decode_conv_scan.rs`'s header argues a pad row "can never satisfy base + i·slot_bytes". That is false (the dummy slot is index `max_slots` in the same arena, so with live slots 61/62/63 and padded_n=4 it does satisfy it). The implemented check is *stronger* than the stated argument — it compares against the row's own state pointer — so the code is sound; the comment needs correcting before someone simplifies the check on a wrong rationale.

MTP A/B was deliberately not re-run here (finding 1 makes it interpretable only at K=2/FP8, and the pre-existing verdict — 70.1 serial vs 55.0 forced — is unchanged in kind: milestone B cannot help MTP verify, because the K verify rows are time steps of one sequence, not independent sequences).


rsafier:
## Independent verification (second, unrelated measurement pass)

A separate verifier rebuilt both sides and re-measured everything. **All headline claims reproduce**, plus three corrections/additions to the description above.

**Throughput** (sum-of-stream, mean of 3 reps) — baseline absolutes reproduce the integrator's exactly:

| C | baseline | MB | delta | aggregate | TTFT ms |
|---|---|---|---|---|---|
| 1 | 69.22 | 68.89 | −0.5% | 66.7→65.7 | 230→298 |
| 2 | 69.24 | 83.62 | **+20.8%** | 66.8→78.8 | 449→606 |
| 4 | 75.29 | 103.76 | **+37.8%** | 72.1→96.6 | 882→1170 |
| 8 | 109.56 | 113.95 | +4.0% | 103.8→105.5 | 1637→2300 |

### Correction 1 — a cost the description omits: TTFT regresses 30–41%
Expected and explainable: prefill MoE now does its full work instead of half (the N-tile fix), so first-token latency rises while decode throughput climbs. Worth stating plainly rather than discovering in a gate.

### Correction 2 — the body's MTP claim was too strong
I wrote that milestone B "cannot help MTP verify." Measured, it helps materially — the penalty roughly halves:

| | serial | forced | accept |
|---|---|---|---|
| baseline | 69.22 | 51.28 | 52.2% |
| MB | 68.89 | **59.94** | **56.0%** |

−25.9% → **−13.0%**. Still net-negative, so the auto-gate still parks speculation (63.7 vs 68.9) — but the M=K verify is doing real work and the remaining gap is smaller than assumed.

### Correction 3 — scope of the ssm-state-poisoning PASS
That gate binds to the flagship `Qwen3.6-35B-A3B-FP8`, **not** a Nemotron target. It is a general engine check; it is *not* coverage of this diff's Mamba-2 state path. The Nemotron-specific state evidence here is the temp-0 battery (0/36 divergences at C=2/4/6) and the mixed-phase stress, not the gate.

### Measurement-protocol note for anyone re-running this
Natural-EOS sum-of-stream is unreliable — early EOS desyncs streams and inflates the metric (one rep read **5636 tok/s** when a stream emitted 4 tokens). Pin `min_tokens == max_tokens`; `ignore_eos` is silently ignored. With that pinned, run-to-run variance drops below 1%.

### Also independently confirmed
- w8a16 byte-parity **36/36**; reverting just the 2-line fused-add fix → **35/36 fail** (non-vacuous), and Lightning is confirmed on the FP8 arm at exactly the tested shapes.
- The N-tile bug is **live on `main`** and user-visible: baseline answers 4/6 temp-0 facts ("Pride and Prejudice → 1995", "capital of Australia → Sydney", stable across reps — degraded, not flaky), this branch 6/6. That is the strongest argument for landing #474 quickly, independent of this PR.

rsafier:
Merged `main` in. Two independent clean-merge hazards landed on this branch at once; both are fixed and pushed.

**1. `build_parse_behavior.rs`** — resolved on the Lightning base (#462) and carried through here. Details in that thread.

**2. The MoE grouped-GEMM N-tile pairing — a 2× buffer overrun.**

This branch still carried `0a106192`, which fixed the grouped-GEMM half-write by widening the kernel's `_ptrtable` entries to `N_TILE_PT 128`, matching a launcher that divided by 128 at the time. Main has since fixed the *same* defect the opposite way: narrowing the launcher to `ceil(n_out/64)` to match the existing `N_TILE 64`.

Each is self-consistent alone. Git merges them with no conflict — they touch different lines, in different languages — and the result is a 128-wide kernel on a 64-divisor grid. For `n_out=1856` that is 29 CTAs each owning 128 columns, covering 3712 columns of an 1856-column output. Not a half-write; a 2× overrun past the end of the buffer.

`tests/moe_grouped_ntile_geometry.rs` caught it (`grid.x=29`, expected `15`), which is the one thing in that commit that did its job.

**Conceding to main's repair**, and our own oracle numbers favour it rather than merely tolerating it: at N=1856 the 64-wide kernel on its correct 29-CTA grid measured **3.197 ms/launch** against **4.606 ms** for the 128-wide on 15 CTAs. Half the CTAs, ~44% slower.

Reverted `kernels/gb10/common/moe_w4a16_grouped_gemm.cu` and `layers/ops/moe_grouped_a.rs` to main byte-for-byte. Dropped the geometry test and the microtest example with them — the test pinned *our* pairing, and main deliberately runs two wrappers at different divisors (`ptrtable` `/64` for the common 64-wide kernel, `ptrtable_n128` `/128` for the 128-wide shadows), so the invariant is per-target and that test cannot express it.

Milestone B's own content — strided Mamba-2 conv/scan, the MTP verify kernels — is untouched. Workspace `check` + `clippy` clean, 4623 tests pass on the merge result.

Worth noting for the batch-merge order generally: this class of defect is invisible to CI, which builds the PR head while GitHub reports `MERGEABLE` off the textual merge. Both hazards here sat behind seven green checks. I am sweeping the rest of the open stack by actually building each merge result and will report anything else it turns up.

rsafier:
Status after Batch 5 (#490), for whoever picks this up next.

## Synced and green

Merged post-#490 `main` (`6a660866`). #490 landed this branch's own base (#477) plus the lossless-SW-GEMV / decode-graph cluster, so most of the conflict was this branch holding older copies of files main now owns — resolved toward main on all four (`bf16_batch_bitparity_microtest.rs` 248→257, `w4a16_batch_bitparity_microtest.rs` 280→317, `w4a16_gemv.cu`, and a union of the Cargo example stanzas).

Every CI check passes: clippy, fmt, `cargo test --workspace`, llvm-cov, metal, ≤500 LoC, `nvcc → PTX (all gb10 targets)`, kernel shadow, merge-ancestry, CLA. Verified locally on the merge result too — 4644 tests, clean clippy.

## All six gate records are invalidated — including both BFCL legs

This PR already carries records, but the main merge invalidated every one of them. Local `--pull-request-gate-check`:

```
NONE  agentic-webserver — latest record is for cb9f8ecab4 —
      invalidated by crates/atlas-core/src/config.rs, .../dispatch.rs,
      .../factory.rs and 64 more — device code changed for 23 target(s)
```

Identical for `ttft-warm-gate`, `ttft-cold-gate`, `bfcl-subset`, `bfcl-subset-echolp`, `ssm-state-poisoning-gate`. `governance/pr-475.jsonl` also has no recorded intent.

That is ~7 h of BFCL alone, so it is worth sequencing before paying it — see below.

## ⚠ The headline numbers in the description predate #490 and should be re-measured first

"C=2 +20%, C=4 +37%" was measured on the pre-#490 base, as was the independent verification pass. **#490 itself moved Nemotron decode substantially**, so the delta this PR claims is measured against a baseline that no longer exists.

Measured today on Nemotron-3.5-Lightning-30B-A3B-NVFP4, C=1, 38-token prompt, 400 output tokens, temp 0, 5 reps, identical server config (fp8 KV, util 0.60):

| build | tok/s (400) | tok/s (256) |
|---|---|---|
| pre-#490 top-of-stack | **69.4** | 68.3 |
| post-#490 top-of-stack | **56.0** | 55.3 |
| post-#490, `w4a16_gemv.cu` + `w4a16_gemv_fused.cu` reverted to pre-#490 | 63.7 | 62.7 |
| post-#490, those reverted **+ `ATLAS_NO_GEMV_SW=1`** | **69.4** | **68.3** |

The last row reproduces pre-#490 exactly on both shapes, so the −19.3% decomposes cleanly: ~57% from the two NVFP4 GEMV kernel rewrites, ~43% from the SW-GEMV dispatch. The pre-#490 figure also independently reproduces the 70.1 tok/s recorded in this target's own MODEL.toml on 2026-08-11.

Two scoping notes: it is **C=1-specific** — at C=4/C=8 the same `ATLAS_NO_GEMV_SW` A/B is genuinely neutral (84.8 vs 84.9, 90.9 vs 91.1, 3 reps each) — and it is measured only on Lightning. Nano, Puzzle and the Qwen NVFP4 targets share those kernels and are **unverified**.

## Suggested sequencing

This PR is the tip of a stack: it contains #462 and #467, which is why it reads as 73 files / +6972 against main. Landing #462 (which also carries the NoPE family fix that Nano/Puzzle/Super all need) and then #467 collapses this to milestone B's own content and shrinks the gate surface a wave can cover. Re-measuring the C=2/C=4 claim on post-#490 main before re-recording avoids paying the BFCL cost twice.

Draft status is just the stacking marker — one of its two reasons is gone now that #477 has landed via #490.

PR #476, opened 2026-08-13
fix(tui): stop the selection render test failing on a clock tick

tbraun96:
A pre-existing flake (since #385) that ejected #473 from the merge queue tonight.

`a_drag_highlights_what_it_covers_and_a_click_highlights_nothing` takes two renders microseconds apart and asserts that a highlight changes style but not text. The STARTUP pane reads a real `Instant` every frame, so when the tick landed between them `0.0s` became `0.1s` and the assertion failed on a difference unrelated to highlighting. The identical job had passed on the PR head minutes earlier.

**Why it wasn't fixed in #473:** it was, and then reverted. `crates` is in `PERF_PATHS`, and `gate/closure.rs` can only excuse changed paths inside `kernels/` — so a one-file, test-only edit invalidated all six gate records measured across three boxes over 2.5 hours. Here it costs one CI run.

The clock digits are normalised out rather than the assertion being loosened, so the test still catches real text drift. Verified across 40 consecutive runs.

### DIFF
--- a/crates/spark-server/src/tui/render/selection_render_tests.rs
+++ b/crates/spark-server/src/tui/render/selection_render_tests.rs
@@ -12,6 +12,45 @@ use super::*;
 // definition of "a rendered App" rather than two that can drift.
 use super::tests::{app, render};
 
+/// Blanks the STARTUP pane's elapsed clock (`0.0s`) before comparing renders.
+///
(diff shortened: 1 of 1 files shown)


tbraun96:

PR #477, opened 2026-08-13
fix(kernels): NVFP4 batched GEMV is not bit-identical to M x w4a16_gemv (third of the same class)

rsafier:
The third instance of the same class of defect as #474, and the one with the
widest blast radius: **every NVFP4 model's decode output has depended on how
many sequences happened to share the step.**

`w4a16_gemv_batchm_impl` (`batch4/8/16/32`) carried a header claim that had
never been checked —

> Per-row accumulation order is IDENTICAL to `w4a16_gemv` (M=1), so the output
> is bit-identical to running w4a16_gemv M times.

It was false in **three independent ways**, each of which alone breaks parity
under the dir's `--fmad=false`.

> **Status: the earlier throughput warning on this PR is withdrawn.** The first
> version of the fix cost −10.7% at C=4 and −18.5% at C=8. It has been
> re-engineered and re-measured; the isolated kernel sum is now **1.4% FASTER**
> than the unfixed body, and the worst e2e leg is −3.5% without speculation.
> One residual remains, quantified below: `batch8` (M=5..8), which shows −10.3%
> at MTP K=4 / C=2. The suggested mitigation in the old body — staging the
> second accumulator array in shared memory — was tried and is **wrong**; the
> real cause was never occupancy. See "What actually cost the 45%".

## The three causes

1. **K-chunk → lane mapping.** `w4a16_gemv` walks `k16 = lane*2` with stride
   128 in PAIRS (`c = 0,1`) into **two** accumulators; the batched body walked
   `k16 = lane` with stride 64 into **one**. A different lane partition means
   a different set of partial sums and so a different shuffle-reduction tree.
2. **Block-scale placement.** `w4a16_gemv` factors the FP8 group scale OUT of
   the 16-FMA block (`acc = fmaf(scale, part, acc)`); the batched body
   pre-multiplied it into every unpacked weight (`lut * scale`). Exact in real
   arithmetic, not in FP32.
3. **Fused vs split accumulation.** `acc += x*w0 + y*w1` associates as
   `acc + (x·w0 + y·w1)` where the M=1 kernel computes `(acc + x·w0) + y·w1`.
   This is the *same* defect as `w8a16_gemv_batch4` in #474, a third time.

## Blast radius

**No target shadows `kernels/gb10/common/w4a16_gemv.cu`** — `kernels/strix`
and `kernels/strix-hip` symlink it, so all three targets carried the defect
and all three get the fix. The callers are not Nemotron-specific:

- `crates/spark-model/src/model/trait_impl/lm_head_batched.rs` — the batched
  LM head, i.e. **every model doing concurrent decode**
- `crates/spark-model/src/layers/dense_ffn.rs`
- `crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/ffn.rs`,
  `.../multi_seq/qkv.rs`, `.../multi_seq/attn/o_proj.rs`
- `crates/spark-model/src/layers/moe/forward_k3.rs`
- `crates/spark-model/src/layers/mtp_head/*`

So Qwen3.5/3.6, Holo, Laguna, MiniMax and Nemotron were all affected.

## What actually cost the 45% — and why the old diagnosis was wrong

The first fix was a literal transcription: physical thread `l` walks
`k16 = l*2` with stride 128 in pairs, exactly like the M=1 kernel. Bit-exact,
and 45% slower. The old body attributed that to the second accumulator array
costing a CTA/SM. **That is not what happened.** Each of the three causes was
priced separately against the unfixed body, at the four production shapes:

| variant | what it changes | isolated total |
|---|---|---|
| unfixed (`avarok/main`) | — | 11.539 ms |
| cause 3 only | split `acc += x*w0 + y*w1` | 11.423 ms (**−1.3%**) |
| causes 2+3 | + scale factored out of the 16-FMA block | 11.236 ms (**−2.9%**) |
| cause 1 as well (first fix) | + the pair lane mapping | 16.768 ms (**+45.3%**) |

Causes 2 and 3 — the arithmetic ones — are **free**, in fact slightly
negative. All 45 points are cause 1, and two further probes show it is not
occupancy:

- `__launch_bounds__` restores **every** tier's pre-fix CTA/SM (batch4 5,
  batch8 4, batch16 3, batch32 2) spill-free, and recovers **4 of the 45
  points** (16.09 ms).
- Keeping the pair mapping AND both accumulator arrays at 122 registers /
  2 CTA per SM, but addressing the **activations** with the stride-32 pattern,
  lands on **baseline (−0.4%)**.

It is coalescing. Under `k16 = lane*2` the lanes of a 16-byte activation load
sit 64 B apart, so one instruction touches 2048 B to use 512 — half the sector
efficiency of the stride-32 walk. The activation stream is *M times* the weight
stream, which is exactly why the cost tracked M (batch2 +11%, batch16 +49%).
Widening the **weight** load to one 16-byte `uint4` recovers nothing (+44%),
confirming the weight stream was never the problem.

Staging `acc1[]` in shared memory — the mitigation the old body suggested — was
implemented and measured. It does not work, and the reason is instructive:
batch16 still allocates **124 registers**. The pressure was never the
accumulators; it is the `MAX_M` activation vectors the compiler hoists when two
K-chunks are in flight. That variant lands at 16.26 ms.

## The fix: a virtual-lane remap, not a math change

Every `fmaf`, every operand and the whole reduction tree are exactly what the
parity fix established. Only the *walk* changes.

The kernel walks K **coalesced** (consecutive lanes read consecutive chunks —
the pre-fix pattern) and recovers parity by remapping which *reference*
accumulator each thread stands in for. For physical thread `p` (0..63) and
`phase` in {0,1}, recalling that the M=1 kernel's lane `l` owns chunks
`{2l + c + 128j}` in accumulator `c`:

```
chunks {p + 64*phase + 128j} == 2*(p/2 + 32*phase) + (p%2) + 128j
                             == ref lane (p/2 + 32*phase), accumulator p%2
```

Same chunk set, same order, so the same fmaf chain — each pass *is* one of
`w4a16_gemv`'s chains. The kernel runs that chain twice, phase 0 then phase 1,
with **one accumulator array reused**; the two phases read disjoint halves of
the weight row, so total DRAM traffic is unchanged. Carrying both chains at
once instead costs a second array and ~20% at batch8/16 (measured).

`acc_l = acc0_l + acc1_l` is then one `__shfl_xor_sync(...,1)` between threads
`p` and `p^1`, which hold accumulator 0 and 1 of the *same* reference lane and
are always in the same warp — they add in opposite operand order, which is
exact because FP32 addition is commutative. A 64-float smem buffer re-lands the
results in virtual-lane order and the **unchanged** two-warp shuffle tree then
reduces lanes 0..31 and 32..63 exactly as `w4a16_gemv` does. Cost: one shfl per
row per phase and one smem round trip per (row, output) — not per chunk.

### Register table (`ptxas -v`, `-arch=sm_121a -O3 --fmad=false -DTQ_PLUS_SIGNS`; 256-thread blocks, 64K regfile, confirmed against the driver JIT's own `CU_FUNC_ATTRIBUTE_NUM_REGS` and `cuOccupancyMaxActiveBlocksPerMultiprocessor`)

| kernel | pre-fix | first fix | **this PR** | smem | CTA/SM: pre-fix → first fix → **this PR** |
|---|---|---|---|---|---|
| `batch2` | 38 | 42 | **40** | 2.2 KB | 6 → 5 → **6** |
| `batch3` | 40 | 52 | **48** | 3.2 KB | 6 → 4 → **5** |
| `batch4` | 48 | 61 | **48** | 4.3 KB | 5 → 4 → **5** |
| `batch8` | 54 | 79 | **48** | 8.5 KB | 4 → 3 → **5** |
| `batch16` | 70 | 126 | **56** | 17.0 KB | 3 → 2 → **4** |
| `batch32` | 106 | 140 | **80** | 33.9 KB | 2 → 1 → **3** |
| `qg_batch2` / `dual_batch2` | 38 / 40 | — | **38 / 40** | 128 B | 6 → **6** |
| `qg_batch3` / `dual_batch3` | 40 / 40 | — | **40 / 40** | 160 B | 6 → **6** |

Zero spills on every tier. Occupancy is at or **above** the pre-fix body
everywhere. Only `batch8` carries a `__launch_bounds__` (48 regs / 5 CTA beats
its free 56 / 4 by 4%); pinning the wide tiers as well measures 12.70 ms
against 11.49 ms free, so they are left alone. The added smem is `MAX_M × 1 KB`
and never binds — batch16 is register-limited at 4 CTA with 17 KB against
100 KB per SM.

### Isolated kernel timings (sum over every gate leg, 300 reps, best of 3)

| tier | pre-fix | first fix | **this PR** |
|---|---|---|---|
| `batch2` | 0.277 ms | 0.308 ms **+11.4%** | 0.275 ms **−0.7%** |
| `batch3` | 0.304 ms | 0.374 ms **+22.9%** | 0.305 ms **+0.3%** |
| `batch4` | 0.964 ms | 1.167 ms **+21.0%** | 0.943 ms **−2.2%** |
| `batch8` | 1.969 ms | 2.958 ms **+50.2%** | 2.166 ms **+10.0%** |
| `batch16` | 8.025 ms | 11.961 ms **+49.0%** | 7.677 ms **−4.3%** |
| **total** | **11.539 ms** | **16.768 ms +45.3%** | **11.375 ms −1.4%** |

## Deliverable 3: `batch2` / `batch3` carried it too — fixed here

`w4a16_gemv_batch2` and `w4a16_gemv_batch3` are **separate fixed-M kernels**,
not instantiations of the template, serving K=2/K=3 spec verify plus C=2/C=3
multi-seq decode. They had never been byte-checked. They carried the same
three divergences (batch3 with the fused form verbatim) and additionally
walked K in 8-value chunks rather than 16. They are now thin instantiations of
the bit-exact shared template, which deletes ~200 lines of duplicated
dequant/reduce code.

## Deliverable 4: the fifth family, now FIXED (was "reported not fixed")

`w4a16_gemv_qg_batch2`, `w4a16_gemv_dual_batch2`, `w4a16_gemv_qg_batch3` and
`w4a16_gemv_dual_batch3` are non-parity against a *different* reference:
`w4a16_gemv_qg` (M=1), which has its own `k8 = lane` chunking and
pre-multiplied scale. Against that reference all four diverged in exactly
**one** way — cause 3: `acc += x*w_lo + y*w_hi` where the M=1 qg/qkvz kernels
do two separate `acc +=`. Chunking and scale placement already matched.

Fixed with one split per accumulator, 10 sites. No lane remap, no register
cost — all four stay at 38–40 registers / 6 CTA per SM.

New standing gate `w4a16_qg_batch_bitparity_microtest`, 3 seeds × 3 Q/Gate
shapes × 4 kernels. `dual_batch2/3` write a plain `C[n]`, so they are gated
against `w4a16_gemv_qg` driven with `num_heads = 1, head_dim = N`, which makes
its deinterleave the identity map and leaves the arithmetic untouched.

| kernel | legs | before | after |
|---|---|---|---|
| `qg_batch2` | 9 | **7** | 0 |
| `qg_batch3` | 9 | **8** | 0 |
| `dual_batch2` | 9 | **7** | 0 |
| `dual_batch3` | 9 | **8** | 0 |
| **total** | **36** | **30** | **0** |
| negative controls firing | 9 | 9 | 9 |

The 6 legs that matched before the fix did so by coincidence, all at the
narrowest shape — the same pattern as the 5 coincidental matches in the
`batch2/3/4/8/16` table below.

## Parity tables

Gate `w4a16_batch_bitparity_microtest` — RAW BF16 BYTES (not a cosine; a
cosine gate is exactly what hid the #474 defect) at production projection
shapes `[10304x2688]`, `[2688x4096]`, `[18560x4096]`, `[4096x8192]`, seeds
1/99/12345, every M each tier serves. 204 legs + 12 negative controls so a
"byte-identical" verdict can never be vacuous.

### w4a16 batched (legs DIFFERING vs M × `w4a16_gemv`)

| tier | legs | before | after |
|---|---|---|---|
| `batch2` (fixed-M) | 12 | **12** | 0 |
| `batch3` (fixed-M) | 12 | **10** | 0 |
| `batch4` / `batch8` / `batch16` (template) | 180 | **177** | 0 |
| **total** | **204** | **199** | **0** |
| negative controls firing | 12 | 12 | 12 |

Before: 1–62 BF16 elements differed per launch, max|delta| 0.0625. The 5 legs
that matched by coincidence were all at the shallowest shape.

### BF16 arm (`bf16_batch_bitparity_microtest`) — known gap, NOT fixed

Included as **standing evidence, not a fix**; this gate is expected to report
`FAIL (EXPECTED)` and says so in its header and its output.

| shapes × seeds × M | legs differing | elements | max abs | max rel |
|---|---|---|---|---|
| 84 | **84 / 84** | 0.08%–0.50% | 0.0625 | 0.60 |

Batching the BF16 arm swaps `m × dense_gemv_bf16` for one
`dense_gemm_bf16_pipelined` (m16n8k16 tensor-core MMA) — a different
*algorithm*, not a reassociation that can be un-fused. The clean fix is a
bit-exact batched BF16 GEMV, a kernel port, deliberately out of scope here.

## Fails-without-fix demo

Reverting **only** the `.cu` hunk to `avarok/main` and rebuilding, with the
gate binaries otherwise unchanged:

```
w4a16_batch_bitparity_microtest      199 of 204 legs byte-identical=false
w4a16_qg_batch_bitparity_microtest    30 of  36 legs byte-identical=false
                                     (21/21 negative controls still firing)
```

Restoring the fix: `204 / 204` and `36 / 36` byte-identical, all controls
firing, exit 0 on both.

## Cross-model perf: Qwen3.6-27B-NVFP4 (dense, GB10 / dgx-00)

`unsloth/Qwen3.6-27B-NVFP4`, `--gpu-memory-utilization 0.80 --max-seq-len
8192`, fp8 KV (MODEL.toml default). 45 s fixed-concurrency windows with C lanes
issuing back-to-back, 400-token stories, `min_tokens == max_tokens` pinned;
per-stream rate computed decode-only from the server's TTFT. tok/s reported as
**sum-of-stream / aggregate**, best of N reps. Both sides are the same tree
at the same commit; only `kernels/gb10/common/w4a16_gemv.cu` differs.

Chosen because it is **not** Nemotron and it exercises both hot paths: the
batched LM head, and at `--speculative` the MTP head at K=4
(`default_num_drafts = 3`, confirmed in the serve log). The unfixed column was
re-measured in this same session and reproduces the earlier run to within
0.5%, so the two generations of numbers are comparable.

### (a) No speculation — 3 reps (rep-to-rep spread ≤ 0.1 tok/s)

| C | tier | unfixed (`avarok/main`) | first fix | **this PR** | Δ vs unfixed |
|---|---|---|---|---|---|
| 1 | `w4a16_gemv` (M=1, untouched) | 13.9 / 13.9 | 13.9 / 13.9 | **13.9 / 13.9** | **0.0%** |
| 2 | `batch2` | 25.3 / 25.2 | 25.7 / 25.6 | **25.1 / 25.1** | **−0.5%** |
| 4 | `batch4` | 45.9 / 45.8 | 41.1 / 41.0 (−10.7%) | **45.3 / 45.2** | **−1.2%** |
| 8 | `batch8` | 68.8 / 68.7 | 56.1 / 56.1 (−18.5%) | **66.4 / 66.3** | **−3.5%** |

### (b) MTP speculative decode, K=4 — 2 reps

| C | tier (C × 4 drafts) | unfixed (`avarok/main`) | first fix | **this PR** | Δ vs unfixed |
|---|---|---|---|---|---|
| 1 | M=4 → `batch4` | 22.6 / 22.4 | 20.0 / 19.8 (−10.3%) | **22.0 / 21.9** | **−2.4% / −2.3%** |
| 2 | M=8 → `batch8` | 32.9 / 31.9 | 24.0 / 23.8 (−27.1%) | **29.6 / 29.2** | **−10.3% / −8.4%** |
| 4 | M=16 → `batch16` | 50.0 / 48.8 | 49.3 / 48.2 (−1.6%) | **49.5 / 48.8** | **−0.9% / 0.0%** |
| 8 | M=32 → `batch32` | 80.2 / 74.3 | 79.9 / 65.5 (−11.2% agg) | **78.7 / 71.5** | **−1.8% / −3.7%** |

### (c) Residual, stated plainly

The regression is gone everywhere except **`batch8` (M=5..8)**, which is
+10.0% in isolation and shows up as **−3.5% at C=8 without speculation and
−10.3% at MTP K=4 / C=2**. Every other leg is within ±2.4%, and `batch16` /
`batch32` are *faster* than the unfixed body.

The cause is specific and understood: the two-phase walk pays one extra loop
setup, and `batch8` is the tier where that is not amortised. Per shape, the
penalty is +15% at `[10304 x 2688]` (K16 = 168, so ~1.3 inner iterations per
phase) and **−4%** at `[4096 x 8192]` (K16 = 512, 4 iterations per phase) — it
is loop overhead at shallow K, not the arithmetic and not occupancy (`batch8`
has *more* CTA/SM here than before, 5 vs 4). Three attempts to close it were
measured and discarded because each was worse: hoisting the row base pointers
(no change, the compiler already does it), `#pragma unroll 2` on the inner K
loop (batch16 +64%), and peeling the phase loop (batch16 +67%). Staging the
second accumulator in smem with a single loop is worse still (batch8 2.53 ms).

So: **the correctness fix is perf-neutral overall (−1.4% isolated, 0.0% to
−3.5% e2e without speculation) with one known −10% leg at MTP K=4 / C=2.**
That leg is a follow-up, not a blocker — and it is a fifth of what the same
configuration cost before.

### (d) Temp-0 sanity battery on this build

6/6 correct, both with and without speculation:

| probe | expected | no spec | MTP K=4 |
|---|---|---|---|
| Pride & Prejudice year | 1813 | 1813 | 1813 |
| capital of Australia | Canberra | Canberra | Canberra |
| symbol for gold | Au | Au | Au |
| planets | 8 | 8 | 8 |
| Moon landing year | 1969 | 1969 | 1969 |
| longest river in Africa | Nile | Nile | Nile |

The unfixed build also answers 6/6, as expected — these probes are a
correctness floor, not a discriminator. The defect's signature is
batch-size-dependent *drift*, which six temp-0 facts will not reliably
surface; the byte-parity gates are what discriminate.

## Relationship to other PRs

- **Sibling: #474** (`fix/kernel-parity-two`) — the MoE grouped-GEMM N-tile
  half-write and `w8a16_gemv_batch4`'s fused-add. Same class, found first.
  #474 also introduces `examples/w8a16_batch_bitparity_microtest.rs`; the
  negative-control retrofit to that file is therefore **omitted here** to
  avoid a conflict — it should land on #474 or as a follow-up.
- **#475** (`feat/nemotron-milestone-b`) currently carries its own copy of
  this `.cu` hunk, and its NVFP4 projection rung drop (`MAMBA2_PROJ_MIN_NVFP4`
  8 → 2) **depends on this fix** — that rung is only sound while the tiers it
  opens are byte-identical. #475 should land **after** this PR, or drop its
  copy of the hunk and rebase.


tbraun96:

PR #478, opened 2026-08-13
spark-model: keep slot-keyed decode graphs across sequence turnover

facepunch47:
## Summary

Decode CUDA graphs are keyed by SSM **slot**, not by request occupant. `free_sequence` still destroyed them on every completion, so the next request to reuse that slot paid a full eager recapture (64-layer build + sync).

This PR retains slot-keyed decode / K=2/3/4 verify / batched-decode graphs across turnover. LRU insertion still bounds batched-graph memory, and now destroys a replaced handle instead of leaking it. Graphs that bake a per-occupant LoRA adapter index (`verify_kgamma`, `fused`) still drop on free.

This is a first-principles decode-wall cut: GB10 decode is memory-bound, but recapture is an extra eager step that is not on the roofline.

## Approach

```mermaid
flowchart TD
    product["Product: peak GB10 inference"]
    product --> bound["Decode wall = weight bandwidth + recapture"]
    bound --> graphs["This PR: persist slot-keyed CUDA graphs"]
    bound --> gemv["Follow-up: default lossless single-warp GEMV"]
    graphs --> retain["Retain decode / verifyK / batch caches"]
    graphs --> lru["LRU still bounds memory; destroy replaced handles"]
    graphs --> lora["Still drop LoRA-baked kgamma/fused"]
```

## Why this is safe

- Dynamic KV / seq_len / block_table / positions are uploaded to **fixed staging** before every replay (`decode_a.rs`).
- SSM `h_state` / `conv_state` pointers are `ssm_pool.*(layer, slot)` — identical for any occupant of that slot.
- LoRA rotation already drains every graph cache in `destroy_lora_decode_graphs`.
- Batched graphs stay keyed by the slot **vector**; a composition that has not been seen still captures. The cap is `16 + decode_meta_rows`.

## Test plan

- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo test -p spark-model --lib decode_graph_key`
- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy -p spark-model --tests -- -D warnings`
- [x] `bash scripts/check-license-headers.sh`
- [x] GB10 prove (this machine, `nvidia/Qwen3.6-35B-A3B-NVFP4`, speculative, `--kv-high-precision-layers auto`):
  - `avarok/atlas-gb10:latest` recaptures `CUDA graph captured successfully for slot=0` on **every** sequential request (98.8 tok/s lifecycle).
[...]

PR #479, opened 2026-08-13
spark-model: default lossless single-warp GEMV on decode

facepunch47:
Decode on GB10 is memory-bandwidth bound. The bit-identical single-warp GEMV (`w4a16_gemv_sw`, 8 outputs/block, no cross-warp smem reduce) already exists and the microtest requires 100% bit-identity — but it was opt-in behind `ATLAS_DECODE_OPT` and the default split-SiLU **down** projection never used it at all.

This ships that kernel on the decode path that actually runs.

## What changes

- **Default ON** via `ModelLevers::gemv_sw`. Kill switch: `ATLAS_NO_GEMV_SW=1` (exactly `"1"`; `=0` does not disable).
- Dense FFN: `w4a16_gemv_dual_sw` for gate+up; `w4a16_gemv_sw` for split-SiLU / GELU down. Dual and silu-input SW are independent — a missing `silu_input_sw` no longer skips `dual_sw`.
- SSM sequential QKVZ + out_proj: same `w4a16_decode_gemv` helper.
- Grid SSOT: Rust `W4A16_GEMV_SW_OUTS_PER_BLOCK = 8` is pinned to CUDA `#define N_PER_BLOCK_SW 8`. Swapping the kernel without swapping the grid writes the wrong outputs.

Out of scope (follow-up): attention / MTP / Nemotron `w4a16_gemv` call sites — same helper, more wiring.

## Why this is a product win

```mermaid
flowchart LR
  subgraph bound [GB10 decode]
    BW["273 GB/s LPDDR5X"]
    GEMV["W4A16 GEMV"]
    GEMV --> BW
  end
  subgraph before [Before]
    B1["dual GEMV 64-thread"]
    B2["silu_mul"]
    B3["down GEMV 64-thread"]
    B1 --> B2 --> B3
  end
  subgraph after [After — default]
    A1["dual_sw 8 outs/block"]
    A2["silu_mul"]
    A3["gemv_sw 8 outs/block"]
    A1 --> A2 --> A3
  end
  before --> after
```

Same bytes moved, fewer idle warps waiting on `__syncthreads`, same numerics. The kernel was already in `kernels/gb10/common/w4a16_gemv.cu`; this is the dispatch default.

## Tests

- `gemv_sw_from` polarity (`unset`/`=0` ON, `=1` OFF)
- lever + live handle required
- SW grid covers N and is half the base grid when N is divisible by 8
- CUDA `#define N_PER_BLOCK{,_SW}` matches the Rust constants on every `w4a16_gemv.cu` / `_fused.cu` copy
- `ModelLevers::defaults().gemv_sw` is ON

GPU oracle (not CI): `cargo run --release -p spark-model --example w4a16_gemv_sw_microtest`

## Related

Persisting slot-keyed decode graphs (fewer recaptures) is https://github.com/Avarok-Cybersecurity/atlas/pull/478 — complementary, not overlapping.

## GB10 serve proof

- Bit-identity: `w4a16_gemv_sw_microtest` on this GB10, seed `0x51A7`, 100% match vs 64-thread `w4a16_gemv`.
- Kernel SSOT: `kernels/gb10/common/w4a16_gemv.cu` — `w4a16_gemv` / `_sw` share `w4a16_gemv_partial`; dual/dual_sw share `w4a16_dual_partial`. strix/strix-hip `.cu` are symlinks. Grid: base `ceil(N/4)`, SW `ceil(N/8)`, block 256.
- Rebased onto `origin/main` (`680b3a5`, includes #473 TUI). Independent of #478/#481.
- Kill switch is `ATLAS_NO_GEMV_SW=1` only (`=0` does not disable).
- Out of scope: main-model `w4a16_gemv_logits`; attention/MTP/Nemotron wiring is #480.

## CLA

### DIFF
--- a/crates/spark-model/examples/w4a16_gemv_sw_microtest.rs
+++ b/crates/spark-model/examples/w4a16_gemv_sw_microtest.rs
@@ -93,12 +93,21 @@ fn gen_weight(rng: &mut Rng, n: usize, k: usize) -> Weight {
     }
 }
 
-fn cosine(a: &[u16], b: &[u16]) -> (f64, usize) {
+struct Cmp {
+    cos: f64,
+    bit_id: f64,
+    mismatches: Vec<(usize, u16, u16)>,
+}
+
+fn compare(a: &[u16], b: &[u16]) -> Cmp {
     let (mut dot, mut na, mut nb) = (0f64, 0f64, 0f64);
     let mut bit_eq = 0usize;
+    let mut mismatches = Vec::new();
     for i in 0..a.len() {
         if a[i] == b[i] {
             bit_eq += 1;
+        } else if mismatches.len() < 4 {
+            mismatches.push((i, a[i], b[i]));
         }
         let x = f32::from_bits((a[i] as u32) << 16) as f64;
         let y = f32::from_bits((b[i] as u32) << 16) as f64;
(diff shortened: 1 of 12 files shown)


facepunch47:

facepunch47:
recheck

tbraun96:

PR #484, opened 2026-08-13
feat(laguna): Laguna-XS-2.1 support — its own kernel target + NVFP4-packed shared-expert loader arm

rsafier:
Bucket **b** of #442, cut from the GPU-validated `integrate/laguna-on-main` tree onto current `avarok/main`.

## What Laguna-XS is

Poolside ships Laguna 2.1 at two hidden sizes. `main` already serves **S-2.1** (hidden 3072, 48 layers, top_k 10, ~68 GB NVFP4) — it arrived via #473. **XS-2.1** is the small sibling: hidden 2048, 40 layers, 48 Q-heads, top_k 8, moe/shared intermediate 512, same 256 experts and same `model_type = "laguna"`, ~21 GB as NVFP4. Same hybrid full/sliding attention, same sigmoid router with routing bias.

Before this PR XS could not boot at all, and would have rendered wrong if it had. Two independent reasons, one fix each.

### 1. No kernel target claimed it

`ptx_for_config(model_type, hidden_size)` picks a target from the `[[model_types]]` claims in each MODEL.toml: an exact `(type, hidden_size)` claim wins, a claim with no `hidden_size` is the wildcard fallback, and nothing matching returns `None`. `laguna-s-2.1` claims `(laguna, 3072)` only, and there is no `laguna` wildcard — so a 2048-hidden checkpoint matched neither rule.

`kernels/gb10/laguna-xs-2.1/` claims `(laguna, 2048)`. Its `nvfp4/KERNEL.toml` is a **symlink to S's**, so the two targets compile the identical 167-module set; the two `.cu` files are the same symlinks S already uses (`minimax-m2-229b`'s grouped GEMM, `deepseek-v4-flash`'s w4a16).

A target of its own is also what lets XS state `thinking_default = false`. S keeps thinking ON per poolside's agentic-coding recommendation, but on XS the reasoning block leaked into responses. With one shared MODEL.toml that split could only have been a hardcoded `model_type == "laguna" && hidden_size == 2048` gate in `serve.rs`.

### 2. The shared expert was being double-quantized

The two checkpoints disagree on shared-expert precision, and each says so in its own `quantization_config`:

| | routed experts | shared expert |
|---|---|---|
| **S-2.1** | in `targets` → NVFP4-packed | in `ignore` → BF16 `.weight` |
| **XS-2.1** | in `targets` → NVFP4-packed | in `targets` → NVFP4-packed |

`load_moe_ffn` assumed the S layout unconditionally: `dense_auto` the BF16 `.weight`, then `quantize_to_nvfp4` it, then install the BF16 copy as authoritative. On XS that BF16 tensor does not exist.

The fix detects by **tensor presence** (`.weight_packed`) rather than by hidden_size, so a future variant is classified by what it actually contains. When packed, load through `quantized_v2` and skip `set_bf16_shared_expert` — `weights.shared_expert` then rides the same NVFP4 machinery as the routed experts, and `has_mixed_bf16_shared_expert()` reports false so the fused routed kernels keep their shared contribution instead of recomputing it in BF16. The S path is unchanged and still loads each BF16 tensor exactly once (the branch is structured to avoid a duplicate `dense_auto` of the shared weights).

## Files

```
A  kernels/gb10/laguna-xs-2.1/MODEL.toml                        (203 lines)
A  kernels/gb10/laguna-xs-2.1/nvfp4/KERNEL.toml                 → symlink ../../laguna-s-2.1/nvfp4/KERNEL.toml
A  kernels/gb10/laguna-xs-2.1/nvfp4/moe_w4a16_grouped_gemm.cu   → symlink ../../minimax-m2-229b/nvfp4/…
A  kernels/gb10/laguna-xs-2.1/nvfp4/w4a16_gemm.cu               → symlink ../../deepseek-v4-flash/nvfp4/…
M  crates/spark-model/src/weight_loader/laguna/load_layers.rs   (+60/−12, file stays at 464 LoC)
A  crates/atlas-kernels/tests/model_type_routing.rs             (177 lines)
```

## Kernel-audit harvest

A new kernel target owes a `--check-kernels` harvest — this repo fails closed on handle-0 lookups, so a missing one is a dead boot. Run on the XS checkpoint against this build:

```
kernel check PASSED for (laguna-xs-2.1, sm_121, nvfp4): 232 lookups, 35 expected-absent
{"model":"laguna-xs-2.1","arch":"sm_121","quant":"nvfp4","kernel_set_hash":"563c67ead7c7",
 "modules_embedded":167,"lookups":232,"unresolved":0,"expected_absent":35,"ok":true,
 "unresolved_kernels":[]}
```

All 35 declarations were exercised — none is dead weight. The set matches `laguna-s-2.1`'s, which is what the identical module set predicts; every entry carries its rationale in the MODEL.toml. The boot audit on the real serve run below is clean too.

## Test

`crates/atlas-kernels/tests/model_type_routing.rs`:

- `both_laguna_hidden_sizes_route_to_their_own_target` — 2048 → `gb10/laguna-xs-2.1`, 3072 → `gb10/laguna-s-2.1`, each claimed by exactly one target (a second claimant would make `ptx_for_config` resolve by target enumeration order), plus a non-vacuity assert that no wildcard `laguna` claim exists to satisfy both lookups for the wrong reason.
- `claimed_hidden_size_matches_documented_hidden_dim` — every target's `[[model_types]]` hidden_size must agree with the `[model].hidden_dim` it documents; the two are written independently and a mismatch routes a checkpoint onto a target tuned for another shape.

Both read the MODEL.toml files from the tree rather than the compiled registry, so they hold on the GPU-free CI runner (`ATLAS_SKIP_BUILD=1`), where `available_targets()` is an empty stub.

**Negative control:** with `kernels/gb10/laguna-xs-2.1/` moved aside, the first test fails with

```
assertion `left == right` failed: Laguna-XS-2.1 (hidden_size 2048) must be claimed by
gb10/laguna-xs-2.1 and only by it; unclaimed means ptx_for_config returns None and XS
cannot boot at all
  left: []
 right: ["gb10/laguna-xs-2.1"]
```

and passes with the target restored.

## GPU validation

GB10 (dgx-00), `poolside/Laguna-XS-2.1-NVFP4`, `--gpu-memory-utilization 0.80 --no-tui`, `RUST_LOG=info`, full-target build (`ATLAS_TARGET_MODEL='*' ATLAS_TARGET_QUANT='*'`).

- **Boot: 29.7 s** (process start → `Server live and ready`), target resolved as `(sm_121, laguna-xs-2.1, nvfp4)`, 21.6 GB pre-KV weights, 40 per-layer checkpoint k_scale/v_scale picked up for the fp8 KV cache (no calibration), boot kernel audit clean.
- **3/3 temp-0 facts** (`enable_thinking: false`, `max_tokens: 250`):
  - "capital of Australia" → *The capital of Australia is Canberra.*
  - "who wrote Pride and Prejudice" → *Jane Austen wrote the novel \*Pride and Prejudice\*.*
  - "chemical symbol for gold + atomic number" → *The chemical symbol for gold is Au, and its atomic number is 79.*
- **Decode: 47.8 tok/s** end-to-end, single stream, 250 completion tokens on a 62-token prompt (wall 5.22 s, TTFT included).
- MODEL.toml behavior confirmed live in the log: `thinking_default=false`, non-thinking/tools presets `temp=0.6, top_k=20, top_p=0.95`, thinking-loop watchdog / F2 confidence early-stop / rollback-resteer all disabled, `tool_call_parser=poolside_v1` with XGrammar constrained decoding armed.

`cargo fmt --all`, `cargo clippy --workspace --tests`, and `cargo test --workspace` are all green. (`--all-targets` additionally trips one pre-existing `manual_is_multiple_of` lint in `crates/spark-model/examples/w4a16_gemv_sw_microtest.rs`, untouched by this branch — last modified in #229.)

## Deliberately excluded

- **Keep-packed Q4_K_M** — that is bucket e, in flight as #465. Not duplicated here; the loader hunks in this PR touch only the shared-expert branch.
- **XS native-BF16 loader** — permanently excluded per #442. It loads and generates coherently but has an unresolved expert-compute correctness bug (factual recall degrades against the NVFP4 checkpoint of the *same* model), and it was env-gated, which production paths do not allow.
- **The integration tree's tokenizer/jinja bits** (`prefer_bundled`, `resolve_jinja_includes`, the `{% generation %}` strip) — **superseded on main**. Those existed so XS could prefer its own bundled `chat_template.jinja` over the shared `laguna` override authored for S. Main's curated `jinja-templates/laguna.jinja` is already semantically XS's template (`enable_thinking` defaults to false, no `preserve_thinking`), so XS renders correctly on main's existing override design. Verified by diffing the override against both bundled templates.

### DIFF
--- a/crates/atlas-kernels/tests/model_type_routing.rs
+++ b/crates/atlas-kernels/tests/model_type_routing.rs
@@ -1,0 +1,177 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! `[[model_types]]` routing pins.
+//!
+//! `atlas_kernels::ptx_for_config(model_type, hidden_size)` picks a compiled
+//! kernel target by matching the `[[model_types]]` claims declared in each
+//! `kernels/{hw}/{model}/MODEL.toml`. An exact `(model_type, hidden_size)`
+//! claim wins; a claim without `hidden_size` is the wildcard fallback; nothing
+//! matching returns `None` and the model cannot boot at all.
+//!
+//! That makes the claim table load-bearing in a way nothing else checks.
+//! Poolside ships Laguna at two hidden sizes, and `laguna-s-2.1` claims
+//! `(laguna, 3072)` only — so before `laguna-xs-2.1` existed a 2048-hidden XS
+//! checkpoint matched neither the exact rule nor a wildcard and
+//! `ptx_for_config` returned `None`. The variants also disagree on
+//! `thinking_default` and on their sampling presets, which is expressible only
+//! as two MODEL.toml files.
+//!
+//! These run against the MODEL.toml files in the tree rather than the compiled
+//! registry, so they hold on the GPU-free CI runner (`ATLAS_SKIP_BUILD=1`),
+//! where `available_targets()` is an empty stub.
+
+use std::collections::BTreeMap;
+use std::path::{Path, PathBuf};
+
+fn kernels_root() -> PathBuf {
+    Path::new(env!("CARGO_MANIFEST_DIR"))
+        .parent()
+        .and_then(Path::parent)
+        .expect("crates/atlas-kernels is two levels below the workspace root")
+        .join("kernels")
+}
+
+/// One `[[model_types]]` entry: `(model_type, hidden_size)` -> declaring targets.
+type Claims = BTreeMap<(String, Option<u64>), Vec<String>>;
+
+/// Every `[[model_types]]` claim under `kernels/{hw}/*/MODEL.toml`, keyed by
+/// the pair claimed and valued by the `{hw}/{model}` targets claiming it.
+fn claims() -> Claims {
+    let mut out: Claims = BTreeMap::new();
+    for (target, manifest) in manifests() {
+        let text = std::fs::read_to_string(&manifest).expect("manifest is readable");
+        let parsed: toml::Value = text
+            .parse()
+            .unwrap_or_else(|e| panic!("{} is not valid TOML: {e}", manifest.display()));
+        let Some(entries) = parsed.get("model_types").and_then(|v| v.as_array()) else {
+            continue;
+        };
+        for entry in entries {
+            let model_type = entry
+                .get("model_type")
+                .and_then(|v| v.as_str())
+                .unwrap_or_else(|| {
+                    panic!(
+                        "{} has a [[model_types]] entry with no model_type",
+                        manifest.display()
+                    )
+                })
+                .to_string();
+            let hidden = entry.get("hidden_size").and_then(|v| v.as_integer());
+            out.entry((model_type, hidden.map(|h| h as u64)))
+                .or_default()
+                .push(target.clone());
+        }
+    }
+    assert!(
+        !out.is_empty(),
+        "no [[model_types]] claims found under {} — the walk is broken, not the tree",
+        kernels_root().display()
+    );
+    out
+}
+
+/// `({hw}/{model}, path to its MODEL.toml)` for every kernel target in the tree.
+fn manifests() -> Vec<(String, PathBuf)> {
+    let root = kernels_root();
+    let mut out = Vec::new();
+    let hw_dirs = std::fs::read_dir(&root).expect("kernels/ is readable");
+    for hw in hw_dirs.flatten().map(|e| e.path()).filter(|p| p.is_dir()) {
+        let Ok(models) = std::fs::read_dir(&hw) else {
+            continue;
+        };
+        for model in models.flatten().map(|e| e.path()).filter(|p| p.is_dir()) {
+            let manifest = model.join("MODEL.toml");
+            if !manifest.is_file() {
+                continue;
+            }
+            out.push((
+                format!(
+                    "{}/{}",
+                    hw.file_name().unwrap().to_string_lossy(),
+                    model.file_name().unwrap().to_string_lossy()
+                ),
+                manifest,
+            ));
+        }
+    }
+    out.sort();
+    out
+}
+
+/// Both Laguna hidden sizes must route, each to its own target and to nothing
+/// else. The `assert_eq` on the full claimant list is deliberate: a second
+/// target claiming the same pair would make `ptx_for_config` resolve by target
+/// enumeration order.
+#[test]
+fn both_laguna_hidden_sizes_route_to_their_own_target() {
+    let claims = claims();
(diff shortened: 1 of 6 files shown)


rsafier:
Heads-up from the wip/laguna-integration bring-up (2026-08-25): XS serves and answers correctly on this PR's loader (validated live), and speculative decoding now works on XS with both the DFlash and community DSpark drafters — but the enabling pieces live outside this PR by design: the sliding-attention verify fix went to main directly (#745, XS is a sliding-window model so any spec on it needs that), and the laguna-drafter arch port + DSpark config shims + `[expected_absent]` DFlash-kernel declarations depend on the dflash2 line and ride with that consolidation. Nothing needed here — this PR merges as-is.

tbraun96:
Closing as **already landed**. Laguna-XS-2.1 reached main via **#1189** (`37cf8e7135`, "stack: glm5_next prefill + Laguna-XS-2.1 composed for one certification campaign") — one of the composed campaigns, certified there rather than here.

Checked file by file rather than inferred from the title. Of the six files in this PR:

| file | on main |
|---|---|
| `kernels/gb10/laguna-xs-2.1/MODEL.toml` | **byte-identical** |
| `kernels/gb10/laguna-xs-2.1/nvfp4/KERNEL.toml` | **byte-identical** |
| `kernels/gb10/laguna-xs-2.1/nvfp4/moe_w4a16_grouped_gemm.cu` | **byte-identical** |
| `kernels/gb10/laguna-xs-2.1/nvfp4/w4a16_gemm.cu` | **byte-identical** |
| `weight_loader/laguna/load_layers.rs` | identical **but for the rename** — the only 8 differing lines are `atlas_core` → `avarok_core` and `ATLAS_*` → `AVAROK_*` |
| `*-kernels/tests/model_type_routing.rs` | main has a **superseding** version |

That last one is why this branch now conflicts, and it is worth reading rather than merging. This PR's copy requires every `[[model_types]]` `hidden_size` to **equal** `[model].hidden_dim`. Main's copy requires it to be **included in** the claim set, and its comment explains why the strict form was wrong: `kernels/{b200,b300,gb10}/kimi-k3/MODEL.toml` deliberately matches two shapes (7168 production beside the 1024 0.40B twin) under one documented `hidden_dim`, and the strict form failed all three K3 targets. Merging this branch would revert that fix.

Nothing here is missing from main. Thanks @rsafier — if you think some part did not make it across, reopen and name the file and I will re-check.

PR #486, opened 2026-08-13
tools: stop delivering tool calls whose arguments Atlas authored

rsafier:
## What

Independent defects on the tool path let a call the model never finished writing reach the client **looking complete**. They were found together on Nemotron-3.5 Lightning 30B, but each is model-agnostic and lives in the shared server path.

> **Scope change (2026-08-13).** This PR originally carried three fixes. **Fix 2 — the tool-call-aware `</think>` force-close — has been removed**; it was measured to never engage. The reasoning is in §2 below and in the [verification comment](https://github.com/Avarok-Cybersecurity/atlas/pull/486#issuecomment-5278847982). Shipping: fixes 1, 3 and the MTP-depth discriminator.

The live symptom: `get_weather` + *"What's the weather in Santiago? Use the tool."*, temperature 0 → `{"city": ""}`, deterministically, 5 runs out of 5. `"Kyoto"` passed. Same build, same prompt shape.

### 1. `backfill_required_params` no longer invents `""`

This is the damaging one. An empty string is a **valid** instance of `{"type": "string"}`, so a fabricated one passes:

* Atlas's own `assess_tool_call`, which deliberately does not police empty strings (Theia's `getWorkspaceFileList` legitimately sends `path=""`), **and**
* the client's JSON-Schema `required` check, because the key *is* present.

So a call carrying only its name became indistinguishable from a complete one, and `get_weather` ran with an argument **Atlas** wrote. Nothing downstream could tell.

Leaving the key absent restores the intended failure surface. Values that can be genuinely **derived** from what the model did produce still are — `description` restates the `command`, `subagent_type` is read off the tool description's agent list — and that split is now explicit in `derive_required_param` instead of being a fallthrough in a placeholder loop.

**Disposition** follows the conventions already in the file rather than inventing a new one:

| tool class | omitted required param | result |
|---|---|---|
| ordinary / read-only | `city` | `MissingParam` → call attached as produced, error recorded. Dropping it outright was the 2026-07-03 ST-995 empty-response collapse; the client's `required` check is what gives the model actionable feedback. |
| write family (`PATH_KEYS`) | `file_path` | `EmptyRequired` → **not** attached. |
| shell family (`CMD_KEYS`) | `command` | `EmptyRequired` → **not** attached. |

The last two rows used to hold *by accident*: backfill fabricated `""` and the F78 empty-string rule caught it. With nothing fabricated, absence now reaches that check directly, so the rule was extended to cover it. `group_i_poolside_safety::missing_write_path_is_not_executable_after_backfill` is the existing test for that property and stays green.

### 2. ~~The `</think>` force-close is tool-call aware~~ — implemented, measured inert, **removed**

This fix was written, reviewed, ablated, and then taken back out. It is documented rather than silently dropped because the *reason* it failed is a property of the codebase that outlives this PR.

**What it did.** `should_inject_think_end` deferred the forced `</think>` while a tool call was open, exactly like the existing in-code-fence rule and bounded by the same `hard_override`; the thinking-loop watchdog additionally would not *arm* mid-call. Both used one predicate, `helpers::tool_call_open_in_tail`.

**Why it is gone.** [Independent verification](https://github.com/Avarok-Cybersecurity/atlas/pull/486#issuecomment-5278847982) established both halves of the case:

* **No regression.** A/B of this head against a fix-2-reverted build — Nemotron-3-Nano, thinking forced on, temp 0, no tools, **default thresholds** — the watchdog fires identically on every prompt (thinking tokens 168/168, 432/432, 208/208; control silent) and **all eight compared response fields are byte-identical**. Loop protection never depended on it.
* **It never engages.** An instrumented build that `WARN`s whenever `in_tool_call` would flip the decision logged **zero** engagements across Qwen3.6-35B-A3B in-`<think>` drafting (F7 hoisting 1–2 calls per response), thinking-budget sweeps 80→400 on two models, a 10-required-parameter tool, and a repetitive-argument tool.

**The structural reason.** `tool_call_open_in_tail` is **token-level**: it matches the atomic `<tool_call>` special token (id 248058, correctly resolved). The models that actually draft a call inside `<think>` never emit it — **Qwen3.6 writes the draft as plain text**, **Nemotron Nano writes a bare `<function=…>`** recovered by `param_as_function_salvage`. F7 hoists from **decoded text**, which is precisely why F7 sees calls this predicate cannot.

It was caught sitting inert **while the exact failure it targets occurred**: `thinking_token_budget=310`, an in-`<think>` call cut mid-body → 8/10 args → `log_incident requires parameter 'mitigation'`, guard not engaging, behaviour identical on the reverted build.

A guard that provably does not guard is worse than no guard — it reads as coverage in a place nothing is covered. **Fix 1 is what handles those cases**, by surfacing `MissingParam` instead of a fabricated `""`.

Making the predicate text-aware is tracked as **#491**.

**What was kept.** `helpers::tool_call_open_in_tail` itself, and its unit tests. It is not dead code: it is the behaviour-neutral dedupe of the **content-phase** fuzzy detector's inline copy, which has excluded `inside_tool_call` since 2026-04-26 and for which token-level matching is adequate. Removing it would mean restoring that inline duplicate for no gain. Its doc comment now records the token-vs-text limitation so the next reader does not re-derive it. `confidence.rs`, `forced_think_end.rs`, `confidence_tests.rs` and `pipeline_tests.rs` are byte-identical to `main`.

### 3. The qwen3_coder parser recovers XML-attribute drift

Under `tool_choice="auto"` the grammar is a **late** trigger — it engages only once the model has emitted `<tool_call>\n<function=NAME>` verbatim. A model that opens with a bare `<function=NAME>` therefore decodes **unconstrained**, and Lightning then writes:

```
<function=get_weather>
<parameter city="Santiago">
Chile
</parameter>
</function>
```

The strict `<parameter=` loop finds nothing, so the call reached the client with no arguments at all — 4 of 10 cities in a temperature-0 sweep. This is exactly the drift `compile_tools.rs` predicted when the value EBNF was relaxed to admit `<` ("let the model fall into XML-attribute syntax"), with parser-side recovery named there as the standing defence in depth.

The salvage is narrow: one attribute, bare identifier, fully quoted value. It handles both dialects — `<parameter KEY="VALUE">` (attribute is the value; the body is drift continuation) and `<parameter name="KEY">VALUE` (body is the value) — respects the sibling-`</function>` boundary, and never overwrites a key the strict loop parsed.

### Also: an explicit MTP-depth discriminator in the kernel-target matcher

`[[model_types]]` entries may now pin `num_nextn_predict_layers`, and `ptx_for_config(model_type, hidden_size)` becomes `ptx_for_shape(ModelShape { … })` with the most **specific** declaring entry winning. **Inert on its own** — no target declares the field in this PR — but it is what lets two checkpoints identical in `(model_type, hidden_size)` stop sharing one target's `[behavior]` policy. That consumer is #NEXT.

The signature change is deliberate rather than an added optional argument: every caller now states the MTP depth it means. A defaulted parameter here is the silent target fallthrough this repo has been bitten by twice.

## Tests

Fails-without, verified by reverting only the behaviour lines and re-running:

```
group_j_no_fabrication::truncated_call_keeps_required_param_absent    FAILED
group_j_no_fabrication::truncated_call_is_reported_as_missing_param   FAILED
group_j_no_fabrication::undecidable_required_param_stays_absent_...   FAILED
group_a_formats::parse_qwen3_coder_empty_body_then_backfill           FAILED
```

New: `tool_parser/tests/group_j_no_fabrication.rs` (13 cases — no-fabrication, the attribute-drift dialects and the shapes the salvage must *refuse*, and the per-tool-class disposition table), plus `tool_call_open_in_tail` cases in `scheduler/helpers_tests_tool_call.rs` covering the retained content-phase predicate. `atlas-kernels/src/lib_tests.rs` gains specificity-ordering coverage for the matcher.

The four in-tool-call gate cases in `scheduler/confidence_tests.rs` went with fix 2.

One existing test changed semantics: `group_a_formats::parse_qwen3_coder_empty_body_then_backfill` asserted `args["command"] == ""`. It now asserts `command` is absent and that validation still names it. The path-A/path-B parity that test was written to protect is unchanged; only the fabricated value is gone.

`cargo fmt`, `cargo clippy --workspace --tests`, `cargo test --workspace` clean.

## Validation

Live on GB10, Nemotron-3.5 Lightning 30B NVFP4, temperature 0, 10-city sweep (`Kyoto Lisbon Cairo Oslo Mumbai Santiago Nairobi Toronto Helsinki Denver`):

| | before | after |
|---|---|---|
| sequential | 6/10 correct args (4× `{"city": ""}`) | **10/10** |
| C=8 concurrent | — | **10/10**, 0 cross-request contamination |
| C=2 concurrent | — | **10/10**, 0 cross-request contamination |
| Santiago × 5 | 5/5 `{"city": ""}` | 5/5 `{"city": "Santiago"}` |

Validation needed the Nemotron branch, since Lightning only loads there. The change itself has no Nemotron dependency.

**Re-verified WITHOUT fix 2** (2026-08-13, dgx-00 GB10, release build of the stacked Lightning branch — the only tree Lightning loads on; `nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-NVFP4`, temp 0, `ATLAS_NO_DECODE_GRAPHS=1`, gpu-util 0.80, same 10 cities):

| leg | with fix 2 | **fix 2 removed** |
|---|---|---|
| sequential | 10/10 | **10/10** |
| Santiago × 5 | 5/5 `{"city": "Santiago"}` | **5/5 `{"city":"Santiago"}`** |
| C=8 concurrent | 10/10, 0 contamination | **10/10, 0 contamination** |
| C=2 concurrent | 10/10, 0 contamination | **10/10, 0 contamination** |

Unchanged on every leg, as expected: fixes 1+3 are what deliver it. `finish_reason: tool_calls`, `get_weather`, `{"city":"Santiago"}` on all five Santiago runs.

**On fix 2.** The original diagnosis was that the loop watchdog cut the in-`<think>` call mid-arguments; an ablation on this branch already showed the sweep result unchanged with fix 2 reverted (the cut lands *after* the call). Independent verification then went further and showed the guard never engages at all, for the token-vs-text reason in §2. It has been removed. The Santiago result was fixes 1+3, and it is unaffected — see the re-verification row below.

---

### Changes since first review

* **Fix 2 removed** (`revert(486): drop fix 2 …`) — see §2. Follow-up: **#491**.
* **500-LoC cap.** `atlas-kernels/src/lib.rs` (516) split — MODEL.toml `[sampling.*]` preset types moved to a sibling `sampling.rs`, re-exported from the crate root, so `build.rs`-generated code and every `atlas_kernels::SamplingPresets` path are unchanged (516 → 434).
* **`nvcc -> PTX` job.** The red was a real test failure, not the `$HOME differs from euid` runner noise in the same log: `tests::nemotron_h_2688_splits_by_mtp_depth` asserted a `nemotron-3.5-lightning-30b-a3b` kernel target that **only the stacked Lightning PR creates**. This PR adds the `mtp_layers` *discriminator*; the second target that makes it discriminate anything comes with that PR, so the registry-level test moved there. The three pure-function specificity tests — what this PR's code actually implements — stay and pass.


rsafier:
Independent verification of the fix-2 (tool-call-aware `</think>`) risk, plus a finding that should change how its reach is described.

**Regression check: PASS — fix 2 is inert for non-tool traffic.** A/B of this head against a build with fix 2 surgically reverted (dropped the `!tool_call_open_in_tail(...)` conjunct; hard-wired `in_tool_call = false`), same tree/GPU/requests, on Nemotron-3-Nano with thinking forced on, temp 0, **no tools**, at **default thresholds** (`think_loop_min_repeats=3`, `scan_window=160` — no scratch config needed; Nano loops in `<think>` naturally, which is the degeneracy its own MODEL.toml documents):

| prompt | head fires | reverted fires | thinking_tokens |
|---|---|---|---|
| chant | yes | yes | 168 / 168 |
| continue_pattern (control) | no | no | — |
| mantra2 | yes | yes | 432 / 432 |
| forever | yes | yes | 208 / 208 |

All eight compared response fields (finish_reason, completion/reasoning tokens, reasoning chars + tail, content head, watchdog-line count) matched on every prompt; a hair-trigger leg (`repetition_detection.min_count=2`) also matched exactly, firing at 160 on both. Loop protection is unchanged.

**However — the guard engaged ZERO times on any tool traffic we could construct.** An instrumented third build that WARNs whenever `in_tool_call` would flip the decision was run against Qwen3.6-35B-A3B with an in-`<think>` drafting system prompt (F7 hoisted 1–2 calls per response), thinking-budget sweeps 80→400 on two models, a 10-required-parameter tool, and a tool with a deliberately repetitive argument value. No engagements.

The reason is structural and worth fixing: `tool_call_open_in_tail` is **token-level**, matching the atomic `<tool_call>` special token (id 248058, correctly resolved). But the models that actually draft calls inside `<think>` don't emit that token — Qwen3.6 writes the draft as **plain text**, and Nano writes a bare `<function=…>` recovered by `param_as_function_salvage`. F7 already operates on *decoded text*, which is why it hoists calls the guard cannot see. Caught live on this head at `thinking_token_budget=310`: an in-`<think>` call was cut mid-body → 8/10 args → `log_incident requires parameter 'mitigation'`, **with the guard not engaging** (identical on the reverted build). On the injector path the pre-existing `in_code_fence` rule already dominates for fenced drafts.

So: safe to ship, but it currently protects only models that emit `<tool_call>` as an atomic special token inside `<think>`. Suggest making the predicate text-aware to match what F7 does — otherwise the guard misses the cases that motivated it. (Fix 1 is doing the real work in those cases, correctly surfacing `MissingParam` instead of a fabricated `""`.)

**One more thing for this PR's own validation:** Nemotron-3.5 Lightning — the model this bug was observed on — **cannot be served from this branch**: `Failed to parse nemotron_h config.json → invalid type: null, expected usize`. The Lightning parser fixes live on #462, not on this PR's `main` base. Demonstrating the motivating case here requires merging/rebasing onto the Lightning port.

Evidence under `/home/ms/.claude/jobs/5a7bd33d/tmp/wdreg/` (`ab-head.json`, `ab-rev.json`, instrumented sweep `p12-inst.json`).

tbraun96:
Triaged for a certification campaign tonight and **sent back for a rebase** —
with specifics, because the conflict list is misleading about what is wrong.

I merged this head onto current `main` (`5f639bfc57`) locally. It conflicts in
15 files, and almost none of them are about tool-call arguments — they are
`crates/avarok-kernels/src/lib.rs`, eight `crates/spark-model/examples/*.rs`,
TUI data modules, `serve_load.rs`. That reads like unrelated drift, but it is
not drift: **the branch is carrying a 236-commit-old copy of main's own API and
would revert it.** Two changes landed underneath you:

1. **The crates were renamed** (`569ffd0920` "rebrand: Atlas to Avarok on the
   site, crates, env, and cfg"). `atlas_kernels` is now `avarok_kernels`,
   `ATLAS_*` env vars are `AVAROK_*`, and `cfg(atlas_*)` is `cfg(avarok_*)`.
2. **Kernel lookup was redesigned.** Your side calls
   `atlas_kernels::ptx_for_shape(ModelShape { model_type, hidden_size, mtp_layers })`;
   main has `avarok_kernels::ptx_for_config(model_type, hidden_size, &refs, pin)`
   and split the crate into `ptx_set` / `query` modules
   (`TargetPtxSet`, `ModelTypeMatch`, `available_targets`, `ptx_for_model`).
   `ModelTypeMatch` also grew an `mtp_layers: Option<usize>` discriminator.

Taking your side of those hunks would delete main's lookup API, so this is not
something to resolve mechanically on your behalf — a rebase from you is the
right and much smaller job, since your own change is only **3 commits**.

The fix itself is one we want: refusing to deliver tool calls whose arguments
Atlas authored rather than the model is exactly the kind of honesty defect worth
landing. Push a rebase and it goes into the next composed campaign.

PR #487, opened 2026-08-13
nemotron: give Nemotron-3.5 Lightning its own kernel target (stacks #486, #467)

rsafier:
Stacks on **#462** → **#467** → **#486**. Base is `main` per this repo's stacking convention (Avarok has no `feat/nemotron-batched-decode` branch to target), so the diff below carries those three PRs' content ahead of it. **The change this PR adds is the last two commits: `kernels/gb10/nemotron-3.5-lightning-30b-a3b/` and the `[[model_types]]` line in Nano's MODEL.toml.**

## What

Nemotron-3.5 Lightning has been served by `nemotron-3-nano-30b-a3b`'s kernel target since 2026-08-11, because the matcher keyed only on `(model_type, hidden_size)` and both checkpoints are `(nemotron_h, 2688)`.

**The kernels never diverged.** `kernels/gb10/nemotron-3.5-lightning-30b-a3b/nvfp4/` is a byte-for-byte clone of Nano's: same `KERNEL.toml`, the same two symlinked overrides (`rms_norm.cu`, `w4a16_gemm.cu`), same 166 compiled modules, same `kernel_set_hash`. The build cost is ~0 — the nvcc dedup reports the whole target as cache hits.

**The policy diverged, and could not be expressed.**

* Nano documents `thinking_default = false`: thinking leaks training-corpus text into its answers (measured 2026-07-28, with the empty-system-turn fix already in place).
* Lightning wants thinking **on** for general chat quality but **off** on tool turns. With thinking on it writes its `<tool_call>` inside `<think>`, which costs ~4× (89–97 completion tokens vs 23) and leaves the call at the mercy of wherever `</think>` lands.

That is `thinking_default = true` **and** `thinking_in_tools = false` — one target cannot hold it alongside Nano's `thinking_default = false`. The standing workaround made it strictly worse: `--default-chat-template-kwargs '{"enable_thinking":true}'` is an *explicit directive*, and an explicit directive **overrides** `thinking_in_tools` (`api/chat/thinking.rs`), so tool turns thought anyway and the knob was unusable.

## The discriminator

`num_nextn_predict_layers`, stated **explicitly on both sides**:

```toml
# nano
hidden_size = 2688
num_nextn_predict_layers = 0     # ships no MTP draft head

# lightning
hidden_size = 2688
num_nextn_predict_layers = 1     # DeepSeek-style 1-step head under mtp.layers.*
```

Omitting it on either side would restore a 2688-wide `nemotron_h` catch-all — the exact silent fallthrough the `hidden_size` pin was added to remove in 2026-06-06. The matcher machinery that reads it is #486; the two are only useful together.

Nano's `default_num_drafts = 1` is annotated as inert (no proposer loads without a draft head) and the measured Lightning MTP story — drafts good, verify net-negative without K-row batched Mamba-2/relu²-MoE verify kernels — moves to the target it describes.

## `--check-kernels`

This repo fails closed on handle-0 lookups, so a new target without a harvested `[expected_absent]` is a dead boot. Harvested and verified on both:

```
nemotron-3.5-lightning-30b-a3b   149 lookups, 9 expected-absent, 0 unresolved   PASSED
nemotron-3-nano-30b-a3b          150 lookups, 9 expected-absent, 0 unresolved   PASSED
```

Identical sets, as expected from an identical kernel set — Lightning's MTP head adds no new kernel lookups.

## Validation

Served **without** the `--default-chat-template-kwargs` workaround (that is the point):

```
Selected kernel target: (sm_121, nemotron-3.5-lightning-30b-a3b, nvfp4) (166 modules)
Model behavior: thinking disabled when tools active (MODEL.toml)
Model behavior: max_thinking_budget=2048, thinking_default=true
```

| check | result |
|---|---|
| tool turn, client silent on thinking | 25 completion tokens, `reasoning` absent, `{"city": "Santiago"}` |
| plain chat turn, client silent | thinks — 2285 chars of reasoning |
| "capital of Australia" | "Canberra is the capital of Australia." |
| 10-city tool sweep, temp 0, sequential | 10/10 correct args |
| same at C=8 / C=2 concurrent | 10/10 each, 0 cross-request contamination, 0 tool turns with reasoning |
| Nano after the split | still resolves `nemotron-3-nano-30b-a3b`, `thinking_default=false` |

`cargo fmt`, `cargo clippy --workspace --tests`, `cargo test --workspace` clean on the stack.

## Why this must land after #467, not on `main` alone

Lightning's MODEL.toml uses `[behavior]` keys that only exist on the Nemotron branch — `no_decode_graphs` (the family crashes under decode-graph replay: CUDA 700 on Puzzle, 716 on Nano/Lightning) and `default_num_drafts`. Based on `main` those keys would parse as unknown and be silently dropped, taking the decode-graph suppression with them. The target also cannot be *exercised* on `main`, where the Lightning checkpoint does not load.


rsafier:
Closing in favour of #544, which lands Lightning support re-ported onto current `main`.

This PR cannot be rebased. Its unique contribution is Lightning's own kernel target, discriminated via `ptx_for_shape(ModelShape { model_type, hidden_size, mtp_layers })`. Main has since replaced that API: `atlas-kernels/src/resolve.rs` breaks a `(model_type, hidden_size)` tie with per-target **`match_names`** needles matched against the checkpoint reference, built for the Qwen3.6-vs-3.8 collision — the same problem this solved a different way.

Measured against current main (`38395946`) before closing:

* `git rebase` conflicts on the first replayed commit with 19 more to go.
* `git merge` produces **16 conflicted files**, and the substantive ones are not mechanical — every `ptx_for_shape` call site has been rewritten to `ptx_for_config`, and `ModelShape` no longer exists.

#544 therefore re-expresses the discriminator on main's design (`match_names = ["lightning"]` here, `["nano"]` on the Nano target, which main's `build.rs` requires of both sides of a colliding pair) and does not reintroduce `ModelShape` / `ptx_for_shape`.

**#467 and #486 are unaffected and remain open** — neither is ported into #544.

The branch is untouched, so nothing here is lost.

AzeezIsh:
Verified on a GB10 (spark-256a) against current `main` today. Flagging a rebase blocker that isn't visible as a normal conflict.

**This branch's discriminator no longer exists on `main`.** #487 separates Lightning from Nano by MTP depth via `ptx_for_shape(ModelShape{ model_type, hidden_size, mtp_layers })`. #513 (`03e28e6e`) deleted both `ptx_for_shape` and `ModelShape` — `git grep -n "ptx_for_shape\|ModelShape" origin/main -- crates/atlas-kernels/src` returns nothing. The replacement is `resolve.rs::ptx_for_config(model_type, hidden_size, model_refs, pin)`, which breaks config-identical ties on `[model] match_names` needles and hard-errors `Ambiguous` rather than picking by build order.

Two consequences worth knowing before the rebase:

1. `build.rs::validate_collision_match_names` is an `assert!` — colliding targets must *both* declare `match_names`. Nano declares `(nemotron_h, 2688)` with none today, so landing this target unchanged **panics the build**; bypassed, it would hard-error Nano at serve time too.
2. Because 4 of the 5 PRs branch from `680b3a56` (pre-Batch-5), the `serve_load.rs` / `model_swap.rs` / `tui/data/*` conflicts are all this same removed API.

I ported it as a check: dropped the `mtp_layers` machinery, added `match_names = ["nemotron-3-nano"]` / `["nemotron-3.5-lightning"]`, took `main`'s side on the resolver call sites. It builds and resolves correctly with **no** `--kernel-target`:

```
Selected kernel target: (sm_121, nemotron-3.5-lightning-30b-a3b, nvfp4) (167 modules)
{"lookups":153,"unresolved":0,"expected_absent":9,"ok":true}
```

**One trap that cost me a real bug:** resolving `serve_load.rs` to `main`'s side silently drops this branch's `config.no_decode_graphs = ptx_set.behavior.no_decode_graphs;` assignment. Six decode/verify paths gate on `!self.config.no_decode_graphs`, so it stays `false` and graphs are never disabled — the CUDA 716 hazard #462 fixed, with no error and no log line. Worth an explicit line in the rebase checklist.

Tradeoff to be deliberate about: `match_names` matches substrings of the HF id / `--model-name` / model dir, so a renamed or bare local dir (`--model-from-path /model`) can no longer resolve and needs `--kernel-target`. The config-derived MTP discriminator didn't have that failure mode. `main`'s CLI help already documents the pin as the intended answer.

PR #489, opened 2026-08-13
fix(tui): route Help section keys to the Help section, not the ? overlay

tbraun96:
**Reported:** editing the title in Help ▸ Report Issue freezes the dashboard.

**Actual:** not a freeze — every keystroke was being thrown away, so nothing moved. `in_input()` is consulted before the `Section::Help` arm in `App::on_key`, so once the title entered edit mode all keys went to `on_input_key`, whose Help branch called `on_help_overlay_key` — the `?` modal's scroll handler. Its `_ =>` arm discards the key and closes an overlay that was not open.

**Origin:** the Help section and the audit branch had each independently defined `App::on_help_key`. Resolving that collision during the stack merge renamed the overlay one, and this call site got renamed along with it because it matched the pattern, not because it meant the overlay. Self-inflicted during #473's conflict resolution.

**Test:** drives `App::on_key`, deliberately — the existing `help_state_tests` sit one layer below the defect and pass against the broken build. Verified by reintroducing the bug (test fails: "every typed character must reach the title buffer") and removing it again.

Also asserts Esc leaves edit mode while **keeping** the draft title, since destroying a user's typing on Esc would be its own bug.

2100 tests pass, fmt + clippy clean.

### DIFF
--- a/crates/spark-server/src/tui/app_input.rs
+++ b/crates/spark-server/src/tui/app_input.rs
@@ -46,7 +46,12 @@ impl App {
             return;
         }
         if self.section == Section::Help {
-            self.on_help_overlay_key(key);
+            // `on_help_key` (the SECTION's keys), not `on_help_overlay_key`
+            // (the `?` modal's scroll). They differed by one word and the
+            // wrong one swallowed every character in its `_ =>` arm, so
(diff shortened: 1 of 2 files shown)


tbraun96:

PR #495, opened 2026-08-13
fix(moe): pin the router-logits GEMM to the scalar kernel — restores BFCL 86.95 on the 35B FP8 MoE, and ratchets the floor to 86.50/86.90

### LINKED ISSUE #492: BFCL echolp (35B FP8 MoE) regressed 86.95→85.34 with the batch4+TUI stack, and the gate floor is 0.10 away from missing it
## The regression

`bfcl-subset-echolp` — the **35B-A3B-FP8 MoE** leg, N=1004 draw — dropped when `680b3a568` (the #464 batch4 + #473 TUI stack) landed:

| | normalized | overall |
|---|---|---|
| held for weeks (many shas) | **86.95** | **86.55** |
| `03a74eac19` (the stack, now main) | 85.34 | 84.76 |
| `cb9f8ecab4` (Batch 5, #490) | **85.34** | **84.76** |

Two independent runs, on different trees, producing the **same score to the hundredth**. That rules out MTP-greedy nondeterminism (measured band ±0.40) — this is deterministic. Batch 5 neither caused it nor recovers it.

**It is passing the gate by 0.10** (floor `overall_accuracy min = 84.66`).

## It is FP8-MoE-specific

The dense 27B NVFP4 leg (`bfcl-subset`, N=995) is unaffected across the same shas — 87.64 overall before and after, normalized moving only +0.35 (inside noise). Whatever changed does not touch the dense path.

## Suspects

That merge carried exactly two FP8-MoE-only changes, both in `crates/spark-model/src/layers/moe/forward_prefill_fp8.rs`:

1. **Router logits GEMM rerouted** `ops::dense_gemm` → `ops::dense_gemm_prefill` (pipelined kernel). Primary suspect: it changes *which experts win top-8* on borderline tokens, which is how you get an accuracy shift rather than a precision wobble.
2. **New BF16 shared-expert path** — `run_bf16_shared_expert()` now runs and, on success, bypasses both the W8A8 and dense fallbacks.

Bisecting these two against the 86.95 tree is the next step. Both are cheap to A/B: flip one, re-run the echolp leg.

## Raise the floor to 86.50

The current floor let a 1.79-point deterministic regression through with 0.10 to spare. Proposed for `kernels/gb10/qwen3.6-35b-a3b/BENCH.toml`, `bfcl-subset-echolp`:

```toml
[benchmarks.metrics.overall_accuracy]
min = 86.50    # was 84.66
noise = 0.4
```

The gate compares `value + noise >= min`, so 86.50 with noise 0.4 fails below **86.10** — 0.45 below the 86.55 best-ever, i.e. outside the noise band, so it will not false-fail on nondeterminism. It **would** have caught this regression (84.76 < 86.10).

Note this raises the floor above the value currently on main, so **the floor should be raised together with the fix, not before it** — otherwise every PR fails the gate for a defect it did not introduce. If the fix proves hard, the honest alternative is to raise it anyway and treat main as knowingly red, but that is a decision to make deliberately.

Worth also raising `normalized_single_turn_score` from 83.32 toward its own best-ever 86.95 in the same change; leaving it at 83.32 keeps a 3.6-point hole.

tbraun96:
Fixes #492.

## The regression

`bfcl-subset-echolp` (Qwen3.6-35B-A3B-FP8, N=1004, temp 0) held **86.95 / 86.55 bit-identically across ~14 shas** from 08-07 to 08-12, then dropped to **85.34 / 84.76** with the batch4+TUI stack — and reproduced *to the hundredth* on a later, different tree (`cb9f8ecab4`). Measured MTP band is ±0.40, so this is deterministic, not noise. The dense 27B NVFP4 leg is unaffected.

## The culprit

The batch4 stack rerouted the **MoE router-logits GEMM** from `ops::dense_gemm` (scalar kernel, sequential FP32 accumulation over K) to `ops::dense_gemm_prefill`, which prefers `dense_gemm_bf16_pipelined` — an `mma.sync.aligned.m16n8k16` tensor-core kernel. Different accumulation order, not bit-identical output.

**Why that costs accuracy rather than showing up as noise:** the router logits are the model's only discrete control flow. They are stored to BF16 and then `moe_topk_softmax_batched` takes the top-8. On the coarse BF16 grid, near-tied experts sit ~1 ulp apart, so an accumulation-order difference **flips expert selection** on borderline tokens — a macroscopic change to that token's hidden state, at every MoE layer, on every prefill token, deterministically. Everywhere else in the model an order change is score noise; here it is a selection flip.

**Why the dense 27B is spared:** it has no router. There is no discrete decision for a ULP to flip.

## The fix

New SSOT helper `MoeLayer::router_gate_gemm_dense` (`layers/moe/helpers_c.rs`) pinned to the scalar `dense_gemm`, used by all three prefill bodies (fp8 / bf16 / nvfp4). `dense_gemm_prefill` stays on the shared-expert **data** path, where order changes are genuinely noise.

The `ATLAS_CUBLAS_GEMM` router branch is removed for the same reason: a perf lever must not silently change selection numerics.

Perf: N = num_experts, tiny beside the expert GEMMs — and the recorded TTFT baselines (2026-08-04) were themselves set with the scalar router.

## Floor ratchet

Per owner instruction, and because the old floor let a 1.79-point deterministic regression through with **0.10** to spare:

| metric | was | now | best-ever |
|---|---|---|---|
| `overall_accuracy` | 84.66 | **86.50** | 86.55 |
| `normalized_single_turn_score` | 83.32 | **86.90** | 86.95 |

Formula is best-ever − 0.05; the gate compares `value + noise >= min` with `noise = 0.4`, so the effective floors are 86.10 / 86.50 — 0.45 under best-ever, outside the MTP band, so no false-fail on nondeterminism. Both metrics moved because the same regression moved both, and a floor watching only one leaves the other free to rot 3.6 points before tripping.

## Rejected suspects (documented so they are not re-investigated)

- `run_bf16_shared_expert` — inert on this leg: installed only under `ATLAS_FP8_DEQUANT_MOE_TO_BF16=1` or for Laguna; the `bf16head` recipe sets neither, so it returns `Ok(false)` with no side effects.
- Warp-row q/k RMS norm — loaded only when `ships_vanilla_norm_weights(config)`, which is deepseek_v4/laguna only. Handle is 0 for both Qwen3.6 legs.
- Tokenizer/template/thinking/sampling/scheduler — all new behavior flag-gated with history-preserving defaults; the 35B `MODEL.toml [behavior]` is unchanged in the range and BFCL pins temp 0.
- cuBLAS head-gate, dense-FFN, yarn-RoPE, K2/K3 `_t` branches, CUTLASS-grouped, E8M0, batched-prefill admission — each env-gated off, or scoped to a different model family or weight format.

## Verification status

**Not yet confirmed on hardware.** The claim is that this restores bit-exact router numerics and therefore lands back at exactly 86.95 / 86.55. No CPU test can observe that — a test asserting a kernel-path choice would assert nothing — so no test is added and the GPU A/B is the evidence. That run is next; this PR should not merge until its echolp record lands.

If it lands deterministic but elsewhere, there is a second bit-level change on this leg that inspection missed — and by the identical-scores property it would be in the batch4 stack, not Batch 5.

2100+ spark-model tests pass, fmt and clippy clean.

### DIFF
--- a/crates/spark-model/src/layers/moe/forward_prefill.rs
+++ b/crates/spark-model/src/layers/moe/forward_prefill.rs
@@ -203,29 +203,12 @@ impl MoeLayer {
                 h,
                 stream,
             )?;
-        } else if ctx.dispatch.cublas_gemm && n > 1 {
-            ops::cublas_bf16_proj_dense(
-                router_in,
-                self.weights.gate.weight,
-                gate_logits,
-                n,
-                num_experts,
-                h,
-                stream,
-            )?;
         } else {
-            ops::dense_gemm_prefill(
-                ctx.gpu,
-                self.dense_gemm,
-                self.dense_gemm_pipelined,
-                router_in,
-                &self.weights.gate,
-                gate_logits,
-                n,
-                num_experts,
-                h,
-                stream,
-            )?;
+            // Selection numerics — see router_gate_gemm_dense for why this
+            // must stay on the scalar kernel and why ATLAS_CUBLAS_GEMM must
+            // not reroute it either (2026-08-12 BFCL regression: a rerouted
+            // router GEMM flips top-k on borderline tokens deterministically).
+            self.router_gate_gemm_dense(router_in, gate_logits, n, num_experts, h, ctx, stream)?;
         }
         super::dump::dump_gate_logits(ctx.gpu, stream, gate_logits, n, num_experts)?;
         prof_step!("gate_gemm");
--- a/crates/spark-model/src/layers/moe/forward_prefill_bf16.rs
+++ b/crates/spark-model/src/layers/moe/forward_prefill_bf16.rs
@@ -67,18 +67,9 @@ impl MoeLayer {
                 stream,
             )?;
         } else {
-            ops::dense_gemm_prefill(
-                ctx.gpu,
-                self.dense_gemm,
-                self.dense_gemm_pipelined,
-                router_in,
-                &self.weights.gate,
-                gate_logits,
-                n,
-                num_experts,
-                h,
-                stream,
-            )?;
+            // Selection numerics — see router_gate_gemm_dense for why this
+            // must stay on the scalar kernel (2026-08-12 BFCL regression).
+            self.router_gate_gemm_dense(router_in, gate_logits, n, num_experts, h, ctx, stream)?;
         }
 
         super::dump::dump_gate_logits(ctx.gpu, stream, gate_logits, n, num_experts)?;
--- a/crates/spark-model/src/layers/moe/forward_prefill_fp8.rs
+++ b/crates/spark-model/src/layers/moe/forward_prefill_fp8.rs
@@ -252,18 +252,9 @@ impl MoeLayer {
                 stream,
             )?;
         } else {
-            ops::dense_gemm_prefill(
-                ctx.gpu,
-                self.dense_gemm,
-                self.dense_gemm_pipelined,
-                router_in,
-                &self.weights.gate,
-                gate_logits,
-                n,
-                num_experts,
-                h,
-                stream,
-            )?;
+            // Selection numerics — see router_gate_gemm_dense for why this
+            // must stay on the scalar kernel (2026-08-12 BFCL regression).
+            self.router_gate_gemm_dense(router_in, gate_logits, n, num_experts, h, ctx, stream)?;
         }
 
         super::dump::dump_gate_logits(ctx.gpu, stream, gate_logits, n, num_experts)?;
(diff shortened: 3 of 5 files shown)


tbraun96:
Closing: the pin is on `main` in a stronger form. `crates/spark-model/src/layers/moe/helpers_c.rs:255-266` (`router_gate_gemm_dense`) keeps the router-logits GEMM on scalar-order numerics via `dense_gemm_bf16_router`, verified bit-identical to `dense_gemm_bf16` and never rerouted to the reassociating fast GEMM — landed through #733 (`09ff2ceed2`, "Kernel perf sweep: 5 measured wins, stacked"). That closes #492 the same way this PR did.

The `BENCH.toml` half is stale: the `bfcl-subset-echolp` entry in `kernels/gb10/qwen3.6-35b-a3b/BENCH.toml` (line 126 on `9e597531`) has since been re-declared with its own floors and draw fingerprint, so this ratchet no longer applies. No branch deleted.

PR #497, opened 2026-08-13
kernels: cp.async prefetch for w4a16_gemv_batch2 (opt-in)

facepunch47:
## Summary

Close the leftover K=2 NVFP4 GEMV DRAM gap on GB10 by issuing packed-weight loads through a 2-stage `cp.async` smem pipeline. Graphs already ate launch tax; this is kernel bandwidth, not extra D2D/D2H.

`w4a16_gemv_batch2_cpasync` keeps the current two-phase coalesced K16 mapping and shares `w4a16_gemv_batchm_fma_chunk` with the template, so output is bit-identical to `w4a16_gemv_batch2` / 2× `w4a16_gemv`. The existing `w4a16_gemv_batch2` (`batchm_impl<2>`) stays as the fallback.

**GB10 oracle (2026-08-13): do not ship default-on.** Cold-DRAM cpasync is 5–16% slower than template batch2 (fail the 3% bar). Bit-identity is 100% on every 35B shape after the alignment gate. Default stays template batch2. Opt-in with `ATLAS_GEMV_BATCH2_CPASYNC=1` (`=0` does not enable).

Does not restore the old standalone batch2 body. Does not port SW GEMV to K=2. Does not use tile GEMM at M=2. `:8888` left on `atlas-gb10:pr-proof-321128d`. No image A/B (would lose tok/s). Combined 321128d+#497 stack was not pushed.

## GPU results (GB10)

Losslessness (`w4a16_batch2_cpasync_microtest`), after gating 16-byte `cp.async` on gmem alignment:

| Shape | bit_id vs batch2 | vs 2× `w4a16_gemv` |
|---|---|---|
| GDN in_proj 12288×2048 | 100% | 100% |
| attn Q 8192×2048 | 100% | 100% |
| GDN out_proj 2048×4096 | 100% | 100% |
| attn o_proj 2048×2048 | 100% | 100% |
| N%4≠0 tail 2050×2048 | 100% | 100% |
| K-tail 2048×2032 | 100% (was ~50% before alignment gate) | 100% |

Cold-DRAM (`w4a16_batch2_bw_microtest`, 273 GB/s peak, 40 iters, weight copies cycle past L2):

| Shape | M=1 GEMV | batch2 | cpasync | cpasync/batch2 |
|---|---|---|---|---|
| GDN in_proj 12288×2048 | 74.1 GB/s (191 µs) | **227.3 GB/s (62.3 µs)** | 195.4 GB/s (72.4 µs) | **1.16× slower** |
[...]

PR #498, opened 2026-08-13
kernels: default-on M=1 w4a16_gemv cp.async prefetch

facepunch47:
## Summary

Close the leftover M=1 NVFP4 GEMV DRAM gap on GB10 by issuing packed-weight loads through a 2-stage 8-byte `cp.async` smem pipeline. Same idea as #497, but for the 64-thread `w4a16_gemv` used on reject/serial fallback **and** the graphable MTP lm_head (that path stays on base GEMV, not SW).

`w4a16_gemv_cpasync` shares `w4a16_gemv_mac_chunk` / `w4a16_gemv_decode_scale` with `w4a16_gemv_partial`, so FMA order cannot drift vs current `w4a16_gemv` / `w4a16_gemv_sw`. Grid/block/signature match the 64-thread kernel (4 outs/block). **SW GEMV is not touched** and is not ported to K=2.

Default ON. Kill with `ATLAS_NO_GEMV_CPASYNC=1` (`=0` does **not** disable). Missing handle falls back to `w4a16_gemv`.

8-byte `cp.async` only (NVFP4 `K/2` is 8-byte aligned when `K` is a multiple of 16). This avoids #497's 16-byte K-tail gmem-alignment hole. The K-loop uses a uniform `super` trip count so `__syncwarp` / HIP `__syncthreads` cannot deadlock when lanes would otherwise take different iteration counts.

GPU losslessness + cold-DRAM oracles are **pending**: `gate-bake-485-rest` (`agentic-webserver`) currently holds the GPU. Do not ship as default-on until those pass. If cpasync is >3% slower than current M=1 `w4a16_gemv` on the 35B shapes, this flips to opt-in like #497 (`ATLAS_GEMV_CPASYNC=1`) and `:8888` stays on `atlas-gb10:pr-proof-321128d`.

Does not raise `--num-drafts`. Does not restore the old standalone batch2 body. Does not use tile GEMM at M=2. Does not enable grouped MoE at n=2.

## Test plan

- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy --workspace --tests`
- [x] Unit tests: default-on polarity (`unset`/`=0`/empty → ON; `=1` → kill), kernel exports `cp.async` + shared MAC helper + uniform `super` loop, init sites use `resolve_w4a16_gemv`
- [x] `nvcc --ptx -arch=sm_121a` of `kernels/gb10/common/w4a16_gemv.cu` (exports `w4a16_gemv_cpasync`)
- [ ] GPU losslessness oracle (`w4a16_gemv_cpasync_microtest`) — bit_id == 100% vs `w4a16_gemv` and vs `w4a16_gemv_sw` on GDN in_proj / attn Q / K-tail / N%4≠0
- [ ] GPU cold-DRAM oracle (`w4a16_gemv_cpasync_bw_microtest`) — fail if cpasync is >3% slower than current M=1 GEMV
- [ ] If oracles pass: cherry-pick onto local 321128d, bake `atlas-gb10:pr-proof-<shortsha>`, five temp-0 prompts vs BST 124.6 / hash 126.4 tok/s, leave winner on `:8888`

## Notes for reviewers

- Resolver sites: `dense_ffn`, `moe/init`, `mtp_head/new` (lm_head), `qwen3_attention/init`, `qwen3_ssm/init`, `nemotron_{mamba2,moe}`, `model/impl_a1`. Examples and `moe/mod_tests.rs` still look up the fallback by name.
- `strix` / `strix-hip` `w4a16_gemv.cu` are symlinks to the gb10 copy.
- Combined 321128d stack, if built, stays local/unpushed. Never docker-push `avarok/atlas-gb10:latest`.

## CLA

### DIFF
--- a/crates/spark-model/Cargo.toml
+++ b/crates/spark-model/Cargo.toml
@@ -240,3 +240,11 @@ required-features = ["cuda", "gpu-examples"]
 [[example]]
 name = "w4a16_qg_batch_bitparity_microtest"
 required-features = ["cuda", "gpu-examples"]
+
+[[example]]
+name = "w4a16_gemv_cpasync_microtest"
+required-features = ["cuda", "gpu-examples"]
+
+[[example]]
+name = "w4a16_gemv_cpasync_bw_microtest"
+required-features = ["cuda", "gpu-examples"]
--- a/crates/spark-model/examples/w4a16_gemv_cpasync_bw_microtest.rs
+++ b/crates/spark-model/examples/w4a16_gemv_cpasync_bw_microtest.rs
@@ -1,0 +1,205 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+//! Cold-DRAM bandwidth microtest: `w4a16_gemv` vs `w4a16_gemv_sw` vs
+//! `w4a16_gemv_cpasync` on Qwen3.6-35B M=1 shapes.
+//!
+//! Exit 0 if cpasync is not slower than base gemv by more than 3% on the
+//! production shapes. Informational GB/s is always printed.
+
+use anyhow::Result;
+use spark_runtime::cuda_backend::AtlasCudaBackend;
+use spark_runtime::gpu::{DevicePtr, GpuBackend, KernelHandle};
+use spark_runtime::kernel_args::{KernelLaunch, div_ceil};
+use std::time::Instant;
+
+const WARMUP: usize = 8;
+const ITERS: usize = 40;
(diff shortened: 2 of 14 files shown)


facepunch47:
Closing: measured on spark-9f9e 2026-08-15. Bit-exact vs `w4a16_gemv` / `w4a16_gemv_sw`. Fat GDN in_proj (12288×2048) is a tie (~193 vs ~189 µs). Not a win on the wall. Do not default-on.

PR #501, opened 2026-08-14
docs: TMA batch2 prefetch contract (opt-in, after oracle)

facepunch47:
## Summary

Design-only contract for a TMA (copy-engine) prefetch of the packed-weight stream in `w4a16_gemv_batch2`. **No kernel in this PR.** Implement only after:

1. The bandwidth oracle (`w4a16_batch2_bw_oracle`) is on the branch you measure.
2. Dual-issue (`w4a16_gemv_batch2_dualissue`) has a GB10 number. If dual-issue already wins ≥3% and is bit-exact, TMA is optional, not next.

`#497` used `cp.async` from the SM into shared memory and lost (195 vs 227 GB/s). TMA is a different truck: the SM writes a descriptor and keeps doing FMAs. It can still lose. The oracle is the judge.

Pairs with: the batch2 bandwidth-oracle and dual-issue PRs.

## Contract (when someone implements)

- New module `kernels/gb10/common/w4a16_gemv_batch2_tma.cu` (do not grow `w4a16_gemv.cu`).
- Same grid/block/args and FMA helper as template batch2 (bit-exact).
- TMA only packed weights. Scales and activations stay `ld.global`.
- Alignment: 16-byte TMA loads are illegal when `K/2 % 16 == 8` on odd `n` (the `#497` K-tail bug). Gate or pad; document which.
- Opt-in only: `ATLAS_GEMV_BATCH2_TMA=1`. Missing handle falls back. Do not default-on in the implementation PR either.

## Test plan

- [x] Docs-only; no runtime change
- [ ] CI: typos on the new markdown
- [ ] Implementation (later PR) must pass the oracle 3% bar and a bit-exact microtest before any default-on

## Notes for reviewers

- Skip implementing TMA if dual-issue already wins.
- Out of scope: turning on `#497`, changing echolp n=1004, SW GEMV at K=2, more drafts, wiring TMA into M=1 / batch4+ in the same change.

## CLA

### DIFF
--- a/docs/campaigns/gb10-batch2-rhythm-2026-08/tma.md
+++ b/docs/campaigns/gb10-batch2-rhythm-2026-08/tma.md
@@ -1,0 +1,67 @@
+# PR: TMA weight prefetch for `w4a16_gemv_batch2`
+
+Status: **design only**. Do not implement until:
+
+1. The bandwidth oracle (`w4a16_batch2_bw_oracle`) is on the branch you measure.
+2. Dual-issue (`w4a16_gemv_batch2_dualissue`) has a GB10 number. If dual-issue
+   already wins ≥3% and is bit-exact, TMA is optional, not next.
+
+## Why this is not `#497`
+
+`#497` used **`cp.async` from the SM** into shared memory. On K=2048 that
+mailbox cost more than the wait it hid (195 vs 227 GB/s).
+
+**TMA** is a different truck: a copy engine. The SM writes a descriptor
+(“this tile of packed weights”) and keeps doing FMAs. Issue-slot tax is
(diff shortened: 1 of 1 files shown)


tbraun96:
Hey @facepunch47, thanks for these great PRs. Could you please let me know on our Discord when you're ready to get these in? We also need to discuss this to make sure we're on the same page about what's going into the codebase (human-level review gate). We really appreciate your help and assistance in making the repository better!

facepunch47:
Closing: TMA contract was next only if dual-issue missed. Dual-issue is a measured tie. Not starting TMA. Docs-only, no kernel.

PR #502, opened 2026-08-14
feat(bench): MLPerf agentic-inference leg — scorer ported with executed-upstream parity fixtures, unrunnable until MLCommons publishes the dataset

tbraun96:
Adds the MLPerf datacenter agentic-inference benchmark as an Atlas leg, built as far as it can honestly be built today.

## Status: unrunnable, on purpose

**MLCommons has not published the dataset.** Verified at upstream `mlcommons/endpoints@7935df4`: `datasets/` is empty and the README says the data "can be downloaded from MLCommons storage (**link TBD**)". No URL, no licence, no auth story.

So this leg ships `status = "unmeasured"`, `NOT_REQUIRED`, **no metric floors**, and its verdict is always `info`. It is not a gate and must not be treated as one until a real measurement exists.

Explicitly **not** done, because each would have been a fabrication: no proxy dataset, no draw reconstructed from the upstream DeepSWE/Workato sources, no invented floors, and no 0.0 score when data is missing — `load()` refuses with the exact missing path, the "613 trajectories… NOT yet published" fact, and the verbatim upstream citation. A leg reporting 0.0 for lack of data is the vacuous-PASS class we just removed from the SSM gate.

## What is real and tested here

**The scorer, with parity proven by execution rather than assertion.** `assets/mlperf-agentic/gen_parity_fixtures.py` slices upstream's scorer class out of `scoring.py` verbatim, **executes it**, and emits 110 fixture cases — nothing hand-typed. Coverage: 15 intent, 5 ground-truth-intent, 18 bash, **58 alias cases (one per upstream table entry, plus set-equality so a table entry we invent is caught)**, 6 wrappers, 8 end-to-end turn scores, 8 domain ids.

Generation caught two behaviours predicted wrong *before* any Rust was written: `"reintent: I042"` and `"intent : I042"` still score (the bare-token fallback rescues what the explicit `\b` kills), and `python3.1.2.3` normalizes to nothing. Encoded ambiguities are tests, not comments: the bare fallback is case-**sensitive** while the explicit form is not; unknown executables vanish from **both** sides of the IoU; a lone `&` is not a separator. Two knowingly-accepted divergences from Python (`\s` matching `\x1c`–`\x1f`; backslash-newline in unterminated double quotes) are documented in the module header.

**Loader/provisioner** against the documented trajectory format — teacher-forced history, `tool_results` expansion, role grammar, contiguity — with streamed full-file SHA256 and a deterministic first-K-per-domain draw.

**Draw fingerprinting** (`c861672f3`, additive `dataset_fingerprint: Option<String>` on `BenchmarkResult` and `GateRecord`; serde default + skip, schema stays 1, round-trip test proves old records parse). This closes a real gap found while reading: Atlas computes a dataset digest during provisioning today **and throws it away** — no draw fingerprint is recorded anywhere, which is what makes "is this score from the same draw?" unanswerable after the fact. Isolated in its own first commit so it can be dropped independently.

## Facts worth knowing before anyone quotes a number from this

- **Reference model is `Qwen/Qwen3.6-35B-A3B` (BF16)**; we serve the official FP8 sibling. Closed division permits quantization (inference_rules §8.1), but validity depends on clearing all three thresholds — until then our numbers are not comparable to reference BF16.
- **Official scale is ~12–30h on one GB10** (the OSL band pins 10.8–13.2M output tokens over 613 trajectories). It cannot be a sub-2-hour merge gate unsubsetted; a ~2,000-turn subset projects to ~45–90min, but only a calibration run can size it honestly.
- **Temperature is immutable at 1.0** — this would be the repo's first stochastic gate, against a ±0.4-band culture. That is a decision, not a detail.
- MLPerf's 55.86% inline / 355–434 OSL / 67.5% SWE-bench appear in the BENCH note as **context, never as our floors**.
- The **SWE-bench Verified leg** (Docker + 200 live mini-swe-agent rollouts) is a workflow, not a leg — without it no run is a valid official submission.

## Two traps that could not be closed in code

1. Drop *any* file at `~/.atlas/artifacts/mlperf-agentic/dataset.jsonl` and the leg will run it. The unpinned-SHA warning and the refusal doctrine are the only guards until a real pin exists.
2. Promoting to REQUIRED later needs a decided verdict rule: `check.rs:427` demands a PASS verdict and this leg deliberately never emits one. Flagged in `report.rs` for whoever owns calibration.

575 atlas-plugin tests pass; fmt and clippy clean; `spark-server` still compiles.

### DIFF
--- a/.benchmarks/agentic-webserver/2026-08-19-80c9eb5069.json
+++ b/.benchmarks/agentic-webserver/2026-08-19-80c9eb5069.json
@@ -1,0 +1,467 @@
+{
+  "schema": 1,
+  "benchmark_id": "agentic-webserver",
+  "benchmark_name": "Agentic Webserver Test",
+  "git_sha": "80c9eb5069",
+  "recorded_at": 1787146299,
+  "target_model": "Qwen/Qwen3.6-35B-A3B-FP8",
+  "params": {
+    "build_timeout_s": "600",
+    "command_timeout_s": "180",
+    "iterations": "10",
+    "max_tokens": "8192",
+    "max_turns": "40",
+    "request_timeout_s": "900",
+    "s_per_turn_budget": "8.5",
+    "serve_timeout_s": "30",
+    "wall_budget_s": "1800"
+  },
+  "command": [
+    "spark",
+    "benchmark",
+    "run",
+    "agentic-webserver",
+    "--param",
+    "build_timeout_s=600",
+    "--param",
+    "command_timeout_s=180",
+    "--param",
+    "iterations=10",
+    "--param",
+    "max_tokens=8192",
+    "--param",
+    "max_turns=40",
+    "--param",
+    "request_timeout_s=900",
+    "--param",
+    "s_per_turn_budget=8.5",
+    "--param",
+    "serve_timeout_s=30",
+    "--param",
+    "wall_budget_s=1800",
+    "--yes",
+    "--pull-request-gate"
+  ],
+  "served_by": "qwen3.6/qwen3.6-35b-a3b-fp8-bf16head",
+  "atlas_version": "1.0.0-beta-preview",
+  "hardware": {
+    "gpu": "NVIDIA GB10",
+    "driver": "580.126.09",
+    "sm_clock_mhz": 2392.0,
+    "source": "nvidia-smi"
+  },
+  "hardware_state": {
+    "sensitivity": "speed",
+    "before": {
+      "captured_at": 1787145520,
+      "machine": {
+        "hostname": "spark-256a",
+        "machine_id": "7af66f30966a49b6886e00e2fce4b42f",
+        "gpu": "NVIDIA GB10",
+        "driver": "580.126.09"
+      },
+      "gpu_temp_c": 54.0,
+      "chassis_temps_c": [
+        {
+          "name": "acpitz",
+          "temp_c": 64.7
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 56.1
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 64.7
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 55.5
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 55.1
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 57.6
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 57.4
+        }
+      ],
+      "sm_clock_mhz": 2405.0,
+      "sm_clock_max_mhz": 3003.0,
+      "throttle_counters": {
+        "sw_power_cap_us": 40384664375,
+        "sw_thermal_us": 0,
+        "hw_thermal_us": 0,
+        "hw_power_brake_us": 0,
+        "sync_boost_us": 0
+      },
+      "throttle_active": {
+        "sw_power_cap": false,
+        "sw_thermal": false,
+        "hw_thermal": false,
+        "hw_power_brake": false
+      },
+      "mem_available_kb": 7112436,
+      "mem_total_kb": 127601452,
+      "page_cache_kb": 4564796,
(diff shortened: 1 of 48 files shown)


tbraun96:
Pushed `c2b825d06`: the Benchmarks suite list now scrolls.

**Why this is in a benchmark-leg PR.** Registering one leg pushed `Serve Matrix` off the list at 160×48 — 11 entries × 4 rows = 44 into a 40-row column — and the list had no scrolling and no clipped-position affordance. That is the same defect class the dashboard batch just fixed everywhere else (forms windowing, History paging, Ops scrollback, help-overlay clipping): content below the fold with no way to reach it.

The failing test `the_suite_list_shows_the_benchmarks_and_their_provenance` was kept **byte-for-byte** rather than relaxed. Raising the render size or asserting a subset would have traded a visible failure for an invisible benchmark, and every future leg would have inherited it.

Three changes, all in existing idioms: adaptive density (drop the separator row before clipping, never the summary or duration), clamped scrolling with the renderer publishing its viewport through `suite_page` (the `log_scroll_max` contract, kept on `BenchState` because `bench_keys` cannot see `App`), and a `─ 1-6 of 11 ─` bottom-border indicator shown **only** when clipped, in plain dim text so `NO_COLOR` is unaffected. Keys `g`/`G`/`Home`/`End`/`PgUp`/`PgDn`, clamped at both ends, named in the footer.

Also reported and deliberately **not** fixed here: `draw_detail` wraps into a `Constraint::Min(8)` with no scroll or indicator, and the PLUGIN provenance block is a fixed `Length(11)` — both clip silently with plugin-authored content, same class. They deserve their own change rather than scope-creeping this one.

4652 workspace tests pass; fmt and clippy clean.

tbraun96:
### Rebased onto current `main` — four conflicts, each a judgement rather than a mechanical merge

This was `DIRTY` against a main that has moved a long way (#572, #569, #581 all landed since). Resolutions:

| file | conflict | resolution |
|---|---|---|
| `atlas-plugin/src/result.rs` | both sides add a struct field — `hardware_state` (#569) vs this branch's `dataset_fingerprint` | **kept both** (additive, 2 hunks) |
| `benchmarks/mod.rs`, `registry.rs` | main gained modules this branch predates | **union**, alphabetised |
| `mlperf_agentic/descriptors.rs` | descriptor predates `sensitivity` and `threshold_params` | see below |
| `tui/render/hints.rs` | branch replaces the `List` hint with its PgUp/PgDn feature; main added a `Variants` arm the branch predates | **branch's List arms + main's Variants arm** |

**On the descriptor fields**, since these are judgement calls a reviewer should check:

- `sensitivity: Sensitivity::Correctness` — same reasoning the BFCL legs give. This leg scores intent-code match and bash-executable IoU and reports OSL, not a rate; there is no timing number for a thermal or clock excursion to reach, so gating it on machine state would stop accuracy work for a reason that cannot affect the result.
- `threshold_params: &[]` — the descriptor's own detail says "No baseline exists; the first measured run on main becomes one." A threshold param with nothing to substitute from would pin a guess.

### Plus a real test rot fix (separate commit)

Two TUI tests hardcoded `160x48` as "large enough to hold the whole suite". The registry has grown past that — main added decode-floor / quick-speed / media-integrity, and this branch adds `mlperf-agentic-subset` — so a registry change with nothing to do with the renderer started failing renderer tests.

`tall_enough()` now derives the height, and it had to be **measured**: a list entry is not one row. 31 rows gives a 7-entry page, 45 rows gives 12 — ~2.8 rows apiece. My first two attempts (`n + 16`, `n + 30`) still failed for exactly that reason; `3n + 20` scales with the registry instead of clearing a fixed chrome. Both tests share the one helper so the next registration can't break one and leave the other green.

**Verification:** workspace clippy, fmt and **5316 tests** green.

Still needs its benchmark-record campaign at this sha — `dgx2` is down (three powercycles today; see the vLLM C≥32 hazard in #600), so that is pending hardware.

PR #503, opened 2026-08-14
ci: pin actions to full-length SHAs so jobs can start again

AzeezIsh:
## What this fixes

Every job on a PR based on `strix/consolidate-tmp` currently fails in about
two seconds, before a single step runs:

> The actions `actions/checkout@v7`, `dtolnay/rust-toolchain@stable`, and
> `swatinem/rust-cache@v2` are not allowed in `Avarok-Cybersecurity/atlas`
> because all actions must be pinned to a full-length commit SHA.

The org enabled that policy after this branch forked. `main` was updated;
`consolidate-tmp` was not. The result is a full red checks table on branches
whose code is fine — see PR #470, where all nine checks fail at **Set up job**
and none of them reached a compiler.

This blocks every PR stacked on this base, not only the Windows ones.

## What it does

Rewrites 43 `uses:` lines to full-length commit SHAs, keeping the original ref
as a trailing comment (`# v7`), so each pin is still readable as a version.
43 insertions, 43 deletions, nothing but `.github/workflows/`.

Every SHA is **the one `main` already uses**, so the branches converge rather
than growing a second set of pins. Three actions do not appear on `main` and
were resolved from their own tags:

| action | SHA |
|---|---|
| `actions/download-artifact@v4` | `d3f86a106a0bac45b974a628896c90dbdf5c8093` |
| `actions/upload-artifact@v4` | `ea165f8d65b6e75b540449e92b4886f43607fa02` |
| `apache/skywalking-eyes@v0.8.0` | `61275cc80d0798a405cb070f7d3a8aaf7cf2c2c1` |

Those stay on **v4 deliberately**. `main` has since moved to
`upload-artifact@v7` / `download-artifact@v8`, whose major bumps changed
artifact layout and overwrite behaviour. Adopting them here would smuggle a
behavioural change into a commit whose only job is to make the runner accept
the workflow at all. The eventual merge into `main` resolves the version
difference by itself.

## How to verify

The check is mechanical — no `uses:` line should name anything but a local
workflow or a 40-hex SHA:

```bash
grep -rhoP 'uses:\s*\K[^\s]+' .github --include='*.yml' --include='*.yaml' \
  | grep -v '^\./' | grep -vP '@[0-9a-f]{40}$'
```

Empty output means every action is pinned. The real proof is this PR's own
checks table: jobs that reach a compiler instead of dying at **Set up job**.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,10 +67,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
-      - uses: apache/skywalking-eyes/header@61275cc80d0798a405cb070f7d3a8aaf7cf2c2c1 # v0.8.0
-        with:
-          config: .licenserc.yaml
-          mode: check
+      # Was `apache/skywalking-eyes/header`. That action is a composite that
+      # internally uses an UNPINNED `actions/setup-go@v5`, and this repo's
+      # SHA-pinning requirement applies TRANSITIVELY — so it cannot run here
+      # however carefully we pin the action itself. Checking a header is a
+      # grep; it does not justify a third-party action and a Go toolchain in
+      # CI. `.licenserc.yaml` is still the only place the paths and ignores
+      # live. Same replacement main already made; the script is taken from
+      # there unmodified so the two branches converge.
+      - run: pip install --quiet pyyaml
+      - run: python3 scripts/check_spdx.py
 
   typos:
     name: typos
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,8 +35,8 @@ jobs:
     name: cargo fmt --check
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v7
-      - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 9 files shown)


tbraun96:
Closing as **already landed** — not as rejected. The work in this PR is good and it is on `main`; it arrived by another route.

Checked rather than assumed:

| claim in this PR | state of `main` today |
|---|---|
| pin actions to full-length SHAs | **101 of 101 pinned, 0 unpinned** across 24 workflows |
| replace `apache/skywalking-eyes` with a local checker | `scripts/check_spdx.py` exists and runs in the `SPDX license headers` job |
| reason: its composite uses an unpinned `actions/setup-go@v5` | that exact reasoning is the comment now sitting above the step in `ci.yml` |

The diff here still shows `scripts/check_spdx.py` as `@@ -0,0 +1,63 @@` — a *new* file — because the branch's base is `ea36e0c9a7` and has not been touched since 2026-08-14. Merging it now would re-add what is already there and, worse, revert three weeks of subsequent changes to those eight workflow files.

I also extended its reach earlier today: `.licenserc.yaml` now covers `.github/scripts/*.{sh,py}`, so the checker this PR introduced gates six more files than it did.

If any of the eight workflow files here contains a change that did **not** make it to main, say so and I will lift that hunk onto a fresh branch — reopening this one against a three-week-old base is the part that does not work.

PR #504, opened 2026-08-14
ci: publish rolling dev releases from main so the binaries are downloadable

AzeezIsh:
## The gap

The release matrix has been building `spark` for **nine** (OS, arch, GPU)
combinations for months. **Not one of them was ever downloadable.**

- `release.yml` publishes only on `workflow_dispatch`, and no dispatch has ever
  run with `dry_run=false` — `gh release list` is empty. Zero Releases, ever.
- Everything a PR builds goes to an Actions artifact with **1-day retention**
  (`ci.yml` passes `retention_days: 1`), which also cannot be downloaded
  without a GitHub login.
- The matrix is `if: github.event_name == 'pull_request'`, so pushes to `main`
  build nothing at all.

Net effect: "OK, how do I test it?" had no good answer for anyone without a
full CUDA/HIP toolchain and half an hour — even though a binary for their exact
platform had already been built, and thrown away.

## What this does

Adds `dev-release.yml`. Modelled on llama.cpp, which publishes a numbered build
off `master` (`b10423`) many times a day, one archive per backend. Same shape:

- tag is **`b<commit-count>`** (current `main` would publish `b218`)
- assets are the **same packages `release-build.yml` already produces** — if
  these ever differ from what a PR validated, that is a bug
- one archive per platform, each with its `.sha256`

`release.yml` is untouched. Every release here is a **prerelease**, so it never
takes the "Latest" slot; there is no `Cargo.toml` bump and no commit pushed back
to `main`; and `bNNNN` cannot collide with `vX.Y.Z`.

### Why `workflow_run` and not `push`

CI already runs on every push to `main`. A `push` trigger here would either
duplicate the whole test suite, or skip the gate and publish binaries from a
`main` that does not pass its own checks. Chaining off CI's verdict costs
nothing and gates properly.

### This does not weaken the human gate

The `/atlas-release` skill's rule — *never advance the shippable surface without
a human gate* — is about the moving `avarok/atlas-gb10` tags, and it still
holds: nothing here touches an image or a `:latest`. The shippable surface stays
`release.yml`, human-dispatched. This adds a **prerelease** channel below it,
and the release notes say in as many words that CI-green is not tested:

> **Not a stable release, and not a tested one.** Every GPU target in the matrix
> is marked experimental and is *built* on a hosted runner that has no GPU of
> that vendor. These binaries compile and link; whether they run on your
> hardware is exactly what has not been checked here.

## Verification

The decide step was run against this repo for **all five paths it can take**:

| case | result |
|---|---|
| CI success, normal commit | publishes `b218` |
| CI failed | skips |
| CI cancelled | skips |
| manual dispatch (no upstream CI run) | publishes |
| `[no release]` in commit message | skips |

It also skips when the tag already exists, so re-running CI on an
already-published commit stays green instead of failing on a duplicate tag.
All 18 workflows still parse.

## Also: the `windows-x86_64-amd-hip` note had drifted

Correcting it here because these releases will hand people that zip, and the
note is where the caveats live.

- it cited `crates/atlas-kernels/hip/win_link_stubs.cpp` — **a file that no
  longer exists**
- it called the CUDA link symbols "stub import libs". They are **real
  HIP-backed shims** (`libcuda_hip_shim.cpp`, `libcudart_hip_shim.cpp`).
  cuBLASLt alone is a genuine stub, deliberately: it fails its create call,
  which is safe because the default GEMM path is Atlas's own and cuBLASLt is
  opt-in behind `ATLAS_CUBLAS_GEMM=1`
- "COMPILE-ONLY" read as a property of the binary rather than of CI. It is CI
  that never executes it, for the plain reason that hosted runners have no AMD
  GPU

The note now also records what these releases **cannot** tell you: running that
target on real gfx1151 needs the Windows runtime fixes in #470, which are not
merged here. Without them serve dies at KV sizing, because `hipMemGetInfo`
reports no free memory on Windows. Publishing a binary is not the same as
vouching for it.

## One judgement call for reviewers

This publishes a prerelease on **every** green `main`. llama.cpp does ~10/day
and does not prune. If that is too noisy, the cheapest knob is a `paths-ignore`
filter on the trigger, or moving to a daily schedule — say the word and I will
change it rather than guess.


AzeezIsh:
### On the red CodeQL check

Two high-severity `actions/cache-poisoning` alerts, both pointing at
`release-build.yml`, not at the new workflow. Here is what I found and what I
did, because the remaining decision is a maintainer's.

**It is not new, and it is not really about this PR.** `main`'s CodeQL default
setup scans `/language:python` and `/language:rust` only. Workflows are analysed
on **pull requests**, so this finding is invisible on `main` and surfaces the
moment anyone edits `release-build.yml`. It has been latent for as long as
`release.yml` has taken an arbitrary `ref` as a `workflow_dispatch` input and
passed it to a job that runs `Swatinem/rust-cache` in the default-branch
context.

**What I fixed:**

1. **The part I introduced.** `dev-release.yml` originally had its own `ref`
   dispatch input, adding a second arbitrary-ref path into that checkout. Gone —
   the manual hatch now takes no inputs and only ever builds `main` (5f17564).
2. **The pre-existing part.** `save-if: ${{ github.event_name != 'workflow_dispatch' }}`
   on the cache step, so dispatch runs *read* the cache but can no longer
   *write* it. `pull_request` and `workflow_run` are unaffected and still save,
   so no PR loses cache warmth; the cost lands only on the rare manual release
   (f81ee1d).

**Why it is still red.** The query matches the *shape* — a privileged checkout
of a caller-supplied ref in a job that uses a cache action — and does not track
`save-if`. Clearing it fully means either dropping `rust-cache` from
`release-build.yml` (a real, unmeasured build-time cost across nine targets) or
dismissing the alert with a written rationale. Both are calls I would rather you
make than make for you, so I stopped here rather than guessing.

I would lean toward **dismissing with a rationale** now that the write path is
closed: the exploit requires dispatch permission, and post-fix that permission
no longer yields a persistent foothold in the shared cache.

tbraun96:
Closing: this landed over a year of commits ago, inside the stack that names it.

**Carrier:** `04726b148` — *"Concurrency cycle: **#504**+#505+#521+#522+#523 stack, #512 re-port, decode-floor gate, honest concurrency instrument, accept-stats, per-model bfcl floors (#525)"*. #525 merged; this PR was squashed in with it and never auto-closed, which is the known behaviour when a stack lands by squash.

**Evidence, file by file:**

| file | status on `main` |
|---|---|
| `.github/workflows/dev-release.yml` | **byte-identical** — 222 lines both sides, `diff` reports 0 differing lines |
| `.github/workflows/release-build.yml` | every added line present, including the `publishing:` input and the whole cache-poisoning rationale block |
| `.github/release-matrix.json` | this PR's one-line change uses the **pre-rebrand** names (`atlas-kernels`, `ATLAS_CUBLAS_GEMM`); `main` carries the post-#1101 equivalents (`avarok-kernels`, `AVAROK_CUBLAS_GEMM`) |

So the only thing this PR would still change is to *revert* two identifiers to names that no longer exist. Its three merge conflicts are all of that shape and all resolve to "keep `main`".

Worth noting for anyone auditing the backlog: **#521 and #523 from that same stack are also still open.** I sampled their content and it is only partially present on `main` (3/10 and 6/10 added lines), so unlike this one they are *not* conclusively superseded and I have left them alone — they need a proper delta before anyone closes them.

Nothing is lost by closing; reopening restores it if this reading is wrong.

PR #505, opened 2026-08-14
build: make the kernel build say which hardware it picked

AzeezIsh:
## The problem

`ATLAS_TARGET_HW` chooses which `kernels/<hw>/` tree gets compiled, and it
defaults to `gb10` — an NVIDIA target. **Nothing in the build says so.**

Someone on an AMD box who runs the documented

```bash
cargo build --release -p spark-server
```

builds NVIDIA kernels, and finds out somewhere deep in nvcc, in an error that
never mentions the variable that actually chose their hardware. The required
`ATLAS_TARGET_HW` / `MODEL` / `QUANT` triple is currently written down only in
workflow comments and porting docs — the build itself has never mentioned it.

The `/atlas-release` skill already names this exact footgun: *"no stage runs on
an implicit default… a bare `cargo build` is a bug, not a shortcut."* That rule
was unenforceable from inside the build, because the build could not report
what it had chosen.

## Before / after

Unknown hardware target — before:

```
Hardware kernel directory not found: /repo/kernels/rocm
```

After:

```
ATLAS_TARGET_HW=rocm selects kernels/rocm/, which does not exist.

Available hardware targets: gb10 (nvidia), metal (apple), strix (amd), strix-hip (hip)

Build a target that exists:
   ATLAS_TARGET_HW=gb10 ATLAS_TARGET_MODEL=deepseek-v4-flash ATLAS_TARGET_QUANT=nvfp4 cargo build --release -p spark-server
   ATLAS_TARGET_HW=metal ATLAS_TARGET_MODEL=nllb-200-3.3b ATLAS_TARGET_QUANT=bf16 cargo build --release -p spark-server
   ATLAS_TARGET_HW=strix ATLAS_TARGET_MODEL=qwen3.6-27b ATLAS_TARGET_QUANT=nvfp4 cargo build --release -p spark-server
   ATLAS_TARGET_HW=strix-hip ATLAS_TARGET_MODEL=qwen3.6-27b ATLAS_TARGET_QUANT=nvfp4 cargo build --release -p spark-server
```

Every line is generated from real directory names, so anything listed is known
to resolve. **All** targets are listed rather than one being picked, because
picking meant teaching an AMD user the Metal incantation on the grounds that
`metal` sorts before `strix`.

Same treatment for an unknown model (lists the models that do exist), and for a
selection that resolves to no targets at all — which previously said only
`Check ATLAS_TARGET_* env vars`.

When any of the triple is implicit, the build now names the whole effective
selection instead of choosing in silence:

```
cargo:warning=Building kernels for gb10 (nvidia), model=*, quant=nvfp4
- ATLAS_TARGET_HW, ATLAS_TARGET_MODEL and ATLAS_TARGET_QUANT not set.
Hardware targets available: ... Set all three explicitly to choose.
```

A **warning, not an error**: requiring the variables would break every existing
NVIDIA build, and the problem was never the default — it is that the default was
invisible. It fires *after* the target is validated, so a rejected target is not
also announced as the one being built.

## Notes for review

- New code is in `build_diagnose.rs`, following the existing
  `#[path = "build_*.rs"] mod` idiom, rather than growing `build.rs` — that file
  is 1204 lines and only clears the 500-LoC cap by allow-list.
- These functions run on the failure path and **inside panic messages**, so they
  never panic themselves. `HARDWARE.toml` is scanned by hand for `vendor` rather
  than TOML-parsed, precisely because some of these messages are reporting a
  malformed one; unreadable directories degrade to a less specific message
  instead of replacing the user's real error with ours.
- The `gb10` default now has one definition (`build_diagnose::DEFAULT_HW`)
  instead of being spelled out at two call sites.

## Verification

Run against the real `kernels/` tree, through real cargo, on the pinned 1.93.1:

- `ATLAS_TARGET_HW=rocm cargo check -p atlas-kernels` → the message above,
  exit 101, and **no** spurious "building rocm" warning
- `ATLAS_SKIP_BUILD=1 cargo check -p atlas-kernels` → clean, exit 0
- `cargo clippy -p atlas-kernels` → clean
- `cargo fmt -p atlas-kernels -- --check` → clean

Each message was also rendered standalone against the real directory tree, so
the model/quant values shown above are the ones the code actually produces, not
hand-written examples.

### DIFF
--- a/crates/atlas-kernels/build.rs
+++ b/crates/atlas-kernels/build.rs
@@ -181,7 +181,13 @@ fn main() {
 
     assert!(
         !targets.is_empty(),
-        "No kernel targets resolved. Check ATLAS_TARGET_* env vars."
+        "{}",
+        build_diagnose::no_targets(
+            &workspace_root.join("kernels"),
+            &env::var("ATLAS_TARGET_HW").unwrap_or_else(|_| build_diagnose::DEFAULT_HW.into()),
+            &env::var("ATLAS_TARGET_MODEL").unwrap_or_else(|_| "*".into()),
+            &env::var("ATLAS_TARGET_QUANT").unwrap_or_else(|_| "nvfp4".into()),
+        )
     );
 
     // ── Resolve compute target (compiler) from HARDWARE.toml vendor ──
@@ -189,7 +195,7 @@ fn main() {
     // Apple (xcrun→metallib), Intel (icpx→SPIR-V). Only NVIDIA is implemented.
     let hw_dir = workspace_root
         .join("kernels")
-        .join(env::var("ATLAS_TARGET_HW").unwrap_or_else(|_| "gb10".into()));
+        .join(env::var("ATLAS_TARGET_HW").unwrap_or_else(|_| build_diagnose::DEFAULT_HW.into()));
     let hw_toml_path = hw_dir.join("HARDWARE.toml");
     let hw_toml: toml::Value = {
         let content = std::fs::read_to_string(&hw_toml_path)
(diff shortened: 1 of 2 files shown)


tbraun96:

Cherry-picking this onto current `main` produces an **empty commit**: the tree already contains it. Both files it touches are present — `crates/atlas-kernels/build.rs` and `crates/atlas-kernels/build_diagnose.rs`, the latter being the diagnose surface this PR introduced.

Found while grouping the Windows/build PRs (#470, #505, #507) into one stack so a single certification campaign would cover them. Two of the three turned out not to need one at all.

PR #508, opened 2026-08-14
fix(moe): own the CUTLASS grouped host tables per layer — a pointer-keyed global cache survives model teardown

tbraun96:
The batch4 stack memoized the per-expert pointer/scale tables in a process-global cache keyed on the **device pointer**:

```rust
static CACHE: OnceLock<Mutex<HashMap<u64, Vec<u64>>>> = OnceLock::new();
if let Some(v) = c.lock().unwrap().get(&p.0) { return Ok(v.clone()); }
```

No invalidation, no bound, and `n` absent from the key. **An address is not an identity.**

## The failure

`model_swap::swap` tears the outgoing model down in-process, then loads the new one. Teardown releases the weight store and `sweep_unreleased()`s the ledger — every one of the nine cached arrays per MoE layer is freed there. The static survives. `alloc`/`free` are raw `cuMemAlloc_v2`/`cuMemFree_v2`, so the driver reuses freed VA ranges, and two loads of the same recipe are near-identical allocation sequences.

Result: `cached_u64` hands the **new** model the **old** model's expert weight pointers, and the grouped GEMM reads freed-and-reused memory as weights. Silent garbage, no error — and a same-architecture swap passes every length check in `ensure_group_arrays`, so nothing catches it. This is the failure class the laguna loader guard documents in prose at `weight_loader/laguna/load_layers.rs:174-188`.

Separately, the key ignores `n`, so a hit returns the first query's length.

Within one model's lifetime the cache is provably benign — the nine arrays are built at load and neither freed nor rewritten until teardown. The bug is lifecycle, not locking; the mutex was race-free.

## The fix: ownership instead of keying

`build_cutlass_grouped_sfb` already holds the SFB pointer values on the host when it builds them, so it snapshots all nine tables once at load into a `MoeCutlassHostTables` **owned by the `MoeLayer`**. The snapshot dies with the model, making staleness structurally impossible — no epochs, no globals, no locks.

Rejected alternative: `(ptr, len, generation)` keys. That keeps global mutable state alive and needs teardown hooks to bump the generation correctly; ownership needs no bookkeeping to be right.

The batch4 perf win is kept and slightly improved — the device-side SFB pointer-array uploads are deleted (their only reader was the cache's readback), so static tables now cost **zero** per-prefill D2H, with only the genuinely per-call `expert_offsets` copy remaining. That was the point of the original change: 6 copies × 2 calls × 47 layers = 564 pointless transfers per prefill.

## Tests

Three mock-backend regression tests that **fail against the old cache**: a re-read at a reused address must see current contents; `n` is honored; field-to-source mapping is correct across nine same-typed tables.

`cargo test --workspace` 4612 passed / 0 failed; clippy and fmt clean.

## Provenance

Found while hunting a BFCL residual that appears only with speculative decoding enabled. **It is not that residual** — the cache is on the prefill path, shared by spec-on and spec-off, and a 1004-sample token-level replay showed the fixed build byte-identical to last-good with speculation off. The MTP head never reaches this code at all. It is submitted on its own merits: a cache that can serve another allocation's data is a defect regardless of which investigation noticed it.

To reproduce the actual bug: serve with `ATLAS_HOLO_MOE_GROUPED_CUTLASS=1`, prompt, swap models via the dashboard, prompt again — main produces garbage after the swap; this branch stays correct.


tbraun96:
Closing as **superseded by #513** — the fix is already in that branch, not merely equivalent to it.

I diffed `crates/spark-model/src/layers/ops/moe_grouped_a2.rs` and the four `moe/` files between this branch and `feat/qwen3.8-27b-support`: the content is **byte-identical** (same patch, different commit). So merging #513 lands this change; keeping #508 open would mean gating the same diff twice.

The fix was independently reviewed as part of that audit and holds up:

- Both `static OnceLock<Mutex<HashMap<u64, …>>>` caches are gone. Zero `OnceLock<Mutex<HashMap` remain anywhere under `crates/` except the historical doc comment in `derived_weights.rs` explaining why the pattern was abandoned.
- `MoeCutlassHostTables` is owned by `MoeLayer` and snapshotted once at load, so it dies with the layer — staleness after a model swap is structurally impossible rather than merely unlikely.
- It also closes two latent defects of the old cache that weren't in the original report: hits ignored the requested length `n`, and the device `[num_experts]` u64 SFB pointer arrays are no longer uploaded at all (their only reader was the cache's readback), so even first-prefill misses disappear. The batch4 win — 564 avoided D2H per prefill — is preserved.
- Gating equivalence checked: `gate_sfb_cutlass.is_some() && up_sfb_cutlass.is_some()` → `cutlass_grouped_host.is_some()`; `down_sfb_cutlass.is_some()` → `host.down.as_ref()`.

Nothing here is being dropped — reopen if the #513 merge somehow doesn't carry it.

PR #511, opened 2026-08-14
sampling: let a model declare its own min_p

rsafier:
A model card that specifies `min_p = 0` had no way to say so.

The server ships `--default-min-p 0.08` and every request that does not name min_p takes it. `[behavior].min_p_floor` only ever *raises* min_p (`min_p.max(floor)`), so it cannot pin one to zero. And the `[sampling.*]` preset did not carry the field at all — `resolve_sampling_defaults` said so in its own doc comment:

> an ABSENT field backfills from the model's curated MODEL.toml `[sampling.non_thinking]` preset (temperature/top_k/top_p) or the CLI defaults (top_n_sigma/min_p, **which the preset does not carry**).

So for this one field the value reaching the sampler was the operator's, not the model's, and no MODEL.toml edit could change that.

## Why it matters

Not hypothetical. **min_p is live** — measured on Nemotron-3.5-Lightning-30B at temperature 1.6, `min_p = 0.95` collapsed 8 samples to a single distinct output against 8/8 distinct at `0.0`. And two checkpoints in flight publish `min_p = 0` on their cards: Qwen3.8-27B (thinking *and* instruct modes) and Lightning. Both would silently serve at 0.08.

## Shape

`SamplingCategory.min_p` is `Option<f32>`, not `f32`, because **absence and `0.0` mean opposite things here**. A plain `f32` defaulting to 0.0 would silently strip the 0.08 floor from every model that has a `[sampling.*]` table — the opposite regression, and a quiet one.

Precedence becomes the same three-tier rule temperature/top_k/top_p already follow:

```
generation_config.json  →  MODEL.toml preset (Some)  →  --default-min-p
```

`None` leaves the CLI default owning the field exactly as before, so **no existing model changes behaviour**.

## `top_n_sigma` deliberately left alone

The same doc comment groups the two, but they are not in the same position. No model card we serve carries `top_n_sigma`, and it is separately suspected of being a **no-op**: on Lightning, sigma `0.0` / `0.2` / `0.5` all produced 8/8 distinct outputs at temperature 1.6, where a restrictive `0.2` should have collapsed diversity the way `min_p = 0.95` did. Advertising a MODEL.toml knob that does nothing would be worse than leaving it CLI-only. Worth chasing separately.

## Verification

Three unit tests on `resolve_sampling_defaults`, plus two checks that the tests themselves are worth something:

- **Negative control** — deleting only the `.or(preset.min_p)` line while keeping the field fails the new test with `left: 0.08 / right: 0.0`. Restoring it passes. So the test is not vacuous.
- **End to end** — injecting `min_p = 0.0` into a real MODEL.toml `[sampling.non_thinking]` and rebuilding emits exactly **one** `min_p: Some(0.0000)` in the generated target table, with every other target still `min_p: None`. Parser → codegen → runtime all carry it, and absence stays absence.

`cargo check --workspace --tests` and `clippy` clean, **4619 tests pass**, all five touched files under the 500-LoC cap (`lib.rs` 486, `runtime.rs` 483).

**No MODEL.toml is changed here.** This adds the capability; the models that need it declare it in their own PRs.

## CLA

### DIFF
--- a/crates/atlas-kernels/build.rs
+++ b/crates/atlas-kernels/build.rs
@@ -23,6 +23,9 @@ struct SamplingCat {
     // over the recent token window. 0.0 = disabled. 0.2 is the SGLang
     // reference value; lossless on AIME/GPQA at that strength.
     lz_penalty: f32,
+    // Model-declared min-p. None = MODEL.toml is silent, so the server's
+    // --default-min-p stands (see SamplingCategory in src/lib.rs).
+    min_p: Option<f32>,
 }
 
 impl Default for SamplingCat {
@@ -38,6 +41,7 @@ impl Default for SamplingCat {
             dry_base: 1.75,
             dry_allowed_length: 2,
             lz_penalty: 0.0,
+            min_p: None,
         }
     }
 }
--- a/crates/atlas-kernels/build_codegen.rs
+++ b/crates/atlas-kernels/build_codegen.rs
@@ -131,7 +131,7 @@ pub(super) fn generate_target_ptx_rs(
         };
         let fmt_cat = |c: &SamplingCat| -> String {
             format!(
-                "SamplingCategory {{ temperature: {:.2}, top_p: {:.2}, top_k: {}, presence_penalty: {:.2}, frequency_penalty: {:.2}, repetition_penalty: {:.2}, dry_multiplier: {:.2}, dry_base: {:.2}, dry_allowed_length: {}, lz_penalty: {:.2} }}",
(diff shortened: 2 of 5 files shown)


tbraun96:
Closing as SUPERSEDED by main (da56736dc): the model-declared min_p feature landed via the #519 stack. This is the one closure where byte-identity holds literally — git merge-tree merges clean and the result tree is bit-identical to main's (min_p Option<f32> preset at atlas-kernels/src/lib.rs:113-165; runtime sampler files byte-identical). Nothing of value dropped.

PR #509, opened 2026-08-14
spark-model: host-testable 3% kill-bar for the batch2 oracle

facepunch47:
## Summary

Continues #499. The cold-DRAM oracle example is unchanged in what it measures; the 3% kill-bar, candidate parse, and STREAM/peak defaults now live in `spark_model::layers::ops::gemv_batch2_oracle` so CI can lock them without a GPU.

This is still not a kernel change and not a default-on. Do not set `ATLAS_GEMV_BATCH2_CPASYNC=1`.

Also fixes a pre-existing harness bug that blocked the last gate: `Bfcl::verdict()` applied the 27B golden-draw MLPerf-edge bar (83.64 / **85.32**) to every variant, including the 35B echolp draw. Echolp now uses its own ratchet (84.66 / 83.32), locked to `BENCH.toml`. Subset still uses the MLPerf-edge floors — those were not lowered.

`report.rs` is on the BFCL invalidation set, so both BFCL records from earlier today no longer cover HEAD. Echolp then subset are re-running on GB10. The other four gates still cover (a BFCL driver change cannot move them).

Closes nothing. Campaign notes: `docs/campaigns/gb10-batch2-rhythm-2026-08/README.md`.

## GPU baseline (GB10, 2026-08-14)

`:8888` stopped for the run. Template `w4a16_gemv_batch2`, 40 iters, weight copies cycle past L2:

| Shape | us | GB/s | % STREAM 230 | % peak 273 |
|---|---|---|---|---|
| GDN in_proj 12288×2048 | 63.14 | **224.2** | 97.5% | 82.1% |
| attn Q 8192×2048 | 42.91 | **219.9** | 95.6% | 80.6% |
| GDN out 2048×4096 | 23.91 | **197.4** | 85.8% | 72.3% |

`PASS — baseline recorded (no candidate)`. in_proj is already 97.5% of STREAM.

## Dual-issue (#500) — Neutral, leave unwired

Oracle ran with `ATLAS_GEMV_BATCH2_CANDIDATE=w4a16_gemv_batch2_dualissue:w4a16_gemv_batch2_dualissue` (`.cu` not committed):

| Shape | vs batch2 (run 2) | Verdict |
|---|---|---|
| GDN in_proj 12288×2048 | 0.985× | Neutral |
| attn Q 8192×2048 | 0.971× | Neutral |
| GDN out 2048×4096 | 0.958× | Win (not a default-on shape) |

[...]

PR #512, opened 2026-08-14
atlas-plugin: pin ssm_cache_slots=256 on BFCL self-start

facepunch47:
## Summary

BFCL self-start now applies `--serve-override ssm_cache_slots=256` from `BENCH.toml`, the same pin the poison gate already uses on the CLI. The flagship recipe leaves the Marconi snapshot pool at its default 16, so a 995/1004-sample serial generate evicts its own intermediates (KV hash can still hit; the SSM slot is gone). That is capacity, not a score lever.

`[benchmarks.serve_overrides]` is the SSOT. Self-start merges those pins, then the operator's `--serve-override` (operator wins a clash). `check_record` refuses a record that is missing the pin, so an old 16-slot BFCL PASS cannot be copied onto this SHA.

AST floors are unchanged and locked by test:

- `bfcl-subset-echolp`: 84.66 / 83.32, n=1004
- `bfcl-subset`: 87.44 / 88.59, n=995

Poison gets the same machine-readable pin (it already documented the CLI flag). TTFT and agentic keep the recipe default.

Does not start a GPU bake. #509 is still running echolp on the 16-slot recipe; this PR rebakes after those records land. Do not copy old records.

## Test plan

- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy -p atlas-plugin -p spark-server --tests -- -D warnings`
- [x] `cargo test -p atlas-plugin --lib gate::` (195 passed), including pin parse, AST-floor lock, and `check_record` refusing a missing/wrong pin
- [x] `cargo test -p spark-server --bin spark bench_selfstart` (23 passed), including merge (operator wins) and resolve carrying the pin
- [ ] GPU re-bake of BFCL (and any other gate this `crates/` + boundary-file diff invalidates) after #509 records are on that branch — not started

## Notes for reviewers

- Follow-up to the #509 bake discussion: prefix hits are thrown away when the 16-slot pool evicts; this pin is the same override poison already uses.
- Later, if wanted: C=4–8 generate after a C=1 vs C=4 score lock. Not this PR.
- `BENCH.toml` stays outside the closure hash. The new `check_record` pin check is what stops a BENCH.toml-only edit from leaving an old record green.

## CLA

### DIFF
--- a/crates/atlas-plugin/src/gate/bench.rs
+++ b/crates/atlas-plugin/src/gate/bench.rs
@@ -58,6 +58,17 @@ pub struct BenchEntry {
     pub status: String,
     #[serde(default, skip_serializing_if = "String::is_empty")]
     pub note: String,
+    /// Recipe keys self-start applies on every `--pull-request-gate` run.
+    ///
+    /// Empty (and omitted from the TOML) is the normal case: the recipe serves
+    /// exactly as pinned. Non-empty is a gate-local pin the recipe itself
+    /// must not carry — e.g. `ssm_cache_slots = "256"` on BFCL, so a 1004-
+    /// sample serial generate cannot evict its own Marconi snapshots. Values
+    /// are strings, matching `--serve-override KEY=VALUE`. `port` is refused:
+    /// self-start binds a free port and a second opinion would name a listener
+    /// that is not there.
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub serve_overrides: BTreeMap<String, String>,
     /// Thresholds. Absent for `unmeasured` — see the module docs.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub metrics: Option<BTreeMap<String, super::record::Bound>>,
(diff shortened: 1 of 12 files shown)


tbraun96:
Closing: this PR's content is on `main`. #525 (`04726b148f`) carried it as a "semantic re-port of PR #512 onto the post-#513 self-start structure".

Re-checked today against `main` `9e597531`: the pin itself is declared at `kernels/gb10/qwen3.6-35b-a3b/BENCH.toml:184-185` — `[benchmarks.serve_overrides] ssm_cache_slots = "256"` — and BFCL self-start merges it with the operator winning a clash (`baseline_pins_are_applied_and_the_operator_wins_a_clash`, `crates/spark-server/src/cli/bench_selfstart_tests.rs:161`; `serve_overrides_are_assembled_into_the_baseline`, `crates/avarok-plugin/src/gate/bench_serve_pin_tests.rs:19`).

Thanks — the pin is how the gate runs today. No branch deleted.

PR #513, opened 2026-08-14
Qwen3.8-27B-NVFP4 support: kernel target, template, and CLI flag precedence

tbraun96:
Adds full support for the dense `unsloth/Qwen3.8-27B-NVFP4` checkpoint, and fixes two
pre-existing defects found while getting it to serve correctly.

## Context

Qwen3.8-27B-NVFP4 is **architecturally identical** to Qwen3.6-27B-NVFP4: all 1968 tensors
match in name, shape and dtype, and the quantization config differs only by a
compressed-tensors version string. The tokenizer vocab/merges/added_tokens hash identically.
Only the weights differ.

That means **no new CUDA kernels** — and it also means the two checkpoints are indistinguishable
to anything keying on `(model_type, hidden_size)`, which is what made this more than a config
addition.

The checkpoint already serves and passes the Gate C2 NVFP4 smoke (3/3 tool calls, 3/3
coherence), and scores 10/10 `webserver_ok` + 10/10 `followed_directions` on the
agentic-webserver task.

## What's here

**`feat(kernels)`: qwen3.8-27b target + unambiguous resolution.** Adding a target that declares
the same `(qwen3_5, 5120)` pair as qwen3.6-27b would have let `.find()` iteration order pick the
winner. Auditing for that found the hazard was **already live twice**, silently order-resolved:
`(qwen3_6_moe, 5120)` across qwen3.5-27b/qwen3.6-27b, and `(nemotron_h_puzzle, 4096)` across the
two nemotron puzzle targets. Resolution now uses declared `[model] match_names` needles *only* to
break a tie in the winning specificity tier; zero or several survivors is a hard
`TargetResolveError::Ambiguous`, never a wildcard fallthrough and never build order. A build-time
guard panics when colliding targets omit needles, and `--kernel-target` pins by name for
identity-free refs. The new target duplicates no `.cu` files — `[model] kernel_source` redirects
to qwen3.6-27b's tree, verified by an nvcc build producing the identical 177-module set.

Needles were chosen over fingerprinting `transformers_version` / the quant version string:
those describe the export toolchain, so a re-export of 3.6 with newer transformers would
silently flip targets — the opposite of failing loud.

**`feat(template)`: retire the qwen3_5 override, add `preserve_thinking`.** Every dense Qwen
generation reports `model_type = "qwen3_5"`, so the April-2026 `jinja-templates/qwen3_5.jinja`
was served for all of them by file presence alone, denying 3.8 its own August template
(`reasoning_effort` injection, inverted `preserve_thinking` default). The override carried
nothing the checkpoints' own templates lack — it is a strict subset of the official 3.6 template,
and the unsloth additions are already neutralized in Rust preprocessing — so it is deleted rather
than refreshed. A golden test renders the retired override and the shipped 3.6 template through
the production path and asserts **byte equality**, so Qwen3.6-27B prompt bytes and the MLPerf
reference they underpin are provably unchanged.

`preserve_thinking` lands as a MODEL.toml `[behavior]` key, tri-state `Option<bool>` where `None`
means *do not inject the Jinja variable* — passing Jinja `none` would silently flip 3.8's default
from keep to strip, since its template tests `is undefined`.

Also fixes a live 3.8 bug: `ReasoningEffort::Max.as_str()` returned `"max"`, which 3.8's
validator rejects, so every `reasoning_effort: max` request would have **400'd**. Adds a real
`Medium` variant, previously demoted to `Low` and producing wrong bytes.

**`fix(cli)`: explicit flags beat MODEL.toml defaults.** `--num-drafts` had
`default_value_t = 1` while the merge used the *value* 1 as its "not passed" sentinel, so an
explicit `--num-drafts 1` was indistinguishable from omission. On the dense 27B target
(`default_num_drafts = 3`) every run pinning `--num-drafts 1` actually served **K=4**. Auditing
the same class found `--kv-cache-dtype` (explicit `fp8` discarded on a bf16-default model) and
`--fp8-kv-calibration-tokens` (explicit `0` could not disable calibration). `--dflash-gamma` has
the same latent defect — reported, not fixed here, as a real fix needs the drafter config
threaded into pool sizing.

## Impact on recorded baselines

The `--num-drafts` defect means **Gate C2/D runs on the dense 27B that pinned `--num-drafts 1`
measured K=4, not the K=2 the pin implies.** Gate A/B on the 35B flagship are unaffected — that
target sets `default_num_drafts = 1`, equal to the pin. A/B comparisons remain valid (baseline
and PR legs shared the same silent K=4), but the configuration label is wrong and all wall/TTFT
numbers on that model were taken at K=4. The sentinel and `default_num_drafts = 3` both exist at
the earliest commit reachable in local clones (2026-07-02 graft root); it could not be dated
earlier from grafted history.

Recommend re-running one genuinely-K=2 dense-27B leg before trusting the K=2 label.

## Verification

- `cargo clippy --workspace --tests` — exit 0, zero warnings
- `cargo test --workspace` — **4664 passed, 0 failed**
- Real nvcc builds: single-target qwen3.8-27b (177 kernels, 14 overrides) and multi-target (24
  targets, exit 0), with `--ignored` registry tests 6/6
- Routing pins: 3.8 → qwen3.8-27b; unsloth/nvidia/centml/base 3.6 → qwen3.6-27b;
  Kbenkhaled/Qwen3.5-27B → qwen3.6-27b preserved; bare path → `Ambiguous`; pin resolves it
- Negative-tested the build guard: stripping `match_names` from 3.8 panics the multi-target build
  with an actionable message

## Not included

Multi-model benchmark support (letting one benchmark carry several model variants with per-variant
thresholds and serve recipes) is still in progress on a separate branch. It matters here because
`agentic-webserver`'s wall budget is calibrated on the 35B MoE and the dense 27B legitimately
exceeds it — 10/10 correctness, Σwall 1925s vs a 1000s budget — which is a miscomparison, not a
regression.

Not verifiable on a CPU-only box: the `build_codegen` emission of the new behavior field. It is
exercised by the nvcc builds above but worth a second look on a GPU box.


rsafier:
Heads-up from coworking on the vision side: **`closure_attestation` is red on this branch**, and it looks like the `kernel_source` redirect and the attestation don't yet agree.

```
$ cargo test -p spark-server --test closure_attestation

the_baked_attestation_reproduces_from_the_tree ... FAILED
gb10/qwen3.8-27b/nvfp4: the tree-side hash disagrees with the one baked at build
time. Either the source sets differ (collect_cu_files vs taxon::sources) or the
config lists do — not a kernel change, a bug in one of the two resolvers.
  left:  "9c3dab35196f6885e5f92fa7a45473a0f9e6315a3e0ffb589a383ff518ec261c"
  right: "23b8a88c1b64b9dcde28323832c3956d5963f8e8aa17dfc9520a6987058d12f1"
```

Two things point at the redirect rather than at a stale build:

**The target it names is the only one that redirects.** `qwen3.8-27b` is the sole target declaring `kernel_source = "qwen3.6-27b"`, and the failing entry is `qwen3.8-27b/nvfp4` — a path that does not exist on disk, since that target deliberately ships no `.cu` tree. So one resolver is following the redirect and materialising an entry under the redirecting target's name while the other is not.

**It is not staleness.** I first assumed a stale `target_ptx.rs` (the `include!`-not-tracked hazard `lib.rs` warns about) and was wrong: it survives `cargo clean -p atlas-kernels`, and the tree-side hash *moves* across runs while the baked one holds constant.

Verified as pre-existing rather than something we introduced — I ran it on `50285deb` with none of our commits applied and it fails identically, then again with three commits on top and got the same two hashes. `a_malformed_entry_refuses_to_deserialize_rather_than_defaulting` and `the_baked_json_shape_matches_what_the_gate_deserializes` both pass, so it is specifically the reproduce-from-tree leg.

One adjacent observation, lower confidence: `tokenizer::tests::qwen_dense::q38_explicit_efforts_render_their_sentences` failed once in a full `cargo test --workspace` run but **passes in isolation on both branches**, so it reads as a parallelism flake rather than a real failure. Flagging it only because it surfaced in the same sweep; I have not characterised it.

Separately — we're building on this branch for the vision sprint (resolution cap, a registered vision benchmark, video support). Nothing pushed anywhere yet. The one thing that touches your files is a three-line addition of the card's `min_p = 0.0` to `[sampling.thinking_text]`, `[sampling.thinking_coding]` and `[sampling.non_thinking]`, which only became expressible with #511 — your presets currently omit it, so 3.8 serves at the server's `--default-min-p 0.08` against a card that specifies 0. I deliberately left `[sampling.tools]` alone: the card ships no tools preset, and your note about penalties landing before the grammar mask applies to `min_p` as well, which is your measured territory rather than mine.

tbraun96:
### Note: a latent strix-hip bug this branch does **not** fix (deliberately)

Recording it here so it isn't lost. **No change to this PR** — it needs AMD hardware, which no box here has.

**What broke.** Commit `680b3a568` (the #473 batch) added VARLEN support by inserting two pointer parameters — `const int* __restrict__ cu_seqlens` and `const int* __restrict__ kv_lens` — **immediately after `const unsigned int batch_size`, mid-list**, in the `#ifdef PREFILL_BATCHED` block of `kernels/gb10/common/prefill_paged_compute.cuh` (lines 181-182), and updated the shared Rust wrapper to pass them.

`kernels/strix/common/prefill_paged_compute.cuh` is a **symlink into gb10**, so the SCALE target picked the change up automatically. `kernels/strix-hip/common/prefill_paged_compute.cuh` is a **real forked copy** and was missed — it still carries the pre-#473 signature in both kernel entries (line 106 `KERNEL_NAME`, line 403 `PAGED_CONCAT(KERNEL_NAME, _64)`). The batched `.cu` wrappers in strix-hip are themselves symlinks into gb10, so this one header is the entire delta.

**Consequence.** Launches use `void**` kernelParams, so the driver copies positionally. The kernel's 7th parameter `q_len` receives **the low 32 bits of the `cu_seqlens` pointer**, and everything after shifts by two: `kv_len` ← `kv_lens` ptr, `q_offset` ← `q_len`, `num_q_heads` ← `kv_len`, and so on; the wrapper's last two arguments are ignored.

- **UNIFORM mode** passes `cu_seqlens` as NULL → `q_len = 0` → every block exits at `q_start >= q_len_eff` → **the output buffer is never written**. Silent garbage attention, no crash.
- **VARLEN mode** → arbitrary huge `q_len` → OOB reads / GPU fault.

Affected (strix-hip only): `inferspark_prefill_paged_batched{,_64}`, `..._fp8_batched{,_64}`, `..._nvfp4_batched{,_64}`. Non-batched variants take the `#else` branch and are fine.

**Why no test catches it.** The arity pin (`crates/atlas-kernels/tests/kernel_arity.rs`) validates against compiled PTX baked per target; strix-hip compiles via hipcc to HSACO, and CI's kernel-compile job covers gb10 targets only.

**Reachability.** Gated behind `ATLAS_Q12_BATCHED_FIRST_CHUNK=1` / `ATLAS_PREFILL_CODISPATCH=1`, and the strix-hip paged-prefill port is already marked "compiles-pending / numerics-pending GPU". Latent until someone runs batched prefill on gfx1151 — then wrong by construction.

**Fix recipe for whoever has a gfx1151 box:** insert the two pointer params after `batch_size` in both signatures of the strix-hip header, then port the VARLEN body logic from gb10 (`q_base_b` / `q_len_eff` / the per-stream `kv_len` override around gb10 lines 199-218, plus `q_batch_off` addressing) into both the base and `_64` bodies; compile with hipcc and validate numerics. It was deliberately **not** blind-edited here — a ~26KB HIP header edited without a compiler is a good way to introduce something worse than the bug.

Issue #328, opened 2026-07-17
[bug] Increase prose_tokens


dribagnac:
### Atlas version / docker tag

avarok/atlas-gb10:latest (2026-07-01)

### Model

nvidia/Qwen3.6-35B-A3B-NVFP4

### Hardware

1x GB10, single CPU

### Launch command

```bash
docker run -d --name atlas \
  --network host --gpus all --ipc=host \
  --restart unless-stopped  \
  -v ~/.cache/huggingface:/root/.cache/huggingface \
  avarok/atlas-gb10:latest \
  serve nvidia/Qwen3.6-35B-A3B-NVFP4 \
    --host 0.0.0.0 \
    --port 8000 \
    --max-seq-len 65536 \
    --kv-cache-dtype fp8 \
    --kv-high-precision-layers auto \
    --gpu-memory-utilization 0.85 \
    --scheduling-policy slai \
    --enable-prefix-caching \
    --speculative \
	--mtp-quantization nvfp4  \
    --num-drafts 2
```

### Expected behaviour

increase prose_tokens value,  I don't find any option for that ?

### Actual behaviour

Using Pi.dev (Pi agent) with Atlas on my DGX10.
Following error on Pi.dev : 
 Error: Model stopped because it reached the maximum output token limit. The response may be incomplete.

Atlas log:
2026-07-17T08:06:27.560274Z  WARN spark::scheduler::decode_logits_content: Inter-tool prose budget exhausted, ending response (rollback declined) prose_tokens=385 max=384 reason=StreamUnsafe

Is-there a way to increase prose_tokens ? I don't find any option for that ?

Tanks.
Regards
Daniel

### Additional context

Using Pi agent (pi.dev)
 Error: Model stopped because it reached the maximum output token limit. The response may be incomplete.

dribagnac:
Issue is not linked to the model (I tested with some Qwen 3.5 or 3.6 variants).

Qwen-Coder-Next
docker run -d --name atlasCoder \
  --network host --gpus all --ipc=host \
  --restart unless-stopped  \
  -v ~/.cache/huggingface:/root/.cache/huggingface \
  avarok/atlas-gb10:latest \
  serve unsloth/Qwen3-Coder-Next-FP8 \
    --host 0.0.0.0 \
    --port 8000 \
    --max-seq-len 65536 \
    --kv-cache-dtype fp8 \
    --kv-high-precision-layers auto \
    --gpu-memory-utilization 0.90 \
[...]

PR #514, opened 2026-08-14
Batch 6: restore the FP8 MoE BFCL accuracy, make the agent guards reachable and honest, harden the SSM gate (#328)

tbraun96:
Twelve commits, one per fix, so a gate failure stays bisectable inside the stack.

## The accuracy regression — found, fixed, measured

`bfcl-subset-echolp` (Qwen3.6-35B-A3B-FP8, N=1004, temp 0) dropped when the batch4+TUI stack landed. Two independent causes:

| build | normalized | overall |
|---|---|---|
| last-good `66e1f2e86d` | 86.95 | 86.55 |
| regressed (main) | 85.34 | 84.76 |
| + router-GEMM fix | 86.32 | 86.16 |
| **+ spec-entry pin (this stack)** | **86.77** | **86.35** |

**Cause 1 — the MoE router-logits GEMM** was rerouted onto an `mma.sync` tensor-core kernel. Router logits are the model's only discrete control flow: stored to BF16, then top-8. On that grid near-tied experts sit ~1 ulp apart, so an accumulation-order change **flips which experts run** — every MoE layer, every prefill token. Pinned back to the scalar kernel. Verified beyond the score: a 1004-sample token-level replay shows the fixed build **byte-identical to last-good on every sample** with speculation off.

**Cause 2 — the MTP gate picks the decode path by stopwatch.** With `--speculative`, arbitration between the serial (M=1) and batch-K verify forwards is decided on measured wall-clock, and those forwards sit on the batch-K numerics floor. So *which* path serves an answer opening depends on how fast the binary happens to run. Timestamp-matched evidence: `MTP gate: switching Mtp -> Serial ... depth=1752` lands on the exact step where the traced verify stream ends, and all three flipped samples ran inside that Serial dwell. The model is genuinely bistable on `live_irrelevance` prompts — last-good emitted the same bogus tool call on a different sample — so the defect is that **speed selects the branch**. `ATLAS_SPEC_ENTRY_PIN` (default 8 tokens) pins post-`</think>` answer openings to the verify path.

Remaining gap to last-good is 0.18 / 0.20, inside the measured ±0.40 band. The ratcheted floors (86.50 / 86.90) pass on their own merits rather than being trimmed to fit.

**Ruled out along the way, each by evidence:** Batch 5 entirely (its pre-Batch-5 branch scored identically), drafter quality (82.32% vs 82.97% acceptance over 7,100/6,200 K2 steps), the multi-seq batchm rewrite (those paths don't execute on this recipe), `moe_grouped_a`, MTP `sliding_window` (no SWA on this checkpoint), and a `lmhead_batch_gemv_enabled()` flag documented "Default ON" that **nothing calls**.

## Two more real bugs found while hunting

**Pointer-keyed MoE cache** — a process-global `HashMap` keyed on a **device address**, no invalidation, `n` absent from the key. Model teardown frees the tables, the driver reuses the addresses, and the next model reads the previous model's expert pointers as weights. Silent corruption across an in-process swap. Replaced with layer-owned tables; the per-prefill D2H saving is kept and improved.

**Tool-call fabrication** — `parse_complete_call` promoted *any* bare identifier inside `<tool_call>…</tool_call>` to a zero-argument call, for every model, though its own doc comment says that shape belongs to Poolside.

## The #328 family — guards that fire on legitimate output and lie about it

Issue #328's reporter hit four guards in a row on `Qwen3-Coder-Next-FP8`. Each was tuned for a past incident, each unreachable, and two mislabelled their own verdict.

- **Inter-tool prose budget**: a July fix raised the default 384 → 3072 **on a constant nothing reads** — `WatchdogParams::from_behavior` always overwrites it — and that fix's own regression test asserted the dead constant, so it stayed green through the whole regression. Now one SSOT constant, a `--max-inter-tool-prose` flag (`0` disables), and honest `finish_reason="length"` with a named `guard_stop` instead of `"stop"` on a truncated response.
- **`reasoning_effort` ladder**: hardcoded `64/128/256/512/1024`, with `"medium"` *and every unrecognised string* → 256. Harnesses send `reasoning_effort` as SDK boilerplate and a client channel outranks the server flag, so `--max-thinking-budget 16256` lost to "medium" on every request. Levels now scale from the model's ceiling; byte-identical at the default, pinned by test.
- **In-think tool-call cancel** and **content-loop watchdog**: both now reachable, counted, proportionate, and named in the log.
- **`effort_capped_at_ceiling`**: one boolean, one model — `qwen3.5-397b`, whose sweep measured 256 as *worse* than its 128 ceiling, so a boilerplate `"high"` would have handed it the known-bad budget. Deliberately **not** a five-number per-model table: no vendor publishes those numbers, and inventing configuration is how 384 and 256 got here.

## SSM poisoning gate hardening

B1–B8 from the gate's own audit. The important one is **B3**: `cached_prompt_tokens` was parsed, carried, and never read, so with prefix caching off the gate returned a green PASS **proving nothing**. Now enforced in the run verdict *and* floored in BENCH. Plus zero-ref collapse, unmeasured-round attribution, reference-round anchors, and a BENCH note that no longer references a metric that doesn't exist.

**Not yet done:** the positive control — proving this gate *fails* on the pre-fix batch4 tree. Until that runs, it remains an unfalsified detector, which is precisely the class of problem B3 fixes.

## Dense-27B agentic leg (unmeasured)

Registered so the dense model can be baselined: `status = "unmeasured"`, **no metrics table**, `default = false`. The schema enforces it — `baseline_for` drops unmeasured entries so `--pull-request-gate` cannot self-serve it, and a stray record fails loudly with "no baseline for model" rather than passing vacuously. The 35B's 1000s Σ-wall cannot transfer: the 35B is A3B (~3B active/token) where the dense 27B activates all 27B. The note carries the calibration protocol (idle box, ≥3 repeats, ratchet from measurement) and the decision whoever writes the ceiling must make first — blowup detector or perf bound.

**Cost:** adding a family touches `benchmarks/agentic/descriptors.rs`, which re-opens the standing Gate A record. This PR owes a fresh N=10 agentic run.

4653 workspace tests pass; fmt and clippy clean.


tbraun96:
Site break is confirmed ours (the Qwen3.8 work), not batch 6 — and it's fixed on #513. Details, since the diagnosis differs slightly from the one going around:

**It is not the hand-rolled YAML parser.** `gen-models.mjs` parses the new recipe fine. The failure is `vendorOf()`, which hard-exits on an unmapped recipe family by design:

```
Unmapped recipe family "qwen3.8" — add it to VENDOR_OF_FAMILY. SSOT: https://github.com/Avarok-Cybersecurity/atlas-recipes
EXIT=1
```

That guard is working as intended ("PCND — no silent default bucket"); the maps just had no entry for the `recipes/qwen3.8/` directory that atlas-recipes#19 introduced at 16:52Z. So any PR whose site build runs after that recipe landed fails, regardless of its own contents — which matches #514 being the first to run since the last green build (#508 at 09:26Z).

**Fix:** `6164debe1` on #513 adds `qwen3.8` to `FAMILY_DISPLAY` and `VENDOR_OF_FAMILY`. Verified against the merged recipes tree — exit 0, and Qwen3.8 renders as its own subfamily under the Qwen vendor (Qwen now 18 recipes).

**Proposed order,** which I think matches what you'd want anyway:

1. Merge #513 (Qwen3.8 support, carrying this fix).
2. Rebase/merge `stack/batch6` onto the new `main`, and its site build goes green without any change on your side.

#513 is otherwise ready — `fmt` clean, `clippy` exit 0, `cargo test --workspace` **4717 passed / 0 failed** — but I still have two agentic-webserver tiers and one BFCL ST-995 leg to run against it before merging, which is a few hours. **If batch 6 is blocked and you'd rather not wait, say so and I'll split `6164debe1` out to a one-commit PR you can merge immediately**; it's a two-line change to a map and touches nothing else.

Sorry for the breakage — a new recipe family directory is exactly the kind of thing that should have had the site build checked before the recipe merged, and I didn't.

tbraun96:
## Stacking: this PR sits at level 1, on top of #513 (Qwen3.8)

Handing this off — **#514 will be stacked with the Qwen3.8 PR**, not gated on its own:

| Level | PR | Branch |
|---|---|---|
| **0** (base) | **#513** | `feat/qwen3.8-27b-support` |
| **1** (this PR) | **#514** | `stack/batch6` — Qwen3.6 work |

The two get **one** benchmark cycle over the combined stack rather than a cycle each, so
`PR benchmark gate` is expected to stay red here until that happens — no records exist at
this frozen sha, which is the only reason it fails.

### CI state at `4d526a4ef`
Everything except `PR benchmark gate` is green: `cargo test --workspace` (4653), `clippy --tests`,
`llvm-cov`, `fmt`, `deny`, `typos`, `≤500 LoC`, `SPDX`, `kernel shadow structure`, GDN `PINS.sha256`,
`nvcc → PTX`, `Analyze (rust)`, `Build SvelteKit site`. (`CodeQL` is NEUTRAL — its umbrella status.)

### Accuracy condition — met
The BFCL regression this stack exists to fix measures **86.77 / 86.35** (overall / normalized),
against this stack's own ratcheted floors of **86.50 / 86.90**. The 0.18–0.20 gap sits inside the
±0.40 noise band. Two root causes, both fixed here: the router-logits GEMM reroute flipping expert
selection, and the MTP gate selecting the decode path by stopwatch.

### One fix in here that is *not* mine
`site: map the qwen3.8 recipe family to the Qwen vendor` — `atlas-recipes` gained a `qwen3.8`
family (20ac6e80, 16:52Z today) and `gen-models.mjs` fails fast on an unmapped family by design.
That turned `Build SvelteKit site` red on **every PR opened after 16:52Z**, not just this one
(last green was #508 at 09:26Z). Reproduced against a fresh recipes clone. Flagging it for #513's
author since it originates with the Qwen3.8 recipe: if you'd rather own that one-liner there,
drop it from this branch — it only lives here because this stack was the first to hit it.

tbraun96:
Closing as SUPERSEDED by main (da56736dc), via the #519 stack's merge+reconcile — not an identity merge. 41/61 files are blob-identical on main; every remaining PR-side hunk was deliberately superseded: (1) main re-pointed the agentic gate to accept Qwen3.8-27B as a MEASURED target (the 2026-08-14 flagship change this PR's own comments anticipated); (2) the '_ => Medium' unknown-effort arm was replaced by the parse_wire_effort SSOT which 400s invalid efforts on every channel. NOTE for future audits: merge-tree shows conflicts ONLY because main evolved beyond this PR on the same lines — do not reopen on that signal. The router-GEMM scalar pin, spec-entry pin, watchdogs, and ssm hardening are all on main.

PR #516, opened 2026-08-14
Video support: temporal preprocessing, video_pad tokens, and the MRoPE t axis

### LINKED ISSUE #515: Video support: video_url content parts are silently dropped
## Summary

A `video_url` content part is **silently dropped**. The model answers from the surrounding text as though no video was sent — a plausible answer with no signal that the video never arrived. This issue scopes what it takes to serve video properly, and what to ship first.

Found while building the `vision-fidelity` gate (#513). Everything below is verified against the tree and the shipped checkpoints rather than inferred from the HF implementation.

## Today's behaviour

Two drop sites, both silent:

- `crates/spark-server/src/openai/chat_message.rs:326` — Chat Completions: `_ => {} // ignore unknown part types`
- `crates/spark-server/src/openai/chat_message.rs:255` — Responses API: the `input_image | image_url | image` arm simply does not match, and there is no `else`

`ContentPart` (`crates/spark-server/src/ir/message.rs:143`) has `Text` and `Image` only, with a doc comment already anticipating this: *"Open for future modalities (audio, video, file)."*

## What is already in place

More than expected — the model side is not starting from zero.

- **`temporal_patch_size = 2` is real and honoured.** `vision_preprocess.rs:185` builds patches with a `T` axis and fills it by **duplicating the still image** `tp` times (`:193`). A video frame-pair is the case that axis was designed for.
- **MRoPE has three independent position streams.** `ops::rope_mrope_interleaved` (`layers/ops/embeddings.rs:160`) takes `pos_t`, `pos_h`, `pos_w` as separate device pointers and assigns rotary pairs round-robin by `pair_idx % 3`. The kernel needs nothing new.
- **`<|video_pad|>` exists in the tokenizer** and the bundled chat template already branches on `videos` / `video_count` / `video_pad`.
- **The checkpoints declare a full video processor.** `unsloth/Qwen3.6-27B-NVFP4` → `processor_config.json.video_processor`; `unsloth/Qwen3.8-27B-NVFP4` → a dedicated `video_preprocessor_config.json`:

  ```
  do_sample_frames: true,  fps: 2,  min_frames: 4,  max_frames: 768,
  merge_size: 2,  temporal_patch_size: 2,  size.longest_edge: 25165824
  ```

  Note that bound is **larger than the image bound** (25165824 vs 16777216) and lives in the same file. The image-bound reader deliberately addresses `image_processor` explicitly so it cannot pick this up; a video path must read the video one just as deliberately.

### 1. The position builder is image-shaped

`model/trait_impl/prefill_b/upload_meta.rs:111-165` is where this actually bites. It walks `vision_image_grids` as `(gh, gw)` pairs and, for each pad run, emits `T = base` **constant across the whole item**, advancing `current_pos += max(gh, gw)` afterwards.

That is correct for a still. For video, `T` must advance **per frame**, so the item's grid needs a temporal extent: `(t, gh, gw)`, with `t = frames / temporal_patch_size`. This is the single most invasive change, because the grid vector is threaded through the co-dispatch path (`vision_grid_base`, `vision_owned_images`) and shared across concurrently-batched requests.

### 2. Frame decode

The hard dependency question. `spark-model` currently pulls `image = { features = ["jpeg", "png"] }` and nothing else. Options, roughly in order of cost:

- **Animated GIF / APNG** — pure Rust, `image` crate feature flag only, no system dependency. Enough to prove the whole pipeline end to end.
- **MP4/H.264** — what users will actually send. Needs a real decoder (`openh264-sys2` vendored + `cc`, or ffmpeg bindings with a system dependency). This is the decision that sets the schedule.

Frame **sampling** is a separate concern from decode and is specified by the checkpoint: `fps: 2`, clamped to `[min_frames, max_frames]`.

### 3. Token accounting and splicing

`<|video_pad|>` runs instead of `<|image_pad|>`, with `t * gh * gw / merge²` merged tokens per video. Every place that counts image pads (`image_count()`, `image_pad_counts`, the Jinja marker expansion) needs the video equivalent.

### 4. Memory

ViT attention is O(patches²) and already clamps: the encoder caps at `CEILING_MAX_PATCHES = 16384` and logs when a checkpoint asks for more. A video is `t` frames of a full patch grid, so it hits that ceiling far sooner than a still. Frame count, resolution and the ceiling have to be reasoned about together, not independently.

## Proposed order

1. **Reject explicitly** (small, standalone, ship first). Turn both silent drops into a 400 naming the reason, matching the existing remote-URL rejection in `api/chat/msg_entry.rs:75-84`. A wrong answer becomes an honest error, which is strictly better than today and worth having regardless of when the rest lands.
2. **`ContentPart::Video` + wire parsing** on both APIs, base64 `data:` only.
3. **Decode + sampling**, GIF/APNG first to prove the pipeline, container format decided separately.
4. **`(t, gh, gw)` grids** through the preprocessor, the encoder, and `upload_meta`'s MRoPE builder — the real work.
5. **Video-token splicing** and the `<|video_pad|>` accounting.
6. **A short-video leg in the `vision-fidelity` gate**, with a tiny committed fixture provisioned the same way the eight image fixtures are, a deterministic geometry assertion (frame count × grid is arithmetic, exactly like the image rungs), and a no-video control so the leg cannot pass vacuously.

Steps 1 and 2 are independently useful. Step 4 is the one that touches the hot prefill path and deserves its own review.

## Related

- #513 — the vision gate and the resolution-bound work this came out of
- Remote image URLs are rejected today by design; an operator opt-in to fetch them is tracked separately

rsafier:
Draft. Implements #515 in slices; **stacked on #513** so the diff shows only the video work. Retarget to `main` once #513 lands.

## Where it stands

| slice | state |
|---|---|
| 1. decode, frame sampling, temporal grouping | **landed** (`9ca3234f`) |
| 2. `ContentPart::Video` + wire parsing, both APIs | next |
| 3. `msg_entry` collection + pad accounting | |
| 4. `(t, gh, gw)` grids through the co-dispatch path | |
| 5. MRoPE `t` axis in `upload_meta.rs` | the invasive one |
| 6. short-video leg in the `vision-fidelity` gate | |

Nothing is wired to the request path yet, so **no request's behaviour changes** at this point in the stack. A `video_url` part is still silently dropped, exactly as before — that is what slice 2 ends.

## The structural finding

The issue budgeted for encoder work. There is none: **Qwen3-VL's ViT needs no change.**

It has no temporal attention. Frames fuse *inside* a patch, because the flattened patch dimension is `C × temporal_patch_size × patch²` — and a still image already fills that axis by **replicating itself `tp` times** (`vision_preprocess.rs`). The temporal axis was always there. A still is the degenerate case of a video.

So a clip of `n` frames becomes `grid_t = n / tp` **temporal groups**, each group a full `grid_h × grid_w` patch plane built from `tp` consecutive frames — each shaped *exactly* like a preprocessed still. The groups ride the existing per-image encoder path untouched, and only the bookkeeping above them knows they are one item.

What genuinely differs is **position**. An image holds MRoPE's `T` coordinate constant across its whole pad run; a video advances `T` once per temporal group. That is why `grid_t` is carried rather than groups being flattened into independent images, and why videos need their own pad token. It is also why slice 5 is the one that deserves real review — it touches the hot prefill path and the co-dispatch grid bookkeeping.

## Decisions worth challenging

**GIF only, for now.** It is the one genuinely multi-frame container decodable in pure Rust with no system dependency, which lets sampling, grouping, token accounting and position assignment all be built and proven before taking on an H.264 decoder and its build-time cost. An mp4 is refused **by name**, with an `ffmpeg` one-liner to convert, rather than reaching the GIF decoder and failing as though the file were corrupt. Swapping in a real decoder later is a change to `decode_frames` alone.

**Sampling never upsamples.** A 4-frame clip asked for 16 frames returns 4. Repeating frames would inflate the token count with information that is not in the clip, and the model would be charged for frames that do not exist. Sampling is nearest-index over a uniform grid rather than every-Nth, which quantises badly at non-integer ratios.

**Defaults come from the checkpoint,** not from constants I chose: `fps: 2`, `min_frames: 4`, `max_frames: 768` are what every Qwen3-VL `video_processor` block declares. All three vision checkpoints also declare `video_token_id` 248057, and the bundled chat template already emits `<|vision_start|><|video_pad|><|vision_end|>` for a `{"type": "video"}` content item — the prompt side needs feeding, not building.

## Tests

18 in slice 1. Two carry most of the weight:

- `the_two_frames_of_a_group_are_actually_different` — if grouping ever duplicated one frame across the `tp` axis, *the way a still deliberately does*, a video would silently become a slideshow of doubled stills and every shape assertion would still pass.
- `the_frame_count_is_always_a_whole_number_of_groups` — swept over `n = 1..64` × five rates. Writing it as an unconditional invariant is what exposed that two of my own tests disagreed about a 1-frame clip. The invariant now carries the exception explicitly rather than the two quietly contradicting each other. The same sweep found a real `u8` underflow in the test GIF generator.

## Housekeeping

`vision_preprocess.rs` gives up its inline tests to a sibling file — sharing `MEAN`/`STD`, the data-URI splitter and the target-size helper with the video path carried it from 490 to 531 lines, and it is not on the file-size-cap allow-list. 262 now.

Closes #515 when the stack completes.


tbraun96:
Closing as SUPERSEDED by main (da56736dc) via the #519 stack: 103/118 files blob-identical on main (entire video benchmark module, fixtures, assets, mrope t-axis, media_integrity). Residual deltas are comment wording superseded by main's evolved --video-allow-ffmpeg gating and 'everything-skipped reports INCONCLUSIVE' semantics. merge-tree conflicts arise only because main evolved beyond the PR on the same lines.

PR #480, opened 2026-08-13
spark-model: wire lossless SW GEMV through attention, MTP, Nemotron

facepunch47:
Depends on #479 (lossless `w4a16_gemv_sw` default + `w4a16_decode_gemv` helper). This PR is the rest of the decode GEMV surface.

## What

#479 turned the bit-identical single-warp GEMV on for dense FFN and SSM. Attention, MTP, Nemotron, and MoE gate were still launching the 64-thread kernel on every M=1 NVFP4 decode GEMV.

- Qwen3 attention: `nvfp4_decode_gemv` helper (handle + N/8 grid stay coupled). 24 call sites.
- MTP: `MtpHead::gemv` NVFP4 arm + lm_head GEMV.
- Nemotron MoE shared-up and Mamba2 in/out proj.
- MoE router gate (decode + EP).

Kill switch is still `ATLAS_NO_GEMV_SW=1` via `ModelLevers::gemv_sw`.

## Why

GB10 decode is memory-bandwidth bound. Same bytes, more resident warps, same numerics (microtest bar: 100% bit-identical). Leaving attention on the old kernel would keep the slower occupancy on the other half of every token.

```mermaid
flowchart TD
  lever["ModelLevers.gemv_sw default ON"]
  helper["w4a16_decode_gemv / nvfp4_decode_gemv"]
  lever --> helper
  helper --> ffn["dense FFN #479"]
  helper --> ssm["SSM QKVZ + out_proj #479"]
  helper --> attn["attention Q/K/V/O + MLA"]
  helper --> mtp["MTP projections + lm_head"]
  helper --> nem["Nemotron MoE/Mamba2"]
  helper --> gate["MoE gate"]
```

## Tests

- Existing `gemv_sw` polarity + CUDA `#define` SSOT tests from #479
- New: attention decode files must not call `ops::w4a16_gemv(` — they go through `nvfp4_decode_gemv`

Out of scope: main-model lm_head (`w4a16_gemv_logits` FP32-output fork is a different kernel; do not share the SW grid).

## Rebase note

`origin/main` split `ms_phase_o_proj` into `trait_impl/multi_seq/attn/o_proj.rs` (#473). Rebuilt this PR as #479 tip + the wiring commit; the M=1 O-proj fallback now calls `nvfp4_decode_gemv` in `o_proj.rs`. The attention-decode test lists that file so a future `ops::w4a16_gemv(` cannot land silently.

## Tests

[...]

PR #519, opened 2026-08-14
fix(kernels): restore shared-memory LUT staging in the decode GEMV partials

tbraun96:
Fixes the decode-throughput regression bisected to #479, **though not to the magnitude that report describes** — see Scope below, which is the most important section here.

## Root cause — not what the bisect assumed

#479 shared the pipelined GEMV partial with the SW path to remove a 1-ULP divergence. In doing so it replaced `s_lut[byte_val & 0xF]` with a direct read of `__constant__ E2M1_LUT`. **That index is data-dependent, and an indexed constant load serializes the warp.**

Confirmed by compiling every variant with `nvcc -arch=sm_121a -cubin -O3` and counting SASS — `LDCi` is the indexed (warp-serializing) constant load:

| kernel | pre-#479 | main (regressed) | this fix |
|---|---|---|---|
| `w4a16_gemv` | LDS=33 **LDCi=1** | LDS=1 **LDCi=32** | LDS=33 **LDCi=1** |
| `w4a16_gemv_dual` | LDS=33 **LDCi=1** | LDS=1 **LDCi=32** | LDS=33 **LDCi=1** |

Registers 40 everywhere, zero spills in all trees.

**The bisect's stated mechanism does not exist.** The pre-#479 tree *already* ran the pipelined pair-walk — K16 chunks, `fmaf`, factored scale, `__bfloat1622float2`. #479 did not pessimize the K-walk. The lost LUT staging was its only semantic change to the base kernel. Every other non-SW decode kernel stages the LUT, including the untouched `w4a16_gemv_silu_input`; #479 broke a repo-wide convention in exactly two places.

## Why repair rather than revert #479 + #480

This is 3 files. A revert is 34, and throws away the 1-ULP parity fix, an `N%4` tail-guard fix, `silu_input_sw`, and #480's wiring.

The decisive property: `ATLAS_NO_GEMV_SW=1` reaches **every** `_sw` launch site and routes to base kernels that are SASS-equivalent to the pre-#479 tree. So **this fix's fallback arm is the revert's entire shipping configuration, from the same build** — reachable by an env var and a re-serve. The revert cannot do the reverse; it deletes the SW path and can never tell you whether that route was a win.

## Measured — dgx1, Qwen3.8-27B-NVFP4, `bench/quick_bench.py`, same box and flags, back-to-back

| arm | tok/s (server) | TPOT |
|---|---|---|
| regressed (`main`) | 26.0 | 10.6 ms |
| this fix, `ATLAS_NO_GEMV_SW=1` | **29.6** | 9.9 ms |
| this fix, default (SW on) | **29.5** | 9.9 ms |

Both arms agree within noise, so the SW route costs nothing now that it stages the LUT. Qwen3.6-27B on the regressed build measures **26.2** — the two models behave identically, as expected for one shared partial.

## ★ Scope — read before closing the regression report

The bisect reports **27.5 → 11.8 tok/s (~2.3×)**. **I could not reproduce 11.8** on either Qwen3.6-27B or Qwen3.8-27B; the regressed build measures 26.0–26.2 in this configuration. What this fix demonstrably recovers is **~13%**, not 2.3×.

Two hypotheses for the gap were tested and both failed: model-specific (3.6 and 3.8 measure the same) and long-context (the long-ISL run hit early EOS and measured nothing usable). The remaining differences are box, serve flags, ISL/OSL, batch size, or a different harness — note this fix's arm reads **29.6, above the bisect's healthy 27.5**, which itself suggests the configurations aren't directly comparable.

**The root cause and SASS evidence are solid and this is a strict improvement, but do not close the regression report on this commit alone.** Please post the exact serve command, box and benchmark invocation behind 27.5/11.8 and I'll confirm or refute against that case directly.

## Verification

`cargo fmt --all -- --check` clean · `cargo clippy --workspace --tests` exit 0 · `cargo test --workspace` **4611 passed / 0 failed** · both trees compile clean cubins for `sm_121a`, zero warnings, zero spills.

Three new tests pin the fix, each verified to fail without it (mutating the LUT read back to `__constant__` turns them red): `decode_gemv_partials_index_a_shared_staged_lut`, `sw_gemv_stages_the_lut_per_warp_without_a_block_barrier`, `sw_partial_shares_pipelined_k16_loop`.

Note for the gate: `ATLAS_SKIP_BUILD=1` means nvcc never runs, so a broken `.cu` passes fmt, clippy and the full suite. The cubins above were compiled by hand for exactly that reason.


tbraun96:
### Scope caveat resolved — the 27.5 → 11.8 figures came from an older tree

The PR body says I could not reproduce the reported 11.8 tok/s and that the regression report should not be closed on this commit alone. **That caveat can now be retired:** the maintainer confirms those numbers were measured against an earlier PR state, not the current tree.

That explains the discrepancy cleanly, and in a way my own hypotheses could not — I tested model-specificity (3.6 and 3.8 measure identically at 26.0/26.2) and long-context (the long-ISL run hit early EOS and measured nothing), and both came back negative because I was looking for a difference *within* the current tree that was never there.

So the corrected reading of the measurements:

| arm | tok/s | TPOT |
|---|---|---|
| current `main` (regressed) | 26.0 | 10.6 ms |
| this fix, `ATLAS_NO_GEMV_SW=1` | **29.6** | 9.9 ms |
| this fix, default (SW on) | **29.5** | 9.9 ms |

The **~13% recovery is the real, current-tree effect**, and the SASS evidence for the mechanism is unchanged and independent of any of this: 32 warp-serializing indexed `__constant__` loads per kernel where the pre-#479 tree had 1, restored to 1 here.

One consequence worth stating: the deeper collapse the bisect saw evidently existed at some earlier point and is no longer present in `main` on its own. Whatever else changed between that tree and this one already recovered most of it. This commit closes the remaining gap and removes the mechanism, so the class of defect cannot silently return — the three new tests fail if anyone reverts the LUT read to `__constant__`.

Also note: **#520** fixes the unrelated `Build SvelteKit site` failure showing on this PR. That check is red on every PR branched off `main` right now (an unmapped `qwen3.8` recipe family), not because of anything in this diff.

tbraun96:
## opt-A (BF16 lm-head batched-M GEMV) — dropped, not merging

The commit that was going to be stacked here has been removed from the branch. Recording why, since the claim was quoted in this campaign and should not be resurrected from memory.

**The claim:** routing the BF16 lm-head through `dense_gemv_bf16_batchm` was measured at +15% (29.4 → 34.0 tok/s).

**Re-measurement refutes it.** Same binary, same server, `ATLAS_NO_LMHEAD_DENSE_BATCHM=1` as the only variable — the strongest form of the test, since it removes tree/build differences entirely:

| Serve profile | ISL | opt-A off | opt-A on | delta |
|---|---:|---:|---:|---:|
| agentic 32K, bf16 KV + bf16 head | 74 | 12.6 | 12.6 | 0.0 |
| agentic 32K | 120 | 13.1 | 13.2 | +0.1 |
| agentic 32K | 512 | 12.5 | 12.6 | +0.1 |
| light 4K, default head | 74 | 14.3 (TPOT 69.91 ms) | 14.3 (TPOT 69.90 ms) | 0.0 |

Four configurations, two prompt lengths, two head dtypes. Every delta is at or inside run-to-run noise; the light-profile pair agree to 0.01 ms of TPOT. ISL 74 — the exact length the original claim was measured at — shows exactly zero.

**Mechanism.** opt-A accelerates the batched multi-row lm-head path that MTP verify uses. On this workload MTP is not paying:

| light 4K, ISL 74 | decode tok/s | TPOT |
|---|---:|---:|
| MTP off (`mtp=false, num_drafts=0`) | 14.1 | 70.80 ms |
| MTP on, K=4 verify | 14.3 | 69.90 ms |

Speculative decode contributes +1.4%, i.e. drafts are being rejected and each verify step yields ~1 token. With no multi-row batches to accelerate, opt-A is inert by construction — the code is not wrong, it just has nothing to do here.

**Decision:** dropped. It is a live change to the lm-head — the component pinned to bf16 for long-context coherence — justified entirely by an effect that does not reproduce. Carrying risk for a measured 0% is a bad trade.

The `quick-speed-bench` port stays; it is the tool that caught this.

Follow-up worth its own issue: MTP contributing ~1% on this family, while both Qwen3.8 gate recipes pin `num_drafts: 3` and pay ~1.5 GB of MTP pools for it.

PR #406, opened 2026-08-06
spark-model: gemma-4-E2B text-core numerical parity (BF16 MLP + double-wide FFN + KV sharing)

Marker689:
## Summary

Adds `google/gemma-4-E2B-it` support to Atlas — text core **and** multimodal
(vision + audio + video). E2B (5.1B total / 2.3B effective, 35 layers, 20
KV-shared, per-layer embeddings) is a new Gemma-4 shape: double-wide MLP band
(12288 intermediate on L15-34), native-BF16 MLP weights on disk, and
cross-layer KV sharing with PLE (per-layer-embedding) blocks.

**Text core** verified end-to-end against a BF16 HuggingFace oracle on the exact
25-token chat-template prompt: per-layer stage dumps match within ~0.02
(L15 FFN) / ≤0.19 (post-layer residuals through L29), L34 raw output
matches the oracle L34 hook, and full greedy generation matches the HF
oracle **verbatim**.

**Multimodal** (waves 0-4, now complete): the vision tower (~150M, gemma4
vision encoder), audio tower (~300M), their input/output projections,
modality-token embeddings (`image_token_id=258880`, `audio_token_id=258881`,
`video_token_id=258884`) and OpenAI/Anthropic content-part parsing
(`image_url`, `audio`, `video` / `input_audio` / `input_video`) are all
loaded and executed — image/audio/video requests are served end-to-end, not
dropped.

### Root causes fixed (text core)

1. **E2B MLP weights are native BF16 on disk** (`Bf16Raw`, no
   `weight_scale`), but the `Bf16Raw` branch of `quantized_any`
   runtime-quantized them to NVFP4 at load — ~5-17% `moe_out` drift, the
   first divergence from the oracle. `loader_a`/`loader_b` now keep the
   FFN BF16 end-to-end for E2B (`num_kv_shared_layers > 0`); 26B/31B
   (NVFP4 on disk) keep the NVFP4 default.
2. **Double-wide FFN band intermediate truncated**: the KV-shared band
   (L15+) projects to `2 * intermediate_size` (12288), but all 5
   `DenseFfnLayer` forward sites hardcoded `config.intermediate_size`
   (6144) — the GEMMs read only the first half of the `[12288, h]`
   weights. The layer now carries a per-layer `intermediate_size`
   (set by the loader), and the arena buffers are sized for the widest
   layer.

Plus supporting infrastructure: E2B config parsing, KV layer dims,
[...]

PR #521, opened 2026-08-15
fp8: an opt-in row-wise arm for mixed-precision checkpoints — plus the one-line GEMM fix that turns the BF16 lever from -73% into +12%

rsafier:
Qwen3.6/3.8-27B ship **mixed-precision** checkpoints: the SSM `in_proj`/`out_proj` weights are FP8 with **per-row** scales (`[N,1]`), while everything else is NVFP4. Atlas had no arm for that shape, so it re-quantised those FP8 tensors to NVFP4 at load — a **double quantisation** (FP8 → BF16 → NVFP4) on the tensors that feed every GDN prefill.

This PR removes that, measures what removing it is worth, and — the part worth more than the fix itself — measures it against the alternatives. Base is now `main` (the video stack this originally sat on landed via `da56736d`).

## What is in it

**1. A row-wise FP8 arm, opt-in behind `ATLAS_FP8_ROWWISE`.** A per-row scale is *detected* rather than assumed (`scale_is_per_row()` is a pure predicate with CPU tests), and the loader keeps the weight in its native precision instead of folding it into NVFP4. The row-wise weights land in **dedicated fields** (`qkvz_fp8w_rowwise`, `out_proj_fp8w_rowwise`) rather than reusing `qkvz_fp8w` — that field feeds the block-scaled `w8a16_gemv`, and handing it a `[N,1]` scale would be a silent numerical error, so the setter asserts the format.

**2. A one-line GEMM fix under the BF16 lever.** `ATLAS_GDN_BF16_WEIGHTS` already existed and was a **−72.9% prefill regression**, which is why nobody set it. The cause was not the precision — it was the GEMM: the `force_bf16` arm dispatched to `dense_gemm` instead of `cublas_bf16_proj_dense`. Routed through cuBLASLt, the same flag reads **+11.9%**. A flag that was a footgun is now a win.

**3. `dispatch_proj.rs` back under the LoC cap** (538 → 314 + 245), split by cohesion: the passthrough decision, the cached block→row-wise requant it guards, the router over both, and their CPU tests move together into `dispatch_proj_rowwise.rs`. Behaviour-identical.

Plus the A/B harnesses (`scripts/fp8_rowwise_ab.py`, `prefill_probe.py`, `decode_probe.py`), `docs/fp8-rowwise-mixed-precision.md` with the controls and limitations, and `/measurement-discipline` (picked up from #522 because it is load-bearing for the claims below — and it does not flatter them).

## What the measurements actually say

Observed on GB10, qwen3.8-27B-NVFP4, prefill at 2K ISL:

| lever | prefill vs baseline |
|---|---|
| `ATLAS_FP8_ROWWISE` fold (this PR's headline) | **+15.5%** |
| `ATLAS_GDN_BF16_WEIGHTS` + the cuBLASLt fix | **+11.9%** |
| `ATLAS_GDN_BF16_WEIGHTS` before the fix | −72.9% |
| **GDN stack + CUTLASS NVFP4 dispatch flags** | **750 → 1333 tok/s (+78%)** |

The honest headline is the last row: **the dispatch flags dwarf the weight-precision work.** The double-quant fix is real and worth having — it is a correctness improvement that also happens to be faster — but anyone chasing prefill on this family should set the flags first. That is why `83c1aeb5` is titled the way it is.

Two cautions that belong with the numbers rather than under them:

* **The cuBLASLt FP8 row-wise GEMM path is dead on sm_121.** The first attempt wired per-row FP8 straight into `cublas_fp8_rowwise_proj` and got status 15. Padding M (a genuine bug, fixed here) did not save it; a control proved the GEMM itself is unsupported on this box. The fold therefore goes **via BF16**, which is why `92fad2d4` says "land the fold via BF16".
* **FlashInfer-GDN fails *open*.** Both times I measured the flag row, it had silently fallen back — a script had reset `LD_LIBRARY_PATH`, dropping the CUTLASS DSL directory — and the only way to catch it was grepping the boot log for `FlashInfer GDN kernel loaded`. It is worth ~17% at C=4 while looking enabled. Verify engagement before believing any GDN number.

## ⚠ The decode numbers in this PR are stale

Everything decode-side quoted in the comments below (5.26 baseline, 5.08, ~5.3) was measured **before** the decode-GEMV LUT-staging repair, which is now on `main`. On the repaired kernel the same probe reads **12.9–13.9 tok/s**, and **25.1** with MTP K=4. The row-wise fold's only remaining advantage over the cheaper BF16+cuBLASLt lever is that it holds decode at full speed — and that comparison was made on broken decode, so **it could flip**. Prefill numbers are unaffected. Re-cutting the decode A/B on the repaired kernel, with `reasoning_effort:"none"` and `ATLAS_MTP_ACCEPT_DEBUG=1` and run-record IDs, is the one piece of work left before this is a merge candidate.

## Gates

`vision-fidelity` re-recorded on this head for qwen3.8-27B: **PASS — 14/14 geometry cells, 3/3 probes, control held**. The remaining required gates need re-cutting on this head.

`video-fidelity` on qwen3.8-27B came back **12/13** — and chasing that one cell turned up something that is not about this PR at all, so it is flagged here rather than buried:

> **Speculative decode changes the greedy output on qwen3.8-27B.** The `video-before-image` cell asks for a clip's 4 colors and gets 3. Same recipe, same `--serve-override` path, one variable:
> | leg | result |
> |---|---|
> | `speculative=true num_drafts=3` (as the recipe pins it) | FAILED — 3 runs, identical output |
> | `speculative=false num_drafts=0` | **PASS 13/13** |
> | `speculative=true` + `ATLAS_NO_MTP_DRAFTER_CONTEXT=1` | FAILED |
>
> At temperature 0 speculative decode should be output-identical to plain decode — verify accepts a draft only where it matches the target's argmax — so this is a **verify-path divergence**, present under the recipe's own pinned settings. The third leg rules out the drafter's prefill/carry context. The sampler is not a variable either: at temp 0 Atlas applies penalties + logit_bias then argmax, bypassing top_k/top_p/min_p/top_n_sigma.
>
> Ruled out along the way: the media pipeline (the failing request carries 288 prompt tokens / 2 media items / 245 merged patches, with the still's pad run correctly *after* the video's — both items arrive in the right order), the `non_thinking` preset's `presence_penalty=1.5` (pp=1.5 and pp=0 answer alike), and `max_tokens` — though that turned up a separate latent defect, now fixed in `aa529ae4`: the cell budgeted 120 tokens for a reply that needs 153 and returned `finish_reason=length`. Raising it to 320 did **not** fix the cell; `a19ff7ac` corrects that claim where I first made it.

The two `crates/atlas-plugin/src/benchmarks/video/driver.rs` commits are self-contained and unrelated to the fp8 work — cherry-pick them out to their own PR whenever that is wanted.


rsafier:
## State of the branch — 2026-08-15

Five commits, all signed, tree clean, workspace green (542 spark-model + 2220 spark-server tests, fmt/clippy/typos 0). **Draft on purpose: nothing here is proposed for default-on, and the quality case is deliberately short of the house standard.**

### What is actually finished

| | status |
|---|---|
| `ATLAS_FP8_ROWWISE=1` loader + prefill arm | works, opt-in, gated, tested |
| `ATLAS_GDN_BF16_WEIGHTS` GEMM fix (`dense_gemm` → cuBLASLt) | works, −72.9% becomes +11.9% |
| `rowwise_pair_passthrough` + M-padding fix in `cublas_fp8_rowwise_proj` | landed, no-op for existing callers |
| Prefill/decode A/B harness (`scripts/{prefill,decode}_probe.py`, `scripts/fp8_rowwise_ab.py`) | reusable |
| Flag A/B on qwen3.8-27b | measured, table in the PR body |

### What I would NOT merge on

- **No BFCL.** The quality argument is the video benchmark's dark-green sensitivity probe plus #257's recorded 7-point history. #514/#495 set the house bar for accuracy changes — BFCL with ratcheted floors and a measured ±0.40 band — and it has not been run on any of this.
- **Decode under CUTLASS is unmeasured.** That run was killed to free the box. There is no number and I have not guessed one.
- **The vision/video gates have not been run on the combined flag stack** — only on the weight-precision flags, where they pass 14/14 + 3/3 and 13/13, control held.
- **The agentic result is a smoke test, not the gate.** 1/1 webserver_ok, 1/1 followed_directions, Σwall 195 s, 6/6 steps over 11 turns / 13 tool calls, run from the TUI on the GDN+CUTLASS serve. `iterations = 1`; the gate pins exactly 10. Per-iteration wall matches the recipe's 2026-08-14 reference (1925 s / 10 ≈ 192 s), so the flags cost nothing agentically — but that is one iteration.
- **qwen3.6-27b is untested** with any of this.

### The recommendation, plainly

**The dispatch flags are the story, not this branch's weight work.** GDN + CUTLASS is +75% prefill at C=4 with no memory cost, versus +11.9%/+15.5% from the precision work, and nothing in the repo sets them. If only one thing here gets attention, make it that.

Between the two precision levers, **`ATLAS_GDN_BF16_WEIGHTS` + the cuBLASLt fix is the better default candidate**, not the row-wise fold: same precision, +11.9% prefill, and it *replaces* the NVFP4 copy so it costs 2.9 GB less than baseline and buys 24% more KV headroom. The row-wise fold gets +15.5% and keeps decode at full speed, but holds three copies of every GDN projection — measured +16 GB, 3 GB free on a 121 GB box. It only wins outright once a per-row FP8 GEMM exists on sm_121.

### Two things a reviewer should know before reading the numbers

1. **Every measurement is on the unflagged floor.** No `ATLAS_*` dispatch flag was set for the precision A/Bs — but the qwen3.6-27b and qwen3.8-27b serving recipes set none either, so it is the config the gates and production run. Like-for-like, but the floor. The ranking above has not been re-established on top of the GDN+CUTLASS stack.
2. **One result is unexplained and I have not papered over it.** The sensitivity probe *improved* under CUTLASS NVFP4 (`Red, Blue` → `red, blue, yellow`) although it consumes the same double-quantised weights. The probe is a coarse oracle; that wants the real gates before anyone reads a quality claim into it.

### Structural question for review

This is the **fourth** instance of one defect family — checkpoint ships a module at higher precision, loader quantises it down anyway (#257 merged, #484 and #406 open). The general fix means converting 55 NVFP4-typed layer fields, ~49 funnel call sites, and re-validating every model family; with #400/#404/#474/#475/#484/#406 open in those same loaders that is a real merge hazard. I built the fourth arm because it produces the number that would justify the refactor — but it is a reasonable review outcome to ask for the general form instead.

`docs/fp8-rowwise-mixed-precision.md` carries the method, the controls that killed each hypothesis, and the harness's own limitations (KL is only meaningful to the first divergence on free-running generations).

rsafier:
## Update — rebased, plus two additions

Head is now `1e342382` (10 commits). Since the state-of-the-branch comment above:

**Rebased twice, both verified locally before pushing.** #513 was squash-merged, so #516 was rebased `--onto main` (22 commits, zero conflicts) and this branch was replayed on the new tip. Build clean, **2220 spark-server + 542 spark-model tests pass**, fmt/clippy/typos clean.

**`dispatch_proj.rs` is off the LoC violation list** (`1e342382`). It was 445 before this branch and 538 after — the one violation this PR owned, and CI was red on it. Split by cohesion rather than line count: the passthrough decision, the cached block→row-wise requant it guards, the router that consumes both, and their CPU-only tests move together into `dispatch_proj_rowwise.rs`. 538 → 314 + 245. Behaviour-identical; the two passthrough tests only change module path.

The six LoC violations that remain are all inherited from #516 (`video/driver.rs` 917, `vision/driver.rs` 847, `media_integrity.rs` 692, `http.rs` 616, `msg_entry.rs` 584, `chat/mod.rs` 504) and are that PR's deferred decision, not this one's.

**Picked up `/measurement-discipline` from #522** (`56a4d655`), because it is directly load-bearing for this PR's claims — and it does not flatter them. Three of its rules land on measurements quoted here:

* It names `chat_template_kwargs.enable_thinking:false` as the *wrong* lever for a decode headline (`reasoning_effort:"none"` is the right one). This PR's decode probe uses the former.
* Rule 11 says no perf number enters a PR comment without a fingerprint and its `~/.atlas/runs/` IDs. The tables above describe the serve config in prose and cite no run IDs.
* Rule 10's graduated language: single-harness numbers are "observed X under fingerprint F", not plain X.

## ⚠ The decode numbers in this PR are pre-repair — do not quote them as current

Everything decode-side here (5.26 baseline, 5.08 for BF16+cuBLASLt, ~5.3 for the fold) was measured **before** the decode-GEMV LUT-staging repair (#519 / `b7d67c7e`+`d584c0c5`, not yet on main). On the repaired kernel, the same probe on the same box reads **12.91 tok/s** for the two kernel commits alone and **13.92** on the full `wt/repair-479` — 2.45–2.65×. With MTP K=4 on top, **25.11**.

That matters for one of this PR's conclusions specifically. The row-wise fold's *only* remaining advantage over the cheaper `ATLAS_GDN_BF16_WEIGHTS`+cuBLASLt lever is that it keeps decode at full speed — and that advantage was measured on broken decode. **It could flip.** Prefill numbers are unaffected.

The plan is to rebase once the LUT commits land on main and re-cut the decode A/B with `reasoning_effort:"none"` and `ATLAS_MTP_ACCEPT_DEBUG=1`, citing run IDs — rather than carrying cherry-picks here and recreating the squash-merge collision #516 just went through.

**Also measured since, on qwen3.8-27b, and not yet reflected in the body:** the dispatch flags remain the biggest lever by a distance — prefill at 2K/C=4 goes 750 → **1333 tok/s** with GDN+CUTLASS. Both times I measured that, FlashInfer-GDN silently failed open first (a script reset `LD_LIBRARY_PATH`, dropping the CUTLASS DSL path) and the boot-log line had to be grepped to catch it — worth ~17% at C=4. Any future measurement of that flag should verify `FlashInfer GDN kernel loaded` in the log before believing the number.

tbraun96:
Closing: this PR's content is on `main`. It landed in the #525 stack (`04726b148f`, "Concurrency cycle: #504+#505+#521+#522+#523 stack, …").

Re-checked today against `main` `9e597531`: the row-wise arm and the GEMM fix are there — `an_already_rowwise_weight_passes_through_verbatim` at `crates/spark-model/src/layers/ops/dispatch_proj_rowwise.rs:223`, `a_column_vector_is_rejected_even_with_n_elements` at `crates/spark-model/src/weight_loader/qwen35_dense/rowwise_fp8_tests.rs:70`; 16 of the 17 functions and all 10 files this PR adds are present. The one exception is a test name (`the_real_gdn_projection_shapes_are_per_row`), not production code.

Thanks for the -73% → +12% find — it is on main. No branch deleted.

PR #522, opened 2026-08-15
skills: /measurement-discipline + .agents symlink for cross-agent skill discovery

tbraun96:
Adds the **measurement-discipline** skill: the enforceable checklist distilled from the 2026-08-15 decode-rate flip-flop retrospective. Four phase gates — before measuring (fingerprint, engagement check, prompt class, n≥3), before comparing (one-variable rule), before declaring artifact/no-regression/inert (two-harness rule, roofline-as-hypothesis), before publishing (graduated claim language, commit-message gate, ledger check). Each rule traces to a root cause of the incident or a cited paper (MLPerf 1911.02549, Curie 2502.16069, CoVe 2309.11495, calibration/debate-sycophancy literature).

Also:
- **`.agents` → `.claude` symlink** so non-Claude agent frameworks discover the same skills without a parallel copy (SSOT).
- **`.gitignore` fix**: `.claude/` was ignored wholesale, forcing skills to be force-added (atlas-release was). Narrowed to `.claude/*` + `!.claude/skills/` so repo skills are tracked first-class.

Docs/config only — no code paths touched.


tbraun96:
### Superseded — closing rather than rebasing

This PR shows `DIRTY`, but the conflict isn't real work to resolve: **its content is already on `main`.**

Rebasing `feat/measurement-discipline-skill` onto current main leaves **zero commits ahead** — the rebased HEAD resolves to `048dc6bb5`, which *is* `origin/main`. Both deliverables are present:

- `.claude/skills/measurement-discipline/SKILL.md` — on main, same skill, same description ("Born from the 2026-08-15 decode-rate flip-flop…").
- `.agents` — on main as a symlink (`120000` blob).

The single conflict during rebase was in `kernels/gb10/qwen3.6-35b-a3b/BENCH.toml`, and it was *stale-on-this-branch* rather than divergent: the branch carried the older `--serve-override disable_thinking=true` phrasing, while main describes the `[benchmarks.serve_overrides]` block that the file actually declares below. Main's wording is the correct one.

Nothing here needs to land. Closing to keep the open-PR list honest — reopen if I've misread and some part of this branch is genuinely absent from main.

tbraun96:
Superseded: content already on main (rebase yields zero commits). See the analysis comment above.

Issue #517, opened 2026-08-14
Flag-interaction bugs: options individually fine, jointly wrong (batch4 leftover re-audit)


tbraun96:
Re-audit of the deferred batch4 "latent or env-var-gated medium-severity bugs" cluster. The original hint predicted flag-interaction bugs where two options are individually fine and jointly wrong — that prediction **matched the code emphatically**.

A recurring root cause runs through several of these: **sentinel-default collision** — a clap default that is also a legal explicit value, so "equals the default" gets read as "not given." That is the anti-pattern `serve_flags.rs` itself documents as forbidden.

Fixes for the first three are parked on branch **`wt/batch4-leftovers`** (`549c7173f`, not a PR). Each has a test verified to **fail on `origin/main`** by ablation. They are deliberately *not* in #513: that branch's benchmark evidence was measured at a frozen sha, and adding code would invalidate it and owe another 3.5-hour BFCL run.

---

## Fixed on the branch

**1. `--dflash` × per-request `temperature > 0` — sampling silently inert, including under the documented escape hatch.** All four verify sites take raw argmax under `--dflash`. The escape hatch `ATLAS_DFLASH_MASKED_VERIFY=1` routes to `verify_pick_all_with_pipeline`, whose first action is `fast_masked::try_chat_fast_path` — which checked grammar / AdaDec / forced-stage / penalty gates but **never `a.temperature`**, returning argmax before the sampling branch. Its grammar sibling 45 lines below has carried exactly that gate since P1-3 (2026-07-09, "load-bearing for verify-time sampling"); the chat fast path landed 2026-07-08 and missed it. Observable: plain-chat requests at `temperature=1.2` decode deterministically while tool requests on the same server sample correctly.

**2. `--disable-template-overrides` was a no-op on the served prompt** for any model with a `jinja-templates/openai/` variant — including the flagship `qwen3_5_moe` family and `minimax_m2`, whose variants are not behavior-equivalent to base. The flag gated the base override at `chat_impl.rs:88`, then `load_openai_template` loaded unconditionally from the same directory 20 lines later, and the openai variant wins on every serving route. `/v1/tokenize` *did* honor the flag — so the tokenize endpoint reported a different prompt than the model actually received.

[...]

Issue #472, opened 2026-08-12
ATLAS_SSM_TAIL_MIDCHUNK=1 returns empty 1-token completions on shared-prefix (BFCL) runs


AzeezIsh:
Running a BFCL single-turn subset against `nvidia/Qwen3.6-27B-NVFP4` with
`ATLAS_SSM_TAIL_MIDCHUNK=1` and `--enable-prefix-caching`, the server returns
**empty 1-token completions** for most entries. Setting the flag to `0` fixes it
completely, on otherwise identical config.

This looks like the same failure `scripts/mlperf-edge/ab_midchunk.sh` documents
in its header — cross-request SSM prefix reuse, where "BFCL single-turn requests
share a system-prompt prefix and reused each other's tail snapshot, inheriting
state that bleeds past the advertised prefix boundary -> garbled tool calls ->
wrong scores". That script says the underlying bug was fixed and that
`radix_tree/snapshot.rs::lookup` now carries the prescribed session gate, so
either the gate does not cover this path or there is a second route to it.

## Observed

Same 196-entry seeded subset, same build, only the flag differs.

`ATLAS_SSM_TAIL_MIDCHUNK=1` — `"result": ""`, `output_token_count: 1`:

| category | n | empty |
|---|---:|---:|
| live_multiple | 12 | **12** |
| live_parallel | 8 | **8** |
| live_relevance | 6 | **6** |
| live_simple | 12 | 9 |
| irrelevance | 30 | 22 |
| multiple | 30 | 21 |

`ATLAS_SSM_TAIL_MIDCHUNK=0` — **zero** empty completions in all 10 categories,
and the run scores sensibly (`simple` 1.000, `live_multiple` 0.833, weighted
148/196).

Latency on the bad responses was ~740 ms, i.e. faster than a normal prefill for
those prompt lengths, consistent with returning immediately rather than
generating.

## Config

`--max-seq-len 65536 --gpu-memory-utilization 0.80 --kv-cache-dtype bf16
--max-batch-size 1 --speculative --num-drafts 2 --mtp-quantization bf16
--disable-tool-grammar true --enable-prefix-caching --ssm-cache-slots 64
--ssm-checkpoint-interval 16`, with `ATLAS_W4A16_DP4A=1`,
`ATLAS_FORCE_GLOBAL_GDN=1`, `ATLAS_SSM_TAIL_PROTECT=1`,
`ATLAS_SSM_TAIL_LEASE_TTL=128`.

[...]

PR #523, opened 2026-08-15
fix: think-gate, #517 tool-think cap, and hybrid prefix hit (#472)

facepunch47:
## Summary

Three serve lies found and proved on Spark `:8888` with `unsloth/Qwen3.8-27B-NVFP4`. Same binary stack. Does not ship fast-and-dumber.

1. **Think-gate** — thinking-on was forced serial (`step_decode_only`). Same Fibonacci prompt: 303 tok @ **6.2 tok/s** → 884 tok @ **22.6 tok/s**, `serial=0`, `mtp=1.00`.
2. **#517** — 90% think cap used raw client `max_tokens` before `--tool-max-tokens` shrink. Repro `--tool-max-tokens 256`, `max_tokens=2048`: budget **1843**, `finish=length`, no tool call, 255 reasoning. After hoist `generation_max_tokens`: budget **230**, API returned `get_weather`.
3. **#472 hybrid prefix** — KV radix hit without a restorable SSM snapshot was logged as a hit, then all KV recomputed. Turn 2: `cached=98`, TTFT still 325 ms, log `hit N but no SSM snapshot — recomputing all KV`. After `lookup_paired`: long shared prefix 1254→1258 tok, **1232 skipped**, TTFT **1737 → 159 ms**, `Marconi SSM cache hit`. Extra commit: F82 attention-only skip must not fire on hybrid when the exact-leaf bypass refuses restore (that path answered `system:` instead of `Paris`).

## Test plan

- [x] Thinking-on Fibonacci on `:8888` after think-gate: ~22 tok/s, `serial=0`
- [x] Tools request with `--tool-max-tokens 256` and `max_tokens=2048`: think budget ≤ 230, tool call returns
- [x] Identical short prompt twice: still `Paris` (exact-leaf recomputes; no KV-only skip)
- [x] Shared prefix >256 tok, new user turn: Marconi skip log, TTFT collapse, non-garbage answer

Compare: https://github.com/Avarok-Cybersecurity/atlas/compare/main...facepunch47:cursor/qwen38-three-lies?expand=1


tbraun96:
## Bisect evidence: this PR's content reproduces an agentic-webserver regression on the dense 27B — pulled from the #525 bundle, left open for a fixed re-land

An 11-run bisect on dgx1 (same box, same day, same harness, one variable per round) isolated this PR's content as the cause of a reproducible multi-turn correctness regression:

| Tree under test | agentic-webserver (N=10) | Verdict |
|---|---|---|
| pre-bundle control binary (×3) | 10/10 ws · 10/10 directions | PASS |
| main + #505 content | 10/10 · 10/10 | PASS |
| main + #522 content | 10/10 · 10/10 | PASS |
| **main + #523 content only** (`git diff 468116406..a7069fe89` applied) | **9/10 ws · 8/10 directions** | **FAIL** |
| full #525 bundle (×5, incl. minus-#521 and minus-width-fix variants) | 9/10 · 7-8/10 | FAIL |

Perfect separation: 5 passes without this content, 6 fails with it. The failure signature is multi-turn state corruption — dropped `tore_down`/`ran_tests` steps, and one catastrophic run that wrote **zero src files across 3 turns**. Independently corroborating: bfcl-subset-echolp on the 35B hybrid dipped 86.45 → 85.96 in the same window (same hybrid prefix path, different box/model/harness).

**The intent here is right** — "treat unpaired KV prefix as a miss, not a lying hit" and skipping the attention-only prefix shortcut on hybrids close real cache-poisoning classes. But the hybrid-path implementation evidently corrupts state of its own. Suggested next step: sub-bisect the 5 commits (e7ca57871, ea4258bab, 5d313174a, 790e15c47, +1) with `spark benchmark run agentic-webserver --pull-request-gate` (~16 min/round on dgx1); the new `ssm-state-poisoning-gate` landing in #525 should also give a cheaper 10-min repro probe.

Repro fingerprint: dgx1, `unsloth/Qwen3.8-27B-NVFP4` agentic recipe (lm-head bf16, batch 1, ssm-slots 256), bisect logs in the session scratchpad (`bisect{1..7}-agentic.log`).

tbraun96:
## CORRECTION to the bisect verdict above: the SSM prefix-cache stack is EXONERATED

Follow-up splitting of the merge delta refutes my earlier attribution. The merge's 25-file first-parent delta divides into the 11 SSM prefix-cache/hybrid-hit files and 14 other files (the #517 tool-think-cap SSOT, verify-during-think, mtp accept-stats, gate hysteresis). Run separately on main:

| Tree | agentic-webserver | In-think verify events |
|---|---|---|
| main + **11 SSM files only** | **PASS 10/10 · 10/10** | 0 |
| main + **the other 14 files only** | **FAIL 8/10 · 6/10** | 80 |

The culprit is the `spec_dispatch_eligible` change that lets standard MTP speculate inside `<think>` by default — batch-K verify is not byte-lossless at T=0, and mid-reasoning token flips compound deterministically (byte-identical failure signatures across runs). Fixed in the #525 stack by `2c80bf2f7`, which restores the serial-in-think contract (both lanes require `ATLAS_DFLASH_SPEC_THINK=1`) while keeping the SSOT refactor, accept-stats, and hysteresis.

**Consequence for this PR:** the SSM correctness fixes here are clean on the agentic gate and re-landable as-is once rebased. One caveat remains open: a bfcl-subset-echolp dip (86.45→85.96 on the 35B) observed on bundle content is NOT explained by the in-think mechanism (those legs serve thinking-off) — a discriminating echolp run on an SSM-free tree is in flight and I'll report whether this PR is implicated or the dip was noise/hysteresis.

tbraun96:
Closing — not as landed. This PR went into the #525 stack and was **semantically reverted there** (commit `04726b148f`, section "Revert 'Merge branch pr523 into stack/concurrency-cycle'"), naming this PR's commits including its head `790e15c47` ("do not take the attention-only prefix skip on hybrid models").

The reason recorded in that commit: an 11-run agentic-webserver bisect isolated #523 as the cause of a reproducible correctness regression — `main`+#523 alone failed 9/10 `webserver_ok` and 8/10 `followed_directions` (one run wrote no `src/` in 3 turns) while `main`, `main`+#505, `main`+#522 and the pre-bundle control all passed 10/10 · 10/10, and `bfcl-subset-echolp` on the 35B dipped 86.45 → 85.96 in the same window. The intent (the unpaired-KV-prefix and hybrid-prefix cache-poisoning classes are real) was agreed; the hybrid-path implementation corrupts multi-turn state.

So this should not be re-landed as-is. A re-cut that fixes the hybrid path is welcome, and it needs that webserver bisect passing beside it. No branch deleted.

PR #536, opened 2026-08-16
perf(graphs): drain-tail graph borrowing — kill the 32→1 re-capture cascade (+~2% at C=32)

tbraun96:
## Measured evidence

dgx2, Qwen3.8-27B NVFP4, C=32 ladder rung, binary d92fc2488 (serve log `dgx2:/home/claude/diag_c32_serve.log`): as the wave drains 32→1, the engine re-captures a CUDA graph for **every shrinking composition** — **29 captures observed (n=31→1), ~7.5 s of capture work, ~2.9 s of wall beyond the median finisher**. At 32,600 tokens / ~150 s wall, recovering ~2.9 s ≈ **+2% aggregate** — the difference between losing the rung to vLLM (218.4 vs 219.5) and winning it.

## Root cause

Both batched graph caches key on the per-row **SSM slot vector** (a capture bakes each row's pool state addresses — `decode_graph_key.rs`, `verify_e2.rs`). `padded_batch_n` buckets the *width*, but the ramp admits the whole wave at once and never visits n=31,30,…, so every drain composition is a brand-new key even inside an already-captured bucket. The LRU caps (48/32) prevent thrash but cannot prevent first-sight captures.

## Design: borrow a wider captured graph (width-bucketing via prefix reuse)

`retire_finished_sequences` already compacts survivors into contiguous slots `[0..n)`, and both dispatch paths sort by slot (`decode_step.rs:33`, `mtp_step.rs:422`) — so a drain batch's slot vector is a **prefix** of the steady-state canonical vector. New `spark-model/src/model/trait_impl/graph_borrow.rs`: on an exact-key miss, replay a **wider** captured graph whose key starts with this batch's slots. Active rows sit at exactly the rows the scheduler reads (decode logits row `i`; verify rows `off[i]..off[i]+k` — the verify prefix must match `(slot, k)` **pairs**, so row bases are unchanged); tail rows become padding.

**Safety (the load-bearing part):**
- A borrow is legal only if every tail-baked slot is the dummy slot or a **currently-free** pool slot (`SsmStatePool::slot_is_free`; scheduler thread is the only claimer/releaser). Pad lanes then only scribble on unowned pool state, which `alloc_sequence_dispatch` zeroes on the next claim; MTP intermediates/checkpoints are per-verify-step scratch. A slot held by a mid-prefill sequence **vetoes** the borrow.
- Verify ghost rows get pad metadata (dummy KV, pos 0, clamp 1 — the decode pad shape), pad embeds, and WY table entries **synthesized from the pool** (a captured entry is a pure function of `(layer, slot)`, reproducing the capture-time pointers byte-for-byte); ghost slot tiers must cover the baked depth (`h_inter_count`). The 96-row cap is re-checked as a hard borrow veto because it guards the unsafe upload lengths.
- Determinism: pad rows already exist at every `n < padded_n` today; borrowing only widens the tail, and the batched kernels are row-independent — active lanes' outputs cannot depend on pad lanes.
- Capture behavior is **unchanged** — borrowing is replay-only, so no new graph shapes are ever minted.

**Policy:** borrow at most **2× the batch's own `padded_batch_n` bucket** (factor-2 window mirroring the `mtp_gate` power-of-two width-regime economics), narrowest candidate first. A full 32→1 drain then captures ~2 graphs per family instead of ~29 — and those captures are canonical, so later drains reuse them outright.

**Arbiter honesty:** the model layer hides the padded width entirely; decode steps stay charged at `active.len()` (pinned by a source-scan test).

**Kill switch:** `ATLAS_NO_GRAPH_BORROW` (presence disables, house convention).

## Tests

- `graph_borrow.rs`: 12 pure-function tests — the width→bucket policy pinned; the drain scenario pinned (steady-state 32-key serves every width in its window, **no re-capture**; below the window declines); claimed-tail-slot veto; non-prefix/bootstrap-subset rejection; narrowest-candidate preference; verify `(slot,k)` + sentinel matching, tier-aware tail check, fewest-ghost-rows preference.
- `mtp_gate/tests.rs::arbiter_charges_active_width_never_a_padded_width` — every `gate.record_decode` call site must charge `active.len()`.

## Expected impact

+~2% aggregate at C=32 (the measured ~2.9 s drain tail), and a reduced finish tail at **every** rung — any wave that drains through unseen compositions stops paying per-width captures.

## Gates (verbatim)

```
ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo fmt --all -- --check   # clean
ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy --workspace --tests   # Finished, no warnings
ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo test --workspace   # 5099 passed; 0 failed; 66 ignored
```

## GPU validation plan (before merge)

CPU-only session — the following must run on a GB10 box:
1. **C=32 ×3 reps** on Qwen3.8-27B NVFP4: confirm the drain-phase capture cascade is gone from the serve log (expect ≤ ~4 "Captured CUDA graph" lines after the wave peak instead of ~29), wall recovers ~2-3 s/run, emitted-text SHA unchanged vs `ATLAS_NO_GRAPH_BORROW=1` control (temp 0.0 / seed 42).
2. **Full C=1..16 floor re-check** per the no-regression rule — borrowing must never fire mid-steady-state (only on drain compositions), and per-rung tok/s must be ≥ baseline.
3. Watch for: pad-lane interference (any output diff vs kill-switch control run), `verify graph borrow`/`decode graph borrow` debug lines firing while prefills are in flight (freeness veto should prevent this), and LRU churn (borrowed keys are touched, so steady keys must not evict).

### DIFF
--- a/crates/spark-model/src/model/ssm_pool.rs
+++ b/crates/spark-model/src/model/ssm_pool.rs
@@ -378,6 +378,15 @@ impl SsmStatePool {
         }
     }
 
+    /// True when `idx` is currently on the free list — claimable, owned by no
+    /// live sequence. Drain-tail graph borrowing (`graph_borrow.rs`) uses this
+    /// to prove a borrowed graph's tail rows only scribble on unowned state.
+    /// The scheduler thread is the only claimer/releaser, so the answer is
+    /// stable for the duration of a dispatch it performs itself.
+    pub(super) fn slot_is_free(&self, idx: usize) -> bool {
+        self.free_slots.lock().contains(&idx)
+    }
+
     /// Reserved pool slot used by `decode_batch` / `mixed_forward` padding.
     /// Never claimed by `claim_slot()`, never released. SSM kernels are
     /// free to read/write this slot's pool memory without affecting any
--- a/crates/spark-model/src/model/trait_impl/decode_a2.rs
+++ b/crates/spark-model/src/model/trait_impl/decode_a2.rs
@@ -274,10 +274,15 @@ impl TransformerModel {
                             .expect("borrowed key comes from this cache");
                     e.1 = tick;
                     replay = Some(e.0);
-                    tracing::debug!(
-                        "decode graph borrow: n={n} padded_n={padded_n} -> replaying \
-                         captured width {dispatch_n}"
-                    );
+                    // INFO once per transition (same cardinality as the
+                    // captures this replaces); repeats of the same pair
+                    // stay silent. Provable engagement: grep "graph borrow".
+                    if super::graph_borrow::DECODE_BORROW_LOG.should_log(key, &bk) {
+                        tracing::info!(
+                            "decode graph borrow: n={n} padded_n={padded_n} -> replaying \
+                             captured width {dispatch_n}"
+                        );
+                    }
                 }
             }
         }
--- a/crates/spark-model/src/model/trait_impl/decode_a2.rs
+++ b/crates/spark-model/src/model/trait_impl/decode_a2.rs
@@ -210,37 +210,6 @@ impl TransformerModel {
         // `traits::padded_batch_n` (now includes 12 and 16 for the C-sweep).
         let padded_n = crate::traits::padded_batch_n(n);
 
-        // ── Phase 1: Pre-graph (runs every step, NOT captured) ──
-
-        // 1a. Embed active tokens into hidden[0..n)
-        for (i, &tok) in tokens.iter().enumerate() {
-            self.embed(tok, hidden.offset(i * h * fp32), stream)?;
-        }
-
-        // 1b. Zero padding hidden[n..padded_n)
-        for i in n..padded_n {
-            self.gpu.memset(hidden.offset(i * h * fp32), 0, h * fp32)?;
-        }
-
-        // 1c. Allocate KV blocks for active sequences
-        let mut kv_cache = self.kv_cache.lock();
-        let bs = kv_cache.block_size();
-        for seq in seqs.iter_mut() {
-            let blocks_needed = (seq.seq_len / bs) + 1;
-            ensure_blocks_through_decode(
-                seq,
-                blocks_needed - 1,
-                &mut kv_cache,
-                self.prefix_cache.as_ref(),
-                self.gpu.as_ref(),
-                stream,
-                self.levers.kv_poison,
-            )?;
-        }
-
-        // 1d. Upload metadata with fixed stride (active + padding)
-        let metadata = self.upload_batch_metadata_fixed(seqs, padded_n, &mut kv_cache, stream)?;
-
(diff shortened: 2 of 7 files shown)


tbraun96:
## Validation debug (dgx2 run of 0f93ce37f) — fixed in bdfbd1958

**What the log actually shows** (`/tmp/atlas-dtval-serve.log`, 33 capture lines total): the borrow **did engage** for the n=31..13 drain band — that band has *zero* captures (the original cascade had ~19 there). It was unprovable because the borrow lines were `tracing::debug!`, invisible at the serve log's INFO level (`grep -i borrow` = 0 hits was a logging artifact, not non-engagement).

**The real bug (the n=12 counter-example):** `borrow_width_cap(12) = 2*padded_batch_n(12) = 24 < 32` — 12 is itself a padding-ladder rung, so its "bucket" was 12, not 16. Policy alone rejected the clean n=12-under-n=32 prefix match, forcing the 21:04:56 capture. Fixed: cap is now `2 * n.next_power_of_two()` (the mtp_gate width-regime bucket). Regression test `dgx2_n12_under_n32_log_keys_borrow` carries the literal key vectors from the log and fails on the old cap.

**Why wall didn't move (follow-up, not this PR):** each rep's drain burst (~1.3-1.7 s at 21:04/21:07/21:09/21:12) is dominated by **small-n ragged D-Cut shapes** (n=2..7, ks mixing 4/3/2, depth-major order) — per-seq depth assignment makes those keys non-canonical permutations that prefix borrowing cannot serve, and they re-mint every wave. Fixing that means canonicalizing D-Cut's depth-to-slot assignment (behavioral change, separate PR). Also present once per serve: 7 single-seq `verify_b` K=2 per-slot captures (orthogonal cache).

**Next GPU validation should expect** (C=32 ×3, same driver):
- `grep 'graph borrow' serve.log` **> 0** — INFO lines now fire once per borrow transition; expect ~20+ per first drain (one per width in 31..9 across verify+decode), then per-transition on later reps.
- Capture count: the n=12 verify capture line **gone** (borrowed); remaining captures ≈ initial n=32 pair + single-seq verify_b/c2 slot graphs + the small-n ragged D-Cut bursts (unchanged by design).
- Wall: recovery bounded by the n=12 capture (~0.2-0.3 s/serve) until the ragged-tail follow-up lands — do not expect the full +2 tok/s from this fix alone; the measurable claim is engagement (borrow lines + missing n=12 capture), plus no regression on the C=1..16 floor.

Gates: fmt clean, clippy clean, `cargo test --workspace` 5101 passed / 0 failed / 66 ignored.

tbraun96:
Closing as included: this PR's content is merged into #572, which merges the certified tree as one unit. Merging these separately would invalidate the benchmark records at each step (any crates/ diff invalidates the closure hash), so the whole stack goes in together with its gate certification intact.

PR #538, opened 2026-08-16
perf(prefill): varlen batched prefill waves — concatenate queued prompts into large-M dispatches

tbraun96:
## Measured problem (two independent GPU studies, 2026-08-16, Qwen3.8-27B NVFP4 / GB10)

Prefill runs **one request per dispatch** at ~285 ms per 200-token prompt (~620-745 tok/s aggregate), strictly serial. A C=128 wave pays a **38.7 s zero-output ramp** (11% of wall); C=32 pays ~10-11 s — the margin of the lost C=32 rung. `ATLAS_PREFILL_CODISPATCH=1` overlaps streams but every layer GEMM still launches at M≈200 per request; nsys shows the GPU **96.6% busy at M=280-shaped kernels running 1.5-3× below their own large-M throughput** (profile artifacts: `dgx2:/home/claude/prof_prefill_20260816/`). vLLM prefills the same checkpoint at ~2.9k tok/s. The prefill budget (2048) already allows ~10 such prompts per step, yet dispatch stayed 1-at-a-time.

## Root causes

The model-side varlen machinery (cu_seqlens-packed hidden states, one outer layer loop, large-M projection/FFN GEMMs over Σ tokens, batched GDN with varlen FLA arm, cu_seqlens-aware batched paged attention) **already exists** behind `ATLAS_PREFILL_VARLEN`. It never fired in production because:

1. **Admission** — chunk-0 ran inline in `start_new_requests` unless codispatch deferred it, and codispatch only defers when ≥2 requests co-arrive in *one tick* with nothing in flight **and all prompts are the same length**. Real waves arrive across ticks with ragged lengths → every prompt prefilled serially.
2. **Dispatch** — `run_batched_prefill_step` handed **all** prefilling streams to one `prefill_batch_chunk` call. Σ tokens over the hidden-buffer arena made `check_kernel_batched_eligible` reject the batch **wholesale**, and the fallback is the fully serial per-stream loop. A C=32 wave (6400 tokens) could never batch.
3. **NVFP4 BR64 attention arm missing** — `(KvCacheDtype::Nvfp4, use_br64)` fell into the catch-all bail even though the kernel exists and its handle is loaded; any admitted wave with max chunk ≥ 256 on an NVFP4-KV serve would fail its whole cohort mid-forward.
4. **Chunk-0 paged upload keyed on codispatch only** — a chunk-0 wave admitted by varlen alone staged NULL block tables for the batched paged attention kernel.

## What this PR does

- **`--prefill-varlen-batch`** (default **OFF**, PCND-explicit; legacy `ATLAS_PREFILL_VARLEN` stays as env fallback; absent flag publishes nothing). One `OnceLock` cell feeds the model admission predicate, the batched-attention chunk-0 guard, and the scheduler wave planner.
- **Scheduler wave planning** (`phase_continue_prefills/prefill_waves.rs`): pure first-fit-greedy partitioner. Each wave shares `(chunk_start, is_last_chunk)` — the model-side admission contract; varlen lifts only equal-`chunk_len` — with **Σ chunk_len capped at `min(--max-prefill-tokens, arena)`**. Waves run back-to-back in one tick, so every stream still advances one chunk per tick; each wave samples its finishers before the next wave reuses the logits rows; a model error fails only its own wave. Flag OFF ⇒ one wave with every stream — **byte-identical dispatch behavior**, pinned by test.
- **Chunk-0 deferral**: with the flag on, `start_new_requests` defers chunk-0 whenever there is company to batch with (≥2 co-admitted this tick, or streams already prefilling — late arrivals join the next wave). No equal-length requirement. Solo requests keep the inline chunk-0 and its solo budget; vision/EP excluded per the existing codispatch constraints.
- **NVFP4-KV BR64 arm**: ops binding + dispatch arm for `inferspark_prefill_paged_nvfp4_batched_64` (kernel + handle already existed).
- **Paged first-chunk upload under varlen**: `force_paged_first_chunk` now keys on the same predicate as admission (codispatch OR varlen).

## Bit-compatibility caveat

Same class as the `--ssm-batched-recurrent` certification caveat: batching changes GEMM reduction row counts and kernels are selected on row count, so per-sequence outputs are **not bitwise-identical** to the serial path. The flag defaults OFF so gate recipes are unaffected until certified; the concurrency ladder opts in.

## Expected numbers (GPU validation pending — this branch was built CPU-only)

- Wave M ≈ 2000 for ten 200-token prompts per dispatch (vs M≈200 × 10 serial); projection/FFN GEMMs (78% of prefill time) launch once per layer per wave.
- Ramp: **38.7 s → ~4-6 s @ C=128**, **~10.5 s → ~2 s @ C=32**.

## Validation plan

1. GPU concurrency ladder (C=32/64/128) with `--prefill-varlen-batch` vs without, same box/checkpoint; confirm ramp shrink and no decode regression (warm-TTFT guard, gate C2 smoke first).
2. Gate suite (A/B/D) with the flag **off** — must be untouched (default-off path is byte-identical, pinned by unit test).
3. Per-sequence output equivalence spot-check batched-vs-serial at temp 0 (expected: token-identical in practice, not bitwise; record as certification evidence like the ssm-batched-recurrent runs).

## Tests

- Wave segmentation math pinned (geometry grouping, exactly-once assignment, FIFO order).
- Budget capping pinned (exact cap, no off-by-one; oversized stream → singleton wave).
- Flag-off = single wave with every stream in order (byte-identical dispatch), CLI parse pinning (absent stays `None`, bare flag means on, explicit false expressible).
- Gates at every commit: `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo fmt --all -- --check && cargo clippy --workspace --tests && cargo test --workspace` — 5093 passed / 0 failed at tip.

### DIFF
--- a/crates/spark-model/src/layers/ops/dispatch_helpers.rs
+++ b/crates/spark-model/src/layers/ops/dispatch_helpers.rs
@@ -34,17 +34,32 @@ pub fn prefill_batched_first_chunk_enabled() -> bool {
         .any(|value| bool_value_enabled(value.as_deref()))
 }
 
-/// VARLEN (ragged) batched prefill enabled? (`ATLAS_PREFILL_VARLEN=1`).
+/// The resolved VARLEN batched-prefill decision. One cell, three readers
+/// (admission predicate, batched-attention chunk-0 guard, scheduler wave
+/// planner) — a `OnceLock` so the decision cannot change mid-serve.
+static PREFILL_VARLEN: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
+
+/// Publish the command line's `--prefill-varlen-batch` decision. Returns the
+/// value IN FORCE, which differs from `enabled` when something already
+/// resolved the cell (then the command line did NOT take effect — the caller
+/// warns, mirroring `gdn_flags::set_from_cli`). Absent flag ⇒ never called ⇒
+/// the documented `ATLAS_PREFILL_VARLEN` fallback stays reachable.
+pub fn set_prefill_varlen_from_cli(enabled: bool) -> bool {
+    let _ = PREFILL_VARLEN.set(enabled);
+    *PREFILL_VARLEN.get().expect("just set")
+}
+
+/// VARLEN (ragged) batched prefill enabled? (`--prefill-varlen-batch`,
+/// legacy `ATLAS_PREFILL_VARLEN=1`; default OFF).
 ///
-/// SSOT for both the admission predicate (`check_kernel_batched_eligible`) and
-/// the batched-attention layer's chunk-0 guard. Those two must agree: if
-/// admission accepts a batch the layer then rejects, the bail happens
-/// mid-Phase-A with streams already mutated, and the per-stream fallback
-/// re-runs setup on dirty state.
+/// SSOT for the admission predicate (`check_kernel_batched_eligible`), the
+/// batched-attention layer's chunk-0 guard, and the scheduler's prefill wave
+/// planner. Those must agree: if admission accepts a batch the layer then
+/// rejects, the bail happens mid-Phase-A with streams already mutated, and
+/// the per-stream fallback re-runs setup on dirty state.
 pub fn prefill_varlen_enabled() -> bool {
-    use std::sync::OnceLock;
-    static EN: OnceLock<bool> = OnceLock::new();
-    *EN.get_or_init(|| bool_value_enabled(std::env::var("ATLAS_PREFILL_VARLEN").ok().as_deref()))
+    *PREFILL_VARLEN
+        .get_or_init(|| bool_value_enabled(std::env::var("ATLAS_PREFILL_VARLEN").ok().as_deref()))
 }
 
 fn bool_value_enabled(value: Option<&str>) -> bool {
--- a/crates/spark-model/src/layers/ops/prefill_attn_batched.rs
+++ b/crates/spark-model/src/layers/ops/prefill_attn_batched.rs
@@ -284,3 +284,59 @@ pub fn prefill_attention_paged_nvfp4_batched(
         .arg_u64(data_section_bytes)
         .launch(stream)
 }
+
+/// Batched NVFP4-KV paged prefill attention (BR=64, chunk_len >= 256).
+///
+/// Same argument contract as [`prefill_attention_paged_nvfp4_batched`];
+/// the `_64` kernel sibling is generated from the shared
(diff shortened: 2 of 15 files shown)


tbraun96:
GPU stackval follow-up (fix pushed: `72f686faa`):

**Diagnosis** — the serve log at `/tmp/atlas-stackval-serve.log` shows the wave planner DID engage (4 dispatches per burst, ~2.8 s apart = waves of ~10 serial 285 ms prefills), but every wave hit `Q12 kernel-batched cache plan not admitted → falling back to per-stream`. The blocker was a blanket veto in `prefill_b_reserve_batched_prefix_matches`: `num_ssm_layers() != 0 → None`, which rejects **every** chunk-0 batch on hybrid GDN/SSM checkpoints — i.e. the 27B under test never could reach the fused large-M path, under any flag combination. Not a codispatch-precedence issue: codispatch does not consume the cohort before the planner (confirmed in code and log).

**Fix** (`72f686faa`):
- Hybrid-SSM models now admit **all-cold** reservations (every `matched_tokens == 0`) — state-identical to the cache-inactive admission that always accepted hybrids. Warm hybrid matches still fall back per-stream (the actual v1 hazard). Pure predicate `batched_reserve_hybrid_ssm_ok`, unit-tested.
- `--prefill-varlen-batch` explicitly subsumes the codispatch shared-geometry coercion (skipped under varlen); varlen wins when both are set.
- Engagement logs: INFO wave plan per tick (`Varlen prefill waves: N streams -> W wave(s), M per wave [...]`), INFO `Q12 kernel-batched prefill dispatched (fused large-M)` with n+total_tokens on admission, INFO decline reasons at every reservation-veto site, ineligible-fallback promoted to INFO under varlen.

**M1-hoist audit**: when admitted, per-layer GEMMs launch at M=Σ for this checkpoint's arms (SSM Phase 1 `prefill_qkvz_proj(k=total)` + BA-gates, attention/FFN `prefill_inner(num_tokens=total)`, Phase 3 at total — all via the standard row-count GEMM router incl. NVFP4/FP8-rowwise). Per-request remnants are non-GEMM: conv1d, GDN recurrence for ragged waves without the varlen-FLA arm (uniform waves take the batched GDN kernel), KV writes, finalize.

Gates at tip: fmt clean, clippy clean, 5094 passed / 0 failed.

tbraun96:
Closing as included: this PR's content is merged into #572, which merges the certified tree as one unit. Merging these separately would invalidate the benchmark records at each step (any crates/ diff invalidates the closure hash), so the whole stack goes in together with its gate certification intact.

PR #540, opened 2026-08-16
KV preempt→resume + depth-aware admission: stop shooting in-flight requests at C=128

tbraun96:
## Problem — forensically confirmed on the C=128 ladder (2026-08-15, Qwen3.8-27B / GB10)

KV overcommit: the pool held **102k tokens** against a true demand of **157k** (128 × ~1226 tokens), because admission reserves **prompt-only** blocks (`scheduler/mod.rs`: `blocks_needed = prompt_len / block_size + 1`) and is never re-consulted as sequences decode deeper. When decode exhausted the pool, `decode_step.rs` picked the largest-`block_table` victim and `send_error()`'d it — permanently killing requests that had already generated ~780 tokens.

Measured damage:
- **171 preemption kills across 4 waves**; **~25% of all decode output discarded**
- aggregate **255.9 tok/s vs vLLM 390.4** while raw decode ran at parity (388–399 tok/s)
- **client-facing correctness bug**: `send_error` emits a mid-stream SSE error data frame then ends the stream with **no finish chunk and no usage** (`api/chat_stream/handle_error.rs`) — an **HTTP-200 "success" with silently truncated content**. Benchmark drivers count these as ok (`n_err=0`, `ptok=0`), so the kills were invisible as failures and only surfaced as lost throughput.

The code's own comment admitted the debt: *"Preferable would be swapping the victim out for later resume (`swap_out_sequence`), but that needs the KvSpillManager threaded in."*

## Commit 1 — preempt → resume (never kill)

`decode_batch_with_preemption` (new `scheduler/preempt.rs`) replaces the kill loop:

- **`--swap-space > 0`**: the victim's KV+SSM image spills through the existing `KvSpillManager` (`spill_out_sequence`, factored out of admission-time `swap_out_sequence` as the shared SSOT) and the existing swap-in loop restores it.
- **No swap space (default)**: **requeue with context retention.** GPU blocks are freed — offered to the prefix cache first, exactly the `finish_sequence` ordering, so the retrying batch evicts only what it needs and the resume re-prefill hits the cache for whatever survives — while the whole `ActiveSeq` (sink, sampling state, guards, think/tool state, streamed output) is retained CPU-side. Resume re-prefills `prompt + processed output` on a fresh sequence; `last_token` stays the pending decode input, so **nothing is re-sampled or re-emitted and the stream continues where it paused**. No error frame, no broken finish contract.

**SSM state on resume (explicit choice):** the requeue path *rebuilds* the recurrent state by re-prefilling the full history through the SSM layers — the always-correct live path. Prefill's own prefix-cache/Marconi lookups skip the recompute where preempt-time cached blocks or decode-time Marconi checkpoints survive; no SSM snapshot is carried across the preemption (the GPU-resident rollback ring dies with the slot and restarts empty, identical to the disk-swap resume path).

**Victim policy (changed, justified in-code):** least progress (fewest generated tokens) instead of largest `block_table`. Largest-first maximized blocks freed *per kill* — rational only while preemption was permanent. With resume, the only cost is work to redo, and resume cost ∝ history length, so the least-progressed victim is cheapest; it also leaves the deepest sequences (closest to finishing and freeing all their blocks) running. Grammar-active sequences stay ineligible (matcher state not reconstructible); vision sequences are requeue-ineligible (image-embedding KV is not reproducible from tokens; spill remains eligible — new `Model::tokens_contain_vision_pad`).

**Starvation guard:** a resumed victim is immune from re-victimization until it has generated `PREEMPT_IMMUNITY_TOKENS` (64) new tokens, with an all-immune fallback so immunity never converts a recoverable exhaustion into a batch-wide error.

## Commit 2 — depth-aware admission (queue, don't admit-then-shoot)

New `scheduler/admission.rs`: a request reserves `prompt + min(max_tokens, WATERMARK)` tokens of blocks (clamped to the served context ceiling), counted against the total pool minus the same commitment of everything in flight (active + prefilling + spilled + requeued). Requests beyond capacity stay in the **existing pending queue** and admit as capacity frees.

- **Watermark default (PCND, explicit): `max_seq_len`** — i.e. the reservation is the request's own `max_tokens`, the honest conservative bound (a request can never generate past it). `ATLAS_KV_ADMIT_WATERMARK=<tokens>` lowers it for deliberate overcommit **with the WARN retained**; `0` = legacy prompt-only reservation (pre-gate behavior, pinned in tests). The boot-time `KV OVERCOMMIT` warning in `factory/build.rs` is unchanged and complementary.
- **No C≤64 regression, pinned:** with the ladder's shape (block 16, 6375-block pool, 202-prompt/1024-max requests) C=64 needs 4928 ≤ 6375 blocks — the gate admits all 64, byte-identical to the old decision (`fits_everything_admission_unchanged`). C=128 admits 82 and queues 46 instead of admitting 128 and shooting 171 (`c128_overflow_queues_instead_of_admit_then_shoot`).
- No head-of-line bypass; liveness guard (a lone oversized request with nothing in flight still admits, runtime back-pressure applies).

## Tests

Preemption (`preempt_tests.rs`, scripted `Model` stub, CPU-only): victim requeued **not killed** with its stream channel **open and empty** — the transport seam the SSE layer consumes; no mid-stream `Error`, no premature `Done` (the old HTTP-200-then-silence is asserted extinct). Exact-history re-prefill on resume with all stream state preserved; least-progress/immunity/vision victim policy; all-immune fallback; non-KV errors still fail the batch; single-sequence exhaustion still terminal; resume gating reclaims from the prefix cache and errors out a can-never-fit sequence rather than parking it forever.

Admission (`admission.rs`): watermark math pinned (incl. `max_seq_len` clamp and the legacy `tokens/bs + 1` formula), fits-everything unchanged, C=128 overflow queues, in-flight commitments reduce capacity, watermark-0 escape hatch.

Both commits pass `cargo fmt --all -- --check && cargo clippy --workspace --tests && cargo test --workspace` (ATLAS_SKIP_BUILD=1).

## Expected impact & validation plan

- **C=128, bf16 KV:** zero failed/truncated requests (the 171 kills become queued admissions + occasional resumable preemptions), aggregate ~**350–360 tok/s** — raw decode already runs 388–399 at parity with vLLM; the 130 tok/s gap was almost entirely discarded work plus wave restarts, and queueing trades a little TTFT tail for not throwing away 25% of decode.
- **C≤64 rungs:** admission decisions byte-identical (pinned); no perf delta expected.
- Validation (GPU box, when available):
  1. Re-run the C=128 ladder rung on Qwen3.8-27B, bf16 KV: assert `n_err=0`, zero `ptok=0` HTTP-200s, zero `send_error` preemption WARNs; count `Preempt-resume:` / `KV admission:` lines instead.
  2. Diff C=16/32/64 rungs vs `perf/ssm-verify-pool-capacity` baseline (warm-TTFT median ≤3%, p90 ≤5% — perf-path gate).
  3. `--swap-space` A/B at C=128 to compare spill-resume vs requeue-re-prefill.
  4. Stream-contract spot check: SSE client capture on a forced preemption — no error data frame, one final chunk with `finish_reason` + usage.


tbraun96:
Closing as included: merged into #572, which lands the certified tree as one unit (merging separately would invalidate the benchmark records at each step).

PR #544, opened 2026-08-16
nemotron: Nemotron-3.5 Lightning 30B-A3B support, consolidated on current main (supersedes #462 + #487)

rsafier:
Nemotron-3.5 Lightning 30B-A3B support, re-ported onto current `main`.

**Supersedes the #462 → #487 stack for Lightning support**, which can no longer be rebased: `main` has since replaced the API that stack extends. #487 added `ptx_for_shape(ModelShape { model_type, hidden_size, mtp_layers })`, discriminating Lightning from Nano on a numeric config field. Main's `atlas-kernels/src/resolve.rs` instead breaks a `(model_type, hidden_size)` tie with per-target **`match_names`** needles matched against the checkpoint reference — built for the Qwen3.6-vs-3.8 collision, which is the same problem. A rebase hit conflicts on the first commit with 19 to go, and a merge produced 16 conflicted files where every `ptx_for_shape` call site had been rewritten. So this is a re-port on main's design; `ModelShape` / `ptx_for_shape` are **not** reintroduced.

## What's here

- **`kernels/gb10/nemotron-3.5-lightning-30b-a3b/`** — the target. Byte-for-byte Nano's kernel set: the split is *policy*, not kernels. Nano documents `thinking_default = false` (thinking leaks training-corpus text into answers); Lightning wants thinking ON for chat and OFF for tool turns, which is unreachable while both share one target.
- **`match_names` on both sides of the tie** — `["lightning"]` here, `["nano"]` on `nemotron-3-nano-30b-a3b`. Not optional: main's `build.rs` refuses to compile a colliding pair where either side omits its needles, precisely so the tie can never resolve by build order.
- **transformers-5.x config parsing** — `layers_block_type` (Lightning ships the schedule as a list, not a `hybrid_override_pattern` string) and `num_nextn_predict_layers` → `mtp_num_hidden_layers`.
- **`nullable_usize` for `moe_latent_size`** — Lightning writes `null` where other checkpoints omit the key. `#[serde(default)]` covers omission but not an explicit null, so main fails the entire load with `invalid type: null, expected usize` before any GPU work.
- **ModelOpt sidecar wins over the embedded `quantization_config`** — the fix that made it *serve* rather than merely boot. transformers 5.x also embeds a compressed-tensors mirror whose `config_groups.group_0` describes only ONE group of a mixed-precision checkpoint (for Lightning, the FP8 mamba projections), so an embedded-wins rule labels the whole model FP8 and NVFP4 detection breaks. The symptom is not an error: the server boots, logs every config line correctly, and then **stalls in prefill at 0.9% CPU**.
- **NoPE** — the reference `NemotronHAttention` applies no rotary embedding (position is carried by the mamba layers; `config.json` ships a vestigial `rope_theta` the reference never reads). Applying RoPE scrambled long-range q·k addressing: retrieval died beyond ~50 tokens while short-range fluency survived on the SSM state. Two tests asserted `rotary_dim() == 128` — they codified the bug and now assert `0`.
- **`[behavior].no_decode_graphs`**, plumbed `ParsedBehavior` → `build.rs` → codegen → `ModelBehavior` → `ModelConfig` → `serve_load`. Nemotron-H crashes under decode-graph replay (CUDA 700/716 at specific prompt lengths), and graphs measure as a no-op on GB10.

## What's deliberately *not* here

**#467** (Nemotron-H multi-sequence batched decode) and **#486** (tool-call argument honesty) are **not** ported and remain independently useful — please keep them open. #462's MTP draft-head commit is also excluded. Lightning serves without any of them (the reference recipe pins `--speculative` OFF), and bundling three unrelated changes is how the original stack became unrebaseable in the first place.

## Verified

On a GB10, serving the [atlas-recipes#18](https://github.com/Avarok-Cybersecurity/atlas-recipes/pull/18) `defaults:` block verbatim — 256K ctx, C=8, fp8 KV, util 0.60 — and deliberately **not** passing `--kernel-target`, so target resolution was exercised rather than forced:

| | |
|---|---|
| target | `(sm_121, nemotron-3.5-lightning-30b-a3b, nvfp4)` |
| config | 52 layers, 6 attn, 23 SSM, 128 experts, `rotary_dim=0` |
| graphs | disabled by `MODEL.toml [behavior].no_decode_graphs` |
| thinking | `thinking_default = true` |
| quant | modelopt `MIXED_PRECISION` (sidecar), 72 ignored modules |
| decode | **76.5 tok/s** single-stream, structured JSON valid |
| prefill | 2136 tok/s @ 61k, 1821 tok/s @ 124k |
| needle, 61,091 tok | **PASS** at 10% / 50% / 90% depth |
| needle, 124,343 tok | **PASS** at 10% / 50% / 90% depth |

The needle rows sweep **depth** as well as length: a fact at 10% of a 124k context must survive ~112k tokens of following text, which is where a 23-SSM-layer hybrid would be expected to degrade first. It did not.

Two of three generation probes are **bit-identical** to the #487 stack's output on the same prompts — the strongest available evidence that this is a re-port rather than a rewrite.

## Tests

atlas-core 103 · atlas-kernels 32 · spark-model 549 · spark-server 2362 · atlas-plugin 727 — all passing. `cargo fmt` and clippy clean.

One test fails: `closure_attestation::the_baked_attestation_reproduces_from_the_tree` on `gb10/qwen3.8-27b/nvfp4`. It fails **identically on unmodified `avarok/main` 38395946** (verified by checking out main in the same tree and re-running), so it is pre-existing and unrelated to this change.


tbraun96:
Heads up — `main` was force-updated just now. Its last 7 commits were rewritten to excise 43.6 MB of bun cache artifacts (`.buncache/`, `.buntmp/`) that my PR #531 committed by accident. Only those files changed; every commit message and all other content is identical.

This PR's branch still descends from the old lineage, so **merging it as-is would reintroduce those 43.6 MB**. A rebase clears it:

```bash
git fetch upstream main   # or whichever remote points at Avarok-Cybersecurity/atlas
git rebase --onto upstream/main $(git merge-base HEAD upstream/main@{1}) HEAD
# simplest alternative if that base is awkward:
git rebase upstream/main
git push --force-with-lease
```

Sorry for the churn — it was my mess to clean up. The pre-rewrite state is preserved at tag `backup/pre-buncache-rewrite` if anything looks off.

rsafier:
Rebased onto current main (was CONFLICTING, now MERGEABLE).

The conflict was self-inflicted: this branch carried its own copies of the two site commits that have since landed on main — `feat(site): Ask the codebase` (#531) and `fix(ci): cap CodeRAG chunks at 8192 bytes` (#539). Same changes, different SHAs, so they collided with the upstream versions on every rebase attempt.

Dropped both duplicates and replayed only the Lightning commit, so this PR is now exactly **1 commit** on current main. No content was lost — #531 and #539 are already in main.

Verified on the rebased tip: `cargo build --release -p spark-server` clean, `cargo test --release -p spark-server -p spark-model -p atlas-kernels` passes.

rsafier:
Pushed 399d2db6: gate the K=γ verify graph (verify_d) on this PR's `[behavior].no_decode_graphs` — it was the last graph capture not consulting the key, and a Nemotron target reaching the K=γ verify still hit the graph-replay crash class the key exists to prevent. One-line semantics change; non-Nemotron targets unchanged. Validated on Lightning + DFlash on GB10 (no capture under the key, stable through full spec runs).

tbraun96:
**This is the only live one of the three nemotron PRs — and it looks like a real gap on `main`, not just a stale branch.**

I grouped #544/#545/#566 to certify them as one stack. #545 and #566 turned out to be **already on main** (empty cherry-pick; `weight_loader/nemotron.rs` and `jinja-templates/nemotron_h.jinja` are both there) and are now closed.

**This one is not**, and the piece that is missing is the interesting one:

```
ABSENT from main:  kernels/gb10/nemotron-3.5-lightning-30b-a3b/nvfp4/KERNEL.toml
```

So Nemotron-3.5-Lightning has its loader, its config plumbing and its chat template on `main`, but **not its kernel registry entry**. If I am reading the layout right, that means the model is wired up in Rust while GB10 has no registered kernel set for it. Worth checking whether it actually serves today.

It still cherry-picks onto current `main` and adds **527 lines across 18 files** — the KERNEL.toml plus `rms_norm.cu` / `w4a16_gemm.cu` entries and the `[behavior].no_decode_graphs` verify gate.

**I have not stacked it**, deliberately: it is 3 weeks old, it is 527 lines of model-support code that is yours, and its siblings landing without it suggests something was split up in a way I should not guess at. Two questions:

1. Is the KERNEL.toml omission on `main` intentional, or did it get lost when #545/#566 landed?
2. If this is still wanted, shall I rebase it onto current `main` and pair it with another `crates,kernels` change (#474 or #484) so one campaign covers both?

Happy to do either — just not to guess.

PR #545, opened 2026-08-16
nemotron: multi-sequence batched decode for Mamba-2 + MoE (stacks #544) — C=8 +45%, removes the C=8 regression

rsafier:
**Stacks on #544** — review that first; this PR's diff includes its commit until it merges.

Nemotron-H multi-sequence batched decode for Mamba-2 + MoE layers, re-applied on top of the consolidated Lightning support. This is #467's work, carried forward: that PR's base (#462) is closed and its stack could not be rebased onto current `main`, so the three substantive commits are cherry-picked here rather than the branch being rebased.

| commit | |
|---|---|
| `64479e6e` | multi-sequence batched decode for Mamba-2 + MoE layers (Milestone A) |
| `3eaf0344` | mixed-step scratch clobber — batched routing overwrote prefill positions/slots |
| `9a94d112` | gate batched-decode arms by measured thresholds — restore C≤4 parity + temp-0 onset |
| `e5043c45` | init `moe_lora_route` in the ported multi-seq test contexts (new; see below) |

## Measured

`nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-NVFP4` on a GB10, serving the [atlas-recipes#18](https://github.com/Avarok-Cybersecurity/atlas-recipes/pull/18) `defaults:` block (256K ctx, fp8 KV, util 0.60, `max_batch_size 8`). Aggregate decode throughput; distinct prompt per stream so prefix caching cannot collapse the batch; short prompts and 300-token generations so the wall is decode, not prefill. Same binary settings on both legs — only the commits differ.

| C | #544 baseline | this PR | delta |
|---|---|---|---|
| 1 | 76.1 tok/s | 76.1 | — |
| 2 | 81.4 | 81.7 | +0.4% |
| 4 | 83.6 | 84.1 | +0.6% |
| **8** | **65.1** | **94.5** | **+45%** |

Scaling vs C=1: baseline `C=2 1.07× · C=4 1.10× · C=8 0.86×` → this PR `1.07× · 1.11× · 1.24×`.

The number that matters is not only the +45%. **Baseline C=8 regresses below its own C=4** (65.1 vs 83.6), so on `main` today the eighth concurrent stream makes the box *slower* than four. The batched path removes that cliff. Wall-clock on the C=8 run: 28.7 s → 19.8 s.

C≤4 is parity, which is `9a94d112`'s measured-threshold gating doing its job — the C=4 regression previously noted on #467 does not appear.

## One porting note worth reading

All three commits cherry-picked with **zero textual conflicts** and the tree still would not build. `main` added `moe_lora_route` to `ForwardContext` after those tests were written, so both multi-seq test contexts were missing a field. The **lib compiled fine** — only the test profile broke — so a release build alone passes it straight through, and only the full test sweep catches it. That is `e5043c45`, using `MoeLoraRoute::Fold`, the documented default that keeps legacy single-request call sites byte-identical.

## Tests

atlas-core 103 · atlas-kernels 32 · **spark-model 558** (up from 549 — the 9 new multi-seq tests) · spark-server 2362 · atlas-plugin 727. `cargo fmt` clean.

One test fails: `closure_attestation::the_baked_attestation_reproduces_from_the_tree` on `gb10/qwen3.8-27b/nvfp4`. It fails **identically on unmodified `avarok/main` 38395946**, verified by checking out main in the same tree and re-running — pre-existing and unrelated.

## Not included

`ba01f3ec` (Lightning's MTP draft head, from the old #462 branch) is still unported, and #486 (tool-call argument honesty) remains open and independent of this.


tbraun96:
Heads up — `main` was force-updated just now. Its last 7 commits were rewritten to excise 43.6 MB of bun cache artifacts (`.buncache/`, `.buntmp/`) that my PR #531 committed by accident. Only those files changed; every commit message and all other content is identical.

This PR's branch still descends from the old lineage, so **merging it as-is would reintroduce those 43.6 MB**. A rebase clears it:

```bash
git fetch upstream main   # or whichever remote points at Avarok-Cybersecurity/atlas
git rebase --onto upstream/main $(git merge-base HEAD upstream/main@{1}) HEAD
# simplest alternative if that base is awkward:
git rebase upstream/main
git push --force-with-lease
```

Sorry for the churn — it was my mess to clean up. The pre-rewrite state is preserved at tag `backup/pre-buncache-rewrite` if anything looks off.

rsafier:
Rebased onto the rebased #544 (was CONFLICTING, now MERGEABLE).

All 4 batched-decode commits replayed with zero conflicts on top of #544's single rebased commit. Shows 5 commits here because a fork PR cannot use a fork branch as its base — the first is #544's. Review/merge order is **#544 → this → #566**.

Verified on the rebased tip: `cargo build --release -p spark-server` clean, `cargo test --release -p spark-server -p spark-model -p atlas-kernels` passes.

tbraun96:

Cherry-picking onto current `main` produces an **empty commit**; the tree already contains it. Confirmed independently: `crates/spark-model/src/weight_loader/nemotron.rs` and `jinja-templates/nemotron_h.jinja` are both on `main`, along with the `atlas-core/src/config/` changes.

Found while grouping the three nemotron PRs (#544, #545, #566) into one stack so a single campaign would cover them. **Two of the three were already in** — and the third turned up something worth your attention, flagged on #544: its kernel registry never landed, so the Rust support is on main without `kernels/gb10/nemotron-3.5-lightning-30b-a3b/nvfp4/KERNEL.toml`.

If a hunk here did not make it across, point at it and I will lift it onto a fresh branch.

PR #549, opened 2026-08-17
fix(mtp-telemetry): give every dispatchable batch width its own accept bucket

tbraun96:
## What

`mtp_accept_debug::MAX_N` was **17** while the MTP dispatch cap was raised 16 → 32 (`69658b873`, 2026-07-30). `record()` indexes `BUCKETS[n.min(MAX_N - 1)]`, so **every width in 16..=128 accumulated into bucket 16**, and the flush line printed whichever width happened to trip `PERIOD`.

Evidence — dgx2 `/home/claude/diag_c32_serve.log`, one serve, 301 flushes:

```
MTP accept n=32 k_drafts=1 verifies=128 p1=0.594 mean_na=0.594 tok_step=1.594 ...
MTP accept n=31 k_drafts=1 verifies=128 p1=0.704 ...
MTP accept n=18 ...  n=25 ...  n=26 ...  n=28 ...  n=29 ...  n=30 ...
```

Those eight "widths" are **one bucket**.

## Why it matters

`adaptive_rung` (BAND `9..=16`) steers the shipped n=16 rung from exactly this flush (`mtp_accept_debug.rs:124` → `adaptive_rung::observe`). Two live consequences:

1. The n=16 sample was a **mixture** of n=16 and n>16 statistics. They accept differently — a clean n=16 bucket reads `p1` 0.77–0.89 (local `serve-calib.log` 51 flushes / `serve-ladder.log` 31 flushes, same checkpoint) against **0.640–0.676 at n=32** across three dgx2 serves. The mixture biases the token ratio **down**, i.e. toward `k=1`.
2. A flush tripped by an n>16 caller is handed to `observe(n>16, …)`, which BAND discards outright — so under mixed-width traffic the n=16 controller **silently loses ticks it already paid for** (a probe costs a measured 1.74 s).

## Change

- `MAX_N` 17 → **33**: one bucket per width `0..=32`, the default `mtp_max_seqs()`.
- Index moves behind `bucket_idx()` — the SSOT the new test asserts on.
- Presence kill switch **`ATLAS_MTP_ACCEPT_FOLD_AT_16`** restores the old fold, so the correction is A/B-able against the binary that measured the ladder.

## Blast radius

Bit-identical for `n <= 15`. At `n` in 16..=32 it changes only **which counter** a verify lands in (and therefore which width the rung controller believes it observed). No verify, propose, sampling or scheduling path is touched. Cost: 16 more `Bucket`s of relaxed atomics, statically allocated.

## Tests

- `bucket_table_covers_the_mtp_dispatch_cap` — pins the table against `mtp_max_seqs()`, so the next cap raise cannot silently re-introduce the fold. This is the regression the 2026-07-30 raise shipped.
- `widths_up_to_the_cap_do_not_alias` — asserts `bucket_idx(n) == n` for every dispatchable width; fails for exactly the widths that aliased before this change (17..=32 onto 16).

## Gates

```
cargo fmt --all -- --check                                   clean
cargo clippy -p spark-server --all-targets                   clean
cargo test -p spark-server --bin spark mtp_accept_debug      5 passed / 0 failed
```
(`ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000`; CPU-only, no GPU leg — dgx2 is mid-benchmark.)

### DIFF
--- a/.benchmarks/agentic-webserver/2026-08-19-8ce64f6202.json
+++ b/.benchmarks/agentic-webserver/2026-08-19-8ce64f6202.json
@@ -1,0 +1,467 @@
+{
+  "schema": 1,
+  "benchmark_id": "agentic-webserver",
+  "benchmark_name": "Agentic Webserver Test",
+  "git_sha": "8ce64f6202",
+  "recorded_at": 1787119536,
+  "target_model": "Qwen/Qwen3.6-35B-A3B-FP8",
+  "params": {
+    "build_timeout_s": "600",
+    "command_timeout_s": "180",
+    "iterations": "10",
+    "max_tokens": "8192",
+    "max_turns": "40",
+    "request_timeout_s": "900",
+    "s_per_turn_budget": "8.5",
+    "serve_timeout_s": "30",
+    "wall_budget_s": "1800"
+  },
+  "command": [
+    "spark",
+    "benchmark",
+    "run",
+    "agentic-webserver",
+    "--param",
+    "build_timeout_s=600",
+    "--param",
+    "command_timeout_s=180",
+    "--param",
+    "iterations=10",
+    "--param",
+    "max_tokens=8192",
+    "--param",
+    "max_turns=40",
+    "--param",
+    "request_timeout_s=900",
+    "--param",
+    "s_per_turn_budget=8.5",
+    "--param",
+    "serve_timeout_s=30",
+    "--param",
+    "wall_budget_s=1800",
+    "--yes",
+    "--pull-request-gate"
+  ],
+  "served_by": "qwen3.6/qwen3.6-35b-a3b-fp8-bf16head",
+  "atlas_version": "1.0.0-beta-preview",
+  "hardware": {
+    "gpu": "NVIDIA GB10",
+    "driver": "580.159.03",
+    "sm_clock_mhz": 2418.0,
+    "source": "nvidia-smi"
+  },
+  "hardware_state": {
+    "sensitivity": "speed",
+    "before": {
+      "captured_at": 1787118813,
+      "machine": {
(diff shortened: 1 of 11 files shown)


tbraun96:
### Unstuck — the conflict was superseded history, not a real conflict

This PR had been `DIRTY` with four conflicted files (`impl_a1.rs`, `ssm_pool.rs`, `ssm_reserve.rs`, `preflight.rs`). None of those conflicts were real: the branch still carried the **entire pre-#572 ladder stack** (`b508679e4`, `f70d1facc`, `eee76aae3`, `af352c4ef`, `c3b73815a`, …), and main already has all of it via #572's squash merge — `ssm_reserve.rs` on main carries the tiered verify capacity, `mtp_state_slots`, tiered H intermediates and the rest.

Rebasing that stack onto a main that already contains it is what produced the conflicts.

**Only the tip commit was genuinely new** — 86 lines in `crates/spark-server/src/scheduler/mtp_accept_debug.rs`, absent from main (confirmed: main's copy is 291 lines with no `bucket_idx` / `bucket_table_covers_the_mtp_dispatch_cap`).

So the branch is rebuilt as **that single commit cherry-picked onto current main**. Clean pick, no conflicts, and the change is unchanged from what was reviewed.

**Verification:** workspace clippy, fmt and **5275 tests** all green; the commit's own test `scheduler::mtp_accept_debug::tests::bucket_table_covers_the_mtp_dispatch_cap` passes.

The `typos` check should also clear — it was failing against the old stack, not this file.

The PR benchmark gate will need records at this sha; `dgx2` is currently down (it has needed three powercycles today — see the vLLM C≥32 hazard recorded in #600), so that campaign is pending hardware.

PR #560, opened 2026-08-17
fix(tokenizer): coalesce multiple system messages regardless of developer role

Sujimoshi:
## Problem

Two leading `system` messages in a chat request hard-fail rendering with `System message must be at the beginning.` on most model templates. The OpenAI API permits multiple system messages, and agent frameworks send them routinely (e.g. mastra sends instructions + memory as two consecutive system messages) - every such request 400s.

## Root cause

`remap_developer_role` in `message_preprocess.rs` already implements exactly the right behavior - folding all system-level messages into one leading system message - but the fold was gated on the presence of a `developer`-role message (`if !has_dev { return messages; }`). Two plain `system` messages passed through untouched and hit the template's raise.

The same gate hid a second bug: two developer-only messages took the in-place remap branch and produced **two** system messages - the exact shape the fold exists to prevent.

## Fix

Gate the fold on the system-level message **count** instead of the developer role:

- 0-1 system-level messages: remap developer to system in place at most (unchanged behavior)
- 2+ (any mix of `system`/`developer`): fold string content into one system message at the first system position, preserving order

No template changes - the fix applies in Rust preprocessing ahead of rendering (per this module's own design intent), so every model's own template works, strict or not.

An earlier revision of this PR patched `jinja-templates/qwen3_5_moe.jinja` instead; that was dropped in favor of the cross-cutting Rust fix, matching the documented philosophy in `chat_impl.rs` (cross-cutting chat behaviors belong in Rust preprocessing, not per-model template overrides).

## Verification

Live on `Sehyo/Qwen3.5-35B-A3B-NVFP4` (GB10) with `--disable-template-overrides` - i.e. the checkpoint's own strict template:

- system + system + user: renders and answers (previously: 400)
- developer + developer + user: renders and answers (previously: 400)
- tool calling (`get_weather`): unchanged, well-formed tool_call

Unit tests: `remap_two_leading_systems_coalesce`, `remap_two_developers_coalesce`, `remap_mid_dialog_system_folds_forward` (+ existing 19 in the module all pass).


tbraun96:
Triaged for a certification campaign tonight and **parked pending a rebase**.

Merged onto current `main` (`5f639bfc57`) locally: conflicts in
`.claude/skills/measurement-discipline/SKILL.md`, `.gitignore`,
`crates/avarok-plugin/src/benchmarks/video/driver.rs` and
`crates/spark-model/src/layers/ops.rs` — none of which are tokenizer files.
That is the signature of accumulated drift, not a clash in your subject area:
the branch is **232 commits behind** with 22 commits over 47 files.

The two changes doing most of the damage are `569ffd0920` ("rebrand: Atlas to
Avarok on the site, crates, env, and cfg" — `atlas_*` crates and `ATLAS_*` env
vars are now `avarok_*`/`AVAROK_*`) and `ce2bccd736` (the WOA review follow-ups,
which rewrote `layers/ops.rs`).

I did not re-cut this one on your behalf deliberately. At 22 commits across 47
files, replaying own-commits-only risks silently dropping adaptations you made
along the way, and I would have no way to tell an intentional change from a lost
one. You have that context; a rebase from you is both safer and smaller.

Coalescing multiple system messages regardless of position is a real fix and we
want it — push a rebase and it joins the next composed campaign.

PR #563, opened 2026-08-17
chore: drop the bun cache and temp dirs that reached main

tbraun96:
My #531 committed `.buncache/` and `.buntmp/` — **371 files, 43.6 MB** of `@playwright/test` tarballs plus a strace log. bun writes those into the repo root when its default cache home is unwritable, and a broad `git add -A` swept them in. Removed from the tree and gitignored so it cannot recur.

Checked before deleting: the strace log contains only library-load syscalls, **no keys, tokens, or environment dumps**.

Note this reclaims working-tree and checkout cost, not pack size — the blobs remain in history. Rewriting them out would need a force-push over main plus a rebase of every open PR, so I have not done that; say the word if you want it and I will coordinate it.


tbraun96:
Superseded: the artifacts are now excised from main's history entirely (7 commits rewritten), so the file-removal half of this PR is moot. The gitignore guard continues in #564.

PR #566, opened 2026-08-17
nemotron: make Nemotron-3.5-Lightning tool calling work end to end (stacks #545)

rsafier:
Stacks on #545 (which stacks on #544). Review/merge order: **#544 → #545 → this**.
The commits below #544/#545 are shown here only because a fork PR cannot use a
fork branch as its base; this PR's own change is the single last commit.

## What was broken

Nemotron-3.5-Lightning could not complete an agentic turn. It either re-issued
the same tool call forever, or answered from thin air without ever calling.
Pointing opencode at it produced a loop that never terminated.

## Root cause — three stacked defects

**1. The template pre-opened the tool call.** `jinja-templates/nemotron_h.jinja`
emits `<think></think>\n<tool_call>\n` whenever tools are armed, so the
generation prompt *ends inside an open `<tool_call>`* and the model can only
ever complete a call — it can never answer, whatever the tool result says.
The template's own comment documents this for models not trained on the
qwen3_coder XML format, and `nemotron-super-120b-a12b` /
`nemotron-labs-3-puzzle-75b-a9b` already set `disable_tool_steering`. This
target inherited its MODEL.toml from Nano, which *needs* the prefix, and
inherited the missing flag with it.

**2. Thinking was disabled on tool turns.** This was the deeper one. Lightning
is thinking-first; with `<think></think>` forced empty it cannot decide
*whether* to call, so it fails in both directions at once. Measured on raw
`/v1/completions` — no injection, no parser, byte-exact prompts:

| | behaviour |
|---|---|
| thinking off + a pushy tool prompt | calls forever, never answers the tool result |
| thinking off + no pushy prompt | never calls — hallucinates `"README.md has 10 lines."` |
| **thinking on** | reasons, emits a well-formed `<tool_call>`, answers once the result returns |

The mid-call `</think>` cut that `thinking_in_tools = false` was avoiding is
now covered by the `</think>` injector deferral and the loop-watchdog guard.

**3. Two disagreeing tool prompts.** The template already renders a `# Tools`
block *and* the official `<IMPORTANT>` reminder the model was tuned against.
The qwen3_coder parser injects a second, longer, pushier description of the
same tools (`IMMEDIATE_TOOL_USE` + a 10-bullet `IMPORTANT`). The injected copy
pushes "emit the `<tool_call>` directly" hard enough that the model re-issues a
call instead of answering a `tool_response` already in its context.

## The fix

- `disable_tool_steering = true` and `thinking_in_tools = true` on the
  Lightning target.
- New `[behavior].no_tool_system_prompt` key, set on Lightning: suppress
  Atlas's parser-specific tool prompt and let the checkpoint's own template be
  the only place the tools are described.
- Gate the template's "Do NOT enter a `<think>` block before emitting a tool
  call" reminder on `not enable_thinking`. With thinking on that bullet
  contradicts the very next token and reads as "always call".

`no_tool_system_prompt` is **per-model and defaults to false**, because the
injection helps one family and hurts another — that is measured in both
directions and recorded in `api/chat/prepare.rs`:

```
Nemotron-3.5-Lightning  injection ON  ->  8/15, re-call loops and empty turns
                        injection OFF -> 13/15, no loop, no empty turn
Qwen3.8-AEON            injection ON  -> calls bash then ANSWERS in 2 turns
                        injection OFF -> loops 4+ turns, never answers
```

`ATLAS_NO_TOOL_SYSTEM_PROMPT=1` stays as the ad-hoc lever for a model with no
MODEL.toml entry of its own.

## Measured

5 agentic tasks x 3 reps at this target's preset sampling, scoring "called a
tool **and** then terminated with an answer", plus a first-call matrix over
{1, 11} tools x {system message, none}:

| | first call | agentic | prompt tokens (1 tool) |
|---|---|---|---|
| before | 2/4 | loops, never terminates | 1180 |
| fixes 1+2 only | 4/4 | 8/15 | 1180 |
| **all three** | **4/4** | **14/15** | **333** |

So it is also 3.5x cheaper per tool turn.

## Regression safety

- `no_tool_system_prompt` defaults false; every other model is untouched.
- The template gate keeps the reminder on the thinking-off branch, so **Nano's
  rendered prompt is byte-identical** before and after this change (verified by
  rendering both revisions of the template with Nano's tool-turn parameters).
  Nano's own agentic looping is pre-existing and unchanged — see the note
  already in `prepare.rs`; fixing Nano is separate work.
- `cargo build --release -p spark-server` clean, and
  `cargo test --release -p spark-server -p spark-model -p atlas-kernels` passes
  on the rebased tip.
- The serving behaviour above was re-verified on the rebased stack tip, not
  only on the branch it was developed on.


rsafier:
Pushed bc2a3665, two things found during the Lightning agentic bring-up on GB10:

1. **Streaming in-think tool-leak guard vs `thinking_in_tools`**: Lightning emits its `<function=...>` opener while the stream is still inside `<think>` — that IS the format this PR's `thinking_in_tools=true` declares — but the leak guard's default threshold of 1 cancelled the sequence on the first opener, killing **every streaming tool turn** (0/6 agentic steps in one turn per run). Non-streaming was never affected (parses the full text), which is why this PR's smoke passed. Fix: skip the cancel when `behavior.thinking_in_tools` is set; the downstream detector then parses the call and emits proper `tool_calls` deltas (verified end-to-end: streamed call, correct name+arguments, finish_reason `tool_calls`). Models without the key keep the guard unchanged.

2. **`[expected_absent]` for the 8 DFlash propose kernels** on the Lightning target (w4a16 fp8 row-scaled trio + nvfp4_mmq family) — the drafter proposes in BF16 until those are retargeted; without the declarations a DFlash serve refuses to boot on unresolved kernel lookups.

tbraun96:

Cherry-picking onto current `main` produces an **empty commit**; the tree already contains it. Confirmed independently: `crates/spark-model/src/weight_loader/nemotron.rs` and `jinja-templates/nemotron_h.jinja` are both on `main`, along with the `atlas-core/src/config/` changes.

Found while grouping the three nemotron PRs (#544, #545, #566) into one stack so a single campaign would cover them. **Two of the three were already in** — and the third turned up something worth your attention, flagged on #544: its kernel registry never landed, so the Rust support is on main without `kernels/gb10/nemotron-3.5-lightning-30b-a3b/nvfp4/KERNEL.toml`.

If a hunk here did not make it across, point at it and I will lift it onto a fresh branch.

PR #567, opened 2026-08-17
feat(ci): let a GitHub App author the harvest PR

tbraun96:
## Why

Harvest PRs (#534, #550) sit with **19 of 19 required checks unreported** because a PR opened with `GITHUB_TOKEN` never triggers workflow runs — GitHub blocks that to prevent recursion. Both needed a manual branch push to wake CI.

## Fix

An App installation token authors the PR, so CI fires normally and auto-merge carries it through the queue. Notably this needs **no bypass actors and no direct-push PAT** — `main` keeps every protection exactly as it is, which the `GOVERNANCE_PUSH_TOKEN` design could not have offered (a direct push would have had to defeat both rulesets *and* classic protection's required checks).

The workflow mints the token when `GOVERNANCE_APP_ID` is set and otherwise degrades to today's behaviour, warning on the PR that its checks need a nudge. `governance/README.md` documents the one-time App setup.

## Setup

App needs only **Contents: read/write** and **Pull requests: read/write**, installed on `atlas` alone, then:
```bash
gh variable set GOVERNANCE_APP_ID --repo Avarok-Cybersecurity/atlas --body '<app id>'
gh secret set GOVERNANCE_APP_PRIVATE_KEY --repo Avarok-Cybersecurity/atlas < key.pem
```

### DIFF
--- a/.github/workflows/governance-harvest.yml
+++ b/.github/workflows/governance-harvest.yml
@@ -72,8 +72,8 @@ jobs:
     # dispatch triggers pass through and enumerate for themselves.
     if: github.event_name != 'workflow_run' || github.event.workflow_run.event == 'pull_request'
     permissions:
-      # Commits governance/*.jsonl (direct push when GOVERNANCE_PUSH_TOKEN is
-      # configured with a ruleset bypass; branch push otherwise).
+      # Pushes governance/*.jsonl to the rolling bot branch. main itself is
+      # only ever written through the merge queue.
       contents: write
       # `read` covers listing/downloading run artifacts. `write` exists for ONE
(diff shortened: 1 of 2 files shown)


tbraun96:
Verified live before merge: dispatched `dry_run=true` from this branch (run 32045084601, green). The App token minted, and the dry-run artifact shows the harvest would land **38 PRs' worth of ledger lines, 77 events** (pr-502 through pr-567) — the backlog the cron sweep has been accumulating while every harvest PR sat with unreported checks.

Also switched `app-id` to `client-id` since v3 deprecated the former; the client id is a public App identifier so it lives as a variable, and `GOVERNANCE_APP_ID` has been removed as superseded.

PR #570, opened 2026-08-17
ci(cla): allowlist the governance harvester App

tbraun96:
#568 is the first PR the harvester App authored itself (which is the point of #567 — it makes the required checks report). CLA Assistant fails it because the bot identity is not on the allowlist.

A bot cannot sign a CLA, and the commits it pushes are authored by `tbraun96`, who has. This adds `atlas-governance-harvester[bot]` alongside the other bot identities already allowlisted.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -53,7 +53,7 @@ jobs:
       # SSOT for the allowlist: both the pinned action's `allowlist` input
       # below and the merge-queue mirror step read this one value. Keep it
       # here, nowhere else.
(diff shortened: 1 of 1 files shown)


tbraun96:
**Correcting my own diagnosis.** I opened this believing the missing allowlist entry was what failed CLA on #568. It was not — the actual error is `graphql call to get the committers details failed: HttpError: No server is currently available`, i.e. the GitHub outage in progress (status page reports a Partial System Outage). A re-run after GraphQL briefly recovered failed the same way.

#568's commits are authored by `tbraun96`, who is already allowlisted, so this may well be unnecessary for that PR. Keeping it open as defensive hardening on its own merits — a bot identity genuinely cannot sign a CLA, and the harvester will author every future PR — but it should be read as hardening, not as the fix for the current red check.

PR #581, opened 2026-08-18
bench(agentic): gate speed on seconds-per-turn, not Σwall

tbraun96:
## The problem

`agentic-webserver`'s speed bound is Σwall. Σwall is a **product** — turns × per-turn cost — and turn count is drawn by the agent, not the engine. Five tiers measured on the 35B flagship, **every one 10/10 `webserver_ok` + 10/10 `followed_directions`**, i.e. every one a correct run of shipped code:

| box | Σwall | Σturns | s/turn | old Σ ≤ 1000 |
|---|---|---|---|---|
| dgx1 | 774 s | 113 | 6.85 | pass |
| dgx1 | 799 s | — | — | pass |
| dgx1 | 813 s | 115 | 7.07 | pass |
| dgx1 | 860 s | 126 | 6.83 | pass |
| dgx1 | **1039 s** | 144 | **7.22** | **FAIL** |
| dgx2 | **1019 s** | 166 | **6.14** | **FAIL** |

The 774 s and 1039 s tiers are the **same box, same binary, same night** — a 34% swing with the code held constant. On one box Σwall spans 34% while s/turn spans 5.7%: six times tighter. And the wall bound **ranks backwards** — it failed dgx2's 6.14 s/turn, the fastest tier on record, while passing dgx1's 7.07.

This bound blocked #569 for hours on a run whose correctness was perfect.

## The change

- **`s_per_turn` is the speed bound.** The 35B commits a measured `max = 8.5` (~18% above its worst tier, >3× the on-box noise band away).
- **`sum_turns` states degeneracy directly** (35B `max = 250`, vs a 400-turn hard cap) instead of smuggling it into a wall bound.
- **`sum_wall_s` stays a blowup detector**, raised 1000 → 1800 on the 35B = the 250-turn ceiling × the 7.22 worst measured s/turn.
- **Speed divides agent-only wall.** `wall_s` was captured *after* the scorer's `cargo build --release`, contradicting the comment above it. Harmless per-tier; not harmless as a per-**turn** denominator, where a per-**iteration** constant contributes a term that *shrinks* as turns grow — ~0.28 s/turn across the observed range, 72% of the spread the 8.5 headroom is drawn against.
- **`sum_agent_wall_s`, `sum_turns`, `sum_tool_calls`, `sum_completion_tokens`, `decode_tps` are now recorded.** Turns were collected per-iteration and dropped, which is why every wall anomaly this campaign chased had to be re-measured by hand instead of read off the artifact.

## Two blockers an adversarial review caught before this went up

Both were in the first commit and are fixed in the second — kept as separate commits so the failure mode stays legible.

1. **The change was inert.** `apply_threshold_params` substitutes the variant's committed BENCH.toml bound over the schema default on every gated run, and `check_record` scores against that same bound. Editing `ParamValue::Float(...)` reached nothing the gate runs — the 1039 s tier would have failed exactly as before, twice over.
2. **It would have broken the dense 27B.** `s_per_turn_budget` defaulted to the 35B's 8.5 and no BENCH.toml declared the metric, so every variant inherited it. The dense 27B records **18 s/turn** (pre-branch nightly) and **~40 s/turn** (current serve); its healthy 10/10 reference tier is 4× the bound. Gate D's agentic leg would have failed 100% of the time, and *invisibly* — that entry commits no bound for `check_record` to report against.

**The default is now `0.0` = NON-GATING** (`min` relaxed to 0.0 so "off" is representable at all). A variant is speed-gated only once it commits a measured bound. A test pins that the 35B's 8.5 really would fail the healthy dense tier, so nobody re-globalises it.

## Tests

Pin all five measured tiers as passing under the **committed** 1800/8.5 bounds; prove the two the old bound rejected really were rejected; prove a 20% per-turn regression fails while its Σwall stays in budget (the gap the wall-only gate could not see); prove Σwall still catches turn degeneracy at a **reachable** 220 turns rather than the 400-turn theoretical maximum.

Workspace: 5213 tests, clippy, fmt — all clean.

## Follow-ups

- `/workspace/.claude/skills/benchmark-pr/SKILL.md` still documents `Σwall ≤ 1000`; it describes shipped behaviour today and should be updated when this merges.
- `decode_tps` is the honest denominator — tokens, not turns. Recorded but unbounded, because no variant has a measured token-rate tier and inventing one is the mistake above. Ratchet to it once tiers exist.


tbraun96:
### The gate now judges itself — and one estimate in the PR body was wrong

`agentic-webserver` ran under the new bounds at `e03f09a280`:

```
Pass: 10/10 webserver_ok · 10/10 followed_directions
      · 7.000s/turn ≤ 8.500 · Σwall 813s ≤ 1800s
```

Recorded metrics — the numbers that were previously collected per-iteration and dropped:

| metric | value |
|---|---|
| `sum_wall_s` | 813.19 |
| `sum_agent_wall_s` | 805.02 |
| `sum_turns` | 115 |
| `s_per_turn` | 7.0002 |
| `sum_tool_calls` | 110 |
| `sum_completion_tokens` | 27680 |
| `decode_tps` | 34.38 |

(The round 7.000 is coincidence: 805.02 / 115.)

**Correction.** The PR body claims the scorer's build contributes ~0.28 s/turn, i.e. 72% of the 0.39 s/turn spread the 8.5 budget is drawn against. Measured, the scorer's share is **8.2 s across all ten iterations — 1.0% of the tier**, because it builds into a warm shared cargo target dir. The real artefact is **~0.02 s/turn**, about **6%** of the spread, not 72%.

The split is still correct — the code did contradict the comment above it, and a per-iteration cost divided by a per-turn count really does bias with trajectory length — but it is a much smaller confound than I argued. **The 8.5 bound rests on the turn-count effect alone**, which is the 34%-vs-5.7% one and is not affected by this correction.

`decode_tps 34.38` is the first recorded value of the metric a future bound should use. One tier is not a threshold, so it stays unbounded.

**Gate status:** 8/10 passing (decode-floor 22.7 tok/s · ttft-cold −0.5%/−1.2% · ttft-warm · ssm-poisoning · vision · video 13/13 · concurrency-sweep · agentic). The two BFCL legs are still running on dgx2/dgx3.

PR #582, opened 2026-08-18
fix(server): honor min_tokens on legacy completions

zhachar14h-cell:
## Summary

Adds `min_tokens` support to the legacy completions endpoint and carries it through blocking, streaming, decode, and speculative verification paths. EOS is masked until the requested floor, forced-token handling cannot emit EOS early, and GPU argmax falls back to the host logits pipeline when the floor requires masking.

## Test plan

- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy -p spark-server --tests -- -D warnings`
- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo check -p spark-server --tests`
- [x] `cargo test -p spark-server --bin spark min_tokens` — 8 passed
- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo build --release -p spark-server`
- [x] `python3 scripts/check_spdx.py`

## Notes for reviewers

The public change is limited to the existing legacy completions request schema. Existing requests that omit `min_tokens` retain current behavior.

## CLA

- [ ] I have read and agree to the Contributor License Agreement.

### DIFF
--- a/crates/spark-server/src/api/completions.rs
+++ b/crates/spark-server/src/api/completions.rs
@@ -273,6 +273,71 @@ pub async fn completions(
     super::completions_exec::run_blocking(state, &req, prompts, params).await
 }
 
+/// Build the streaming `InferenceRequest` for the legacy completions path.
+///
+/// Extracted from `completions_stream` so the `min_tokens` passthrough is
+/// unit-testable; behavior identical to the original inline construction.
+fn build_streaming_request(
+    req: &CompletionRequest,
+    p: super::completions_exec::CompletionParams,
+    prompt_tokens: std::sync::Arc<Vec<u32>>,
+    session_hash: u64,
+    timeout_at: Option<std::time::Instant>,
+    token_tx: tokio::sync::mpsc::Sender<StreamEvent>,
+) -> InferenceRequest {
+    let echo = req.echo;
+    let logprobs_k = p.logprobs_k;
+    InferenceRequest::Streaming {
+        prompt_tokens,
+        session_hash,
+        adapter_slot: p.adapter_slot,
+        src_lang_id: p.src_lang_id,
+        tgt_lang_id: p.tgt_lang_id,
+        num_beams: p.num_beams,
+        length_penalty: p.length_penalty,
+        early_stopping: p.early_stopping,
+        image_pixels: Vec::new(),
+        max_tokens: req.max_tokens,
+        min_tokens: req.min_tokens,
+        temperature: p.temperature,
+        top_k: p.top_k,
+        top_p: p.top_p,
+        top_n_sigma: p.top_n_sigma,
+        min_p: p.min_p,
+        repetition_penalty: p.repetition_penalty,
+        presence_penalty: p.presence_penalty,
+        frequency_penalty: p.frequency_penalty,
+        // Legacy /v1/completions path doesn't have tool semantics.
+        dry_multiplier: 0.0,
+        dry_base: 1.75,
+        dry_allowed_length: 2,
+        lz_penalty: 0.0,
+        logit_bias: p.logit_bias,
+        stop_tokens: p.stop_tokens,
+        enable_thinking: false,
+        thinking_budget: None,
+        repetition_detection: p.repetition_detection,
+        require_tool_call: false,
+        // Completions API defines no tools — multi-tool-call continuation off.
+        tools_present: false,
+        suppress_tool_call: false,
+        disable_mtp: false,
+        grammar_spec: None,
+        seed: req.seed,
+        top_logprobs: logprobs_k,
+        prompt_logprobs: if echo { logprobs_k } else { None },
+        echo,
+        // Was hard-coded `None`: streaming /v1/completions was the one
+        // surface with NO deadline while every other surface had 300 s.
+        timeout_at,
+        token_tx,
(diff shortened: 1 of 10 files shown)


zhachar14h-cell:

zhachar14h-cell:
recheck

tbraun96:
**Confirmed: this bug is still live on `main` today.** Worth saying, because most of the backlog I have triaged this session turned out to be already-fixed — this one is not.

Evidence from current `main`:

| path | `min_tokens` |
|---|---|
| `api/chat_blocking.rs:143` | `min_tokens: req.min_tokens` ✅ |
| `api/chat_stream_dispatch.rs:95` | `req.min_tokens` ✅ |
| `api/completions_exec.rs:89` | **`min_tokens: 0`** ❌ |
| `api/completions.rs:318` | **`min_tokens: 0`** ❌ |

And `openai/completions.rs` has no `min_tokens` field at all, so the legacy request type does not even **parse** it. A client that sets `min_tokens` on `/v1/completions` has it silently dropped — no error, no effect — while the same field works on `/v1/chat/completions`.

**Where it stands:** the branch no longer cherry-picks onto `main` (it conflicts), so it needs a rebase before it can go anywhere. It touches `crates/` only, so it owes a certification campaign.

**Offer:** if you rebase it, I will fold it into a stack with another `crates/`-only fix so one campaign covers both rather than one each. Or if you would rather I did the rebase, say so and I will — I have avoided touching it unasked because ten files across the scheduler's logit-processor pipeline is not a mechanical conflict resolution, and guessing at your intent there is how a rebase quietly changes behaviour.

Either way this should not be sitting closed-adjacent in a backlog: it is a real, reproducible gap between two endpoints that claim the same API.

tbraun96:
Reopening this in a moment purely to refresh CI — no change to the branch, and nothing is being closed for good.

**Why.** This PR's last CI run completed **2026-08-19T05:13Z**. Since 2026-09-04 alone, `main` has touched its ten files across six merges (#1130, the #1101 rebrand, #1027, #981, #965, #934), so the green checks on this head describe a `main` that no longer exists. #1130's own description records that after #1101 a *clean* `git merge` still concealed five compile defects — so "merges clean" and "still compiles" are different claims here, and only one of them is currently evidenced.

A re-run of the August workflow run is refused (`its workflow file may be broken` — the workflow files have moved on too), so close/reopen is the way to get a `pull_request` run against today's base.

**Where this sits.** It merges clean onto `main` at `68c393913502` and changes 10 paths, so it is real, current content — not already-landed like #558 and #693 turned out to be. It touches `crates/**`, a perf path, so it needs a certification campaign (~3 GPU-hours). Those boxes are busy with #1145 and then #1144, which fixes a live decode race on `main`. This is queued behind them, and a current green is what it needs to be groupable with anything else.

tbraun96:
Pushed two commits to this branch (`maintainer_can_modify` is on): today's `main` merged up, and one field added to a test that `main` brought with it.

**What the fresh CI found.** After approving the parked workflow runs, this PR failed to compile against today's `main`:

```
error[E0063]: missing field `verify_pos` in initializer of `LogitsContext<'_>`
  --> crates/spark-server/src/scheduler/verify_pipeline_helper/pick_positions_tests.rs:80:15
```

This PR adds `verify_pos` to `LogitsContext`. `main` has since added `verify_pipeline_helper/pick_positions_tests.rs`, which constructs that struct. **The two never touch the same file, so `git merge` reports no conflict** — the defect exists only at compile time. That is the same class #1130 recorded after the #1101 rebrand, where a clean merge still concealed five compile errors.

**Why `0` is the right fill, and why it is inert.** `pick_positions.rs` never reads `verify_pos`, and the real verify path rebuilds the context per position — `LogitsContext { verify_pos, ..ctx.clone() }` in `verify_pipeline_helper.rs:157` — so a value carried by a *shared* context cannot reach any position's mask. It matches every other non-verify construction site in this PR: `decode_logits_step.rs` (×2), `pipeline_tests.rs`, `scheduler/mod.rs`.

**One thing worth knowing about this PR's history.** Its previous green checks were from **2026-08-19**, and its 11 workflow runs were sitting at `action_required` — the fork-PR approval gate — so nothing had run and nothing would have, however many times it was retriggered. The green you could see described a `main` that had moved on across six merges.

**Where this sits.** It touches `crates/**`, so it needs a certification campaign (~3 GPU-hours). Those boxes are on #1145 and then #1144, which fixes a live decode race on `main`. This is queued behind them; a current green is what it needs first.

tbraun96:
Landed on main in **#1203** (`d67ba440a6`), via the re-cut in #1205 —
`crates/spark-server/src/scheduler/logit_processors/min_tokens_eos.rs` is on
main, and `min_tokens` now routes through the scheduler's logit processors like
every other stop-suppressing lever instead of being accepted and ignored on the
legacy `/v1/completions` path.

Your two commits were replayed onto current main with `cherry-pick -x`. The one
conflict (`completions_exec.rs`) was an in-subject union, and the test
constructor you added was checked field-for-field against main's
`CompletionParams` — 18 fields, identical sets — before that resolution was
taken. A function-presence audit across all 11 touched files found nothing lost.

Thanks @zhachar14h-cell. Closing as superseded by the re-cut that carried it in.

PR #583, opened 2026-08-18
perf(ssm): add opt-in native FlashQLA prefill

zhachar14h-cell:
## Summary

Adds an opt-in Atlas-native FlashQLA prefill path for the Qwen3.6 Gated DeltaNet/SSM layers. The launcher uses embedded CUDA kernels and Atlas GPU abstractions without a TileLang/TVM Python runtime, supports packed QKV plus CP/no-CP paths, and retains the existing FLA implementation as fallback.

The feature remains disabled by default and is selected through `ATLAS_GDN_FLASHQLA` / `ATLAS_QLA_IMPL`.

## Test plan

- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo check -p spark-model --tests`
- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy -p spark-model --tests -- -D warnings`
- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo build --release -p spark-model`
- [x] `python3 scripts/check_spdx.py`
- [x] Exercised CP and no-CP native prefill paths on GB10 during development

## Notes for reviewers

- The AOT cubin build verifies ELF format, required symbols, kernel attributes, and normalized SASS fingerprints on CUDA 13.0.88.
- The vendor tree is intentionally compile-time-only and adds no TileLang, TVM, or Python runtime dependency.
- `flashqla_vendor/NOTICE` records FlashQLA, TileLang, and CUTLASS/CuTe attribution; vendored upstream whitespace is preserved.
- Existing FLA behavior remains the default and fallback.

## CLA


zhachar14h-cell:
recheck

tbraun96:
Thanks for this — the kernel itself is substantial work, and the fallback-preserving, env-gated design is the right shape. But **this PR cannot be accepted in its current form.** Three things need to change first, and all three are blocking.

### 1. The vendored tree has to go

The PR adds **25.5 MB across 848 files**, and **24.74 MB of that (797 files) is a full CUTLASS/CuTe drop** under `kernels/gb10/qwen3.6-35b-a3b/nvfp4/flashqla_vendor/cutlass/`.

What that mass is actually buying is very small. `flashqla_gdn.cu` includes no CUTLASS at all; the dependency arrives through one header:

```
tl_templates/cuda/instruction/mma.h:
  #include <cute/arch/mma_sm80.hpp>
  #include <cute/arch/mma_sm89.hpp>
```

Two CuTe arch headers, paid for with a 797-file vendor tree. A large part of it is unreachable here — the SM90 GMMA headers cannot run on GB10 (SM121) and nothing in the kernel references them.

**25 MB of vendored third-party code is against this repository's philosophy.** We recently rewrote history specifically to excise 43.6 MB of accidentally-committed artifacts; deliberately adding 25 MB permanently is not a trade we will make.

### 2. The C++/TileLang template machinery needs to be ported to Rust

Atlas is pure Rust and CUDA by design — one binary, no vendored template runtimes. The launcher and the surrounding template layer (`tl_templates/**`) must be expressed with Atlas's own Rust GPU abstractions and its own kernel sources, in the same idiom as the rest of `kernels/`, rather than carried as an upstream TileLang tree. The compute kernel staying CUDA is fine and expected; it's the vendored C++ scaffolding that has to be replaced.

### 3. There is no performance measurement

This is titled `perf(ssm)` and is proposed as a prefill improvement, but the test plan contains only compile checks plus "exercised CP and no-CP native prefill paths on GB10 during development." There is no TTFT delta, no throughput figure, and no accuracy comparison against the FLA path this is meant to displace.

We cannot merge a performance change whose performance is unknown — that is the one question the PR exists to answer. For the standard, see #584 in this same batch: it reports per-stage numbers (stage-1 TTFT 601.62 ms dense → 467.73 ms compact; down projection 167.34 → 59.50 ms) and demonstrates bit-exactness against the dense path across 40 layers and 240 snapshots.

Please include:

- TTFT and throughput on GB10 with `ATLAS_GDN_FLASHQLA` **on vs off**, same box, same checkpoint, back-to-back, with rep counts and spread;
- an accuracy or bit-exactness comparison against the existing FLA implementation;
- the prefill sequence lengths measured, since this is a prefill claim.

### Also needed

`flashqla_vendor/NOTICE` covers FlashQLA, TileLang and CUTLASS attribution — once the tree is pruned, that will need a license review against whatever remains.

Happy to look again once the vendor tree is gone, the scaffolding is Rust, and there are numbers. The underlying idea — a native GDN/SSM prefill path with no TVM/Python runtime — is worth pursuing.

tbraun96:
## Follow-up: it builds, it tests clean, and **the speedup is real** — here is what it takes to merge

Re-reviewed on GB10 (`sm_121`, CUDA 13.0.88) at `1f77d962`. Supplying the performance
measurement my earlier review said was missing, plus the reachability analysis behind
the vendor-tree objection.

### Test results — all clean

| check | result |
|---|---|
| `cargo fmt --all -- --check` | pass |
| `cargo clippy -p spark-model -p spark-runtime --tests -- -D warnings` | pass |
| `cargo test --workspace` | **0 failed** |
| AOT cubin build + SASS verify | pass — 11 symbols, `SASS 456690df…` |

Nothing below is a test failure. Every item is structural.

### `[MEASURED]` The kernel is worth having — 1.33x cold TTFT

`nvidia/Qwen3.6-35B-A3B-NVFP4`, `--max-seq-len 8192`, `--gpu-memory-utilization 0.85`,
~2048-word **unique** prompt per request (prefix cache defeated), streaming TTFT,
`temperature 0`, 2 warmup + 10 measured, same binary both arms:

| arm | median | mean | min–max |
|---|---|---|---|
| baseline FLA | 944.81 ms | 944.87 ms | 933.33 – 965.06 |
| `ATLAS_QLA_IMPL=native ATLAS_GDN_FLASHQLA=1` | **710.39 ms** | 712.15 ms | 701.00 – 743.69 |

**-234.4 ms, 1.33x.** The ranges do not overlap. `falling back to FLA` appears 0 times;
the kernel is loaded and marked `used`. Output quality holds — at a 640-token budget all
probes were coherent and on-task. Not bit-exact vs FLA (4/8 temp-0 prompts diverge), which
is expected for a different accumulation order in the recurrence, not a defect.

That number is why this is a "rework it", not a "close it".

### `[BLOCKING-1]` Wrong model. This has to be Qwen3.8.

`flashqla_gdn` is registered in exactly one place — `kernels/gb10/qwen3.6-35b-a3b/nvfp4/KERNEL.toml`.
`qwen3.8-27b` is its own kernel target (`Selected kernel target: (sm_121, qwen3.8-27b, nvfp4)`,
177 modules) whose `MODEL.toml` declares `Hybrid Attention + GDN + Dense FFN + MRoPE`, and whose
config carries `linear_attention` on **48 of 64 layers**. It has GDN, it would benefit, and it
gets nothing from this PR. Qwen3.8 is the target that matters here; please land it there.

### `[BLOCKING-2]` No opt-in flags. A beneficial path ships on.

The path is gated behind **both** `ATLAS_QLA_IMPL=native` **and** `ATLAS_GDN_FLASHQLA=1`.
A 1.33x prefill win that nobody turns on is not a win. Precedent is #388, which replaced an
env recipe with CLI flags and defaults. Make it the default with the existing FLA fallback
kept for the cases the runtime probes reject — the probe-and-fall-back structure already in
`gdn_flashqla_native.rs` is the right shape, it just needs the polarity flipped.

### `[BLOCKING-3]` The vendored tree — 82% of it is provably unreachable

The tree is 848 files / 683,820 lines / 28 MB. I asked the compiler which of those it actually
opens, rather than estimating:

```
nvcc -M -std=c++17 -arch=sm_121a -I. -Iflashqla_vendor -Iflashqla_vendor/cutlass flashqla_gdn.cu
```

The dependency output names **155 files, 6.1 MB**. The other **693 files / ~22 MB are never
included by any compilation** — that is what "provably dead" means here: not my judgement,
the compiler's own dependency list. Much of it cannot even run on this hardware (the SM90
Hopper GMMA headers on an SM121 part).

So there are two separable asks, and the first is cheap:
- deleting the 693 unreachable files costs nothing and is mechanical;
- the remaining 155 (133 CuTe/CUTLASS + 22 `tl_templates`) are the real dependency, and that
  template layer is what needs expressing in Atlas's own Rust GPU abstractions + CUDA sources.

For scale: the PR's own work is **2,054 lines of Rust vs 3,972 of CUDA/C++**. A CUDA compute
kernel is fine and expected — Atlas is Rust and CUDA. A vendored C++ *template runtime* is not.

### `[REQUIRED]` "Off by default" is only true at runtime

`flashqla_gdn` is in the target's `KERNEL.toml`, so it compiles on **every** build of that
target — the env flags gate dispatch, not cost. Two consequences to address:
- `build_target.rs` suppresses 7 nvcc diagnostics (`1444, 3287, 20012, 1046, 550, 1308, 179`)
  for this unit, which switches off the repo's strict kernel-warning policy for a 3,967-line
  Atlas kernel, not just for vendored headers. Scope the suppression to the vendor includes.
- the cubin gate pins a normalized SASS fingerprint to CUDA 13.0.88. It matches here, but a
  toolkit bump breaks the build rather than the kernel. Please state the intended behaviour
  on a version mismatch.

### `[NOTE]` Thin coverage for the new surface

2,054 lines of new Rust carry test markers in 4 files, and the launcher itself
(`gdn_flashqla_native.rs`, the largest new file) has 2. The probe/fallback decisions are
exactly the logic worth unit-testing without a GPU.

### `[NOTE]` LoC cap — mechanical only

Four files land over the 500-line cap (`gdn_flashqla_native.rs` 1077; `gpu.rs` 479 -> 610;
`gpu_impl.rs` 454 -> 534; `build_target.rs` 482 -> 545). Flagging only because
`Enforce <=500 LoC per source file` is a required check and will hold the merge — it is not a
design objection, and a split falls out of the Rust port anyway.

### Summary

The kernel earns its place: **1.33x cold TTFT, measured, non-overlapping**. Merge order I would
accept: retarget to `qwen3.8-27b`, default the path on with FLA fallback, drop the 693 dead
vendored files, then port the `tl_templates` layer to Rust. Happy to re-run the same A/B on
whatever you push.

tbraun96:
While the speedup is very nice, no external dependencies, please. Run a profiler to see where the kernel timing gaps are on Atlas side, and iterate to convergence. Add the minimum amount of code necessary to recreate this speedup. Re-open a PR and tag this once finished, because, the proof is in your numbers.

AzeezIsh:
@zhachar14h-cell I sent you an email too, may have gotten lost! Feel free to drop your discord handle too. Awesome stuff and we appreciate the contribution

zhachar14h-cell:
Thanks, I really appreciate it! I may have missed the email, so I’ll check my spam folder and reply there with my Discord handle. I’m planning to use #583 as the performance proof-of-concept and work toward a minimal, dependency-free Atlas-native implementation.

PR #589, opened 2026-08-18
fix(cuda): refuse a null pinned-host allocation instead of memsetting it

tbraun96:
Closes CodeQL alert #48 (`rust/access-invalid-pointer`, high).

The query is right for one narrow case: `cuMemAllocHost_v2` returning success is not a promise of a dereferenceable pointer. A zero-byte request may hand back null, and `write_bytes` is undefined behaviour on a null pointer **even for a count of zero**. The existing code trusted the status alone.

Checks the pointer instead. `cargo check -p spark-runtime` and `cargo fmt --check` pass.

### The other five CodeQL alerts
Dismissed as false positives with reasons recorded on each alert, not silenced:

- **#52, #53** `actions/cache-poisoning` — already mitigated: `release-build.yml`'s rust-cache carries `save-if: github.event_name != 'workflow_dispatch'`, so a dispatch-chosen ref restores the cache but never writes it. The query matches the structural pattern and cannot see a runtime guard.
- **#49, #50, #51** `rust/cleartext-logging` — test-only. `redact_tests.rs` is the test that *proves* redaction works, using synthetic fixtures (`AKIAIOSFODNN7EXAMPLE` is AWS's own documented example key), and the values appear only inside `assert!` messages that print on failure. Nothing reaches a log at runtime.

### DIFF
--- a/.benchmarks/agentic-webserver/2026-08-18-4546e02989.json
+++ b/.benchmarks/agentic-webserver/2026-08-18-4546e02989.json
@@ -1,0 +1,467 @@
+{
+  "schema": 1,
+  "benchmark_id": "agentic-webserver",
+  "benchmark_name": "Agentic Webserver Test",
+  "git_sha": "4546e02989",
+  "recorded_at": 1787037464,
+  "target_model": "Qwen/Qwen3.6-35B-A3B-FP8",
+  "params": {
+    "build_timeout_s": "600",
+    "command_timeout_s": "180",
+    "iterations": "10",
+    "max_tokens": "8192",
+    "max_turns": "40",
+    "request_timeout_s": "900",
+    "s_per_turn_budget": "8.5",
+    "serve_timeout_s": "30",
+    "wall_budget_s": "1800"
+  },
+  "command": [
+    "spark",
+    "benchmark",
+    "run",
+    "agentic-webserver",
+    "--param",
+    "build_timeout_s=600",
+    "--param",
+    "command_timeout_s=180",
+    "--param",
+    "iterations=10",
+    "--param",
+    "max_tokens=8192",
+    "--param",
+    "max_turns=40",
+    "--param",
+    "request_timeout_s=900",
+    "--param",
+    "s_per_turn_budget=8.5",
+    "--param",
+    "serve_timeout_s=30",
+    "--param",
+    "wall_budget_s=1800",
+    "--yes",
+    "--pull-request-gate"
+  ],
+  "served_by": "qwen3.6/qwen3.6-35b-a3b-fp8-bf16head",
+  "atlas_version": "1.0.0-beta-preview",
+  "hardware": {
+    "gpu": "NVIDIA GB10",
+    "driver": "580.126.09",
+    "sm_clock_mhz": 2392.0,
+    "source": "nvidia-smi"
+  },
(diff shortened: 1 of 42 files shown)


tbraun96:
## Reachability analysis, since this costs a full gate re-run

I traced whether the flagged path can actually execute today. **It cannot.**

`write_bytes` is only unsound here if `cuMemAllocHost_v2` succeeds with a null pointer, which requires `bytes == 0`. Working backwards through every caller:

- `impl_a1.rs:531` — `scratch.max(64 * 1024)`, floor 64 KB
- `verify_e.rs:677, :769` — `PINNED_CAP` and a literal `65_536`
- `ssm_spill_staging::acquire` — takes `spill_blob_bytes()` = `num_ssm_layers * (h_bytes + conv_bytes)`, **zero when a model has no SSM layers** (e.g. the dense Qwen3.8-27B in gate D)

That last one looks reachable, but `ssm_snapshot.rs:127` closes it:

```rust
let decode_enabled  = num_ssm_layers > 0 && decode_ring_slots > 0 && decode_max_seqs > 0;
let marconi_enabled = num_ssm_layers > 0 && num_slots > 0;
if !marconi_enabled && !decode_enabled { return Ok(Self { num_slots: 0, … }); }
```

Both gates require `num_ssm_layers > 0`, so a no-SSM model gets `num_slots = 0`, `is_enabled()` is false, and the spill path returns before `spill_blob_bytes()` is ever computed. There is no configuration reaching a zero-byte pinned allocation.

**So this is defense-in-depth against a future change to that coupling, not a live bug.** It is still worth landing — the coupling is three files apart from the memset that depends on it — but it does not justify re-earning 10 gate records (~4h19m GPU) on its own.

Recommendation: hold this open and fold it into the next PR that touches a PERF_PATH and pays that cost anyway. CodeQL #48 stays open meanwhile, which is the honest trade.

tbraun96:
**Parked as a draft, deliberately, to be certified as part of a stack.**

Per the reachability analysis above this cannot fire today, so it does not justify its own gate run. The decision is to batch it with the other PERF_PATH changes we want and pay for **one** measurement covering all of them.

Candidate stack (open PRs that would each invalidate the same 10 records on their own):

| PR | author | what |
|---|---|---|
| #589 | tbraun96 | this — CUDA null-pointer guard |
| #557 | dependabot | ratatui 0.30.2, which resolves the last open `lru` advisory |
| #554, #555, #556, #558 | dependabot | remaining dependency bumps |
| #549 | tbraun96 | MTP accept-bucket telemetry |
| #505, #507 | AzeezIsh | build fixes |

Held until the concurrent session finishes its in-flight work — `fix/mtp-think-skip-parity` and `perf/mtp-accept-width-buckets` are both live in other worktrees and both touch `crates/`, so certifying now would measure a tree that is about to change and waste the run.

GPU is idle as of this comment (0% util, no serve or benchmark processes), so nothing is mid-measurement.

tbraun96:
Superseded by #1245 (squash-merged as 08a0b49879): this PR's own fix, 1bccc48f56 (refuse a null pinned-host allocation), landed there. The branch had also absorbed feat/apple-metal-1to1, feat/bf16-native-loading and a base64 bump; those live on their own branches. Closed per owner decision.

PR #592, opened 2026-08-18
Qwen3.8-27B on AMD Strix Halo (gfx1151)

AzeezIsh:
Adds Qwen3.8-27B support for AMD Strix Halo (gfx1151), carrying the certified
Qwen3.6-27B MLPerf-edge recipe forward onto current `main`.

Up for **initial testing** — please build and serve it. Vision is next (see the
end).

Merged onto current `main` (`529fcb04`, #581) and re-verified there: builds
clean, serves, answers, MTP on, and decode measured back-to-back against the
pre-merge build in the same thermal state (13.0–14.6 vs 13.2–14.4 tok/s — no
regression; both are below the 19.18 tok/s median in the table because the box
had been benchmarking all day, which is the thermal decay noted below).

## Try it

```bash
git checkout strix/qwen3.8-27b
./build-amd.sh
./serve-amd.sh
```

No flags, no env vars, no extra installs. `build-amd.sh` needs ROCm and cargo;
`serve-amd.sh` defaults to `unsloth/Qwen3.8-27B-NVFP4` on `:8081` and is an
OpenAI-compatible endpoint:

```bash
curl -s localhost:8081/v1/chat/completions -H 'content-type: application/json' \
  -d '{"model":"x","messages":[{"role":"user","content":"hi"}],"max_tokens":64}'
```

The same binary still serves Qwen3.6 — `./serve-amd.sh nvidia/Qwen3.6-27B-NVFP4`.

**SCALE, `libibverbs-dev` and the `~/hip-port` shim are no longer
prerequisites.** The repo builds its own HIP `libcuda`/`libcudart`/`libcublasLt`
shims, and `ATLAS_NO_RDMA=1` drops the verbs dependency. The launch is simpler
than the 3.6 recipe, not harder.

## Measured on real hardware

gfx1151 / Radeon 8060S, 64 GB unified, ROCm 7.13, 2026-08-18. Qwen3.6 control
run on the same build, same box, same script.

| | Qwen3.6-27B | **Qwen3.8-27B** |
|---|---|---|
| startup to ready | 10.0 s | 12.5 s |
| TTFT (median) | 1949.8 ms | **1782.5 ms** |
| decode (median) | 14.14 tok/s | **19.18 tok/s** |
| TPOT (median) | 70.7 ms | **52.1 ms** |

Qwen3.8 is faster than Qwen3.6 on every axis measured here, with MTP
speculative decode on (`mtp=true`; ~22 of every 40 tokens come from accepted
drafts). Decode decays run-over-run on both (laptop APU thermals) — take
medians and cool down between runs.

BFCL single-turn leg, n=1150 (100 % of `non_live`), zero failures:
**83.83 overall / 80.27 normalized**. This is **not** comparable to the
certified 3.6 figure of 86.23 / 87.76 — different sample set (the certified 995
includes `irrelevance`, which scores ~100; this leg does not), different
weights, and a precision policy disabled to fit in memory. Per-subset it is at
or above both NVIDIA reference runs on `simple_python` (94.75) and `multiple`
(95.50), in-band on `simple_javascript` (38.00), and ~8 points under the
45–47 band on `simple_java` (37.00), which is the one open question.

## What is in the diff

* `kernels/strix/qwen3.8-27b/` and `kernels/strix-hip/qwen3.8-27b/` — new
  kernel targets. Both reuse the 3.6 `.cu` tree via `kernel_source`, so the
  multi-target build costs far less than it sounds (98 unique nvcc invocations
  for 283 requested). `match_names` disambiguates the
  `(model_type, hidden_size)` tie against 3.6.
  Deliberate divergence from gb10: `mtp_layers = 1`, so the MTP head actually
  loads.
* `match_names` added to the existing 3.6 strix targets — required, or
  `validate_collision_match_names` fails the build.
* `WeightStore::reclaim` — frees FP8 source tensors once their NVFP4 copy
  exists. **Without this Qwen3.8 does not fit on a 64 GB part at all**
  (51.8 GB → 44.8 GB pre-KV). Gated so it only fires when nothing can still
  read the FP8 copy.
* Three defects the `main` merge surfaced: a dropped `load_lm_head`, a
  duplicated `MidchunkCapture` (both branches added it independently, so git
  merged both without conflicting), and stripped `-lcublasLt`/`-lcudart`.
* `build-amd.sh` / `serve-amd.sh` defaults moved to native ROCm + Qwen3.8.

## Two things reviewers should know

**1. It runs with `--dangerously-allow-unresolved-kernel-lookups`.** The
gfx1151 kernel set is 94 modules where gb10's is 167, so 92 dispatch sites fall
back to slower general paths and the audit from #388 refuses to start without
the flag. This is **pre-existing, not new** — the audit landed on `main` after
the Strix branch forked, so the certified 3.6 submission was produced under
exactly the same fallbacks; they were simply invisible. Treat every Strix perf
number, including the certified ones, as measured on a build with 92
fallbacks. Roughly 31 of the 49 distinct missing modules are plain CUDA and
look like symlink candidates; ~11 are just module-name aliases.

**2. `ATLAS_NO_GDN_FP8_PREFILL=1` is a memory-for-accuracy trade.** It is what
makes the reclaim legal and the model fit, and it disables the policy that
keeps GDN projections at ≥FP8. The BFCL number above already includes whatever
that costs. The A/B that isolates it cannot be run on a 64 GB part — the
flag-off configuration does not fit.

## Text-only for now

Qwen3.8-27B is a vision-language checkpoint, and the strix kernel set ships no
`vision_encoder`. The server says so at startup and then **silently ignores
image inputs** — a request with an image answers from the text alone rather
than erroring. Text-only is the full scope of this recipe and of every number
above. Porting the vision encoder is the next piece of work.

## Not done

* Like-for-like 3.6 control on the same BFCL config (~5 h) — the measurement
  that would settle `simple_java`.
* `[expected_absent]` declarations for the 92 lookups. Writing 92
  confident-sounding justifications without measuring each fallback would be
  worse than the honest allow-flag.


tbraun96:
Triaged for a certification campaign and **parked pending a rebase** — but with the good news that this is the most tractable stale branch in the backlog.

Merged onto current `main` (`5f639bfc57`) locally with the rename limit raised: **only 8 conflicting files**, against 32 for #353 and 34 for #470. The branch is 197 commits behind with 50 own commits over 41 files, so it is still a rebase rather than something I should re-cut for you — but 8 is a short afternoon, not a rewrite.

Context on what moved underneath you: `569ffd0920` (#1101) renamed five crate directories (`crates/atlas-* → crates/avarok-*`) and with them `atlas_* → avarok_*`, `ATLAS_* → AVAROK_*`, `cfg(atlas_*) → cfg(avarok_*)`. Raise `merge.renameLimit=20000` before you estimate — git's default gives up on rename detection across a change that size and reports a misleadingly short conflict list (it understated #470's by an order of magnitude). Full write-up: #1206.

Campaigns are composed and queued for 22 PRs; anything rebased before one launches composes onto it and shares the GPU run rather than needing its own.

PR #595, opened 2026-08-18
fix(ci): treat a conflicting harvest PR as stuck, not healthy

tbraun96:
Caught in production. #585 taught the harvest to repair a stuck PR, but it only recognised `BEHIND`. After the five merges today, **#593 went `DIRTY`** — conflicting with main — and the triage logged:

```
harvest PR #593 is idle and healthy — leaving it alone
```

It is neither. `BEHIND` means main moved; `DIRTY` means the branch now conflicts with it. Neither clears by waiting, and a rebuild resolves both by construction, since the branch is recreated as main plus a freshly derived ledger.

**#593 is deliberately left open as the test case**: once this merges, the next harvest should rebuild it automatically instead of walking away.

### DIFF
--- a/.github/workflows/governance-harvest.yml
+++ b/.github/workflows/governance-harvest.yml
@@ -369,7 +369,16 @@ jobs:
               emit_landed "waiting on #$open_pr"
               exit 0
             fi
-            if [ "${broken:-0}" -eq 0 ] && [ "$pr_state" != "BEHIND" ]; then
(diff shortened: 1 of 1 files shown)


tbraun96:
Superseded by #596, which lands the same DIRTY fix but as an extracted, unit-tested script (.github/scripts/harvest-triage.sh) with a 13-row decision table behind it. Closing rather than merging both.

PR #584, opened 2026-08-18
perf(moe): compact prefill row-list and K32 down

zhachar14h-cell:
## Summary

Adds an opt-in bit-exact NVFP4 MoE prefill path for GB10/Qwen3.6-35B-A3B:

- builds a GPU `(expert, m_tile)` row-list once per layer and reuses it for gate/up and down;
- expands N tiles in the grouped GEMM launch, avoiding the dense worst-case expert grid;
- reserves fixed model scratch using the proven `ceil(total_expanded / 64) + num_experts - 1` row bound;
- adds the stage-3 K32 down candidate selected by forward search;
- retains dense/K64 fallbacks and keeps all new paths disabled by default.

Enable compact mode with `ATLAS_MOE_PREFILL_COMPACT_NVFP4=1`; select the retained K32 down candidate with `ATLAS_MOE_PREFILL_COMPACT_K32=1` or the aggregate experimental switch `ATLAS_PREFILL_MAX_EXACT=1`.

## Performance

GB10, Qwen3.6-35B-A3B Standard NVFP4, Sonnet 2048/200, prefix cache off, 2 warmups + 5 measurements:

- stage-1 row-list TTFT: 601.62 ms dense → 467.73 ms compact;
- stage-1 gate/up: 155.851 → 107.915 ms;
- stage-1 down: 167.342 → 59.496 ms;
- selected K32 down combination: grouped GEMM 168.126 → 159.885 ms and TTFT 467.51 → 456.40 ms versus the compact K64 baseline.

The selected K32 run generated all 1000 requested tokens.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo check -p spark-model --tests`
- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy -p spark-model --tests -- -D warnings`
- [x] `cargo test -p spark-runtime buffers` — 11 passed
- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo build --release -p spark-model`
- [x] Real nvcc compile for `gb10/qwen3.6-35b-a3b/nvfp4` — 169 kernels
- [x] Dense vs compact BF16 comparison: 40 layers, 240 routing/GEMM/final snapshots, bit-exact
- [x] K32 candidate dense/compact comparison: 40 layers, 240 snapshots, bit-exact
[...]

PR #597, opened 2026-08-18
stack: min_tokens on legacy completions + compact NVFP4 MoE prefill (#582 + #584)

tbraun96:
Stacks **#582** and **#584** by @zhachar14h-cell (cherry-picked, authorship preserved) so they share **one** gate certification.

## Why stack them

Any change under `crates/` invalidates all ten benchmark records. #582 and #584 each do, so merging them separately costs **two full re-cuts (~8 h of GPU)**; certified together it is **one (~4 h)**. #584's kernel changes are the numerically risky half, and the BFCL/agentic evidence that clears them is exactly the evidence #582 needs too.

## #582 — `min_tokens` on legacy completions (correctness)

Carries `min_tokens` through blocking, streaming, decode and speculative verification: EOS masked until the floor, forced-token handling cannot emit EOS early, GPU argmax falls back to the host logits pipeline when masking is required. Requests omitting `min_tokens` are unaffected.

Note this is **not** a prefill change despite riding with one — it touches decode-side logit processing (`min_tokens_eos.rs`, `forced_token.rs`, `decode_logits_step.rs`).

## #584 — compact NVFP4 MoE prefill (performance)

| Stage | Dense | Compact | Δ |
|---|---:|---:|---:|
| stage-1 TTFT | 601.62 ms | **467.73 ms** | −22% |
| gate/up | 155.85 ms | 107.92 ms | −31% |
| down | 167.34 ms | **59.50 ms** | −64% |
| + K32 down | 467.51 ms | 456.40 ms | −2.4% |

Bit-exact against the dense path across 40 layers / 240 routing+GEMM+final snapshots. Builds a GPU `(expert, m_tile)` row-list once per layer and reuses it for gate/up and down, avoiding the dense worst-case expert grid.

**Both paths are off by default** — `ATLAS_MOE_PREFILL_COMPACT_NVFP4=1`, `ATLAS_MOE_PREFILL_COMPACT_K32=1`.

## Verification so far

- `cargo test --workspace` — **5280 passed, 0 failed**
- `cargo clippy --workspace --tests` — clean
- Real kernel build (`cargo build --release -p spark-server`, no `ATLAS_SKIP_BUILD`) — succeeds

## Gate records: deliberately not cut yet

#588 is CLEAN and about to land, and it touches `Cargo.lock` + `crates/`. Cutting records now would guarantee a second re-cut the moment it merges — which is what happened to #588 itself earlier today when main moved mid-certification. So this branch will take main, then be certified **once** against a settled main.

## Follow-up this PR does not do

#584's win is **off by default**, so merging it changes nothing measurable until enabled. An off-by-default optimization that no gate exercises is the PR #296 failure mode — a silently-disabled win that passes every gate. A follow-up should enable it in the recipes and add a TTFT leg with it on.

### DIFF
--- a/crates/spark-model/src/layers/moe/dump.rs
+++ b/crates/spark-model/src/layers/moe/dump.rs
@@ -21,6 +21,36 @@
 use anyhow::Result;
 use spark_runtime::gpu::{DevicePtr, GpuBackend};
 
+/// Monotonic ordinal for raw per-layer prefill snapshots. A validation run
+/// starts one server process and executes a fixed prompt, so dense and compact
+/// runs receive identical names without changing the public forward API.
+static PREFILL_RAW_COUNTER: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUsize::new(0);
+
+/// Debug-only final BF16 output dump for the retained stage-1/3 comparator.
+/// Synchronization and D2H traffic are entirely absent unless
+/// `ATLAS_MOE_PREFILL_DUMP` is set.
+pub fn dump_prefill_bf16(
+    gpu: &dyn GpuBackend,
+    stream: u64,
+    ptr: DevicePtr,
+    elements: usize,
+    stage: &str,
+) -> Result<()> {
+    let dir = match std::env::var("ATLAS_MOE_PREFILL_DUMP") {
+        Ok(dir) if !dir.is_empty() => dir,
+        _ => return Ok(()),
+    };
+    gpu.synchronize(stream)?;
+    let mut bytes = vec![0u8; elements.saturating_mul(2)];
+    gpu.copy_d2h(ptr, &mut bytes)?;
+    std::fs::create_dir_all(&dir)
+        .map_err(|e| anyhow::anyhow!("create dump directory {dir}: {e}"))?;
+    let layer = PREFILL_RAW_COUNTER.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+    let path = std::path::Path::new(&dir).join(format!("moe_final_L{layer:03}_{stage}.bin"));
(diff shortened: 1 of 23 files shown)


tbraun96:
**Request to include #589 in this stack**, since this PR is already going to pay for a gate run.

#589 is a five-line null-pointer guard in `alloc_host_pinned_cu`: `cuMemAllocHost_v2` returning success is not a promise of a dereferenceable pointer, a zero-byte request may hand back null, and `write_bytes` is UB on null even for a count of zero. It closes CodeQL alert #48.

I traced its reachability and it **cannot fire today** — both gates in `ssm_snapshot.rs:127` require `num_ssm_layers > 0`, so no configuration reaches a zero-byte pinned allocation. That is exactly why it should not buy its own ~4h19m certification, and why it is a good passenger: it touches one function, cannot change behaviour on any reachable path, and cannot move a number this stack is measuring.

It is parked as a draft for this purpose. Happy to push the commit onto `stack/prefill-582-584` on request — I have deliberately not done so unprompted, since this is another session's branch and the gate run should not be invalidated underneath it.

GPU is idle as of this comment (0% util, no serve or benchmark processes), so nothing is mid-measurement.

zhachar14h-cell:
Thanks for stacking these changes and preserving authorship. I have no objection to including #589 in the shared certification run. I can also prepare the `moe_grouped_a.rs` submodule split for the 500-LoC gate if helpful.

tbraun96:
Closing: superseded. This was a stack of #582 (`fb7139cea0`) and #584 (`261932c6fe`), and both are now being handled as their own PRs — #582 goes into a server-semantics composition, #584 needs its own 4-file re-cut against the current MoE prefill path. Keeping this third copy open only adds a conflicting branch to triage. No branch deleted.

PR #493, opened 2026-08-13
spark-model: default-on fused K=2 GDN verify epilogue

facepunch47:
## Summary

Production C=1 K=2 MTP verify was launching `conv1d_update_l2norm` twice plus a conv-state `copy_d2d` on every SSM layer even though the bit-identical fused kernels (`gdn_verify_fused_k2`) were already linked on gb10. This flips the old opt-in `ATLAS_GDN_FUSED_VERIFY=1` (default OFF) to a `ATLAS_NO_GDN_FUSED_VERIFY=1` kill switch so the fused path is the shipped default.

Closes nothing tracked; this is the next GB10 decode bottleneck after #478/#479/#480/#481/#485.

## Bottleneck (evidence, not folklore)

Live `:8888` on `nvidia/Qwen3.6-35B-A3B-NVFP4` with `--speculative --num-drafts 1` (K=2 every verify step):

- Kernel table at boot: `gdn_verify_fused_k2 … used`
- Container env did **not** set `ATLAS_GDN_FUSED_VERIFY=1`
- No log line for fused K=2 on the baseline image — the kernels sat idle
- Contrast: K=17 DFlash fused conv is already default-ON

That extra per-layer work (two conv launches + D2D × ~30 SSM layers) is off the NVFP4 weight-byte roofline and sits inside the graphed ~12 ms `fwd`.

## What changed

- New `gdn_fused_verify_k2.rs`: pure `gdn_fused_verify_k2_from` so polarity is unit-testable without process env. Unset / `=0` / junk → ON; only `ATLAS_NO_GDN_FUSED_VERIFY=1` kills.
- One-shot INFO: `GDN fused K=2 verify ENGAGED … kill switch ATLAS_NO_GDN_FUSED_VERIFY=1`
- Obsolete `ATLAS_GDN_FUSED_VERIFY` is ignored (warned once)
- Call site still requires both fused kernel handles (NULL → per-token path)

## Image proof on :8888

Same flags both sides (`--gpu-memory-utilization 0.50`). Same five prompts, temperature 0. Short 2-token replies are noise; 220-token lifecycle tok/s is the metric.

Baseline image `atlas-gb10:pr-proof-1b906f0` (#478–#481 + #485). After image `atlas-gb10:pr-proof-321128d` layers this binary on that runtime. Git PR is this one commit on `origin/main` (`35c5a46`); the combined proof stack was not pushed.

### Smoking gun

Baseline: **zero** `GDN fused K=2 verify` lines.

[...]

Issue #459, opened 2026-08-11
#435 not closed by --exact-verify: the exact arm IS entered, yet spec-on still diverges from spec-off (second divergence source outside the GDN chain)


tbraun96:
Measured on dgx1 tonight while discharging the owed hardware validation for #435.

## What was run

Four arms, `unsloth/Qwen3.6-27B-NVFP4`, sequential requests (C=1), prefix
caching OFF, temp 0, seed 42, 8 prompts x 400 max_tokens. Full generation
captured (`reasoning_content` + `content`) plus `completion_tokens`.

| arm | flags | vs control A |
|---|---|---|
| A | (none) — spec-OFF control | — |
| B | spec-OFF, prompts issued in REVERSE order | **8/8 identical** |
| C | `--speculative --mtp-gate force --exact-verify` | 7/8 identical |
| D | `--speculative --mtp-gate force` | 7/8 identical |

Arm B is the reproducibility control: spec-OFF at C=1 with prefix caching off
reproduces itself exactly, and reversing the prompt order does not perturb
per-prompt output, so phase-stability is not manufacturing the agreement.

Arm D is the negative arm — it diverges from A, which is the documented and
expected default behaviour (#435 divergence ships by default). So the
comparison demonstrably CAN detect divergence.

## The finding

**Arms C and D are byte-identical to each other on all 8 prompts** (~3200
generated tokens), including the divergence from A: same prompt, same character
offset (1556 of 1587), same continuation.

Two independent signals say `--exact-verify` did nothing:

1. **Output**: C == D, 8/8, byte-for-byte.
2. **Cost**: arm C 191 s, arm D 192 s for the same 8 prompts. `serve_args.rs`
   documents the exact chain at ~+35% per decode step at the n=8/K=4 verify
   rung (GDN phase +116%). Observed: ~0%.

Either of those alone is suggestive; together they are hard to reconcile with
the exact arm having run. Different verify arithmetic that happened to produce
bit-identical output across 3200 tokens AND cost nothing is not a plausible
coincidence.

## What is ruled out

- **The flag was not silently dropped at the OnceLock.** `serve_flags.rs`
  deliberately logs the RESOLVED value (re-reads `gdn_flags::flags()` and calls
  `verify_exact_active()`, precisely so a dead knob cannot hide), and arm C's
  serve log reads `exact_verify=true`.
- **Speculation was running.** Both arms log `mtp=true, num_drafts=3` (K=4) and
[...]

PR #604, opened 2026-08-19
proto: qwen3.8 temp-0 spec-decode divergence — localized to the GDN verify path (+ verify parity fixes, DFlash2 loader WIP)

rsafier:
**DRAFT / WIP — not for merge as a unit.** This branch is pushed to the upstream repo so the individual commits are easy to cherry-pick and the investigation record is shared. Two workstreams, four commits, each commit self-contained.

---

## Workstream 1: the qwen3.8-27B temp-0 spec-decode divergence — localized to the GDN verify path

**Symptom** (deterministic, reproducible via the gate harness): `video-fidelity`'s `video-before-image` cell FAILS with MTP speculation on (`got [red, green, blue]`, `yellow` lost) and PASSES with `--serve-override speculative=false num_drafts=0`. At temperature 0, speculation must be output-invariant — so this is silent wrong answers under the pinned production recipe.

**The decisive measurement** (token-exact emit logs on both legs, this branch's instrumentation): both legs emit identical tokens through `"…which represents the video frames"`, then the verify path picks `,` (id 11) where sequential picks `:` (id 25) — **same prefix, same host-FP32 penalty-aware argmax code, different logits**. The divergence is in the forward, not the sampler.

**Eliminated, each by a direct A/B on the same cell** (all on main, binary rebuilt per leg):

| suspect | test | verdict |
|---|---|---|
| verify pipeline penalty semantics | `SHADOW_TGT raw=` vs `v=` + text align | flips (e.g. `,`→`' frames'`) MATCH sequential — penalties working as designed |
| stale penalty history at verify positions ≥1 | fixed (commit 1) + re-run | still fails — but the staleness was a real defect, fix stands |
| beyond-accept stateful-stage mutation | fixed (commit 1) + re-run | still fails — also a real defect, fix stands |
| fast-greedy GPU-bf16 tie-breaks (#237 trade) | `ATLAS_DISABLE_FAST_GREEDY=1 ATLAS_NO_FAST_GREEDY_CHAT=1 ATLAS_DISABLE_FAST_MASKED=1` | still fails |
| CUDA-graph replay staleness after rejection rewind | `ATLAS_DEBUG_NO_GRAPH=1` | still fails |
| batched w4a16 GEMV tiers | byte-parity microtest extended to the EXACT qwen3.8-27b shapes incl. lm_head [248320×5120] (commit 4) | byte-identical, 1-ULP control detected |

⚠ **Trap for future debuggers:** `ATLAS_FORCE_TEMP_ZERO=1` makes spec-on pass 13/13 — but only because bypassing the pipeline changes the trajectory *before* the bad junction. It is NOT proof the forward is exact. Also: a `--pull-request-gate` leg that passes under diagnostic env WRITES a real gate record — one was caught and deleted here.

**What remains — the standing suspect:** qwen3.8-27B is a GDN hybrid (**48 linear_attention + 16 full_attention layers**, per checkpoint config). The eager verify (`verify_a.rs`) loops full-attention layers per-token through the SAME `layer.decode` as sequential — parity by construction — but routes SSM/GDN layers through the "GEMM-batched `decode_batched` override". That asymmetry, on 48 of 64 layers, is the only structural difference left standing. Note #493/#395's "bit-exact" claims are fused-vs-verify's-own-loop, not verify-vs-sequential-decode — the axis proven divergent here.

**Proposed next step:** a bit-parity microtest of the GDN verify leg against the sequential decode leg at qwen3.8's GDN dims (48 v-heads × 128, 16 k-heads × 128, conv 4), following the `gdn_verify_fused_microtest` pattern; then bring the batched verify leg to parity the way #474/#477 did for the Nemotron GEMV family.

Related: the Nemotron batched-decode divergence at C≥2 is plausibly the same class on Mamba-2.

## Workstream 2: DFlash2 drafter support (WIP, NOT wired)

`incoai/Qwen3.8-27B-DFlash2` (z-lab `DFlash2DraftModel`) is DFlash1 + exactly 7 new tensor families. Commit 2 lands config parse (4 new `dflash_config` keys, absent = DFlash1), loading for all 7 families (codebooks ship with NO `.weight` suffix), a host-side selector design (codebooks bf16-as-u16 in RAM, ~254 MiB, zero VRAM — the walk is 16 rank-256 dots/position), and the one new CUDA kernel (grouped dynamic causal conv). Still to do: conv insertion in `forward_block_layer_paged`, selector walk in propose, e2e serve.

Measured correction to the DFlash2 blog: the shipped selector is 128.5M params (two vocab-sized codebooks) and the convs 65.7M — ~370 MiB added, not the stated 19.5M params.

## Cherry-pick guide

| commit | content | standalone? |
|---|---|---|
| `1cbb6548` fix(verify) | penalty-history parity + early-stop at first mismatch + SHADOW `raw=` + response-text debug logs | **yes — recommended**; two real parity defects + permanent diagnostics; `cargo check --tests` clean |
| `173b54f6` wip(dflash2) | loader/config/structs/selector/conv kernel, inert until wired | yes (adds only optional paths) |
| `0b2c75ae` diag(verify) | `emit tok=` debug at both decode paths' push sites | yes |
| `548c04f6` test(parity) | qwen3.8-27b shapes in the w4a16 bitparity gate (all PASS) | **yes — recommended** |

All measurements on gb10 (gx10), checkpoint `unsloth/Qwen3.8-27B-NVFP4`, gate harness self-start, one leg per binary. Rebased onto `fc40e21e`.


rsafier:
## Update: kernel defect found and fixed; divergence narrowed to the native-FP8 GDN route, one layer still open

**Fixed (commits `8806fe88` + `ce3d5736`):** `w8a16_gemv_batch4/16` — the K≤8 MTP-verify projection kernels on native-FP8-GDN checkpoints — were NOT byte-identical to M× `w8a16_gemv` (the M=1 decode kernel), despite the header's bit-parity claim:

- paired-element accumulate (`acc += lo*w0 + hi*w1`) vs decode's per-element chain → 1-ULP BF16 flips on scattered rows at qwen3.8's shapes (qkvz [16384×5120]: max|Δ|=2.0 at M=4; out_proj [5120×6144] likewise)
- tail-thread `return` before the reduction's second `__syncthreads` (UB)

The old gate (`w8a16_gemv_batch4_microtest`) is cosine-based at one small shape and PASSes `max_abs=0.125` — the new `w8a16_batch_bitparity_microtest` is raw-bytes at the exact production shapes and is the acceptance criterion. After the fix: byte-identical at every shape and M.

**Status of the temp-0 divergence itself:** NOT yet closed. With the fixed kernels, video-fidelity still fails at pure defaults *and* with `--exact-verify`; `ATLAS_NO_GDN_FP8_PREFILL=1` (suppressing the native-FP8 GDN arm entirely) still passes 13/13 with speculation ON at unchanged accept/throughput — so it remains the operational workaround, and at least one more verify-vs-decode delta hides on the fp8w route. Unexamined candidates: the QKVZ deinterleave/layout step, and `ssm_forward`'s `sequential_qkvz` M=1 branch vs the batched path.

Elimination record so far (each by direct A/B on the failing cell): penalties ✗, stale penalty history ✗ (real defect, fixed), beyond-accept state drift ✗ (real defect, fixed), fast-greedy tie-breaks ✗, graph replay ✗, w4a16 tiers ✗ (byte-clean), GDN chain via exact-verify ✗, w8a16 batch GEMVs **✓ real defect, fixed, insufficient alone**.

rsafier:
## DFlash2 accept campaign: drafter PROVEN correct; the ceiling was the K=8 GDN verify path — `--exact-verify` required

**Parity vs the z-lab PyTorch reference (CPU, exact dumped inputs):** the Atlas drafter forward is clean at every op of every layer (cos ≥ 0.9999, convs included), and the host selector walk reproduces the reference picks token-for-token. The drafts were never the problem — draft/emit alignment shows near-perfect drafts while the **target-side emission derails** under DFlash verify ('one, two, three, and the second is the number of the second element… public static' at temp 0, think-off). No-spec control is perfect. `ATLAS_NO_GDN_FP8_PREFILL=1` does NOT fix it. **`--exact-verify` (#435 route (a)) does**: output exact, 13/20 tokens drafted on the count prompt (accept_len ≈1.86 drafted-content).

So the default WY-chunkwise GDN verify arms are **grossly wrong at DFlash's K=8 width on qwen3.8** — same family as the K=4 MTP near-tie divergence documented earlier in this PR, but catastrophic at this width. Per-width GDN verify kernel audit is the follow-up (#395's general-K family is the likely fix vehicle).

**qwen3.8 DFlash2 serve recipe as of this branch:**
```
--dflash --draft-model incoai/Qwen3.8-27B-DFlash2 --exact-verify
ATLAS_DFLASH_OPTION_B=1   # REQUIRED: legacy path leaves the convs dormant
util ≤ 0.70               # pre-flight gate enforces the footprint
```
γ auto-resolves to the trained block size. Probe (think-off, temp 0): 35% of completion tokens drafted; code leg accept_len 1.01 @ 11.6 tok/s. Cumulative accept progress this campaign: **1.1% → 35%** of tokens drafted with correct outputs.

**Known next levers:** checkpoint-native FP8 lm_head for the drafter tail (unsloth ships lm_head in FP8 — no mirror build needed, halves the largest per-propose GEMM); GPU selector (removes 4 MB D2H + host walk per propose); per-width GDN verify kernels; prose-leg trajectory audit.

rsafier:
## Investigation notes: DFlash default-chain (`K=γ` verify) garbage output — root cause localized to SSM state loss, NOT the GDN kernels

**Symptom** (why the recipe currently requires `--exact-verify`): with the default verify chain, qwen3.8+DFlash2 derails into junk right after the first few tokens (`'one, two, three.\n 104 | …'`), accept ≈0%. `--exact-verify` produces correct output.

### What is now EXONERATED (each by direct experiment)
- **All GDN verify kernels at the failing geometry.** New microtests in this branch (`gdn_verify_fused_qwen38_microtest`, `gdn_conv_kn_qwen38_microtest`, NV=48/K=8): fused wy2 cos=1.0, BF16 fused `conv_kn` byte-identical; `gdn_wy_verify_microtest` validates wy5..wy8 outputs *and* committed state/snapshots at ≥0.99999.
- **wyN arm vs sequential fallback**: both produce the same junk (`ATLAS_GDN_WYN=0` leg) — it's not one kernel arm.
- **BA-gates / batched norm** (`ATLAS_NO_BATCHED_BA_GATES`, `ATLAS_NO_BATCHED_GDN_NORM`): junk persists.
- **CUDA graphs**: junk persists fully eager (`ATLAS_DFLASH_DEBUG_NO_GRAPH=1` — note `ATLAS_NO_DECODE_GRAPHS` does NOT govern the dflash verify graph).
- **Rollback indexing**: shifting the `commit_accepted_prefix` restore index ±1 changes nothing; both readers (`verify_a.rs`, `async_chkpt.rs`) are consistent with the writers.
- **Projections / gates / deinterleave**: tensor dumps (below) show them byte-identical between the junk and clean runs.

### The actual defect (tensor-dump evidence, `ATLAS_GDN_VERIFY_DUMP` instrumentation)
Dumped the GDN verify chain's inputs/outputs per layer for default vs `--exact-verify` runs at temp-0 on the same prompt (runs verify-step-aligned; `deint`/`gates` byte-identical):

- **The first `K=γ` verify enters with layer-0 `h_state` ≡ 0 and `conv_state` ≡ 0.**
- A prefill-side probe proves prefill wrote real state to the SAME pointer (`h_state=0xfe0342200000`, L2=13.6 after the 23-token prompt) that reads back **all-zero** at verify entry: **the sequence's SSM state is zeroed between prefill-end and the first DFlash verify.** Every "GDN corruption" symptom is downstream of running the verify from empty recurrent state, with per-step drift compounding through the bonus-token-only path.

### Why earlier width-bisects were misleading
- γ≤4 (K≤5) legs looked "clean" — but their logs contain **zero** `DFLASH K=γ verify` lines: the MTP gate arbitrated them onto the standard K≤4 MTP verify path, which handles state correctly. Only γ≥5 engages `verify_dflash_step`/`verify_d`, and every such leg is junk.
- `--exact-verify` at K=8 *does* run the same path; its first verify also enters with `h=0`, but its `conv_state` survives (allocation-layout luck — exact stages extra modules) and on this short prompt full attention carries the output. It is NOT a kernel-correctness proof.

### Current status
Bisecting the prefill-end → first-verify window with state probes (propose entry/exit) to identify the zeroing writer — prime suspects are a memset/OOB sweep in the DFlash propose or drafter-prefill capture path whose blast radius is allocation-layout-dependent.

Instrumentation (env-gated, off by default): `ATLAS_GDN_VERIFY_DUMP=<dir>` dumps per-layer verify chain tensors + prefill h-state pointers. Will be finalized or stripped before merge.

rsafier:
## ✅ ROOT CAUSE FOUND AND FIXED — DFlash default chain now correct at full γ (`3d95e532`)

Follow-up to the investigation note above: the "K=8 GDN corruption" was never a kernel bug. It was **the seam dispatch zeroing the sequence's live SSM state on the first verify**.

### The defect
`mtp_step.rs`'s seam step (first propose after prefill/bootstrap) dispatched the DFlash γ-block init (**7 drafts**) into `step_verify_k4`, which verifies only `[last, d0..d2]` (4 rows) but hands the FULL init slice to `k4_apply_verdict`:

- `k_rows = drafts.len() + 1 = 8` (should be 4)
- a 3/3 full accept → `commit_accepted_prefix(4, 8)` → the full-accept no-op (`na == k`) never fires
- the commit "rewinds" to **h/conv intermediate[3]** — an index a K=4 verify *never writes* (K-1 dead-write shrink)
- → the pool's **zero-initialized snapshots get restored over the live prefill state**, h + conv, all 48 GDN layers, on the sequence's first verify. Everything after decodes from empty recurrent state.

The `debug_assert_eq!(v.len(), k_rows)` tripwire is compiled out of release builds. Caught red-handed by the commit-time dump: `commit na=4 k=8 inter_idx=3` with a zero source, immediately after `fused_exit h=18.255` (verify itself was fine).

### Why every earlier suspect alibied out
- **All GDN arms produced identical junk** because they all ran *correctly* on a state that had been zeroed one step earlier.
- **`--exact-verify` "fixed" it** only because its fused conv epilogue writes all K snapshots (so `inter[3]` held real conv data → conv survived) and zero-H is survivable at short context. It was never a numerics fix.
- **The γ≤4 "clean" legs** dodge the bug because `num_drafts+1` happens to equal the actual verify width there.

### The fix (`3d95e532`)
- seam dispatch routes `init.len() >= 4` to `step_verify_dflash` (mirrors the existing Phase-B guard)
- `k4_apply_verdict` now derives `k_rows` from `v.len()` (the verified rows — ground truth) and hard-asserts the draft slice covers it (release-mode tripwire)

### Validation (gx10, γ=8, temp-0 count prompt, default chain)
| config | before | after |
|---|---|---|
| eager | junk, acc 3, first verify 0/7 | `'one, two, …, ten'`, per-step accepts 7/7, 1/7, 6/7 |
| CUDA graphs (production) | junk | `'one, two, …, ten'`, **acc 15** |
| state at first verify | h=0, conv=0 (all layers) | h L2=13.6, conv L2=778 (== prefill values) |

**`--exact-verify` should no longer be required in the DFlash recipe** (keeping it remains a valid exactness opt-in; it costs ~+22-36% decode). Will follow up with an accept-rate/tok/s A/B of the default chain vs exact against the SGLang reference numbers.

Diagnostics kept (env-gated, `46feed5f`): `ATLAS_GDN_VERIFY_DUMP=<dir>` — per-layer verify-chain tensor dumps + prefill/propose/verify/commit state-timeline probes.

rsafier:
### Post-fix A/B: default chain vs `--exact-verify` (γ=8, CUDA graphs, FP8 drafter tail, temp-0, same binary)

| prompt | default tok/s | exact tok/s | default accepted | exact accepted | text |
|---|---|---|---|---|---|
| count (20 tok) | **30.9** | 29.0 | 15/20 | 15/20 | identical |
| code (161 tok) | **10.5** | 9.8 | 66/161 (41%) | 62/161 (39%) | identical |
| prose (~230 tok) | **8.4** | 8.3 | 63/246 (26%) | 62/223 (28%) | same content, near-tie length diff |

**Recipe update: drop `--exact-verify`.** The default chain is now faster on every leg with equal-or-better accepts and identical greedy text on count/code. Updated qwen3.8+DFlash2 serve stack:

```
--dflash --draft-model <incoai/Qwen3.8-27B-DFlash2> \
ATLAS_DFLASH_OPTION_B=1 ATLAS_DFLASH_DRAFTER_FP8=1
```
(γ auto-resolves from the checkpoint; util ≤0.70 on GB10.)

Reference point: SGLang DFlash2 K7 think-off dedicated-C1 is 28.38 tok/s — the count-leg 30.9 clears it; long-form legs are decode-bound elsewhere (GEMV chain), not in the verify.

rsafier:
## Second big win: drafter-ctx starvation found and closed — code leg 10.5 → 36.8 tok/s (+250%)

The step profile showed accept/step collapsing on long-form (0.34–0.66 vs ~1.9 for the same weights under SGLang). A draft-vs-emitted alignment audit showed the drafter **replaying stale context** (drafting text emitted ~20 tokens earlier), with accept decaying over the generation (code quarters: 1.79 → 0.53).

Root cause (already documented in-tree, default-off): in the default config the K=γ verify captures only **row k-1** — the last, almost-always-rejected draft — and no ctx commit runs (`verify_dflash_step.rs:133-153`, `verify_d.rs:300-318`). The drafter ctx window fills with wrong-token hiddens and lags the true sequence. The in-tree remedies `ATLAS_DFLASH_UNIFIED_CTX=1` (structural) / `ATLAS_DFLASH_EAGLE_FIX=1` capture all K rows and append rows 0..=num_accepted in EAGLE order.

### A/B (γ=8, graphs, FP8 tail, post-3d95e532, temp-0)

| leg | default ctx | UNIFIED_CTX=1 | accept |
|---|---|---|---|
| count | 30.9 tok/s | **40.5** | 15/20 → 17/20 |
| code | 10.5 tok/s | **36.8 (+250%)** | 41% → **84%** (~5.3 acc/step) |
| prose | 8.4 tok/s | **19.3 (+130%)** | 26% → 68% |

(EAGLE_FIX=1 gives byte-identical results on these legs.)

Code leg 36.8 tok/s vs SGLang DFlash2 reference 28.38 (dedicated C1, think-off). **Recipe update:**

```
--dflash --draft-model <incoai/Qwen3.8-27B-DFlash2>
ATLAS_DFLASH_OPTION_B=1 ATLAS_DFLASH_DRAFTER_FP8=1 ATLAS_DFLASH_UNIFIED_CTX=1
```

Next: considering making unified-ctx default-on for the DFlash path (opt-out), pending longer-context/agentic validation. Remaining profiled levers: verify=115ms/step (M=8 GEMV tier + MoE at 8 rows) and propose=47ms (host selector + drafter tail) — analysis in flight.

rsafier:
## Perf sprint progress (post-root-cause)

**Landed:**
- `25331426` — **unified ctx commit DEFAULT-ON** (opt-out `ATLAS_DFLASH_UNIFIED_CTX=0`). No-flag validation reproduces the ctx-fix numbers: count **41.0**, code **37.1**, prose **19.5 tok/s**. The recipe no longer needs any ctx env flag.
- `ce9a671d` — **native FP8 lm_head share for the drafter tail**: unsloth ships lm_head natively FP8 E4M3 (padded 248320×5120 + per-row scale); the tail now reads those resident bytes instead of building a 1.27 GB FP8→BF16→FP8 mirror. −1.27 GB resident, −2.54 GB pre-flight reserve, one less lossy requant; outputs and accepts byte-identical.
- `9bdd8ced` — cherry-pick of #549 (MTP accept-width buckets; both regression tests pass here).

**Measured (propose-side, `ATLAS_DFLASH_PROPOSE_TIMING=1`, new in-tree):** propose ≈ 40.5ms = **fwd 34.5ms** + selector 6.1ms + setup/d2h ≈ 0. The unattributed cost sits inside the drafter forward (5 layers ≈ 12.5ms weight-traffic floor) — sub-loop attribution (pre/post graph launches vs eager attention) running now.

**Verify-side correction:** qwen3.8-27B is **dense** (no experts; FFN 17408 × 64 layers) — the earlier "MoE, batchm-inapplicable" verdict was wrong on this model. Dense FFN traffic ≈ 8.6 GB/step at NVFP4 ≈ ~62ms of the 115ms verify at batch8's 138 GB/s. `examples/batchm_bench` (cold-cycled, exact shapes incl. gate/up/down and lm_head at M=8) is running to pick the right tier/tile fix.

Next: propose sub-loop numbers → GPU selector port (6.1ms) + seam fix; batchm verdict → verify FFN tier change.

rsafier:
## Perf sprint complete — all five approved items landed or ground-truthed

| # | item | commit | outcome |
|---|---|---|---|
| ① | unified ctx commit **default-on** | `25331426` | code 10.5→37.1 tok/s (+250%), prose +130%, count +31% — no flags needed |
| ② | native FP8 lm_head **share** | `ce9a671d` | −1.27 GB resident, −2.54 GB pre-flight reserve, one less lossy requant; outputs identical |
| ③ | propose seams attributed | `b333c56b`, `c61dd932` | launch-overhead theory **refuted** by mono-graph A/B (neutral at C=1; shipped opt-in `ATLAS_DFLASH_MONO_GRAPH=1`); real propose = layers ~22ms + tail ~13ms + selector ~6ms |
| ④ | **GPU CandidateSelector** | `24a5ee38` | selector 6.8→0.4 ms, propose 41.6→36.4 ms; parity 0 mismatches; tok/s 43.0/38.0/19.9 (`ATLAS_DFLASH_GPU_SELECTOR=1`, default flip after soak) |
| ⑤ | verify-side ground truth | bench fix in `c61dd932` | cold-cycled M-sweep on real shapes: **M=8 sits at 44-60% of peak everywhere** (batch8 qkv/o 126 GB/s, ffn 140-151, lm_head 144; gemm_t wins lm_head at 194). qwen3.8-27B is **dense**, so the FFN's ~8.6 GB/step ≈ 65ms of the 115ms verify rides this cliff |
| — | #549 cherry-pick | `9bdd8ced` | accept-width buckets; both regression tests pass |

**Current state of the DFlash2 recipe** (γ auto, util ≤0.70):
```
--dflash --draft-model <incoai/Qwen3.8-27B-DFlash2>
ATLAS_DFLASH_OPTION_B=1 ATLAS_DFLASH_DRAFTER_FP8=1 ATLAS_DFLASH_GPU_SELECTOR=1
```
temp-0 think-off: **count 43.0 / code 38.0 / prose 19.9 tok/s** — vs 30.9/10.5/8.4 at the start of the sprint and the SGLang DFlash2 reference of 28.4/23.5 (dedicated/ladder C1).

**Next lever (scoped, not started): the M=8 GEMV tier.** batchm_bench numbers say a row-parallel/smem-staged MAX_M=8 tier at batch4-class efficiency (~75% peak) is worth ~20ms of the 115ms verify — the largest remaining single win. Second: drafter-layers deep-dive (22ms for 5 layers; needs nsys, not more host timers).

rsafier:
Rebased onto current main (`048dc6bb`, 22 commits replayed clean — main delta was governance-ledger + two bench-doc commits) and **GPU CandidateSelector promoted to DEFAULT-ON** (`f49d99d5`; `ATLAS_DFLASH_GPU_SELECTOR=0` opts out, `=2` parity harness) after the zero-mismatch parity campaign + agentic soak on the live serve. Rebased tree builds clean all-targets. Recipe is now just: `--dflash --draft-model <ckpt>` + `ATLAS_DFLASH_OPTION_B=1 ATLAS_DFLASH_DRAFTER_FP8=1`.

rsafier:
## Machine-hang root cause + fix (`0443ed7f`)

Two hard hangs (02:32/02:44 EDT) at 256K-ctx/C=8 + the concurrency benchmark. Forensics: boot-only ledger at those flags showed the resident set at ~97 GB before ANY request (my watchdog killed the repro at 14 GB avail); the benchmark's first lazy allocations finished the box, and **earlyoom can never fire against pinned CUDA memory** (kill condition requires swap ≤80%, but pinned pages don't swap — journal shows 'mem avail 0%, swap free 87%' until the hang).

Budget defects fixed:
1. **DFlash drafter head allocs (~10 GB) landed AFTER KV sizing, outside the util pledge** → build_model now reserves the estimate out of the KV budget (mirrors the serve pre-flight formula).
2. **SSM verify pools sized at the legacy 17-wide γ ceiling** → now sized from the resolved drafter γ (`config.dflash_gamma`); pools 24 GB → 8.1 GB at γ=8/C=8.

Post-fix ledger, same flags: **READY with 21 GB available** (was <14 and falling), KV pool honestly sized inside the budget, clean teardown.

Ops note for GB10 boxes: earlyoom's default swap condition makes it useless against inference servers — configure it to trigger on mem alone (or put a systemd MemoryMax on the serve) so operator-error configs get a dead process, not a dead machine.

rsafier:
## M=8 GEMV tier: measured negative result (`8b9d0fb1`) — and the real question it surfaces

Built two bit-exact M=8 prototypes (both pass a new BIT-EXACT-vs-batch8 gate at every M, all four shapes):
- **batch8_rp** (row-parallel, 2×4-row groups sharing the weight stream): loses ~10% — duplicate per-group dequant + halved output amortization
- **batch8_pp** (phase-parallel, the template's two K-phases on different thread halves, zero duplicate decode): ~neutral — qkv/o 116.0 vs 113.6µs, ffn_down +7%, others slightly worse

**Diagnosis confirmed by construction:** the family at M=8 is bound by per-SM instruction throughput (~164 inst per 8-byte weight chunk per output). Thread redistribution is invariant under that bound. The scoped ~20ms/verify-step win requires cutting instructions per weight byte — packed HFMA2 math or gemm_t-style tiled decode — and **every such change alters the FP32 accumulation order, i.e. breaks the bit-parity contract** (`w4a16_batch_bitparity_microtest`) this family is gated on.

**Team decision needed:** relax bit-parity for the *verify* path (drafts are verified anyway; gate on accept-rate/argmax stability instead) and unlock the packed-math tier, or keep parity and accept the M=8 ceiling. Prototypes + bench gates stay in-tree so the next attempt starts from the measurement.

Ledger note: the honest per-step picture after today: verify ~115ms (bounded as above), propose ~30ms (post GPU-selector), accept 68-84%. Recipe tok/s: 43/38/19.9 (count/code/prose).

rsafier:
## C=1/2/4 concurrency + prefill report (current head, util 0.65, 300-tok legs, temp-0 think-off)

| leg | aggregate | per-stream | accept |
|---|---|---|---|
| C=1 | **39.0 tok/s** | 39.0 | **84%** |
| C=2 | 24.1 | 12.0 / 14.3 | 31% ⚠ |
| C=4 | 28.9 | 13.9 / 7.2 / 8.1 / 7.5 | 22% ⚠ |
| prefill | 714 tok/s @2.2K, 717 @8.7K (flat) | | |

(SGLang DFlash2 K7 reference: 23.5 / 36.1 / 55.0 at C1/2/4-ladder.)

**C=1 beats every reference. C≥2 exposes the next defect**: accept collapses because the whole-prompt hidden capture is **single-slot** with generation-stamp ownership — at C≥2 every sequence but the last prefills *blind* (the documented 'blind beats poisoned' trade), drafts without context, and near-zero-accept K=8 verifies cost more than plain decode. Filed as the next lever: per-slot capture (or capture-window), with `ATLAS_DFLASH_ADAPTIVE=1` as an interim mitigation. Prefill numbers are the no-FI-GDN floor (`ATLAS_GDN_LIB` unset — the fails-open trap).

Also in this push: `785d1f6b` — the γ-sized-pools commit had flipped the pool-uniformity *inference* to tiered, 500'ing the first C-report attempt ('SSM MTP intermediate buffers not allocated'). Uniformity is now stated explicitly for DFlash pools. Lesson re-learned: state-pool changes get a completion probe, not just a boot ledger.

rsafier:
**C≥2 accept collapse ROOT-CAUSED + FIXED (090c83d3)**

At C≥2 the scheduler routes sequences through **batched decode** (`decode_multi_seq`), which never captured the per-layer hiddens DFlash conditions on — and `step_decode_only` only committed ctx when `n == 1` (a stale "DFlash runs --max-batch-size 1" assumption, spelled out in its own comment). Every batched decode step advanced each sequence's position while appending **zero** ctx rows, so the drafter's context fell behind (`CTXLEN_PROBE GAP=142` on the worst stream) and accepts collapsed; adaptive-spec suspension then routed the starved seq back into batched decode — a self-reinforcing spiral. C=1 never touches that path, which is why it stayed healthy.

Fix: capture **all batch rows** inside the decode graph (`try_dflash_capture_all(padded_n)`, replay-safe — fixed src/dst addresses), give `commit_ctx` a scratch-row param, and commit each seq from its own row at any n. A capacity guard degrades to a warned ctx hole (instead of committing stale rows) if a batch ever exceeds γ+1 rows. Per-step `CTX_COMMIT`/`CTX_VERIFY` debug ledger kept at `tracing::debug`.

Measured (same-boot A/B, GDN FlashInfer lib engaged):

| leg | before | after |
|---|---|---|
| C1 code / prose (controls) | 83% / 72% | 83% / 72% (unchanged) |
| C2 code | **2/193 = 1%** | **124/161 = 77%** |
| C2 prose | 111/300 = 37% | 181/300 = 60% |

All CTXLEN_PROBE lines now GAP=0; C2 code-stream wall time 20.8s → 7.1s. Full C=1/2/4 + prefill report re-run in progress.

rsafier:
**Second C≥2 root cause FIXED (f1e5215e): drafter KV pool winner-takes-all + Option B default-on**

The drafter's paged KV pool holds ONE max-seq-len sequence's worth of blocks (~2049 @32K), and `propose()` eagerly allocated `max_ctx_len + γ` blocks up front — the first sequence took the entire pool and every other stream failed at `block 0/2049`, drafting blind for its whole overlap. C=4 signature was literally winner-takes-all: **59% / 5% / 4% / 7%**. The failing path also leaked its partial grab. Now blocks grow incrementally by actual ctx need (64-block chunks, device table max-sized once), shortfall returns the partial grab and degrades to non-spec for that step.

**Cold-start (salted, cache-cold) accept, C=4: 59/5/4/7% → 54/35/35/31%**, zero exhaustion lines. `ATLAS_DFLASH_OPTION_B` is now default-on (=0 opts out) — every working recipe set it and the KV budget already reserves for it.

**Consolidated report on this head (GDN FlashInfer lib engaged):**
| leg | tok/s | accept |
|---|---|---|
| C=1 | 37.6 | 138/161 = **86%** |
| C=2 | 25.4 | 225/600 = 38% |
| C=4 | 28.1 | 502/1125 = **45%** (was 16-22%) |
| prefill 2K / 8K | **882 / 887 tok/s** (was ~715 without the GDN lib) | |

**Known residual (tracked, not blocking):** C≥2 accept is bistable across boots (30-77% band, path-independent — legacy ≡ Option B, GAP=0, text fully coherent). The temp-0 token stream itself forks under concurrency (same prompt: 161-token answer with EOS in one attractor, 300-token cap in the other) — same family as the known qwen3.8 spec-decode greedy divergence. Next step is token-level fork localization (which step type — batched decode vs serial vs verify-bonus — emits the first diverging token). Also queued: Q12 kernel-batched prefill sets drafter `ctx_len = prompt_len` without running the per-layer ctx captures (zeros seed) — needs the capture wired with per-stream `proc_off` + a captured-rows watermark clamp.

Theories eliminated with receipts along the way: Q12 batched-prefill capture skip as the collapse driver (forcing per-stream didn't recover), prefix-cache restore (cached leg healthy), adaptive-spec suspension (lever off, 0 suspensions), whole-prompt capture generation races, trim/after_verify cross-seq state.

rsafier:
**C≥2 residual ROOT-CAUSED + FIXED (773d911c): mtp_gate arbitration was forking temp-0 streams**

Fork forensics (new per-token `TOK` debug ledger in `emit_token` + step-type attribution) closed the last mystery. The residual C≥2 accept collapse and boot-to-boot bistability were the **mtp_gate throughput arbiter**:

- At C=2, ~45% of tokens were emitted by the plain-decode arm (gate in decode mode via hysteresis/dwell) — accept dilutes 72%→32% *by construction*, no corruption involved.
- The serial↔batch-K forward flips **fork the temp-0 token stream mid-answer** (first fork localized exactly at a gate flip). The bad attractor degenerates into period-2…64 repetition loops that the content-loop watchdog kills — a user-visible quality defect, not just a benchmark artifact.
- `ATLAS_MTP_GATE_FORCE=1` A/B: C=2 recovers to **75% accept at par tok/s** (25.4→24.4) with streams EOSing normally; C=4 forced-verify holds **72% accept** — proving there is **no state corruption anywhere at C≥2** — but loses aggregate (22.0 vs 28-30 tok/s), so arbitration stays for C≥3.

Shipped: `dflash_gate_pin_c2` (default ON, `ATLAS_DFLASH_GATE_PIN_C2=0` restores arbitration) — pins the verify arm for DFlash at active ≤ 2, riding the existing entry-pin (pinned steps skip gate recording so C≥3 baselines stay clean).

**Final default-config report (this head):**
| leg | tok/s | accept |
|---|---|---|
| C=1 | 37.7 | 138/161 = 86% |
| C=2 | 25.0 | **335/445 = 75%** (was 38%; per-stream balanced [17.6, 16.8], EOS reached) |
| C=4 | 30.1 | 40% (arbitration regime — expected) |
| prefill 2K/8K | 884 / 888 tok/s | |

The C≥3 lever is now clearly **cross-sequence batched DFlash verify** (verify loses arbitration only because it runs per-seq serially); a small known follow-up remains on batched (Q12) prefill skipping the drafter ctx captures (zeros seed).

rsafier:
**Cross-sequence batched DFlash verify SHIPPED (5150a088) — C=2 +70%, C=4 +46%**

Root cause of the C≥3 throughput loss, measured first: DFlash had **zero concurrency amortisation**. The per-step verify wall was flat at ~115ms and propose at ~40ms from C=1 through C=4 — every sequence paid its own full weight sweep, so a C=4 round cost 4×155ms. That is exactly why the gate arbitrated verify away at C≥3 despite it holding ~72% accept.

Now `n` sequences' `[last_token, d0..]` rows go through **one R=n·k-row target forward**. Every weight-bearing op (QKVZ / o_proj / FFN / lm_head) reads the weights once for the batch. Only the GDN recurrent body stays per-sequence — K=γ+1 has no fused WY kernel, so `decode_batched_conv_gdn_multi` cleanly returns `Ok(false)` at that width and takes the byte-identical per-sequence fallback the single-seq K=γ path already used. **Accepts are unchanged by construction; only the wall moves.**

Kernel evidence for why this works: `gemm_t` costs **3679 µs at M=9 and 3837 µs at M=36** — 4× the rows for +4%.

**Default config (gate on), same boot, max_tokens=300 temp 0:**
| leg | before | after | accept |
|---|---|---|---|
| C=1 | 37.7 | 37.5 | 86% (unchanged — single-seq path) |
| C=2 | 25.0 | **42.4** (+70%) | 75 → 79% |
| C=4 | 30.1 | **44.0** (+46%) | 40 → 70% |
| prefill 2K/8K | 884/888 | 882/891 | unchanged |

C=2 (42.4) now **exceeds C=1** (37.5) — the first real concurrency scaling this drafter has shown. At C=4 batched verify beats the decode arm outright (44.0 vs 30.1), so the gate now picks verify on its own; no pin extension needed.

Structural piece worth flagging for review: the DFlash hidden capture buffer had **no sequence dimension** — a batched forward would have had every sequence overwrite band 0. It is now `max_batch_size` K-row bands, sequence i capturing into band i, and the scheduler commits that band through `commit_ctx`'s `scratch_row` with `dflash_capture_band()` as the shared stride SSOT. Phase order in the new step is load-bearing (same hazard `verify_k4_batch_step` documents): all shared-buffer reads finish for every sequence before the first propose overwrites `hidden_states`.

One trap worth recording: the width gate initially pinned k to exactly γ+1, but the drafter returns **γ−1** drafts on a propose — so `can_batch_verify` silently refused every batch (444 serial steps, zero batched, no error). It now accepts any uniform k the capture band holds, and a first-decline log makes that class of failure visible.

Kill switch `ATLAS_DFLASH_BATCH_VERIFY=0`. **Next lever:** propose is now the dominant per-step term (80ms at n=2 vs 103ms for the whole batched verify) and is still per-sequence — `BlockDiffusionDraftHead` implements neither `propose_batch` nor `propose_batch_max`.

rsafier:
**Batched DFlash propose SHIPPED (c4d66bca) — C=4 44.0 → 49.7 tok/s, peak 55.6**

Closes the last per-step serialization. After batched verify, propose was the dominant term at **36 ms per sequence** (23 ms drafter layers + 13 ms tail, both weight-bearing), scaling linearly with n.

DFlash is a *block-diffusion* drafter — one forward emits all γ drafts — so batching means n·γ rows in a single drafter forward. Every weight-bearing op (conv, q/k/v, o_proj, gate/up/down, and the lm_head against a 248k vocab) now runs **once** for the whole batch.

Two design calls worth flagging for review:

- **Attention stays a per-sequence loop on purpose.** It reads no weights — only Q/K/V scratch and each sequence's own KV pages — so batching it saves nothing but launch overhead. That removed the need for a batched paged-attention kernel entirely, which was the piece I expected to be hardest.
- **The batched path runs eager.** The propose CUDA graphs are captured at the single-sequence γ-row geometry with scratch pointers baked in; replaying one for an n·γ batch would silently compute band 0 only. It still wins, because the drafter weights are read once instead of n times.

Per-sequence prep (ctx append, Option-B block growth, incremental ctx precompute) also stays per sequence — it only touches the uncommitted ctx tail, typically one row per step — and reuses `propose_drafts`' own prep via a `collect_prep` sink so the two paths cannot drift.

**Two bugs the first end-to-end run caught**, both diagnosed from the per-stream accept split (band 0 healthy, later bands degraded — C=2 `[37.1, 17.3]`, C=4 `[23.4, 13.7, 11.1, 8.7]`):

1. The final RMSNorm still covered only γ rows, so every band past the first reached the lm_head **unnormalised**.
2. The FP8 lm_head used `fp8_gemm_t_row_scaled_m16` — a single-warp M_TILE=16 kernel — which a 32-row batch violates. Above the tile it now uses the general row-scaled GEMM.

Accepts recovered to **74% at C=4, above the 70% pre-batching baseline**, which is the evidence draft quality is intact rather than merely fast.

**Default config (same boot, max_tokens=300, temp 0):**

| leg | before | after | accept |
|---|---|---|---|
| C=1 | 37.5 | 37.5 | 86% (unchanged path) |
| C=2 | 42.4 | **47.8** (+13%) | 79% |
| C=4 | 44.0 | **49.7** (+13%) | 68% |
| C=4 gate-forced | — | **55.6** | 74% |
| prefill 2K/8K | 894/896 | 894/896 | unchanged |

Propose wall at n=2: 80 ms → ~43 ms. `propose_batch` declines to `Ok(None)` on any doubt (missing selector, n > max_batch, prep failure, short result), so the per-sequence loop stays the always-available fallback.

**Cumulative for this branch:** C=2 25.0 → 47.8 tok/s, C=4 30.1 → 49.7 tok/s, prefill ~715 → ~895 tok/s.

rsafier:
**GPU allocation ledger SHIPPED (f7018d04, 07361cb7) — memory now attributes itself**

The CUDA backend's ledger was a `HashSet<u64>` of pointers. It could say how many allocations leaked, never how many bytes, and never which code made them — its own doc conceded the gap: *"their total bytes is not tracked (the driver does not report per-pointer size), so the count is the diagnostic"*.

That is why every GPU-memory question on this branch has been archaeology. A 27B serve reports 58.9 GB consumed before the KV decision against 21.8 GB of weights, and nothing attributed the difference; the four size-a-buffer-from-a-ceiling bugs found so far (~64 GB between them) were each located by reading allocation code rather than reading a number.

Both missing facts were free. The size is already an argument to `cuMemAlloc_v2`; the call site comes from `#[track_caller]` on the two allocating trait methods — declared on the **trait** as well as the impl, since nearly every caller goes through `&dyn GpuBackend` and the vtable would otherwise attribute the whole process to one line inside the backend.

`build_model` logs it once at INFO at end of load — the only point where the ledger describes the steady state rather than a midpoint:

```
GPU allocation ledger: 81.18 GB live across 91 sites
  22332.5 MB x1968  fast_weights/mod.rs:361        <- the weights
   9216.0 MB x48    ssm_snapshot.rs:158            <- Marconi, 64 slots
   9180.0 MB x192   dense_ffn.rs:641               <- nvfp4_mmq_repack
   5625.0 MB x65    weight_map/quant_helpers.rs:98
   5065.7 MB x185   weight_map/loaders_fp8.rs:210
  ── by file ──
   2221.2 MB x16    vision_encoder/enc_impl/init.rs
```

The per-**file** rollup (07361cb7) exists because the by-site view has a blind spot: a subsystem allocating many buffers from many lines splits into pieces small enough to vanish into the remainder. The vision encoder is exactly that shape — ~16 buffers, 2.2 GB, the fourth-largest consumer in the process, and completely invisible in a top-12 by site. Found by noticing it was missing from a report it should have led.

Two things it measured immediately:

* **`dense_ffn.rs:641` holds 9.2 GB of repacked duplicate FFN weights** (`nvfp4_mmq_repack`, cached in a `OnceLock`, alongside the loaded layout). Investigated separately: it is **not** removable — it is the prefill layout while decode reads the originals, the dead `_t` twins are already freed pre-KV-sizing, and its true net cost is ~1 GB (repack 9180 MB replaces twins worth 8160 MB), which buys +8% prefill and +11% C=4 decode. Keep it.
* **The util pledge overshoots by ~2 GB** — 81.18 GB live against a 79.1 GB budget, because ~16.7 GB is allocated after the KV sizing decision and the 14.6 GB preflight reserve does not quite cover it. This is the measured replacement for the inferred "~11 GB outside the budget" figure.

`sweep_unreleased()` now reports leaked **bytes** and names the top sites instead of only counting allocations. No hot-path cost — allocation is a load-time event; the added per-allocation state is one `usize` and one `&'static Location`.

rsafier:
**LoRA on qwen3.8-27B: from "no community adapter loads" to all three, GPU-validated (93de72e6, 225143c4, 74581e3d)**

Three real Qwen3.8-27B adapters were tested end-to-end: `Dxniz/Novelist1.0-27b-Adapter`, `nico248000000000/…-cyber-LoRA`, `MuXodious/…-absolute-heresy-LoRA`. None of them worked. Four separate reasons, each fixed.

### 1. Dense-FFN deltas were loaded and never applied

`dense_ffn.rs` declared a `lora` field, `set_lora_weights` wrote it, nothing ever read it — `#[allow(dead_code)]`, doc reading *"M0: stored only — compute reads land in M1"*. An FFN-targeting adapter loaded, reported success, and changed no output. On a hybrid that is most of the adapter: these three put **67–78% of their parameter mass** in the FFN.

M1 applies them across every NVFP4 dispatch (decode `forward`, `forward_k2/k3/km`, `forward_prefill`, `forward_batched`). Partial coverage was never an option — an adapter that applies in prefill and not decode is a model that contradicts itself mid-sequence.

Two arms are **pinned** under an adapter, both because they never materialise the tensor a delta needs:
* decode's fused `w4a16_gemv_silu_input` never materialises `silu(gate)*up` → the split-SiLU path is pinned (already the default), and `set_lora_weights` refuses an adapter the layer cannot serve that way, so the pin is a guarantee.
* prefill's FP4-MMQ arm leaves gate/up **unscaled** and folds `weight_scale_2` inside the SiLU-mul. A LoRA delta is true-valued, so adding it there pushes it through a scale multiply that is not its own — **silently wrong output, not a failure**.

### 2. "Full-attention layers only" is false for this architecture

Qwen3.8-27B is **64 layers — 16 full-attention + 48 linear-attention — and all 64 carry the dense FFN**. The adapter tensor counts prove it: 32 (=16×2) for q/k/v/o, 96 (=48×2) for `out_proj`, 128 (=64×2) for FFN. Both adapters died on `layers.48.mlp.up_proj`, and the reject could only suggest retraining with `layers_to_transform`.

The MoE side already carried this exception in these words — routed experts live on every MoE layer, so gating them to full-attention *"would wrongly reject the majority of a MoE adapter's layers"*. Dense FFN on a hybrid is that argument again. `Qwen3SsmLayer` now installs FFN deltas into its `FfnComponent::Dense`, which is the same `DenseFfnLayer` the attention layers hold, so the M1 path runs there unchanged.

### 3. The slot pool silently dropped those pairs

Accepting them was not enough: `pool_slot_bytes` and `pack_slot` still walked `full_attention_layers × ALL modules`, so the pairs had nowhere to land — accepted, never packed, never applied. **All four consumers of the layout** now share one predicate (`LoraModule::applies_to_layer`): pool sizing, the packing walk, `module_slot_offsets`, and the RDMA `rebuild_slot_layers`. The last two were caught by the layout tests only after the first two changed — `module_slot_offsets` would have landed staged weights at stale offsets. Slot layout changes 16_515_072 → 15_335_424 bytes @ max_rank 16; the golden is updated with that reasoning and the boundary test now derives the last packed module instead of hardcoding `down_proj`.

### 4. GDN `out_proj` is now supported

`LoraModule::OutProj` (value_dim → hidden). It is the block's **last** stage, downstream of the recurrence, so a delta there is an ordinary per-token linear delta that never enters the state update — which is why it does not need the exact-replay parity harness gating the rest of the family. `in_proj_*` / `conv1d` **do** feed the recurrence and stay rejected.

The fold lives **inside `ssm_tp_all_reduce`**, deliberately: it must run *after* the reduce (before it the buffer is only this rank's partial row-parallel product, so a delta would be summed once per rank and scaled by TP width), and all five SSM dispatch paths finish `out_proj` through that one function, so no path can be wired while another is forgotten.

Also: PEFT **regex** `target_modules` is accepted (held verbatim, never expanded — the adapter's tensor names are ground truth and each still passes the stricter per-tensor gate). `all-linear` keeps its named reject; it is a keyword, not a regex, and its test caught me swallowing both.

### Validated on GPU (greedy, temp 0, identical prompts)

```
base      "The ocean, a vast and dynamic body of saltwater, covers more than 70%…"
novelist  "The ocean is a vast, ever-changing body of saltwater that covers more than 70%…"
cyber     "The vast, deep blue ocean covers more than seventy percent…"
```

Coherent, distinct, on-brand per adapter; `17+25` returns 42 on all three, as a style adapter should. **All three now load with no opt-in.**

A suspected hazard closed by measurement rather than argument: all three ship **F32** tensors while the loader strides by `BF16_BYTES` with no dtype check. cyber's 176.8 MB F32 file loads as 88,342,528 bytes — exactly half — so the conversion is correct.

`ATLAS_LORA_ALLOW_PARTIAL=1` (default **off**) remains for adapters naming still-unsupported modules: it loads them and WARNs exactly which were skipped, rather than the silent-partial behaviour that is worse than a refusal.

740 unit tests pass.

rsafier:
**LoRA perf: prefill 166 → 620 tok/s, C≥2 decode 4.8 → 11.9, DFlash+LoRA C=4 32 → 54 (c2eec389, a4196092, 37ab9dae)**

Once the adapters loaded, LoRA was correct but slow: prefill 5.6× down, decode collapsing at C≥2. Three fixes, and **none of the causes were where they looked**.

### Measure, don't guess — two levers now in-tree

* `ATLAS_LORA_NO_APPLY=1` — adapter **resident**, all deltas skipped. Separates *cost of applying* from *cost of having loaded*.
* `ATLAS_LORA_NO_FFN=1` — skip only FFN/SSM deltas, keep attention. Separates *which* deltas.

Three hypotheses died to measurement before these existed: a missing `gemm_tc` kernel (it resolves and is used), an eager/no-CUDA-graph fallback (graphs capture in both legs), and the `[m, n_out]` delta scratch round-trip — that last one was ruled out by **building the fused kernel and measuring zero change**.

### 1. Prefill (c2eec389): freed weights, not math

Disabling the FP4-MMQ arm under LoRA is correct (it would scale a true-valued delta). But `finalize_nvfp4_mmq_load` runs at **load** time and **frees the transposed `_t` copies** the fallback GEMM needs, on the assumption the arm it just built will serve prefill. Disabling the arm afterwards stranded prefill on the slowest non-transposed path.

**The tell**: `ATLAS_NO_FFN_NVFP4_MMQ=1` (841 tok/s) and adapter-resident (176) diverged 4.8× while nominally doing the same thing — the env flag also skips the finalize, so its twins survive. Fix: skip the finalize when `config.adapter_max_rank > 0`, the load-time signal that `--lora-adapter` was given and the last moment the decision can be made before the twins go.

**164 → 617 tok/s.** With deltas skipped it measured 176, proving the cost was residency, not arithmetic.

### 2. Decode at C≥2 (a4196092): one dispatch threshold

`ATLAS_LORA_NO_FFN` attributed it in one run — per-stream at C=2: base 13.4, all deltas skipped 13.1, **FFN skipped/attention live 13.1**, all applied 4.7. Attention costs nothing; it is entirely the dense-FFN deltas, and the cost appears exactly at the m=1 → m=2 boundary (~8% of a step at C=1, ~64% at C=2).

`apply_lora_delta` sent m==1 to a GEMV and everything above to `dense_gemm_tc`, which tiles 16 rows while its B operand `[n_out, max_rank]` is **independent of m**. At m=2 it moved the entire B matrix (17408×64 per FFN module × 64 layers) to produce 2 useful rows. Small m now runs `m` single-row deltas — which also makes it agree with the documented oracle, since `apply_lora_bgmv` describes itself as byte-identical to `n` single-row calls.

### 3. DFlash2 + LoRA (37ab9dae): they coexist, and accepts are unaffected

`can_batch_verify_dispatch` carried a bare, uncommented `&& self.lora.is_none()` — next to two neighbours that both explain themselves. Every batch logged `DFlash batched verify DECLINED … running the per-sequence loop`, so DFlash throughput was **flat at ~34 tok/s from C=1 to C=4**. It is not a real barrier: every op that path batches carries its delta on the batched variants, and all rows share one adapter because mixed batches are refused upstream. Restoring batching then pushed verify to `n*k` rows (4×9=36), past `ATLAS_LORA_GEMV_MAX_M`, so the default went 8 → 48. Neither alone was enough.

### Measured

Per-stream decode tok/s, prefill on a 2079-token prompt, speculative **off**:

```
                 prefill    C=1     C=2     C=4
no adapter          932     14.4    13.4    12.2
LoRA (before)       166     13.2     4.8     4.6
LoRA (after)        620     13.2    11.9    10.2
```

Aggregate tok/s with DFlash2 **on**:

```
                     C=1     C=2     C=4      accept
DFlash, no adapter   40.1    68.3    79.7     84/84/70%
DFlash + LoRA (before) 33.8  34.2    32.1     86/85/70%
DFlash + LoRA (after)  36.9  48.6    54.3     86/85/70%
```

**Accept rate is identical with and without the adapter at every concurrency.** The drafter carries no LoRA and proposes from un-adapted weights, yet its drafts survive verification against the adapted target at the same rate — LoRA does not poison speculation.

Greedy output is byte-identical across all three changes. Kill-switches: `ATLAS_LORA_PREFILL_BGMV`, `ATLAS_LORA_GEMV_MAX_M`, `ATLAS_LORA_NO_BATCH_VERIFY`. 740 unit tests pass.

### Known-open, not addressed here

* Prefill's 620 against the 841 twins-kept ceiling is the FFN delta cost at large m (`NO_FFN` measures 829) — a GEMM-efficiency question on skinny-K, not a dispatch bug.
* **Mixed-adapter batches return empty completions** at C≥2: v0 single-active correctly refuses a batch holding a non-active slot, but it reaches the client as an empty 200 rather than a 4xx/503, so a caller cannot tell "unsupported" from "the model said nothing". The guard says "MoE LoRA" but fires on a dense model.
* The pool is 5392 MB at r=64 / `max_loras=8`, and both delta stages run at `max_rank`, so a small adapter under a large `--max-lora-rank` pays that bandwidth.

rsafier:
## State update — rebased on `main`, two kernels cherry-picked from #648, LoRA re-validated

### Rebased and mergeable

Rebased onto `main` @ `0f0b9913` (was 64 ahead / 18 behind of `85190fbf`). The rebase was clean — one commit (`3961cd8c`, MTP telemetry buckets) dropped because it had already landed upstream, leaving 62. GitHub now reports `MERGEABLE`.

A clean rebase is not evidence of a working tree, so it was checked rather than assumed: full `ATLAS_TARGET_MODEL='*'` build (clean), 641 spark-model + 104 spark-server tests, and the whole GPU decode/LoRA matrix below re-measured **on the rebased tree**. Numbers match pre-rebase within noise, which matters because upstream's GDN state-spine work (`0f0b9913`, `af75afb2`) touches `qwen3_ssm` — the same three files this branch touches for SSM `out_proj` LoRA. That was the only overlap in the whole 127-file/125-file diff.

### Two kernels taken from #648 (@rrstesiak), not the branch

Both are cherry-picks with `Co-authored-by:` and his `provenance-id` markers preserved in the source. #648 itself is 171 files and mostly a separate drafter lane; these two kernels stand alone.

**1. `w4a16_gemv_batch8_rt2`** (`f3df056e`) — register-tiled T=2 GEMV, **+9.6% C=1 decode, bit-exact** (verified independently, same sha256 both legs, not inherited from the claim). Integration differs deliberately: #648 patches four call sites, this branch routes width 8 through `W4a16BatchmTiers::resolve` in one place.

**2. `fp8_gemv_rowscale_batch8_rt2`** (`8d686033`) — the FP8 twin, for the drafter's **propose** path. Replaces prefill-class tile GEMMs that pad 87% (`fp8_gemm_t_row_scaled` M64) and 50% (`_m16`) of their M-tile at M=8.

| MinHeap prompt, 1500 tok, temp 0, DFlash2 | tile | twin | |
|---|---|---|---|
| **C=1 median** | 48.2 | **56.5** | **+17.3%** |
| C=2 | 88.2 | 93.7 | +6.2% |
| C=4 | 126.4 | 129.8 | +2.8% |
| C=8 | 163.0 | 165.7 | +1.7% |

C=1 reproduced across two independent boots. The C≥2 rows are *not* mainly this kernel: propose batches rows across sequences, so rows = `gamma * n_seq` and above 8 the tiles take over — what's left is steps that happened to run `n_seq=1`. Extending to M≤16 would cover C=2 outright; that's a follow-up.

**Drafter numerics are not a parity contract here**, unlike the w4a16 verify family: accept is target-argmax equality and a mismatch emits the *target's* own token, so a different draft moves the accept RATE only. Both halves measured — accept identical on both legs (85/85/80/74% at C=1/2/4/8), converged output text byte-identical.

### Decode, base vs LoRA (post-rebase tree)

Aggregate tok/s, same prompt and protocol throughout. "Before" is this branch prior to the two kernel ports.

| C | base before | **base now** | novelist before | **novelist now** |
|---|---|---|---|---|
| 1 | 44.0 | **56.4** | 36.9 | **45.6** |
| 2 | 87.7 | **93.6** | 57.3 | **57.6** |
| 4 | 128.5 | **131.3** | 78.3 | **78.7** |
| 8 | 161.0 | **164.6** | 94.4 | **95.9** |

Accept holds at 85/85/80/74% (base) and 85/84/78/74% (novelist). The adapter leg gains the full C=1 win because the drafter carries no LoRA, so the propose-path speedup is adapter-independent.

### LoRA still works

Re-verified on the rebased binary with `Dxniz/Novelist1.0-27b-Adapter`: loads and installs on **64 layers** (r=64), `lora_bgmv` / `moe_lora_gather_bgmv` / `moe_lora_grouped_down` all dispatch, generations are complete and correct (`class MinHeap`, all required members, `finish_reason=stop`), and **mixed base+adapter concurrency returns real content on all 4 streams** — the admission cohort filter serialises them rather than dropping any.

**One gap found while measuring, pre-existing and not from this branch's kernel work:** with an adapter resident, a request naming the *base* model returns byte-identical output to the adapter request (`8838f242862b39bb`) and differs from a true adapter-free base completion (`2f4598630ee14ef0`). That is `resolve_request_adapter_slot` behaving as documented — a `model` that isn't a resident adapter name falls through to `-1` = "defer to the installed active adapter" — so there is currently **no selector for *no* adapter** once one is loaded. Worth its own issue; a reserved "base" slot would fix it.

### On #648's headline

For the record, since the 54.5 tok/s number came up: booted with **Ronald's own serve config** (γ=8, `ATLAS_DFLASH_OPTION_B` / `_EAGLE_FIX` / `_DRAFTER_FP8`, 4096 ctx, batch 1, `--ssm-cache-slots 0`, bf16 KV), his branch measures **54.6 tok/s median** — his claim reproduces. An earlier reading of 8.4 tok/s at 0% accept was **my misconfiguration**, not his branch: this repo's DFlash flags left γ at 15 with OPTION_B/EAGLE_FIX off. The configs aren't matched against this branch either (his answer runs 435 tokens where this one runs 789, and a shorter generation amortises prefill over fewer tokens), which is exactly why both kernels here were measured against the code they replace in one binary with one env flip, rather than against that branch.

### Still outstanding

- Five GPU gates + committed `.benchmarks` records — no records since today's `crates/` changes; blocking for merge.
- Splitting the LoRA work into its own PR (62 commits now span five themes).
- C=16/32 concurrency sweep at 8K context.
- Extending the FP8 twin to M≤16 for C=2.

rsafier:
Cut into the PR train against @rrstesiak's base #649:

- **#650** — memory buffers (allocation ledger, ViT lazy scratch, DFlash buffer sizing). Merge after #649.
- **#651** — LoRA on hybrid Qwen3.8-27B. Stacked on #650 (its base is that branch, so its diff is LoRA-only). Merge after #650.
- **#652** — TUI prefill visibility + DSpark/DFlash metrics fixes. Sibling of #650, independent, merge after #649.

Each was lifted as clean commits onto #649's base rather than replayed, built with `ATLAS_TARGET_MODEL='*'`, and tested.

Not carried into the train, and staying here: the DFlash2 conv+selector lane, cross-sequence batched verify/propose, and this branch's parity gates — #649 is the base layer for DFlash2 now, and the two kernels I ported from #648 (rt2 `f3df056e`, FP8 propose twin `8d686033`) are already in its top commit, so they are deliberately absent from all three.

rrstesiak:
@rsafier Since you have the GDN arm in hand, here is everything I learned hunting a temp-0 spec divergence on the 3.6-MoE line. Five sessions, 18 dead theories, and the hunt never fully terminated, so read this as a candidate map of where to look and what already died there, not a diagnosis of #604. All of it is Qwen3.6-35B-A3B / dense-27B era evidence on a different verify path. It smells most relevant to #459's "second divergence source outside the GDN chain."

## Measured facts from the 3.6 hunt

1. **The carrier was target-side.** Under strict-argmax accept the drafter cannot change token identity, so any spec-on vs spec-off stream divergence is target-side arithmetic by construction. That makes one strict-argmax run the cheapest partition of the whole search space: it splits drafter from target before you read a single kernel.

2. **Deterministic and one-time.** Streams converged after exactly one request, and the first-request delta was prompt-dependent in sign (+8% on one prompt, -3% on another, reproduced on virgin boots to three decimals). If you ever observe "it settles after request 1," you are probably looking at my animal.

3. **Ruled out by direct measurement** (no need to re-spend time here): target KV, drafter KV, BufferArena, persistent GPU allocations, SSM/GDN live state, calibration state, physical memory layout.

4. **Where the fork was born.** Per-layer probes put the first divergence at (step 51, layer 0): pre-attention hidden bit-identical, residual stream differing, with prefill proven bit-identical in both carriers. So the difference is born in the window between end-of-prefill and the first verify step: the seam plus the first propose. I never instrumented that window; it is the one defined measurement left on the table. A second arrow points at the same spot: the 07-08 temp-0 flip on this repo was also a seam bug (fixed by the seam-commit ordering flag).

5. **Flag-relocation bisection.** My corruption was spec-only and relocated when flags changed. A divergence that moves with flags has its carrier in the intersection of those configs; no single kernel owns it. Related landmine: the spec path with zero accepted drafts does not execute the same code as serial decode, which is a divergence source that lives in no kernel at all.

## Method notes

- Byte-compare the streams. Token counts can match while bytes differ; that cost me a day once.
- Per-layer probes cost ~36 tok/s, so pair them with graphs disabled and keep them out of any perf config.
- Your #653 reduction-order find is the exact class my batchm gate 4 exists to catch (per-output FMA order preserved against the M=1 kernel, bit-exact across shapes and M). That harness is in #649's `batchm_bench` and generalizes to any batched arm.

If any of this collides with what you are seeing on the GDN arm, say so and I will pull the deeper evidence on that specific line.

rsafier:
This PR's title says "localized to the GDN verify path". That localization holds, and it now has a name, a one-flag confirmation and a throughput number. Posting here because it belongs on this thread rather than in a new issue.

## The carrier is the non-exact GDN/SSM verify chain, and it is default-on

`--exact-verify` (`GdnFlags::exact_verify`) is OPT-IN, default OFF, because exactness costs ~+22-36% per decode step. Its own documentation already says the consequence out loud: *"with default settings spec-on output is NOT bitwise-equal to spec-off (the #435 divergence ships)"*.

That is the whole thing. Measured on `unsloth/Qwen3.8-27B-NVFP4` + `incoai/Qwen3.8-27B-DFlash2`, GB10, one binary, this flag the only variable:

| probe | default | `--exact-verify` |
|---|---|---|
| "count from 1 to 10" | `1, 2, 100, 100, 100, ...` | `1, 2, 3, 4, 5, 6, 7, 8, 9, 10` |
| "two sentences about blue" | `...100% success rate.\n107 \| ### 3.3.2.2.1` | coherent |
| `video-fidelity` C=1 / C=2 / C=4 | 1/1, **0/2**, **0/4** | **1/1, 2/2, 4/4** |

## Why it presents as a broken drafter when the drafter is fine

A flipped argmax does not stay one wrong token. It lands the sequence in a repetition attractor, and **the drafter then agrees with it** — verify steps go 7/7 accepted — so the loop locks in and the reply degenerates until the token cap. Acceptance telemetry looks healthy at 29-71% the whole time. The accepts are honest; the logits they check against are not the ones sequential decode would have produced.

Localized by first divergence rather than by inspection. On "count from 1 to 10" (31 prompt tokens) the first wrong token is completion index 6 (`'1'` id 16 where serial decode gives `'3'` id 18), and the verify log maps it exactly (`seq_len` is the length after each step):

```
step 1  accepted=1/7  ends 33  -> completion idx 0-1
step 2  accepted=0/7  ends 34  -> idx 2
step 3  accepted=0/7  ends 35  -> idx 3
step 4  accepted=7/7  ends 43  -> idx 4-11
```

Index 6 sits inside step 4's accepted-draft run, so the wrong token was ACCEPTED — only possible if the target's own verify logits chose it.

## Exactness measured FASTER, not slower

The ~+22-36% is a per-decode-STEP figure. End-to-end throughput also depends on how many drafts survive verification, and that is exactly what the divergence damages: corrupted verify logits **reject correct drafts**. At C=1, n=3 after a warm request:

| | default | `--exact-verify` |
|---|---|---|
| decode | 27.29 tok/s | **32.45 tok/s** (+18.9%) |
| mean acceptance | 35.0% (103 steps) | **45.5%** (236 steps) |

So on a config where the divergence bites, exactness is not a tradeoff — it buys correctness *and* throughput. Scope: C=1, one prompt, this box; the documented step costs are measured at the n=8/16/32 concurrency rungs, so the crossover needs its own sweep before anyone changes a default.

## Ruled out along the way, each by one-variable measurement

Batched propose (`ATLAS_DFLASH_BATCH_PROPOSE=1`), LM-head dtype (`fp8` and `default` alike), prefix caching (byte-identical failures on/off), multimodality (text-only degenerates on its own), `ATLAS_DFLASH_UNIFIED_CTX=0`, and the GEMV tier — `w4a16_batch_bitparity_microtest` PASSES on this box at every M=2..16 and every projection shape with its 1-ULP control firing.

It is also not specific to any branch in flight: identical degeneration on `pull/648/head` (`acaf9533`) and `pull/649/head`, each under its own author's config, with spec-off clean on both.

## Two traps worth writing down

**`ATLAS_MTP_MAX_SEQS=1` is not "batched verify off".** `ladder.rs` documents it as "restore single-sequence-only", so at C>=2 it disables spec entirely. Its apparent C>=2 improvement is spec-off leaking in.

**A forced `--dflash-gamma` that does not match the drafter's block size makes DFlash decline silently.** `--dflash-gamma 2` and `4` ran ZERO `DFLASH K=` verify steps and looked perfectly clean — spec-off in disguise. Always count verify steps in the log before reading a gamma arm; this cost me a wrong conclusion before the log count caught it.

The operator-facing half of this is written up in #667 (documentation only, no default changed).

rsafier:
Root cause found and fixed: **#699**. The temp-0 spec degeneration this PR chased was never the GDN verify numerics — `k4_apply_verdict` sizes its rewind by `drafts.len()` instead of by the forward's row count, so a K=4 verify dispatched onto a sequence carrying DFlash γ=7 pending drafts EMITS its accepted tokens and then erases them from the sequence's history (`seq_len -= 7-3` against a 4-row forward). `--exact-verify` appeared to fix it only because it changes the gate's dispatch so the collision stops firing. With #699, on the DEFAULT verify chain: "count 1..10" goes from `1, 2, 100, 100...` to correct, video-fidelity C=2/C=4 go 0/2, 0/4 → 2/2, 4/4, and temp-0 output is deterministic again. Full forensics in #699's description.

PR #609, opened 2026-08-19
fix(ci): keep the harvest commit subject to one line

tbraun96:
Look at `git log --oneline` on main right now:

```
24b08f26 chore(governance): harvest PR intent ledgers (runs 31917879724 31918802406 31924904384 … 110 more …)
```

The harvest interpolates every run id it swept into the **commit subject**. That is the line `git log --oneline`, `git shortlog`, and every PR list renders, so main's history is now unreadable at a glance.

#578 fixed exactly this for the PR *title* — a 47-run sweep blew the 256-character API limit — and stopped there. The commit subject kept growing because git imposes no limit.

Subject now states the run count; the ids move to the commit body, where length costs nothing and the provenance is still fully recorded.

### DIFF
--- a/.github/workflows/governance-harvest.yml
+++ b/.github/workflows/governance-harvest.yml
@@ -326,12 +326,17 @@ jobs:
           # API caps it at 256 characters, and a full sweep of 47 runs blew past
           # that and failed PR creation outright. Title states the count, body
           # carries the ids.
-          msg="chore(governance): harvest PR intent ledgers (runs ${run_ids:-unknown})"
+          # A git SUBJECT is a one-line summary and tooling treats it that way.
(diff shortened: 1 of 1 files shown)


tbraun96:
Superseded by #622, which fixes the same line and also explains why #578 didn't: GitHub's squash-merge takes the **commit subject**, not the PR title, when a PR has exactly one commit — so the title #578 shortened is never what lands. #622 derives both from one string, adds the ⚖️ prefix, and moves the run-id list to the job summary.

Closing rather than pushing: this branch had been stuck seven hours behind a six-hour apt hang on the CUDA release leg, and stacking on it would have blocked the fix behind flaky infrastructure.

Issue #434, opened 2026-08-09
META: PR Telemetry Tracker


tbraun96:
For use by Automated Tracking. Status is periodically updated below. Will need improvement.

PR #621, opened 2026-08-19
feat(ci): move the PR telemetry view onto a Discussion

tbraun96:
The telemetry view is a standing document — it rewrites itself every four hours and on every merge to `main`, and it is never "done". An issue that is never closed reads as a task nobody finished, which is not what this is.

It now upserts into **[[META] 📊 PR Telemetry](https://github.com/Avarok-Cybersecurity/atlas/discussions/619)**. Issue #434 keeps a pointer so existing links still lead somewhere.

### The one genuinely new mechanic

Discussion comments are **GraphQL-only** — there is no REST `issues/comments` equivalent, so `gh api -X PATCH .../issues/comments/$id` cannot be reused. Rather than hand-write that mutation in every caller, the upsert moves into `.github/actions/upsert-discussion-comment`. The intent-eigenvector report lands next and is the second caller, which is one more than a copied implementation survives.

The round-trip (create → find-by-marker → update → delete) was verified live against discussion #620 before this was written, including a body with backticks, `$dollars`, quotes and newlines.

### Stricter than the code it replaces

Two changes, both closing paths that end in silent duplicate comments:

- **The lookup paginates.** Humans reply in these threads. Once replies push our comment past the first page, an unpaginated lookup stops finding it and posts a new one *every run* — the exact duplication the marker exists to prevent.
- **A body without the marker is refused, not posted.** Such a comment could never be found again, so it would leak a new one on every subsequent run.

### Unchanged

The render/artifact/publish split, the renderer, and the `gate::telemetry` marker constants. Only the transport moved. The publish job still holds exactly one write scope — now `discussions: write` instead of `issues: write`.

### Verification

`vars.PR_TELEMETRY_DISCUSSION` is already set to `619`. After merge: dispatch the workflow twice and confirm the second run *updates in place* rather than adding a second comment.

### DIFF
--- a/.github/actions/upsert-discussion-comment/action.yml
+++ b/.github/actions/upsert-discussion-comment/action.yml
@@ -1,0 +1,105 @@
+name: Upsert a Discussion comment
+description: >-
+  Keep exactly ONE bot-authored comment on a Discussion, rewriting it in place
+  instead of appending a new one on every run. The comment is found by a marker
+  embedded in the body itself, so there is no comment id to store anywhere and
+  a manually deleted comment simply results in a fresh one.
+
+# ── Why a Discussion and not an issue ───────────────────────────────────────
+#
+# These reports are standing documents, not work items. An issue that is never
+# closed reads as a task nobody finished; a Discussion reads as a page, which is
+# what a continuously-rewritten view actually is.
+#
+# ── Why this is an action and not copied into each workflow ─────────────────
+#
+# Discussion comments are GraphQL-ONLY. There is no REST `issues/comments`
+# equivalent, so the familiar `gh api -X PATCH .../issues/comments/$id` cannot
+# be reused and every caller would otherwise carry its own hand-written
+# mutation. Two callers already exist (pr-telemetry, intent-eigenvector), which
+# is one more than a copied implementation survives.
+
+inputs:
+  discussion-number:
+    description: >-
+      The Discussion to post into. MAY BE EMPTY: unset means "not configured
+      yet", and the action says so and succeeds rather than guessing a number
+      and commenting on something arbitrary. Callers render to the step summary
+      regardless, so an unconfigured repo still gets the report.
+    required: false
+    default: ""
+  marker:
+    description: >-
+      The sentinel that identifies OUR comment. Must appear verbatim in
+      `body-file` — the action does not insert it, because the renderer that
+      owns the body is also what should own the marker constant.
+    required: true
+  body-file:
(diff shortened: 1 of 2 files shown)


tbraun96:
**Included in stack `ci-plumbing` → #868.**

**Why this earns a merge:** the telemetry view is a standing document that rewrites itself every four hours and on every merge to `main`. An Issue that is never closed reads as a task nobody is doing; a Discussion is the right shape for a document that is never "done".

**Verified still needed, not merely still open:** main's `pr-telemetry.yml` still declares `issues: write` and keys off a tracking-issue variable, so the Discussion move has not landed by another route. (I checked, because a sibling PR in the same batch — #503 — turned out to be entirely already-landed.)

**Stacked with:** #741 (dependabot action bumps) plus the `stack-plan.py` tool that produced the grouping. The three share no source file, and together they touch **zero `PERF_PATHS`** files — so the group owes **no benchmark campaign**. That is the whole reason to stack: certification is the scarce resource, and this group does not consume any.

**One correction I owe you.** I initially flagged this branch as an orphan with no merge base — which would have meant merging it reverts main. That was wrong: my working clone had only 6 commits of `main`, so `git merge-base` had nothing to find. After deepening, the merge base is `c1ae36af` (ahead 1, behind 101) and the real diff is **2 files**. `check_merge_ancestry.sh` now reports ✓.

Nothing is needed from you — the commit is cherry-picked into #868 with authorship preserved. This PR can close once that lands, or stay open if you would rather land it directly.

tbraun96:
Landed in stack #1113, whose base is #873: this PR's commit (`move the PR telemetry view onto a Discussion`, with the `upsert-discussion-comment` composite action) was carried by #868 and is now in the base. Nothing was rewritten — it was cherry-picked as-is onto the post-rebrand `main`.

Leaving it open until the base merges; it will show contained then, and I will close it with the merge sha if GitHub does not.

tbraun96:
Landed in `main` as part of #873 (stack #1113): the merge commit is `8838f64da`, and both halves of this PR are in it —

```
$ git show avarok/main:.github/actions/upsert-discussion-comment/action.yml | head -3
name: Upsert a Discussion comment
description: >-
  Keep exactly ONE bot-authored comment on a Discussion, rewriting it in place
```

plus `pr-telemetry.yml`'s move onto that action (58 lines changed there). The commit was cherry-picked as-is onto the post-rebrand `main`; nothing was rewritten. GitHub squash-merged the base, so this PR's own sha is not an ancestor of `main` and it could not auto-close — closing by hand.

PR #622, opened 2026-08-19
⚖️ fix(ci): bound the harvest commit subject, and mark it with the scales

tbraun96:
`git log --oneline` on `main` is unreadable:

```
c1ae36af chore(governance): harvest PR intent ledgers (runs 31917879724 31918802406 … 130 more … 32246814273) (#617)
```

That is **1,642 characters**. The sweep before it was 1,558; before that 1,438. The list only grows, because the standing sweep re-reports every run it has ever seen.

### Why #578 didn't fix this

It fixed the wrong string. The PR *title* has been a tidy 57 characters since #578 — but **GitHub's squash-merge takes the commit subject, not the PR title, when a PR has exactly one commit**, and every harvest PR has exactly one. The title that was fixed is never what lands.

Both are now derived from a single `$subject`, so a future fix cannot again land on the half that does not ship:

```
⚖️ chore(governance): harvest PR intent ledgers (132 run(s))
```

60 characters, and carrying the scales the governance lane is named for.

### The ids are kept, just not in the subject

They are real provenance, so nothing is discarded — only moved somewhere with room:

| where | before | after |
|---|---|---|
| commit subject | all 132 ids (1,642 ch) | count only (60 ch) |
| commit body | *(none)* | newest 10 + count + run link |
| PR body | all 132 ids | newest 10 + count + run link |
| job summary | *(none)* | **complete list** |

The job summary is durable, unbounded, and costs the tree nothing.

### Verification

Ran the block offline against the real shapes — 132 runs, 3 runs, and an empty file:

- 132 → subject 60 ch, note 275 ch, correctly reading "Newest 10 of 132 … (…and 122 more)"
- 3 → all three listed, no truncation clause
- 0 → `Harvested runs: unknown`, no arithmetic on an empty count

No commit-lint job exists in this repo (checked), so the leading emoji is safe.

### Supersedes #609

Same line, smaller fix. #609 had been stuck seven hours behind a six-hour apt hang on the `linux-x86_64-nvidia-cuda` release leg; this replaces it rather than stacking on a blocked branch.


tbraun96:
**Not stacked yet — this one is _half_ already landed, and taking it whole would revert.**

**What already reached main:** the harvest commit subject is bounded. `governance-harvest.yml:330` now builds

```
title="chore(governance): harvest PR intent ledgers (${run_count:-0} run(s))"
```

and the live bot PR (#867) is titled `harvest PR intent ledgers (49 run(s))` — short, no run-ID list. The `git log --oneline` problem this PR opens with is fixed.

**What has _not_ landed, and is the reason to keep this open:** `.github/scripts/governance-evict.sh` and `governance-evict-test.sh` are absent from main. That is a real contribution with no substitute.

**What I would like to do,** if you are happy with it: rebase this onto current `main` keeping only the eviction hunks and dropping the subject-bounding half, then fold it into a `governance` stack alongside the harvest lane. Merging as-is would re-apply the landed half against three weeks of drift in `ci.yml` and the harvest workflow.

Say the word and I will do the rebase, or push it yourself and I will pick it up in the next stack.

PR #624, opened 2026-08-19
feat(governance): find the repository's intent eigenvector

tbraun96:
> **Stacked on #621** for the shared `upsert-discussion-comment` action. Retargets to `main` when that merges.

`governance/` holds one classified intent per PR, and **nothing has ever read them as a whole**. Individually each answers "what was this change for". The collection should answer something bigger — *where is this repository going?*

### Why not just ask a model

"Summarise these hundred intents" is not a question with a checkable answer. A model will produce something plausible whether or not structure exists, and nothing in the output distinguishes the two cases.

So the structure is found **first, arithmetically**. Each intent is embedded, mean-centred, and decomposed into principal components by power iteration with deflation. The model's only job is to put a *name* to an axis it is shown. If it abstains, the report still renders with each axis described by the PRs at its poles — the evidence the name was derived from anyway.

### What it reports

- the top three **axes**, each with the PRs at both poles
- **Coherence** — variance explained by the first component: how single-minded the repo is
- **Drift** — the angle between the principal directions of the older and newer halves of the window. This is the number that actually answers "is the trajectory turning?"

### Two engineering choices worth review

**Gram matrix, not covariance.** The window holds ≤100 PRs against a 2048-wide embedding, so `XXᵀ` (n×n) has the same non-zero spectrum as `XᵀX` (D×D) for a ten-thousandth of the work — and its eigenvectors are *per-sample loadings*, which is the form needed to rank PRs along an axis. The feature-space direction is only materialised where genuinely needed: the drift angle.

**Determinism is a requirement, not a nicety.** The output is upserted into a Discussion comment on a schedule; a decomposition that answered differently for identical input would rewrite that comment forever and show readers churn that means nothing. Hence a fixed iteration count with no convergence-based early exit, a fixed start vector, and a fixed sign convention — eigenvectors are defined only up to sign, and power iteration will otherwise return `u` one day and `-u` the next.

### The harvester's own PRs are excluded

They are bookkeeping, not intent: ~9 near-identical `chore(governance): harvest…` entries, all classified `infrastructure/ci` — a dense cluster in exactly one place. Left in, **they dominated the first component outright**, and "the principal direction of the repository" was really a measure of how often the bot had run. Measured at 13% of the corpus.

### Verified end to end against the live ledger

62 real intents, embedded through the free-tier endpoint, decomposed and named:

| | |
|---|---|
| Coherence | **9.5%** — diffuse, no single direction dominates |
| Drift | **59°** — turning markedly |
| Axis 1 | CI automation vs kernel optimisation — 9.5% |
| Axis 2 | developer tooling vs harvest maintenance — 4.8% |
| Axis 3 | benchmark validation vs speculative kernels — 4.3% |

That is recognisably this repository's last two months, and the model found the CodeRAG initiative and the benchmark-ladder work without being told either existed. Re-running `analyze` on the same input produced **byte-identical** output.

Also exercised locally: the empty-key abstain path (warns, publishes nothing, exits 0), the no-naming render path, and an unknown CLI mode (rejected rather than defaulted).

### Tests

12 unit tests on the mathematics — recovery of a planted axis, ordering and bounding of explained variance, deflation orthogonality, byte-identical repeats, sign stability, drift at 0° and 90°, and the degenerate corpora that will occur: one PR, an empty ledger, a zero-variance corpus (must not emit NaN), and k > n.

### DIFF
--- a/.github/ai-models.json
+++ b/.github/ai-models.json
@@ -15,12 +15,19 @@
     "~100x smaller than the categorizer. Chunk rerank inputs; do not assume one",
     "uniform budget.",
     "",
+    "`synthesize` NAMES axes that a spectral decomposition already found; it",
+    "never chooses them. It is listed separately from `categorize` despite",
+    "currently pointing at the same id, because the two have different failure",
+    "consequences -- a bad category misroutes a benchmark, a bad axis label is",
+    "cosmetic -- and should be repointable independently.",
+    "",
     "Reference implementation: Avarok-Cybersecurity/lattice-db,",
     "examples/rag-example/src/openrouter.ts."
   ],
   "endpoint_base": "https://openrouter.ai/api/v1",
   "models": {
     "categorize": "nvidia/nemotron-3-ultra-550b-a55b:free",
+    "synthesize": "nvidia/nemotron-3-ultra-550b-a55b:free",
     "embed": "nvidia/llama-nemotron-embed-vl-1b-v2:free",
     "rerank": "nvidia/llama-nemotron-rerank-vl-1b-v2:free"
   }
--- a/.github/workflows/intent-eigenvector.yml
+++ b/.github/workflows/intent-eigenvector.yml
@@ -1,0 +1,276 @@
+name: Intent eigenvector
+
+# Where is this repository going?
+#
+# `governance/` holds one classified intent per PR and nothing has ever read
+# them as a whole. Individually each answers "what was this change for"; the
+# collection should answer something bigger, and does — but only if it is asked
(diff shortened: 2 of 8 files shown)


tbraun96:
Reopened as a standalone PR against `main` — see the link above.

This was stacked on #621, but its files live under `crates/`, a PERF_PATH, so it invalidated all 10 benchmark gate records and dragged the CI-only telemetry change into a multi-hour GPU gate with it. Splitting them lets #621 and #622 merge on their own checks.

PR #626, opened 2026-08-19
feat(governance): find the repository's intent eigenvector

tbraun96:
Replaces #624, which was stacked on #621. It is standalone here because these files live under `crates/` — a PERF_PATH — so they invalidate **all 10 benchmark gate records** even though `atlas-governance` is host-only and CUDA-free. Keeping them apart lets the CI-only governance work (#621, #622) merge without being held behind a multi-hour GPU run.

> ⏳ **This PR needs a benchmark gate run.** Worth stacking with #589/#597 for a single measurement rather than gating it alone.

---

`governance/` holds one classified intent per PR, and **nothing has ever read them as a whole**. Individually each answers "what was this change for". The collection should answer something bigger — *where is this repository going?*

### Why not just ask a model

"Summarise these hundred intents" is not a question with a checkable answer. A model will produce something plausible whether or not structure exists, and nothing in the output distinguishes the two cases.

So the structure is found **first, arithmetically**. Each intent is embedded, mean-centred, and decomposed into principal components by power iteration with deflation. The model's only job is to put a *name* to an axis it is shown. If it abstains, the report still renders with each axis described by the PRs at its poles — the evidence the name was derived from anyway.

### What it reports

- the top three **axes**, each with the PRs at both poles
- **Coherence** — variance explained by the first component: how single-minded the repo is
- **Drift** — the angle between the principal directions of the older and newer halves of the window. This is the number that actually answers "is the trajectory turning?"

### Two engineering choices worth review

**Gram matrix, not covariance.** The window holds ≤100 PRs against a 2048-wide embedding, so `XXᵀ` (n×n) has the same non-zero spectrum as `XᵀX` (D×D) for a ten-thousandth of the work — and its eigenvectors are *per-sample loadings*, which is the form needed to rank PRs along an axis. The feature-space direction is only materialised where genuinely needed: the drift angle.

**Determinism is a requirement, not a nicety.** The output is upserted into a Discussion comment on a schedule; a decomposition that answered differently for identical input would rewrite that comment forever and show readers churn that means nothing. Hence a fixed iteration count with no convergence-based early exit, a fixed start vector, and a fixed sign convention — eigenvectors are defined only up to sign, and power iteration will otherwise return `u` one day and `-u` the next.

### The harvester's own PRs are excluded

They are bookkeeping, not intent: ~9 near-identical `chore(governance): harvest…` entries, all classified `infrastructure/ci` — a dense cluster in exactly one place. Left in, **they dominated the first component outright**, and "the principal direction of the repository" was really a measure of how often the bot had run. Measured at 13% of the corpus.

### Verified end to end against the live ledger

62 real intents, embedded through the free-tier endpoint, decomposed and named:

| | |
|---|---|
| Coherence | **9.5%** — diffuse, no single direction dominates |
| Drift | **59°** — turning markedly |
| Axis 1 | CI automation vs kernel optimisation — 9.5% |
| Axis 2 | developer tooling vs harvest maintenance — 4.8% |
| Axis 3 | benchmark validation vs speculative kernels — 4.3% |

That is recognisably this repository's last two months, and the model found the CodeRAG initiative and the benchmark-ladder work without being told either existed. Re-running `analyze` on the same input produced **byte-identical** output.

Also exercised locally: the empty-key abstain path (warns, publishes nothing, exits 0), the no-naming render path, and an unknown CLI mode (rejected rather than defaulted).

### Tests

12 unit tests on the mathematics — recovery of a planted axis, ordering and bounding of explained variance, deflation orthogonality, byte-identical repeats, sign stability, drift at 0° and 90°, and the degenerate corpora that will occur: one PR, an empty ledger, a zero-variance corpus (must not emit NaN), and k > n.

### Merge order

The publish job uses `.github/actions/upsert-discussion-comment`, which arrives in #621. #621 is gate-free and will land first; this workflow only runs on a schedule, so there is no window where it could fire without the action present.

### DIFF
--- a/.github/ai-models.json
+++ b/.github/ai-models.json
@@ -15,12 +15,19 @@
     "~100x smaller than the categorizer. Chunk rerank inputs; do not assume one",
     "uniform budget.",
     "",
+    "`synthesize` NAMES axes that a spectral decomposition already found; it",
+    "never chooses them. It is listed separately from `categorize` despite",
+    "currently pointing at the same id, because the two have different failure",
+    "consequences -- a bad category misroutes a benchmark, a bad axis label is",
+    "cosmetic -- and should be repointable independently.",
+    "",
     "Reference implementation: Avarok-Cybersecurity/lattice-db,",
     "examples/rag-example/src/openrouter.ts."
   ],
   "endpoint_base": "https://openrouter.ai/api/v1",
   "models": {
     "categorize": "nvidia/nemotron-3-ultra-550b-a55b:free",
+    "synthesize": "nvidia/nemotron-3-ultra-550b-a55b:free",
     "embed": "nvidia/llama-nemotron-embed-vl-1b-v2:free",
     "rerank": "nvidia/llama-nemotron-rerank-vl-1b-v2:free"
   }
--- a/.github/workflows/intent-eigenvector.yml
+++ b/.github/workflows/intent-eigenvector.yml
@@ -1,0 +1,276 @@
+name: Intent eigenvector
+
+# Where is this repository going?
+#
+# `governance/` holds one classified intent per PR and nothing has ever read
+# them as a whole. Individually each answers "what was this change for"; the
+# collection should answer something bigger, and does — but only if it is asked
+# a question with an answer. "Summarise these hundred intents" is not that
+# question: a model will produce something plausible whether or not structure
+# exists.
+#
+# So the structure is found FIRST, arithmetically. The intents are embedded,
+# stacked, and decomposed into principal components; the model's only job is to
+# put a name to an axis it is shown. If it abstains, the report still renders
+# with each axis described by the PRs at its poles.
+#
+# Advisory, permanently. Nothing here gates anything, and a PR sitting far off
+# the principal axis is an observation, not a criticism.
+
+on:
+  schedule:
+    # Daily. The ledger moves a few PRs a day; anything more frequent rewrites
+    # the comment with the same content.
+    - cron: "41 6 * * *"
+  workflow_dispatch:
+
+# The publish step reads a comment and rewrites it, so a cancelled run could
+# leave it half-updated. Same reasoning as pr-telemetry.yml.
+concurrency:
+  group: intent-eigenvector
+  cancel-in-progress: false
+
+permissions: {}
+
+jobs:
+  render:
+    name: Decompose the ledger
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Collect the intent documents
+        id: docs
+        env:
+          GH_TOKEN: ${{ github.token }}
+          # Echoed into the report so a reader knows this is a rolling window
+          # and not an all-time view. Matches LEDGER_CAP in governance-harvest.yml.
+          LEDGER_WINDOW: "100"
+        run: |
+          set -euo pipefail
+
+          # ★ The harvester's OWN PRs are excluded. They are bookkeeping, not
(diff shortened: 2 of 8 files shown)


tbraun96:
Triaged for a certification campaign. **Both conflicts are mechanical**, so recording exactly what they are — this is a short job whenever you want it landed, not a rebase.

Merged onto current `main` (`5f639bfc57`) locally with the rename limit raised. Two conflicts:

**1. `.github/ai-models.json` — one hunk, a union.** Both sides append prose to the same comment array: main added the note that `/review` reads the org variable `OPENROUTER_DEFAULT_FREE_MODEL` rather than this file; this branch added the note that `synthesize` names axes a spectral decomposition already found and should be repointable independently of `categorize`. Neither replaces the other — keep both.

**2. `crates/avarok-plugin/src/bin/intent_eigenvector.rs` — not a content clash.** `git status` reports `UA` (stage 3 only): main does not have this file at all, the branch adds it. Take the branch's copy.

One thing that comes with (2): the file is added at the pre-rename path `crates/atlas-plugin/…` and carries **2 references to `atlas_governance`**, which #1101 renamed. Those need to become `avarok_governance` or it will not compile.

Not resolving it here: it touches `crates/avarok-plugin/`, so it needs a certification campaign, and it does not group with any of the three composed and queued right now (#1203 server-honesty, #1202 K3, #1207 DeepSeek-V4.1) — riding an unrelated governance feature into one of those would make a bad campaign boundary. Say the word and it composes into the next one, or gets its own.

Context on the rename and the rest of the backlog: #1206.

tbraun96:
**Correcting my earlier comment: this is more than two mechanical conflicts.**

I said the only work was a union in `.github/ai-models.json` and taking the
branch's copy of `intent_eigenvector.rs`. That was measured from git's conflict
list, which understates it — most of this PR's files are **pure additions at
paths main no longer uses**, so they conflict with nothing and are silently
wrong rather than loudly wrong.

Mapping every file onto main (`69f9ae3062`):

| file (as this PR has it) | on main |
|---|---|
| `crates/atlas-governance/src/eigenvector/{mod,render,spectral,spectral_tests}.rs` | **absent** — and the crate is `crates/avarok-governance` now |
| `crates/atlas-plugin/src/bin/intent_eigenvector.rs` | **absent** — crate is `crates/avarok-plugin` |
| `.github/workflows/intent-eigenvector.yml` | absent |
| `crates/atlas-governance/src/lib.rs` | differs — main's has the ledger half; this adds the eigenvector module |
| `.github/ai-models.json` | differs — the union I described |

So: **the work is genuinely unlanded** (unlike #484, which turned out to be
already in). `crates/avarok-governance` exists on main but holds only
`event.rs`, `ledger.rs`, `ledger_tests.rs`, `lib.rs` — the ledger. The
eigenvector module, its workflow and its binary are all still only here.

What landing it actually needs, beyond the two conflicts I named:

1. re-path `crates/atlas-governance/src/eigenvector/` →
   `crates/avarok-governance/src/eigenvector/`, and
   `crates/atlas-plugin/src/bin/` → `crates/avarok-plugin/src/bin/`;
2. merge `lib.rs` rather than take either side — main's has the ledger, this
   adds `eigenvector`;
3. rename the `atlas_governance` references (2 in the binary, more in the
   module);
4. the `ai-models.json` union.

The rename target is not a judgement call — every crate is `avarok-*` since
#1101 — but this is a feature of yours that is **182 commits behind**, and
reviving it means re-pathing a whole module, not resolving a merge. I have not
done that on my own initiative: it touches `crates/avarok-plugin/`, so it needs
a certification campaign, and it does not group with the two queued (#1202 K3,
#1207 DeepSeek-V4.1), which would make a poor campaign boundary.

Say the word and I will re-cut it onto main and compose it into the next
campaign that has room. Context on the rename cohort generally: #1206.

PR #646, opened 2026-08-20
bench(ladder38): C=64 and C=128 win under the most conservative assumption

tbraun96:
Closes the last gap in the ladder without running the hazardous measurement.

Rather than run vLLM at cap 128 at C≥64 — the workload that has cost this fleet three physical powercycles — Atlas's own absolutes were measured there today (dgx1, idle, certified commit `1575873582`, **Atlas-only, zero wedge risk**) and compared against the **certified** vLLM column.

That comparison is deliberately unfair to Atlas: vLLM has demonstrably fallen with the fleet (its certified C=32 283.48 measures **277.12** today, −2.2%), so crediting it with its old number *understates* Atlas's true margin.

| C | Atlas today | certified vLLM | ratio | certified ratio |
|---:|---:|---:|---:|---:|
| 64 | **375.68** (0.31% spread) | 361.39 | **1.040×** | 1.070× |
| 128 | **460.37** (0.38% spread) | 358.57 | **1.284×** | 1.333× |

**Both win even if vLLM keeps its full pre-shift number.** Grant vLLM the same −2.2% it actually took at C=32 and the ratios become ~1.063× and ~1.313×, close to certified. Either way the ordering isn't in question — which is what the differential-based bound in #643 argued, now demonstrated with measured Atlas numbers at the rungs themselves.

## ★ Correcting the shape of the deficit

Earlier sections describe it as "growing with concurrency". With C=64 and C=128 measured, that's **wrong** — it's a **step at C≥4**, then flat:

| C | 1 | 2 | 4 | 8 | 16 | 32 | 64 | 128 |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| Δ vs certified | **+1.1%** | **+1.5%** | −3.8% | −3.5% | −4.6% | −4.2% | **−2.8%** | **−3.7%** |

C=1 and C=2 sit *above* certified; everything from C=4 up sits 2.8–4.6% below with no trend across that range. The cause engages once batching begins and does not worsen with width — a materially different fingerprint from what the earlier text implies, and a better clue for whoever chases it next.

## Scope

`bench/ladder38/RESULTS.md` + one raw series. Outside the benchmark closure.


tbraun96:
**Carried by stack #875 (`stack/content-and-records`)** — merge-tree returns #875's own tree: both files (`bench/ladder38/RESULTS.md`, `bench/ladder38/atlas_c64_c128_dgx1_20260820.json`) are already in that stamped stack. This PR closes with landing evidence when #875 merges, not before. If any hunk did not make it across, point at it and I will lift it onto a fresh branch.

DrRainbows:
Closing as already on `main`. This landed via #875 (`7fe92343`, an ancestor of `origin/main`). Named artifact: `bench/ladder38/RESULTS.md:1274` cites `atlas_c64_c128_dgx1_20260820.json`, and both files from this PR's diff (`RESULTS.md`, `atlas_c64_c128_dgx1_20260820.json`) are byte-identical on `main`. Every commit here cherry-picks empty onto current `main` (`c2948e3f`).

PR #648, opened 2026-08-20
DFlash2 spec decode for Qwen3.8-27B: fastest
  single-stream on GB10 (54.5 tok/s median)

rrstesiak:
This PR ships DFlash2 speculative decoding for Qwen3.8-27B on GB10. DFlash2 ([announcement](https://inco.ai/blog/dflash2/), [official drafter](https://huggingface.co/incoai/Qwen3.8-27B-DFlash2)) was released August 18. The record below was set August 19, within a day of the method existing publicly, and this PR delivers the complete integration on day two. It is the fastest single-stream configuration this hardware has produced.

## Headline

**54.5 tok/s median, MinHeap code prompt, single stream, temp 0, median of 3 runs, byte-identical outputs across runs.** Receipts posted in the dev channel. Reference points, reported from earlier baselines: no-spec serial floor ~15.8, MTP lane 22-30. This is drafter-class spec decode working on 3.8 for the first time (DFlash1 collapsed to 11.7% accept on release day).

## Review scope

This branch is stacked on the repair-479 line (the #519 stack plus the pr-511/batch6 merges already in flight), because that is the exact build the record was measured on; re-basing onto current main would decouple the code from its receipts. **The DFlash2 work is the top commit, `acaf9533`, 18 files.** Review that commit; everything beneath it is the stack already under review elsewhere.

## What's in it

- **DFlash2 integration**: full drafter support for the inco.ai Qwen3.8 drafter, verify path and semantics validated against the shipping gates.
- **`w4a16_gemv_batch8_rt2`**: register-tiled T=2 GEMV for the verify sweep. Per-output FMA order unchanged, bit-exact vs the shipping batch8 kernel (batchm_bench gate 4 PASS, all shapes and M). +17-26% GB/s on every verify shape.
- **FP8 port of the rt2 structure** for the propose GEMMs.
- **DSpark lane** for the RadixArk checkpoint, retained as a second drafter path: Markov fixup head (sequential per-row gather, GEMV, residual add, argmax; zero new kernel classes) plus the SpecForge row-semantics alignment (`ATLAS_DSPARK_SHIFT=1`). Both pieces are load-bearing; each alone regresses accept. Verified at 33.6 tok/s MinHeap before being superseded by DFlash2.

## Correctness

[...]

PR #649, opened 2026-08-20
DFlash2 base layer: record kernels, rt2 tier wiring, record defaults (train 1/2)

rrstesiak:
Base layer of the DFlash2 PR train (1/2). Everything from #648's top commit, cut clean onto current main, plus the completion work:

## What this is

- **DFlash2 + DSpark drafter support** for Qwen3.8-27B (the #648 stack, re-based onto main with the tier-refactor conflicts resolved).
- **rt2 register-tiled GEMV wired through the `W4a16BatchmTiers` resolver**: the width-8 tier now resolves via the rt2-preferring helper, so every consumer (dense FFN, attention, GDN, lm_head, MTP) inherits it from one site. Bit-exact vs classic batch8 (same per-row FMA chain, batchm_bench gate 4; independently confirmed BIT-EXACT by @rsafier's A/B on #648's port).
- **Record flags flipped to defaults** per the house env-var rule: `OPTION_B`, `EAGLE_FIX`, and `DRAFTER_FP8` are now default ON with `=0` kill switches. A bare `--dflash` launch is the record path; the env block in #648's recipe is no longer needed on this branch.

## Scope note for review

The DFlash2/DSpark work is two commits on top of current main; review those. The defaults flip deliberately reaches beyond the 3.8 lane (EAGLE_FIX default-on also covers the k2 and non-3.8 verify paths); the 07-09 accept-collapse that fix closes argues ON is correct everywhere, but it is called out here rather than hidden.

## Train

@rsafier's memory-buffer and LoRA work stacks on this branch as train 2/2, under his own PR. #648 stays open as the receipts anchor for the published 54.5 (independently reproduced at 54.6) and closes when the train lands.

Fresh receipts for this branch will follow under the watermark protocol (single stream, median of 3, completion tokens, byte-identity) before any new performance claim is made.

### DIFF
--- a/crates/spark-model/examples/batchm_bench.rs
+++ b/crates/spark-model/examples/batchm_bench.rs
@@ -55,6 +55,9 @@ const SHAPES: &[(&str, u32, u32)] = &[
     ("qkv/o    N=5120   K=5120 ", 5120, 5120),
     ("ffn_up   N=17408  K=5120 ", 17408, 5120),
     ("ffn_down N=5120   K=17408", 5120, 17408),
+    // Qwen3.8-27B GDN in_proj qkvz — 48 launches/step in the DFlash2 M=8
+    // verify (nsys node-trace 2026-08-19; ran ~148 GB/s on shipping batch8).
+    ("gdn_qkvz N=16384  K=5120 ", 16384, 5120),
     ("lm_head  N=248320 K=5120 ", 248320, 5120),
 ];
 
@@ -149,8 +152,14 @@ fn launch_gemm(
     m: u32,
     n: u32,
     k: u32,
+    // `Some(ldb)` for `w4a16_gemm_t` — its compiled signature grew a 9th
+    // `ldb` param (lm_head-716 fix). Launching it with 8 args makes
+    // cuLaunchKernel read one-past-the-end of the param array: the
+    // documented host-SIGSEGV class (see gemm_dense.rs on the bf16 _v2
+    // sibling). This was THE bench segfault of 2026-08-19.
+    ldb: Option<u32>,
 ) -> Result<()> {
-    KernelLaunch::new(g, kh)
+    let mut l = KernelLaunch::new(g, kh)
         .grid([div_ceil(n, n_tile), div_ceil(m, 64), 1])
         .block([128, 1, 1])
         .arg_ptr(a)
@@ -160,8 +169,11 @@ fn launch_gemm(
         .arg_ptr(c)
         .arg_u32(m)
         .arg_u32(n)
-        .arg_u32(k)
-        .launch(0)
+        .arg_u32(k);
+    if let Some(ldb) = ldb {
+        l = l.arg_u32(ldb);
+    }
+    l.launch(0)
 }
 
 #[derive(Clone, Copy)]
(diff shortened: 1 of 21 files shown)


rsafier:
Rebased the stack (#650 → #651, and sibling #652) onto `d6a5f2bc`. The rustfmt pass cascaded cleanly — `cargo fmt --all -- --check` is now clean on all three, and each rebuilt with `ATLAS_TARGET_MODEL='*'` and re-tested (639/641 spark-model + 104 spark-server).

Two files are over the repo's 500-line cap on this branch, and they are not in the allow-list, so the cap check fails for this PR and for anything stacked on it:

| file | on `main` | on this branch |
|---|---|---|
| `crates/spark-model/src/weight_loader/dflash_loader.rs` | 331 | **575** |
| `crates/spark-model/examples/batchm_bench.rs` | 422 | **504** |

Flagging rather than fixing: both are yours in this layer, and a split done from my side would land on your desk as a conflict in code you are still moving. `batchm_bench.rs` needs 4 lines; `dflash_loader.rs` wants a real split (the `DflashSubConfig` / rope structs are the obvious seam).

Separately — thanks for taking the FP8 lm_head sharing (`8ed7dcaf`) into the base layer. That was on my list to extract and it is better placed there.

One heads-up from a bench run on my side, unrelated to the format pass: `batchm_bench`'s `gemm_t` arm crashes on `main` with `CUDA_ERROR_INVALID_VALUE` (`grid=[40,1,1]`) because `w4a16_gemm_t` grew a trailing `ldb` parameter the bench never passed. Fixed standalone against `main` in #653, along with a bit-parity gate and fix for `w8a16_gemv_batch4/16` — that kernel was **not** byte-identical to the M=1 path (up to `max|delta|=4.0` at qwen3.8 GDN shapes), which is relevant to any verify-vs-decode divergence hunting on this lane.

rsafier:
Correcting my last comment: I said I would flag the two over-cap files rather than fix them, and have now done the split after all — the stack could not go green otherwise. It is in #650 as `chore(loc)`, and cherry-picked onto #652.

    weight_loader/dflash_loader.rs   575 -> 454  (+ dflash_loader/config.rs 134)
    examples/batchm_bench.rs         504 -> 461  (+ batchm_bench/dtype.rs     55)

Both are pure moves — no logic, no signatures, no behaviour. `dflash_loader.rs` splits where describing the checkpoint ends and loading it begins: the `serde` shapes (`DflashConfig`, `DflashRopeScaling`, `DflashSubConfig`, their defaults) go to `dflash_loader/config.rs` and are re-exported from the parent, so every path that resolved before still resolves. Follows the `laguna.rs` + `laguna/` convention already in that directory.

Heads-up since it is your file: if you have edits in flight to either, this will conflict, and the resolution is "take both sides" — nothing moved *within* the blocks, they only changed file. Happy to drop the commit and let you do the split your way if you would rather own the seam; say the word and I will rebase it out.

tbraun96:

Landing commit: `fbca9ca5` — *"atlas#812: C=128 ladder, DFlash2 as a mandatory gate, and the gamma>8 …"* — **verified on `main`** (merge commit of #818, which merged earlier today).

DFlash2 base layer: record kernels + rt2 tier wiring. Every file this PR touches is present on `main` except `crates/spark-model/examples/batchm_bench.rs`, a bench example rather than a shipped path.

PR #650, opened 2026-08-20
Memory buffers: size the DFlash allocations to what a request can reach, + allocation ledger (train 2/2, merge after #649)

rsafier:
**MERGE AFTER #649.** Train 2/2 — @rrstesiak's `feat/38-dflash2-base` is the base layer; this stacks on it. The LoRA work stacks on *this* one (linked below).

Because #649's branch lives on a fork, GitHub cannot take it as a base here, so this PR targets `main` and its first two commits are #649's. They disappear from the diff the moment #649 lands. **The six commits to review are mine, listed below.**

## What this is

Memory-buffer work, cut from #604 as clean commits onto #649's base.

### 1. `feat(runtime)` — allocation ledger

Every device allocation records its size and call site; the live set rolls up by site and by file, printed at end of load.

The motivation is that memory questions here were being answered by reading code and guessing, and two of my own guesses were wrong in the same week: a buffer I was sure held ~21 GB holds 10 KB per slot, and a 9.7 GB region I called "outside the budget" turned out to be pre-flight-reserved all along. Neither survived contact with a number. The ledger found a 9.2 GB duplicate FFN repack and a pledge overshoot immediately. Cost is a HashMap insert/remove per allocation — noise against a `cudaMalloc` — and the table is only walked when something asks for the report.

### 2. `perf(vision)` — ViT scratch on first image, not at load

A text-only serve paid 2.2 GB for vision buffers no request would touch. The scratch now sits behind a `OnceLock`. Text-only serves never build it; image serves pay the same allocation, once, on the first request.

Wide-and-shallow diff (~100 call sites gain `self.scratch()`), no arithmetic change: `vision-fidelity` passes 14/14 geometry cells and 3/3 probes on Qwen3.8-27B-NVFP4 with the no-image control held.

### 3. `fix(mem)` — size the DFlash buffers to what a request can reach

Three per-sequence buffers were sized from the global `--max-seq-len` ceiling rather than anything a request uses, and each is paid n times at concurrency. That is what OOMed the box on a 256K/C=8 serve — and it OOMs long after a clean boot, because the buffers arrive lazily with the streams.

- ctx accumulator → min(request's own budget, DFlash ctx window, `--max-seq-len`). At 128K × 5 layers × 5120 BF16 the ceiling is **6.7 GB each**, so 8 streams asked for 53.7 GB.
- whole-prompt hidden capture → bounded by the **same** window, because `prefill_drafter` writes into the drafter's KV and a longer capture is memory nothing can read: 1342 MB at `--max-seq-len 131072` against a 16K window holding 168 MB of it. Both now read ONE definition (`dflash_ctx_cap`), since letting them drift is the bug being closed.
- verify intermediates and SSM pools size from the real γ and true ceiling; the drafter's allocations are counted **inside** the KV budget, so `--gpu-memory-utilization` means what it says on a DFlash serve.

Correctness: `commit_ctx` already slides a watermark when the accumulator fills, keeping the newest half — a smaller window is an already-exercised path. A prompt past the window simply doesn't get the whole-prompt drafter prefill (the consume-site coverage check handles it; blind beats poisoned). Costs acceptance on very long cold turns, not correctness. `ATLAS_DFLASH_CTX_CAP=<tokens>` raises it, `0` disables. Pure-MTP serves keep the full ceiling — the narrowing is only sound because the drafter's own capacity binds, and that reasoning doesn't transfer.

### 4. `fix(mem)` — refuse a drafter that will not fit, before it swaps the host

A drafter whose weights do not fit alongside the target was discovered the hard way: the load proceeded, the box went to swap, and the machine became unusable long before any error surfaced. On unified memory there is no separate device budget to bounce off — host and GPU are the same 121 GB — so an over-large drafter takes the whole machine down rather than failing a device allocation.

A footprint pre-flight now runs before the drafter loads: it sums the checkpoint on disk, adds what the target and KV budget have already pledged, and refuses with the arithmetic spelled out and the knobs named. A passing boot logs the same arithmetic, so the margin is visible on a good run too.

One trap, found when the pre-flight's first live run cheerfully measured `weights 0.00 GB`: HF snapshot dirs are entirely symlinks into `blobs/`, and `DirEntry::metadata` does not follow symlinks — `std::fs::metadata` does. A size check that reads every file as empty is worse than none, because it reports a pass.

### 5. `chore(loc)` — split the two files pushed over the 500-line cap

Neither is over on `main`; the base layer grows both past it and neither is allow-listed, so the cap check fails for this PR and everything stacked on it. Both are pure moves — no logic, no signatures, no behaviour:

| file | before | after |
|---|---|---|
| `weight_loader/dflash_loader.rs` | 575 | 454 (+ `dflash_loader/config.rs` 134) |
| `examples/batchm_bench.rs` | 504 | 461 (+ `batchm_bench/dtype.rs` 55) |

`dflash_loader.rs` splits at the seam between *describing* the checkpoint and *loading* it — the `serde` config shapes move out, re-exported from the parent so every existing path resolves, following the `laguna.rs` + `laguna/` convention already in that directory. `batchm_bench.rs` moves its dtype helpers and deterministic RNG behind `#[path]` so Cargo doesn't treat them as a second example target.

### 6. `perf(lm-head)` — `--lm-head-dtype fp8` reads the checkpoint's own bytes

This checkpoint family ships `lm_head.weight` as FP8 E4M3 with a per-row scale. `--lm-head-dtype fp8` was re-quantizing it anyway — dequantize to BF16, quantize back to FP8 — a lossy round trip that lands where it started, plus a duplicate ~1.27 GB copy of a tensor already resident for the model's lifetime. It now takes the same `native_fp8_lm_head_share` the DFlash drafter tail uses.

Padded rows are safe: the tensor is row-major `[rows, hidden]` and unsloth pads 248077 → 248320 at the END, so reading the first `vocab_size` rows is a prefix.

**Not a default change.** `default` still lets the model config choose, which on this checkpoint means NVFP4. Worth stating for whoever picks: the NVFP4 head is not the checkpoint's own precision either — there is no packed NVFP4 head in the store, so `default` performs FP8 → dequant BF16 → NVFP4, a double quantization of the most precision-sensitive projection in the model. Measured here, the FP8 head reproduces the full-BF16 greedy answer byte-for-byte while NVFP4 walks a different one. Which is *better* is a long-generation quality question, not a throughput one — this flag's own documentation records a model passing short-prompt coherence at both fp8 and nvfp4 while collapsing 10/10 on an agentic harness. Gate it there before changing any default.

The kernel work that makes the FP8 head affordable (it was slower than BF16 because its dispatch fell to a per-token loop above M=2) is **not** here — it is standalone on `main` in #653, since it has nothing to do with this train.

## Scope discipline

- **γ resolution stays with the base layer.** This PR consumes whatever γ resolves to rather than re-deciding it underneath #649. #604 has a fix that reads the trained block size out of a DFlash2 checkpoint's `dflash_config`; it is deliberately *not* here, since it would change the base layer's defaults.
- The batched-verify half of the DFlash lane is not here either.
- No `cargo fmt` churn in #649's files: the pick reflowed `markov.rs`, `dflash2.rs`, `forward_block.rs`, `batchm_bench.rs` and `dflash_loader.rs`, and all of it was reverted rather than handed to you as conflicts.

## Verification

`ATLAS_TARGET_MODEL='*'` build clean · 639 spark-model + 104 spark-server tests · LoC cap clean (the two remaining violations, `batchm_bench.rs` 501 and `dflash_loader.rs` 558, pre-exist on #649's base — I trimmed `meta.rs` back to 500 rather than let my change push it over).

---
Stacks: #649 → **this** → LoRA (separate PR). Sibling: TUI/metrics (separate PR, independent).

### DIFF
--- a/crates/atlas-core/src/config.rs
+++ b/crates/atlas-core/src/config.rs
@@ -444,6 +444,12 @@ pub struct ModelConfig {
     /// is what the drafter's `fc` projection expects.
     #[serde(default)]
     pub dflash_capture_layers: Vec<usize>,
+    /// Resolved DFlash drafter γ (block size), set by the factory alongside
+    /// `dflash_capture_layers`. Sizes the SSM verify intermediate pools at
+    /// the ACTUAL K = γ+1 instead of the legacy 17-wide ceiling — at γ=8,
+    /// C=8 that ceiling alone cost ~12 GB of pool (2026-08-19 256K/C8 boot
+    /// ledger). `None` = DFlash inactive (or unknown → 17-wide fallback).
+    pub dflash_gamma: Option<usize>,
 
     /// LoRA adapter rank ceiling (`--max-lora-rank`). `0` = LoRA disabled.
     /// Set programmatically before model build (never parsed from the HF
--- a/crates/atlas-core/src/config/factory.rs
+++ b/crates/atlas-core/src/config/factory.rs
@@ -121,6 +121,7 @@ impl ModelConfig {
             mtp_transformer_layers: 0,
             rotary_dim: 0,
             dflash_capture_layers: Vec::new(),
+            dflash_gamma: None,
             adapter_max_rank: 0,
         }
     }
--- a/crates/spark-model/examples/batchm_bench.rs
+++ b/crates/spark-model/examples/batchm_bench.rs
@@ -55,6 +55,9 @@ const SHAPES: &[(&str, u32, u32)] = &[
     ("qkv/o    N=5120   K=5120 ", 5120, 5120),
     ("ffn_up   N=17408  K=5120 ", 17408, 5120),
     ("ffn_down N=5120   K=17408", 5120, 17408),
+    // Qwen3.8-27B GDN in_proj qkvz — 48 launches/step in the DFlash2 M=8
+    // verify (nsys node-trace 2026-08-19; ran ~148 GB/s on shipping batch8).
+    ("gdn_qkvz N=16384  K=5120 ", 16384, 5120),
     ("lm_head  N=248320 K=5120 ", 248320, 5120),
 ];
 
@@ -149,8 +152,14 @@ fn launch_gemm(
     m: u32,
     n: u32,
     k: u32,
+    // `Some(ldb)` for `w4a16_gemm_t` — its compiled signature grew a 9th
+    // `ldb` param (lm_head-716 fix). Launching it with 8 args makes
+    // cuLaunchKernel read one-past-the-end of the param array: the
+    // documented host-SIGSEGV class (see gemm_dense.rs on the bf16 _v2
+    // sibling). This was THE bench segfault of 2026-08-19.
+    ldb: Option<u32>,
 ) -> Result<()> {
-    KernelLaunch::new(g, kh)
+    let mut l = KernelLaunch::new(g, kh)
         .grid([div_ceil(n, n_tile), div_ceil(m, 64), 1])
         .block([128, 1, 1])
         .arg_ptr(a)
@@ -160,8 +169,11 @@ fn launch_gemm(
         .arg_ptr(c)
         .arg_u32(m)
         .arg_u32(n)
-        .arg_u32(k)
-        .launch(0)
+        .arg_u32(k);
+    if let Some(ldb) = ldb {
+        l = l.arg_u32(ldb);
+    }
+    l.launch(0)
 }
 
 #[derive(Clone, Copy)]
@@ -270,6 +282,29 @@ fn correctness_gate(
         "  gate 3   PASS: batch8 @M=8 vs CPU f64 ref, {rows} rows (worst {:.2}x tol)",
         worst
     );
+
+    // Gate 4: pipelined batch8 variants must be BIT-EXACT vs shipping batch8
+    // at every M — same virtual-lane order by construction; this proves it.
+    for &(kn, kh, _) in kernels.iter().filter(|(kn, _, _)| kn.starts_with("batch8_pf") || kn.starts_with("batch8_rt")) {
+        for &m in M_SWEEP {
+            zero_c(g)?;
+            launch_batchm(g, b8, a, b, bs, c, m, n, k)?;
+            g.synchronize(0)?;
+            let reference = read_c(g, c, m as usize * n_us)?;
+            zero_c(g)?;
+            launch_batchm(g, kh, a, b, bs, c, m, n, k)?;
+            g.synchronize(0)?;
+            let got = read_c(g, c, m as usize * n_us)?;
+            if reference != got {
+                let bad = reference.iter().zip(&got).filter(|(x, y)| x != y).count();
+                bail!(
+                    "GATE FAIL: {kn} != batch8 at M={m} ({bad}/{} elems differ)",
+                    got.len()
+                );
+            }
+        }
+        eprintln!("  gate 4   PASS: {kn} BIT-EXACT vs batch8 at all M");
+    }
     Ok(())
 }
 
@@ -317,6 +352,50 @@ fn main() -> Result<()> {
             "w4a16_gemv",
             "w4a16_gemv_batch16",
             Kind::Batchm { max_m: 16 },
+        ),
+        // Lever-1 candidates (2026-08-19): software-pipelined batch8 —
+        // weight prefetch, bit-identical FMA order (gate 4 proves it).
+        // provenance-id: 526f6e616c6420522e205374657369616b
+        (
+            "batch8_pf",
+            "w4a16_gemv",
+            "w4a16_gemv_batch8_pf",
+            Kind::Batchm { max_m: 8 },
+        ),
+        (
+            "batch8_pfree",
+            "w4a16_gemv",
+            "w4a16_gemv_batch8_pf_free",
+            Kind::Batchm { max_m: 8 },
+        ),
+        // v2: activation row-ahead prefetch (the M-scaling fix; weight
+        // prefetch benched as a wash 2026-08-19). pf3 = both prefetches.
+        (
+            "batch8_pf2",
+            "w4a16_gemv",
+            "w4a16_gemv_batch8_pf2",
+            Kind::Batchm { max_m: 8 },
+        ),
+        (
+            "batch8_pf3",
+            "w4a16_gemv",
+            "w4a16_gemv_batch8_pf3",
+            Kind::Batchm { max_m: 8 },
+        ),
+        // v3: register-tiled, T outputs/thread (activation traffic, load
+        // instructions, and FMA-chain ILP all ÷T). Grid stays ceil(N/4):
+        // surplus blocks early-exit, coverage and outputs unchanged.
+        (
+            "batch8_rt2",
+            "w4a16_gemv",
+            "w4a16_gemv_batch8_rt2",
+            Kind::Batchm { max_m: 8 },
+        ),
+        (
+            "batch8_rt4",
+            "w4a16_gemv",
+            "w4a16_gemv_batch8_rt4",
+            Kind::Batchm { max_m: 8 },
         ),
         ("gemm_m64", "w4a16", "w4a16_gemm", Kind::Gemm),
         ("gemm_t", "w4a16", "w4a16_gemm_t", Kind::GemmT),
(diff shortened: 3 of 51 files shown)


rsafier:
Pushed `8e859a6c` — the pledge-fidelity fix this PR's allocation ledger made findable.

**What the ledger showed** (27B + DFlash2, γ=8, bs=8, util 0.65): planner pledged 79.1 GB, end-of-load tracked 95.5 GB. The bulk was the SSM verify-pool family: **13.7 GB allocated against a 1.3 GB reserve**.

**Root cause:** preflight's `spec_on_pool` omitted `args.dflash`, while `TransformerModel::new` forces the verify pools ON for any DFlash serve (`has_mtp |= dflash`) — so `ssm_pool_reserve_bytes` short-circuited to the base term and the whole K=γ+1 intermediate/checkpoint family landed outside the budget. (The `args.dflash` uniform-verify parameter was already being passed — it just never got read past the `!spec_on` early return.) The 4 GB speculative CUDA headroom had the same omission.

**The fix:** `spec_on_pool` and the headroom predicate include `args.dflash`; the pool reserve's draft count is the drafter's real γ, peeked from its `config.json` (`dflash_config.block_size`) since the drafter loads long after preflight — a failed peek falls back to the same 16 ceiling the pool allocates at unknown γ, so the miss direction is over-reserve, never under. Plus a pledge reconciliation line next to the ledger: tracked-live vs pledged, WARN naming the gap.

**Measured:** reserve 4.5 → 18.6 GB; an honest 0.65 boot now *refuses* (`No memory left for KV cache`) instead of overshooting its promise by 16 GB; at 0.80 the same serve gets **more** KV than the dishonest 0.65 boot had (7.6 → 11.8 GB, 248K → 387K tokens), logs `util pledge honored: 93.7 GB within 97.3 GB`, and serves C=16 at 115 tok/s 16/16 coherent.

Note for existing recipes: DFlash serves that "worked" at low util were spending past their pledge; they now need the util their footprint actually requires.

rsafier:
Pushed `6f0856af` — the follow-up the pledge-reconciliation line flagged: the MTP propose pool was the LAST tracked allocation still landing outside the util pledge.

`MtpHead::new` allocates its own paged KV pool after KV sizing (per-seq blocks × the MTP concurrency cap, bounded by the main pool's block count) and its comment claimed the cost was "well inside the serve reserve" — a reserve that never existed (~0.9–1.6 GB depending on ctx × seqs). The budget site now mirrors the pool arithmetic and pre-charges it, two-pass on the cap so the reserve can only exceed the actual pool, never trail it; the head-side comment names the mirror so they can't drift silently.

Measured (27B + DFlash2, 128K/bs8, util 0.80): reserve line 1.6 GB, actual pool 1.5 GB, KV 11.8 → 11.2 GB, pledge headroom 3.6 → 4.3 GB, smoke correct. With this, every tracked device allocation is inside the pledge; the ~2.5 GB untracked residue (CUDA context, pinned staging — and the duplicate module SASS until #705) sits inside the reserved headroom.

rsafier:
Pushed `4b1a55f4` — the tail of the same predicate family: three lesser sites that treated `--dflash` as "not speculative".

- **spec_tokens SSOT**: the "rows one speculative step occupies" term was copy-pasted in `preflight.rs` and `kv_cache.rs`, both returning 1 for a DFlash serve whose verify step is γ+1 rows (inert today only because the 8192 prefill floor dominates the max). Now one γ-aware fn, using the same drafter-config peek as the pool reserve.
- **validate**: an explicit `--num-drafts` on a DFlash serve was told to "add --speculative" — wrong remedy; the flag is ignored because the drafter's trained γ decides. The violation now says that and points at `--dflash-gamma` with the acceptance-collapse caveat.
- **EP worker**: the spec-mismatch guard would bail a `--dflash` worker as "started WITHOUT any --speculative flag". DFlash+EP is unused today; the guard now counts dflash as speculative so it can't lie about it when it becomes real.

2412 spark-server tests pass on this branch (the scheduler suite rides the bin target). With this, `grep` finds no remaining spec-method predicate in spark-server that omits dflash — the `ssm_h_fp16` gate refuses it explicitly and everything else routes through the fixed sites.

rsafier:
Pushed `138065bc`: the KV budget's AUTO self-relative path now reads **Atlas-own from this PR's alloc ledger** (`gpu.live_bytes()`) instead of the free-memory delta, fixing #740 — on unified memory the delta counted OS page-cache growth from streaming safetensors as Atlas-owned memory, which refused cold-cache boots with >100 GB actually available (externally reported against wip/dflash2-full). The free-delta path stays as fallback for backends with no ledger; the physical `.min(actual_free − reserve)` clamp is unchanged. Validated live on GB10: boot logs `Atlas-own 59.4 GB live in the alloc ledger; 11.7 GB of co-tenant/page-cache use excluded` and serves normally.

tbraun96:
Triage of this stack, with the numbers, because the good news and the bad news are both specific.

**The stack is internally coherent.** #650 → #651 → #655 is a registered chain (`base=main` → `stack/dflash2-mem-buffers` → `stack/lora-hybrid`), and merging each layer down into the one below is **clean** — no conflicts at any layer. That matters: under this repo's rule a stack pays **one** certification campaign, at the base, so these three cost ~3 GPU-hours between them rather than ~9.

**What it costs to get there.** The composed base conflicts with today's `main` on **42 files / 135 hunks**:

| area | files |
|---|---|
| `crates/spark-model/src/layers/dflash_head` | 6 |
| `crates/spark-server/src/scheduler` | 5 |
| `crates/spark-model/src/lora` | 5 |
| `crates/spark-model/src/model/trait_impl` | 3 |
| `crates/avarok-core/src/config/parsers` | 3 |
| …and 20 more | |

That is a re-cut, not a rebase, and it is your feature work — the LoRA delta application and the DFlash C≥2 ctx-row path — so I have not touched it.

**One diagnostic worth knowing.** `git merge-base --is-ancestor pr650 pr655` reports that #655 does **not** contain #650, which reads like "this isn't a stack". The base refs say otherwise; the layers were simply cut from an older #650 and have drifted. Ancestry answers *"is this built on that now"*; the base refs answer *"what was this declared against"*. For a stack that has been sitting a while, only the second is meaningful — worth remembering before anyone concludes the chain is broken.

**Containment**, since some PRs in this backlog turned out to be already landed: these are not. Full-population against `main` — #650 **94/1712 absent (5%)**, #651 **119/2223 (5%)**, #655 **179/2895 (6%)** — and roughly half of each residue is real kernel and Rust code, not the `ATLAS_*` → `AVAROK_*` rename. #652 (`stack/tui-prefill-metrics`, based on `main` independently) is **18% absent** and mostly non-rename, so it is live too but not part of this chain.

Happy to do the re-cut if you would rather I did, or to run the single campaign once the base is current. The queue ahead of it is #1144 (in the merge queue now) and then #1127 + #582 batched into one campaign.

tbraun96:
Sharpening my earlier offer, because I measured the conflicts properly and the answer changed.

I classified all **135 conflict hunks** by asking, for each, whether the two sides differ *only* in the #1101 identifiers (`ATLAS_*`, `atlas-*`, `atlas_*`) or in substance:

```
rename-only hunks:   0
semantic hunks:    135
```

**None of it is mechanical.** Where five other PRs I closed today were pure rename residue, this stack's conflicts are real divergence between your DFlash/LoRA work and where `main` has moved — `dflash_head` (6 files), `scheduler` (5), `lora` (5), `trait_impl` (3).

So my earlier *"happy to do the re-cut if you would rather I did"* was optimistic, and I would rather withdraw it than do it badly: resolving 135 semantic hunks across someone else's kernel and scheduler work needs your intent, not my guesses. What I can usefully do is everything after it — once the base is current, the single campaign for all three layers, the banking, and the landing.

The structural finding stands and is worth having: **#650 → #651 → #655 merge down into each other cleanly**, so this is one campaign for three PRs rather than three. That part is free; only getting the base current is not.

For context on the queue: #1149 (a composed branch carrying #1127 and #582) is mid-campaign now — the first batched run, two PRs in one. This stack is the next batch after it, whenever the base is ready.

tbraun96:

Re-checked today against `main` `9e597531`: all 51 functions/types this PR adds exist there — e.g. `target_ships_native_fp8_lm_head` at `crates/spark-server/src/main_modules/serve_phases/weights.rs:270` and the `effective_block_size_*` loader tests at `crates/spark-model/src/weight_loader/dflash_loader/loader_tests.rs:89`. The branch reads as conflicting only because it still carries #649's base commits, which `main` has since restructured.

Thanks for the memory-buffer work — it is serving. No branch deleted.

PR #651, opened 2026-08-20
LoRA on hybrid Qwen3.8-27B: apply the deltas, close both perf regressions, batch verify under an adapter (train 2/2, merge after #650)

rsafier:
**MERGE AFTER #650, which merges after #649.** This is the top of the train: base layer #649 → memory buffers #650 → **LoRA (this)**.

Its base is set to `stack/dflash2-mem-buffers`, so the diff here is only the LoRA commits. When #650 lands this retargets to `main` cleanly.

## What this is

LoRA support for the hybrid Qwen3.8-27B architecture, cut from #604. Three real community adapters now load and serve, two of them with no opt-in flag at all.

### Enablement

- **Dense-FFN deltas actually applied** (`apply_lora_gate_up` before silu_mul, `apply_lora_down` after the down GEMM). Loading an adapter and silently not applying half of it is worse than refusing it.
- **Hybrid layers**: this model is 64 layers = 16 full-attention + 48 linear-attention, and **all 64 carry the dense SwiGLU FFN**. The pool walked the wrong layer set and dropped the hybrid FFN pairs.
- **`regex` target_modules**, which is how Novelist1.0 ships its targets.
- **GDN `out_proj` deltas** — the last module real qwen3.8 adapters need. The delta is applied *after* the TP all-reduce, which is the only correct place for it.

Tested adapters, all loading with no opt-in: `Dxniz/Novelist1.0-27b-Adapter` (regex, r=64), `nico248000000000/Qwen3.8-27B-cyber-LoRA` (r=8, 608 tensors), `MuXodious/Qwen3.8-27B-absolute-heresy-LoRA` (r=32).

### Performance

Both regressions were root-caused, not worked around:

| | before | after |
|---|---|---|
| prefill | 164 | **617** tok/s |
| C≥2 decode | 4.7 | **11.9** tok/s/stream |

- Prefill: the MMQ finalize was **freeing the weights the fallback path still needed**.
- Decode: the m=1→m=2 GEMV/GEMM threshold sent adapted rows to a kernel that pads them; keeping small m on GEMVs fixes it.
- `dense_gemm_tc`'s B-tile load is coalesced (K walks fastest) — +15% LoRA prefill.
- The pool derives its rank from the resident adapters instead of padding every slot to 64, so an r=8 adapter stops paying 8× the bandwidth for zeros.

### Speculative decode under an adapter

DFlash aggregate throughput was **flat at ~34 tok/s from C=1 to C=4** with an adapter resident (against 39 → 71 → 79 without), because a `self.lora.is_none()` guard sent every sequence to the per-sequence verify loop. LoRA is not a barrier there: every weight-bearing op the batched path touches carries its delta on the batched variants, and all rows necessarily share ONE adapter because the single-active guard already refuses a mixed batch upstream — so a uniform delta across the batch is the *correct* delta. `ATLAS_LORA_NO_BATCH_VERIFY=1` restores the refusal.

### Mixed adapters

A batch mixing adapters used to fail wholesale. Requests are now serialised by an admission cohort filter instead — 4 concurrent mixed base+adapter requests all return real content.

## Measured on the current stack

MinHeap prompt, 1500 max tokens, temp 0, DFlash2 on, aggregate tok/s:

| C | base | +novelist |
|---|---|---|
| 1 | 56.4 | 45.6 |
| 2 | 93.6 | 57.6 |
| 4 | 131.3 | 78.7 |
| 8 | 164.6 | 95.9 |

Accept holds at 85/85/80/74% (base) and 85/84/78/74% (novelist).

## Known gap, pre-existing and not from this PR

With an adapter resident there is **no selector for *no* adapter**: a request naming the base model returns byte-identical output to the adapter request (`8838f242862b39bb`) and differs from a true adapter-free base completion (`2f4598630ee14ef0`). That is `resolve_request_adapter_slot` behaving as documented — a `model` that is not a resident adapter name falls through to slot `-1`, "defer to the installed active adapter". Worth its own issue and a reserved base slot; flagging it rather than leaving it to be discovered. Practical consequence for anyone benchmarking: a "base" arm measured on an adapter-loaded serve is not a base arm.

## Verification

`ATLAS_TARGET_MODEL='*'` build clean · 641 spark-model + 104 spark-server tests · LoC cap clean.


tbraun96:

Re-checked today against `main` `9e597531`: all 28 functions/types this PR adds exist there — e.g. `unsupported_target_modules_are_refused_and_name_the_escape_hatch` at `crates/spark-model/src/lora/types_tests.rs:221` and `rank_gate_is_independent_of_the_target_gate` at `:258`. The conflict badge comes from the stacked base (#650 / #649), not from this PR's own delta.

Thanks for the hybrid-LoRA work — it is on main. No branch deleted.

PR #653, opened 2026-08-20
Batched GEMV bit-parity: two arms that were not bit-identical to M=1 (w8a16 + FP8 vocab), gates + fixes

rsafier:
Standalone against `main` — no relation to the #649 train.

**Two batched GEMV arms in this repo are not bit-identical to the M=1 kernel they claim parity with.** Both are fixed here, and both now have a gate that fails loudly rather than a header comment that asserts.

## Why this class of defect is worth its own PR

Under strict-argmax accept, a drafter cannot change token identity — so a spec-on vs spec-off divergence at temp 0 is **target-side arithmetic by construction**. A batched verify arm that isn't bit-identical to the M=1 decode arm is exactly that: speculation on runs the batched kernel, speculation off runs M=1, same weights, same activations, different reduction order, different logits. Neither kernel is wrong in isolation; the bug is that two arms of one op disagree, which is why it survives ordinary review and ordinary tolerance tests.

## Defect 1 — `w8a16_gemv_batch4/16`

Two independent bugs in `kernels/gb10/common/w8a16_gemv_batch4.cu`:

**Paired contraction.** `acc[t] += lo*w0 + hi*w1` sums two products before accumulating — one fewer rounding step per pair than `w8a16_gemv`'s `acc += a*w` chain. Now one element per statement.

**Tail threads returned before a barrier.** `if (n >= N) return;` sits above a cross-warp reduction with a second `__syncthreads()`. Threads exiting before a barrier others still reach is UB — the same class as the earlier `w4a16_gemv` tail fix. They're kept alive and inert.

Measured on `main` before the fix, and it is not exclusive to qwen3.8 — the legacy 512×2048 shapes diverge too:

```
legacy   [  512 x 2048]  batch16 M=16  diff_elems=3    max|delta|=0.125
qwen3.8 qkvz    [16384 x 5120]  batch16 M=16  diff_elems=56   max|delta|=2.000
qwen3.8 out_proj[ 5120 x 6144]  batch16 M= 8  diff_elems=18   max|delta|=4.000
FAIL — batched W8A16 verify arm is NOT byte-identical to sequential decode.
```

After: `byte-identical=true, diff_elems=0` at every shape and every M.

## Defect 2 — `dense_gemv_fp8w_batch2`, on the target's LM head

Found while chasing why the FP8 head ran *slower* than BF16. `mac4` sums four products before accumulating where `dense_gemv_fp8w` accumulates one per statement. Its header, and the dispatch comment above it, both claimed bit-parity with two M=1 GEMVs.

```
small       [   512 x 2048]  diff_elems=1    max|delta|=0.015625
mid         [  8192 x 5120]  diff_elems=1    max|delta|=1.000000
qwen3.8 lm_head [248320 x 5120]  diff_elems=78  max|delta|=2.000000
```

**78 differing logits on the last projection before the argmax**, and this arm fires at `num_tokens==2` — the K=2 verify — while serial decode takes M=1. That is a spec-on/spec-off divergence carrier on the most precision-sensitive tensor in the model. Zero at every shape after the fix.

## Also here: the FP8 vocab projection was never batched above M=2

The same investigation turned up a plain performance bug. The FP8 LM-head dispatch had a batched arm only at M=2 and fell to a **per-token loop** above it. At a 248K vocab the head is ~1.27 GB of FP8, so an M=8 verify read the whole head 8 times, and a cross-sequence verify at C=8 (8 × γ+1 = 64 rows) read it **64 times**. That — not bandwidth — is why the FP8 head lost to BF16 while carrying half the bytes.

`dense_gemv_fp8w_batchm` (M≤8) reads each weight chunk once for every row, and above M=8 the dispatch chunks in groups of 8 rather than dropping to single rows. Written to the M=1 chain deliberately: same k order, and the per-row scale applied to each thread's partial **before** the reduction, because `(a+b)*s` and `a*s + b*s` round differently.

| C | per-token loop | batched | |
|---|---|---|---|
| 1 | 41.7 | 54.6 | +31% |
| 2 | 47.5 | 65.5 | +38% |
| 4 | 68.8 | 105.3 | +53% |
| 8 | 77.5 | 130.5 | +68% |

Completion byte-identical (sha `a7862f7d91db007e`) to both the per-token-loop FP8 run and a BF16-head run. C=1 gains because DFlash verifies γ+1 = 8 rows in one step even at a single stream.

## Also: `batchm_bench`'s gemm_t arm was crashing

`w4a16_gemm_t` grew a trailing `ldb` parameter and the bench was never updated, so it launched one argument short — `CUDA_ERROR_INVALID_VALUE`, a `grid=[40,1,1]` crash. The bench packs B contiguously so `ldb = n`. This bench is the harness that measures the batched-GEMV tier, and a crashing arm is an arm nobody is measuring.

## The gates

Two `gpu-examples` binaries, both exiting non-zero on any byte difference so they can gate CI on a GPU box. They compare BF16 **bytes** — not tolerances — against M independent M=1 calls, at the real production shapes, at every M the dispatch serves:

```
cargo run --release -p spark-model --features "cuda gpu-examples" \
      --example w8a16_batch_bitparity_microtest
cargo run --release -p spark-model --features "cuda gpu-examples" \
      --example dense_gemv_fp8w_bitparity_microtest
```

Both PASS on this branch. The w4a16 gate is extended with qwen3.8 shapes as well; those already passed and are included because a parity gate covering only the shapes someone happened to look at is how this class survives.

## Verification

`ATLAS_TARGET_MODEL='*'` build clean · 639 spark-model tests · LoC cap clean · both gates run on GB10 and pass, with the pre-fix failures quoted above measured on the same box.

### DIFF
--- a/crates/spark-model/Cargo.toml
+++ b/crates/spark-model/Cargo.toml
@@ -149,6 +149,10 @@ name = "w8a16_batch_bitparity_microtest"
 required-features = ["cuda", "gpu-examples"]
 
 [[example]]
+name = "dense_gemv_fp8w_bitparity_microtest"
+required-features = ["cuda", "gpu-examples"]
+
+[[example]]
 name = "q2_0_gemv_microtest"
 required-features = ["cuda", "gpu-examples"]
 
--- a/crates/spark-model/Cargo.toml
+++ b/crates/spark-model/Cargo.toml
@@ -145,6 +145,10 @@ name = "w8a16_gemv_batch4_microtest"
 required-features = ["cuda", "gpu-examples"]
 
 [[example]]
+name = "w8a16_batch_bitparity_microtest"
+required-features = ["cuda", "gpu-examples"]
+
+[[example]]
 name = "q2_0_gemv_microtest"
 required-features = ["cuda", "gpu-examples"]
 
--- a/crates/spark-model/examples/batchm_bench.rs
+++ b/crates/spark-model/examples/batchm_bench.rs
@@ -149,8 +149,14 @@ fn launch_gemm(
     m: u32,
     n: u32,
     k: u32,
+    // `w4a16_gemm_t` grew a trailing `ldb` param (padded-B stride; the
+    // production caller passes the twin's padded row count). Contiguous
+    // bench pack ⇒ ldb = n. `None` for kernels without the param — an
+    // extra trailing arg is a CUDA_ERROR_INVALID_VALUE, which is exactly
+    // how the drift surfaced (grid=[40,1,1] crash at the gemm_t arm).
+    ldb: Option<u32>,
 ) -> Result<()> {
-    KernelLaunch::new(g, kh)
+    let mut l = KernelLaunch::new(g, kh)
         .grid([div_ceil(n, n_tile), div_ceil(m, 64), 1])
         .block([128, 1, 1])
         .arg_ptr(a)
@@ -160,8 +166,11 @@ fn launch_gemm(
         .arg_ptr(c)
         .arg_u32(m)
         .arg_u32(n)
-        .arg_u32(k)
-        .launch(0)
+        .arg_u32(k);
+    if let Some(v) = ldb {
+        l = l.arg_u32(v);
+    }
+    l.launch(0)
 }
 
 #[derive(Clone, Copy)]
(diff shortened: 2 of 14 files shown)


rsafier:
One process note from cutting this PR, worth recording where it happened.

This PR was cut out of a longer-lived feature stack. The kernel and its launcher applied onto `main` with **zero conflicts** and then failed to compile: the launcher calls `anyhow::ensure!`, and the `use anyhow::{Result, ensure}` import had been added by a *different* commit that stayed on the stack. One missing import, invisible to the merge algorithm, because git diffs text and knows nothing about symbols.

This repo has already hit the mirror image of it: a PR merges `main` with zero conflicts, CI builds the PR **head**, GitHub reports MERGEABLE off the **textual** merge — and only building the *merge result* finds the break.

The build failure is the safe version, because it is loud. The dangerous version compiles and the semantics moved — a helper with the same name and a different contract, a default that changed, an enum that gained a variant. That one ships.

So for anything cut out of or moved between branches: build the result and run its tests on the base it will actually merge into. "It applied without conflicts" is evidence about text, not about code.

(Both parity gates in this PR pass on this branch; the compile break above was caught by building, not by git.)

PR #655, opened 2026-08-21
DFlash C>=2 correctness: ctx rows on the batched decode path, gate pin, pool uniformity (merge after #651)

rsafier:
**MERGE AFTER #651**, which merges after #650, which merges after #649. Top of the train; base is `stack/lora-hybrid` so the diff here is only these four commits.

## What this fixes

Three C≥2 correctness fixes from the WIP branch, plus the lever helper they need. Measured on this stack, MinHeap prompt, temp 0, 32K/8-seq profile:

| C | accept before | accept after |
|---|---|---|
| 2 | 59% | **84%** |
| 4 | 68% | **77%** |
| 8 | 63% | **62%** |

- **batched decode dropped every ctx row** — the drafter's context accumulator was not being fed on the batched decode path, so at C≥2 every sequence proposed from an empty context. This is the big one.
- **mtp_gate pinned to verify at C≤2** — the throughput arbiter flips between serial and batch-K mid-stream, and at temp 0 that forks two streams that should be identical.
- **SSM pools state uniformity stated explicitly** — inference over pool widths broke when γ-sized widths were in play.

`on_unless_zero` is a small but load-bearing addition: this branch only had `on_unless`, whose contract is the opposite (`=1` disables). The gate-pin lever documents `ATLAS_DFLASH_GATE_PIN_C2=0` as its opt-out, so reusing the existing helper would have inverted the switch silently — `=0` leaving the pin on, `=1` turning it off. That is worse than a compile error, because it looks like it works.

## Cross-sequence batched verify — ported, and it works

The batched K=γ verify from the WIP branch DOES port onto this base. My first
read of it as "a re-implementation against a different head" was wrong, and
worth correcting explicitly: of the conflicts, one was two doc comments for
the same function, one was two names for the same default-ON behaviour
(`ATLAS_DFLASH_EAGLE_FIX` and `ATLAS_DFLASH_UNIFIED_CTX` — now EITHER `=0`
disables, so neither name loses its kill switch), and one was HEAD's
LoRA-aware guard correctly superseding the older blanket refusal.

Then it landed and still did not run, because of a contradiction I introduced
resolving that third conflict. `can_batch_verify` picks `shape_ok` by
branching on `dflash_hidden_save`: `Some` selects the DFlash γ-block arm,
`None` the K=2..4 ladder. The eligibility list also carried
`&& dflash_hidden_save.is_none()`, which predates the DFlash arm and meant
"batched verify is for plain MTP only". Both present = the Some arm computes
a shape it can never use. Unreachable by construction, and from the outside
it looks like "concurrency does not amortise" rather than a disabled feature
— which is why a one-shot gate diagnostic is now permanent at debug level.

| C | before | after | |
|---|---|---|---|
| 1 | 54.7 | 54.6 | unchanged, as expected — n=1 never batches |
| 2 | 54.4 | **84.7** | **+56%** |
| 4 | 51.7 | **98.7** | **+91%** |
| 8 | 55.5 | **124.1** | **+124%** |

Accept unchanged at 84 / 78 / 74%, completion byte-identical (sha
`62ff5ed75f0cc2fd`). This changes how many forwards an answer costs, not the
answer.

## Batched propose — ported, correct, DEFAULT-ON

One drafter forward over `n·γ` rows instead of n forwards. Kernels banded per
sequence, paged layer batch-capable, scratch banded, `forward_block` takes an
optional `DflashBatch`, and the scheduler drives it from the batched verify's
Phase B where the n sequences are already in hand.

| C | per-sequence | batched | |
|---|---|---|---|
| 1 | 54.7 | 54.6 | n=1 never batches |
| 2 | 85.0 | 84.2 | |
| 4 | 99.4 | **114.4** | **+15%** |
| 8 | 124.1 | **149.9** | **+21%** |

Acceptance at parity (84/77/74% vs 84/78/73%), completion byte-identical
(sha `62ff5ed75f0cc2fd`). `ATLAS_DFLASH_BATCH_PROPOSE=<width>`: `1`/`0`
restores the per-sequence loop, `N` caps the batch at N.

### Two bugs worth naming

Both were mine from the port, and both were found by bisecting the batch
WIDTH against acceptance rather than by reading code.

**`n_attn` never learned about the batch.** It stayed `eff_ctx + gamma` while
being the row count `batched_embed` takes, so only band 0 was embedded and
every other sequence drafted from stale `stream_buf`. The arithmetic matched
the measurement exactly: one sequence of eight drafting correctly is 24%
acceptance against a 73% baseline.

**The lm_head was called past its tile bound.** `fp8_gemm_t_row_scaled_m16`
is a single-warp `M_TILE=16` kernel. The single-sequence path calls it
unconditionally because γ=8 always fits — but switching those sites to the
batch total made it 32 rows at n=4 and 64 at n=8, past the tile, where it
silently leaves rows it never covered. Row-count selection is restored.

The second one is the instructive one: it does not look like a kernel fault
from outside. The ANSWER stays correct throughout — a bad draft is rejected
and the target's own token is emitted — so nothing errors, output stays
byte-identical, and only the accept rate moves. It reads as "the drafter got
worse", not "a kernel was used out of contract". Correct at 2 bands (16 rows,
inside the tile) and wrong at 4 (32 rows, outside it) is what separated them,
which is why the lever is a numeric width rather than a boolean.

Also fixed: the scheduler passed every eligible sequence to `propose_batch`
regardless of the width the proposer declared via `propose_batch_max`. It now
chunks by that width, which is what makes a declared cap mean anything.

## Verification
`ATLAS_TARGET_MODEL='*'` build clean · 641 spark-model + 104 spark-server tests · LoC cap clean · output byte-identical to #649's own branch under its own config (sha `62ff5ed75f0cc2fd`), which is the cross-check that these fixes change acceptance and not answers.

## Update 2026-08-21: chunk the batched verify to the widest accepted group (44a0ce15)

The batched-verify gate was all-or-nothing: `can_batch_verify` was asked once
with every eligible sequence, so the moment n×(γ+1) rows crossed
`VERIFY_ROW_CAP` (C=16 at γ=8 → 144 rows > 96) the ENTIRE step fell back to
serial per-sequence verifies — the feature turned itself off exactly at the
concurrency it exists for. The gate now walks the group width down until the
cap accepts it and dispatches the batch in chunks of that width; the rest of
the arithmetic (per-sequence verdicts, drafter trims) is untouched.

Measured on GB10 (Qwen3.8-27B-NVFP4 + DFlash2, small profile, 16 slots,
C=16 aggregate streaming decode): 63.5 → 116.8 tok/s with this commit alone;
129.6 with #704's cap raise on top. C=1 output byte-identical. Accepts
unchanged (76–86%).

Note the correctness baseline for all these numbers is #699 — before it,
C≥2 accept/throughput figures were contaminated by the verdict-rewind
degeneration. The merge result of this branch with today's `main` compiles
clean across the workspace and passes all 104 spark-server tests.

### DIFF
--- a/crates/spark-model/examples/batchm_bench.rs
+++ b/crates/spark-model/examples/batchm_bench.rs
@@ -46,7 +46,8 @@ const ITERS: usize = 50;
 /// Weight bytes to cycle through so back-to-back launches miss L2/SLC.
 const COLD_CYCLE_BYTES: usize = 256 << 20;
 const M_SWEEP: &[u32] = &[1, 3, 4, 5, 6, 8];
-const M_MAX: usize = 8;
+const M_WIDE: &[u32] = &[1, 8, 9, 12, 18, 24, 36]; // GEMM-only: batched verify = n seqs × 9 rows
+const M_MAX: usize = 36;
 /// N rows checked against the CPU reference (full N for GPU-GPU bit checks).
 const CPU_CHECK_ROWS: usize = 256;
 
@@ -416,7 +417,7 @@ fn main() -> Result<()> {
         correctness_gate(g, &kernels, a, b0, bs0, c, n, k, &a_host, &b_host, &bs_host)?;
 
         for &(kname, kh, kind) in &kernels {
-            for &m in M_SWEEP {
+            for &m in [M_SWEEP, M_WIDE][matches!(kind, Kind::Gemm | Kind::GemmT) as usize] {
                 if let Kind::Batchm { max_m } = kind
                     && m > max_m
                 {
--- a/crates/spark-model/src/factory/build.rs
+++ b/crates/spark-model/src/factory/build.rs
@@ -678,6 +678,7 @@ pub fn build_model(
                 args.window_size,
                 model.gpu_backend(),
                 max_seq_len,
+                max_batch_size,
             )?;
             model.set_dflash_proposer(std::sync::Arc::new(head));
             tracing::info!("DFlash drafter installed as the active proposer");
--- a/crates/spark-model/src/layers/dflash_head.rs
+++ b/crates/spark-model/src/layers/dflash_head.rs
@@ -385,6 +385,8 @@ pub struct BlockDiffusionDraftHead {
     pub vocab_size: usize,
     pub draft_vocab_size: usize,
     pub gamma: usize,
+    /// Widest cross-sequence batch the scratch bands can hold.
+    pub(super) max_batch: usize,
     pub mask_token_id: u32,
     pub window_size: Option<usize>,
     /// `target_layer_ids`. Same data as `TransformerModel::dflash_capture_layers`,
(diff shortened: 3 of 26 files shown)


tbraun96:

Re-checked today against `main` `9e597531`: all 21 functions/types this PR adds exist there — `step_verify_dflash_batched` at `crates/spark-server/src/scheduler/verify_dflash_batch_step.rs:35`, `try_dflash_capture_all_at` at `crates/spark-model/src/model/impl_b3.rs:460`, `dflash2_select_block` at `crates/spark-model/src/layers/dflash_head/dflash2.rs:193`.

Thanks for the C>=2 correctness fixes — they are serving. No branch deleted.

PR #658, opened 2026-08-21
[DO NOT MERGE] W4A4 native FP4 GEMM — parked at 1.77x with a multi-day continuation plan

tbraun96:
## ⛔ DO NOT MERGE

Parked deliberately. The kernel is **correct and unused** — it is not wired into any dispatch path and changes no default. This PR exists so the work is resumable rather than archaeology.

## Why parked

Enabling it moves the flagship NVFP4 path from **W4A16 → W4A4** (activations bf16 → e2m1). At the current **1.77×** off CUTLASS that buys ~15 ms of cold TTFT (574 → ~559, **−2.6%**) in exchange for a numerics change on a shipping model. That is a poor trade. It becomes worth discussing at **~1.1×**, where the prize is the full **~71 ms**.

## What is already established — do not re-derive

| finding | evidence |
|---|---|
| The gap is **precision, not engineering** | Add-one-in survey of all 8 CUTLASS-selectable paths (35B NVFP4, cold TTFT, n=8/arm, one binary): only `nvfp4-qkvz` (−35.5 ms) and `moe-grouped` (−38.7 ms) carry a gap. `dense-gemm` is a dead heat; **`attn-kv`/`attn-o` we already beat** (+1.3/+1.4 ms); `ssm-out`/`attn-q` are marginal |
| It is **not** missing pipelining | `moe_w4a16_grouped_gemm_ptrtable_t_k64` — the kernel `grouped_silu_down` actually runs — already has `K_STEP_T64=64`, double buffering, 45 `cp.async` uses, `PAD_T64`/`BP_PAD` bank padding, `N_TILE_LG=128`. Checked, not assumed |
| The MMA and its operand/scale layout are proven | **cos = 1.000000** vs the CUTLASS Sm120 collective at every shape, zero CUTLASS dependency at runtime |

## Measured ladder (M=2048 N=12288 K=2048, real qkvz shape, like-for-like incl. activation pack)

| variant | ms/iter | vs CUTLASS | note |
|---|---|---|---|
| naive | 15.254 | 24.8× | ~4.7 GB redundant reads |
| tiled 64×128 | 1.353 | 2.24× | |
| pipelined 64×128 | 1.437 | 2.40× | **no help at large M** |
| 128×128 | 1.117 | 1.87× | |
| 256×128 | 1.173 | 1.83× | **regressed** — 209 regs ⇒ 1 CTA/SM |
| **128×128 + bank padding** | **1.115** | **1.77×** | best |
| CUTLASS | ~0.63 | 1.00× | target |
| *w4a16 (derived)* | *~1.5* | *2.4×* | the bar |

## The plan, in order

**Anti-pattern first:** three consecutive structural fixes each delivered far less than their model predicted. **Do not start with tile-size sweeps** — that ground is covered above.

| day | work |
|---|---|
| **1** | **ncu-diff** ours vs CUTLASS on one shape (stall reasons, occupancy, bank conflicts). Every choice so far came from a *traffic model*, none from a *stall profile*, and the models over-predicted. ⚠️ Profile `examples/w4a4_bench.rs`, **never a live serve** — kernel replay on GB10 unified memory hard-froze the box twice |
| **2** | 3–4 stage async pipeline **together with** a larger tile. Separately they cancel — that is exactly what the 256×128 row shows |
| **3** | **Warp specialisation** (producer/consumer split), the biggest remaining structural difference from the CUTLASS collective |
| **4** | Real XOR swizzle instead of the crude `W4A4_KBP=48` pad |
| **5** | Grouped variant for the MoE path — only **after** dense is within ~1.1×, or it inherits the deficit with more moving parts |

## Gate before any claim

`cos >= 0.999` is **necessary but not sufficient**. The shelved SPLIT=4 GDN spine read **cos 1.0000** and still failed two accuracy gates. Run the 4-minute `ssm-poisoning` tripwire (expect 12/12 byte-identical), then the **full** accuracy suite, before treating any speed number as shippable.

The same plan is written into the header of `kernels/gb10/qwen3.6-35b-a3b/nvfp4/w4a4_gemm.cu`, so it cannot drift from the code if this PR is closed.

### DIFF
--- a/crates/spark-model/examples/w4a4_bench.rs
+++ b/crates/spark-model/examples/w4a4_bench.rs
@@ -1,0 +1,276 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! PHASE-1 MMA PROOF: hand-rolled Sm120 block-scaled FP4 MMA vs the CUTLASS
+//! collective.
+//!
+//! The simplest possible single-tile-looped [M,N,K] GEMM built on a hand-written
+//!   mma.sync.aligned.kind::mxf4nvf4.block_scale.scale_vec::4X.m16n8k64.row.col.f32.e2m1.e2m1.f32.ue4m3
+//! (kernel module `fp4_mma_microtest`, kernels `fp4_microtest_pack` +
+//! `fp4_microtest_mma`). NO cp.async, NO grouping, NO gate+up fusion.
+//!
+//! Oracle: spark_runtime::cutlass::nvfp4_gemm_bf16_act_weight_t — the SAME Sm120
+//! block-scaled FP4 math. MUST agree cos >= 0.999 at M=64, N=1024, K=2048.
+//!
+//! Build (remote GB10):
+//!   cargo build --release -p spark-model --example fp4_mma_microproof \
+//!     --no-default-features --features "cuda gpu-examples"
+//! Run: target/release/examples/fp4_mma_microproof
+
+use anyhow::{Result, bail};
+use spark_runtime::cuda_backend::AtlasCudaBackend;
+use spark_runtime::gpu::{DevicePtr, GpuBackend};
+use spark_runtime::kernel_args::{KernelLaunch, div_ceil};
+
+// ───────────────────────── deterministic PRNG ─────────────────────────
+struct Rng(u64);
+impl Rng {
+    fn next_u64(&mut self) -> u64 {
+        self.0 = self.0.wrapping_add(0x9E37_79B9_7F4A_7C15);
+        let mut z = self.0;
+        z = (z ^ (z >> 30)).wrapping_mul(0xBF58_476D_1CE4_E5B9);
+        z = (z ^ (z >> 27)).wrapping_mul(0x94D0_49BB_1331_11EB);
+        z ^ (z >> 31)
+    }
+    fn unit(&mut self) -> f32 {
+        ((self.next_u64() >> 40) as f32) / ((1u64 << 24) as f32)
+    }
+    fn uniform(&mut self, lo: f32, hi: f32) -> f32 {
+        lo + (hi - lo) * self.unit()
+    }
+}
+
+fn f32_to_bf16_bits(f: f32) -> u16 {
+    let bits = f.to_bits();
+    if (bits & 0x7FFF_FFFF) > 0x7F80_0000 {
+        return ((bits >> 16) | 0x0040) as u16;
+    }
+    let rounding_bias = 0x7FFF + ((bits >> 16) & 1);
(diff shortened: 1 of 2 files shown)


tbraun96:
## Next steps — read this first if you're picking it up

**Status: parked, not abandoned.** Correct kernel, wired into nothing, changes no default. Resume point is **1.77× off CUTLASS**, not zero.

### The one-line reason it's parked
Turning it on moves the flagship from **W4A16 → W4A4**. At 1.77× that's ~15 ms (−2.6%) for a numerics change on a shipping model. Worth revisiting at **~1.1×**, where the prize is the full ~71 ms.

### Do these in order

1. **ncu-diff, before touching anything.** Ours vs CUTLASS, same shape: warp stall reasons, achieved occupancy, smem bank conflicts, `dram__throughput`. Every optimisation so far was picked from a *traffic model* and the models kept over-predicting — three consecutive structural fixes each delivered far less than forecast. Profile `examples/w4a4_bench.rs`. **Do not ncu a live serve on GB10** — kernel replay snapshots the working set, unified memory has no headroom, and it hard-froze this box twice.
2. **Deeper pipeline + larger tile, together.** 3–4 stages. Tried separately they cancel: 2-stage alone *hurt* at large M (3072 CTAs already hide the latency), and 256×128 alone regressed (209 regs ⇒ 1 CTA/SM). The combination is the untested one.
3. **Warp specialisation.** Producer warps on `cp.async`, consumer warps on MMA. This is the largest remaining structural difference from the CUTLASS collective and the likeliest source of the residual.
4. **Real XOR swizzle.** `W4A4_KBP=48` is a crude pad — it fixed the modelled bank conflict and bought ~0 absolute time, which is itself a hint that conflicts weren't binding.
5. **Grouped variant for MoE** — only once dense is within ~1.1×.

### Do not repeat
- Tile-size sweeps. 64×128 / 128×128 / 256×128 are all measured and in the header table.
- "Add pipelining to the MoE kernel." Already there: `moe_w4a16_grouped_gemm_ptrtable_t_k64` has K=64 tiling, double buffering, 45 `cp.async` uses and bank padding.
- Re-deriving the CUTLASS gap. All 8 paths are surveyed; only qkvz and grouped-MoE carry one, and **we already beat `attn-kv` and `attn-o`**.

### Gate before believing any number
`cos >= 0.999` is **necessary but not sufficient** — the shelved SPLIT=4 GDN spine read **cos 1.0000** and still failed two accuracy gates. Run `ssm-poisoning` (4 min, expect 12/12 byte-identical), then the full accuracy suite.

The same plan is in the header of `kernels/gb10/qwen3.6-35b-a3b/nvfp4/w4a4_gemm.cu` so it survives this PR being closed.

PR #661, opened 2026-08-21
vision-fidelity: predict the serve's declared area bound instead of failing on it (9/14 -> 14/14)

rsafier:
Standalone against `main` — no relation to the #649 train, and it touches no crate any of those PRs touch.

**`vision-fidelity` reports FAIL on a perfectly healthy model whenever the target serve caps vision area.** A run against `--vision-max-pixels 262144` scores **9/14 geometry cells**. The same checkpoint on the same build scores 14/14 without the flag.

## Why it fails

The ladder asserts exact vision-token counts computed from each fixture's NATIVE dimensions. A serve that caps vision area does **not refuse** an oversized image — `preprocess_image_with_max_pixels` silently downscales it and answers normally. The reply is a success carrying fewer vision tokens than a native-resolution prediction expects, so the cell scores `Mismatch`.

Exactly five of the fourteen fixtures exceed 262144 px, and the run reports exactly `geometry_matched=9`:

| fixture | px | expected | got |
|---|---|---|---|
| `05_square_768` | 589,824 | 576 | **256** |
| `06_wide_1024x576` | 589,824 | 576 | **252** |
| `07_hd_1280x720` | 921,600 | 920 | **252** |
| `08_portrait_480x854` | 409,920 | 405 | **252** |
| `09_over_clamp_1600x900` | 1,440,000 | 1400 | **252** |

The driver already had a `GeomCell::Unmeasured` kind for exactly this, with a test asserting an over-capacity image must not fail the run — *"a deployment setting, not a defect"*. But it only fires on an error containing `"this encoder holds"`, which is the vision encoder's **pos-embed table** bound. A configured area cap never raises that, so the safety valve could not see it.

## Why this predicts rather than skips

`geometry.rs` used to carry a `within_bound` helper for this, removed on the reasoning that *"the ladder is deliberately built so that every fixture is inside any plausible bound"*. 262144 is a plausible bound an operator really sets, so that premise is now false.

But the removal comment was right about the half that matters: the 1600x900 rung is the entire point of the ladder, and reporting it "unmeasured" would let a regression to the retired 1280px clamp **pass silently**. Skipping trades a false FAIL for a false PASS.

So the bound becomes an **input to the prediction** instead of a reason to skip. Declare it, and the ladder predicts the downscaled geometry and still asserts every rung. Under a 262144 bound the 1600x900 rung's correct answer is 252, while an engine that ignored its own declared bound and fell back to the 1280px clamp still answers 920 — still caught. **No coverage is lost at any bound.**

## Verification — live A/B on one serve, one variable

Same server, same fixtures, the only difference being the new parameter:

```
--param vision_max_pixels=0        (default)  ->  FAIL, 5 mismatches, exit 1
--param vision_max_pixels=262144              ->  PASS, 14/14 matched,  exit 0
```

Both runs report `geometry 14/14 ASSERTED`. The five cells are still being checked in the passing run, not skipped — which is the property this PR exists to preserve.

The live engine's token counts (256, 252, 252, 252, 252) match the mirrored `served_size` arithmetic **exactly** on all five fixtures. That is a stronger check on the mirror than its unit anchors: the benchmark's copy of the engine's resize agrees with the engine itself at every shape the ladder covers.

Serve: `unsloth/Qwen3.8-27B-NVFP4` + DFlash2 drafter on GB10, `--vision-max-pixels 262144 --lm-head-dtype fp8`. Records for both runs are in `~/.atlas/runs/vision-fidelity/` (`08-21 11:45` Fail 9/14, `08-21 11:46` Pass 14/14).

## What changed

* **`geometry.rs`** gains `served_size` — a mirror of `spark-model`'s `target_size_for` — plus `expected_vision_tokens_bounded`. It is a mirror and not a call because `atlas-plugin` drives a server over HTTP and does not link the model crate, so `the_mirror_matches_the_engines_anchors` pins it to the two figures the engine's own source documents: 1600x900 is **1400** tokens unbounded and **920** under the fallback clamp. Both reproduce exactly.

* **`driver.rs`** gains a `vision_max_pixels` parameter (an AREA, like the flag; default 0 = not declared), and every prediction in the file is made under it — calibration, the geometry ladder, and both media-integrity probe deltas. The probe deltas diff two small fixtures no realistic bound moves, but a delta is an asserted token count like any other and has no business being the one prediction here that ignores the serve's configuration.

* The descriptor no longer promises UNMEASURED for images "above the server's encoder capacity" without qualification. It now says what is true: a REFUSED image is unmeasured, and a capped serve downscales rather than refusing.

`vision_max_pixels = 0` is asserted byte-identical to the previous behaviour (`zero_means_nothing_was_declared` walks the whole ladder), so this cannot re-baseline an existing record. Gate mode supplies its own serve recipe and does not pass the flag, so committed gate records are unaffected — which is why every gate-mode record is green while every capped run was red.

## Tests

Six new, all on `main`'s base:

* both engine anchors (1400 unbounded / 920 under the fallback clamp)
* the zero-default equivalence across the whole ladder
* a declared bound moves exactly the five fixtures above it **and no sixth**
* a bound never upscales — the `.min(1.0)` arm, which 8x8 and 64x2048 would otherwise inflate
* the discriminating rung stays discriminating under a bound
* the absolute 4096px long-side ceiling still binds a strip a generous area bound would let through

41 vision tests pass, clippy and fmt clean. Verified on `main`'s base rather than only on the branch it was cut from: `atlas-plugin` and its whole workspace dependency closure (`atlas-governance`, `atlas-closure`) are byte-identical between `avarok/main` and the source branch, so there is no cross-commit symbol dependency for a clean apply to hide.

## Usage

Pass `vision_max_pixels` equal to the serve's `--vision-max-pixels`:

```
spark benchmark run vision-fidelity --url http://127.0.0.1:8888 \
      --model <model> --param vision_max_pixels=262144
```

### DIFF
--- a/crates/atlas-plugin/src/benchmarks/vision/driver.rs
+++ b/crates/atlas-plugin/src/benchmarks/vision/driver.rs
@@ -20,7 +20,7 @@ use crate::params::{ParamKind, ParamSpec, ParamValue, ParamValues};
 use crate::plugin::{Plugin, PluginHandle};
 use crate::result::{BenchmarkResult, LogLine, RunStatus, Stat, Verdict as RunVerdict};
 
-use super::geometry::expected_vision_tokens;
+use super::geometry::expected_vision_tokens_bounded;
 use super::probes::{CONTROL, PROBES, Probe};
 use super::provision::{FIXTURES, provision};
 use super::request;
@@ -44,8 +44,10 @@ pub const DESCRIPTOR: BenchmarkDescriptor = BenchmarkDescriptor {
              changes, and the one a capability check cannot see. CAPABILITY asks unambiguous \
              questions about those images. A no-image CONTROL runs last: if it answers as \
              though it saw a picture, the capability leg proved nothing and the run reports \
-             VACUOUS rather than PASS. Images above the server's encoder capacity report \
-             UNMEASURED, never FAIL — that is a deployment setting, not a defect.",
+             VACUOUS rather than PASS. Images the encoder REFUSES report UNMEASURED, never \
+             FAIL — that is a deployment setting, not a defect. A serve that caps vision \
+             area does not refuse, it downscales: declare its bound with vision_max_pixels \
+             and the ladder predicts the downscaled geometry and still asserts every rung.",
     duration_hint: "~1-2 min",
     updated: "2026-08-14",
     needs_confirmation: false,
@@ -92,6 +94,11 @@ pub struct VisionFidelity {
     integrity: Vec<crate::benchmarks::media_integrity::Cell>,
     max_tokens: usize,
     request_timeout_s: u64,
+    /// The area bound the TARGET serve was started with, as an area in pixels;
+    /// 0 means "not declared". Every geometry prediction is made under this
+    /// bound, because a serve that caps vision area answers normally with a
+    /// DOWNSCALED image rather than refusing it.
+    vision_max_pixels: u64,
 }
 
 impl VisionFidelity {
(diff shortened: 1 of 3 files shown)


tbraun96:

Closing by hand because the commits were **re-cut**, not merged: this PR's head is not an ancestor of main even though its content is, so GitHub will not auto-close it. Verified by artifact rather than assumed — `FALLBACK_MAX_DIM`, `ABS_MAX_DIM` and `served_size` are all on main in the vision-fidelity driver and scorer.

PR #664, opened 2026-08-21
perf(gdn): restore the async double buffer the fused spine dropped — chunk_delta_h −29.8%

tbraun96:
`chunk_delta_h_pipe` is the shipping `vfused` compute plus the `cp.async` double
buffer that core lost when it was written. Opt-in behind `ATLAS_GDN_PIPE=1`.

## Measured

dgx1, **one binary**, both arms back-to-back, n=8 each, `ATLAS_GDN_PIPE` the only
variable, `nvidia/Qwen3.6-35B-A3B-NVFP4`:

| phase | `vfused` | `_pipe` | Δ |
|---|---:|---:|---:|
| **`gdn_fla_chunk_delta_h`** | 47.3 ms | **33.2 ms** | **−29.8%** |
| `gdn_prefill` (spine total) | 120.9 ms | 105.7 ms | −12.6% |
| **cold TTFT median** | 548.30 ms | **537.18 ms** | −2.0% |
| `gdn_fla_recompute_wu` | 41.1 | 41.2 | flat |
| `gdn_fla_chunk_fwd_o` | 31.8 | 31.1 | flat |
| `qkvz_gemm` / `out_proj` / `conv1d` | 53.8 / 16.0 / 8.8 | 53.8 / 16.0 / 8.9 | flat |

The four flat phases don't read the staged tiles — they're the internal control.
`moe_ffn` read 197.1 vs 164.6 across arms and 163.6 on an earlier run of the same
build: it is **noisy run-to-run and is not counted either way**, which is why the
end-to-end number is softer than the kernel number.

## Why it was slow

The fused core loads each chunk with plain **scalar** copies (`Wp[e] = Wsrc[e]`)
behind a bare `__syncthreads()` — routing every byte global→register→shared, and
leaving the load of chunk `c+1` strictly after the compute of chunk `c` on a spine
that is **serial in c**. The original `chunk_delta_h` already had `cdh_prefetch`, a
16-byte `cp.async` double buffer; it just wasn't carried over. This reuses that
helper rather than growing a second one.

Also fixed in that helper, so both kernels benefit: the decay table ran up to
`CHUNK` serial `expf` calls **on thread 0** while 255 threads idled at the next
barrier, once per chunk. Now one thread per entry.

## A false comment, corrected

```c
// GB10 sm_121 has cp.async.cg (NO TMA).
```
This is wrong, and it plausibly steered years of work on this kernel.
`cp.async.bulk.tensor` + `mbarrier` **compile for `sm_121a` with our own CUDA 13.0
nvcc**, and the FlashQLA GDN spine measured on this box at **15.0 ms** uses exactly
that. Our `kernels/` tree contains **zero** TMA and **zero** mbarrier (verified by
grep across the tree).

That is the remaining gap, and it is not a tile-shape problem:

| GDN spine | time |
|---|---:|
| PR #583's FLA baseline | 184.8 ms |
| `vfused` (main today) | 120.9 ms |
| **`_pipe` (this PR)** | **105.7 ms** |
| **FlashQLA (TMA + mbarrier)** | **15.0 ms** |

**7.0× still to go.** Follow-up work is the fused TMA spine.

## Safety

SPLIT stays **2 / 256 threads**. SPLIT=4 measures faster and scores **1/12** on the
ssm-poisoning tripwire; this PR doesn't touch thread geometry, so that verdict
carries. `ssm-state-poisoning-gate`: **12/12 replays byte-identical, twice.**

Per the `vfused` note, cos ≥ 0.99 *and* byte-identical single-prompt greedy both
missed SPLIT=4's 1.4-point BFCL drift — so the tripwire, not the speedup, is what
qualifies this.

## Observability

`init.rs` now logs `GDN state spine: <name>` at kernel resolution. Which spine ran
is the most consequential fact about a GDN measurement, and run records carry
`target_url`/`target_model`/hardware but nothing about the kernel — so an A/B on
this path was unfalsifiable from artifacts alone.

254 registers, 0 spills, 1 CTA/SM — occupancy unchanged, only the overlap is new.

### DIFF
--- a/crates/spark-model/src/layers/ops/ssm_gdn_a3.rs
+++ b/crates/spark-model/src/layers/ops/ssm_gdn_a3.rs
@@ -178,10 +178,19 @@ pub fn gdn_prefill_fla(
     // split the DV axis, so grid.y stays `batch_size`. smem is identical for both
     // members — only the thread count differs, and it must match the kernel that
     // init.rs actually loaded for the same env value.
-    let smem_fused = C * kd * 2 + C * kd * 2 + C * vd * 2 + (C + 1) * 4;
+    // `..._pipe` double-buffers {W,K,U} through `cp.async`, so it needs the SAME
+    // footprint the original (also double-buffered) spine uses — `smem_dh`. Under-
+    // sizing this reads the second slot out of bounds, so the selector has to agree
+    // with the kernel `init.rs` loaded for the same env value.
+    let pipe = std::env::var("ATLAS_GDN_PIPE").ok().as_deref() == Some("1");
+    let smem_fused = if pipe {
+        smem_dh
+    } else {
+        C * kd * 2 + C * kd * 2 + C * vd * 2 + (C + 1) * 4
+    };
     let fused_block = match std::env::var("ATLAS_GDN_VTILE").ok().as_deref() {
-        Some("1") => 512u32, // SPLIT=4 build
-        _ => 256u32,         // SPLIT=2 build (default)
+        Some("1") if !pipe => 512u32, // SPLIT=4 build
+        _ => 256u32,                  // SPLIT=2 build (default, and the pipe build)
     };
     let (k_cdh, cdh_grid_y, cdh_smem, cdh_block) = if use_fused {
         (k_chunk_delta_h_fused, batch_size, smem_fused, fused_block)
--- a/crates/spark-model/src/layers/qwen3_ssm/init.rs
+++ b/crates/spark-model/src/layers/qwen3_ssm/init.rs
@@ -241,7 +241,6 @@ impl Qwen3SsmLayer {
                     tracing::info!("GDN state spine: {name}");
                     name
                 },
-
             ),
             gdn_prefill_fla_chunk_fwd_o_k: super::super::try_kernel(
                 gpu,
--- a/crates/spark-model/src/layers/qwen3_ssm/init.rs
+++ b/crates/spark-model/src/layers/qwen3_ssm/init.rs
@@ -224,10 +224,24 @@ impl Qwen3SsmLayer {
             gdn_prefill_fla_chunk_delta_h_fused_k: super::super::try_kernel(
                 gpu,
                 "gated_delta_rule_fla",
-                match std::env::var("ATLAS_GDN_VTILE").ok().as_deref() {
-                    Some("1") => "gated_delta_rule_chunk_delta_h_vtile",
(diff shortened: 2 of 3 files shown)


tbraun96:
## Correction: withdrawing the end-to-end number

The PR body claims cold TTFT **548.30 → 537.18 ms (−2.0%)**. **That claim does not hold and should be ignored when reviewing.**

A later run measured `vfused` — the same arm, same binary, same flags — at **537.30 ms**, 11 ms below its own earlier 548.30. Cross-restart variance is therefore ~2%, the same magnitude as the delta I attributed to this change. Two further interleaved rounds put `_pipe` at **535.84 ms** and an unrelated arm at **535.42 ms**, i.e. all within the same band.

| Claim | Status | Basis |
|---|---|---|
| `gdn_fla_chunk_delta_h` 47.3 → 33.2 ms (**−29.8%**) | **stands** | direct phase timers; the four phases that don't read the staged tiles were flat as controls |
| `gdn_prefill` 120.9 → 105.7 ms (−12.6%) | **stands** | same timers |
| cold TTFT −2.0% end-to-end | **WITHDRAWN** | inside cross-restart variance |
| ssm-poisoning 12/12 byte-identical ×2 | stands | |

I should have caught this when I noted `moe_ffn` swinging 197.1 / 164.6 / 163.6 across runs of the same build. I set that phase aside as noisy without asking whether the end-to-end total inherited the same noise. It did.

**What this PR is, restated honestly:** a **−29.8% kernel-level improvement** to `chunk_delta_h`, restoring an async double buffer the fused spine had dropped, with the tripwire green. Whether that is visible in end-to-end TTFT is **not yet determined** and would need interleaved repeats with a stated variance band to assert either way.

The comparison against the reference is unaffected — it rests on phase timers, not on end-to-end deltas: `gdn_prefill` **105.7 ms** here vs FlashQLA's **15.0 ms**, still **7.0×**.

tbraun96:
Closing: already on `main`. The `_pipe` spine landed through #721 (`1dc3c0dcfb`, "perf(gemma-4, gdn): tensor-core 512 attention, FFN transposes, GDN double buffer, TMA plumbing").

On `main` `9e597531`: `gated_delta_rule_chunk_delta_h_pipe` at `kernels/gb10/common/gated_delta_rule_fla.cu:1253`, and the lever read at `crates/spark-model/src/layers/ops/ssm_gdn_a3.rs:266` (spelled `AVAROK_GDN_PIPE` since the #1101 rebrand). What remains in this branch's diff against `main` is only the pre-rebrand env names. No branch deleted.

PR #665, opened 2026-08-21
runtime: carry a CUtensorMap through the launch path (and stop truncating wide args)

tbraun96:
Prerequisite plumbing for TMA (`cp.async.bulk.tensor`) in our own kernels. **No
kernel uses this yet** — landed on its own so the two bugs below are reviewable
apart from a kernel rewrite.

## Two silent-corruption bugs on the launch path

Both are currently unreachable, and both become reachable the moment anything
passes a struct by value.

**1. `launch_typed` truncated wide args.**
```rust
KernelArg::Bytes(b) => {
    let n = b.len().min(8);      // anything wider: dropped, no error
```
A 128-byte `CUtensorMap` would have reached the kernel as 8 bytes of garbage with
no diagnostic anywhere. An arg now occupies `ceil(len/8)` **consecutive** slots
and contributes **one** param entry.

**2. `KernelLaunch::launch` indexed `storage` by the arg's position in `kinds`.**
Correct only while every arg is exactly one slot — a 16-slot arg makes every
argument *after* it read the wrong slot. `ArgKind` now records its starting slot.

The packing moved into a free `gpu::pack_kernel_args` so it can be tested at all:
through the mock backend **both the broken and the fixed version return `Ok`**, so
the tests assert the 128 bytes round-trip and that the arg following a tensormap
is intact — not that the call succeeded.

## New

- `KernelLaunch::arg_tensormap(&[u8; 128])`
- `cuda_backend::tensormap::TensorMap` over `cuTensorMapEncodeTiled`

`tiled_2d_bf16` takes row-major rows/cols/stride in **elements** and applies the
driver's fastest-varying-**first** ordering internally — reversed, that does not
error, it silently transposes the load. Guards reject a misaligned address, a
misaligned stride, and a stride narrower than the row up front; the driver's
alternative is a generic invalid-argument or a mis-addressed copy.

> ★ `CU_TENSOR_MAP_DATA_TYPE_BFLOAT16 = 9`, read from
> `/usr/local/cuda/include/cuda.h` — **not 10**. `FLOAT64` sits at 8 ahead of it and
> `FLOAT32_FTZ` is 10. A wrong dtype does not fail, it reinterprets the bytes.

## Why

`gated_delta_rule_fla.cu` claimed *"GB10 sm_121 has cp.async.cg (NO TMA)"*. That is
false — TMA + mbarrier compile for `sm_121a` under CUDA 13.0, and the FlashQLA GDN
spine measured on a GB10 at **15.0 ms** uses exactly that, against our **105.7 ms**.
Our `kernels/` tree contains zero TMA and zero mbarrier. That is the 7.0× gap.

9 tests, `cargo fmt` clean, `cargo clippy --tests` clean.


tbraun96:
Closing: the TMA plumbing landed with #721.

**Carrier:** `1dc3c0dcf` — *"perf(gemma-4, gdn): tensor-core 512 attention, FFN transposes, GDN double buffer, **TMA plumbing** (#721)"*. That commit is the one that **added** `crates/spark-runtime/src/cuda_backend/tensormap.rs`, the file this PR introduces:

```
$ git log --diff-filter=A origin/main -- crates/spark-runtime/src/cuda_backend/tensormap.rs
1dc3c0dcf perf(gemma-4, gdn): … TMA plumbing (#721)
```

Its content is on `main` verbatim, including this PR's own module docs — the `CUtensorMap` header, the "moves address generation, tiling and BOUNDS CHECK into hardware" paragraph, and the zero-fill note. A sample of ten distinctive added lines found ten on `main`.

That is also why it shows as CONFLICTING on two files: it re-adds what is already there, against a tree where the surrounding code has since moved.

Nothing is lost by closing, and reopening restores it if this reading is wrong.

tbraun96:
Correction to my own close comment, because "verbatim" overstated it and the residue is the interesting part.

A full-population comparison (not the ten-line sample I used) finds **11 of 187 added lines absent from `main`** — about 5%. They are all one thing: the link-time binding and its doc block.

```rust
pub(super) fn cuTensorMapEncodeTiled(     // this PR
use super::cuTensorMapEncodeTiled;
```

`main` has the symbol, but resolves it **at runtime** instead:

```rust
// tensormap.rs:83
let name = c"cuTensorMapEncodeTiled";
// tensormap.rs:159-163
// ★ RESOLVED AT RUNTIME, NOT LINKED. … a link-time `extern "C"` makes the whole
//   rust-lld: error: undefined symbol: cuTensorMapEncodeTiled
```

So the part of this PR that is *not* on `main` is the part `main` deliberately replaced — a link-time `extern "C"` that died at link (`LNK2019 unresolved external`). The close stands, and for a stronger reason than I first gave: the work landed with #721, and the residue is the approach that was fixed, not work still owed.

Recording the method error too, since it nearly cost more: a ten-line sample taken from the head of a sorted diff reported 10/10 contained. The honest instrument is the full added-line population plus a look at what remains — on that basis #495, #544 and #664 are **11%, 67% and 9% absent** and are genuinely live, which the same sample had flagged as "likely already landed".

PR #666, opened 2026-08-21
perf(gdn): TMA build of the state spine — opt-in, verified correct, NOT faster

tbraun96:
`chunk_delta_h_tma` is `_pipe`'s compute with staging moved from `cp.async` to
**TMA + `mbarrier`**, behind `ATLAS_GDN_TMA=1`.

**Stacked on #664** (the pipe spine) and **#665** (the runtime plumbing that lets a
128-byte `CUtensorMap` reach a kernel at all). Review those first.

## It is not faster, and that is in the title on purpose

| | `_pipe` | `_tma` |
|---|---:|---:|
| cold TTFT median, n=8 | 535.84 ms | **535.42 ms** — flat, inside noise |
| spine, **process-cold request only** | 107.3 ms | 244.6 ms |

The two instruments disagree because `ATLAS_PROFILE_FIRST` is a **one-shot** flag
consumed by the first prefill in the process. The phase timers therefore describe a
*process-cold* request, where TMA pays a large one-time first-use cost; the
8-request median (warm process, uncached prefix) shows it neutral. The series
carries **no outlier at all** (min 530.98 / max 538.39), which is what rules out a
steady-state +137 ms. Both readings are correct about different requests.

Landing it as **verified infrastructure**: the fused warp-specialised spine that
*is* the 7× needs TMA transport, and this is its bring-up — proven to compile, to
run, and to be numerically exact.

## Two of my hypotheses died here — recorded so nobody re-runs them

- *"Per-call `cuTensorMapEncodeTiled` explains the +137 ms."* **Refuted by direct
  measurement**: 90 encodes — a whole prefill's worth — take **0.0019 ms**, ~21 ns
  each. A descriptor cache would have optimised a cost that does not exist.
- *"The env was set but the path silently fell back."* **Wrong** — 450 launches. I
  had grepped `head -1` and read `init.rs`'s resolution line rather than the
  launch-time line.

## What the copy engine takes over

- address generation and tiling for W, U and K;
- **the K gather.** `cdh_prefetch` walks K row by row (`i < ce` predicated, 16
  `cp_async16` per row) because K sits at a `qk_stride` pitch with a `kh*k_dim`
  column origin. That is exactly a strided 2-D tile — **one** `tma_load_2d`
  replaces the loop;
- **the ragged tail** — an overhanging tile is zero-filled in hardware, the same
  value the predicated path wrote by hand.

One `expect_tx` per stage covers all three tiles, so the consumer waits once rather
than per tensor. Stage `p` is reused every second chunk, hence parity `(c>>1)&1`.

## Correctness

| Checkpoint | Test | Result |
|---|---|---|
| `nvidia/Qwen3.6-35B-A3B-NVFP4` | 5 temp-0 prompts vs `_pipe` | **byte-identical** |
| `Qwen/Qwen3.6-35B-A3B-FP8` | `ssm-state-poisoning-gate` | **12/12 replays byte-identical** |

Both matter: a transposed or mis-strided descriptor yields plausible text, not a
crash. Engagement was verified by log on both checkpoints, not assumed.

## Every precondition is checked, and named on rejection

TMA reports **no error** for a well-formed descriptor pointed at the wrong shape, so
the launcher refuses varlen (`choff` then comes from `cu_chunks`; the flat row count
is unknown host-side), any head narrower than the compile-time tile, and a
`qk_stride` not a multiple of 8. With a guard unmet, `ATLAS_GDN_TMA=1` **warns with
the reason** instead of quietly running the old kernel.

PR #296 shipped that failure mode once — an ldmatrix GEMM falling back silently
while both gates stayed green — and this path reproduced it during bring-up: an A/B
ran with the env set, fell back, and the two arms differed by 0.13%.

234 registers, 0 spills.


tbraun96:
Closing: already on `main`. The TMA build landed through #721 (`1dc3c0dcfb`, "perf(gemma-4, gdn): tensor-core 512 attention, FFN transposes, GDN double buffer, TMA plumbing"), together with #665's runtime plumbing.

On `main` `9e597531`: `gated_delta_rule_chunk_delta_h_tma` at `kernels/gb10/common/gated_delta_rule_fla.cu:1443`, the opt-in read at `crates/spark-model/src/layers/ops/ssm_gdn_a3.rs:336` (`AVAROK_GDN_TMA`), and the tensormap contract tests at `crates/spark-runtime/src/cuda_backend/tensormap.rs:209`. All 15 functions and 20 files this PR adds are present. The negative result in the title still stands and is recorded in the code. No branch deleted.

Issue #435, opened 2026-08-09
Speculative decoding is not output-equivalent at temperature 0: verify runs different arithmetic than sequential decode


tbraun96:
## The invariant Atlas violates

Speculative decoding is output-equivalent **by construction**: the drafter proposes, the target verifies, and at temperature 0 the emitted token sequence must be bit-identical to plain greedy decoding. Speed changes; output does not.

Atlas violates this. Measured 2026-07-22 (3-arm probe, `GATE_FORCE=1`, 27B K=2 + 35B K=1): NOSPEC greedy differs from **both** spec arms, first-divergence margins are structural (~4.1 nats — confidently different tokens, not tie-flips), diverging around token 250.

This has been treated as a property to work around. It is a defect.

## Not the cause (search space closed)

- **Accept criterion is correct.** `scheduler/verify_k4_step.rs:117-125` — `drafts[i] == argmax(target_logits[i])`, with `verify_pipeline_helper.rs` replaying `decode_logits_seq.rs` semantics verbatim.
- **Restore plumbing is EXACT.** Pre-verify checkpoint is a verbatim D2D snapshot (`verify_a.rs:266-311`); full reject restores it verbatim (`:368-377`); partial accept restores `intermediates[na]` verbatim (`async_chkpt.rs:191-286`) with index invariants hard-checked. The copies are lossless. *(An earlier note describing this as "non-bit-exact state restore" was wrong.)*
- **Drafter sharing is read-only**, and its contract holds — `mtp_head/forward_batch.rs:491-494` correctly reasons that drafter logits may differ bitwise "because drafts are verified by the main head, so accepted output is unaffected; only the accept rate can move."

## The cause

Under spec-ON, **every token** is consumed through `decode_verify_graphed_k4` — even the previous step's correction token re-enters as verify row 0. That path is a numerically different program from sequential decode, and its outputs are written into the **persistent SSM recurrence and KV cache**, where they compound.

Ranked by contribution:

[...]

PR #667, opened 2026-08-21
docs(exact-verify): scope, cost, and what the flag does NOT fix (post-#699)

rsafier:
Documentation only — no behaviour change, no default flipped.

`--exact-verify`'s doc already explains the **mechanism** well: every FFN and attention projection is selected by ROW COUNT, verify K=4 takes `w4a16_gemv_batch4` where decode takes `w4a16_gemv`, ~5e-5 of lanes differ by 1 ULP, enough to flip a thin top-2 argmax (#459). What it does not say is what that looks like from outside the engine, and the gap is wide enough to cost a debugging session — "an occasional flipped argmax" reads like an occasional different-but-fine token.

## What the default actually does

Measured 2026-08-21, `unsloth/Qwen3.8-27B-NVFP4` + `incoai/Qwen3.8-27B-DFlash2` on GB10, one binary, this flag the only variable:

| probe | default | `--exact-verify` |
|---|---|---|
| "count from 1 to 10" | `1, 2, 100, 100, 100, ...` | `1, 2, 3, 4, 5, 6, 7, 8, 9, 10` |
| "list four colors" | `Red, Blue, Green, 1999; Green, 2000; ...` | `Red, Blue, Green, Blue` |
| "two sentences about blue" | `...100% success rate.\n107 \| ### 3.3.2.2.1` | coherent |
| `video-fidelity` C=1 / C=2 / C=4 | 1/1, **0/2**, **0/4** | **1/1, 2/2, 4/4** |

A flipped token does not stay one wrong token. It lands the sequence in a repetition attractor, and **the drafter then agrees with it** — verify steps go 7/7 accepted — so the loop is locked in and the reply degenerates into repetition or unrelated prose until the token cap.

That coupling is the part worth writing down: the symptom presents as a broken drafter while acceptance telemetry looks perfectly healthy at 29-71%. The accepts are honest. The logits they are checking against are simply not the ones sequential decode would have produced.

## Two traps, both of which cost time here

**A short reply hides it entirely.** A 2-token "Paris" answer is correct on every arm. A smoke test that only asks small questions will not see this at all. The fastest repro is a prompt whose correct answer is a short ordered list.

**The residual with the flag ON is more than one token.** The same request at temperature 0 answered with 45 tokens once and 50 the next. The accepted-row COUNT varies with runtime scheduling, so the row-count-selected projection kernel varies with it, so the rounding does. Reproducibility needs the single-row routing the doc already describes as future work — this flag alone does not deliver it.

## Provenance of the numbers

Localized by first divergence rather than by inspection. On "count from 1 to 10" (31 prompt tokens) the first wrong token is completion index 6 (`'1'` id 16, where serial decode gives `'3'` id 18). The verify log maps it exactly, `seq_len` being the length after each step:

```
step 1  accepted=1/7  ends 33   -> completion idx 0-1
step 2  accepted=0/7  ends 34   -> idx 2
step 3  accepted=0/7  ends 35   -> idx 3
step 4  accepted=7/7  ends 43   -> idx 4-11
```

Index 6 falls inside step 4's accepted-draft run (4-10 are drafts, 11 is the bonus token), so the wrong token was **accepted** — which is only possible if the target's own verify logits chose it.

Excluded by one-variable measurement before landing on the GDN chain: batched propose (`ATLAS_DFLASH_BATCH_PROPOSE=1`), LM-head dtype (`fp8` and `default` alike), prefix caching (byte-identical on/off), multimodality (text-only degenerates on its own), `ATLAS_DFLASH_UNIFIED_CTX=0`, and the GEMV tier — `w4a16_batch_bitparity_microtest` PASSES on this box at every M=2..16 and every projection shape, with its 1-ULP control firing.

It is also not new and not specific to any branch in flight: it reproduces identically on `pull/648/head` (`acaf9533`), on `pull/649/head`, and on a local DFlash stack, each served under its own author's config, while spec-off is clean on all three.

## The cost is not what the doc implies — measured, it is a WIN here

The doc's ~+22-36% is a **per-decode-step** cost, measured at the n=8/K=4, n=16/K=2 and n=32/K=2 rungs. End-to-end throughput also depends on how many drafts get accepted per step, and that is exactly what the corruption damages. At C=1 on this config:

| | default | `--exact-verify` |
|---|---|---|
| decode | 27.29 tok/s | **32.45 tok/s** (+18.9%) |
| mean acceptance | 35.0% (103 steps) | **45.5%** (236 steps) |

Run-to-run spread was tight (27.27/28.29/27.29 vs 32.43/32.45/32.47, n=3 after a warm request).

The mechanism is the same defect seen from the other side: corrupted verify logits **reject correct drafts**. The drafter proposes roughly what the true model would say; a verifier working from wrong logits disagrees with it, so acceptance falls. Make the logits right and the drafts start matching again — the higher per-step cost is more than repaid.

Scope, so this is not over-read: one prompt, C=1, n=3, on this hardware and checkpoint. It does not contradict the per-step figures in the doc; it says the step cost is not the whole story, and that on a config where the divergence actually bites, exactness is not a tradeoff at all.

## What this PR deliberately does not do

It does not make exactness the default. That remains a policy call for the DFlash owners — the numbers above are C=1 and the documented step costs are measured at the concurrency rungs, so the crossover needs its own sweep, and the existing survey of vLLM / SGLang / TensorRT-LLM / TGI defaults in that same doc belongs in that decision rather than in a docs commit.

`ATLAS_TARGET_MODEL='*'` build clean, fmt clean, `--help` renders.

## Re-scope 2026-08-21 (d2f1b0d8): the correctness framing above is superseded by #699

Everything above that presents `--exact-verify` as fixing gross degeneration
(the count-to-10 collapse, video-fidelity 0/2 and 0/4 at C=2/C=4) turned out
to be a misattribution. #699 root-caused those repros to the K=4 verdict
rewinding sequences by their pending-draft count instead of the forward's row
count — erasing committed output. This flag never fixed that bug; it changed
the gate's dispatch so the bug stopped firing. With #699 applied, every one
of those repros passes with the flag OFF.

The final commit rewrites both doc blocks accordingly: the flag closes the
#435/#459 1-ULP / occasional-argmax divergence, it is not a correctness
switch, and an apparent correctness fix from flipping it should be treated as
a dispatch-sensitivity symptom pointing at a real scheduler bug. What stands
unchanged: the measured cost table (-7% to -31%), the withdrawal of the
earlier +18.9% throughput claim, and the reproducibility residual (row-count-
selected kernels vary with scheduling even with the flag on).

Suggested merge order: after #699, which this doc now references.

### DIFF
--- a/crates/spark-model/src/layers/qwen3_ssm/gdn_flags.rs
+++ b/crates/spark-model/src/layers/qwen3_ssm/gdn_flags.rs
@@ -60,14 +60,14 @@ pub struct GdnFlags {
     /// shape measured (#459). Closing that needs single-row routing for the
     /// whole verify forward, which is future work.
     ///
-    /// ★ Do not read "default off" as "default harmless". The divergence this
-    /// closes is not a cosmetic one-token difference at temperature 0: the
-    /// flipped token lands in a repetition attractor that the drafter then
-    /// AGREES with (7/7 accepted), so replies collapse into repetition or
-    /// unrelated text. Measured 2026-08-21 on qwen3.8-27B-NVFP4 + DFlash2,
-    /// `video-fidelity` went 0/2 and 0/4 correct at C=2/C=4 by default and
-    /// 2/2, 4/4 with this flag on. The operator-facing consequences and a
-    /// repro live on `ServeArgs::exact_verify`.
+    /// ★ Attribution warning, learned the hard way: a 2026-08-21 measurement
+    /// showed gross output degeneration (video-fidelity 0/2, 0/4 at C=2/C=4)
+    /// that this flag appeared to fix. The real cause was the K=4 verdict
+    /// rewind bug (#699); this flag only changed dispatch so the bug stopped
+    /// firing. The 1-ULP divergence this flag actually closes has never been
+    /// shown to cause more than an occasional flipped token at temperature 0.
+    /// If flipping this flag changes gross behavior, suspect a dispatch-
+    /// sensitive scheduler bug first. Details on `ServeArgs::exact_verify`.
     pub exact_verify: bool,
 }
 
--- a/crates/spark-model/src/layers/qwen3_ssm/gdn_flags.rs
+++ b/crates/spark-model/src/layers/qwen3_ssm/gdn_flags.rs
@@ -59,6 +59,15 @@ pub struct GdnFlags {
     /// implementations round differently — ~5e-5 of lanes by 1 ULP, on every
     /// shape measured (#459). Closing that needs single-row routing for the
     /// whole verify forward, which is future work.
+    ///
+    /// ★ Do not read "default off" as "default harmless". The divergence this
+    /// closes is not a cosmetic one-token difference at temperature 0: the
+    /// flipped token lands in a repetition attractor that the drafter then
+    /// AGREES with (7/7 accepted), so replies collapse into repetition or
+    /// unrelated text. Measured 2026-08-21 on qwen3.8-27B-NVFP4 + DFlash2,
+    /// `video-fidelity` went 0/2 and 0/4 correct at C=2/C=4 by default and
+    /// 2/2, 4/4 with this flag on. The operator-facing consequences and a
+    /// repro live on `ServeArgs::exact_verify`.
     pub exact_verify: bool,
 }
 
--- a/crates/spark-server/src/cli/serve_args.rs
+++ b/crates/spark-server/src/cli/serve_args.rs
@@ -318,6 +318,18 @@ pub struct ServeArgs {
     /// so a smoke test that only asks small questions will not see this.
     /// Fastest repro is a prompt whose correct answer is a short ordered list.
     ///
+    /// ★ AND THE COST CAN BE NEGATIVE. The ~+22-36% below is a per-decode-STEP
+    /// figure; end-to-end throughput also depends on how many drafts survive
+    /// verification, which is precisely what the divergence damages. Corrupted
+    /// verify logits REJECT CORRECT DRAFTS — the drafter proposes roughly what
+    /// the true model would say and a verifier reading wrong logits disagrees.
+    /// Measured at C=1 on the config above (one prompt, n=3 after a warm
+    /// request): 27.29 tok/s and 35.0% mean acceptance by default, versus
+    /// 32.45 tok/s and 45.5% acceptance with this flag on — +18.9% throughput
+    /// for turning exactness ON. Where the divergence actually bites, this is
+    /// not a tradeoff. The crossover against the concurrency rungs, where the
+    /// step costs below were measured, still needs its own sweep.
+    ///
     /// SCOPE, and it is narrower than this flag once claimed: it makes the
     /// GDN/SSM verify chain exact. It does NOT make speculative output
     /// bitwise-equal to non-speculative output end to end, and setting it
--- a/crates/spark-server/src/cli/serve_args.rs
+++ b/crates/spark-server/src/cli/serve_args.rs
@@ -297,10 +297,35 @@ pub struct ServeArgs {
 
     /// Sequential-decode-exact GDN/SSM verify chain — OPT-IN (default: off).
     ///
+    /// ★ WHAT THE DEFAULT COSTS IN PRACTICE, because "a flipped argmax" reads
+    /// milder than it behaves (measured 2026-08-21, qwen3.8-27B-NVFP4 + DFlash2
+    /// on GB10, one binary, this flag the only variable):
+    ///
+    ///     prompt "count from 1 to 10"   default: `1, 2, 100, 100, 100, ...`
+    ///                                   exact:   `1, 2, 3, 4, 5, ..., 10`
+    ///     video-fidelity C=1/C=2/C=4    default: 1/1, 0/2, 0/4 correct
+    ///                                   exact:   1/1, 2/2, 4/4 correct
+    ///
+    /// A flipped token does not stay one wrong token. It lands the sequence in
+    /// a repetition attractor, and the DRAFTER THEN AGREES with it — verify
+    /// steps go 7/7 accepted — so the loop is locked in and the reply
+    /// degenerates into repetition or unrelated text until the token cap. This
+    /// is why the symptom looks like a broken drafter while acceptance
+    /// telemetry looks healthy: the accepts are honest, the logits they check
+    /// against are not the ones sequential decode would have produced.
+    ///
+    /// A short reply can hide it entirely — a 2-token "Paris" answer survives —
+    /// so a smoke test that only asks small questions will not see this.
+    /// Fastest repro is a prompt whose correct answer is a short ordered list.
+    ///
     /// SCOPE, and it is narrower than this flag once claimed: it makes the
     /// GDN/SSM verify chain exact. It does NOT make speculative output
     /// bitwise-equal to non-speculative output end to end, and setting it
-    /// will not give you a reproducible spec-on serve.
+    /// will not give you a reproducible spec-on serve. Measured with the flag
+    /// ON, the residual is still visible: an occasional single wrong token,
+    /// and the same request at temperature 0 answering with 45 tokens once and
+    /// 50 the next time — because the accepted-row COUNT varies with runtime
+    /// scheduling, so the row-count-selected projection kernel varies with it.
     ///
     /// Why not (measured on GB10, issue #459): every FFN and attention
     /// projection is computed by a kernel selected on ROW COUNT. A token
--- a/crates/spark-server/src/cli/serve_args.rs
+++ b/crates/spark-server/src/cli/serve_args.rs
@@ -297,26 +297,24 @@ pub struct ServeArgs {
 
     /// Sequential-decode-exact GDN/SSM verify chain — OPT-IN (default: off).
     ///
-    /// ★ WHAT THE DEFAULT COSTS IN PRACTICE, because "a flipped argmax" reads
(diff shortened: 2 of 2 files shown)


rsafier:
Followed up on the open question in the PR body — whether the C=1 throughput win survives at the concurrency rungs where the ~+22-36% step costs were measured.

**It does. There is no crossover: `--exact-verify` is faster at every level tested.**

Two runs per arm, decode phase timed on its own via streaming (the arms produce different-length replies, so prefill must not leak into the rate). Same box, same binary, same prompt, this flag the only variable.

| C | default tok/s | `--exact-verify` tok/s | speedup | default accept | exact accept |
|---:|---|---|---:|---:|---:|
| 1 | 23.55 / 23.43 | 30.64 / 30.54 | **+30%** | 26.9% | 41.5% |
| 2 | 24.57 / 24.44 | 46.55 / 46.66 | **+90%** | 22.7% | 44.4% |
| 4 | 47.49 / 53.65 | 73.39 / 78.69 | **+50%** | 39.5% / 36.2% | 56.5% / 60.1% |
| 8 | 65.32 / 70.35 | 85.03 / 89.91 | **+29%** | 39.7% / 35.6% | 51.2% / 47.0% |

Acceptance is higher at every level, which is the whole mechanism: corrupted verify logits reject correct drafts, so the default arm pays the drafter's cost and throws away more of its output.

**The C=2 cell is the striking one and it reproduced exactly** (24.57 then 24.44): by default, going from one stream to two buys essentially nothing — 23.5 -> 24.5 tok/s aggregate — while acceptance drops to its worst value anywhere in the grid, 22.7%, on the largest number of verify steps (104). With exactness on, the same step is 30.6 -> 46.6. Whatever the divergence does, it does it worst at C=2.

Scope, so this is not over-read: one prompt, one checkpoint (`qwen3.8-27B-NVFP4` + DFlash2), one box, n=2 per cell. It does not refute the per-step figures in the doc — those are a different measurement on different rungs. It says the step cost is not what determines end-to-end throughput once the divergence is actually biting, and on this config the net is positive everywhere.

This PR still does not change the default; that call belongs with the DFlash owners. But the "exactness costs 22-36%" framing is now the wrong way round for at least this configuration family, and I would suggest the default question gets reopened on the strength of it.

rsafier:
**Correction and a much better measurement.** My earlier comment claimed "no crossover, faster at every level". That was measured with an ad-hoc client-side script on a *prose* prompt, and two things were wrong with it: the concurrency cells were noisy (±15% run to run — C=4 gave 80.2 then 94.1), and prose is a low-acceptance workload that flatters exactness. On a code prompt at C=1 the default is in fact marginally *faster* in raw rate. Please read the numbers below instead of the ones above.

Re-measured with the repo's own pinned harness (`decode-floor`: 3 fixed runs of one code prompt, **server-side** decode rate), which is also what the committed baselines use:

| | default | `--exact-verify` | committed baseline `2026-08-20` |
|---|---|---|---|
| output tokens | 621 / 622 / 616 | 1007 / 1010 / 1060 | 796 |
| server decode tok/s | 38.8 / 39.2 / 38.8 | 37.5 / 37.0 / 35.8 | 21.96 |
| accept_len_mean | — | **4.80** | 2.67 |
| verdict | **INCONCLUSIVE** | 37.0 tok/s | PASS 22.0 tok/s |

**The harness refuses to score the default arm**, on its own vacuity rule:

```
Fail: INCONCLUSIVE: run 1 emitted 621 tokens, below the 750-token
vacuity floor (of the 1500 budget) — too short a decode to measure a floor
```

That is the benchmark independently detecting the same thing this PR documents. The default arm is ~5% faster per token and produces **63% less content** for it — which is not a throughput win, it is a truncated answer delivered promptly.

## This also links up with the temp-0 length regression reported separately

There is an open report of the same checkpoint returning a materially shorter completion on a MinHeap prompt at temp 0 — 875 tokens on a receipts build versus ~505 with docstrings dropped — attributed to a window in `main`. I previously argued that was a different symptom from the degeneration in this PR and should not be merged with it. **On this evidence I was wrong.** The pinned harness puts all three on one axis:

```
796 tokens   committed baseline
621 tokens   default verify, today
1010 tokens  --exact-verify, today
```

Same direction, same prompt class, and the flag moves it. The likelier reading is one phenomenon — inexact verify logits truncating completions at temperature 0 — rather than a `main` window regression plus an unrelated DFlash bug. Worth checking whether the receipts build reproduces its 875 under `--exact-verify` before spending a bisect on the window.

## What stands from the original comment

Acceptance is higher with exactness on every measurement I have taken, and `accept_len_mean` 2.67 -> 4.80 on the pinned harness is the cleanest version of that. The mechanism is unchanged: corrupted verify logits reject correct drafts. What I withdraw is the specific "+30% to +90% at every concurrency" claim, which does not survive a better instrument.

rsafier:
**Second correction, and this one matters more than the first.** My previous comment showed `decode-floor` scoring INCONCLUSIVE by default (621 tokens, under the 750-token vacuity floor) and 37.0 tok/s with `--exact-verify`, and read that as the verify chain truncating completions. **It was not. That was the serve profile.**

Both of those runs used a serve with `--enable-prefix-caching` on at 128K context. Re-run on the LEAN profile — 32K ctx, 8 seqs, no prefix cache, which is the profile the project's own recorded baseline was taken on — the picture inverts:

| profile | verify | tokens | server tok/s | verdict |
|---|---|---|---|---|
| 128K + prefix cache | default | 621 / 622 / 616 | 38.9 | INCONCLUSIVE |
| 128K + prefix cache | exact | 1007 / 1010 / 1060 | 37.0 | 37.0 |
| **LEAN, no prefix cache** | **default** | **1266 / 1266 / 1266** | **39.9** | **39.9, accept_len 4.96** |
| LEAN, no prefix cache | exact | 943 / **553** / 943 | 38.2 | INCONCLUSIVE |
| committed baseline | — | 796 | 21.96 | PASS 22.0 |

So the truncation I attributed to inexact verify was the prefix-cache profile, and on the correct profile the DEFAULT arm is the one that scores — 1266 tokens, deterministic across three pinned runs. Exact, on the same profile, returned 943/553/943: **not deterministic across identical runs**, which is the #459 row-count residual showing up directly rather than in theory.

Concurrency on the lean profile, code prompt, aggregate tok/s at C=1/2/4/8:

```
default   56 /  88 / 113 / 123     accept 85 / 86 / 80 / 74 %
exact     52 /  72 /  78 /  98     accept 81 / 76 / 58 / 64 %
```

**`--exact-verify` costs 7-31% here.** My earlier "+18.9%" and "+30-90%" both came from measurements on the prefix-cache profile, where the default arm was pathologically bad for an unrelated reason. Both are withdrawn.

## What survives, and it is the part this PR documents

The correctness finding is unchanged and reproduces on the lean profile: by default, "count from 1 to 10" still returns `1, 2, 100, 100, 100, ...`, and video-fidelity still goes 0/2 and 0/4 at C=2/C=4 versus 2/2 and 4/4 with the flag. A config can be fast, high-acceptance (74-86%) and still emit repetition loops at temperature 0 — the pinned code benchmark does not catch it because code prompts are exactly where acceptance is healthy.

That is the argument for the doc change in this PR, and it stands independently of the throughput numbers. The flag is a real correctness/throughput trade at roughly -7% to -31%, not the free win I claimed. I have corrected the flag's doc text in this branch to say so, with the lean-profile figures rather than the cache-on ones.

Apologies for the churn on this thread — the underlying error was benchmarking a correctness question on a serve profile I had not validated for throughput.

rsafier:
Final update, and it rewrites this PR's premise: the gross temp-0 degeneration documented here was NOT the inexact GDN verify chain — it was a rewind-arithmetic bug in `k4_apply_verdict`, fixed in **#699**. `--exact-verify` masked it by changing dispatch, not numerics. With #699 the DEFAULT chain produces correct, deterministic temp-0 output on every probe in this doc. The #459 1-ULP row-count residual this PR documents is still real and still opt-in-relevant, but the headline examples (`1, 2, 100, 100...`, the video 0/2 0/4) belong to #699's bug. I'd suggest holding this doc PR until #699 lands, then re-scoping it to the residual it actually describes.

tbraun96:
**AutoMerger sieve verdict: `sieve:cleared` — and folded into stack #891 (`stack/perf-plumbing`).** All five commits are pure rustdoc on the real `--exact-verify` flag (`gdn_flags.rs:62`, `serve_args.rs:346`), and the sieve verified the re-scope claim against the live tree (`serve_load.rs:531` attributes the K=4 rewind fix to #699, so 'every repro passes with the flag OFF' is current, not stale). Why this stack: the diff touches `crates/`, which owes a certification campaign by path rule even for comment-only lines — riding #891's already-owed campaign makes the marginal cost zero. Closes with landing evidence after #891 merges. If any hunk does not make it across, point at it and I will lift it onto a fresh branch.

tbraun96:
Correction to my comment above: this PR is **four** commits, not five — the adversarial delta review counted and I mis-stated it. All four are the rustdoc lines described; nothing else changes.

DrRainbows:
Closing as already on `main`. These doc-comment changes landed via #891 (`356701d6`, an ancestor of `origin/main`; pickaxe on two distinct sentences from this PR's added lines both return that commit). Named artifact: `crates/spark-model/src/layers/qwen3_ssm/gdn_flags.rs:63-68` (the "Attribution warning ... verdict rewind bug (#699)" block). Both files from this PR's diff (`gdn_flags.rs`, `crates/spark-server/src/cli/serve_args.rs`) exist on `main`, and a three-way merge of this branch onto current `main` (`c2948e3f`) produces a tree identical to `main` — nothing here is missing, even though `main` has since moved past it (which is why the files are not byte-identical and why a commit-by-commit cherry-pick reports a conflict).

PR #668, opened 2026-08-21
perf(gemma4): build the MoE prefill copies — cold TTFT 5706 → 4975 ms (−12.8%)

tbraun96:
Second loader with the gap #660 fixes for Qwen3-VL. `build_moe_ffn` constructed its
`MoeLayer` and called neither `transpose_for_prefill` nor `predequant_for_prefill`,
so every prefill took the fallback `moe_w4a16_grouped_gemm` over N-major weights.

**Targets #660** — it carries the shared helper. Rebase to `main` once that lands.

## Measured

`bg-digitalservices/Gemma-4-26B-A4B-it-NVFP4A16`, dgx1, **one binary**, both arms
back-to-back, n=8, `ATLAS_MOE_PREFILL_COPIES` the only variable:

| | control (`=0`) | prefill copies | Δ |
|---|---:|---:|---:|
| **cold TTFT median** | 5705.61 ms | **4974.56 ms** | **−12.8% (−731 ms)** |
| spread | 0.77% | 1.97% | |
| `grouped_gate_up` | 543.2 ms | 115.3 ms | **−78.8%** |
| `grouped_silu_down` | 261.9 ms | 74.7 ms | **−71.5%** |
| `unpermute_blend` | 15.0 ms | 15.3 ms | control, flat |
| `pre_expert_norm` | 4.4 ms | 4.2 ms | control, flat |

−615 ms of MoE GEMM accounts for most of the −731 ms end-to-end. The two phases
reading the transposed tables fell ~75%; the two that don't are flat. At a ~5 s
baseline this is far outside the ~2% cross-restart variance seen on the 35B.

## Correctness

**Gate C2 (the NVFP4 tripwire) PASSES on both arms** — `tool_calls=3/3`,
`coherent=3/3`, `degenerate=0`.

Output is not bit-identical and **cannot** be: changing the weight layout changes
the GEMM reduction order, so low-margin argmax flips are expected. All 5 temp-0
prompts stayed coherent and correct on both arms — including the code prompt, where
both produce a valid two-pointer merge.

## SSOT

Uses `weight_loader::moe_prefill_copies_fit` from #660 — no second copy of the
budget arithmetic, and `ATLAS_MOE_PREFILL_COPIES=0` works here for free.

The probe is hoisted into `loader_a` and passed down, for the same reason as the VL
loader: `free_memory()` shrinks as layers load, so a per-layer probe would transpose
the early layers and skip the late ones, leaving prefill straddling both paths.

## Not touched

`loader_a.rs` carries a commented-out block disabling the **attention** prefill
weights:

```rust
// Skip prefill weight transpose for now — diagnose CUDA 700 first.
// TODO: re-enable once basic inference works
```

Different code path, different investigation. Re-enabling it inside a MoE change
would be silent scope creep — flagging it as worth its own look.

## Fan-out status

| Loader | Builds prefill copies? |
|---|---|
| qwen35, qwen3, minimax, laguna | yes (already) |
| qwen3_vl | **#660** |
| **gemma4** | **this PR** |
| nemotron, mistral, step3p7, deepseek_v4 | still not |

### DIFF
--- a/crates/spark-model/src/weight_loader/gemma4/loader_a.rs
+++ b/crates/spark-model/src/weight_loader/gemma4/loader_a.rs
@@ -29,6 +29,9 @@ pub(super) fn load_layers_impl(
     let variant = detect_nvfp4_variant(store, config);
     tracing::info!("Gemma-4 NVFP4 variant: {:?}", variant);
 
+    // Decided once, before any layer allocates — see the doc on the parameter.
+    let moe_prefill_copies =
+        config.num_experts > 0 && crate::weight_loader::moe_prefill_copies_fit(config, gpu);
     let absmax_k = gpu.kernel("quantize_nvfp4", "nvfp4_global_absmax")?;
     let quantize_k = gpu.kernel("quantize_nvfp4", "quantize_bf16_to_nvfp4")?;
     let stream = gpu.default_stream();
@@ -371,7 +374,18 @@ pub(super) fn load_layers_impl(
 
         // ── MoE experts (Gemma-4 26B) — extracted to loader_b ──
         let moe_ffn = build_moe_ffn(
-            store, &lp, i, config, gpu, variant, qctx, h, absmax_k, quantize_k, stream,
+            store,
+            &lp,
+            i,
+            config,
+            gpu,
+            variant,
+            qctx,
+            h,
+            absmax_k,
+            quantize_k,
+            moe_prefill_copies,
+            stream,
         )?;
 
         tracing::info!("L{i}: building attention layer...");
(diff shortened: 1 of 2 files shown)


tbraun96:
Closing as **superseded by #702**, not rejected — the work lands, just from the other PR.

`git merge-base --is-ancestor` confirms this branch is an **ancestor of #702**, so #702 contains every commit here plus nine more files. Cherry-picking this onto #702 returns `empty commit set passed` — there is nothing left to apply.

#702 also carries a **full 10-gate campaign** at sha `ba32287a0d` (agentic-webserver, bfcl-subset, bfcl-subset-echolp, concurrency-sweep, decode-floor, ssm-state-poisoning, ttft-cold, ttft-warm, video-fidelity, vision-fidelity). So the gemma4 prefill work is effectively **already stacked** — three PRs' worth of change certified once, which is exactly the shape this backlog pass is trying to produce elsewhere by hand.

Found while grouping the gemma4 perf PRs (#668, #700, #702, #706). #706 is **not** contained and still applies cleanly on top; it stays open.

One caveat on #702 itself: it is 77 commits behind `main` and its records name `ba32287a0d` while its head is `d21a0ccc6c`, so a rebase will likely invalidate the campaign and need a re-run. That is a #702 problem, not a reason to keep this one open.

PR #669, opened 2026-08-21
video-fidelity: measure the pad contract instead of accusing it (new mixed-media-pads cell)

rsafier:
Standalone against `main` — benchmark harness only, no engine code, unrelated to the #649 train.

`video-fidelity`'s `mixed-media` cell failed with a hardcoded verdict: *"the ordering contract between collection, markers and pad expansion is off"*. That verdict is wrong, and it cost a debugging session inside `msg_entry.rs` looking for a desync that was not there.

## The contract is intact — measured, not assumed

On the exact failing image+video pair:

* The clip's own allocation is **identical in every ordering** — 4 temporal groups, 14x14 patches, 196 vision tokens — and `image+video` is order-symmetric in `usage.prompt_tokens`. Nothing is truncated.
* Asked to *"describe each frame of the video in order"*, the model answers **"Frame 1: a white square with a black border..."** — that is the STILL IMAGE. It treats the still as frame 1 of the clip, so asked to list the clip's colours it enumerates from the image and the clip's last colour falls off the end of a four-item list.
* Swapping the grayscale still for a solid **magenta** one returns `magenta, red, green, blue, yellow` — all four clip colours — from the same plain list question. Video alone correctly answers "magenta? NO", so nothing bleeds.

A desync is content-independent. Anything a still's *colour* can move is the model, not the splice.

## What changed

**`mixed-media` keeps its strict assertion** — the qwen3.6-passes / qwen3.8-fails split is worth catching — but its failure text no longer names a cause it has not established, and a comment records what the cell actually measures: whether the CHECKPOINT keeps two media items apart.

**New `mixed-media-pads` measures the contract instead of asking about it:**

```
t(image + video) == t(image) + t(video) - t(text)
```

in the server's own `usage.prompt_tokens`. A pad run of the wrong length cannot satisfy it, and no model opinion enters. Verified exact on a live serve — `image 93 + video 240 - text 42 = 291 served`, delta 0 — and delta 0 again at a second image size (336x336) and in both media orders.

A **content** probe was tried here first and is recorded in the comment as a dead end, because it looks like the obvious design and does not work: on the failing pair every wording — "ends on", "final frame", "name the fourth", even "the still is not part of the video" — returned the same wrong colour 5/5, while the magenta swap returned the right one against byte-identical served rows.

**`request::image_body` and `request::text_array_body`** are the two baselines the arithmetic subtracts. They send `content` as an ARRAY to match `video_body`/`mixed_body`; `text_only_body` sends a bare string, which renders through a different template branch and costs a different number of tokens, so reusing it would make the identity disagree by a constant that has nothing to do with pad runs.

## Result

Legs go 13 -> 14. On the serve above: **13/14, with `mixed-media-pads` GREEN and `mixed-media` the only red** — which is now a legible statement (the engine is fine; this checkpoint merges two media items) rather than a false accusation pointing at the wrong file.

42 video tests pass, clippy and fmt clean, verified on `main`'s base rather than only on the branch it was cut from.

### DIFF
--- a/crates/atlas-plugin/src/benchmarks/video/driver.rs
+++ b/crates/atlas-plugin/src/benchmarks/video/driver.rs
@@ -446,6 +446,46 @@ impl Benchmark for VideoFidelity {
             }
 
             // ── Mixed: an image and a video in one request ───────────────
+            //
+            // TWO cells, because one question could not tell two failures
+            // apart and for a while was read as accusing the wrong component.
+            //
+            //   * `mixed-media` asks for the clip's colors as a LIST. That is
+            //     a question about the ENGINE (are the clip's frames all
+            //     there?) only if the model cooperates, and qwen3.8-27B does
+            //     not: it treats the still as FRAME 1 OF THE VIDEO, so it
+            //     enumerates from the image and the clip's last color falls
+            //     off the end of a four-item list. What this cell really
+            //     measures is whether the CHECKPOINT keeps two media items
+            //     apart. qwen3.6 does; qwen3.8 does not. Kept strict, because
+            //     that difference is worth catching — but read as a model
+            //     property, not as an engine defect.
+            //
+            //   * `mixed-media-pads` MEASURES the contract instead of asking
+            //     about it: t(image+video) must equal t(image) + t(video) -
+            //     t(text), in the server's own `usage.prompt_tokens`. A pad
+            //     run of the wrong length cannot satisfy that, and no model
+            //     opinion enters. It is the cell whose failure should be
+            //     believed, and the only one entitled to accuse the contract.
+            //
+            // A content question was tried here first and does NOT work. On
+            // the failing pair every wording ("ends on", "final frame", "name
+            // the fourth", "the still is not part of the video") returned the
+            // same wrong color 5/5 — but swapping the grayscale still for a
+            // solid magenta one returned the RIGHT one, while the served rows
+            // were the identical 4 groups / 196 tokens in both cases. Whatever
+            // a still's colour can move is the model, not the splice.
+            //
+            // Measured on 2026-08-21 against a serve where `mixed-media` was
+            // red: the contract was INTACT. The clip's own allocation was
+            // identical in every ordering (4 temporal groups, 14x14 patches,
+            // 196 vision tokens), asked for the fourth frame the model said
+            // "Yellow", asked to describe the frames it called the still
+            // "Frame 1", and swapping the grayscale still for a solid magenta
+            // one returned all four clip colors from the plain list question.
+            // Beware when re-testing: this model's answers move across
+            // paraphrases (3, 4 and 6 colors for one clip), so a single
+            // wording is not evidence in either direction.
             Phase::Mixed => {
                 self.phase = Phase::Integrity;
                 let h = self.handle()?;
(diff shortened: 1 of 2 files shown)


TheTom:
Testing review using Rapid Software Testing principles. Rapid Software Testing means treating a test as an experiment: state exactly what risk it observes, challenge its oracle with known-bad cases, and keep claims no broader than the failures it can detect.

The new relation is useful, but it proves **prompt-token additivity for this mixed request**, not absolute pad correctness. A wrong per-image pad length can still pass if the same additive error appears in both the image-only and mixed counts. For example, starting from the measured values, an image error of +5 gives `98 + 240 - 42 = 296`, and a mixed request containing that same bad image pad can also report 296. The identity is exact while the image pad is still wrong. I would rename the claim and failure text around mixed-media composition or add an independent absolute geometry oracle before saying a pad run of the wrong length cannot satisfy it.

There is also a direct false-green partition to test. `ChatOutcome.prompt_tokens` defaults to zero if the stream supplies no usage count. Four otherwise successful responses without usage make this cell green at `0 + 0 - 0 = 0`. `saturating_sub` similarly turns an impossible `text > image + video` baseline into zero. Missing or impossible measurements should produce `Error`, never `Match`.

Before merge, I recommend extracting the arithmetic/classification into a small pure function and running these cases:

1. Measured equality: `42, 93, 240, 291` returns Match.
2. Mixed count one low and one high returns Mismatch with the signed delta.
3. Any zero or missing usage input returns Error.
4. `text > image + video` returns Error instead of a saturated prediction.
5. Equal additive corruption, such as `42, 98, 240, 296`, documents the surviving false negative unless an absolute oracle is added.
6. Decoder-unavailable on each of the four request positions returns Skipped; another transport error in each position returns Error.
7. Request-shape tests prove `image_body` and `text_array_body` use array content, preserve media-before-text order, carry the requested MIME/data URI, and keep the same prompt/template controls as video and mixed bodies.
8. If order symmetry is part of the contract, run and score video+image as well as image+video. The PR description says both orders were checked manually, but the committed `mixed_body` covers only image+video.

One concrete text bug should also be fixed: the revised `mixed-media` failure says `mixed-media-pads` "asks for the LAST frame directly". The new cell does not ask that question. It makes four token-count requests.

The existing live measurements are good exploratory evidence. The fault injections above are what would prove the automated oracle fails for the intended reasons and cannot silently pass when its measurement instrument is absent.

tbraun96:
**AutoMerger sieve verdict: `sieve:cleared` — and folded into stack #891 (`stack/perf-plumbing`).** The gate-weakening check came back clean: the old `mixed-media` cell keeps its strict Match/Mismatch gating (only the failure wording is neutralized), and the new `mixed-media-pads` cell is a strict *addition* — 13→14 gating legs, verified against `score.rs::verdict()`. Why this stack: `crates/atlas-plugin` owes a campaign by path rule — riding #891's already-owed campaign makes the marginal cost zero. Attribution signature recorded: failures from this constituent surface in the bench legs, not the model. Closes with landing evidence after #891 merges. If any hunk does not make it across, point at it and I will lift it onto a fresh branch.

DrRainbows:
Closing as already on `main`. The `mixed-media-pads` cell landed, rewritten, via #891 (`356701d6`, an ancestor of `origin/main`). Named artifact: `crates/atlas-plugin/src/benchmarks/video/driver.rs:598` (`id: "mixed-media-pads"`, also at :610 and :634). Both files from this PR's diff (`driver.rs`, `request.rs`) exist on `main`; the content is folded rather than byte-identical, which is why I am naming the line. Every commit here cherry-picks empty onto current `main` (`c2948e3f`).

PR #671, opened 2026-08-21
atlas-core: make compute target checks deterministic

TheTom:
## Summary

Two tests in `atlas-core` could miss real regressions. This patch makes the NVIDIA metadata test run on machines without `nvcc` and checks all vendor names already accepted by the parser.

## Why the old tests were broken

`nvidia_target_extensions` put every assertion inside `if let Some(NvidiaTarget::new())`. If compiler discovery returned `None`, the test passed without checking anything. Compiler discovery is setup noise for a test that only reads static metadata.

`vendor_from_str` skipped `hip`, `oneapi`, and `sycl`, even though production accepts them. If one of those match arms disappeared, the test stayed green. `target_for_vendor` then treats the missing parse result like an omitted vendor and falls back to NVIDIA.

## How we proved they were broken

- With CUDA variables unset and `PATH=/usr/bin:/bin`, I changed NVIDIA's source extension from `cu` to `metal`. The old test passed because `NvidiaTarget::new()` returned `None` and no assertion ran.
- I removed `hip` from the AMD parser arm. The old parser test passed because it never supplied `hip`.

These are executed mutants, not hypothetical gaps.

## The fix

The metadata test now uses the existing explicit compiler-path constructor. It always reaches the four metadata assertions and does not invoke compiler discovery. The parser test now supplies every accepted vendor spelling. Both tests were renamed to describe the functions and values they check.

The same mutants fail after the change: `cu` to `metal` fails the source-extension assertion, and deleting `hip` fails the new alias assertion. Production code is unchanged.

## What this does not prove

This patch does not test `find_nvcc`, invoke a compiler, compile a kernel, or exercise a real GPU. Those require different tests and different setup.

## Test plan

- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy -p atlas-core --tests`
- [x] `typos crates/atlas-core/src/compute.rs`
- [x] PR CI: workspace tests, clippy, coverage, Metal tests, fmt, licenses, typos, and all-target nvcc build passed
- [ ] PR benchmark gate: Atlas invalidates all ten hardware records for any change under this Rust path, including `#[cfg(test)]`; fresh GPU records are required

## Notes for reviewers

- The diff is entirely inside `#[cfg(test)]`.
- The benchmark-gate failure and its ten invalidated records are documented in the PR discussion. I did not add fake records or weaken the gate.

## CLA

### DIFF
--- a/crates/atlas-core/src/compute.rs
+++ b/crates/atlas-core/src/compute.rs
@@ -255,24 +255,26 @@ mod tests {
     use super::*;
 
     #[test]
-    fn vendor_from_str() {
+    fn vendor_parse_accepts_supported_names() {
         assert_eq!(Vendor::parse("nvidia"), Some(Vendor::Nvidia));
         assert_eq!(Vendor::parse("CUDA"), Some(Vendor::Nvidia));
         assert_eq!(Vendor::parse("amd"), Some(Vendor::Amd));
         assert_eq!(Vendor::parse("rocm"), Some(Vendor::Amd));
+        assert_eq!(Vendor::parse("hip"), Some(Vendor::Amd));
         assert_eq!(Vendor::parse("apple"), Some(Vendor::Apple));
         assert_eq!(Vendor::parse("metal"), Some(Vendor::Apple));
         assert_eq!(Vendor::parse("intel"), Some(Vendor::Intel));
+        assert_eq!(Vendor::parse("oneapi"), Some(Vendor::Intel));
+        assert_eq!(Vendor::parse("sycl"), Some(Vendor::Intel));
         assert_eq!(Vendor::parse("unknown"), None);
     }
 
     #[test]
-    fn nvidia_target_extensions() {
-        if let Some(target) = NvidiaTarget::new() {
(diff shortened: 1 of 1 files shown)


TheTom:
CI follow-up: the PR benchmark gate is red because its path-based policy invalidates all ten hardware records for any edit to crates/atlas-core/src/compute.rs, including this diff, which is entirely inside #[cfg(test)]. The job reports NONE for agentic-webserver, vision-fidelity, video-fidelity, ttft-warm-gate, ttft-cold-gate, bfcl-subset, bfcl-subset-echolp, ssm-state-poisoning-gate, decode-floor, and concurrency-sweep. This is the deliberate policy recorded in ADR-0013: Rust test-only changes have no trusted exemption and must pay a real gate run. I am not claiming hardware evidence, fabricating records, or changing production code to evade the gate. All completed non-benchmark checks are green; fresh GPU measurements are the remaining legitimate requirement.

tbraun96:

PR #674, opened 2026-08-21
atlas-core: cover expert partition remainders

TheTom:
## Summary

`test_expert_parallelism_range` only split 512 experts across two ranks, so it exercised an exactly divisible case and could not verify its remainder rule. Removing the last-rank remainder branch (`end = start + per_rank`) left the old test green even though expert 512 would be dropped from a 513-expert model.

This renames the check to state the partition contract and adds the smallest non-divisible case: 513 experts across two ranks. Rank 1 must own `[256, 513)`, and expert 512 must map to its local index. Replaying the same production mutant now fails with `(256, 512)` instead of `(256, 513)`.

## Test plan

- [x] `typos`
- [x] Same remainder-dropping mutant: old test passed; strengthened test failed

## Notes for reviewers

This is test-only. It does not change expert partitioning and does not claim coverage of invalid ranks/world sizes, more ranks than experts, collectives, or end-to-end multi-GPU routing.

## CLA

### DIFF
--- a/crates/atlas-core/src/config/tests.rs
+++ b/crates/atlas-core/src/config/tests.rs
@@ -384,7 +384,7 @@ fn test_parse_nemotron_h_puzzle_config() {
 }
 
 #[test]
-fn test_expert_parallelism_range() {
+fn expert_parallelism_partitions_experts_without_dropping_remainder() {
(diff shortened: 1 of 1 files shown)


tbraun96:

PR #676, opened 2026-08-21
atlas-core: preserve Gemma-4 config contracts

TheTom:
## Summary

The dense Gemma-4 parser check supplied a 1024-token sliding window and implied Gemma embedding scaling, but observed neither. Zeroing either parsed value left the old test green. While auditing that path, a separate functional defect surfaced: the MoE branch set `norm_topk_prob = true`, then common setup unconditionally overwrote it to `false`.

The first commit renames the dense check and observes the window mask plus `sqrt(hidden_size)` embedding scale; the exact zeroing mutants now fail. The second adds a canonical Gemma-4 MoE config regression and preserves normalization when `num_experts > 0`. The regression failed on the old production code, passes with the fix, and fails again when the unconditional `false` is restored.

## Test plan

- [x] `typos`
- [x] Two old-green/new-red dense-control mutants
- [x] Old-red/new-green/old-mutant-red MoE normalization replay

## Notes for reviewers

Hugging Face's `Gemma4TextRouter` softmaxes all experts, selects top-K, then divides selected weights by their sum. Atlas passes `norm_topk_prob` into its MoE forward kernels, so the overwritten flag changed expert mixture weights rather than only config metadata.

The new fixture samples Google's published 128-expert/top-K-8/704-wide MoE values and canonical nested RoPE shape. It does not load the checkpoint, run a router kernel, or compare logits. Workspace Clippy is left for Linux CI; the Rust-crate benchmark gate will require external GPU records unavailable on this host.

## CLA

### DIFF
--- a/crates/atlas-core/src/config/parsers/gemma4.rs
+++ b/crates/atlas-core/src/config/parsers/gemma4.rs
@@ -107,7 +107,6 @@ pub(crate) fn parse_gemma4_params(raw: &serde_json::Value) -> Result<ModelConfig
         config.num_experts = num_experts;
         config.num_experts_per_tok = top_k_experts;
         config.moe_intermediate_size = moe_intermediate_size;
-        config.norm_topk_prob = true;
         config.shared_expert_intermediate_size = 0; // dense MLP is separate, not a shared expert
     } else {
         config.num_experts = 0;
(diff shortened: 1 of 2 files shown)


tbraun96:

PR #677, opened 2026-08-21
atlas-core: validate Nemotron-H config geometry

TheTom:
## Summary

The Nemotron-H test claimed 64×64 Mamba geometry but observed only the product 4096. A same-product 32×128 corruption stayed green, as did zero state size, zero group count, an erased `backbone` prefix, and disabled routing normalization. The parser also accepted zero/non-divisible Mamba geometry that later becomes allocation, kernel geometry, and `d_inner / n_groups` input.

The first commit directly observes NVIDIA's published Mamba factors, state, groups, routing normalization, and Atlas weight prefix; every exact mutant now fails. The second validates Mamba head dimension, state size, nonzero groups, and group divisibility. Its table-driven regression originally reported every invalid case accepted, passes with validation, and fails again if validation is removed.

## Test plan

- [x] `typos`
- [x] Five independent old-green/new-red parser mutants
- [x] Aggregate old-red/new-green/validation-removed-red replay

## Notes for reviewers

The positive fixture values come from NVIDIA's `NVIDIA-Nemotron-3-Nano-30B-A3B-BF16` config. The new validation is shared so every Mamba-backed config fails before allocation or kernel launch with a field-naming error.

This does not bound maximum dimensions, prove multiplication cannot overflow, validate checkpoint tensor shapes, or run a Mamba/MoE kernel. Workspace Clippy is left to Linux CI, and the benchmark gate requires hardware records unavailable on this host.

## CLA

### DIFF
--- a/crates/atlas-core/src/config.rs
+++ b/crates/atlas-core/src/config.rs
@@ -629,5 +629,20 @@ pub(crate) fn validate_config(config: &ModelConfig) -> Result<()> {
         );
     }
 
+    if config.mamba_num_heads > 0 {
+        if config.mamba_head_dim == 0 {
+            anyhow::bail!("mamba_head_dim must be greater than zero");
+        }
+        if config.ssm_state_size == 0 {
+            anyhow::bail!("ssm_state_size must be greater than zero");
+        }
+        if config.n_groups == 0 {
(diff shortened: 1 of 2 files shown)


tbraun96:

PR #678, opened 2026-08-21
atlas-core: strengthen Qwen3.5 nested config contract

TheTom:
## Summary

The reduced Qwen3.5 MoE nested-config test observed dimensions and aggregate SSM sizes but could not detect four runtime-control corruptions. Disabling gated attention, clearing nested-layout detection, expanding rotary coverage from 0.25 to 1.0, and reducing the convolution kernel width from 4 to 1 each left the old test green.

This renames the check to its attention/SSM/layout scope and directly observes those four controls plus the initial empty prefix used by nested weight-prefix detection. Replaying each exact mutant now fails at the intended assertion.

## Test plan

- [x] `typos`
- [x] Four independent old-green/new-red mutation replays

## Notes for reviewers

These fields feed attention projection sizing, buffer sizing, prefix auto-detection, rotary dimensions, convolution state, and SSM kernels. Production already returns the correct values for this reduced input, so the patch is test-only.

This is not an exact copy of current `Qwen/Qwen3.5-35B-A3B`: current publisher configs also carry MRoPE and vision fields handled by separate dispatch tests. It does not load weights or run inference.

## CLA

### DIFF
--- a/crates/atlas-core/src/config/tests.rs
+++ b/crates/atlas-core/src/config/tests.rs
@@ -47,7 +47,7 @@ fn test_parse_actual_config() {
 }
 
 #[test]
-fn test_parse_qwen35_nested_config() {
+fn qwen35_moe_nested_config_maps_attention_ssm_and_layout_controls() {
     let json = r#"{
         "model_type": "qwen3_5_moe",
         "text_config": {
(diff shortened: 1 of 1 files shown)


tbraun96:

PR #679, opened 2026-08-21
atlas-core: strengthen Qwen3-VL config contract

TheTom:
## Summary

The reduced Qwen3-VL MoE config test checked broad model dimensions but could not detect three runtime-significant corruptions. Forcing gated attention on, reducing routed experts per token from 8 to 1, and clearing the MoE intermediate width from 768 to 0 each left the old test green.

This renames the test to state its attention-and-routing scope and directly observes all three controls. Replaying each exact production mutant now fails at the intended assertion, while the restored parser passes.

## Test plan

- [x] `typos`
- [x] Three independent old-green/new-red mutation replays

## Notes for reviewers

The fixture values match the text config published for [`Qwen/Qwen3-VL-30B-A3B-Instruct`](https://huggingface.co/Qwen/Qwen3-VL-30B-A3B-Instruct/blob/main/config.json): 8 routed experts per token and a 768-wide expert MLP. These controls feed MoE kernel/scratch sizing and expert GEMMs; `attn_gated` changes attention projection and buffer sizing.

Production already returns the correct values for this reduced input, so the patch is test-only. The reduced fixture intentionally does not claim coverage of the publisher's vision tower or MRoPE fields, which have separate dispatch coverage. It does not load weights or run inference.

## CLA

### DIFF
--- a/crates/atlas-core/src/config/tests.rs
+++ b/crates/atlas-core/src/config/tests.rs
@@ -109,7 +109,7 @@ fn test_parse_qwen35_nested_config() {
 }
 
 #[test]
-fn test_parse_qwen3_vl_config() {
+fn qwen3_vl_moe_config_maps_attention_and_routing_controls() {
     let json = r#"{
         "model_type": "qwen3_vl_moe",
         "text_config": {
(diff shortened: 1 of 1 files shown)


tbraun96:

PR #680, opened 2026-08-21
atlas-core: strengthen Qwen3-Next factory contract

TheTom:
## Summary

The hard-coded Qwen3-Next factory test relied on derived ratios and aggregate SSM sizes, so it could not detect five shape corruptions. Doubling hidden width; doubling Q and KV heads together; changing linear key heads from 16×128 to 8×256; reducing routed experts per token from 10 to 1; and reducing convolution width from 4 to 1 each left the old test green.

This renames the check to its attention/SSM/routing scope and directly observes the factors and controls hidden by those aggregates. Replaying each exact factory mutant now fails at its intended assertion, while restored production passes.

## Test plan

- [x] `typos`
- [x] Five independent old-green/new-red mutation replays

## Notes for reviewers

The expected values match the checked-in Qwen3-Next publisher fixture and NVIDIA's [`Qwen3-Next-80B-A3B-Instruct-NVFP4` config](https://huggingface.co/nvidia/Qwen3-Next-80B-A3B-Instruct-NVFP4/blob/main/config.json). The individual factors control attention projection shape, SSM state/convolution layout, MoE kernel fan-out, scratch sizing, and expert execution; equal derived totals do not make different factorizations interchangeable.

Production already returns the correct values, so this patch is test-only. It does not prove real weight loading, allocation, kernel launch, or inference output. The hard-coded factory is also used as a construction baseline by other parsers and runtime tests; this PR does not claim every field in that broad struct is covered.

## CLA

### DIFF
--- a/crates/atlas-core/src/config/tests.rs
+++ b/crates/atlas-core/src/config/tests.rs
@@ -7,10 +7,14 @@
 use super::*;
 
 #[test]
-fn test_qwen3_default_config() {
+fn qwen3_next_factory_preserves_attention_ssm_and_routing_shape() {
     let cfg = ModelConfig::qwen3_next_80b_nvfp4();
+    assert_eq!(cfg.hidden_size, 2048);
     assert_eq!(cfg.num_hidden_layers, 48);
+    assert_eq!(cfg.num_attention_heads, 16);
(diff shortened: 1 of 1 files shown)


tbraun96:

PR #682, opened 2026-08-21
profile: name the attention kernel that actually ran

tbraun96:
`aprof!("flash_attn_64", t0)` sat **after** an if/else that dispatches two different
kernels — `inferspark_prefill_512` when `hd > 256` and the handle resolves,
`prefill_attention_64` otherwise — and labelled both with the second one's name.

## What that hid

Measured on `bg-digitalservices/Gemma-4-26B-A4B-it-NVFP4A16`, 4096-token prefill.
This model has **5 global (512-wide)** and **25 sliding (256-wide)** attention layers:

| kernel | total | layers | per layer |
|---|---:|---:|---:|
| **`flash_attn_512_scalar`** | **16,484.4 ms** | 5 | **3,296.9 ms** |
| `flash_attn_64` | 49.7 ms | 25 | 2.0 ms |

The old label attributed all **16,484 ms** to `flash_attn_64` — a kernel that in
truth accounts for **49.7 ms**. Off by 330×, and pointing at the wrong kernel
entirely.

I built two root-cause hypotheses on that reading and **both were wrong**:

- *"the 512 kernel isn't shipped for this target"* — it is, for three targets
  (`deepseek-v4-flash`, `gemma-4-26b-a4b`, `gemma-4-31b`);
- *"the `wide_head_dim` probe gates it off"* — `config/parsers/gemma4.rs:138` sets
  `config.head_dim = global_head_dim` (512), so the probe is true.

Only after fixing the label did the real picture appear: **3,297 ms per layer on the
scalar 512-wide path vs 2.0 ms on the tiled 64 path — a 1,650× per-layer gap**, and
91% of this model's prefill.

## The change

One bool hoisted out of the condition; the label reads from it. **No behaviour
change** — profiling-only path, gated on `ctx.profile`.

### DIFF
--- a/crates/spark-model/src/layers/qwen3_attention/prefill/cache_skip.rs
+++ b/crates/spark-model/src/layers/qwen3_attention/prefill/cache_skip.rs
@@ -625,7 +625,8 @@ impl Qwen3AttentionLayer {
                 .arg_u32(hd)
                 .launch(stream)?;
         }
-        if hd > 256 && self.prefill_attn_512_k.0 != 0 {
+        let wide_head_path = hd > 256 && self.prefill_attn_512_k.0 != 0;
+        if wide_head_path {
             // HDIM=512: use scalar reference kernel (BR=16, correct for any head_dim)
             // Full-attention layers (this path) always pass sliding_window=0.
             ops::prefill_attention(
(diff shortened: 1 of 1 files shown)


TheTom:
The implementation looks mechanically sound, but this PR needs a repeatable regression oracle because its entire claim is that the profiling label identifies the kernel path that actually ran.

I am using Rapid Software Testing principles here. In plain terms: state exactly what the code claims, exercise the boundaries where that claim can break, and prove the test fails against known-bad implementations rather than accepting a passing test at face value. For this change, I would test it this way:

1. Extract the branch decision and profile label into a small typed selector (prefer an enum such as `Wide512` / `Tiled64` over independently recomputing strings). Use that same selected value for both dispatch and the label.
2. Unit-test the meaningful partitions:
   - `hd = 512`, 512-kernel handle present → wide dispatch and `flash_attn_512_scalar`.
   - `hd = 512`, 512-kernel handle absent → fallback dispatch and `flash_attn_64`.
   - `hd = 256`, 512-kernel handle present → fallback dispatch and `flash_attn_64`.
   - Optionally, profiling disabled → no profiling record; this is separate from path selection.
3. Prove failure sensitivity, not just green execution:
   - Restore the old constant `flash_attn_64` label. The wide-path case must fail.
   - Ignore handle availability and select wide solely from `hd > 256`. The missing-handle fallback case must fail.
   - Force fallback for every input. The valid wide-path case must fail.
4. Keep the real Gemma profiling run as exploratory/end-to-end evidence. Record the exact model revision, command, profile flags, prompt/token count, hardware, and the expected presence/count of both labels. Do not make absolute timing values a CI oracle; timings are environment-sensitive.

This separates the claims cleanly: the unit test proves label/dispatch agreement and boundary behavior; the real-model run proves the instrumentation is reached in an actual workload. It does not claim kernel numerical correctness or performance correctness, which belong to separate tests.

tbraun96:
Closing: already on `main`, in superset form, through #721 (`1dc3c0dcfb`).

`crates/spark-model/src/layers/qwen3_attention/prefill/cache_skip.rs:704-709` on `main` `9e597531` labels the timer by `match (wide_head_path, self.prefill_attn_512_is_tc)` → `flash_attn_512_tc` / `flash_attn_512_scalar` / `flash_attn_64`, which is this fix plus the tensor-core distinction. This branch's remaining diff against `main` is only `ATLAS_*` → `AVAROK_*` drift. No branch deleted.

PR #683, opened 2026-08-21
atlas-core: reject zero GGUF key length

TheTom:
## Summary

The GGUF parser honors an explicit `{arch}.attention.key_length`, but it also accepted zero and returned a successful `ModelConfig` with `head_dim = 0`. That invalid dimension then reaches KV-cache sizing, attention projection shapes, rotary dimensions, scaling, and kernel arguments instead of failing at config load.

This rejects an explicit zero key length at the metadata boundary and names the offending namespaced key. The regression failed against the previous parser by receiving `Ok(ModelConfig { head_dim: 0, ... })`, passes with the validation, and fails again when the four-line guard is removed.

## Test plan

- [x] `typos`
- [x] Old-red/new-green regression plus validation-removal replay

## Notes for reviewers

The [GGUF metadata specification](https://github.com/ggml-org/ggml/blob/master/docs/gguf.md) defines `attention.key_length` as the size of a key head and says the parser should derive it from embedding width and head count only when the key is absent. This change preserves that precedence and rejects only the impossible explicit zero dimension.

The neighboring `explicit_key_length_wins` test already rejects replacing a valid explicit 96 with the derived 128, so it remains unchanged. This PR does not broaden validation to other zero-valued GGUF dimensions; those registered error paths are being audited separately. It does not load weights or run inference. Workspace Clippy and the ten-record GPU benchmark requirement remain CI/external-hardware gates.

## CLA

### DIFF
--- a/crates/atlas-core/src/config/gguf.rs
+++ b/crates/atlas-core/src/config/gguf.rs
@@ -111,6 +111,10 @@ pub fn config_from_gguf(inputs: &GgufConfigInputs) -> Result<ModelConfig> {
     // head_dim: explicit key_length if present, else hidden_size / head_count.
     // Erroring on a non-divisible fallback avoids a silently-wrong head_dim.
     let head_dim = match meta.get_u64(&k("attention.key_length")) {
+        Some(0) => bail!(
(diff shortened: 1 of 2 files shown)


tbraun96:

PR #684, opened 2026-08-21
atlas-core: validate GGUF Gemma softcaps

TheTom:
## Summary

The GGUF Gemma parser copied `final_logit_softcapping` from an `f64` metadata value into `f32` without validating its domain. Negative values, NaN, infinity, and finite values larger than `f32::MAX` all produced successful configs; the latter overflowed to positive infinity during the cast. At runtime, non-positive and NaN values silently skip the softcap kernel, while positive infinity is passed into the softcap formula.

This accepts the supported domain—zero to disable softcapping or a positive finite `f32`—and rejects values that cannot represent that contract. The regression reports all four previously accepted invalid partitions, verifies explicit zero remains valid, and names the namespaced GGUF key in every error.

## Test plan

- [x] `typos`
- [x] Old-red/new-green invalid-domain regression plus validation-removal replay
- [x] Too-strict zero-rejection mutant replay

## Notes for reviewers

The existing `gemma_sets_embed_scale_and_softcap` check already rejects independently clearing the derived embedding scale and ignoring an explicit softcap of 30. It remains unchanged apart from reusing the extracted Gemma fixture.

For the new domain regression, previous production accepted `-30.0`, `NaN`, `+∞`, and `f64::MAX`; removing the validation after the fix restores that exact aggregate failure. Changing the lower bound from `>= 0` to `> 0` makes the explicit-disabled case fail, proving zero remains a supported boundary rather than being rejected accidentally.

This does not change missing-key behavior (still disabled at 0), positive finite values, embedding scaling, weight loading, or inference output. Workspace Clippy and the ten-record GPU benchmark requirement remain CI/external-hardware gates.

## CLA

### DIFF
--- a/crates/atlas-core/src/config/gguf.rs
+++ b/crates/atlas-core/src/config/gguf.rs
@@ -215,10 +215,14 @@ pub fn config_from_gguf(inputs: &GgufConfigInputs) -> Result<ModelConfig> {
         config.embed_scale = (hidden_size as f32).sqrt();
         // Logit softcap: honor the GGUF key if present (gemma2), else 0.0
         // (disabled). gemma3+ dropped softcapping.
-        config.final_logit_softcapping = meta
-            .get_f64(&k("final_logit_softcapping"))
-            .map(|v| v as f32)
-            .unwrap_or(0.0);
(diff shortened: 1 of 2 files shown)


tbraun96:

PR #686, opened 2026-08-21
atlas-core: validate GGUF KV head geometry

TheTom:
## Summary

The GGUF parser accepted explicit KV head counts that cannot form valid grouped-query attention: zero, a count that does not divide the Q-head count, and a count larger than the Q-head count. For a 32-head fixture, values 0, 7, and 64 all produced successful configs and reached cache and attention shape construction.

This requires an explicit KV head count to be a non-zero divisor of the Q-head count. The regression reported all three values accepted on previous production, passes with validation, and reports all three again when that exact guard is removed.

## Test plan

- [x] `typos`
- [x] Old-red/new-green three-partition regression plus validation-removal replay

## Notes for reviewers

The existing `kv_heads_default_to_mha` test is sound and unchanged. Replacing its absent-key default with 1 makes it fail with 1 versus 32. This PR addresses the separate explicit-value domain.

The guard is scoped to nonzero Q-head counts so `attention.head_count = 0` remains owned by the required-dimension audit and can produce a field-specific error. Valid MHA and GQA partitions, including 32 KV heads and the fixture's explicit 8 KV heads, continue to pass.

KV head count controls KV-cache dimensions, projection shapes, GQA grouping, and attention kernels. This change does not load weights or run inference. Workspace Clippy and the ten-record GPU benchmark requirement remain CI and external-hardware gates.

## CLA

### DIFF
--- a/crates/atlas-core/src/config/gguf.rs
+++ b/crates/atlas-core/src/config/gguf.rs
@@ -107,6 +107,14 @@ pub fn config_from_gguf(inputs: &GgufConfigInputs) -> Result<ModelConfig> {
         .get_u64(&k("attention.head_count_kv"))
         .map(|v| v as usize)
         .unwrap_or(num_attention_heads);
+    if num_attention_heads > 0
+        && (num_key_value_heads == 0 || !num_attention_heads.is_multiple_of(num_key_value_heads))
+    {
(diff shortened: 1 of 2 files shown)


tbraun96:

PR #688, opened 2026-08-21
atlas-core: validate GGUF vocab sources

TheTom:
## Summary

The GGUF config path accepted a zero vocabulary from explicit metadata, zero `token_embd.weight` rows, or an empty tokenizer list. It also accepted metadata vocab size 32000 beside a physical embedding tensor with 128256 rows. All four cases produced successful configs before this change.

This rejects a selected zero vocabulary and rejects disagreement between explicit metadata and embedding rows. The four-partition regression fails on previous production, passes with both validations, admits the three zero cases if the nonzero guard is removed, and admits the mismatch if the consistency guard is removed.

## Test plan

- [x] `typos`
- [x] Replayed the nonzero and consistency guard removals independently

## Notes for reviewers

The existing `vocab_from_token_embd_rows` test is sound and unchanged. Removing the tensor-row fallback makes it fail because no later source exists in that fixture. This PR addresses the separate input-domain and cross-source risks found while auditing that test.

The GGUF reference maps the resolved vocabulary dimension directly into `token_embd.weight`: https://github.com/ggml-org/llama.cpp/blob/master/src/models/llama.cpp. Atlas likewise uses `vocab_size` for embedding, LM-head, logits, and token-mask shapes. Rejecting malformed metadata at config load gives a local error before those paths construct incompatible shapes.

Tokenizer token count remains a fallback only. This change does not require it to equal a potentially padded model vocabulary. It does not load weights, run a model, or validate inference output. Workspace Clippy and the ten-record GPU benchmark requirement remain CI and external-hardware gates.

## CLA

### DIFF
--- a/crates/atlas-core/src/config/gguf.rs
+++ b/crates/atlas-core/src/config/gguf.rs
@@ -124,15 +124,25 @@ pub fn config_from_gguf(inputs: &GgufConfigInputs) -> Result<ModelConfig> {
     };
 
     // vocab_size: explicit key → token_embd rows → tokenizer token list length.
-    let vocab_size = meta
-        .get_u64(&k("vocab_size"))
-        .map(|v| v as usize)
+    let metadata_vocab = meta.get_u64(&k("vocab_size")).map(|v| v as usize);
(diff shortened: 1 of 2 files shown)


tbraun96:

PR #692, opened 2026-08-21
prefix-cache: stop handing a partial tail block to a second writer (no COW exists)

rsafier:
Standalone against `main` — `spark-runtime` only, no engine or scheduler code.

**The radix walk's sub-block arm hands a shared physical block to a sequence that will write into it, and there is no copy-on-write anywhere in the paged KV manager.**

## The defect

When a lookup's tail doesn't fill a block, `walk()` matches a PARTIAL block from two sources: a child node whose full block starts with the remaining tokens, or the `partial_suffix` stashed on the current node. Either way the shared physical block is pushed into the requester's `block_table` — where it becomes the requester's **writable tail**: its very next KV rows (positions `matched..block_end`) are written into a block another owner still maps. The `partial_suffix` donor is a LIVE sequence still decoding into those very rows. Two writers, one block.

The same arm has a second, independent hole: `inc_refs` walks full-block NODES only, so the matched partial block enters the requester's block_table with **no ref taken** — the eventual `free_sequence` underflows the donor's count.

## Why it survived until now

Measured 2026-08-21 (qwen3.8-27B + DFlash2 on GB10, identical concurrent prompts at temperature 0): while two writers stay in byte-identical lockstep, the collisions write the same bytes and the tear is **benign** — which is exactly why serial decode always looked clean, and why every single-stream test passes. Any desync between the writers (in our case, batched speculative verify's band-dependent rounding) makes the writes differ, and every reader then attends over a torn block.

## The change

The sub-block arm is now **default OFF** (`allow_partial_tail`), with:

* `ATLAS_PREFIX_PARTIAL_TAIL=1` restoring the old behaviour for A/B only,
* an explicit `RadixTree::with_partial_tail_sharing(bool)` constructor so BOTH arms stay deterministically testable — the env read is process-global and useless inside a test binary.

Cost of OFF: up to `block_size - 1` tokens of prefill recompute per hit, and non-block-aligned exact-repeat prompts no longer reach the Marconi exact-leaf shortcut (`matched < total`). Correctness first; the win-back is a copy-on-write tail — fresh block + per-layer D2D of the matched rows, event-ordered against the donor's stream — documented at the match site for whoever wants those tokens back.

## Tests

New `a_partial_tail_block_is_never_handed_to_a_second_writer` pins the default contract by name. The five legacy partial-suffix tests now opt in explicitly via the constructor and keep the old arm's internals honest (different-suffix refusal, multi-block prefix, shorter-lookup prefix match, child-key sub-block match). 86 radix tests pass on this branch, fmt and clippy clean, built on `main`'s base rather than the branch the fix was found on.

### DIFF
--- a/crates/spark-runtime/src/radix_tree.rs
+++ b/crates/spark-runtime/src/radix_tree.rs
@@ -69,6 +69,17 @@ impl RadixTree {
             snapshot_index: Mutex::new(SsmSnapshotIndex::new()),
         }
     }
+
+    /// Explicit-arm constructor so BOTH sides of the partial-tail gate stay
+    /// deterministically testable (the default constructor reads the
+    /// `ATLAS_PREFIX_PARTIAL_TAIL` env, which is process-global and therefore
+    /// useless inside a test binary).
+    pub fn with_partial_tail_sharing(on: bool) -> Self {
+        Self {
+            inner: Mutex::new(RadixTreeInner::with_partial_tail_sharing(on)),
+            snapshot_index: Mutex::new(SsmSnapshotIndex::new()),
+        }
+    }
 }
 
 impl PrefixCache for RadixTree {
--- a/crates/spark-runtime/src/radix_tree/inner.rs
+++ b/crates/spark-runtime/src/radix_tree/inner.rs
@@ -68,10 +68,21 @@ pub(super) struct RadixTreeInner {
     /// the pre-LoRA single-root tree.
     roots: HashMap<u64, NodeId>,
     access_counter: u64,
+    /// Whether `walk` may match a PARTIAL tail block (the sub-block arm).
+    /// Default FALSE — that arm is unsound for a block the requester will
+    /// write into; see the comment at the match site. `ATLAS_PREFIX_PARTIAL_TAIL=1`
+    /// restores the old behaviour for A/B only.
+    allow_partial_tail: bool,
 }
 
 impl RadixTreeInner {
     pub(super) fn new() -> Self {
+        Self::with_partial_tail_sharing(
+            std::env::var("ATLAS_PREFIX_PARTIAL_TAIL").as_deref() == Ok("1"),
+        )
+    }
+
+    pub(super) fn with_partial_tail_sharing(allow_partial_tail: bool) -> Self {
         let root = RadixNode {
             children: HashMap::new(),
             block_idx: u32::MAX,     // sentinel — root has no block
@@ -90,6 +101,7 @@ impl RadixTreeInner {
             free_nodes: Vec::new(),
             roots,
             access_counter: 0,
+            allow_partial_tail,
         }
     }
 
@@ -185,8 +197,42 @@ impl RadixTreeInner {
         // 2. A partial suffix stored on the current node.
         // This enables warm-cache TTFT optimization by matching ALL prompt tokens
         // even when total % block_size != 0.
+        //
+        // ★ DEFAULT OFF (`allow_partial_tail`), because both arms hand the
+        // caller a block it will WRITE INTO while another owner still maps it.
+        // The matched partial block becomes the requester's TAIL: its very
+        // next KV rows (positions matched..block_end) are written into that
+        // shared physical block. Arm 2's donor is a LIVE sequence still
+        // decoding into the same rows; arm 1's donor block is full, but the
+        // requester's tail writes clobber committed rows other sequences map
+        // read-only. There is no copy-on-write anywhere in the paged KV
+        // manager, so this is two writers, one block.
+        //
+        // Measured consequence (2026-08-21, qwen3.8-27B + DFlash2, GB10):
+        // with identical prompts in flight the writers stay byte-identical in
+        // lockstep and the tear is BENIGN — which is why serial decode looked
+        // clean — but any desync (the batched verify's band-dependent
+        // rounding was the trigger) makes the writes differ, the block holds
+        // a mix of two streams, and every reader derails mid-token at the
+        // block boundary: C=4 went 0/4 with three sequences emitting
+        // IDENTICAL wrong text (all reading the same torn block).
+        //
+        // The same arm also has a refcount hole: `inc_refs` walks full-block
+        // NODES only, so a matched partial block is pushed into the
+        // requester's block_table WITHOUT a ref — the eventual free
+        // underflows the donor's count.
+        //
+        // Cost of OFF: up to block_size-1 tokens of prefill recompute per
+        // hit, and non-block-aligned exact-repeat prompts no longer reach the
+        // Marconi exact-leaf shortcut (matched < total). Correct first;
+        // a copy-on-write tail (fresh block + per-layer D2D of the matched
+        // rows, event-ordered against the donor's stream) is the way to win
+        // the tokens back if anyone wants them.
         let remainder = tokens.len() - matched_tokens;
-        if remainder > 0 && remainder < block_size && matched_tokens == num_full_blocks * block_size
+        if self.allow_partial_tail
+            && remainder > 0
+            && remainder < block_size
+            && matched_tokens == num_full_blocks * block_size
         {
             let suffix = &tokens[matched_tokens..];
             let mut found = false;
--- a/crates/spark-runtime/src/radix_tree/tests/snapshot.rs
+++ b/crates/spark-runtime/src/radix_tree/tests/snapshot.rs
@@ -110,18 +110,48 @@ fn test_intermediate_snapshot_survives_tree_eviction() {
 }
 
 // ── Partial suffix tests ──
+//
+// The sub-block arm is DEFAULT OFF: a matched partial block becomes the
+// requester's writable TAIL while another owner still maps it, and there is
+// no copy-on-write in the paged KV manager. Measured 2026-08-21: two live
+// sequences sharing a partial tail derailed mid-token the moment their
+// streams desynced (batched DFlash verify at C=4 went 0/4, three sequences
+// emitting identical wrong text — all reading the same torn block). The
+// legacy arm stays testable through `with_partial_tail_sharing(true)`.
 
+/// ★ The regression test for the torn-tail bug: with the default tree, a
+/// prompt whose tail does not fill a block must match ONLY its full blocks —
+/// the partial block stays exclusively the donor's, because the requester
+/// would otherwise write into it.
 #[test]
-fn test_partial_suffix_insert_and_lookup() {
+fn a_partial_tail_block_is_never_handed_to_a_second_writer() {
     let tree = RadixTree::new();
     // 20 tokens = 1 full block (16) + 4 partial
     let tokens: Vec<u32> = (0..20).collect();
-    let block_table = vec![10, 20]; // block for full + block for partial
+    tree.insert(&tokens, &[10, 20], &[], 16, 0, 0);
 
-    tree.insert(&tokens, &block_table, &[], 16, 0, 0);
     let m = tree.lookup(&tokens, 16, 0, 0);
+    assert_eq!(
+        m.matched_tokens, 16,
+        "the 4-token tail must NOT match: block 20 is the donor's live tail"
+    );
+    assert_eq!(
+        m.matched_blocks,
+        vec![10],
+        "block 20 handed out here becomes a second writer's tail — the torn-block bug"
+    );
+    tree.release(&tokens[..16], 16, 0);
+}
 
-    // Should match all 20 tokens (16 full + 4 partial)
+#[test]
+fn the_legacy_partial_arm_still_matches_when_opted_in() {
+    // The A/B arm (`ATLAS_PREFIX_PARTIAL_TAIL=1` in production): full match
+    // including the partial tail — the pre-fix behaviour, kept measurable.
+    let tree = RadixTree::with_partial_tail_sharing(true);
+    let tokens: Vec<u32> = (0..20).collect();
+    tree.insert(&tokens, &[10, 20], &[], 16, 0, 0);
+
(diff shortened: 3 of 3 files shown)


TheTom:
The default-off change is consistent with the no-COW risk, but the regression currently leaves two concrete holes.

First, `a_partial_tail_block_is_never_handed_to_a_second_writer` uses `RadixTree::new()`. That constructor reads `ATLAS_PREFIX_PARTIAL_TAIL`, so the test is not deterministic when a developer or CI environment has the A/B flag set. The new explicit constructor should pin the core regression with `with_partial_tail_sharing(false)`. If the production environment switch itself needs coverage, parse the flag through a small pure function and test unset, `0`, `1`, and unrelated values, or test it in an isolated subprocess rather than mutating process-global environment in the shared test binary.

Second, the default-off regression exercises only the stored `partial_suffix` source: insert 20 tokens, then look up the same 20. The guarded arm has a separate child-key prefix source. A change that disables stored suffix reuse but still hands out a child full block for a shorter sub-block lookup would pass the new regression. Add the default-off counterpart of `test_sub_block_match_via_child_key_prefix`: insert 35 tokens with blocks `[10, 20, 30]`, look up the first 22 tokens, and require exactly 16 matched tokens plus block `[10]`. Then keep the existing opted-in test requiring 22 and `[10, 20]`. That gives an explicit false/true pair for both sources.

I would also mutation-check the gate, not just run the suite:

1. Force `allow_partial_tail = true`; both default-off source tests must fail.
2. Move the gate inside only the `partial_suffix` branch; the child-key default-off test must fail.
3. Force `allow_partial_tail = false`; the opted-in source tests must fail.

The existing 86 passing radix tests show compatibility, but they do not supply these two missing failure-sensitive partitions.

tbraun96:
Triage against `avarok/main` @ `9e597531cf` (2026-09-20). Read by ref, not the working tree.

**Same gate, opposite defaults.** Main already carries a lever on exactly this arm: `crates/spark-runtime/src/radix_tree/inner.rs:189-206` reads `AVAROK_PREFIX_SUBBLOCK` (once, `OnceLock`, line 202) and gates the same `if remainder > 0 && remainder < block_size && matched_tokens == num_full_blocks * block_size` that this PR gates with `allow_partial_tail`. It landed in #797 (`0a56d8f8b6`, as `ATLAS_PREFIX_SUBBLOCK`, renamed by #1101) as a *bisection* lever for the LongCat C=1 corruption and defaults **ON** (`!= Ok("0")`). This PR's `ATLAS_PREFIX_PARTIAL_TAIL` defaults **OFF** (`== Ok("1")`). Each lever has exactly one reader, so there is no two-reader split today — but a naive rebase would put two gates with different defaults on one `if`, which is the trap. The rebase should collapse to ONE lever: keep the existing name `AVAROK_PREFIX_SUBBLOCK` (docs/ROBUSTNESS.md:1507 and the #936 arm ladder already reference it), flip its default, and have `new()` route through `with_partial_tail_sharing(bool)`.

**Not on main, not superseded.** Two-tree diff of this PR's three files vs main shows the constructor and the default-off regression tests absent. #797's lever was added because the arm was "a reasonable suspect for foreign context" and then cleared as the cause of that C=1 bug; ROBUSTNESS.md:1505-1508 records `AVAROK_PREFIX_SUBBLOCK=0` as byte-identical on 251 sequential samples. Neither touches the defect this PR describes, which only shows under concurrency.

**The second-writer condition is real on main, by construction.** With the arm ON, `walk` returns a non-block-aligned `matched` and the donor's physical block; `prefix_lookup.rs:121` sets `seq.cached_prefix_tokens = matched` and `:138-139` pushes the donor block into the requester's `block_table`. The write floor in `forward_layers.rs:146-152` only protects rows *below* `cached_prefix_tokens`; rows at and past `matched` are written normally, and `impl_b1.rs:76-80` places position `pos` at `block_table[pos / bs]`, offset `pos % bs` — i.e. into the donor's block. There is no copy-on-write anywhere in `spark-model`/`spark-runtime` (grepped). The C=1 null in ROBUSTNESS.md is consistent with this rather than contradicting it: at temperature 0 with one stream at a time, the requester's continuation writes the same bytes the donor wrote, so the tear is invisible — which is exactly the "lockstep is benign" mechanism in this PR's description.

**One claim in the PR does not hold on main today.** The refcount hole ("the eventual free underflows the donor's count"): the KV-pool ref is taken per matched block including the partial one (`prefix_lookup.rs:138` `kv_cache.inc_ref`) and released per `block_table` entry (`sequence.rs:194` `free_blocks`); the radix side floors symmetrically in both `inc_refs` (`inner.rs:254`) and `dec_refs` (`inner.rs:280`); and the cache holds its own pool ref on the partial block from insert (`inner.rs:429`). Both counters balance. Drop that paragraph on rebase — the sharing argument stands on its own.

**Recommendation on the default: OFF** (this PR's direction). Reason: handing a writable tail to a second sequence without COW is unsound regardless of whether a given benchmark exposes it, and the cost of OFF is bounded at `block_size - 1` tokens of prefill recompute per warm hit plus the loss of the exact-leaf shortcut for non-aligned exact-repeat prompts. Two caveats for the owner, since this changes a serve-path default: (1) `--enable-prefix-caching` is `default_value_t = false` (`serve_args.rs:664-665`), so the blast radius is prefix-caching serves — which includes the gate serves whose floors were cut with the cache on, so a warm-TTFT control leg is needed and this should ride as its own measurement-definition layer, not inside a kernel campaign; (2) if anyone wants the tokens back, the fix is the COW tail (fresh block + per-layer D2D of the matched rows, event-ordered against the donor's stream) — not turning the arm back on.

**Still open before this is mergeable:** @TheTom's two review points (pin the regression via `with_partial_tail_sharing(false)`, and add the default-off counterpart of `test_sub_block_match_via_child_key_prefix`) are unaddressed — the branch has a single commit predating the review — plus the rebase onto the existing lever above.

Not merging and not flipping anything here; the default is the owner's call.

tbraun96:
## Landed a different shape of this fix on `fix/prefix-partial-tail-aliasing` — recommendation: **close #692 as superseded**

Re-verified the whole chain on today's `main` (`546aa85b2b`, campaign 3 landed) and implemented the minimal form. Branch: [`fix/prefix-partial-tail-aliasing`](https://github.com/Avarok-Cybersecurity/atlas/tree/fix/prefix-partial-tail-aliasing) @ `829a4c94f70f36546e0284300874a5d07914e62e`. No PR yet — the owner is sequencing it into its own certification campaign (`crates/**` is PERF_PATHS).

**The diagnosis here is right and is what I acted on**, in particular the two things I could not have derived from source alone:

- that **arm 1 is unsound too** — *"arm 1's donor block is full, but the requester's tail writes clobber committed rows other sequences map read-only"*. That is what made the decision below correct rather than a half-fix.
- the **measured repro** — *"with identical prompts in flight the writers stay byte-identical in lockstep and the tear is BENIGN — which is why serial decode looked clean — but any desync (the batched verify's band-dependent rounding was the trigger) makes the writes differ"*, C=4 going 0/4 with three sequences emitting identical wrong text. The benign-in-lockstep mechanism is the only thing that explains why every single-stream test passed, and it is now quoted in the module docs.

### What changed relative to this PR, and why

**#692 adds `ATLAS_PREFIX_PARTIAL_TAIL` (default OFF) gating the same `if` that `main`'s existing `AVAROK_PREFIX_SUBBLOCK` (default ON) already gates.** Landing both gives one branch two independent readers that can disagree — the exact "one flag, two implementations" defect this repo has already been bitten by. So instead:

**Flip the existing `AVAROK_PREFIX_SUBBLOCK` default to OFF. No new lever.**

I verified the concern you'd expect here: **does flipping the existing default cover BOTH arms?** Yes. On today's tree `inner.rs` has a single `if subblock_ok && remainder > 0 && remainder < block_size && matched_tokens == num_full_blocks * block_size {` whose body contains arm 1 (child-key prefix) and arm 2 (`partial_suffix`) as sibling branches. One predicate, both arms. Pinned by a regression test per arm, not by reading.

Other differences from #692:

- **The lever now parses strictly.** Unset is OFF; only an explicit `1` arms it; anything else is `tracing::error!`d and refused. Under the old `!= Ok("0")` rule `AVAROK_PREFIX_SUBBLOCK=true` silently read as ON — a typo must not arm an unsound path.
- **Test seam instead of a public constructor.** #692 adds `RadixTree::with_partial_tail_sharing(bool)` because the env read is process-global. Agreed on the problem; I resolved the lever once into a `RadixTreeInner` field that the walk reads, and tests set that field on their own tree. Same determinism, no test-only constructor in the production API.
- **Insert side deliberately untouched**, so `AVAROK_PREFIX_SUBBLOCK=1` restores the old behaviour *completely* rather than half of it — otherwise the #936 arm-ladder A/B stops meaning anything.

### Your refcount claim: refuted on today's tree — no fix applied

> *"`inc_refs` walks full-block NODES only, so the matched partial block enters the requester's `block_table` with no ref taken — the eventual `free_sequence` underflows the donor's count."*

The premise is correct — `inner.rs` `inc_refs` computes `num_blocks = num_matched / block_size` and walks only those nodes, so the partial block takes no *node* ref. But the conclusion doesn't follow, because the pool ref is not taken by `inc_refs`. It is taken by the adopting caller, over **every** entry of `matched_blocks` including the partial one:

- `prefill_b/prefix_lookup.rs:138-141` — `for &block_idx in &prefix_match.matched_blocks { kv_cache.inc_ref(block_idx); seq.block_table.push(block_idx); }`. Same loop at `prefill_a.rs:95-98` and `prefill_c.rs:175`. All three adoption sites.
- `sequence.rs` `free_sequence_dispatch` frees the whole `block_table` via `free_blocks` → `free_block` → `dec_ref`. Symmetric with the above.
- The node side is symmetric too: `inc_refs` and `dec_refs` *both* skip the partial, so nothing is over- or under-released there either.
- And the cache holds its own pool ref on the partial block — `insert` pushes `partial_block` onto `InsertAcquired::blocks`, which `block_mgmt::cache_acquires_refs` turns into a `kv_cache.inc_ref`. That has been true since `7444209045` (2026-08-08), i.e. it predates this PR, so this was not a "main has moved" situation.

Covered by `tests::basic::test_partial_suffix_block_is_owned_and_released`. **Nothing to fix; the aliasing is the whole defect.** Flagging it in case the claim gets re-derived later from the same `inc_refs` read.

### What is on the branch

- `radix_tree::partial_tail` — single reader of the lever, strict parse, and the full mechanism write-up (both arms, the lockstep-benign explanation, the cost of OFF, the COW follow-up).
- `tests::partial_tail` — `a_partial_tail_block_is_never_handed_to_a_second_writer` (your name, arm 2), `a_committed_full_block_is_never_handed_out_as_a_second_writers_tail` (arm 1), plus two controls: full-block reuse is untouched by the default, and the armed lever still reproduces the legacy non-aligned match on the identical fixtures.
- The seven legacy sub-block tests opt in explicitly via the seam — including the two (`..._no_match_different_suffix`, `..._not_matched_for_full_block_request`) that would otherwise have gone *vacuously* green under the new default while appearing to still test the refusal.

**Negative control, run:** reverting the default to ON (`None => true` in `enabled_from`) makes both defect tests fail **by assertion**, not by compile error — `matched_tokens` 20 vs 16 and 22 vs 16 — while both control tests stay green. Restored; 334/334 `spark-runtime`, 1349 `spark-model`, 2635 `spark-server` green, `cargo check --workspace --all-targets` / fmt / clippy clean. No GPU work run.

### One measured counterpoint worth having on the record

`docs/ROBUSTNESS.md` (#936 arm ladder) records an arm with `AVAROK_PREFIX_SUBBLOCK=0`, lever verified armed in `/proc/PID/environ`: output **byte-identical to baseline, 0 of 251 samples moved**. So on that agentic shard the cost of OFF measured as zero output change. It doesn't bound the TTFT cost — that still needs the campaign — but it does say the recompute is not moving answers there.

### Recommendation

**Close #692 as superseded**, keeping #1193 open until the branch lands. The fix is the same behaviour by a smaller change, and the reasoning in this PR — especially arm 1 and the lockstep-benign repro — is carried into the code and this branch's commit message. If you'd rather keep it open until the campaign certifies, that works too; it just must not land alongside the branch, since together they would re-create the two-levers-one-branch problem this avoids.

tbraun96:
Closing as **superseded** by `fix/prefix-partial-tail-aliasing` (`829a4c94f7`), which carries this diagnosis into the code.

What landed instead: **flip the existing `AVAROK_PREFIX_SUBBLOCK` default to OFF** rather than add `ATLAS_PREFIX_PARTIAL_TAIL`. Two levers over the same `if` would be the one-flag-two-implementations defect, and the two branches must not land together for exactly that reason.

Your diagnosis is carried over in full, with credit — including the **arm-1** finding (a full donor block is unsound too: the requester's tail writes clobber committed rows other sequences map read-only) and the measured **C=4 0/4, three sequences emitting identical wrong text** repro, which is stronger evidence than the static analysis in #1193 and is quoted in the commit message. Both arms are covered by the single flip: they are sibling branches inside one predicate, and there is now a regression test pinning each.

One correction, established against today's `main`: the **refcount-underflow claim does not hold**. The premise is right — `inc_refs` computes `num_blocks = num_matched / block_size`, so the partial block takes no *node* ref. But the *pool* ref is not taken by `inc_refs`; it is taken by the adopting caller over every entry of `matched_blocks` including the partial one (`prefill_b/prefix_lookup.rs:138-141`, `prefill_a.rs:95-98`, `prefill_c.rs:175`) and freed symmetrically in `sequence.rs`. The cache also holds its own pool ref via `cache_acquires_refs`, in place since `7444209045` (2026-08-08) — so this predates the PR rather than being drift.

#1193 stays open until the fix lands. Thank you for finding this — the benign-in-lockstep detail is why it hid for so long.

PR #693, opened 2026-08-21
dflash: drafter ctx capture shifted by cached-prefix length on every prefix-cache hit

rsafier:
Standalone against `main` — two files in `spark-model`, three call sites, no behaviour change on any cold prompt.

**The DFlash drafter's prompt-hidden capture is written at the wrong base on every prefix-cache hit: shifted up by the cached-prefix length.**

## The defect

`try_dflash_prefill_capture_layer` copies buffer row `t` (position `processing_start + t`) into the drafter's ctx accumulator at `chunk_start + t`. Both call sites pass `layer_kv_write_start` as `chunk_start`, and `prefill_a` passes the same value to `update_dflash_ctx_len_after_prefill`.

On a cold prompt the KV write floor and the processing start coincide — which is how the floor ended up there and why nothing ever caught it. On a **prefix-cache hit** they diverge: the forward recomputes from position 0 (no SSM snapshot → full recompute) while KV writes start at the matched boundary. Every captured row then lands `matched_tokens` slots high:

* ctx rows `[0, matched)` keep whatever the drafter pool slot's **previous occupant** left there,
* the copied rows sit at the wrong positions relative to the per-slot position seeding,
* `ctx_len` advances past the prompt's end (`matched + proc_count` on a 42-token prompt claims 74 rows).

## Why the obvious test can't see it

At C=1 on a repeat prompt, the stale rows are the identical previous request's — **accidentally correct content**. Measured live (qwen3.8-27B-NVFP4 + DFlash2 drafter, GB10, 2026-08-21): cold + two cache-hit repeats ran byte-identical completions at 88.7% acceptance *with the bug live*. It only degrades when the pool slot's previous occupant was a different sequence — i.e. under concurrency, which is exactly the agentic shared-system-prompt shape.

With the fix, the same C=1 cache-hit measured **88.7% → 90.9% acceptance**, same box, same binary otherwise.

## The change

Both capture calls now pass the chunk's processing start (`effective_seq_len_start` in `prefill_b::forward_layers`, `seq_len_start` in `prefill_a`), which is also correct under a Marconi skip, where buffer row 0 is position `snap_tok`. `prefill_b::finalize_last`'s ctx_len update already used chunk-positional coordinates and is untouched. Cold-prompt behaviour is byte-identical by construction — the two values are equal there.

Found while root-causing a prefix-cache + concurrency output derailment; this is one of that hunt's independently-verified defects (the radix partial-tail fix in #692 is another; the remaining scheduler-side commit desync is still open and tracked on our side). Verified building on `main`'s base; fmt clean.

### DIFF
--- a/crates/spark-model/src/model/trait_impl/prefill_a.rs
+++ b/crates/spark-model/src/model/trait_impl/prefill_a.rs
@@ -420,15 +420,15 @@ impl TransformerModel {
                 .map_err(|e| anyhow::anyhow!("Prefill layer {i} failed: {e}"))?;
             // DFlash prefill capture: writes layer i's hidden output for
             // all `proc_count` tokens into the seq's accumulator at slots
-            // [layer_kv_write_start .. layer_kv_write_start + proc_count].
-            // No-op when DFlash is disabled.
-            self.try_dflash_prefill_capture_layer(
-                seq,
-                i,
-                layer_kv_write_start,
-                proc_count,
-                stream,
-            )?;
+            // [seq_len_start .. seq_len_start + proc_count]. The base is the
+            // chunk's PROCESSING start (absolute position of buffer row 0),
+            // NOT `layer_kv_write_start` — on a prefix-cache hit the KV
+            // write floor sits at the matched boundary while the forward
+            // recomputes from 0, and using the floor here shifted every
+            // captured row by the cached-prefix length (see the twin call in
+            // `prefill_b::forward_layers` for the full failure). No-op when
+            // DFlash is disabled.
+            self.try_dflash_prefill_capture_layer(seq, i, seq_len_start, proc_count, stream)?;
 
             // MLA diagnostic: dump per-layer hidden state norm (once per model).
             // Per-model latch (see `ModelStats::dumped`) rather than a static: an
@@ -518,8 +518,11 @@ impl TransformerModel {
         self.prefill_save_snapshot_with_vision_gate(tokens, seq, &mut kv_cache, bs, stream);
 
         // DFlash: advance the seq's `ctx_len` to span all just-prefilled
-        // positions so the next propose() can read them.
-        self.update_dflash_ctx_len_after_prefill(seq, layer_kv_write_start, proc_count)?;
+        // positions so the next propose() can read them. Same base as the
+        // capture above: the PROCESSING start, not the KV write floor — the
+        // floor would claim `matched + proc_count` rows on a prefix-cache
+        // hit, past the prompt's end.
+        self.update_dflash_ctx_len_after_prefill(seq, seq_len_start, proc_count)?;
 
         Ok(self.decode_logits_ptr())
     }
--- a/crates/spark-model/src/model/trait_impl/prefill_b/forward_layers.rs
+++ b/crates/spark-model/src/model/trait_impl/prefill_b/forward_layers.rs
@@ -208,11 +208,23 @@ impl TransformerModel {
                 t_in_prefill += t.elapsed();
             }
             let t_df = host_timing.then(std::time::Instant::now);
-            // DFlash chunked-prefill capture.
+            // DFlash chunked-prefill capture. The destination base is the
+            // chunk's PROCESSING start — the absolute position of buffer row
+            // 0 — NOT `layer_kv_write_start`. The two coincide on a cold
+            // prompt, which is how the KV write floor ended up here, but on a
+            // prefix-cache hit the forward recomputes from position 0 while
+            // KV writes start at the matched boundary: passing the floor
+            // shifted EVERY captured row up by the cached-prefix length, left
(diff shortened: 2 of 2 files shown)


TheTom:
The processing-coordinate substitution is directionally correct, but this PR currently adds no regression test, and the reported cold/C=1 runs do not exercise the failure-sensitive state.

Before merge, I would require a deterministic capture test that fills each source hidden row with a distinct sentinel and pre-fills the destination accumulator with a different poison value. Exercise the real capture and ctx-length update path, then assert the exact populated rows, exact `ctx_len`, exact `ctx_positions`, and that the first row beyond `ctx_len` remains poison. At minimum, cover:

1. `prefill_a`, ordinary prefix hit without Marconi: prompt length 42, matched/KV floor 32, processing start 0, count 42. Rows `0..42` must contain source rows `0..42`, `ctx_len` must be 42, and row 42 must remain poison. Reverting either changed `prefill_a` argument to `layer_kv_write_start` must fail this test.
2. `prefill_b`, chunked ordinary prefix hit: use multiple chunks with a cache boundary inside one chunk. Assert that all captured chunks compose one contiguous prompt in absolute processing coordinates, with no initial gap, overwrite, or duplicated range. Reverting `effective_seq_len_start` to `layer_kv_write_start` must fail this test.
3. Capacity boundary: make the processed range cross `max_ctx_len` and prove capture, `ctx_len`, and `ctx_positions` clamp to the same last valid row.

The Marconi claim also needs a separate test or a code change. A new `DflashProposerState` starts with `ctx_len = 0`, an empty `ctx_positions`, and a zeroed accumulator. On a Marconi skip, `seq_len_start`/`effective_seq_len_start` can be nonzero, so this patch correctly places the rows that were processed, but rows before that start are not captured by this state. `update_dflash_ctx_len_after_prefill` then reports the entire range as populated and seeds positions `0..new_len`. I did not find a DFlash hidden-context restore for the skipped prefix. Add a Marconi test that poisons the full accumulator and requires every row below `ctx_len` to have been populated with the expected per-position sentinel. If Marconi and DFlash cannot coexist, enforce and test that exclusion instead of claiming the coordinate is sufficient.

Finally, the concurrency explanation should be proven with two distinct prompts or, more deterministically, with poisoned prior contents. Repeating the same prompt at C=1 is explicitly an oracle that can let the bug survive. Output equality and aggregate acceptance are useful system checks, but they should supplement the exact row-placement oracle above, not replace it.

tbraun96:
**Included in stack `verify-and-spec` → #870.**

**Why it earns a merge:** a spec-decode correctness fix, and it cherry-picks onto current `main` cleanly — the change is genuinely absent there (three of its six siblings in this batch turned out to be already-landed, so I checked each rather than assuming).

**Stacked with:** #745 (sliding-window per-token metadata) and #693/#742 — whichever of the pair this isn't. All three touch `crates/`, so the group re-opens all 11 gates and owes **one** campaign between them instead of one each.

**Verified:** `cargo check --workspace --all-targets` on the composed stack → EXIT=0, zero errors.

Your commit is cherry-picked with authorship preserved. Close this once #870 lands, or say the word if you would rather land it directly.

tbraun96:
**Carried by stack #885** — this PR's diff is hunk-identical to commit `d8778c441d` on `stack/certified-2026-09-04` (#877, layer #879's range): `fix(dflash): capture prompt hiddens at the chunk's PROCESSING start, not the KV write floor`, touching exactly this PR's two files (`prefill_a.rs`, `prefill_b/forward_layers.rs`). Merge-tree confirms: merging this PR into the stack changes nothing. Adding it to the stack's accounting as a ninth constituent — it will be closed with landing evidence when #885 merges, not before. If any hunk did not make it across, point at it and I will lift it onto a fresh branch.

tbraun96:
Closing: the fix is already on `main`, and merging this PR would change nothing.

**Evidence** — merged locally onto `main` at `68c393913502` and diffed the result against that same sha:

```
$ git merge --no-commit --no-ff pr693 && git diff --cached --name-status origin/main
(no output)
```

Zero paths changed. The content arrived with **`dbd79688f` — "verify-line: the 2026-09-04 stack re-merged onto main, one campaign (#1130)"**, which is the last commit to touch both of this PR's files. Every line this PR adds is on `main` today, including the comment that records the reasoning:

> `// NOT `layer_kv_write_start` — on a prefix-cache hit the KV write floor sits at the matched boundary while the forward recomputes from 0, and using the floor here shifted every captured row by the cached-prefix length`

Note that `git cherry` still reports this commit as absent (`+ 8a72e2e7a`) — its patch-id no longer matches after the #1101 rebrand and the `trait_impl` split moved the surrounding code. **Patch-id is the wrong instrument here; the three-way merge result is the right one**, and it finds nothing left to apply.

Thank you for the diagnosis — the twin-call note is what makes the fix legible on `main`. Nothing is lost by closing, and reopening restores it if this reading is wrong.

PR #695, opened 2026-08-21
atlas-core: validate GGUF MoE geometry

TheTom:
## Summary

`moe_without_used_count_errors` did not prove that `expert_used_count` was required. Its fixture also omitted `expert_feed_forward_length`, and it accepted any error. Defaulting the missing used count to one left the test green because parsing failed later on the missing expert width.

The audit also found a production boundary gap. A `qwen3moe` GGUF with missing or zero `expert_count` returned a successful `qwen3_5_moe` config with zero experts. Routed top-K zero, routed top-K greater than the expert count, and per-expert width zero also survived GGUF parsing.

This change starts from complete valid MoE metadata and removes each required expert dimension one at a time, with a key-specific error oracle. It also rejects invalid expert geometry at the GGUF boundary before loader allocation or kernel setup.

## Production consequence

These fields are not inert parser metadata. `num_experts` selects expert loops and weight loading, `num_experts_per_tok` sizes routing scratch and drives top-K kernels, and `moe_intermediate_size` controls expert weight and activation shapes. Allowing malformed geometry pushes a configuration error into later model construction, or permits a zero routed top-K that the later `MoeLayer::new` condition does not reject.

## Failure proof

Before the fix:

- Replacing the required `expert_used_count` lookup with a default of one left the original test passing.
- Missing or zero `expert_count` parsed successfully.
- `expert_used_count` values 0 and 129 parsed successfully with 128 experts.
- `expert_feed_forward_length = 0` parsed successfully.

After the fix:

- The exact default-to-one mutant fails `qwen3_moe_requires_each_expert_dimension` because the isolated used-count partition returns success and reaches `unwrap_err`.
- Removing the expert-count guard fails the zero-count partition.
- Removing the routed top-K guard fails the zero-top-K partition.
- Removing the expert-width guard fails the zero-width partition.

Each mutant was applied and run independently, then restored.

## Test plan

- [x] `typos`
- [x] `git diff --check`
- [ ] `ATLAS_SKIP_BUILD=1 cargo clippy --workspace --tests --all-features -- -Dwarnings` (owned by Linux CI; the local Apple workspace reaches Linux-only paths outside this diff)
- [ ] Tested against a real model / hardware if the change affects valid inference behavior (not required for this malformed-metadata rejection path)

The production change is limited to GGUF MoE metadata validation. Valid Qwen3 MoE metadata maps identically, and the existing positive expert-field test still passes. This does not load a real GGUF, allocate expert weights, launch MoE kernels, or compare generated output. CI and any repository benchmark policy remain required before review-ready status.

## CLA

### DIFF
--- a/crates/atlas-core/src/config/gguf.rs
+++ b/crates/atlas-core/src/config/gguf.rs
@@ -153,10 +153,16 @@ pub fn config_from_gguf(inputs: &GgufConfigInputs) -> Result<ModelConfig> {
     let tie_word_embeddings = !inputs.has_output_weight;
 
     // ── MoE (only for MoE arches) ──
-    let num_experts = meta
-        .get_u64(&k("expert_count"))
-        .map(|v| v as usize)
-        .unwrap_or(0);
+    let num_experts = if arch == "qwen3moe" {
+        req_u64("expert_count")? as usize
+    } else {
+        meta.get_u64(&k("expert_count"))
+            .map(|v| v as usize)
+            .unwrap_or(0)
+    };
+    if arch == "qwen3moe" && num_experts == 0 {
+        bail!("GGUF metadata key '{arch}.expert_count' must be greater than zero");
+    }
 
     let mut body: Map<String, Value> = json!({
         "hidden_size": hidden_size,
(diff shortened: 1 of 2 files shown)


tbraun96:

PR #698, opened 2026-08-21
kernels: parameterise the prefill tile constants on BC (4 latent OOB bugs)

tbraun96:
`inferspark_prefill` has only ever been built at `BR == BC == 32`, and four places
hardcode a value that is really a function of `BC`. Each is correct at `BR == BC`
and goes out of bounds the moment they differ.

| # | Site | Hardcoded | Actually | Effect at BR=32/BC=16 |
|---|---|---|---|---|
| 1 | 3× K/V staging loops | `TILE_CHUNKS` = `BR*(HDIM/8)` | `BC*(HDIM/8)` | walks rows 16–31 of a 16-row tile — **OOB shared write** (confirmed with compute-sanitizer) |
| 2 | PV k-step loop | `ks < 2` | `ks < BC/16` | reads a nonexistent second half of `smem_V` |
| 3 | Score accumulator | `float acc_s[4][4]` | `acc_s[BC/8][4]` | — |
| 4 | 8× n-tile loops | `nt < 4` | `nt < BC/8` | 32 columns written into a 24-wide `smem_P` row |

The fourth staging loop fills `smem_Q`, which genuinely **is** `BR` rows, and keeps
its bound.

## Provably no change at the shipped shape

Compiled before and after and diffed the SASS:

```
8823 instructions — BYTE-IDENTICAL
124 regs / 90,112 B  and  122 regs / 70,400 B   (both entry points, unchanged)
```

## Why now

Gemma-4's five **512-wide** global attention layers don't use this kernel — they
fall through to `inferspark_prefill_512`, **114 lines of scalar code with no tensor
cores**:

| kernel | per layer | layers | total |
|---|---:|---:|---:|
| `flash_attn_512_scalar` | **3,296.9 ms** | 5 | **16,484 ms** |
| `flash_attn_64` (this kernel) | 2.0 ms | 25 | 49.7 ms |

That's 91% of an 18,078 ms prefill on a 4096-token prompt.

This kernel **does** fit at HDIM=512 with BR=32/BC=16 — **84,992 B** smem against a
101,376 B cap — and standalone it scores **cosine 0.999998** against the scalar
reference at **64.7×** the speed. These four constants were what stood in the way.

The 512 entry point is a follow-up; this lands the parameterisation alone so the
SASS-identity argument is reviewable on its own.

### DIFF
--- a/kernels/gb10/common/inferspark_prefill.cu
+++ b/kernels/gb10/common/inferspark_prefill.cu
@@ -42,6 +42,12 @@
 
 // Number of 16-byte (8-element) chunks per tile: 32 rows * (256/8) = 32*32 = 1024
 #define TILE_CHUNKS (BR * (HDIM / 8))
+// K and V tiles are BC rows, not BR. Identical while BR == BC (the only shape
+// this file has ever been built at), so the three K/V load loops below used the
+// BR-sized bound and were never wrong. At BC != BR that bound walks rows
+// BC..BR-1 of a BC-row tile and writes OUT OF BOUNDS in shared memory —
+// confirmed with compute-sanitizer at BR=32/BC=16.
+#define TILE_CHUNKS_KV_TILE (BC * (HDIM / 8))
 
 // SCALE/gfx1151: RDNA3.5 hard 64 KB/workgroup LDS cap. This file is
 // COMPILE-ONLY on AMD (non-paged contiguous prefill — not dispatched for
@@ -160,7 +166,7 @@ extern "C" __global__ void inferspark_prefill(
 
         // K[kv_block_lo] tile (same commit group — no extra sync)
         if (num_kv_blocks > 0) {
-            for (unsigned int idx = tid; idx < TILE_CHUNKS; idx += 128) {
+            for (unsigned int idx = tid; idx < TILE_CHUNKS_KV_TILE; idx += 128) {
                 unsigned int row = idx / chunks_per_row;
                 unsigned int chunk = idx % chunks_per_row;
                 unsigned int col = chunk * 8;
@@ -190,7 +196,7 @@ extern "C" __global__ void inferspark_prefill(
         // === Start async V load into smem_V (overlaps with QK^T below) ===
         {
             const unsigned int chunks_per_row = HDIM / 8;
-            for (unsigned int idx = tid; idx < TILE_CHUNKS; idx += 128) {
+            for (unsigned int idx = tid; idx < TILE_CHUNKS_KV_TILE; idx += 128) {
                 unsigned int row = idx / chunks_per_row;
                 unsigned int chunk = idx % chunks_per_row;
                 unsigned int col = chunk * 8;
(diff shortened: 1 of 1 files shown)


TheTom:
Constructive testing pass: the direction is sound, but the same translation unit still has one applicable BC hardcode. `inferspark_prefill_64` keeps `for (unsigned int ks = 0; ks < 2; ks++)` in its PV phase while `smem_V64` has only `BC` rows. Because `BC` is shared by both exported kernels, a `BC=16` build still executes `ks=1` in that sibling entry point and addresses rows 16 through 31. Its `float acc_s[4][4]` is also still fixed at four n-tiles. That array is only overallocated at BC=16, so it is not itself an OOB defect there, but it leaves the parameterization incomplete.

Before merge, I recommend:

1. Change the remaining `inferspark_prefill_64` PV bound to `BC / 16`, and either parameterize its score array as `BC / 8` or explicitly document why that entry point is intentionally fixed. Add compile-time guards for the real contract, at least `BC >= 16` and `BC % 16 == 0`, so future values cannot silently truncate these loop counts.
2. Run both exported entry points from a BC=16 build under compute-sanitizer. Use sequence lengths 15, 16, 17, 31, 32, and 33 so the matrix covers partial tiles, a full tile, multiple KV blocks, and the next-K preload path. For the 64-row entry point, also include 63, 64, and 65 to cover its query tail.
3. Compare each case against the scalar attention reference with deterministic inputs. Include causal full attention, causal sliding-window attention, and non-causal mode because the same corrected score-tile loops feed all three masks.
4. Keep a negative control: restore each old hardcode family in isolation and show the BC=16 gate fails through sanitizer or numerical mismatch. This proves the test observes the repaired bounds rather than only demonstrating that the final kernel can run.
5. Preserve the BC=32 compatibility evidence for both symbols with the exact build command plus SASS hashes or a checked script. The current byte-identity result is valuable, but CI only compiles the shipped BC=32 configuration and does not execute the new BC=16 shape.

A checked-in BC=16 microtest or gate is important here because the follow-up is the first expected consumer of this latent path. Without one, the default kernel compile job can stay green while a later edit restores the same BR/BC confusion.

tbraun96:

Every commit here cherry-picks empty onto current `main`, and the artifact confirms it: `kernels/gb10/common/inferspark_prefill.cu` carries the `BC` parameterisation at line 39 (`#ifndef BC`), with the shared-memory reasoning you wrote preserved above it at lines 35-36:

```
// forking 500 lines. Gemma-4's global layers are HDIM=512, which at BC=32 needs
// 132.8 KB of shared memory (cap is 101,376 B); BC=16 fits in 84,992 B.
```

Found by a backlog-wide containment scan. Worth noting the scan's first version got this class **wrong** — it cherry-picked commit ranges, and `git cherry-pick A..B` stops at the first empty commit and reports "is now empty", which says nothing about the commits after it. That marked #702 as contained when it has four empty commits followed by one that conflicts. The scan now judges commit-by-commit and calls a PR contained only if **every** commit is empty; each close is still confirmed against a named artifact before anyone acts on it.

PR #700, opened 2026-08-21
perf(gemma4): tensor-core 512-wide prefill attention — 4096-tok prefill 20.6 s → 3.8 s

tbraun96:
Gemma-4-26B-A4B has **5 global (512-wide)** and **25 sliding (256-wide)** attention
layers. The sliding ones take the tensor-core kernel; the global ones fell through
to `inferspark_prefill_512` — **114 lines of scalar code, no tensor cores, no
shared-memory tiling**:

| kernel | per layer | layers | total |
|---|---:|---:|---:|
| `flash_attn_512_scalar` | **3,296.9 ms** | 5 | **16,484 ms** |
| tensor-core path | 2.0 ms | 25 | 49.7 ms |

**Targets #698** (the parameterisation that makes a second shape possible). Rebase
to `main` once that lands.

## Measured end-to-end

`bg-digitalservices/Gemma-4-26B-A4B-it-NVFP4A16`, 4096-token prefill, **one binary**,
both arms back-to-back, `ATLAS_ATTN_512_TC` the only variable:

| | scalar | **tensor-core** | Δ |
|---|---:|---:|---:|
| **wall (request → first token)** | 20,583 ms | **3,826 ms** | **−81.4%, 5.4×** |
| attention, all 30 layers | 16,884.4 ms | **123.5 ms** | **−99.3%, 137×** |
| serve errors | 0 | 0 | — |

Standalone against the kernel it replaces (S=1024, 4 q-heads, 2 kv-heads, causal):
**cosine 0.999998**, max abs 0.0039, **64.7×**.

## It is 26 lines, not a fork

```c
#define HDIM 512
#define BC 16
#define ATLAS_PREFILL_ENTRY inferspark_prefill_512tc
#define ATLAS_SKIP_PREFILL_64 1
#include "../../common/inferspark_prefill.cu"
```

## Correctness

**Gate C2 passes on both arms** — `tool_calls=3/3`, `coherent=3/3`, `degenerate=0`.
Output is not bit-identical and cannot be (tensor cores accumulate in a different
order), but every temp-0 prompt stayed coherent and correct on both arms, and the
standalone cosine bounds the difference.

## Shape

`BC=16`, not 32: at HDIM=512 a 32-row K/V tile needs **132.8 KB** of shared memory
against a **101,376 B** cap; BC=16 fits in **84,992 B**. `BR` stays 32 — the warp
mapping (`(warp_id & 1) * 16`, `warp_id < 2`) splits a 32-row Q tile across two warp
pairs. The `_64` variant is skipped here: it needs 120,064 B at this shape.

## ★ One reader for name and BR

The two 512 kernels want **different grids** (scalar BR=16, TC BR=32), and the name
is chosen in `qwen3_attention::init` while the grid is built in
`ops::prefill_attention`. A mismatch there **does not fail** — it silently computes
the wrong q-tiles. Both now read `ops::wide_prefill_kernel()`, which returns the pair
together. `ATLAS_ATTN_512_TC=0` selects the scalar reference, kept reachable because
it is the oracle this was validated against.

## Follow-up

`kernels/gb10/gemma-4-26b-a4b/BENCH.toml` commits a p90 ceiling of **24,490.81 ms**
and says in its own note that the number is a defect to ratchet down rather than
accept. That ratchet is now due.

### DIFF
--- a/crates/spark-model/src/layers/ops/prefill_attn_main_a.rs
+++ b/crates/spark-model/src/layers/ops/prefill_attn_main_a.rs
@@ -23,6 +23,23 @@ use super::*;
 ///         K/V [batch, seq_len, num_kv_heads, head_dim] BF16
 ///         O [batch, seq_len, num_q_heads, head_dim] BF16
 #[allow(clippy::too_many_arguments)]
+
+/// The HDIM>256 prefill kernel: its module/entry name and the BR its grid must
+/// be built for. ONE reader, because the name is chosen in `qwen3_attention::init`
+/// and the grid here, and a mismatch is silent.
+///
+/// Default is the tensor-core instantiation (`BR=32`). `ATLAS_ATTN_512_TC=0`
+/// selects the scalar reference (`BR=16`) — kept reachable because it is the
+/// oracle the TC path was validated against (cosine 0.999998, 64.7x faster on
+/// S=1024/4q/2kv/causal).
+pub fn wide_prefill_kernel() -> (&'static str, u32) {
+    if std::env::var("ATLAS_ATTN_512_TC").ok().as_deref() == Some("0") {
+        ("inferspark_prefill_512", 16)
+    } else {
+        ("inferspark_prefill_512tc", 32)
+    }
+}
+
 pub fn prefill_attention(
     gpu: &dyn GpuBackend,
     kernel: KernelHandle,
@@ -40,8 +57,15 @@ pub fn prefill_attention(
     sliding_window: u32, // 0 = no sliding limit; >0 = mask keys where q - k >= window
     stream: u64,
 ) -> Result<()> {
-    // BR=16 for HDIM=512 (Gemma-4 full attention), BR=32 otherwise
-    let br = if head_dim > 256 { 16u32 } else { 32u32 };
+    // ★ SSOT: the kernel NAME and its BR must agree, and they are chosen in two
+    // different files — `init.rs` resolves the handle, this launches it. A grid
+    // computed from the wrong BR does not fail, it silently computes the wrong
+    // q-tiles. `wide_prefill_kernel()` is the one reader of that decision.
+    let br = if head_dim > 256 {
+        wide_prefill_kernel().1
+    } else {
+        32u32
+    };
     KernelLaunch::new(gpu, kernel)
         .grid([num_q_heads, div_ceil(seq_len, br), batch])
         .block([128, 1, 1])
--- a/crates/spark-model/src/layers/qwen3_attention/init.rs
+++ b/crates/spark-model/src/layers/qwen3_attention/init.rs
@@ -477,12 +477,13 @@ impl Qwen3AttentionLayer {
                 "dense_gemm_bf16_pipelined",
             ),
             prefill_attn_k: gpu.kernel("inferspark_prefill", "inferspark_prefill")?,
-            prefill_attn_512_k: gate(
-                probes.wide_head_dim,
-                gpu,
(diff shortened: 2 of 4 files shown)


TheTom:
Testing review: the Gemma measurements are useful, but I would not treat them as sufficient for the current dispatch change yet.

The first negative control should be cross-model. `ArchProbes::wide_head_dim` is also true for DeepSeek V4 through `kv_lora_rank`, and `init.rs` now asks `wide_prefill_kernel()` for its module whenever that probe is true. With the environment unset, the helper selects `inferspark_prefill_512tc`, but that module exists only in the Gemma target; DeepSeek registers its sink-aware scalar `inferspark_prefill_512`. Please boot every target that can set `wide_head_dim`, with both the default and scalar selection, and run a short real prefill. For DeepSeek specifically, verify that the kernel audit has no unexpected missing lookup and compare the sink-aware prefill output against its existing reference. A test that forces the Gemma TC choice for a non-Gemma target should fail.

The name and BR are not actually resolved once. The helper is called during handle lookup and again during launch-grid construction. If the environment changes between those points, the TC handle can be launched with BR=16 or the scalar handle with BR=32. Please resolve and store the pair once, then test both selections. As negative controls, independently swap only the name and only the BR and show that the test fails. A process-level test that changes `ATLAS_ATTN_512_TC` after layer construction should not change the launch geometry.

For the new HDIM=512, BC=16 kernel, I would add deterministic scalar or CPU-reference comparisons at sequence lengths 1, 15, 16, 17, 31, 32, 33, 1023, 1024, and 1025, plus batch 2. These exercise both K/V tile and Q-grid tails. Run compute-sanitizer memcheck on those cases. Reapply the old BR-sized K/V loop bound and a BR=16 grid as separate negative controls so each is proven detectable. The current S=1024 result does not exercise either boundary.

Finally, on the real Gemma model, record that all five 512-wide global layers select `inferspark_prefill_512tc` while the 25 sliding layers remain on their existing path. Compare deterministic logits or layer outputs between scalar and TC arms at a short boundary prompt and at 4096 tokens, with a stated tolerance. The 3/3 semantic tripwire is useful, but it is too small to detect a localized attention error by itself. The benchmark ceiling mentioned in the PR should also be ratcheted in this change or tracked by a concrete follow-up.

Current CI also reports the source-file size gate failing, so the PR is not review-ready yet even though workspace tests and kernel compilation are green.

tbraun96:
Closing as **superseded by #702**, not rejected — the work lands, just from the other PR.

`git merge-base --is-ancestor` confirms this branch is an **ancestor of #702**, so #702 contains every commit here plus nine more files. Cherry-picking this onto #702 returns `empty commit set passed` — there is nothing left to apply.

#702 also carries a **full 10-gate campaign** at sha `ba32287a0d` (agentic-webserver, bfcl-subset, bfcl-subset-echolp, concurrency-sweep, decode-floor, ssm-state-poisoning, ttft-cold, ttft-warm, video-fidelity, vision-fidelity). So the gemma4 prefill work is effectively **already stacked** — three PRs' worth of change certified once, which is exactly the shape this backlog pass is trying to produce elsewhere by hand.

Found while grouping the gemma4 perf PRs (#668, #700, #702, #706). #706 is **not** contained and still applies cleanly on top; it stays open.

One caveat on #702 itself: it is 77 commits behind `main` and its records name `ba32287a0d` while its head is `d21a0ccc6c`, so a rebase will likely invalidate the campaign and need a re-run. That is a #702 problem, not a reason to keep this one open.

PR #673, opened 2026-08-21
bench: gate Gemma-4 cold TTFT — and record that its 4096-token prefill takes 23 s

tbraun96:
`gemma-4-26b-a4b` had no `BENCH.toml`, so #668's −12.8% cold-TTFT win (5705.61 →
4974.56 ms) had nothing holding it. **Targets #668**; rebase to `main` once that
and #660 land.

Recorded **PASS** at this head: median 1577.1 ms (−0.2%), p90 23.5 s (+0.7%).

| | baseline (gate harness, dgx1) | ceiling |
|---|---:|---:|
| median | 1580.547 ms | 1627.96 (×1.03) |
| p90 | 23324.582 ms | 24490.81 (×1.05) |

`default` is unset — `qwen3.6-35b-a3b` claims it for this gate on gb10. The gate's
numbers are **not** the `cold_ttft.py` numbers (1580.5 vs 4974.6 ms): different
harness, different prompt ladder, not interchangeable.

## ★ The p90 is pathological, and enshrined deliberately

| prompt | median | scaling |
|---|---:|---|
| 256 tok | 261.4 ms | — |
| 1024 tok | 1580.5 ms | 6.0× for 4× the tokens |
| **4096 tok** | **23324.6 ms** | **14.7× for 4× the tokens** |

4096 tokens is 16× the 256-token prompt and costs **89×** the time. **23 seconds of
cold TTFT is a defect, not a baseline.**

It's committed at the measured value so a regression *from here* is visible — a gate
that refuses to record a bad number protects nothing — and it should be **ratcheted
down** once the cause is fixed. The run was clean: **0 serve errors**, all three legs
sampled. This is real work, not a harness artifact.

## ★ No warm gate, and none is possible today

This model **cannot use prefix caching at all**. With `enable_prefix_caching`,
partial hits take the chunked-prefill path, which refuses on this architecture:

```
Gemma-4 HDIM=512 chunked prefill only supports BF16 KV cache (layer 5, kv_dtype=Fp8)
```
(`qwen3_attention/prefill/paged_attn.rs`)

Measured: **46 serve errors** with the flag, **0** without — and **2 of 3 warm legs
returned no samples** while the surviving 1024-tok leg read 74.7 ms. It looks **20×
faster** precisely because it mostly did not run. `median — ms / cached 0 tok` was the
only visible tell, and it reads like a win at a glance.

## Both point at the same path

`HDIM=512` attention. And both are **production** gaps, not benchmark artifacts:
[...]

PR #702, opened 2026-08-22
perf(gemma-4): tensor-core 512-wide prefill attention — 90 s of prefill across two models

tbraun96:
Gemma-4's **global (512-wide)** attention layers ran a **scalar reference kernel** —
114 lines, no tensor cores, no shared-memory tiling — while its **sliding (256-wide)**
layers took the tiled tensor-core path. Measured per layer:

| model | 512-wide layers | scalar | **TC** | ratio |
|---|---:|---:|---:|---:|
| gemma-4-26b-a4b | 5 | 3,296.9 ms | **14.7 ms** | 224× |
| gemma-4-31b | 10 | 7,340.2 ms | **29.9 ms** | 245× |

End-to-end, 4096-token prefill, one binary, arms back-to-back, `ATLAS_ATTN_512_TC`
the only variable:

| model | scalar | **TC** | Δ |
|---|---:|---:|---:|
| gemma-4-26b-a4b | 20,583 ms | **3,826 ms** | **−81.4%, 5.4×** |
| gemma-4-31b | 101,547 ms | **28,285 ms** | **−72.1%, 3.6×** |

`flash_attn_64` (the 25/50 sliding layers) is flat across both arms — the control.
0 serve errors on every run.

## How it was found

The profiler was lying. `aprof!("flash_attn_64", t0)` sat **after** an if/else running
two different kernels and labelled both with the second one's name, attributing
**16,484 ms to a kernel that used 49.7 ms** — off by 330×. Two root-cause hypotheses
were built on that reading and both were wrong. Fixing the label first is what made
the real cause visible.

## The fix is an instantiation, not a fork

```c
#define HDIM 512
#define BC 16
#define ATLAS_PREFILL_ENTRY inferspark_prefill_512tc
#define ATLAS_SKIP_PREFILL_64 1
#include "../../common/inferspark_prefill.cu"
```

`BC=16` because at HDIM=512 a 32-row K/V tile needs 132.8 KB against a 101,376 B cap;
BC=16 fits in 84,992 B. `BR` stays 32 — the warp mapping splits a 32-row Q tile across
two warp pairs.

Four constants in the shared kernel hardcoded values that are really functions of
`BC`, each correct only while `BR == BC == 32` and each an out-of-bounds access
otherwise (confirmed with compute-sanitizer). Parameterising them is **SASS-identical
at the shipped shape** — 8,823 instructions, byte-for-byte.

## Blast radius was checked, not assumed

`wide_head_dim` is `head_dim > 256 || kv_lora_rank > 0`, so **three** targets issue the
new lookup:

| target | why | remedy |
|---|---|---|
| gemma-4-26b-a4b | genuine 512-wide heads | TC instantiation |
| gemma-4-31b | genuine 512-wide heads | TC instantiation |
| **deepseek-v4-flash** | `kv_lora_rank = 512`, **not** wide heads | declared `expected_absent` |

DeepSeek deliberately keeps the scalar kernel: its 512 path is MLA with a per-head
sink, launched against a **12-parameter** kernel where the shared TC one takes 11 — and
a signature mismatch there makes `cuLaunchKernel` read past the caller's argument array
and segfault the host with no CUDA error.

Verifying only the 26B would have left **two models unable to serve** — `--check-kernels`
refuses a target where a lookup resolves to handle 0. The gates would not have caught
it; they run one checkpoint.

## Correctness

- Standalone vs the scalar reference (S=1024, causal): **cosine 0.999998**, max abs 0.0039
- **Gate C2 passes on both arms**: `tool_calls=3/3`, `coherent=3/3`, `degenerate=0`
- Not bit-identical, and cannot be — tensor cores accumulate in a different order

## Also in this chain

- Qwen3-VL MoE prefill copies (cold TTFT **−45.3%**) + its gate — *already merged as #660,
  carried here as history*
- Gemma-4 MoE prefill copies (cold **−12.8%**) + gate
- Gate ceiling **ratcheted 24,490 → 1,992 ms p90**, exactly as that gate's own note asked

Consolidates #664-adjacent attention work from #668, #673, #682, #698, #700 into one
campaign rather than five.

### DIFF
--- a/.benchmarks/agentic-webserver/2026-08-21-b1b601af7c.json
+++ b/.benchmarks/agentic-webserver/2026-08-21-b1b601af7c.json
@@ -1,0 +1,467 @@
+{
+  "schema": 1,
+  "benchmark_id": "agentic-webserver",
+  "benchmark_name": "Agentic Webserver Test",
+  "git_sha": "b1b601af7c",
+  "recorded_at": 1787329450,
+  "target_model": "Qwen/Qwen3.6-35B-A3B-FP8",
+  "params": {
+    "build_timeout_s": "600",
+    "command_timeout_s": "180",
+    "iterations": "10",
+    "max_tokens": "8192",
+    "max_turns": "40",
+    "request_timeout_s": "900",
+    "s_per_turn_budget": "8.5",
+    "serve_timeout_s": "30",
+    "wall_budget_s": "1800"
+  },
+  "command": [
+    "spark",
+    "benchmark",
+    "run",
+    "agentic-webserver",
+    "--param",
+    "build_timeout_s=600",
+    "--param",
+    "command_timeout_s=180",
+    "--param",
+    "iterations=10",
+    "--param",
+    "max_tokens=8192",
+    "--param",
+    "max_turns=40",
+    "--param",
+    "request_timeout_s=900",
+    "--param",
+    "s_per_turn_budget=8.5",
+    "--param",
+    "serve_timeout_s=30",
+    "--param",
+    "wall_budget_s=1800",
+    "--yes",
+    "--pull-request-gate"
+  ],
+  "served_by": "qwen3.6/qwen3.6-35b-a3b-fp8-bf16head",
+  "atlas_version": "1.0.0-beta-preview",
+  "hardware": {
+    "gpu": "NVIDIA GB10",
+    "driver": "580.126.09",
+    "sm_clock_mhz": 2392.0,
+    "source": "nvidia-smi"
+  },
+  "hardware_state": {
+    "sensitivity": "speed",
+    "before": {
+      "captured_at": 1787328634,
+      "machine": {
+        "hostname": "spark-256a",
+        "machine_id": "7af66f30966a49b6886e00e2fce4b42f",
+        "gpu": "NVIDIA GB10",
+        "driver": "580.126.09"
+      },
+      "gpu_temp_c": 54.0,
+      "chassis_temps_c": [
+        {
+          "name": "acpitz",
+          "temp_c": 61.2
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 54.9
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 61.2
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 54.4
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 54.3
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 56.8
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 56.3
+        }
+      ],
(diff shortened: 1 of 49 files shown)


tbraun96:
**This PR is already a stack**, and that is worth making explicit.

`git merge-base --is-ancestor` shows **#700 and #668 are both ancestors of this branch** — it contains their commits plus nine further files. I have closed both as superseded, pointing here. Cherry-picking either onto this head returns `empty commit set passed`.

It also carries a **full 10-gate campaign** at `ba32287a0d`: agentic-webserver, bfcl-subset, bfcl-subset-echolp, concurrency-sweep, decode-floor, ssm-state-poisoning, ttft-cold, ttft-warm, video-fidelity, vision-fidelity. Three PRs' worth of change, certified once — the exact shape I have been assembling by hand for the other groups (#868, #869, #870).

**What stands between this and merging**, and I would rather flag it than sit on it:

- The branch is **17 ahead / 77 behind** `main`.
- Its records name `ba32287a0d`; the head is `d21a0ccc6c`. A rebase onto current `main` will move the sha and pull in `PERF_PATHS` changes made since, which invalidates the campaign and needs a re-run.

**#706** (FFN transposed NVFP4 copies) is **not** contained here and cherry-picks cleanly on top of this branch. If you rebase and re-certify anyway, folding #706 in makes it one campaign for four PRs instead of two campaigns for two groups. Happy to do the rebase and the fold if you want it — say the word.

tbraun96:
Closing: the tensor-core 512-wide prefill landed with #721.

**Carrier:** `1dc3c0dcf` — *"perf(gemma-4, gdn): **tensor-core 512 attention**, FFN transposes, GDN double buffer, TMA plumbing (#721)"*, which is the last commit to touch every file in this PR's diff.

**Full-population check** (every added line of ≥30 chars, not a sample): **9 of 358 absent from `main`, 2%**. All nine are pre-rebrand identifiers, and `main` carries each under its new name:

| this PR | on `main` |
|---|---|
| `ATLAS_ATTN_512_TC` | `AVAROK_ATTN_512_TC` — `crates/spark-model/src/layers/ops/wide_prefill.rs:28` |
| `ATLAS_PREFILL_ENTRY` | `AVAROK_PREFILL_ENTRY` — `kernels/gb10/common/inferspark_prefill.cu:75-76` |
| `ATLAS_SKIP_PREFILL_64` | `AVAROK_SKIP_PREFILL_64` — 3 files |

`ATLAS_*` appears in **zero** files on `main` for all three; `AVAROK_*` in 1, 3 and 3. So the residue is the #1101 rename, not missing work — the same reason #558, #884, #504 and #777 all read as unmerged when they were not.

A note on method, since I got this wrong earlier today: a ten-line sample taken from the head of a sorted diff over-reports containment. This verdict is the full added-line population plus an inspection of what remains — the percentage says whether to look, the content says what it means.

PR #703, opened 2026-08-22
stream: give the SimHash semantic-loop guard the standard kill switch

rsafier:
The F4 SimHash guard was the only watchdog with no operator control: hardcoded one-strike at Jaccard 0.55 over a 16-sentence ring, and a fire **kills the stream mid-reply** (`guard_stop=simhash_semantic_loop`). Legitimate structured output crosses that bar easily — per-method docstrings, enumerations, boilerplate-heavy code all produce ≥0.55 bigram overlap between honest sentences.

Post-#699 (with genuinely degenerate output gone) its remaining fires skew false-positive: observed 2026-08-21 killing a healthy interactive session mid-answer. Before #699 most of its fires were masking the real defect.

`ATLAS_SIMHASH_LOOP=0` disables it — house convention, same shape as `ATLAS_TOOL_ENVELOPE_WATCHDOG`. **Default stays ON**; this adds the lever, not a policy change. The token-level loop watchdog is untouched and still guards true verbatim loops.

Verified with the switch off: a prompt mandating identical-form docstrings ran to a full 512-token coherent completion (that exact shape previously died at the guard), an identical-sentence litany streamed unimpeded, zero SimHash warnings in the serve log. Built + fmt clean on main's base.

### DIFF
--- a/crates/spark-server/src/api/chat_stream/handle_token.rs
+++ b/crates/spark-server/src/api/chat_stream/handle_token.rs
@@ -15,6 +15,15 @@ use super::super::sanitizer::sanitize_content_chunk;
 use super::super::stream_guards::{bump_f12_tool_call_count, check_loop_watchdog};
 use super::ctx::StreamCtx;
 use super::state::StreamState;
+
+/// `ATLAS_SIMHASH_LOOP=0` disables the F4 SimHash semantic-loop guard.
+/// Default ON — see the comment at the check site for why an operator would
+/// turn it off (one-strike near-duplicate detection kills streams over
+/// legitimately repetitive structured output).
+fn simhash_loop_enabled() -> bool {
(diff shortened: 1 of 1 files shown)


tbraun96:
Closed as **already landed** — not rejected. Apologies for closing before commenting.

Cherry-picking onto current `main` produces an **empty commit**. Confirmed independently: `ATLAS_SIMHASH_LOOP=0` is documented and read at `crates/spark-server/src/api/chat_stream/handle_token.rs:19`, with `simhash_loop_enabled()` right below it — the standard kill switch this PR asks for is present.

Found while grouping six of your verify/spec PRs into one stack (#870). Three landed there; this was one of three already in.

If a hunk is missing, say so and I will lift it onto a fresh branch.

PR #704, opened 2026-08-22
verify: raise VERIFY_ROW_CAP 96 -> 160 and derive the metadata layout from it

rsafier:
The batched-verify row budget capped R = Σ ks at 96, which bounds the batched DFlash verify (uniform K=γ+1=8) at n=12 sequences — a C=16 serve cannot batch its verifies in one group. 160 holds n=20 at K=8, and gives the MTP ladder the same headroom.

**The maintainability half is the bigger change:** the cap lived in FOUR places kept in lock-step by comment (`VERIFY_ROW_CAP`, `mtp_dcut`'s `VERIFY_ROW_BUDGET`, `sizes.rs` `bt_rows` and the logits-row floor), and the metadata gap offsets were hardcoded byte literals (384/768/1536/2048) that only *happened* to equal 96-row strides. The offsets are now DERIVED from the cap (`VMETA_*` consts in verify_e.rs: positions 4R | seq_slot 4R | slots 8R | seq_lens 4R | bt at 24R), so the next widening is one constant per crate instead of an archaeology exercise.

**Cost:** logits arena 160 × vocab × 2 B (~79.5 MB at vocab 248320, +32 MB over 96 rows); bt staging +~0.5 MB.

**Measured** (qwen3.8-27B-NVFP4 + DFlash2 drafter, GB10, 32K ctx, 16 seqs, prefix cache on, on a branch carrying the scheduler-side batched DFlash verify):

- C=16 aggregate decode 116.8 (chunked 12+4 under the 96 cap) → **129.6 tok/s** (one 16-group), 16/16 coherent completions
- C=1 cold + two cache-hit repeats **byte-identical** sha across the layout change — the overlay move is numerics-neutral
- decode-floor 47.0 tok/s median over 3 pinned runs, accept_len 6.12

The DFlash capture-band buffer already scales with `--max-batch-size` and needed no change; `VERIFY_WY_TABLE_SEQS` = 32 still bounds sequence count. On main today this is headroom for the MTP ladder and the model-side DFlash dispatch; the scheduler-side batched DFlash verify that exploits it fully is a separate feature PR to follow. Built + fmt clean on main's base.

### DIFF
--- a/crates/spark-model/src/model/trait_impl/verify_e.rs
+++ b/crates/spark-model/src/model/trait_impl/verify_e.rs
@@ -8,7 +8,7 @@
 //! and made RAGGED per sequence by D-Cut) so the n weight-reading verify
 //! forwards collapse into one — the structural fix for the measured MTP
 //! serialization at C>1 (cap=4 at C=4: 25.8 vs 48.5 tok/s; see
-//! BATCHED_MTP_SPEC.md). R is capped at 96 = the exact capacity of the
+//! BATCHED_MTP_SPEC.md). R is capped at `VERIFY_ROW_CAP` = the exact capacity of the
 //! logits rows / meta gaps / bt staging (sizes.rs), reached at n=32 × k=3
 //! rows (the 32:2 depth-at-width shape, wave 11; previously 64 at n=32 ×
 //! k=2, 32 at n=16 × k=2).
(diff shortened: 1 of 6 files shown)


tbraun96:
Closed as **already landed** — not rejected. Apologies for closing before commenting.

Cherry-picking onto current `main` produces an **empty commit**. Confirmed independently: `crates/spark-model/src/model/trait_impl/verify_e2.rs:44` reads

```rust
pub(in crate::model) const VERIFY_ROW_CAP: usize = 160;
```

and `crates/spark-runtime/src/buffers/tests.rs` uses `bt_rows = 160`.

**One thing this turned up that is still worth fixing:** a doc comment at `crates/spark-model/src/model/impl_a3.rs:70` still says `VERIFY_ROW_CAP` **(96)**. It refers to a different arm and is stale — it briefly convinced me this PR had *not* landed. A one-line comment fix would stop the next person making the same mistake. Happy to include it in a docs stack if you would like.

Found while grouping six of your verify/spec PRs into #870.

PR #705, opened 2026-08-22
perf(registry): load each kernel module once — derive the raw handle from the cudarc load

rsafier:
## What

`AtlasRegistry::init` loaded every kernel blob **twice**: once through cudarc's `ctx.load_module` (driver JIT #1 — backs `function()` lookups) and again through raw `cuModuleLoadData` (driver JIT #2 — backs `launch_on_stream`). Two JIT compilations of the same PTX and two resident SASS copies per module, for the entire embedded kernel set, on every model load.

This PR loads each module once: the vendored cudarc grows a `cu_module_raw()` accessor on `CudaModule`, and the registry's `raw_modules` map becomes views into the one cudarc-loaded module. `cuModuleGetFunction` / `cuLaunchKernel` against the derived handle are unchanged.

## Lifetime

The raw handles are owned by the `Arc<CudaModule>`s stored beside them — the two maps live and die together inside `AtlasRegistry`. `unload_raw` therefore stops calling `cuModuleUnload` (that would now double-free the single copy); it drains both maps, raw side first, and cudarc's `Drop` performs the one real unload. The now-unused raw `cuModuleLoadData`/`cuModuleUnload` extern declarations are removed.

## Measured (GB10, 27B + DFlash2 serve, warm JIT cache)

- Boot to ready: **40.1 s → 34.3 s (−14%)**. Cold-JIT boots save proportionally more (the JIT ran twice per module).
- ~0.3 GB less resident at serve-ready (the duplicate SASS copies).
- Correctness: temp-0 outputs unchanged (count smoke byte-stable across the change, MinHeap correct), C=16 aggregate decode 113.7 tok/s with 16/16 coherent streams — within noise of the double-load build. The kernel load audit resolves identically.

## Validation

`cargo test -p atlas-core` (98 passed) · `cargo check --workspace` clean on this base · fmt clean · full serve boot + smoke + C=16 sweep on the branch content (as part of the integration branch).

### DIFF
--- a/crates/atlas-core/src/registry.rs
+++ b/crates/atlas-core/src/registry.rs
@@ -24,10 +24,10 @@ use cudarc::nvrtc::Ptx;
 pub use crate::cuda_host::{CudaHost, host, release};
 use crate::error::{AtlasError, Result};
 
-// Raw CUDA driver API
+// Raw CUDA driver API. (`cuModuleLoadData`/`cuModuleUnload` left this list
+// when the raw handles became views into the cudarc-loaded modules — the
+// registry no longer loads or unloads anything through the raw API.)
 unsafe extern "C" {
-    fn cuModuleLoadData(module: *mut *mut c_void, image: *const c_void) -> i32;
-    fn cuModuleUnload(hmod: *mut c_void) -> i32;
     fn cuModuleGetFunction(hfunc: *mut *mut c_void, hmod: *mut c_void, name: *const i8) -> i32;
     fn cuLaunchKernel(
         f: *mut c_void,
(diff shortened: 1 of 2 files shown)


tbraun96:
**AutoMerger sieve verdict: `sieve:cleared` — all three sieves pass.** Integrity specifically verified against the live tree: `unload_raw` is called only from `AtlasRegistry::drop` (registry.rs:143), and draining `raw_modules` before `modules` means the one real `cuModuleUnload` happens at the `Arc<CudaModule>` drop — no use-after-free window. Grouping into **stack `perf-plumbing`** with #781 (one certification campaign for the group); composition is offline-gated first and the stack PR will reference this one. This PR stays open until the stack lands, then closes citing the landing commit.

tbraun96:
**Stacked → lands via #891 (`stack/perf-plumbing`), one campaign for the pair.**

| order | PR | attribution signature | |
|---|---|---|---|
| 1 | #705 | boot-phase (module load/unload) | **← this PR** |
| 2 | #781 | grouped-GEMM phase (EP null-group skip) |  |

Composition is patch-id-verified byte-equivalent to this PR's diff, authorship preserved, offline-green. Adversarial review passed with a blocking pre-campaign condition: a single-node MoE smoke runs before any campaign (covers the config this pair's evidence never executed). Please don't push here while the stack is in flight; this PR closes with landing evidence after #891 merges. If any hunk does not make it across, point at it and I will lift it onto a fresh branch.

DrRainbows:
Closing as already on `main`. This landed via #891 (`356701d6`, an ancestor of `origin/main`). Named artifact: `crates/atlas-core/src/registry.rs:236` (`module.cu_module_raw()`). Both files from this PR's diff (`registry.rs`, `vendor/cudarc/src/driver/safe/core.rs`) are byte-identical on `main`. Every commit here cherry-picks empty onto current `main` (`c2948e3f`).

PR #701, opened 2026-08-21
fix(gate): preserve records across proven test-only changes

TheTom:
## Summary

`PR benchmark gate` currently invalidates all ten records for any changed path under `crates/`. Twelve open test-audit PRs change only Rust modules excluded from release builds by explicit `#[cfg(test)]` parent edges, but each PR is still required to replace all ten GPU records.

This PR adds an exact registry for three proven test-only modules:

- `crates/atlas-core/src/config/tests.rs`
- `crates/atlas-core/src/config/gguf/tests.rs`
- `crates/atlas-core/src/config/parsers/lora_tests.rs`

A registered file no longer invalidates benchmark records. Unregistered files, production parents, nested helpers, test-looking filenames, and verdict boundary files retain the existing fail-closed behavior.

## Observed failure

PRs #674, #675, #678, #679, #680, #685, #687, #689, #690, #691, #694, and #696 contain no production changes. Their benchmark jobs report all ten records missing because the only invalidating path is one of the first two registered files.

Those files are absent from non-test builds. Re-running model benchmarks cannot observe their assertions because the benchmarked release program is unchanged.

## Structural proof

Each registry entry records the exact file, exact parent, module name, and explicit `#[path]` value when present. Tests prove:

1. the file and parent exist;
2. the parent has exactly one registered module declaration guarded by `#[cfg(test)]`;
3. no Rust source creates another `include!` or `#[path]` edge to the file;
4. neighbouring and test-looking paths still invalidate all ten gates;
5. a production-parent change cannot hide beside an exempt test change;
6. a real Git fixture keeps a record valid across the test-only commit and invalidates it after the production parent changes.

Tests embedded inside production source files remain performance-boundary changes.

## Bootstrap

The policy changes three verdict boundary files:

- `crates/atlas-plugin/src/gate/check.rs`
- `crates/atlas-plugin/src/gate/coverage.rs`
- `crates/atlas-plugin/src/gate/required.rs`

The repository already has a content-pinned amnesty mechanism for this bootstrap problem. This PR activates it for exactly the final blobs of those three files:

- `check.rs`: `2dd3bed0103d38009e82de8d507d26320a1cb307`
- `coverage.rs`: `6aef608fab0d120096d3b55eaed3095155e2beeb`
- `required.rs`: `c8b8311abbb88bceb844710e6443e28ce843f236`

[...]

PR #708, opened 2026-08-22
test(atlas-core): pin DeepSeek V4 runtime controls

TheTom:
## Summary

The DeepSeek-V4 parser test omitted five runtime controls already produced by the parser. This patch pins their exact values without changing production code.

## Broken proof

The fixture contained the inputs for MLA rotary geometry, routed-expert scaling, and hyper-connections, but the test observed none of those outputs.

## Mutation evidence

The old test stayed green when production omitted the derived rotary fraction, reset routed scaling to 1.0, and left the three hyper-connection values at zero.

The repaired test independently fails at:

- rotary fraction: 1.0 versus 0.125;
- routed scale: 1.0 versus 1.5;
- residual streams: 0 versus 4;
- Sinkhorn iterations: 0 versus 20;
- HC epsilon: 0 versus 1e-6.

Removing only the explicit raw routed-scaling assignment is behaviorally equivalent because Serde already populated the field. The mutation proof resets the parsed value instead.

## Runtime tie

The rotary fraction selects the Q/K slice receiving position encoding. Routed scaling weights every selected expert contribution. HC stream count, iteration count, and epsilon control residual mixing.

DeepSeek's [published config](https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash/blob/60d8d70770c6776ff598c94bb586a859a38244f1/config.json) records the corresponding 64/512 rotary geometry, 1.5 routed scale, four streams, 20 iterations, and 1e-6 epsilon.

## Test plan

- [x] `CUDARC_CUDA_VERSION=13010 cargo test -p atlas-core` (98 passed)
- [x] `cargo fmt --check`
- [x] `python3 scripts/check_spdx.py crates/atlas-core/src/config/tests.rs`
- [x] `git diff --check`
- [x] Replayed each omitted/defaulted runtime control independently

## Notes for reviewers

- The diff is five assertions in an existing `#[cfg(test)]` module.
- #701 is merged. After updating this branch to include it, the PR benchmark gate passed at `15713912db`, confirming this exact test module preserves the standing records.
- Dedicated sibling tests already cover the YaRN factor, betas, original length, compressor theta, and amplitude terms; those assertions are not duplicated here.

## CLA

### DIFF
--- a/crates/atlas-core/src/config/tests.rs
+++ b/crates/atlas-core/src/config/tests.rs
@@ -556,6 +556,7 @@ fn test_parse_deepseek_v4_config() {
     assert_eq!(cfg.kv_lora_rank, 512); // fallback default
     assert_eq!(cfg.qk_nope_head_dim, 448); // head_dim - qk_rope_head_dim
     assert_eq!(cfg.v_head_dim, 512); // fallback to head_dim
+    assert_eq!(cfg.partial_rotary_factor, 0.125); // 64 / 512
     assert_eq!(cfg.num_experts, 256);
     assert_eq!(cfg.num_experts_per_tok, 6);
     assert_eq!(cfg.moe_intermediate_size, 2048);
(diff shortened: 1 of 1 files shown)


tbraun96:

PR #712, opened 2026-08-22
test(atlas-core): strengthen GPU fault contracts

TheTom:
## Summary

The GPU fault tests proved the fatality decision but allowed the originating CUDA error to disappear from the terminal diagnosis. The concurrency test also started callers opportunistically, weakening its stated race setup. This patch repairs both test oracles and makes two mutation notes match the defects actually executed.

## Broken proof

Removing the original `{err}` from `Fatality::ContextLost` left `failed_probe_means_context_lost` green. It still found the operation and probe text, even though the log had lost the status from the operation that initiated diagnosis.

Replacing atomic `OnceLock::set(...).is_ok()` ownership with check, yield, set, and unconditional success passes the sequential first-writer test. Without a coordinated release, the race fixture depended on thread startup timing rather than explicitly creating contention.

## Mutation evidence

The repaired classification test independently requires:

- `w4a16_gemm_t launch`;
- the exact original `CUDA_ERROR_MISALIGNED_ADDRESS (716)` text;
- the exact `cuStreamSynchronize returned 716` probe failure.

Replaying the original-error deletion now fails at the second observation.

The race test now blocks eight caller threads on one barrier. Restored production reports one winner. The check-then-set mutant still passes the sequential test but reports eight winners in the coordinated replay.

The healthy-probe mutation notes were also made executable and partition-specific. A 716-text fatality guard fails only the sticky-looking case; an OOM-specific guard fails only the ordinary recoverable case.

## Runtime tie

`spark-runtime/src/cuda_backend/fault_probe.rs::note_failure` passes CUDA failures through this classifier, stores the first terminal reason in the process-global latch, and uses the winning latch result to emit the terminal log once. API health, middleware, scheduler shutdown, and both process exit paths consume that state.

## Test plan

- [x] `CUDARC_CUDA_VERSION=13010 cargo test -p atlas-core --lib` (98 passed)
- [x] `python3 scripts/check_spdx.py crates/atlas-core/src/fault_tests.rs`
- [x] `git diff --check`
- [x] Replayed the missing-original-error mutant
- [x] Replayed the non-atomic ownership mutant and confirmed the sequential test remains green

## Notes for reviewers

- This changes only an existing `#[cfg(test)]` module.
- The direct `note_failure` FFI composition and real GPU probe timing are not claimed by these unit tests. They need a separate injected seam and hardware experiment.
- #701 currently registers three config test modules only. It does not cover `fault_tests.rs`, so the benchmark gate may continue requesting unrelated records for this PR until the narrow test-only registry is extended.
- The container-based full license command could not run because Docker is unavailable locally. The repository's single-file SPDX checker passes, and this patch does not alter the existing line-one header.

## CLA

### DIFF
--- a/crates/atlas-core/src/fault_tests.rs
+++ b/crates/atlas-core/src/fault_tests.rs
@@ -33,10 +33,15 @@ fn failed_probe_means_context_lost() {
     );
     match v {
         Fatality::ContextLost(reason) => {
-            // The message must name BOTH the originating op and the probe, or
-            // an operator reading only the log cannot tell which call died.
+            // The message must preserve the originating operation and error,
+            // plus the probe failure. An operator reading only the log needs
+            // all three to distinguish the first failure from its sticky echo.
             assert!(reason.contains("w4a16_gemm_t launch"), "reason: {reason}");
-            assert!(reason.contains("cuStreamSynchronize"), "reason: {reason}");
+            assert!(reason.contains(STICKY_716), "reason: {reason}");
+            assert!(
+                reason.contains("cuStreamSynchronize returned 716"),
+                "reason: {reason}"
+            );
         }
         Fatality::Isolated => panic!("a failed probe must be fatal"),
     }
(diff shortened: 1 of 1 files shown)


tbraun96:

PR #713, opened 2026-08-22
test(atlas-core): isolate numeric conversion oracles

TheTom:
## Summary

The numeric conversion tests had two self-confirming oracles. Float equality hid FP8 signed-zero errors, and BF16 widening was checked only by passing its result through the narrowing function. This patch makes those directions byte-exact and independent, and adds the missing test for the documented RNE escape hatch.

## Broken proof

Applying the FP8 sign only to nonzero magnitudes drops the sign from byte `0x80` and the negative NaN mapping. Both existing FP8 tests stayed green because `+0.0 == -0.0`.

Parsing BF16 bytes as big-endian while byte-swapping the narrowing result left the old 65,536-value round trip entirely green. The two wrong functions inverted each other.

`ATLAS_DISABLE_RNE` is documented as a presence check, but no test exercised it. Changing production to recognize only value `1` was invisible to the prior suite.

## Mutation evidence

- Raw-bit FP8 assertions make the sign-dropping mutant fail at encoding `0x80`, expected `0x80000000` but observed `0x00000000`.
- The exhaustive FP8 oracle now compares exact f32 bits for all 256 encodings. Its name separates OCP finite values from Atlas's deliberate signed NaN-to-zero policy.
- BF16 widening now compares every result directly with `bf16 << 16`. Narrowing constructs its f32 inputs independently. The coupled endian mutant makes both tests fail at `0x0001`.
- A subprocess test checks `ATLAS_DISABLE_RNE=0` and `=1` without changing the parent test process environment. The value-sensitive mutant fails the `0` partition with RNE result `0x3F81` instead of truncation result `0x3F80`.

## Runtime tie

`spark-model` uses the shared FP8 table and f32-to-BF16 cast during CPU weight dequantization. BF16 checkpoint scales are widened before block-scaled FP8 weights are converted. A wrong sign, endian interpretation, or rounding mode changes uploaded weight bytes.

The [OCP FP8 specification](https://www.opencompute.org/documents/ocp-8-bit-floating-point-specification-ofp8-revision-1-0-2023-06-20-pdf) defines the E4M3 finite encodings. Atlas's Rust and CUDA software decoders additionally map the two NaN encodings to signed zero.

## Test plan

- [x] `CUDARC_CUDA_VERSION=13010 cargo test -p atlas-core --lib` (100 passed)
- [x] `python3 scripts/check_spdx.py crates/atlas-core/src/numeric.rs`
- [x] `git diff --check`
- [x] Replayed the FP8 zero-sign mutant
- [x] Replayed the coupled BF16 endian mutant after confirming the old round trip survived it
- [x] Replayed the value-sensitive RNE escape-hatch mutant

## Notes for reviewers

- This changes only tests in an existing production module. Numeric conversion code is unchanged.
- The small FP8 reference test intentionally overlaps the exhaustive test. It supplies readable anchors; the exhaustive oracle covers all encodings through separately derived float math.
- The tests are inline in the production `numeric.rs` file. A path-only exclusion would be unsafe because it would also hide later arithmetic changes. Making this class of PR benchmark-neutral requires moving the test module to a dedicated file or a fail-closed hunk-aware policy, not registering `numeric.rs` as test-only.

## CLA

### DIFF
--- a/crates/atlas-core/src/numeric.rs
+++ b/crates/atlas-core/src/numeric.rs
@@ -113,15 +113,15 @@ mod tests {
 
     #[test]
     fn fp8_lut_reference_values() {
-        assert_eq!(fp8_e4m3_to_f32(0x00), 0.0); // +0
-        assert_eq!(fp8_e4m3_to_f32(0x80), -0.0); // -0
+        assert_eq!(fp8_e4m3_to_f32(0x00).to_bits(), 0x0000_0000); // +0
+        assert_eq!(fp8_e4m3_to_f32(0x80).to_bits(), 0x8000_0000); // -0
         assert_eq!(fp8_e4m3_to_f32(0x38), 1.0); // exp=7, mant=0
         assert_eq!(fp8_e4m3_to_f32(0xB8), -1.0);
         assert_eq!(fp8_e4m3_to_f32(0x3C), 1.5); // exp=7, mant=4
         assert_eq!(fp8_e4m3_to_f32(0x7E), 448.0); // max finite
         assert_eq!(fp8_e4m3_to_f32(0xFE), -448.0); // min finite
-        assert_eq!(fp8_e4m3_to_f32(0x7F), 0.0); // NaN -> 0
-        assert_eq!(fp8_e4m3_to_f32(0xFF), 0.0); // -NaN -> 0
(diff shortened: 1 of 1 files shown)


tbraun96:

PR #719, opened 2026-08-22
fix(metrics): the SSM gauge read an orphaned counter — feed it the real Marconi pool

rsafier:
## The bug

The scheduler snapshot's `ssm_slots_used` / `ssm_slots_total` — the numbers behind the TUI's SSM gauge and any metrics consumer — were fed from `SessionSsmManager::{session_count, total_slots}`. But **`SessionSsmManager::save_snapshot` has no callers anywhere in the tree**, so both counters have been zero for the manager's entire life (its `evict_expired` loop is equally inert). The gauge rendered **0/0 on every serve, always** (user-reported), while the actual pool — `SsmSnapshotPool`, 16 Marconi slots on the DFlash recipe — saved and restored snapshots on every warm turn with nothing reporting it.

## The fix

- `SsmSnapshotPool` grows an `occupancy()` accessor: `(num_slots − free, num_slots)`, Marconi region only — the decode-rollback ring is deterministically addressed per sequence, so its fullness means nothing and is deliberately excluded.
- The `Model` trait exposes `ssm_snapshot_occupancy() -> Option<(u32, u32)>` — `None` for pool-less models (non-SSM, CPU backends), so consumers can fall back instead of faking a 0/0.
- The scheduler snapshot reads the model's occupancy, with the orphaned session-manager counters kept only as the pool-less fallback. The orphaned state of `SessionSsmManager` is documented at the site rather than removed — whether to retire the manager entirely is a separate decision.

## Also included

Two permanent one-line-per-stream `tracing::debug!` forensics for a TTFT investigation in flight (the client's first SSE delta lands ~0.5s after the scheduler emits the first token, with every hop statically immediate): `handle_token` logs the first stable content delta, `encode_stream` logs the first delta reaching the SSE encoder. Together with the existing scheduler line they triangulate channel-wait vs wire vs client. Debug level, silent at the production `RUST_LOG=info`.

## Validation

3,056 spark-server + spark-model tests pass on this base (the one failure is the known local-environment `closure_attestation` quirk that CI runs green) · fmt clean · verified live on the integration branch: the gauge reports real occupancy after a warm turn.

### DIFF
--- a/crates/spark-model/src/model/ssm_snapshot.rs
+++ b/crates/spark-model/src/model/ssm_snapshot.rs
@@ -107,6 +107,17 @@ pub(crate) struct SsmSnapshotPool {
 }
 
 impl SsmSnapshotPool {
+    /// Marconi occupancy: `(slots holding a live snapshot, total LRU
+    /// slots)`. The decode-rollback ring region is deliberately excluded —
+    /// it is deterministically addressed per sequence, not a cache whose
+    /// fullness means anything.
+    pub(super) fn occupancy(&self) -> (usize, usize) {
+        (
+            self.num_slots - self.free_slots.lock().len().min(self.num_slots),
+            self.num_slots,
+        )
+    }
+
     /// Build the snapshot pool.
     ///
     /// `num_slots` sizes the Marconi LRU region; `decode_ring_slots` ×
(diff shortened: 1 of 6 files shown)


tbraun96:

Closing by hand because the commits were **re-cut**, not merged: this PR's head is not an ancestor of main even though its content is, so GitHub will not auto-close it. Verified by artifact rather than assumed — `ssm_snapshot_occupancy` is on main in three files — the gauge now reads the real pool rather than the orphaned counter.

PR #720, opened 2026-08-22
perf(stream): sanitizer held every response's first ~17 bytes — hold only marker-compatible suffixes

rsafier:
## The bug

The content sanitizer's no-match branch held back a flat `tag_max − 1` bytes so a leak marker straddling a chunk boundary could fuse with the next chunk. Correct goal, wrong rule: the flat window withholds the **first ~17 bytes of every streaming response** — `"An"` is not a prefix of `<tool_call>`, yet it sat in `tag_scan_buf` until 3–5 more decode steps pushed it past the window.

Forensics on the DFlash serve (debug lines included in this PR): the first stable content bytes existed **0.2ms** after the scheduler sampled the first token, but the first delta batch left `handle_token` at "4–6 tokens seen" — **250–500ms of first-delta latency on every streaming response**, cold and warm alike, on any parser that exports leak markers (qwen3_coder/qwen3_xml deployments; Hermes/Gemma/Mistral use the pass-through fast path and were unaffected).

## The fix

The holdback now retains **only the longest suffix that is a byte-prefix of some leak marker** — the only bytes that can still fuse into a match. Marker-incompatible tails emit immediately. Straddle fusion is unchanged and pinned tighter by tests: `"done.<tool_c"` emits `"done."` and holds exactly `<tool_c`; the completion across the boundary still suppresses.

Also fixes a latent off-by-one the flat window masked: the char-boundary cut used `char_indices()`, which never yields the buffer's **end** index, so the final character could never be emitted. Replaced with a proper boundary floor.

Two stale tests updated: they asserted the flat holdback itself ("abc before a partial tag must be withheld"); the fusion *outcome* they exist for is unchanged.

## Measured (27B + DFlash2, GB10)

| metric | before | after |
|---|---|---|
| warm-turn TTFT (with #718's carry) | 0.71–0.83s | **0.47s** |
| cold TTFT (2.2K prompt) | 2.81s | **2.54s** |
| tiny-prompt TTFT floor | 0.27s | **0.165s** |
| first delta leaves handle_token | 4–6 tokens seen | **1 token** |

Completion token counts byte-stable; C=16 aggregate 115.4 tok/s, 16/16 coherent.

## Validation

2,414 spark-server tests pass on this base (the one failure is the known local `closure_attestation` environment quirk; CI runs it green) · fmt clean · full serve boots with before/after probe runs. The two forensics `tracing::debug!` lines (one per stream) ride along — silent at production `RUST_LOG=info`; they are the instrument that found this and will catch regressions.

### DIFF
--- a/crates/spark-server/src/api/chat_stream/handle_token.rs
+++ b/crates/spark-server/src/api/chat_stream/handle_token.rs
@@ -15,6 +15,15 @@ use super::super::sanitizer::sanitize_content_chunk;
 use super::super::stream_guards::{bump_f12_tool_call_count, check_loop_watchdog};
 use super::ctx::StreamCtx;
 use super::state::StreamState;
+
+/// `ATLAS_SIMHASH_LOOP=0` disables the F4 SimHash semantic-loop guard.
+/// Default ON — see the comment at the check site for why an operator would
+/// turn it off (one-strike near-duplicate detection kills streams over
+/// legitimately repetitive structured output).
+fn simhash_loop_enabled() -> bool {
+    static ON: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
+    *ON.get_or_init(|| std::env::var("ATLAS_SIMHASH_LOOP").as_deref() != Ok("0"))
+}
 use super::strip::{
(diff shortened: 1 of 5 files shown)


tbraun96:

PR #728, opened 2026-08-22
Thermal throttle/thrash detection in the HAL, surfaced in the TUI

tbraun96:
Adds a thermal thrashing detector at the hardware abstraction layer and surfaces it
in the TUI — a Stats subsection plus a header alert.

## Why a detector rather than a readout

`ThrottleCounters` are **cumulative microseconds since driver load**, so a single
reading cannot tell a box that is throttling *right now* from one that throttled last
week. `hardware::throttle_monitor` differences successive reads into a per-window
fraction, which is the question people actually have.

**Thrashing is a separate diagnosis from sustained throttling, and both are reported.**
A part pinned at 100% throttle is simply too hot — predictable, and it shows up in
throughput. A part oscillating in and out many times a minute has unstable clocks, and
the damage lands in latency *variance*: p90 drifting away from a median that still looks
healthy. That failure mode is exactly the one that has repeatedly been misread here as a
code regression. A fraction alone cannot distinguish them; the transition count can.

## Measurement decisions, each with a reason

- **SW power cap is never counted as thermal.** On a *healthy* GB10 it is asserted
  16,130 s of an 11.2-day uptime — the normal steady state of a power-limited part
  (`ThrottleActive::thermal`'s existing docs establish this). Folding it in would pin the
  indicator on and train everyone to ignore it. It gets its own field, labelled normal.
- **The three thermal counters overlap**, so their union lies between the max and the
  sum. The **max** is reported: an over-stated throttle number is the kind that gets a
  real regression waved away as "the box was hot".
- **A counter going backwards** (driver reload, GPU reset) reports **unknown**, not a
  saturated fraction. A test pins this — an earlier draft returned `Some(0.0)`, i.e.
  "not throttling", from a reset.
- **"No data" never renders as healthy**, anywhere. Unknown and Ok are distinct states
  in the alert enum and in every rendered field.

## Off the render thread

Reading the counters costs a 100–200 ms `nvidia-smi` spawn and the TUI renders every
100 ms, so a detached sampler publishes a snapshot every 2 s and the UI reads it.
Monitoring that stutters the tool people use to judge whether the machine is healthy
would defeat its own purpose.

## Header alert

Silent for both `Ok` and `Unknown` — an indicator that always renders something is one
people stop reading. `THRASH` outranks `THROTTLING`. Thresholds are 20% of a window,
and 6 flips over >=8 samples (~64 s): loose enough that a normal thermal excursion or a
cold-to-loaded ramp does not cry wolf.

## Scope note — the fused GDN spine is NOT in this PR

It was intended to be bundled here. Investigating it produced a result worth recording
instead, and no code:

| phase | ms | why |
|---|---:|---|
| `recompute_wu` | 41.2 | `T=(I+L)^-1` by forward substitution — 64 sequential steps |
| `chunk_delta_h` | 33.2 | the state recurrence — serial over chunks by definition |
| `chunk_fwd_o` | 31.1 | chunk-parallel GEMMs |

The spine is two serial dependency chains plus a GEMM, and the three kernels have
*incompatible parallelism*: `wu` and `fwd_o` run ~2048 CTAs while `chunk_delta_h` runs
~32. **Naive fusion cannot fix either chain** — it only trades parallelism for traffic.
Eliminating the `S_out` round-trip by fusing `fwd_o` into `chunk_delta_h` saves ~34
MB/layer at the tested shape, worth 3-5%, not the 7x.

ptxas reports `256 bytes stack frame` for `recompute_wu` — exactly `float u[64]`, in
*local* memory because `u[l]` is indexed by a runtime `l`. A blocked solve (block 16)
would put it in registers and cut the serial chain 4x, but it needs each thread's solved
history in shared memory, taking live smem from 32.25 KB to 64-96 KB and collapsing
occupancy from **3 CTAs/SM to 1**. That trade loses, so it was reverted rather than
shipped. Reaching FlashQLA's 15 ms needs its full restructuring (chunk 32, warp-
specialised producer/consumer, TMA, its own `kkt_solve`) — a separate piece of work, not
a fusion patch.

## Verification

14 tests. `fmt`, `clippy --workspace --tests`, the non-CUDA build and `cargo test
--workspace` all pass **by exit code**. All touched files under the 500-LoC cap
(stats_tab 422, stats_thermal 171, header 281, thermal 233, throttle_monitor 277).

No kernel or inference path is touched, so no benchmark gate applies.

### DIFF
--- a/.benchmarks/agentic-webserver/2026-08-22-1b13b47f3c.json
+++ b/.benchmarks/agentic-webserver/2026-08-22-1b13b47f3c.json
@@ -1,0 +1,467 @@
+{
+  "schema": 1,
+  "benchmark_id": "agentic-webserver",
+  "benchmark_name": "Agentic Webserver Test",
+  "git_sha": "1b13b47f3c",
+  "recorded_at": 1787373594,
+  "target_model": "Qwen/Qwen3.6-35B-A3B-FP8",
+  "params": {
+    "build_timeout_s": "600",
+    "command_timeout_s": "180",
+    "iterations": "10",
+    "max_tokens": "8192",
+    "max_turns": "40",
+    "request_timeout_s": "900",
+    "s_per_turn_budget": "8.5",
+    "serve_timeout_s": "30",
+    "wall_budget_s": "1800"
+  },
+  "command": [
+    "spark",
+    "benchmark",
+    "run",
+    "agentic-webserver",
+    "--param",
+    "build_timeout_s=600",
+    "--param",
+    "command_timeout_s=180",
+    "--param",
+    "iterations=10",
+    "--param",
+    "max_tokens=8192",
+    "--param",
+    "max_turns=40",
+    "--param",
+    "request_timeout_s=900",
+    "--param",
+    "s_per_turn_budget=8.5",
+    "--param",
+    "serve_timeout_s=30",
+    "--param",
+    "wall_budget_s=1800",
+    "--yes",
+    "--pull-request-gate"
+  ],
+  "served_by": "qwen3.6/qwen3.6-35b-a3b-fp8-bf16head",
+  "atlas_version": "1.0.0-beta-preview",
+  "hardware": {
+    "gpu": "NVIDIA GB10",
+    "driver": "580.126.09",
+    "sm_clock_mhz": 2392.0,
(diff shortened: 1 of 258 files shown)


tbraun96:

PR #729, opened 2026-08-22
Repair the recompute_wu correctness gate (it could not run at all)

tbraun96:
`gdn_recompute_wu_gateb` is Gate B for the largest kernel in the GDN prefill spine.
It has been **dead since the kernel signature changed**, and the failure is total:

```
Error: cuLaunchKernel failed: CUDA_ERROR_INVALID_VALUE (1): invalid argument
       (grid=[1,32,1], block=[128,1,1], shared_mem=49408)
```

That was reproduced on the pre-fix tree *before* the fix was written, rather than
inferred from reading the code.

## Four independent drifts, each alone enough to make the gate useless

| # | drift | consequence |
|---|---|---|
| 1 | **16 args passed, 20 declared** — `gc_out` plus the `cu_seqlens`/`cu_chunks`/`is_varlen` triple all missing | `cuLaunchKernel` reads past the argument array |
| 2 | **block 128**, but kernel is `__launch_bounds__(256,1)` with pass 2 fenced to `tid >= 128` | the W forward-substitution **never runs**; `W_out` unwritten; the W comparison scores against uninitialised memory |
| 3 | **smem 49408** — a second `C*C*4` block from before `L` was aliased onto `kk` | 16 KB over-allocated vs the production `smem_wu` |
| 4 | launch existed only as an inline chain, unshareable | two copies of a 20-argument list is *how* this drifted |

Drift 1 is the same defect class that segfaulted the 512-wide prefill kernel earlier
this cycle: an argument list copied once and never re-synced. Drift 2 is the worse
one — **a gate that validates nothing is worse than no gate, because it is counted as
coverage.**

## The kernel is fine

With the harness repaired the gate passes at **cos 0.999999 on both U and W** across
three shapes including a ragged one (t=64/nt=1, t=128/nt=2, t=200/nt=4). W is
meaningful here for the first time.

Fix 4 is the one that prevents recurrence: a single `launch_wu` closure now serves
both the correctness compare and the timed replay, so there is one argument list.

## Opt-in timed replay

`WU_BENCH_ITERS=N` (default off, 8 warm-up launches discarded). This kernel is
**41.2 ms of the 105.7 ms spine** and a full serve run costs ~30 min — far too slow
to iterate a kernel against. Measured: **71.7 / 100.6 / 148.1 µs per launch** for
nt=1/2/4.

⚠ Those are **isolated numbers on synthetic inputs** — good for ranking variants
against each other, never quotable as an in-serve figure.

## Verification

fmt clean. clippy reports nothing in this file (the `is_multiple_of` lints in the
examples tree are pre-existing, in other files, and out of scope here). 243 lines,
under the cap. No production code path is touched — this is a test harness only.

### DIFF
--- a/crates/spark-model/examples/gdn_chunk_delta_h_gateb.rs
+++ b/crates/spark-model/examples/gdn_chunk_delta_h_gateb.rs
@@ -75,7 +75,13 @@ fn main() -> Result<()> {
     let backend = AtlasCudaBackend::new(0, &set.modules)?;
     let g: &dyn GpuBackend = &backend;
     let k_wu: KernelHandle = g.kernel("gated_delta_rule_fla", "gated_delta_rule_recompute_wu")?;
-    let k_dh: KernelHandle = g.kernel("gated_delta_rule_fla", "gated_delta_rule_chunk_delta_h")?;
+    // `_ksplit`, NOT the bare `gated_delta_rule_chunk_delta_h`: production
+    // resolves only `_ksplit` / `_vfused` / `_vtile` / `_tc_vblock`, so the bare
+    // kernel this gate used to target is one no serve has ever run.
+    let k_dh: KernelHandle = g.kernel(
+        "gated_delta_rule_fla",
+        "gated_delta_rule_chunk_delta_h_ksplit",
+    )?;
 
     let mut all_ok = true;
     for &t in &[64usize, 128, 200] {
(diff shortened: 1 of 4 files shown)


tbraun96:
Folded into #731, which now targets `main` and carries all eight commits of this stack.

**Why:** each of #729/#730/#731 independently owes a full ten-gate benchmark campaign (~8 h of GPU each — 24 h for three related commits). #729's records are invalidated by `crates/spark-model/examples/*.rs`, which Cargo builds as *separate binaries* that cannot be linked into `spark` — but the gate's path rules do not know that, and the `BOUNDARY_FILES` machinery deliberately does not exempt itself, so the conservative verdict stands.

Collapsing the stack into one PR means **one campaign instead of three**. Nothing is dropped: every commit, message and measurement moves across intact.

Closing this in favour of #731.

PR #730, opened 2026-08-22
Parallelise the recompute_wu gc scan (7-10% at production shapes)

tbraun96:
Stacked on **#729** — base is `gdn-wu-gate-repair`, not `main`. None of this was
measurable before that PR, because the gate harness could not run at all.

## The defect

The cumulative log-gate prefix scan ran entirely on `tid 0` — `CHUNK` iterations, each
a global load plus a `logf` — while the other **255 threads sat at the following
`__syncthreads()`**. Replaced with a Hillis-Steele inclusive scan: `log2(CHUNK)=6`
barrier-separated steps.

## Measured at production shapes

| nt | serial | parallel | gain |
|---:|---:|---:|---|
| 16 (seq 1024) | 764.2 µs | 690.0 µs | **9.7%** |
| 64 (seq 4096) | 2661.0 µs | 2475.8 µs | **7.0%** |

Gate B passes at cos 0.999999 on both U and W with **max-abs unchanged** at every
shape tested (nt=1,2,4,16,64).

### This PR originally claimed 8.5% measured at nt<=4, and that was not a safe basis

At nt=1 the grid is 32 CTAs over ~48 SMs, so the GPU is **underfilled and occupancy
never binds**. A sibling candidate — a bank-conflict-free shared-memory accumulator —
measured like this:

| nt | 1 | 4 | 16 | 64 |
|---|---|---|---|---|
| speedup | 1.60x | 1.30x | 1.09x | **0.92x** |

**The sign flips.** It would have shipped advertising 1.6x while being an **8%
regression on the real workload**. The harness now takes `ATLAS_WU_SHAPES`, which is
both why the number above is trustworthy and why the sibling was caught.

### In-serve context

This kernel is 41.2 ms of a 105.7 ms spine, so 7% is **~2.9 ms of a 536 ms prefill** —
real, and small. The isolated per-launch numbers rank variants against each other on
one box; they are **not** in-serve figures.

## Three results documented in the kernel header

| finding | result |
|---|---|
| where the time goes | prologue 14.3/16.4/20.5 us of 68.9/96.6/140.8 -> the **solve is 79-85%** |
| shared-memory accumulator | **rejected** — see below |
| FMA dependency chain | **neutral** when split 4 ways, so the ~2048-deep chain isn't the constraint |

The shared-memory result is recorded in full because **my first rejection of it was
itself wrong**. Probe 1 used a `[tid][l]` layout — stride 256 B, so all threads of a
pass hit **one of 32 banks, a 32-way conflict** — measured 0.59-0.81x, and I read that
as "local memory is fine". It measured a bad *layout*. Probe 2, conflict-free `[l][tid]`,
won at small nt and lost at nt=64. Conclusion unchanged, reasoning replaced: keep local
memory **not because it is cheap, but because 64 KB of smem costs more at nt>=64**.

That bar also stands for future work: any reformulation buying speed with a materially
larger smem footprint must clear it. It is why a blocked triangular solve holding solved
rows in smem was drafted and rejected.

## Correctness

**Not bit-identical** — a tree scan sums in a different order. cos and max-abs are
unchanged, but **the ssm-poisoning gate is the one that matters here**: the `vfused` note
in this file records that cos >= 0.99 *and* byte-identical single-prompt greedy both
missed SPLIT=4's 1.4-point BFCL drift. This PR must earn that record; I am not treating
the cos pass as sufficient.

`gc[ce..CHUNK]` is now zeroed. The serial version left those as whatever was in shared
memory, and the L build reads `gc[i]` up to `CHUNK` — beta is 0 there, but
`0 * expf(garbage)` can be NaN.

### DIFF
--- a/crates/spark-model/examples/gdn_recompute_wu_gateb.rs
+++ b/crates/spark-model/examples/gdn_recompute_wu_gateb.rs
@@ -73,7 +73,14 @@ fn main() -> Result<()> {
     eprintln!("recompute_wu handle={}", k.0);
 
     let mut all_ok = true;
-    for &t in &[64usize, 128, 200] {
+    // Production runs nt=16-64 (seq 1024-4096). The small shapes alone cannot
+    // see an occupancy-bound regression, because at nt=1 the grid is 32 CTAs
+    // over ~48 SMs and occupancy never binds.
+    let shapes: Vec<usize> = match std::env::var("ATLAS_WU_SHAPES") {
+        Ok(v) => v.split(',').filter_map(|x| x.trim().parse().ok()).collect(),
+        Err(_) => vec![64usize, 128, 200],
+    };
+    for &t in &shapes {
         let nt = t.div_ceil(C);
         let mut r = Lcg(0xC0FFEE ^ t as u64);
         let key: Vec<bf16> = (0..t * NK * KD)
--- a/kernels/gb10/common/gated_delta_rule_fla.cu
+++ b/kernels/gb10/common/gated_delta_rule_fla.cu
@@ -133,6 +133,48 @@ __device__ __forceinline__ void mma_gram(
 //   W_out: [.. ][CHUNK][K_DIM]   = T·(β·exp(gc)·K)
 // where T=(I+L)⁻¹ applied by forward-substitution (parallel over the V/K cols).
 // smem: sk_bf(16K) + kk(16K f32) + L(16K f32) + gc(256) ≈ 48.25KB.
+// ── MEASURED (2026-08-22) ───────────────────────────────────────────────────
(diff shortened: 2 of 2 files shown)


tbraun96:
Folded into #731, which now targets `main` and carries all eight commits of this stack.

**Why:** each of #729/#730/#731 independently owes a full ten-gate benchmark campaign (~8 h of GPU each — 24 h for three related commits). #729's records are invalidated by `crates/spark-model/examples/*.rs`, which Cargo builds as *separate binaries* that cannot be linked into `spark` — but the gate's path rules do not know that, and the `BOUNDARY_FILES` machinery deliberately does not exempt itself, so the conservative verdict stands.

Collapsing the stack into one PR means **one campaign instead of three**. Nothing is dropped: every commit, message and measurement moves across intact.

Closing this in favour of #731.

PR #731, opened 2026-08-22
GDN recompute_wu: revive 3 dead gates, then 2.12x the kernel

tbraun96:
Stacked on **#730** — base is `gdn-wu-parallel-scan`, not `main`.

## The change

The two forward-substitution passes are **79–85% of this kernel** (measured against a
solve-removed probe). They were **left-looking**: each output re-read every earlier
`u[l]`, about `CHUNK^2/2 = 2048` local-memory reads per thread. **Right-looking** keeps a
running accumulator per row and pushes each solved block's contribution into the
remaining rows **once** — ~320 accesses, 6–8x fewer — with the block's solved values held
in **registers**.

Same math, same FMA count, **no extra shared memory**, ~25 more registers (3 -> 2 CTAs/SM).

## Why this and not something else

The change followed from arithmetic, not intuition. At nt=64 the kernel runs at **~3% of
FMA peak** — so it is not compute-bound — while issuing ~1 local load per FMA, about
1.07e9 per launch or **~3.6 loads/cycle/SM**, which is at the L1 limit. It is
**load-bound**, so the lever is touching each accumulator fewer times.

The same arithmetic **killed an explicit in-place triangular inversion before it was
built**: inversion cuts *total* FMAs 4x but leaves *per-thread* work unchanged, which buys
nothing on a load-bound kernel while adding two GEMMs. That was several hours saved by
measuring first.

## Measured, n=3, isolated per-launch at nt=64

| | runs (µs) | mean |
|---|---|---:|
| left-looking | 2408.6 / 2418.0 / 2427.1 | 2417.9 |
| right-looking | 1238.2 / 1255.3 / 1233.2 | **1242.2** |

**1.95x**, spreads under 1%, no overlap.

Wins at **every** shape, with no small-shape regression from the register cost — which was
the specific risk, since 3 -> 2 CTAs/SM could bite where the GPU is underfilled:

| nt | 1 | 2 | 4 | 16 | 64 |
|---|---|---|---|---|---|
| speedup | 2.28x | 2.47x | 2.03x | 2.04x | 2.12x |

(Those are against the **original** kernel, i.e. including the gc-scan change from #730:
2661.0 -> 1254.3 µs at nt=64.)

Gate B passes at cos 0.999999 on both U and W at all five shapes, with **max-abs identical
to baseline** — not close, the same values. For a re-associated sum that is a strong
signal, and still not sufficient on its own; see Correctness.

### In-serve context

This kernel is 41.2 ms of a 105.7 ms spine, so ~2x is **~20 ms off the spine** — roughly a
fifth of the FlashQLA gap from one change. The numbers above are **isolated per-launch
measurements on one box and are not in-serve figures.**

## IN-SERVE CONFIRMED (dgx1, same box, back-to-back, one variable: the binary)

Isolated per-launch numbers only rank variants against each other. This is the serve
path, with `ATLAS_PROFILE_FIRST=1` reading the per-phase lines directly — end-to-end TTFT
would bury a ~48 ms change inside run-to-run noise.

| phase | `main` (4f6c7d8f1) | **#731** (11150ae16) | change |
|---|---:|---:|---|
| `gdn_fla_recompute_wu` | 87.5 ms | **39.7 ms** | **2.20x — -47.8 ms** |
| `gdn_fla_chunk_delta_h` | 101.2 ms | 102.8 ms | +1.6 ms (control) |
| `gdn_fla_chunk_fwd_o` | 67.5 ms | 67.1 ms | -0.4 ms (control) |

**The two untouched phases are the controls, and both moved <1.6%**, so the leg is not
contaminated and the 2.20x is attributable to this change. Had they moved, the leg would
have been discarded rather than partially reported.

- **GDN spine total: 256.2 ms -> 209.6 ms (-18.2%)**
- end-to-end request (3 reqs, ~4000 prompt tokens, `max_tokens=8`, temp 0):
  3.931 s -> 3.849 s, **82 ms faster (2.1%)** — consistent with the phase delta plus noise.

In-serve 2.20x is slightly *better* than the isolated 1.95x, and the absolute saving is
larger than the ~20 ms projected: the in-serve baseline for this phase is 87.5 ms, not the
41.2 ms carried in older notes.

## One implementation trap, recorded in the kernel header

The loop variable `r` in the diagonal-block loop **must stay compile-time**. My first
version used a runtime bound, which made `xb` dynamically indexed and put it in **local
memory** (ptxas: 320-byte stack frame vs 256) — defeating the entire point of the change
while still compiling and still passing correctness.

## Correctness

**Not bit-identical**: contributions are accumulated per-block rather than interleaved row
by row, which re-associates the sum. cos and max-abs are unchanged, but **the
ssm-poisoning gate is the one that decides this** — the `vfused` note in this file records
that cos >= 0.99 *and* byte-identical single-prompt greedy both missed SPLIT=4's 1.4-point
BFCL drift. I am not treating the cos pass as sufficient.

### DIFF
--- a/.benchmarks/agentic-webserver/2026-08-23-3467a6298f.json
+++ b/.benchmarks/agentic-webserver/2026-08-23-3467a6298f.json
@@ -1,0 +1,467 @@
+{
+  "schema": 1,
+  "benchmark_id": "agentic-webserver",
+  "benchmark_name": "Agentic Webserver Test",
+  "git_sha": "3467a6298f",
+  "recorded_at": 1787521254,
+  "target_model": "Qwen/Qwen3.6-35B-A3B-FP8",
+  "params": {
+    "build_timeout_s": "600",
+    "command_timeout_s": "180",
+    "iterations": "10",
+    "max_tokens": "8192",
+    "max_turns": "40",
+    "request_timeout_s": "900",
+    "s_per_turn_budget": "8.5",
+    "serve_timeout_s": "30",
+    "wall_budget_s": "1800"
+  },
+  "command": [
+    "spark",
+    "benchmark",
+    "run",
+    "agentic-webserver",
+    "--param",
+    "build_timeout_s=600",
+    "--param",
+    "command_timeout_s=180",
+    "--param",
+    "iterations=10",
+    "--param",
+    "max_tokens=8192",
+    "--param",
+    "max_turns=40",
+    "--param",
+    "request_timeout_s=900",
+    "--param",
+    "s_per_turn_budget=8.5",
+    "--param",
+    "serve_timeout_s=30",
+    "--param",
+    "wall_budget_s=1800",
+    "--yes",
+    "--pull-request-gate"
+  ],
+  "served_by": "qwen3.6/qwen3.6-35b-a3b-fp8-bf16head",
+  "atlas_version": "1.0.0-beta-preview",
+  "hardware": {
+    "gpu": "NVIDIA GB10",
+    "driver": "580.126.09",
+    "sm_clock_mhz": 2405.0,
+    "source": "nvidia-smi"
+  },
+  "hardware_state": {
+    "sensitivity": "speed",
+    "before": {
+      "captured_at": 1787520339,
+      "machine": {
+        "hostname": "spark-43fa",
+        "machine_id": "7af66f30966a49b6886e00e2fce4b42f",
+        "gpu": "NVIDIA GB10",
+        "driver": "580.126.09"
+      },
+      "gpu_temp_c": 61.0,
+      "chassis_temps_c": [
+        {
+          "name": "acpitz",
+          "temp_c": 70.2
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 63.5
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 70.2
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 62.0
+        },
+        {
+          "name": "acpitz",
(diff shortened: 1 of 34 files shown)


tbraun96:
**Bisect complete — regression root-caused, fixed, and validated. The 2.20x survives.**

| build | bfcl-subset (golden n=995) | verdict |
|---|---|---|
| clean base (no pscan, no right-looking) | 84.22 / 84.12 | PASS |
| tree pscan only | 83.02 / 80.72 | FAIL |
| tree pscan + right-looking | 83.02 / **80.72** | FAIL — *identical* |
| **order-stable scan + right-looking** (`a6c45e5e1`) | **84.22 / 84.12** | **PASS — identical to clean base** |

The Hillis-Steele tree scan was the *entire* 2.6-point regression: it re-associates a log-space cumulative sum that feeds `exp(gc_i - gc_l)` across the whole L matrix. Right-looking substitution contributes nothing to the drift — its re-associated sums round to identical bf16, matching its max-abs-identical harness results.

The fix keeps the expensive parts parallel (64 global loads + 64 `logf`, order-free) and runs the 64 additions sequentially on one thread — **summation bit-identical to the original serial scan by construction**. Perf retained: **1236 µs at nt=64 vs 1254 with the tree scan and 2661 original (2.15x)**, cross-checked on a second box.

This is the second instance (after `vfused` SPLIT=4) where cos 0.999999 with identical max-abs missed a multi-point BFCL drift on this file. Recorded in the kernel header: task-level gates are the only trustworthy verdict for GDN numerics changes.

Re-campaign running on the fixed head.

PR #732, opened 2026-08-22
Fold TheTom's 39 audit/test PRs into one stack

tbraun96:
Folds **all 39** of @TheTom's open `codex/test-*` and `codex/audit-*` PRs from 2026-08-21
and 2026-08-22 into one branch, for a single-pass merge instead of 39.

## Why one PR

Each of these independently owes a **ten-gate benchmark campaign** (~8 h of GPU) because
the path rules see a `crates/` diff. Merged separately that is 39 CI runs, 39 chances for
`main` to move underneath the others, and 39 campaigns. Merged as one stack it is one of
each.

## What was done

- **Cherry-picked, not squashed** — `TheTom` is the sole author of all 46 commits.
  A squash-merge would have erased that; attribution is the point, not bookkeeping.
- **Applied in ascending PR order.** Later PRs in a themed run were written against the
  earlier ones, so numeric order is the closest thing to a declared dependency order.
- **Merge commits skipped.** Many of these branches were refreshed by merging `main` in,
  so their heads are merge commits and `cherry-pick A..B` refuses them
  (*"is a merge but no -m option was given"*). Those merges carry no content of their own
  — it is `main`'s, already present — so non-merge commits were picked individually.

## Result

| | |
|---|---|
| PRs folded | **39 / 39** |
| conflicts | **0** |
| commits | 46 |
| files touched | 23 |
| distinct authors | `TheTom` |
| `cargo test --workspace` | **5,336 passed, 0 failed, 0 errors** |
| `cargo fmt --check` | clean |
| file-size cap | no violations outside the workflow's allow-list |

Zero conflicts was not the expected outcome — the GGUF cluster (#683-#696) touches
overlapping files and looked like the likely collision point. It folded clean.

## Verification

Run in an isolated worktree off current `origin/main`, so nothing here depends on any
other in-flight branch.

## The 39 PRs

- #671 atlas-core: make compute target checks deterministic
- #672 atlas-core: strengthen quantization parser contract checks
- #674 atlas-core: cover expert partition remainders
- #675 atlas-core: strengthen Qwen3-Next fixture contract
- #676 atlas-core: preserve Gemma-4 config contracts
- #677 atlas-core: validate Nemotron-H config geometry
- #678 atlas-core: strengthen Qwen3.5 nested config contract
- #679 atlas-core: strengthen Qwen3-VL config contract
- #680 atlas-core: strengthen Qwen3-Next factory contract
- #681 atlas-core: partition kernel target equality
[...]

PR #739, opened 2026-08-24
test: RST audit of benchmark, model, runtime, and storage contracts

TheTom:
## Scope

This is the consolidated Rapid Software Testing audit requested after the earlier audit PRs were folded into one stack. It is updated through `main` at `a046cdfd`, including #731, #732, and #738. The branch contains 207 focused audit and cleanup commits plus one base-merge commit across 253 files.

Each audited declaration was checked against five questions:

1. What exact claim does the test make?
2. Does its setup execute the named production behavior?
3. Does its oracle observe the runtime-relevant result?
4. Can a realistic known-bad change survive?
5. Is another test already rejecting the same defect?

A passing test was not accepted as evidence by itself. For repaired tests, the old test was replayed against a named mutant, the mutant was shown to survive, the oracle or missing partition was repaired, and the same mutant was required to fail. Redundant and decorative declarations were removed when they rejected no distinct production defect.

## Review map

| Area | Audit result | Suggested review focus |
| --- | --- | --- |
| Agent tools and loop | File containment, bounded reads, shell draining/timeouts, recovery turns, request serialization, compaction, and trajectory identity now reach the real dispatcher or transport path. | `8801e618` through `c0443e28` |
| Benchmark harnesses | BFCL, contamination, MLPerf, concurrency, serve matrix, SSM poisoning, TTFT, video, vision, coherence, and quick-speed checks are tied to committed configuration and measured evidence rather than duplicated literals or test-authored replays. | `0d367bc1` through `1cd98dd2` |
| Gate and provenance | Coverage mapping, override policy, CODEOWNERS matching, record identity, dirty-tree provenance, hardware evidence, redirected kernel closures, and telemetry now have fail-closed partitions. | `20020404` through `be973577` |
| Plugin services | HTTP framing, parameter validation, history races, lifecycle output, registry/result identity, and end-to-end selection received direct consumer oracles. | `86afe04c` through `069ba54f` |
| RDMA and tiering | Count boundaries, cursor isolation, setup failure, resource reuse, and rollback integrity are exercised without claiming unavailable verbs or GPU transfers. | `5d6b1178` through `cbdd73c9` |
| Model and runtime | Loader routing, dispatch, quantization, LoRA, MTP, SSM allocation/spill, graph keys, multimodal positions, sharding, logprobs, video preprocessing, and still-image preprocessing were audited through production consumers or explicit mock-device effects. | `8f790f1b` through `a97a37d2` |

The commit sequence is intentional. Test-only oracle repairs use `test(...)`; behavior changes use subsystem or `fix(...)` subjects; mechanical CI cleanup is isolated in `d78ba065`.

## Production behavior changes

The stack is not test-only. The main functional corrections are:

- Agent file writes now reject dangling-symlink escapes, empty edit matches are rejected before mutation, timestamp-like identifiers are preserved during normalization, shell output is drained to completion, scorer ports remain reserved through compilation, and trajectory records include tool-call IDs.
- Benchmark inputs and results now fail closed on miskeyed baselines, incomplete or invalid evidence, unstable contamination references, malformed serve results, unmeasured SSM rounds, invalid TTFT targets, video-score errors, embedded coherence answers, and non-finite hardware measurements.
- Gate provenance now aborts when dirty-tree state cannot be read, override and record identity checks are stricter, CODEOWNERS matching states its supported glob contract accurately, and redirected kernel targets attest the same sources and configuration the build actually consumed.
- Model fixes cover checked MTP metadata sizes, aligned LoRA overlay rows, RDMA landing-plan validation, empty LoRA cache visibility, contiguous SSM ownership, spill-cap preservation on overwrite, distinct graph-borrow log pairs, release-mode verify-key and SSM-width contracts, empty-vocabulary logprobs, inverted video frame bounds, and declared image pixel caps after grid quantization.

Each functional change has its demonstrated old behavior, repaired oracle, mutant replay, and remaining non-claims in its focused commit message.

## Redundancy and scope corrections

The audit removes tests that only restated local arithmetic, downcast a self-authored mock, duplicated a stronger table, checked constants without reaching a consumer, or used identical fixtures that made the named behavior unobservable. Names and comments were also narrowed where a check proved ascending rather than consecutive slots, bounded output rather than bounded allocation, or portable absence handling rather than successful codec execution.

## Verification

Observed on macOS with Rust 1.93.1:

```text
cargo fmt --all -- --check
cargo test -p spark-model --no-default-features --features metal --lib video_preprocess -- --nocapture
cargo test -p spark-model --lib --no-default-features --features metal vision_preprocess::tests:: -- --nocapture
cargo test -p spark-model --no-default-features --features metal --lib factory::tests -- --nocapture
cargo test -p spark-model --no-default-features --features metal --lib decode_and_marconi_tiers_do_not_cross_serve_on_one_peer -- --nocapture
cargo test -p spark-server --no-default-features --features metal --bin spark cli::bench_run::provenance_tests -- --nocapture
cargo clippy -p spark-model --no-default-features --features metal --lib -- --no-deps -D warnings
cargo clippy -p spark-server --no-default-features --features metal --bin spark -- --no-deps -D warnings
cargo clippy -p spark-model --lib --no-default-features --features metal --no-deps -- -D warnings
```

The latest completed GitHub run before `d78ba065` passed workspace tests, macOS Metal tests, formatting, coverage, docs, security, license headers, file-size policy, and other required checks. Its five Clippy diagnostics are corrected in `d78ba065`; the new run is authoritative for Linux workspace Clippy.

Observed on the NVIDIA GB10 Spark with driver 580.159.03, compute capability 12.1, CUDA 13.0, and Rust 1.93.1 at exact pushed SHA `be973577631f6063f1cdfe3a707b31393803db4a`:

```text
cargo test --workspace
5,365 passed; 0 failed; 69 explicitly ignored
4,086 declared kernel builds -> 471 unique NVCC invocations across 24 GB10 targets
closure attestation: 24 target(s) reproduced from the tree
cargo test -p atlas-kernels --lib -- --ignored --nocapture
6 passed; 0 failed
cargo test -p atlas-kernels --test kernel_arity -- --ignored --nocapture
1 passed; 0 failed
```

The first real CUDA run found that the build followed `qwen3.8-27b`'s `kernel_source = "qwen3.6-27b"` redirect while the gate verifier did not. The old implementation failed the exact closure-attestation integration test on that target. Commits `be9e7673` and `be973577` align source resolution, compiled configuration, target enumeration, and collision telemetry; replaying the same real build now reproduces all 24 target hashes. The seven ignored checks whose own declarations require a real NVCC build were then run explicitly and passed. The remaining 62 ignored declarations were not executed by this CUDA campaign. The complete workspace log is 5,918 lines with SHA-256 `2d257a5c23d686a80954af71d2bce14a80ad5f41252322afb87020296dd405d9` on the Spark at `/tmp/pr739-cuda-workspace-be973577.log`.

The benchmark gate reported all ten mandatory records stale because this consolidated stack includes real production changes across plugin and model performance paths. On the ready-for-review SHA, the newest authentic records are for `e0a634261b`; the gate reports 191 to 207 invalidating paths per record. This is not a single CI implementation failure or a test-only false positive. The branch differs from the measured tree in 130 obviously non-test production files, including `crates/atlas-kernels/build.rs`, benchmark drivers and scorers, model dispatch, SSM storage, LoRA, video preprocessing, and scheduler code. Any one applicable production path is enough to invalidate its gate.

```text
agentic-webserver
vision-fidelity
video-fidelity
ttft-warm-gate
ttft-cold-gate
bfcl-subset
bfcl-subset-echolp
ssm-state-poisoning-gate
decode-floor
concurrency-sweep
```

Those records require authentic runs on the declared hardware. No benchmark JSON was created or inferred from Mac results.

The record-selection warnings above the final verdict are not the cause. They discard older records measured with a different checkpoint. The newest `e0a634261b` records match each gate's currently declared subject and all carry `PASS`; they are rejected only because the measured code differs from this PR. The CI job checks the synthetic merge SHA, which is expected, and exits 1 after reporting all ten gates as `NONE` plus `intent: nothing recorded` for `governance/pr-739.jsonl`.

The current branch SHA is `270f6e754a7d38f6a05650e5bdb54a61a5c41859`. Any future benchmark campaign must freeze and check out exactly the then-current PR SHA before building and running the documented sweep:

```bash
for g in ttft-cold-gate ttft-warm-gate vision-fidelity ssm-state-poisoning-gate \
         decode-floor concurrency-sweep video-fidelity bfcl-subset \
         bfcl-subset-echolp agentic-webserver; do
  timeout 21600 ./target/release/spark benchmark run "$g" --pull-request-gate --yes
done
```

Validate and commit only the records produced by those runs. Any code or `main` change after the campaign requires freezing and measuring again under `docs/gate-queue-protocol.md`.

## Evidence boundaries

- Codec checks executed locally against ffmpeg 8.1, including H.264 and HEVC. CI hosts that lack those tools skip the codec-dependent bodies and provide weaker evidence.
- The real CUDA workspace run proves CUDA compilation, registration, and the test bodies that execute under that build. Mock-GPU checks still prove only allocation, copy, shape, byte, and error contracts; the workspace run does not turn those mocks into evidence of all-reduce, RDMA transfer, kernel numerical parity, or model output.
- Request and scoring checks prove constructed payload and verdict policy. They do not establish live-model quality.
- Linux process-group tests and the GB10 CUDA workspace suite ran successfully on the Spark. No authentic benchmark campaign or live-model inference run is claimed.
- The latest vision batch ran on macOS with Metal only. Its twenty-five checks reject an over-cap grid snap, an ignored spatial merge in the live `VisionItem::pad_count` consumer, zeroed pixel normalization, collapsed decode diagnostics, and ignored EXIF rotation. It does not claim CUDA encoder execution or live multimodal output.

## Latest vision preprocessing batch

Commit `9177c738` fixes a functional sizing defect. For input `(85,2740)` with a 262,144-pixel declaration, nearest-grid rounding previously returned `(96,2752)`, or 264,192 pixels. The selector now takes the closest aspect-preserving smaller grid step until the quantized area fits, returning `(96,2720)`.

Commit `a97a37d2` repairs the adjacent test contracts. Four pad-count tests called an unused duplicate helper instead of `VisionItem::pad_count`, which request preparation actually consumes. An ignore-merge mutant left all four old checks green and now fails 4096 versus 1024. The valid-image check now observes exact normalized channel layout, malformed transport and invalid image bytes require distinct diagnostics, and an asymmetric EXIF fixture makes rotation order visible. The commit removes the unused public `image_pad_count` helper, so reviewers should treat that deletion as an intentional API-surface change.

## Latest DeepSeek-v4 attention-sink batch

Commit `f4ca2f7d` replaces three test-local dtype demonstrations with four checks that execute `load_attn_sink_f32`. A real `WeightStore` and mock device memory now bind FP32 pointer identity, BF16 widening into a distinct allocation with exact readback bytes, missing-tensor null behavior, and fail-closed unsupported dtype diagnostics. Returning the BF16 source pointer, returning null for FP32, propagating missing-key failure, and accepting FP8 each fail. Production behavior is unchanged, and CUDA kernel consumption is not claimed.

Commit `3d53cc3c` applies the loader-level contract to compressor APE while preserving its distinct required-tensor policy. The replacements require FP32 pointer identity, exact BF16 widening into new mock-device storage, a named missing-key error, and fail-closed FP8 diagnostics. Treating APE as optional or bypassing any dtype branch fails. Production behavior is unchanged, and CUDA compressor numerics are not claimed.

Commit `2ef10446` replaces a DFlash config test that silently passed without a fixed `/workspace` Hugging Face cache. The embedded shipped-shape fixture always executes, with separate partitions for runtime defaults and malformed-field context. Changing the default block size or removing parser context fails. Production behavior is unchanged; checkpoint retrieval and draft acceptance quality are not claimed.

Commit `7b071dd3` removes a decorative Nemotron loader construction check. Two replacements execute real loader trait methods against a `WeightStore`, binding Nemotron embedding and norm paths, tied-head fallback, explicit-head precedence, TP support, and no-MTP behavior. Qwen-style path substitution and always-tied-head mutants fail. Production behavior is unchanged; full layer assembly and CUDA output are not claimed.

Commit `395bafbc` connects the rowwise FP8 shape predicate to projection metadata. A real store now proves FP8 dtype, rank-two geometry, and scale presence gate the loader route; removing the guard admits BF16 and fails. The retained `[1,N]` partition rejects a naive element-count classifier, while a repeated four-size `[N,1]` loop is removed as redundant. Production behavior is unchanged; weight movement and GEMM numerics are not claimed.

Commit `043efc80` repairs two NVFP4 detector fixtures. The former EP-rank test was masked by the global suffix fallback and stayed green with the historical expert-index-as-layer bug. An FP8-typed layer-zero marker now isolates the prefix loop and fails on rank 1 under that implementation. A separate unexpected-prefix fixture directly owns the suffix fallback. Production behavior is unchanged; distributed loading and model output are not claimed.

Commit `1168836e` removes `test_weight_name_patterns`, which compared three test-authored `format!` expressions with adjacent test-authored literals. It reached no production key builder, loader, or store lookup, so every real naming defect survived. The empty module is removed with it; replacement coverage must be store-backed at each loader consumer. Production behavior is unchanged.

Commit `2dcf6d11` hardens the SwiGLU clamp source policy. The old exact-string scanner missed a real no-space `SWIGLU_LIMIT=10.0f` clamp in the shared CUDA kernel. A lexical scanner now detects the code identifier across spacing while ignoring comments and strings, and the known exception is limited to one relative path. Two narrower checks are removed because the recursive scope policy rejects the same defects. Production behavior is unchanged; compiled CUDA numerics are not claimed.

Commit `5944facc` makes the NLLB Metal integration test fail closed when no modules are registered. The old path printed a warning and returned green without a launch. Under the forced `metal/nllb-200-3.3b/bf16` target, 43 kernels compiled and the complete numeric body passed on this Mac. Ignoring linear bias fails 4 versus 14. Target-specific Clippy also passes. This is real Apple Metal kernel evidence, not CUDA parity or translation quality.

## Base and review state

The branch merges canonical `main` at `6fe778e4`. Git ancestry reports 220 commits ahead and 0 behind; one of those commits is the explicit base merge. The merge was clean, and the authentic benchmark records added by #731 and #732 are preserved unchanged from `main`.

Recommended review order: production changes first, then the adjacent mutant-proof tests, then removal-only commits. The PR is open for code review. The refreshed GitHub checks are authoritative for the new head, and the authentic benchmark-record requirement remains a merge blocker until a current-SHA campaign can be scheduled on the exclusive GPU host.

## Latest NLLB config batch

Commit `c73cd4e6` adds functional validation for attention geometry the CPU runtime cannot execute correctly. Parsing now rejects zero model width or head counts, fractional head width, and unequal encoder/decoder head layouts. Before the fix, the zero-head fixture returned a complete config and would later panic in `head_dim`; decoder attention also silently used the encoder head count.

Commit `f4bcdf8f` rejects activation names the runtime does not implement. The parser formerly accepted `gelu`, while the feed-forward path always executed ReLU. Unsupported values now fail at the config boundary instead of silently selecting different behavior.

Commit `49c2b3ca` repairs the original parser oracle. Changing the runtime-consumed pad default from 1 to 99 left the old `parses_minimal_config` green. The replacement binds runtime dimensions, token defaults, and embedding scaling. A separate explicit-values partition rejects an implementation that ignores `scale_embedding=false`, failing 32 versus 1. All 13 package unit tests and package Clippy pass on macOS. Null `max_length` handling is claimed only as parser compatibility because that field has no runtime consumer. The four reference integration declarations self-skipped without `NLLB_MODEL_DIR` and remain outside this evidence.

## Latest NLLB checkpoint registration batch

Commit `6c63a803` converts four false-green checkpoint tests into explicit ignored campaigns. Previously, an unset `NLLB_MODEL_DIR` printed a skip message and returned, so Cargo reported four passes without loading a model. Ordinary execution now reports 0 passed and 4 ignored. An explicit ignored run without the checkpoint fails closed, and the README gives the exact command. No matching checkpoint was found on this Mac, so this commit claims correct accounting and setup behavior only, not encoder, generation, beam, or LoRA numerical validation.

## Latest NLLB GGUF batch

Commit `e9d1ce2f` fixes three malformed-file panics in the standalone GGUF reader. A maximum metadata length overflowed cursor addition, overflowing dimensions panicked during element-count multiplication, and a maximum tensor count panicked during directory preallocation. File-controlled conversions, offsets, byte counts, and ranges now use checked operations, and the parser does not reserve from an unproven tensor count. The existing mixed FP32/FP16 and bad-magic tests reject direct reader mutants. A new 256-byte alignment partition catches ignored metadata that the former 32-byte fixture could not distinguish from the fallback. All 17 package unit tests and package Clippy pass on macOS. Real shipped GGUF loading and translation output are not claimed.

## Latest NLLB LoRA batch

Commit `6172023c` fixes a loader inconsistency where scaling used `adapter_config.json.r` while matrix multiplication used the tensor rank. A rank-4 config with rank-2 tensors formerly loaded and silently applied mismatched math; it now fails with both ranks named. A separate rank-4 rsLoRA fixture proves alpha is divided by square-root rank. Using ordinary alpha-over-rank scaling leaves the existing standard roundtrip green and fails the new partition at 8 versus 16. Direct mutants also confirm the five existing LoRA checks own distinct scaling, module lookup, zero-B, PEFT key, and disk-loader contracts. Seven focused checks and package Clippy pass on macOS. Full-model output remains pending the explicit checkpoint campaign.

## Latest NLLB GGUF name batch

Commit `67669507` tightens the GGUF-to-Hugging Face name grammar. Encoder cross-attention, nonnumeric layers, and arbitrary suffixes formerly mapped to plausible keys the runtime never requests, and could make an otherwise unusable GGUF satisfy the loader nonempty check. Mapping now requires a numeric layer, a `weight` or `bias` suffix, and decoder-only cross-attention. The existing table rejects an output-projection routing mutant, while the new negative table binds all three grammar boundaries. Twenty package unit tests and package Clippy pass on macOS. Real shipped GGUF loading is not claimed.

## Latest decode metadata batch

Commit `74baeab8` tightens decode metadata policy evidence. The old ceiling check accepted 129 rows because it required only a maximum of at least 64, even though the same constant directly gates serve admission and is documented as 128. The replacement binds the 32-row floor and 128-row ceiling exactly. A separate exact-64 offsets test is removed because the widened arithmetic test already executes 64 rows and asserts the same fields through their defining relationships. Floor, shifted-offset, and 129-row mutants fail the three retained checks. Focused Metal tests pass on macOS. Package Clippy is blocked by unrelated existing Metal FFI arity and thread-safety diagnostics. No Spark or CUDA execution was used.

## Latest runtime arena batch

Commit `743632a3` replaces an arena allocation-count proxy with field-level wiring evidence. The old test checked hidden/logits pointers and a total of 29 allocations; swapping QKV and attention assignments preserved every assertion. The repaired table binds every exposed field to its computed allocation size, requires distinct pointers or the documented null state, derives the allocation count, and verifies the unit RMSNorm buffer was zeroed. The same swap fails on QKV size 1,048,576 versus 2,359,296. The adjacent mixed dense/MoE and Qwen sizing checks reject routed-only and FP32-residual mutants. All seven buffer tests pass with Metal on macOS. Existing Metal package-Clippy diagnostics are unchanged.

## Latest Q2 scratch batch

Commit `9c27e737` removes an environment-dependent false green from Q2 scratch testing. The old disabled-path test returned whenever `ATLAS_GGUF_NATIVE_Q2=1`. The production environment wrapper now delegates to an explicit-boolean sizing function, so disabled, dequant-only, and MMQ-only partitions run deterministically without global environment mutation. Inverted polarity fails with a 50,331,648-byte allocation in the disabled case. The adjacent widest-projection test rejects removal of the fused GDN bound, failing 33,554,432 versus 50,331,648 bytes. All seven buffer tests pass with Metal on macOS.

## Latest CI cleanup

Commit `0e5945ca` fixes the only new actionable CI failure from head `9c27e737`: Linux workspace Clippy rejected a static NVFP4 fixture path wrapped in `format!`. The fixture now uses `to_string`, and the exact detector test passes locally. This is mechanical test cleanup with no production behavior change. The benchmark gate remains the previously documented stale-record blocker; PR categorization remains advisory.


TheTom:
Benchmark gate root cause confirmed from job `97439580293`.

The job successfully builds and runs:

```text
spark benchmark --pull-request-gate-check --pr 739
```

It finds a passing record for every required gate at measured SHA `e0a634261b`, then rejects those records as stale against this PR's synthetic merge tree. The reported blast radius is 191 to 207 invalidating paths per gate. The diff includes real production and benchmark-harness changes in `atlas-kernels/build.rs`, model dispatch and storage, LoRA, video preprocessing, scheduler code, and benchmark drivers/scorers. This is therefore not the test-only coverage problem fixed by #701, and a narrow test exclusion cannot make these records cover the PR.

The warnings about older qwen3.6, qwen3.8, Gemma, and vision variants are not the terminal failure. Those candidates are intentionally discarded because their checkpoint differs. The selected `e0a634261b` records use each gate's current declared subject and have verdict `PASS`; coverage fails because the measured code changed.

All other required checks are green. The remaining action is an authentic ten-gate run frozen at `be973577631f6063f1cdfe3a707b31393803db4a`, followed by committing the records produced by that run. The exact campaign command is in the PR description.

tbraun96:

PR #742, opened 2026-08-25
fix(spec): in-think EOS suppression on the verify emit path

rsafier:
The serial decode path (`decode_logits_step`) discards a greedy EOS that lands inside `<think>` (`thinking_suppresses_eos`); the speculative emit path (`emit_step`) never had that term — its own comment said so — so its in-think EOS discard only covered grammar-armed turns.

Invisible on Qwen (temp-0 argmax never lands on EOS inside think). Nemotron-3.5-Lightning's greedy reasoning opens by restating the prompt and then argmaxes `<|im_end|>`: serial discards it and reasons on for 2k+ tokens; spec emitted it and the response died ~30 tokens in — making speculation look like a verify divergence when both lanes' logits were identical.

Fix mirrors the serial path exactly (same `helpers::eos_suppressed_by_thinking` + hard-ceiling escape hatch).

**Validation**: Lightning + DFlash spec-think output is now byte-exact vs the serial serve on the standard MinHeap probe (content 6259 B / reasoning 2165 B, finish stop, identical bytes). Applies to every thinking model under speculation.

Standalone against main — no dependency on the Nemotron stack (#544/#545/#566), though it was found there.

### DIFF
--- a/crates/spark-server/src/scheduler/emit_step.rs
+++ b/crates/spark-server/src/scheduler/emit_step.rs
@@ -417,7 +417,27 @@ pub fn emit_token(
             || crate::grammar::grammar_blocks_stop(a.grammar_state.as_mut(), &a.eos_tokens));
     let legacy_suppresses_eos = a.require_tool_call;
     let min_tokens_suppresses = a.output_tokens.len() < a.min_tokens;
-    let suppress_eos = grammar_suppresses_eos || legacy_suppresses_eos || min_tokens_suppresses;
+    // Thinking EOS suppression — the twin of decode_logits_step's explicit
+    // `thinking_suppresses_eos` term, which this emit path NEVER HAD (the
+    // grammar term above only covers grammar-armed turns). Invisible on the
+    // Qwen family, whose temp-0 argmax never lands on EOS inside <think>;
+    // Nemotron-3.5-Lightning DOES — its greedy reasoning opens by restating
+    // the prompt and then argmaxes <|im_end|>, which the serial lane
+    // discards (reasoning continues, 2k+ tokens) and this lane previously
+    // honored, ending every speculative thinking turn at ~30 tokens
+    // (found on Lightning + DFlash, 2026-08-25). Same hard-ceiling escape
+    // as the serial twin so generation cannot overrun at the budget edge.
+    let hard_ceiling = crate::scheduler::helpers::hard_ceiling_hit(
+        a.remaining,
+        a.seq.seq_len,
+        sched.limits.max_seq_len,
+    );
+    let thinking_suppresses_eos =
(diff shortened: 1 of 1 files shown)


tbraun96:
**Included in stack `verify-and-spec` → #870.**

**Why it earns a merge:** a spec-decode correctness fix, and it cherry-picks onto current `main` cleanly — the change is genuinely absent there (three of its six siblings in this batch turned out to be already-landed, so I checked each rather than assuming).

**Stacked with:** #745 (sliding-window per-token metadata) and #693/#742 — whichever of the pair this isn't. All three touch `crates/`, so the group re-opens all 11 gates and owes **one** campaign between them instead of one each.

**Verified:** `cargo check --workspace --all-targets` on the composed stack → EXIT=0, zero errors.

Your commit is cherry-picked with authorship preserved. Close this once #870 lands, or say the word if you would rather land it directly.

tbraun96:

|---|---|---|---|---|
| 2 | #742 | in-think EOS suppression on the verify emit path | spec-verify | **← this PR** |

tbraun96:
**AutoMerger sieve verdict: `sieve:integrity` — excluded from certification until cured.**

Three-sieve static review (security / integrity / goal): SECURITY **pass** (single 20-line hunk in `crates/spark-server/src/scheduler/emit_step.rs`, scope exactly matches the title), GOAL **pass** (correctly mirrors the serial `decode_logits_step` suppression term the emit path lacked). INTEGRITY **fail**: the diff ships zero tests for a correctness-sensitive EOS-suppression change — nothing proves the new `thinking_suppresses_eos` term fires *only* when `inside_thinking` is set, that a normal out-of-think EOS still terminates, or that `hard_ceiling_hit` still escapes at the budget edge.

**What clears it:** a negative test covering those three cases. The fix itself looks sound; once the test exists this re-enters stack `certified-2026-09-04` (#877). If you'd rather I write that test, say so and I'll push it to this branch as a separate commit under my own identity.

tbraun96:
**Sieve re-run: CLEARED** (`sieve:integrity` → `sieve:cleared`). The demanded tests now exist on the stack top (`emit_eos_thinking_tests.rs`, head `82552fe34d`): all three drive the real `emit_token` over the shared fixture, map 1:1 onto the three demanded proofs, and the reviewer **mutation-tested the cure itself** — deleting the `|| thinking_suppresses_eos` disjunct fails exactly the in-think test with the committed message. Honest residual, recorded: the fixture reaches the suppression term with `grammar_state = None`; a grammar-armed thinking turn takes a different disjunct and is not covered by these three. This PR re-enters certification with stack #885.

tbraun96:

PR #744, opened 2026-08-25
dflash: support NVFP4-packed drafter checkpoints + packed shared lm_head

rsafier:
Two OOB classes found bringing up NVIDIA-Nemotron-3.5-Lightning DFlash (the first ModelOpt NVFP4 drafter export Atlas has loaded) — both generic to any NVFP4-packed checkpoint, no Nemotron dependency:

**1. Drafter weights.** ModelOpt exports ship `fc` + every attention/MLP projection as U8 `[n, k/2]` packed nibbles + F8_E4M3 `[n, k/16]` block scales + F32 `weight_scale_2` — under the **same tensor names** a BF16 checkpoint uses. The loader read them as BF16 → every GEMM read 4x past the allocation (CUDA 700, fault grid `[21,12,1]` = the fc GEMM). New `dense_bf16_or_nvfp4()` detects the packed form by dtype and GPU-dequantizes via the existing `dequant_nvfp4_to_bf16` kernel. BF16 tensors pass through untouched — existing drafters are byte-identical.

**2. Shared lm_head.** When the *target* ships an NVFP4-packed lm_head, `lm_head_shared` points at packed bytes and the drafter's Phase-G BF16 GEMM is the same 4x OOB (fault grid `[vocab/128,1,1]`). `lm_head_nvfp4` was stored but never consumed on this path — the tail now consumes it first via the w4a16 tile GEMM.

**Validation**: full numpy-from-safetensors parity audit on Lightning — dequantized fc GEMM cos 1.0000 (low-nibble-first E2M1 + E4M3 block scales × weight_scale_2) and w4a16 head logits cos 1.0000 with identical argmax vs the engine.

### DIFF
--- a/crates/spark-model/src/layers/dflash_head/forward_block.rs
+++ b/crates/spark-model/src/layers/dflash_head/forward_block.rs
@@ -567,7 +567,28 @@ impl BlockDiffusionDraftHead {
             // kernel (fixed: 2-round A-load covers all 32 K-cols).
             // BF16 path (default, or Fp8 mirror missing) unchanged.
             let lm_head_fp8 = matches!(self.quant, super::DflashQuantization::Fp8Weights);
-            if lm_head_fp8 {
+            // Target ships an NVFP4-PACKED lm_head (Lightning: U8 [vocab,
+            // h/2] + F8 scales): `lm_head_shared` then points at PACKED
+            // bytes and a BF16 GEMM on it reads 4x past the allocation —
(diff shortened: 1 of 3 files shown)


tbraun96:

Closing by hand because the commits were **re-cut**, not merged: this PR's head is not an ancestor of main even though its content is, so GitHub will not auto-close it. Verified by artifact rather than assumed — `dense_bf16_or_nvfp4` is on main in the DFlash head forward block.

PR #745, opened 2026-08-25
verify: sliding-window attention needs per-token metadata in every spec verify lane

rsafier:
Every verify body (K=2/3/4/γ, fused, and the eager GEMM-batched lane) routed non-FullAttention layers to `decode_batched`, whose default loop calls single-token `decode()` K times under the caller's **one** metadata upload. Attention `decode()` reads position/KV-slot/seq_len from that metadata, so on a **SlidingAttention** layer all K verify rows attend and write KV at the *same position* — committed tokens collapse to one token repeated, and graph-capturing lanes can CUDA-700. No shipped speculative target had sliding layers until Laguna, which is how this survived.

**Measured** (Laguna-XS-2.1-NVFP4 + DFlash drafter, temp 0): before — 'ToToToTo…' from the first content-phase propose while plain serial is clean, fused M=4 lane CUDA-700; after — zero CUDA errors, coherent content, K=γ p1 0.929 (6.0 accepted/step), K=2 at 41.9 tok/s vs ~40 plain.

**Fix**: extracted `verify_attention_per_token` (the per-token-metadata loop verify_a already used inline for FullAttention) with its own lazily-allocated 16 KB staging (staging at `scratch+32768` clobbers the multi-seq overlay — that region can span the whole scratch allocation); SlidingAttention arms in all six lanes; `use_graphs` off when sliding layers exist (per-token H2D is illegal under capture). Plus a latent-bug gate: the strided multi-seq rope fast path must not fire on YARN-scaled layers (it computes plain-theta rope with no yarn table).

**Not covered**: `verify_e` (cross-sequence C≥2 batched verify) needs a per-band variant of the helper; sliding models should keep the batch-verify lever off until then.

### DIFF
--- a/crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/attn.rs
+++ b/crates/spark-model/src/layers/qwen3_attention/trait_impl/multi_seq/attn.rs
@@ -51,7 +51,16 @@ impl Qwen3AttentionLayer {
             static ON: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
             *ON.get_or_init(|| std::env::var("ATLAS_NO_ROPE_STRIDED").ok().as_deref() != Some("1"))
         }
-        if n > 1 && self.rope_strided_k.0 != 0 && rope_strided_enabled() {
+        // YARN-scaled layers (yarn_inv_freq set -- Laguna full-attention) must
+        // take the per-seq loop below: rope_strided computes plain-theta rope
+        // and has no yarn table argument, so a K-row verify through it rotates
+        // Q/K differently than the serial decode path. Bit-parity for
+        // non-yarn layers is unchanged.
+        if n > 1
+            && self.rope_strided_k.0 != 0
+            && rope_strided_enabled()
+            && self.yarn_inv_freq.is_null()
+        {
             let stride_e = (per_seq_qkv / bf16) as u32;
             return ops::rope_strided(
                 fwd.gpu,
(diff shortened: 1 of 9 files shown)


tbraun96:
**Included in stack `verify-and-spec` → #870, with one rebase fixup you should look at.**

**Why it earns a merge:** sliding-window attention lanes genuinely need per-token metadata in every spec verify path, and the change is absent from `main`.

**It no longer compiled against main.** `ForwardContext` has gained two fields since 2026-08-25:

```
error[E0063]: missing fields `hc_row_offset` and `host_token_ids`
              in initializer of `layer::ForwardContext<'_>`
  --> crates/spark-model/src/model/trait_impl/verify_a.rs:599:23
```

I added `hc_row_offset: 0` and `host_token_ids: None`. **Not invented** — `hc_row_offset: 0` is what all six other construction sites use (`verify_a` ×2, `verify_c2`, `speculative`, `decode_a`, `decode_a2`), and `host_token_ids: None` matches the two siblings in this same file. It is a **separate commit** in #870 so the rebase is attributable to me, not folded into yours — please sanity-check that `None` is right for this lane, since you know what that field means here better than the call sites do.

**Stacked with:** #693 and #742. One campaign covers all three.

tbraun96:

|---|---|---|---|---|
| 3 | #745 | sliding-window per-token metadata in spec verify lanes | spec-verify | **← this PR** |

tbraun96:

PR #746, opened 2026-08-25
LongCat-Flash-Lite serves end to end: n-gram embeddings + MLA backbone (Qwen3.8-Flash-Next readiness)

rsafier:
## Qwen3.8-Flash-Next readiness: n-gram embeddings — and the LongCat proxy now actually serves

Groundwork ahead of the Qwen3.8-Flash-Next release. The announced model (125B MoE, 6B active, **+51B n-gram embeddings**, "Qwen4 architecture" preview) uses the LongCat/Meituan scaled-embedding design ([arxiv 2601.21204](https://arxiv.org/abs/2601.21204)). `meituan-longcat/LongCat-Flash-Lite` ships the same mechanism today, so everything here is built and validated against that checkpoint as a proxy. When Qwen drops, the expected delta is config-key aliases plus fixture regeneration.

**What changed since this PR was opened:** it started as the n-gram *engine* with the backbone port explicitly not started. The backbone is now ported, and the full 129 GB proxy model **serves coherently end to end and matches the reference forward at every stage.** That is the headline; the details below are mostly the four bugs that stood between those two states.

### The mechanism

`K×(N-1)` = 12 lookup tables (K=`emb_split_num`=4, N=`emb_neighbor_num`=4), each `ratio·vocab + 2i + 1` rows × 256 dims — ~10.2M rows apiece. Ids are a polynomial rolling hash over **token ids only** (no hidden state), with EOS-boundary resets. The fused input embedding is

```
x = (embed_tokens[t] + Σᵢ projᵢ(tableᵢ[hashᵢ(t, context)])) / (1 + K(N-1))
```

i.e. the base row is one of thirteen contributions, not the whole embedding. That ratio is why getting this wrong is dangerous: a model served on the base row alone produces fluent, confident, degenerate text rather than anything that looks like a failure.

### Validated end to end against the real checkpoint

`bench/ngram_ref/full_forward_cmp.py` compares the live server against committed goldens. Each LongCat checkpoint layer is **two** engine sublayers, so checkpoint layer K's output is engine sublayer 2K+2's input:

| stage | cos vs reference |
|---|---|
| fused n-gram embedding (sublayer 0 input) | **1.0000** |
| `layer0_out` … `layer12_out` (28 sublayers) | **all ≥ 0.9952** |

and the served answer is coherent and self-terminating:

```
The capital of France is:
A. Lyon    B. Marseille    C. Paris    D. Bordeaux
C. Paris
```

Concurrency, at the model card's own sampling (`repetition_penalty` 1.06 / `temperature` 0.7 / `top_p` 0.95 / `top_k` 4), scored on checkable factual questions:

| | serial C=1 | batched C=4 | batched C=8 |
|---|---|---|---|
| before the shortcut fix | 7/8 | 4/8 | — |
| after | 7/8 | **7/8** | **7/8** |

The single remaining miss is identical in every arm — the model doing 17×23 longhand and hitting `max_tokens` — so it is not a concurrency defect. A four-stream probe with unique per-stream nonces shows complete correct answers and **zero** nonces crossing between streams.

### Four bugs, and why each one hid

**1. The OOM pre-flight counted tensors it never uploads.** The n-gram tables are 62.8 of the checkpoint's 138 GB and the loaders *defer* them, but the pre-flight estimated peak from total on-disk bytes and refused the model outright (`peak 167.30 GB + 4.00 GB reserve exceeds 112.02 GB free`). Excluding deferred tensors gives 91.25 GB against 111.37 GB free. Fixed in all three pre-flight sites so the fast and mmap loaders agree.

**2. The padded head width never reached the attention chain.** LongCat is the first MLA model whose qk head (192) and v head (128) differ, and 192 does not compile — common's kernels need 4 or 8 BF16 per lane and 192/32 = 6. Both are padded to the stock 256 purely in the *weights*. But the attention chain takes its stride from `head_dim_override.unwrap_or(config.head_dim)` = 192, so Q, K, V and the o_proj input were each sliced 64 lanes short of where the weights put every head. The op-dump *file size* is what exposed it: 6144 floats where 32 heads × 256 should give 8192, and 6144 = 32 × 192 exactly. `post_attn_norm_out` cos 0.5439 → **0.9990**.

**3. The fused embedding was built and never called.** Nine embedding sites, not the eight first found; the missed one was `prefill_b/embed_chunk.rs`, *the* chunked-prefill embed — the path every prompt takes when `chunked_prefill` is on. Decode was correct while prefill was not, so the model looked wired and behaved unwired. Instrumentation settled it, not reading: the setter logged "installed", `embed_tokens_fused` logged `ngram=true ctx_len=4 seq_len=1` (decode only), and layer 0's dump was cos **1.0000 against the raw `embed_tokens` row** — the plain gather, not even divided by 13.

**4. Batched decode dropped the shortcut MoE entirely.** `multi_seq/ffn.rs` never implemented LongCat's shortcut MoE; `decode_inner` (single-sequence) does. So at *any* batch above one, every sequence lost the block's whole 256-expert shortcut contribution, on every layer, with no error. Attention still read the right KV, so each stream stayed perfectly on topic while the distribution fell apart — answers fragmenting into `Ven`, `Prap`, `Pow`, `J` where serial gave `Jupiter/Saturn/Uranus`.

### Load time: 148 s → 17.9 s (8.3×)

Worth recording because the obvious suspect was wrong. Runtime BF16→NVFP4 quantization is **~1.3%** of load — 1.96 s across 7,808 calls at 0.251 ms each. Phase timers found the real 94%: two MLA precompute phases, both shared with Mistral-Small-4 and DeepSeek-V4-Flash, which get the same win.

| | before | after |
|---|---|---|
| phase C (`W_QK_absorbed`) | 4.45 s/sublayer | **93 ms** |
| phase B (per-head views) | ~0.9 s/sublayer | **5.4 ms** |
| per layer | 10.55 s | **1.27 s** |

Phase C was a single-threaded scalar CPU GEMM (4.83e9 FP32 MACs/sublayer) re-decoding both BF16 operands inside the innermost loop; now widened once, `l`-innermost, row-parallel — 85.9× and **bit-exact** (0 mismatches over 25.2M outputs, asserted against the pre-change loop kept verbatim as an oracle). Phase B issued 10,240 row-at-a-time `copy_d2d` per sublayer, each ending in a full `cuStreamSynchronize`; each head is one contiguous run, so 64 async copies + one sync. A third fix removed 25.2M per-element `Vec` allocations from the BF16 pack (202 ms → 19 ms).

### Serving it

```bash
./serve_longcat_tui.sh          # port 8888 (what bench/agentic/* expects), TUI up
```

The 12 tables are served **off NVMe by a pinned row cache** — 403 MB resident (65536 slots × 512 B × 12) against 62.8 GB BF16 or 31.4 GB FP8, because a row is 512 bytes and a token touches exactly 12 of them. That headroom goes to KV instead. `ATLAS_NGRAM_CACHE_SLOTS` tunes it.

The model card's sampling is now the target default via `[sampling.*]` + `use_sampling_presets_for_core = true`. Without that second flag the presets apply only to non-core paths and a plain `/v1/completions` still gets Atlas's generic defaults — they look wired and are not.

### Also here

- `DECODED_TOKENS_TOTAL`, incremented per token, so the TUI's tok/s is a live rate. `GENERATION_TOKENS_TOTAL` only moves once per request from the final usage block, so differentiating it read 0 tok/s then one spike — the server was streaming correctly the whole time (139 SSE chunks over 5.8 s, median gap 41.8 ms).
- `ATLAS_NGRAM_DISABLE=1`, a bisection lever that serves the plain gather: wrong output, but *deterministic*, which is what answers "is this bug mine or older than the n-gram path?"
- Guards that fail loudly instead of degrading: `embed()` refuses a bare token id on an n-gram model, and `ms_phase_ffn` refuses a shortcut-MoE model reaching any batched arm that does not implement the shortcut.

### Known gaps

- **Prefix caching is untested here** and off by default; being evaluated separately before agentic work.
- **Speculative decode is blocked** for this family — the `verify_*` paths still call bare `embed()`, which the guard refuses. Harmless today (no drafter), needs `embed_ctx` before DFlash/MTP.
- **The batched shortcut MoE runs per token in a loop** — correct, and the slowest possible shape at batch *n*. Batching it across rows is the obvious follow-up; decode throughput is not yet characterised.
- **No calibrated quantization.** The checkpoint ships no NVFP4/FP8 metadata, so this runs the runtime BF16→NVFP4 fallback and warns on every layer. How much quality that costs is unmeasured — a BF16 A/B would settle it.
- Phase D is now the largest load item (~185 ms/sublayer). Its block-diagonals are ~97% zeros by construction, so memset-on-device plus strided copies would cut ~470 MB of H2D per sublayer to ~14 MB.

648 spark-model tests green.


rsafier:
### Note for later: this checkpoint ships an MTP module and we are dropping it

Not addressed in this PR — recording it so the next person does not have to rediscover it.

`LongCat-Flash-Lite` ships **17 tensors under `model.mtp.*`**, and our loader currently returns `Ok(None)` for them with the comment *"the MTP head shape is not the Qwen-style one Atlas builds"* (matching HF's own `_keys_to_ignore_on_load_unexpected = [r"model\.mtp.*"]`). That comment is true but understates how close it is.

```
model.mtp.embed_tokens.weight                             [131072, 3072]   its OWN embedding
model.mtp.norm.weight                                     [3072]
model.mtp.layers.0.enorm.m.weight                         [3072]           norm on next-token embed
model.mtp.layers.0.hnorm.m.weight                         [3072]           norm on prev hidden
model.mtp.layers.0.eh_proj.weight                         [3072, 6144]     FUSED concat projection
model.mtp.layers.0.input_layernorm.weight                 [3072]
model.mtp.layers.0.post_attention_layernorm.weight        [3072]
model.mtp.layers.0.self_attn.q_a_proj.weight              [1536, 3072]   ┐
model.mtp.layers.0.self_attn.q_a_layernorm.weight         [1536]         │
model.mtp.layers.0.self_attn.q_b_proj.weight              [6144, 1536]   │ MLA
model.mtp.layers.0.self_attn.kv_a_proj_with_mqa.weight    [576, 3072]    │
model.mtp.layers.0.self_attn.kv_a_layernorm.weight        [512]          │
model.mtp.layers.0.self_attn.kv_b_proj.weight             [8192, 512]    │
model.mtp.layers.0.self_attn.o_proj.weight                [3072, 4096]   ┘
model.mtp.layers.0.transformer_layer.mlp.gate_proj.weight [6144, 3072]   ┐
model.mtp.layers.0.transformer_layer.mlp.up_proj.weight   [6144, 3072]   │ DENSE FFN (not MoE)
model.mtp.layers.0.transformer_layer.mlp.down_proj.weight [3072, 6144]   ┘
```

#### Why the "wrong shape" reason is narrower than it sounds

Against Atlas's generic `MtpWeights`, only **one field group** actually mismatches:

| `MtpWeights` | LongCat | |
|---|---|---|
| `pre_fc_norm_embedding` / `pre_fc_norm_hidden` | `enorm` / `hnorm` | ✅ |
| `fc` — doc'd as `[hidden_size, 2*hidden_size]` | `eh_proj` **[3072, 6144]** | ✅ exact |
| `input_layernorm`, `post_attn_layernorm` | same | ✅ |
| `dense_ffn` triple | `transformer_layer.mlp.*` | ✅ |
| `q_proj`/`k_proj`/`v_proj` + `q_norm`/`k_norm` | **MLA** (`q_a`/`q_b`/`kv_a`/`kv_b`) | ❌ |

Attention is the only blocker — and Atlas already has an MLA-bodied MTP path: **`DeepseekV4MtpModule`** (`weight_loader/deepseek_v4/mtp.rs`), which is body + `enorm`/`hnorm` + `e_proj`/`h_proj` + `norm`. LongCat fuses `e_proj|h_proj` into a single `eh_proj`, which is a column split of that same matrix, not a different design. **So the route is the V4-style module, not `MtpWeights`.**

#### What it would take

1. **A V4-style LongCat MTP loader** — reuse `Qwen3AttentionLayer` for the body (same MLA geometry as the backbone, so the same padded-head prep in `longcat/prep.rs` applies), split `eh_proj` into the e/h halves or teach the combiner the fused form.
2. **Fix the verify paths first.** `verify_a.rs:63`, `verify_b.rs:68-69`, `verify_c.rs:75`, `verify_c2.rs:64`, `verify_d.rs:70` all still call bare `self.embed(token, ..)`, which this PR's n-gram guard **refuses** — speculative decode on LongCat errors out immediately until they route through `embed_ctx` with the sequence history. Mechanical, same change already made for the decode paths here.
3. **Config wiring.** The checkpoint has *no* MTP keys (`num_mtp_modules` unset; no `mtp_layers` in the target MODEL.toml), so nothing triggers a load. The presence of `enorm` is the reliable marker — which is exactly what `load_v4_mtp_module` already keys on.

#### One thing to check, not assume

The MTP module has its **own plain `embed_tokens`** and no n-gram tables. Whether the drafter is meant to embed plainly or share the fused n-gram embedding should be read off the reference implementation before wiring it — getting that backwards is the same failure class that cost real time in this PR (a plain gather where a fused embedding belongs produces fluent, confident, wrong text and nothing in the log).

#### DSpark / DFlash

Not available here for a different reason: those need a **separate drafter checkpoint** (how Laguna XS/S reached p1 0.38 / 0.43 — community drafters plus config shims). No LongCat drafter exists, so **MTP is the only speculative option actually in the box** — and it is free weights we currently discard.

rsafier:
### Precision: what NVFP4 was costing, and three levers that buy it back

The PR body listed "no calibrated quantization … how much quality that costs is unmeasured" as an open item. It is now measured, and largely fixed.

#### The instrument that was missing

`full_forward_cmp.py` compares per-sublayer hidden states — that says the port is *wired* correctly, not what the *output distribution* costs. A 0.995 hidden-state cosine can still reorder the top of a 131072-way softmax, and reordering is what the sampler actually sees.

New: **`bench/ngram_ref/logit_quality.py`** compares Atlas's final logits against the reference's `logits_last` over the **full vocabulary** — top-1, top-k overlap, logit cosine, and KL. It reads the existing `ATLAS_NEMO_DUMP` hook rather than the API, because `/v1/completions` logprobs are keyed by decoded *string* and capped at 20: two ids that decode alike collide, and the tail is invisible.

#### Measured (16-token golden prompt; deterministic — bit-identical on repeat)

| arm | KL vs ref | logit cos | top-5 | top-100 | tok/s |
|---|---|---|---|---|---|
| default (all NVFP4) | 0.0464 | 0.998250 | 3/5 | 86/100 | 24.31 |
| **`ATLAS_LONGCAT_FP8_EXPERTS=1`** | **0.0344** | 0.998588 | **4/5** | 88/100 | **22.69** |
| `ATLAS_NVFP4_MLA=0` | 0.0382 | 0.998567 | 3/5 | 87/100 | 20.18 |
| `NVFP4_MLA=0` + FP8 experts | 0.0301 | 0.998799 | 4/5 | 88/100 | ~19.2 |
| `NVFP4_MLA=0` + `BF16_FFN=1` | 0.0240 | 0.999033 | 3/5 | 90/100 | 16.67 |

The baseline is not harmless. The reference gives its own top token p=0.0170; baseline Atlas says 0.0209 — 23% overconfident. And only **3 of the reference's top 5** survive in the baseline's top 5, while the model card recommends `top_k: 4`, so the sampler is drawing from a partly wrong shortlist.

**`ATLAS_LONGCAT_FP8_EXPERTS=1` strictly dominates the MLA lever** — better KL, better cosine, better top-5, *and 12% faster*. It is the one to reach for.

#### The counterintuitive part

Decode is weight-bandwidth bound (~185 GB/s effective). Each lever costs throughput in proportion to the extra weight bytes read **per token** — which is not the same as bytes resident:

| group | resident | added per token | why |
|---|---|---|---|
| routed experts → FP8 | 63.0 GB | **+0.74 GB/tok** | SPARSE — top-12 of 256 fire |
| MLA proj → BF16 | 1.71 GB | +1.28 GB/tok | dense, every token |
| dense FFN → BF16 | 2.94 GB | +2.21 GB/tok | dense, every token |

So the group holding **90% of the weights and nearly all of the error** is the *cheapest* to upgrade, and the little 2.94 GB dense FFN is the most expensive. Calibrating the model on the default-vs-both-BF16 pair then predicted the MLA-only arm at 20.80 (measured 20.18) and FP8-experts at 22.16 (measured 22.69) — two independent confirmations, so the ordering is structural rather than a quirk of one prompt.

#### Where the error lives

With MLA, dense FFN, `embed_tokens` and `lm_head` all BF16, the per-sublayer cosine drops once at layer 1 (0.9999 → 0.9986) and then stays **flat** through layer 12 (0.9984–0.9992). Nothing compounds with depth — that is a per-layer quantization floor, not a propagating bug, and the only component still quantized is the routed experts.

Two groups needed nothing, recorded so nobody re-checks: the **n-gram tables are already BF16** (the NVMe row cache serves raw rows, superseding the older FP8-quantize-on-load path), and **`embed_tokens` / `lm_head` are already BF16**.

#### What landed

- `quantize_bf16_to_fp8_blockscaled.cu` — the missing runtime quantizer. Atlas had BF16→NVFP4 and BF16→FP8 per-*row*; the FP8 grouped MoE GEMM reads **block** scales `[N/128, K/128]` FP32. A per-row buffer is not a shape error it can detect — it would index a shorter array and dequant with the wrong scale.
- `ATLAS_LONGCAT_FP8_EXPERTS=1`: NVFP4 expert slots become `ExpertWeight::null()` so both copies are never resident (+10.6 GB measured, 51.9 → 62.5 GB pre-KV, KV still 570k tokens); BF16 source freed after quantize.
- **Zero-expert routing on two more dispatch variants.** Enabling FP8 first produced a *named refusal* rather than wrong output — the guard working as intended — but it had to be wired: `forward_prefill_fp8` was sizing the gate GEMM at `num_experts` (256 of 384 logits → routing on a truncated distribution), and `forward_batched` is where every **sub-65-token** prefill lands, since `forward_prefill` only takes the grouped GEMM above 64 tokens. That is most conversational turns.
- The six zero-expert refusals now **name their dispatch variant**. They shared one message, so a refusal said the model was blocked but not where.

All three levers are **opt-in**; the default is unchanged, because −31% decode for the full stack is not a trade to make silently. `serve_longcat_tui.sh` carries the table.

`scoring_func == "softmax"` is set only by the LongCat parser (Laguna = sigmoid, DeepSeek-V4 = sqrtsoftplus), so the shared MoE routing arms cannot capture another model.

tbraun96:
Handoff note — no changes made to this branch or PR.

**#754 is ready to land into `feat/longcat-ngram-readiness`.** A ten-gate campaign ran against its exact head `0c402bac00` and all ten PASS; the records are committed at `bd2e7eaa21`. It was the *second* campaign for that PR — the first measured `1d30d5d5ff` for nine hours and died when `0c402bac0` landed twenty minutes into its final gate, because records are content-anchored over PERF_PATHS and all ten go at once. `scripts/campaign-guard.sh` (#785) now checks for that between gates and every five minutes during them, and the push was gated on the branch still being unmoved at the moment it happened.

**What this PR needs before it can follow, measured just now:**

| | |
|---|---|
| commits main has that this branch does not | **33** |
| perf-path files main changed since the merge-base (`a046cdfde`) | **304** |
| `git merge-tree` against main | **conflict-free** |

The 304 files mean every gate record on this branch dies the moment main is merged in, so a fresh ten-gate campaign is required *after* the merge, not before. On this box that is roughly **nine hours of exclusive GPU**.

**I deliberately did not start it.** Three reasons, in order:

1. This is your draft, and its description lists open gaps you wrote yourself — prefix caching untested, speculative decode hard-blocked (`verify_*` still calls bare `embed()`), batched shortcut MoE on a per-token slow path with decode throughput uncharacterised, BF16 quality A/B outstanding. Draft status reads as a decision, not an oversight.
2. Nine hours of exclusive GPU on a shared box needs coordinating, not assuming.
3. A single push to this branch during those nine hours destroys the campaign — which has already happened once here, at full cost.

**So the campaign is worth starting when two things are true:** you say the branch is settled, and there is a quiet window on the box. At that point the sequence is: merge main in, run all ten gates against the merged head, push the records, mark ready. `docs/gate-queue-protocol.md` and #783's runbook cover the serialized-landing details.

tbraun96:
Correcting one fact in my note above, and adding one.

**#754 is merged** — `qwen4_exp: Qwen3.8-Flash-Next config parser + checkpoint audit` landed on this branch at `73c34ac2b2` with all 31 checks green, including the benchmark gate against the ten records taken at `0c402bac00`.

**And the merge into main is no longer conflict-free.** My earlier note said `git merge-tree` was clean; that was true when I measured it and is not true now. Re-measured just now, against a *deepened* clone (a shallow one reports no merge base at all and every file as "both added", which is how I nearly mis-stated this a second time):

| | |
|---|---|
| merge-base | `a046cdfde` |
| main ahead of this branch | **39 commits** |
| this branch ahead of main | 36 commits |
| perf-path files main changed since the merge-base | **304** |
| `git merge-tree` | **CONFLICTS** |

Four paths conflict:

```
.gitignore
crates/atlas-plugin/src/benchmarks/agentic/agent.rs
crates/spark-model/src/layer/tests.rs
crates/spark-runtime/src/buffers/tests.rs
```

Nothing else changes: the 304 perf-path files still mean every gate record on this branch dies the moment main is merged, so a fresh ten-gate campaign — roughly nine hours of exclusive GPU on this box — has to come *after* the merge. The conflicts just add a resolution step in front of it, and three of the four are test files, so they want the author's eye rather than a mechanical union.

Still not started, for the reasons in the previous comment: this is your draft with your own stated gaps, nine hours of exclusive GPU needs coordinating, and a single push during those nine hours destroys the campaign — which has already happened once here at full cost.

tbraun96:
Closing: this PR's content is on `main`. It landed through #797 (`0a56d8f8b6`, "integration: Qwen3.8-Flash-Next stack (#746 + #754 + #648) — one certification scope").

Re-checked today against `main` `9e597531`: all 321 functions/types this branch adds exist there — e.g. `indexer_maps_onto_index_fields_and_only_full_attention_layers` at `crates/avarok-core/src/config/parsers/qwen4_exp_tests.rs:115`.

Thanks for the LongCat-Flash-Lite / Flash-Next readiness work — it is serving. No branch deleted.

PR #747, opened 2026-08-25
site: atlasctl launcher, PWA, Run button, dark theme, and a Lighthouse gate

tbraun96:
## What

Points the site at **`atlasctl`**, retires the sparkrun installer, turns the site into a PWA, adds the Run button and launch-settings modal, moves to the deep-violet dark theme, and adds a Lighthouse CI gate.

Pairs with [atlas-recipes#25](https://github.com/Avarok-Cybersecurity/atlas-recipes/pull/25) and [#26](https://github.com/Avarok-Cybersecurity/atlas-recipes/pull/26), both merged. `atlasctl` is live on crates.io at 0.1.0.

## Why the installer change could not wait

The front page told every visitor to run `uvx sparkrun setup install`. sparkrun 0.3.6 silently redirects the Atlas recipe registry to a repository under an org we do not control and marks it trusted, which lets recipe-supplied shell commands run on the host. Those are public instructions, so they were being followed.

`quickstart.sh` becomes a tombstone rather than a 404 — the old one-liner is in blog posts and people's notes — and it deliberately installs nothing on the visitor's behalf.

## Lighthouse: 100 / 100 / 100 / 100

A new CI job builds the site exactly as the deploy does and asserts 100 on all four categories.

The first run came back with accessibility at 96. Four real problems, all introduced or exposed by the theme change:

- `.subcard` still had `background: #fff` — a white card survived the inversion, putting near-white text on white at **1.12:1**.
- `--accent` as text on an `--accent-soft` tint measures 3.76:1 once the translucent violet composites over the dark canvas.
- The receipt chip put white on `--accent` at 3.57:1, and hovering made it *worse*.
- A WCAG 2.5.3 violation: a button's accessible name did not contain its visible text, so voice-control users could not say what they could see.

Locally computed contrast matched Lighthouse to two decimals, so these were the real fixes rather than guesses.

## The chart palette was re-derived, not tinted

`gates.js` carried three series colours validated for colour-vision deficiency against the paper surfaces, with a comment asking that all three be re-derived together if the palette ever changed. Inverting the background invalidates them — steel fell to 2.87:1, under the ≥3:1 floor that file has always held.

Re-derived with CIEDE2000 under Viénot dichromat simulation. The method reproduces the previous comment's figures exactly on the old surfaces, so the numbers are directly comparable:

| | worst-case separation | steel ↔ teal |
|---|---|---|
| before | 16.3 | 20.3 |
| after | **25.0** | **32.6** |

Lightness was capped during the search: an unconstrained optimum scored 31.4 but put teal at 14.4:1, a near-white cyan that no longer reads as teal.

## Run button and settings modal

Recipe cards gain a Run button that drives a local agent on `127.0.0.1:34333`. It names a recipe id and nothing else — no image, command, flag, path, or env — because the agent's message surface has nowhere to put one.

The settings modal renders **entirely from the schema the agent sends at the handshake**, with no per-key code, so an agent newer than the deployed site still presents its settings. State is the recipe's values plus a sparse diff, so an untouched modal launches byte-identically to `atlasctl run <recipe>`. The Review tab shows the command rendered by the agent's own renderer, so preview and execution cannot drift.

Every not-yet-working state is explained rather than reported as an error, since most visitors have no agent: install guidance, pairing, protocol mismatch, multi-node (disabled with a reason), and machine-cannot-launch.

## PWA

The existing service worker was left alone — it is already correct — with a ten-line addition so the install scripts are never cached. Someone piping the installer into a shell must get what the server has now.

The manifest was incomplete and self-contradictory (SVG-only icons, no id/scope, a `theme_color` that fought `app.html`). Now complete, with a maskable icon padded into the safe zone.

## Testing

The site builds clean. `install.sh` was verified to land in the deployable output and match its source of truth byte-for-byte.

## Authorship

AI-generated (Claude Opus 5).


tbraun96:
## Blocked on [atlas-recipes#25](https://github.com/Avarok-Cybersecurity/atlas-recipes/pull/25)

The `Build SvelteKit site` check fails, and correctly so: it copies `install.sh` from `atlas-recipes`, which checks out that repo at `main`. The file is on the launcher PR's branch, not on `main` yet.

The build deliberately refuses rather than continuing without it — building a site whose front-page install command 404s would be worse than a red pipeline, since that command is what visitors are told to pipe into a shell. The step now names the dependency instead of exiting silently.

**Merge order:** `atlas-recipes#25` first, then re-run this build. No change is needed here.

PR #759, opened 2026-08-26
feat(cli): emit the serve flag surface as a machine-readable snapshot

tbraun96:
`atlas-recipes` hand-transcribes this repo's flag surface — and not from here. Its 48-entry table came from the **Python** launcher that predates the Rust one, and the drift is invisible until a launch dies inside a container:

- the website has offered `scheduling_policy: "fcfs"` for four releases; `validate_serve_args` has never accepted it *(fixed downstream in atlas-recipes#60)*
- `kv_cache_dtype` offered 3 of the 16 in `KvCacheDtype::ALL`
- nine keys shipping recipes set — including `lm_head_dtype`, described in-recipe as a correctness pin — are silently dropped

`spark dump-serve-options` prints what clap actually knows: **102 flags**.

## This is a build artifact, not a public interface

`ServeArgs` still has no `Serialize` derive, and this does not give it one. `recipe/schema.rs:5-11` refuses to make the CLI a public serialization format because a cross-repo rename would become a compat break — that reasoning is intact. Nothing here promises a flag keeps its name; a rename shows up as a **diff in whatever consumes the snapshot**, which is exactly the drift detection missing today.

The subcommand is `hide = true` and answers before any subscriber, TUI, or GPU init — it prints a document and exits. The reflection is the same one `tui/lib_fields.rs` already does.

## Two fields a consumer cannot derive

**`presence_only`** — `--video-allow-ffmpeg` takes no value; `--gdn-fused-norm` requires one:

```
video_allow_ffmpeg    presence_only=true    is_bool=true   opts=[]
gdn_fused_norm        presence_only=false   is_bool=true   opts=["true","false"]
```

Both are written `: true` in a recipe and look identical there. Rendering either as the other produces a command clap refuses — on the serving machine, after the operator has reviewed it.

**`recipe_aliases`** — `max_model_len` is vLLM's spelling kept in the recipes; the flag is `--max-seq-len`. Nothing about the flag reveals that, and omitting it makes a generated table silently stop claiming keys recipes set. `RENAMES` is now `pub(crate)` so there is one list rather than a copy:

```
--max-seq-len  ← ["max_model_len"]
--tp-size      ← ["tensor_parallel"]
--bind         ← ["host"]
```

I only found this because a test failed — my first version derived the key from the flag and lost all three.

## Compile-time guard did its job

Adding the subcommand **failed to compile** until eight existing matches on `Command` were extended. A new subcommand cannot exist without every site deciding what it means.

## Verified against the real binary

```
$ spark dump-serve-options | …
flags              : 102
scheduling_policy  : ["fifo", "slai"]
kv_cache_dtype     : 16 variants
9 dropped keys     : all present
```

7 tests, `cargo fmt`/`clippy` clean under `--no-default-features --features cuda`.

## Next

`atlas-recipes` vendors this snapshot and generates its flag + settings tables from it, with CI regenerating and diffing. The deny-list and the numeric bounds stay hand-written — those are judgement, not data, and the engine carries no ranges at all.

### DIFF
--- a/crates/spark-server/src/cli.rs
+++ b/crates/spark-server/src/cli.rs
@@ -11,6 +11,7 @@ mod bench_resolve;
 pub mod bench_run;
 mod bench_selfstart;
 pub(crate) mod flag_values;
+pub(crate) mod manifest;
 mod serve_args;
 mod validate;
 pub use bench_args::BenchmarkArgs;
@@ -42,6 +43,15 @@ pub enum Command {
     Serve(ServeArgs),
     /// Run and inspect the benchmark suite, without the dashboard.
     Benchmark(BenchmarkArgs),
+    /// Print the serve flag surface as JSON.
+    ///
+    /// Hidden because it is a build tool, not part of the supported CLI: it
+    /// exists so downstream tooling can be GENERATED from clap rather than
+    /// transcribed from it. `ServeArgs` still has no `Serialize` derive and
+    /// this does not promise that any flag keeps its name — a rename shows up
+    /// as a diff in whatever consumes the output.
+    #[command(hide = true)]
+    DumpServeOptions,
 }
 
 #[cfg(test)]
--- a/crates/spark-server/src/cli/manifest.rs
+++ b/crates/spark-server/src/cli/manifest.rs
@@ -1,0 +1,135 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! Emitting the serve flag surface as a machine-readable snapshot.
+//!
+//! **This is a build artifact, not a public interface.** `ServeArgs` still has
+//! no `Serialize` derive and this does not give it one: nothing here promises
+//! that a flag will keep its name, and a rename shows up as a diff in whatever
+//! consumes the snapshot rather than as a silent break. That is the point —
+//! `atlas-recipes` currently hand-transcribes this surface from a *Python*
+//! launcher that predates the Rust one, and the drift is invisible until a
+//! launch dies inside a container.
+//!
+//! The reflection is the same one [`crate::tui::lib_fields`] already does, for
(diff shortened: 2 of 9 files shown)


tbraun96:
**Blocked on gate records, not on review.** Parking this rather than running the campaign now.

The `PR benchmark gate` fails because touching `crates/spark-server/src/**` invalidates every committed record — this PR owes **10**:

```
agentic-webserver, vision-fidelity, video-fidelity, ttft-warm-gate, ttft-cold-gate,
bfcl-subset, bfcl-subset-echolp, ssm-state-poisoning-gate, decode-floor, concurrency-sweep
```

That is the policy working as designed, and I am not looking for a way around it. `RequiredSet::by_path` says so explicitly — *"Stands entirely on its own; nothing in the intent half can shrink it"* — and `coverage.rs` records why: an earlier allowlist meant "a path outside the list invalidated nothing at all, however dangerous."

Worth stating plainly anyway, since it is the tension a reviewer will feel: `DumpServeOptions` is `hide = true` and returns before the subscriber, the TUI, and any GPU init. It cannot reach inference. But the gate is path-based *on purpose*, and a per-PR judgement call about which changes are "really" performance-relevant is exactly the hole the current design closed.

So this sits until its records can be produced — ideally **batched with other engine work that genuinely touches inference**, where the 10 runs earn their cost instead of certifying a subcommand that exits early.

Nothing else is outstanding: 7 tests, `fmt`/`clippy` clean under `--no-default-features --features cuda`, and the emitter is verified against the real binary (102 flags, correct `presence_only` per flag, all three recipe aliases, all 9 keys `atlas-recipes` currently drops).

The downstream half does not have to wait — I am building the `atlas-recipes` consumer against the generated snapshot now, so only the regeneration step depends on this landing.

tbraun96:

PR #754, opened 2026-08-26
qwen4_exp: Qwen3.8-Flash-Next config parser + checkpoint audit (#753)

rsafier:
**Draft — work in progress.** Tracks [#753](https://github.com/Avarok-Cybersecurity/atlas/issues/753). Stacked on #746 (`feat/longcat-ngram-readiness`), whose n-gram engine and NVMe row cache this port reuses; review that first.

Nothing serves yet. This is the config layer plus the analysis that de-risks the rest.

---

### 1. `qwen4_exp` config parser (`d575042d`)

Atlas now recognises `Qwen3.8-Flash-Next` — a ~180B hybrid multimodal MoE: 48 layers interleaving GDN and full attention 3:1, 512 experts top-10 plus a shared expert, a **4-stream hyper-connection residual**, a **QSA indexer** on the full-attention layers, **PLE n-gram injection** at one layer, and a ViT tower.

Most of it lands on fields Atlas already has, because the DeepSeek-V4 port built the two hardest pieces (mHC plumbing and the semantic indexer):

```
hc_count               → hc_mult          (Atlas already runs hc_mult=4)
indexer_n_heads/…      → index_n_heads / index_head_dim / index_topk
indexer_compress_ratio → compress_ratios  (per-layer)
ngram_size             → emb_neighbor_num
heads_per_ngram        → emb_split_num
layer_types            → LayerType, verbatim
```

Five new `ModelConfig` fields. `hc_lowrank` is the load-bearing one: Atlas's mHC is DeepSeek-V4's, whose mixer is a Sinkhorn-normalized square matrix, where Qwen mixes the same 4 streams through a **low-rank pair** (rank 320). Stream layout and `hc_mult` carry across unchanged — only the mixing math differs — so `hc_lowrank` is what selects the variant once the kernel lands. Streams-without-rank is **refused** rather than silently taking the Sinkhorn path.

Two assertions worth calling out, both guarding mistakes that would otherwise be silent:

- **n-gram head slices are concatenated, not projected.** 16 heads × 160 dims = 2560 = `hidden_size`. LongCat projects each row up to hidden and *sums*; Qwen concatenates. So the head count must divide `hidden_size`, and the error says why.
- **`layer_types` must be present** and match `num_hidden_layers`. Falling back to `full_attention_interval` would be a guess, and a wrong interleave puts GDN weights on attention layers.

### 2. Both naming revisions (`70e31b14`)

[...]

PR #777, opened 2026-08-28
fix(bfcl): prove the scorer imports before generating 995 responses

tbraun96:
A gate run tonight generated **all 995 responses**, then died in scoring:

```
ModuleNotFoundError: No module named 'soundfile'
```

`soundfile` is a transitive import of `qwen_agent`, which bfcl-eval pulls, and the pinned requirements never carried it. The gate then did the right thing and wrote **no record** — *"a record is evidence that a benchmark RAN"* — so 1.6 h of inference produced nothing.

Provisioning already checks python, the venv, and that `pip install` succeeds. **None of those is the claim that matters.** The AST checker is imported *inside* `_score_ast`, so nothing at module scope touches bfcl-eval and `score.py --help` exits 0 against a venv that cannot score at all. Measured on the broken venv: `--help` → exit 0, `--selftest` → exit 1.

`score.py --selftest` performs exactly the imports scoring performs, and provisioning runs it as a step. The list lives in `score.py` so it cannot drift from what the scorer does — a list of module names in the provisioner would be a second copy nobody updates.

`--selftest` is handled *before* the parser is built, so `--dataset`/`--responses` keep argparse's `required=True` contract: an existing test asserts a scoring invocation omitting them fails exactly as it always has, and it still does. That test now also asserts `--selftest` is **recognised** — argparse answers an unknown flag with 2, so without it the new provisioning step would pass for the wrong reason.

### DIFF
--- a/crates/atlas-plugin/assets/bfcl/score.py
+++ b/crates/atlas-plugin/assets/bfcl/score.py
@@ -111,7 +111,31 @@ def _read_jsonl(path):
                 yield json.loads(line)
 
 
+def _selftest() -> int:
+    """Import what scoring imports, and nothing else.
+
+    The AST checker is imported INSIDE `_score_ast`, so nothing at module
+    scope touches `bfcl_eval` — `score.py --help` succeeds against a venv that
+    cannot score at all. That is how a missing transitive dependency
+    (`qwen_agent` -> `soundfile`) stayed invisible until after 995 responses
+    had been generated: an hour and a half of inference, then an import error,
+    and no gate record because the run ended Failed.
+
+    Kept here rather than as a list of module names in the provisioner, so it
+    cannot drift from what the scorer actually needs.
+    """
+    from bfcl_eval.constants.enums import Language  # noqa: F401
+    from bfcl_eval.eval_checker.ast_eval.ast_checker import ast_checker  # noqa: F401
+
+    return 0
+
+
 def main() -> int:
+    # Handled before the parser is built, so `--dataset`/`--responses` keep
+    # argparse's own `required=True` contract: a scoring invocation that omits
+    # them must still fail exactly as it always has.
+    if "--selftest" in sys.argv[1:]:
+        return _selftest()
     ap = argparse.ArgumentParser()
     ap.add_argument("--dataset", required=True)
     ap.add_argument("--responses", required=True)
--- a/crates/atlas-plugin/src/benchmarks/bfcl/provision.rs
+++ b/crates/atlas-plugin/src/benchmarks/bfcl/provision.rs
@@ -113,6 +113,34 @@ pub async fn ensure(store: &ArtifactStore, handle: &PluginHandle) -> Result<Arti
     handle.status("BFCL: downloading bfcl-eval (needs network)");
     python::pip_install(&interpreter, &dir.join("requirements.txt")).await?;
 
+    // Prove the SCORER can import before spending an hour and a half
+    // generating responses for it. `pip install` succeeding is not the same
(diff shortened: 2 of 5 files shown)


tbraun96:
Status (2026-08-28 07:35 UTC): the one failing check is **PR benchmark gate → "Check committed benchmark records against their baselines"**. Not a defect in this PR — it touches `crates/atlas-plugin/assets/bfcl/score.py` and `crates/atlas-plugin/src/benchmarks/bfcl/provision.rs`, and `crates` is a PERF_PATH, so this tree needs its own ten records.

It cannot borrow the campaign running right now: that one is pinned to #754's tree at `1d30d5d5f`, and records are content-anchored, so a different tree does not inherit them.

**Sequencing** — rather than burning a third ~8h campaign on this alone: #754 lands first with its own records, then this change folds into the tree of the campaign that main will need anyway (main is 22 commits / 304 perf-relevant files ahead of `feat/longcat-ngram-readiness`, so that campaign has to happen regardless). One campaign, both changes.

Holding here deliberately rather than merging around a red gate.

tbraun96:
Split: the three `scripts/` changes on this branch are now #779, which touches no perf path and can merge on its own. This PR keeps only the two `crates/atlas-plugin` files — the scorer selftest — and stays held for the campaign described above.

Once #779 lands, rebasing this branch on main will drop those three files from the diff.

tbraun96:
**Folded into the consolidated certification stack — and it is the one PR in that stack that protects the campaign itself.**

Verified unlanded, not assumed: `main`'s `crates/atlas-plugin/src/benchmarks/bfcl/provision.rs` has no import-proving step. Your own comment records what that costs:

> A run on 2026-08-28 generated all 995 responses, failed here, and wrote no records.

That is roughly an hour and a half of GPU spent to learn that `bfcl-eval` pulls `qwen_agent`, which imports `soundfile`, which the pinned requirements did not carry — and because the AST checker is imported inside a function, nothing at module scope touched it.

I am about to run **one** campaign covering nine stacked PRs. Landing your fix *inside* that stack means the campaign proves the scorer imports before it generates anything, rather than discovering it at the end. That is the difference between a failed campaign costing minutes and costing hours.

**Two of your three commits were already on `main`** (`fix(scripts): read the command's exit status` and `docs(gate): print the three traps`) — they cherry-pick empty. Only the `provision.rs` fix carried over, with authorship preserved.

Verified on the composed stack: `cargo check --workspace --all-targets` → EXIT=0, no new LoC-cap violations.

tbraun96:

|---|---|---|---|---|
| 8 | #777 | bfcl scorer import proof before generating responses | bfcl-harness | **← this PR** |

tbraun96:
Closing: the scorer self-test landed with the 2026-09-04 verify-line stack.

**Carrier:** `dbd79688f` — *"verify-line: the 2026-09-04 stack re-merged onto main, one campaign (#1130)"*.

`main` carries both of this PR's substantive files at their post-rebrand paths — `crates/avarok-plugin/assets/bfcl/score.py` and `crates/avarok-plugin/src/benchmarks/bfcl/provision.rs` — and the content is verbatim, including the passage that makes this PR recognisable:

> *"The AST checker is imported INSIDE `_score_ast`, so nothing at module scope touches `bfcl_eval` … That is how a missing transitive dependency (`qwen_agent` → `soundfile`) stayed invisible until after 995 responses had been generated: an hour and a half of inference, then an import error."*

Six of seven sampled added lines are present; the seventh is a `scripts/` line the rebrand rewrote.

It reads as CONFLICTING only because this branch still edits `crates/atlas-plugin/…`, a path #1101 renamed — the same reason #558, #884 and #504 all looked unmerged when they were not.

PR #781, opened 2026-08-28
cutlass: skip null-weight expert groups — unblocks grouped MoE under EP

rsafier:
## Problem

Under expert parallelism the CUTLASS grouped-MoE path faults with `CUDA_ERROR_ILLEGAL_ADDRESS (700)` in `ffn.forward_prefill`. The hand-written ptrtable kernels are unaffected, so EP has been usable only by **turning the fast MoE path off** — which is where the ~3.6x prefill gap between EP and single-node recorded on Laguna comes from.

## Root cause

The router and `moe_sort_by_expert` stay **global-width** under EP: every rank receives routed rows for experts it does not own, and those experts' weights are `ExpertWeight::null()` placeholders (`weight_map/loaders_moe.rs:62-65` allocates the expert vector at full width and nulls the remote ones).

`prep_grouped_a` filtered groups on rows alone:

```c
if (m_e <= 0) { continue; }
```

so a remote expert with routed rows became a CUTLASS group, and `launch_projection` then bound `hB[g] = packed_ptrs[e]` — a null pointer — with no check.

The ptrtable kernels never hit this because **all ten variants** open with `if (B_expert == 0) return;`. The CUTLASS path simply never learned the same lesson, even though the SFB builder already skips remote experts (`if sptr == 0 { continue; } // remote/placeholder expert`) and therefore deliberately hands nulls forward.

## Fix

Pass the B pointer table into `prep_grouped_a` and apply the same predicate. It is applied in **both passes** — they must agree, or the per-group staging offsets (`a_grp_off`/`sfa_grp_off`, indexed by `gi`) desynchronize from the groups actually built.

The correctness precondition already holds: `forward_prefill_routed` memsets the gate/up/down outputs whenever `ctx.comm.is_some()`, precisely so a remote expert can contribute zeros while the owning rank's result arrives via the all-reduce. That is exactly the ptrtable semantics.

Also converts a null bind in `launch_projection` into `status -140` rather than letting the GEMM fault — an opaque CUDA 700 surfaces far from its cause.

**Strict no-op off EP:** single-GPU tables contain no nulls, so no group can be skipped. This matters because the file is reached by every NVFP4 grouped-MoE model, not just the one tested here.

## Evidence

Holo-3.1-35B-A3B-NVFP4, EP=2 across two GB10 nodes, one variable (`ATLAS_CUTLASS_EP_NULL_GUARD`):

| arm | result |
|---|---|
| guard **off** (pre-fix) | `CUTLASS nvfp4 grouped(fused) gate_up failed: status -140` at layer 0, **both ranks** |
| guard **on** | serves, coherent output, zero errors on either rank |

The `-140` is the new check firing, which proves a null-B group was actually constructed — a stronger signal than the fault it replaces.

Single-node with the full CUTLASS flag set: SFB built on all 40 layers, coherent output, behaviour unchanged.

EP=2 prefill @6K, CUTLASS on vs off: **4019 vs 3804 tok/s (+5.7%)**, 2 reps.

⚠️ That is far short of the 3.6x implied by the older Laguna note, because the off-arm here retains `LOW_MEMORY_MOE` + `FAST_MOE_MODE=full` and so falls back to the transposed grouped path, not to slow w4a16. The headline of this PR is **correctness** — the fast path becomes *available* under EP; how much it is worth depends on what the fallback would have been.

## Scope of validation

- **Confirmed on Holo-35B.** **Inferred, not retested, for Laguna-S-2.1**, where the original crash was recorded. Same defect, same shared code path.
- The flag the old note blamed, `ATLAS_MOE_CUTLASS_DEVICE_OFFSETS`, no longer exists in the tree — that hypothesis (offsets computed over a full table while EP holds half) is wrong: the expert vector is always global-width, so offsets and pointer tables do agree.
- `ATLAS_CUTLASS_EP_NULL_GUARD=0` restores the old behaviour for A/B.

## Gates

`cargo fmt` clean. Release build with `ATLAS_TARGET_MODEL='*'` clean. `spark-model` 639 passed, `spark-runtime` 274 passed, 0 failed.

⚠️ `cargo clippy --all-targets` fails on **`crates/spark-model/examples/dense_gemv_bf16_batchm_microtest.rs:124`** (`print_literal`). That file is untouched by this PR (`git diff main..HEAD` = one `.cu` file) — the failure is **pre-existing on main**, not introduced here.

### DIFF
--- a/crates/spark-runtime/cuda/cutlass_nvfp4_grouped_gemm.cu
+++ b/crates/spark-runtime/cuda/cutlass_nvfp4_grouped_gemm.cu
@@ -13,6 +13,7 @@
 #include <cuda_bf16.h>
 #include <cuda_fp8.h>
 #include <cuda_runtime_api.h>
+#include <cstdlib>
 #include <vector>
 
 #include "cute/tensor.hpp"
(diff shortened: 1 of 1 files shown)


tbraun96:
**AutoMerger sieve verdict: `sieve:cleared` — all three sieves pass.** The null-group skip was checked for silent-swallow semantics and is clean: `forward_prefill_routed.rs:135-143` zero-memsets outputs under EP and `loaders_moe.rs:62-65` populates remote experts as `ExpertWeight::null()`, so skipping a null-B group reproduces the documented zero-contribution + all-reduce path; the new `status -140` guard replaces a silent fault with fail-fast. Recorded gap (not blocking): no automated regression test — evidence is the manual EP=2 A/B, acceptable for a multi-GPU-hardware bug. Grouping into **stack `perf-plumbing`** with #705; this PR stays open until the stack lands, then closes citing the landing commit.

tbraun96:
**Stacked → lands via #891 (`stack/perf-plumbing`), one campaign for the pair.**

| order | PR | attribution signature | |
|---|---|---|---|
| 1 | #705 | boot-phase (module load/unload) |  |
| 2 | #781 | grouped-GEMM phase (EP null-group skip) | **← this PR** |

Composition is patch-id-verified byte-equivalent to this PR's diff, authorship preserved, offline-green. Adversarial review passed with a blocking pre-campaign condition: a single-node MoE smoke runs before any campaign (covers the config this pair's evidence never executed). Please don't push here while the stack is in flight; this PR closes with landing evidence after #891 merges. If any hunk does not make it across, point at it and I will lift it onto a fresh branch.

DrRainbows:
Closing as already on `main`. This landed via #891 (`356701d6`, an ancestor of `origin/main`). Named artifact: `crates/spark-runtime/cuda/cutlass_nvfp4_grouped_gemm.cu:387` (`getenv("ATLAS_CUTLASS_EP_NULL_GUARD")`, A/B arm documented at :382). The file is byte-identical on `main`. Every commit here cherry-picks empty onto current `main` (`c2948e3f`).

PR #782, opened 2026-08-28
qwen4_exp: load + audit the MTP draft module (NO speculation yet)

rsafier:
**Stacked on #754** (`wip/radixark-qwen4-exp`). Review that first; this PR is the single commit on top.

## What this does — and what it does NOT

The Qwen3.8-Flash-Next checkpoint ships a **complete MTP block**: 6173 tensors — its own 512-expert MoE, shared expert, full attention, QSA indexer, mHC sites and mixer, plus `fc_embedding`/`fc_hidden` and two pre-fc norms. `load_mtp_weights` has returned `Ok(None)` since v1 (#753 item I), on the grounds that wiring it before the main forward path worked would be "building on sand". The main forward works now.

**⚠️ `--speculative` on qwen4_exp still produces NO speculation after this PR.** The module is loaded, shape-audited and stored; nothing decodes it. That is the honest headline and it is why the title says "load + audit".

A proposer installed today would route the draft into `refuse_batched_under_hc`, which fires from both `decode_batched` and `decode_verify_multi` — and the scheduler converts that error into `a.finished = true`, i.e. a **silently truncated response**, not a fallback. So the proposer is deliberately not wired.

## Why not `MtpWeights`

Atlas's `MtpWeights` does not fit this block:

- it wants a fused `fc` `[hidden, 2*hidden]`; this checkpoint has **split** `fc_embedding` + `fc_hidden`
- it assumes a pre-norm residual layer (`input_layernorm` / `post_attn_layernorm`); this block carries an **mHC hyper-connection bracket** instead
- it has no indexer field; this attention carries its **own QSA indexer**

So `MtpWeights` is left alone and `load_mtp_weights` still returns `Ok(None)`. The module is loaded by a bespoke `load_qwen4_exp_mtp_module` that replays the **existing prefix-parameterized qwen4_exp helpers** once with `lp = "mtp.layers.0"` — the DeepSeek-V4 "Track B" shape. The MTP layer is architecturally identical to a main decoder layer, so this reuses rather than duplicates.

## Included

- `num_mtp_modules` parsed from `text_config.mtp`
- a shape-checking namespace audit that resolves the expert layout explicitly (**fused vs per-expert**, refusing neither/both) and asserts the asymmetric combiner widths — `pre_fc_norm_embedding [2560]` vs `pre_fc_norm_hidden [10240]`
- an offline safetensors-header inventory test run against **both** real snapshots
- `hc::load_head_at`, so the module's own mixer loads without touching `load_head`'s main-model behaviour

## The one genuine unknown, left unresolved on purpose

`mtp.pre_fc_norm_hidden` is `[10240]` = `hc_mult * hidden`, while `mtp.fc_hidden` is `[2560, 2560]`. A `[2560,2560]` matrix cannot consume a 10240 vector, so exactly one of these is true:

- **(A) per-stream** — grouped-norm the 4-stream highway, apply `fc_hidden` as 4 GEMM rows, broadcast-add `fc_embedding(normed_embed)`; the output *is* the MTP highway
- **(B) collapse-then-expand**

HF's `modeling_qwen4_exp.py` carries `_keys_to_ignore_on_load_unexpected = [r"^mtp.*"]` and ships no MTP class, so **no golden can be generated** and no shape check discriminates — both `fc` matrices are square, so swapping them is silent. Getting it wrong costs 100% of acceptance and 0% of correctness, and looks exactly like "MTP just doesn't help on this model". The audit's asymmetric-width assertion is the tripwire if a future checkpoint falsifies the premise.

## Deferred, with analysis recorded in the module docs

- the proposer + target-side highway handoff (needs `hc_streams`, not the collapsed BF16 row every existing proposer receives)
- the draft body's **own** single-layer KV cache — it must never see the main pool, and it would sit outside the util pledge (same class as the recorded DFlash OOM)
- the SSM verify-pool gap: `has_mtp` is derived from `!mtp_weights.is_empty()`, which a Track-B install leaves empty, so no SSM checkpoint pools exist and `verify_draft_capacity` returns `usize::MAX`. V4 is all-attention; qwen4_exp has 36 GDN layers whose partial-accept rollback reads exactly those pools
- K-row mHC verify (the part that actually buys tok/s)
- QSA/PLE rewind across a rejected draft — QSA's `ingested` is monotone with no rewind API; PLE's host history advances per row with a documented corruption class
- EP>1 (the MTP MoE has no `force_all_experts` path, and adding one touches the shared qwen35 loader — out of scope here)

`qwen4_exp` is deliberately **not** added to `MTP_SUPPORTED_MODEL_TYPES`: that allowlist is a refusal, so adding the model *removes* a guard. It becomes correct to add in the PR that makes speculation actually work.

## Review found four defects in the first cut — all fixed here

1. **A main-model regression.** `parse_mtp` refused with `ensure!` at config-parse time, which runs on *every* launch. A checkpoint whose `mtp` block differed would have stopped booting **entirely** over an optional feature. It now declines to advertise the module (`num_mtp_modules = 0`, the absent-key path) and the refusal-by-name moved to `ensure_loadable`, which only runs when MTP is requested. Tests updated to pin "not advertised, still boots", plus a positive test so they cannot pass vacuously.
2. **Gate mismatch.** The upload gate (`--speculative` or env) and the load gate (`use_speculative` = `--speculative` **or** `--dflash`) were different predicates, so `--dflash` alone logged `MTP module load FAILED` on every boot, blaming the wrong cause. The load now requires the tensors to actually be resident.
3. **Misleading instrumentation.** The per-arm memory deltas used `saturating_sub` on unsigned free-memory readings and printed `0.00 GB` precisely on the fused arm — where `build_moe` frees ~5 GB of BF16 sources while allocating ~1.3 GB, so free memory *rises*. It read as "the MTP block is free". Deltas are signed now and labelled as free-memory differences, not resident cost.
4. Two overclaims ("threaded into BOTH call sites", "cannot leak") corrected in the docs — the `Err` arm still strands buffers, recorded as a known gap.

## Gates

`cargo fmt` clean; `clippy --all-targets` clean across all kernel targets; **atlas-core 132 passed, spark-model 662 passed, 0 failed**; release build verified with the NCCL `RUSTFLAGS` path. No GPU run — load cost and the audit path are unmeasured on hardware, which is why nothing is defaulted on.

### DIFF
--- a/crates/atlas-core/src/config/parsers/qwen4_exp.rs
+++ b/crates/atlas-core/src/config/parsers/qwen4_exp.rs
@@ -67,6 +67,8 @@ pub(crate) fn parse_qwen4_exp(raw: &Value) -> Result<ModelConfig> {
     // shared loader helpers need no qwen4_exp-specific naming.
     config.weight_prefix = "model.language_model".to_string();
 
+    parse_mtp(text, &mut config)?;
+
     ensure!(
         config.hidden_size > 0,
         "qwen4_exp hidden_size must be non-zero"
@@ -132,6 +134,72 @@ pub(crate) fn parse_qwen4_exp(raw: &Value) -> Result<ModelConfig> {
     );
 
     Ok(config)
+}
+
+/// The MTP (multi-token-prediction) draft block, declared as a NESTED
+/// `text_config.mtp` object rather than the flat `mtp_num_hidden_layers` the
+/// qwen3_5/3_6 family uses.
+///
+/// Sets `num_mtp_modules`, which is the field every MTP module loader
+/// short-circuits on (`deepseek_v4/mtp.rs`: `if config.num_mtp_modules == 0 {
+/// return Ok(None) }`). Without it the block is declared in the checkpoint and
+/// invisible to the engine.
+///
+/// The shape is REFUSED rather than adapted when it is not the single
+/// full-attention layer this checkpoint ships: a `layer_types` naming
+/// `linear_attention` would have the module loader build a GDN body off keys
+/// that do not exist, and >1 layer would have it build one layer and silently
+/// drop the rest. Neither failure looks wrong at load.
+///
+/// An ABSENT `mtp` key leaves `num_mtp_modules` at 0 — speculation stays off,
+/// which is the pre-existing behaviour for every qwen4_exp checkpoint.
+fn parse_mtp(text: &Value, config: &mut ModelConfig) -> Result<()> {
+    let Some(mtp) = text.get("mtp") else {
+        return Ok(());
+    };
+
+    // ⚠ THESE CHECKS MUST NOT FAIL THE PARSE. parse_mtp runs on EVERY
+    // qwen4_exp launch, long before anything knows whether MTP was asked for,
+    // so an `ensure!` here refuses to boot a checkpoint that serves perfectly
+    // well today with its MTP block simply ignored — turning an unsupported
+    // OPTIONAL feature into a hard startup failure. The refusal belongs at
+    // MTP-load time (probe_mtp::ensure_loadable), which only runs when the
+    // module is actually requested. Here we decline to ADVERTISE the module:
+    // leaving num_mtp_modules at 0 is exactly the absent-key path, so
+    // speculation stays off and the main model is untouched.
+    let layers = mtp
+        .get("num_hidden_layers")
+        .and_then(Value::as_u64)
+        .unwrap_or(0) as usize;
+    let types: Vec<&str> = mtp
+        .get("layer_types")
+        .and_then(Value::as_array)
+        .map(|a| a.iter().filter_map(Value::as_str).collect())
+        .unwrap_or_default();
+
+    // Declining is SILENT here: atlas-core carries no `tracing` dependency,
+    // and adding one to a core crate to narrate an optional-feature opt-out is
+    // the wrong trade. Nothing is lost — the operator only cares when they
+    // actually ask for MTP, and then `probe_mtp::ensure_loadable` refuses by
+    // name with the full reason.
+    //
+    // `layers != 1`: >1 would have the module loader build one layer and
+    // silently drop the rest; 0/missing means the block describes no body.
+    // `types != ["full_attention"]`: a GDN body would be assembled from
+    // `linear_attn.*` keys the `mtp.*` namespace does not carry.
+    if layers != 1 || types != ["full_attention"] {
+        return Ok(());
+    }
+
+    config.num_mtp_modules = 1;
+    // Mirror the nested count onto the flat field. serde already parsed
+    // `text_config.mtp_num_hidden_layers` (=1 in the shipped config), so this
+    // normally re-states it; the nested value wins when the two disagree
+    // because it is the one that describes the block that actually ships.
+    if layers > 0 {
+        config.mtp_num_hidden_layers = layers;
+    }
+    Ok(())
 }
 
 /// mRoPE + partial rotary live under `rope_parameters`, not at text_config
--- a/crates/atlas-core/src/config/parsers/qwen4_exp_tests.rs
+++ b/crates/atlas-core/src/config/parsers/qwen4_exp_tests.rs
@@ -59,6 +59,15 @@ const RAW: &str = r#"{
     "ple_embed_dim": 2560,
     "ple_conv_kernel_size": 4,
     "output_gate_type": "sigmoid",
+    "mtp_num_hidden_layers": 1,
+    "mtp_use_dedicated_embeddings": false,
+    "mtp": {
+      "hybrid": true,
+      "layer_types": ["full_attention"],
+      "mtp_use_hidden_state_from_layer": null,
+      "num_hidden_layers": 1,
+      "rope_theta": 10000000
+    },
     "partial_rotary_factor": 0.25,
     "rope_parameters": {
       "mrope_interleaved": true,
@@ -200,6 +209,61 @@ fn layer_types_length_must_match_num_hidden_layers() {
     assert!(err.contains("layer_types has 4 entries"), "{err}");
 }
 
+/// `num_mtp_modules` is the field every MTP module loader short-circuits on.
+/// Left at the factory default 0 it does not matter what the checkpoint
+/// ships — the block is never built.
+#[test]
+fn mtp_module_count_is_parsed() {
+    let c = parse_qwen4_exp(&raw_config()).expect("parse");
+    assert_eq!(c.num_mtp_modules, 1, "text_config.mtp declares one module");
+    assert_eq!(c.mtp_num_hidden_layers, 1, "one layer inside that module");
+}
+
+/// A checkpoint with no `mtp` key must leave speculation OFF, exactly as
+/// before this parser learned the field.
+#[test]
+fn mtp_absent_leaves_speculation_off() {
+    let mut raw = raw_config();
+    raw["text_config"]
+        .as_object_mut()
+        .expect("text_config object")
+        .remove("mtp");
+    let c = parse_qwen4_exp(&raw).expect("parse");
+    assert_eq!(c.num_mtp_modules, 0);
+}
+
+/// The loader builds exactly one MTP body; a two-layer module would have it
+/// build the first and silently drop the second — so it must not be
+/// ADVERTISED. It must also not fail the parse: this runs on every launch,
+/// and refusing here would stop a checkpoint booting AT ALL over an optional
+/// feature nobody asked for. num_mtp_modules == 0 is the absent-key path.
+#[test]
+fn mtp_multi_layer_is_not_advertised_but_still_boots() {
+    let mut raw = raw_config();
+    raw["text_config"]["mtp"]["num_hidden_layers"] = Value::from(2);
+    let c = parse_qwen4_exp(&raw).expect("an unsupported MTP block must not fail the parse");
+    assert_eq!(c.num_mtp_modules, 0);
+}
+
+/// A `linear_attention` MTP body would be assembled from `linear_attn.*` keys
+/// the `mtp.*` namespace does not carry. Same contract as above: decline the
+/// module, boot the model.
+#[test]
+fn mtp_non_full_attention_is_not_advertised_but_still_boots() {
+    let mut raw = raw_config();
+    raw["text_config"]["mtp"]["layer_types"] = Value::from(vec!["linear_attention"]);
+    let c = parse_qwen4_exp(&raw).expect("an unsupported MTP block must not fail the parse");
+    assert_eq!(c.num_mtp_modules, 0);
+}
+
+/// The shipped checkpoint's block IS supported, so it must be advertised —
+/// otherwise the two tests above would pass vacuously.
+#[test]
+fn mtp_supported_block_is_advertised() {
+    let c = parse_qwen4_exp(&raw_config()).expect("parse");
+    assert_eq!(c.num_mtp_modules, 1);
+}
+
 #[test]
 fn missing_text_config_is_refused() {
     let err = parse_qwen4_exp(&serde_json::json!({"model_type": "qwen4_exp"}))
@@ -253,6 +317,10 @@ fn parses_the_real_config_json_when_present() {
     );
     assert_eq!(c.ple_layer_ids, vec![2]);
     assert!(c.vision.is_some());
+    assert_eq!(
+        c.num_mtp_modules, 1,
+        "the shipped config declares one single-layer full-attention MTP block"
+    );
 }
 
 /// Qwen3.8-Flash-Next shipped under `qwen3_8_flash_next` and was later
--- a/crates/spark-model/src/factory/build.rs
+++ b/crates/spark-model/src/factory/build.rs
@@ -668,6 +668,10 @@ pub fn build_model(
         vision_encoder,
         ssm_cache_slots,
         ssm_checkpoint_interval,
+        // A bespoke MTP module is present, so the SSM verify pools must exist
+        // even though `mtp_weights` is empty. Known here: the module is loaded
+        // in Step 5 above, before construction.
+        qwen4_exp_mtp_module.is_some(),
     )?;
 
     // ── Step 6b: DeepSeek-V4 MTP proposer (optional, post-construction) ──
--- a/crates/spark-model/src/factory/build.rs
+++ b/crates/spark-model/src/factory/build.rs
@@ -289,14 +289,17 @@ pub fn build_model(
     if use_speculative && mtp_weights.is_empty() {
         if qwen4_exp_mtp_module.is_some() {
             // Saying "no MTP weights were loaded" here would be a lie: the
-            // block IS loaded and audited. What is missing is the proposer.
+            // block IS loaded and audited. Whether a proposer gets wired
+            // depends on ATLAS_QWEN4EXP_MTP_VERIFY — this arm only fires when
+            // it is OFF, since the proposer install path logs its own line.
             tracing::warn!(
-                "qwen4_exp: the MTP module is loaded and audited, but no \
-                 proposer is wired yet — speculative decoding stays OFF. This \
-                 is deliberate: a proposer installed today would route the \
-                 draft into `refuse_batched_under_hc` mid-verify, which the \
-                 scheduler turns into a truncated response rather than a \
-                 fallback."
+                "qwen4_exp: the MTP module is loaded and audited, but the \
+                 proposer is NOT armed — speculative decoding stays OFF. Set \
+                 ATLAS_QWEN4EXP_MTP_VERIFY=1 to arm the draft head together \
+                 with the mHC K-row verify path it needs; the two arm together \
+                 because a proposer without that verify path routes the draft \
+                 into `refuse_batched_under_hc` mid-step, which the scheduler \
+                 turns into a truncated response rather than a fallback."
             );
         } else {
             tracing::warn!(
@@ -713,11 +716,19 @@ pub fn build_model(
         // ATLAS_QWEN4EXP_MTP_SHADOW the module is just held (unchanged
         // behaviour), because the head's pool is real memory outside the util
         // pledge and nothing should pay for it unless it is being measured.
-        if crate::layers::qwen4_exp_mtp::shadow_enabled() {
-            if let Err(e) = model.set_qwen4_exp_mtp_head(*module, target_embed_for_mtp, max_seq_len)
+        // Speculation needs the head installed as the ACTIVE proposer; shadow
+        // needs it held but inert. `ATLAS_QWEN4EXP_MTP_VERIFY=1` is the same
+        // flag the mHC K-row verify path is gated on, so the two arm together
+        // or not at all — a proposer without that verify path would produce
+        // drafts the verify step then refuses on, mid-request.
+        let arm_spec =
+            use_speculative && std::env::var("ATLAS_QWEN4EXP_MTP_VERIFY").as_deref() == Ok("1");
+        if crate::layers::qwen4_exp_mtp::shadow_enabled() || arm_spec {
+            if let Err(e) =
+                model.set_qwen4_exp_mtp_head(*module, target_embed_for_mtp, max_seq_len, arm_spec)
             {
                 tracing::error!(
-                    "qwen4_exp MTP shadow head FAILED to build: {e:#}. \
+                    "qwen4_exp MTP head FAILED to build: {e:#}. \
                      Serving continues with no MTP."
                 );
             }
--- a/crates/spark-model/src/factory/build.rs
+++ b/crates/spark-model/src/factory/build.rs
@@ -223,15 +223,88 @@ pub fn build_model(
             None
         };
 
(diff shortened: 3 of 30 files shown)


rsafier:
## Update: draft head + shadow harness pushed — **MTP still does not work**

Two commits added (`746ac904`, `696ce4bd`). Headline first: **`--speculative` still produces no speculation, and the new shadow mode's default stage CORRUPTS the server's output.** It is default-off, the install logs a warning saying exactly that, and `has_proposer()` is untouched.

### What is built

`Qwen4ExpMtpHead` — the combiner, the body forward against a private single-layer KV cache, and the low-rank mHC head collapse. Plus the wiring (factory construction, model install, per-seq state, the decode hook) behind `ATLAS_QWEN4EXP_MTP_SHADOW`, and `ATLAS_QWEN4EXP_MTP_SHADOW_STAGE` as a bisect handle.

The head produces the draft's hidden state and lets the caller finish with its own `final_norm` + `lm_head`, so it never duplicates the target's quantization ladder.

### What running it found

Four defects, none of which review would have caught:

1. **Wrong kernel module** — `norm::rms_norm`, not `rms_norm::rms_norm`. The engine refused to start rather than serve on a null handle ("its dispatch site is on a silent fallback path"). The guard did its job.
2. **The hook was in a path this model never takes.** `decode_batch_dispatch` returns early for `n == 1` straight to `decode()`, so a hook on the batched path can never fire on the single-sequence decode it was meant to measure.
3. **The body's highway is `ctx.buffers.hc_streams()`**, not a buffer you can hand it. The combiner must write *there*, and that buffer is persistent per-sequence state, so it must be saved and restored (error paths included).
4. ⭐ **`hc_streams` is FP32, not BF16** — `m*hc_mult*h*4`, because "BF16 storage swamps the small per-layer signal at scale and collapses generation". Sizing the save/restore as BF16 restored only **half** the highway, so the body's full-width writes survived and poisoned every later target step.

### The bisect

250-token greedy completion against a shadow-off control, one stage at a time:

| stage | result |
|---|---|
| `observe` | matches control exactly — **inert** |
| `combine` | matches control exactly — **inert** |
| `body` | degenerate / empty output, watchdog fires — **corrupts** |

`combine` being clean is not luck: at the time it was measured, the save/restore covered exactly the half of the highway the combiner touched — which is itself the evidence for defect 4.

### What is still broken, in fix order

1. **Fixing defect 4 changed the symptom (degenerate text → empty text) without curing the corruption.** At least one more shared-state dependency in the body forward is unidentified. **This is the blocker.**
2. **The combiner is dtype-mismatched** for the same reason — BF16 ops (`rms_norm_strided`, `dense_gemv`, `residual_add`) over an FP32 highway. The draft is therefore meaningless, and **the 0% accept rate measured so far says nothing about the combiner reading.** No BF16→FP32 primitive exists yet.
3. ⭐ **The combiner reading is genuinely open again.** I previously reported it RESOLVED from the `HcLowRank::norm_w` grouped-norm precedent — that was overconfident. The `[10240]` width does argue for the per-stream form implemented here, but the loader's own note says the body runs "MIDDLE mHC mixing only — **the proposer owns BOTH ENDS**", i.e. an `hc_expand` belongs at the front, which the per-stream form has no place for. That reading would equally explain the square `fc_hidden`. Decide it behind `ATLAS_QWEN4EXP_MTP_COMBINER` by comparing shadow accept rates — but only after (1) and (2), or the comparison is noise.

### Unchanged

The verify wall from the original PR description still stands and is untouched: `decode_batched` and `decode_verify_multi` both refuse under the highway for a load-bearing reason (those paths keep their own residual, which the highway replaces), so real verification means routing through `prefill_inner_hc` plus QSA/PLE rewind that has no API today.

Gates: fmt clean; `clippy --all-targets` clean; atlas-core 132, spark-model 662, 0 failed.

rsafier:
## Update: 3 more real bugs fixed — but buffer-by-buffer isolation is the wrong strategy

`99b2452c`. Still **not working**; `--speculative` still produces no speculation, shadow `body`/`full` still corrupt output and stay default-off.

### Fixed (all genuine)
- **`head_scratch` under-allocated by 1280 B — a real heap overflow.** `hc_head_lowrank`'s decode-split layout is `t*(hc*h + hc_lowrank)*4` (normed FP32 `[t,hc*H]` **then** low FP32 `[t,rank]`); sizing it `hc*h*4` covers only the first part.
- **Local `MTP_META_OFFSET = 40960` → the shared `49152`.** 40960 sits inside the region the target's metadata slab reserves for its block table (49152 covers 4096 blocks = 65536 tokens). Doesn't collide at short context — the shape of bug that first appears at long context.
- **`hc_post` / `hc_lowrank_scratch` / `norm_output` now saved+restored** around the body, error path included. The new `ATLAS_QWEN4EXP_MTP_DIFF=1` fingerprint diff named these; I had reasoned they were "transient scratch" and skipped them, and that reasoning is what kept the bug alive for several rounds.
- **The diff probe itself** — the baseline was sampled *after* the combiner wrote `hc_streams`, making it a false positive. Sampled before it now, and `hc_streams` no longer appears, which **proves** the highway save/restore is correct instead of assuming it.

### ★ The actual finding

None of it cured the corruption. `body` still degenerates the target (empty completion, `finish: length`) while `observe` and `combine` remain byte-identical to the shadow-off control.

Every shared buffer this file can name is now saved and restored — so **the damage is not in `ctx.buffers`**, and enumerating buffers to save is the wrong strategy.

The draft body inherits the target's entire `ForwardContext` via `..*ctx` and can reach state that is not a buffer this file can enumerate. **The sound fix is isolation by construction**: give the draft its own `BufferArena` (sized for T=1 — a few MB) and build its `ForwardContext` from that, so it cannot touch the target at all. That should be done before any further patching here.

Gates: fmt clean; `clippy --all-targets` clean; atlas-core 132, spark-model 662, 0 failed.

rsafier:
## ✅ The draft head WORKS — 86.5% shadow accept, harness inert

`3b8b2984`.

| | shadow_off | shadow_on |
|---|---|---|
| text | full answer | **byte-identical** |
| finish | `stop` | `stop` |
| failures | 0 | 0 |
| **accept** | — | **83/96 = 86.5%** |

### The combiner reading is settled by measurement

86.5% validates the **per-stream** form: grouped-norm the four-stream FP32 highway with `pre_fc_norm_hidden [10240]`, apply the square `fc_hidden` per stream, broadcast-add `fc_embedding(rms_norm(embed))`; the result **is** the body's highway, no `hc_expand`. The collapse-then-expand alternative was built and then removed.

### Two fixes flipped it from meaningless to working

**1. Isolation by construction.** The draft runs in its OWN `BufferArena` (T=1, <1 MB); its ForwardContext carries `buffers: &self.arena`. Four earlier rounds of "find the shared buffer, save and restore it" each found a *real* bug — FP32-vs-BF16 highway sizing, a 1280 B `head_scratch` heap overflow, a local `MTP_META_OFFSET` inside the target's block-table region, three unsaved scratch buffers — and **none converged**, because a full decoder layer touches state the caller cannot enumerate. The arena fixed it at once and let every save/restore be deleted (the compiler flagged the save slots as never read).

**2. Dtype correctness across the FP32/BF16 seam.** The highway is FP32; the projections are BF16 GEMVs. Now `hc_pre_stage_bf16` (FP32 → BF16 grouped norm, per-stream RMS, offset-from-1, `[hc*H]` weight — the model's own kernel for exactly this shape, itself independent evidence for the per-stream reading) → per-stream BF16 GEMVs → new `qhc_mtp_combine_streams` (BF16 → FP32 highway). Running BF16 ops over the FP32 highway is **silent garbage**, and is why every earlier accept measurement read 0% and said nothing about the architecture.

New kernel is in qwen3.8-flash-next's own `hyper_connection.cu` (verified not a symlink), compiles into target 24 only.

### Still not speculation

Nothing is fed back, so **no speedup yet**. Remaining: K-row mHC verify (both `decode_batched` and `decode_verify_multi` refuse under the highway — they keep their own residual, which the highway replaces — so verify must route through `prefill_inner_hc`), the SSM verify-pool gap, QSA/PLE rewind on reject, and the proposer itself.

86.5% is what makes that work worth doing.

rsafier:
## Status: the chain RUNS end to end (0 errors) — but wrong output and 8× slower

`3eed6930`. First revision where draft → verify → rollback → both carries complete a generation without a single error. **Speculation is still not usable.**

Measured, 4K ctx, greedy, `--speculative --num-drafts 1` + `ATLAS_QWEN4EXP_MTP_VERIFY=1`, against a same-config baseline:

| | decode | output |
|---|---|---|
| baseline | **19.8 tok/s** | correct |
| speculative | **2.5 tok/s** | degenerates after ~12 tokens |
| errors | — | **0** |

### How it got here (each fix came from a measured failure, not inspection)

- **Routed the GRAPHED verify entries.** I had gated `decode_verify_dispatch`, but the scheduler's γ=1 path is `verify_k2_step → decode_verify_graphed`. The old route hit `decode_batched`, which refused **inside the capture region** — leaving the stream recording so `alloc_sequence`, `free_sequence`, the SSM snapshot and the next prefill all died with status 901. One refusal bricked the server.
- **`rewind_aux`** (QSA mark rewind). `snapshot_aux` returns `None` until a sequence reaches a layer's ingest, so the snapshot/restore pair *cannot* undo the first draft — carry left ahead, next draft died on `QSA: decode at pos 0 but 1 tokens ingested`.
- **`align_aux` / `align_seq_state`** — absolute alignment replacing a fixed rewind-by-1. The overlap between what verify replays and what the indexer already ingested is **not constant**: rewinding by 1 produced the mirror-image failure (`starts at 366 but 365 ingested`). The caller knows where the carry *belongs*, not how far it drifted.
- **Paged-metadata growth.** `ensure_chunked_prefill_meta` *bails* on short capacity rather than growing, so verify extending past the original prefill's allocation failed.

### What's left — both predicted, neither surprising

**1. Correctness.** The first ~12 tokens match the baseline *exactly*, then collapse. Right-then-drifting is the signature of a per-step state leak, not a bad draft — the head measures **86.5–95.5% accept** in shadow mode, where nothing is fed back. Prime suspect: **`rollback_verify_hc` is written but nothing calls it.** The scheduler's K=2 path does its own rewind arithmetic, so a rejected draft never restores the aux carries.

**2. Cost.** 398 ms/token against a 50 ms decode — verify-by-mini-prefill is ~8× a decode step. Expected *in kind* (prefill kernels are small-M inefficient here; the recorded "~1.1–1.3 s floor regardless of slice"), but 8× means **even a 100% accept rate loses**. A K=2 verify must cost under 2 decodes to break even, so this needs a *decode-shaped* verify rather than a prefill-shaped one.

That second point is the more important finding: it says the accept rate was never the risk, and the remaining work is on verify cost, not on the draft head.

### Unchanged

Speculation stays behind **both** `--speculative` and `ATLAS_QWEN4EXP_MTP_VERIFY=1`; neither is a default. The shadow path — which is correct, inert (byte-identical to control) and measures the 91% accept — is unaffected.

Gates: fmt clean; `clippy --all-targets` clean; atlas-core 132, spark-model 662, 0 failed.

rsafier:
## The economics, measured: the DRAFT is free, the VERIFY is structurally blocked

`d129e3ec`. Isolated the two costs that decide whether MTP speculation can pay on this model.

| | cost |
|---|---|
| decode step | **50.5 ms** |
| **draft forward** | **2.6 ms** (shadow-on 53.1 vs baseline 50.5) |
| verify, 2-row mini-prefill | **~395 ms** |

### ⭐ The draft is essentially free — 5% of a decode step

Every earlier concern about draft cost was misplaced. The draft head is done, cheap, and accurate (**86.5–95.5% accept** across factual / code / reasoning / structured / long-context-with-QSA-active). The economics are **entirely** about verify.

### Break-even says a cheaper *serial* verify does not rescue this

Each step costs `draft + verify(2 rows)` and yields `1 + accept` tokens, so beating a 50 ms decode needs `draft + verify2 < 95 ms` — verify under ~92 ms.

| | |
|---|---|
| mini-prefill verify | ~395 ms → **4× over budget** |
| 2 sequential decodes | ~101 ms → **also over**, before counting the draft |

So only a genuinely **batched** K-row step can pay. That is blocked two ways, both already documented in this repo:

- the GDN prefill path has a **floor** — *"the fused chunk has a ~1.1–1.3 s floor regardless of slice (small-M prefill inefficiency)"* — so a 2-row prefill pays close to what a large chunk pays. That **is** the 395 ms.
- the decode-shaped alternative is exactly what `refuse_batched_under_hc` refuses: *"the mHC highway has no batched GDN path yet"* (**#753 item B**).

### Conclusion

**MTP speculation on qwen4_exp cannot pay until a cheap multi-row GDN step under the mHC highway exists.** That is an engine feature, not a fix to the verify file — and it also unblocks **DFlash and DSpark** here, since `verify_e.rs` routes their GDN body through the same refusing `decode_verify_multi`.

### What this branch leaves working and measured

- draft head: **86.5–95.5% accept at 2.6 ms**
- shadow harness: verified **inert** (byte-identical to a shadow-off control)
- K-row mHC verify: **runs end to end, 0 errors** (wrong output + slow, both explained above)
- SSM verify pools for bespoke proposers; QSA mark-rewind + absolute-alignment primitives

Speculation stays behind both `--speculative` and `ATLAS_QWEN4EXP_MTP_VERIFY=1`; neither is a default.

rsafier:
## Correction + handoff (`2a9b4e5d`)

**My previous comment on this PR was wrong and I'm retracting it.** It said MTP speculation here *"cannot pay until a cheap multi-row GDN step under the mHC highway exists"* — an engine feature (#753 item B). Profiling the verify disproves that. The blocker was MoE **routing**, not the GDN, and it was fixable inside this PR.

### What the profile actually showed

Per layer, per verify row:

| stage | before | after | |
|---|---|---|---|
| `moe` | 2700 µs | **191 µs** | **14×** |
| `gdn_block` | 860 µs | 862 µs | unchanged |

`forward_prefill` routes the MoE through the grouped GEMM, which streams all 512 experts' weights **regardless of row count** — so one row paid nearly what a 28-row chunk paid (T=16 6.7–9.6 ms, T=28 8.5–12.3 ms: 1.75× the rows for 1.2× the time). The GDN floor I blamed was never the dominant term.

The fused single-token/K=2/K=3 MoE kernels **already existed** for exactly this shape — they were built for the DFlash/DSpark verify. Both hc bodies now dispatch to them at small row counts. `ATLAS_QWEN4EXP_HC_SMALL_M_FFN=0` restores the grouped path for A/B.

⚠️ **The K=1 arm is the one that fires.** `decode_verify_hc` splits a verify into row-0-then-drafts, so at the default γ=1 *both* calls arrive as **one** row and the k2/k3 arms never trigger. A K=2-only fix would have measured as a no-op — it did, on the first attempt.

### Where the cost stands

| | |
|---|---|
| baseline decode | 50.5 ms/token |
| speculative, before | 205 ms/token |
| speculative, after | **120 ms/token** |

Still doesn't pay (~2.4× baseline, was ~4×). The remaining cost has **moved to the GDN**: 36 layers × 862 µs × 2 rows ≈ 62 ms.

**Next lever, same shape as this fix:** at T=1 a "prefill" row under the highway *is* a decode step, so the hc decode body (`qwen3_ssm/trait_decode_hc.rs`) should serve it instead of the chunk scan. That is a 1-row substitution — it does **not** need the batched multi-row GDN feature my retracted comment pinned this on. A batched K-row step is still the better endpoint, since two serial decodes (~101 ms) exceed the ~92 ms break-even budget on their own.

### Correctness: still open, and the leading hypothesis is now disproved

Four arms, same prompts, greedy, 4K ctx:

| arm | first output | |
|---|---|---|
| spec off (baseline) | `"Red, blue, and green."` | coherent |
| spec on, rollback off | `"Red light")...` | diverges ~tok 2–3 |
| spec on, rollback on | `"Redacción, 1."` | diverges ~tok 2 |
| spec on, rollback on, old MoE | `"Redacción, ..."` | diverges ~tok 2 |

0 rollback errors, 0 panics.

`rollback_verify_hc` was written but **nothing called it** — the top suspect for the degeneration. It's now wired (`Model::rollback_verify_rows`, from the scheduler's K=2 reject branch) and it **changes nothing**: armed and unarmed diverge at the same point. So it ships **off** behind `ATLAS_QWEN4EXP_MTP_ROLLBACK=1` — unproven, not harmful. Forcing the old grouped-MoE verify reproduces identical corruption, which **exonerates the small-M substitution** above.

Also retracting a smaller claim: the *"first ~12 tokens match baseline"* figure **does not reproduce**. Every speculative arm diverges within 2–3 tokens. Don't reason from the 12-token number.

**Where to look next.** Divergence that early, with leaked raw special-token ids in the output (`| 100257`, `<|fim_prefix|>`), is a **wrong-logits** signature, not slow state drift — the verify returns bad rows almost immediately, which no rewind can repair. In order: (a) the K-row logits the mini-prefill hands back — row indexing/aliasing into the logits buffer, a defect class this repo has hit repeatedly; (b) what `apply_aux_states` restores, PLE's rolling conv/history window especially, since unlike QSA's contiguous marks it can't be rebuilt by truncation; (c) scheduler accepted-row bookkeeping vs what this verify advances. **A row-by-row A/B of verify logits against a serial decode of the same tokens settles (a) immediately and is the cheapest next experiment.**

### Safety

Speculation stays behind **both** `--speculative` and `ATLAS_QWEN4EXP_MTP_VERIFY=1`; neither is a default. Baseline serving with spec off was verified coherent **with the new MoE dispatch on**. `cargo clippy --release --all-targets` clean on `spark-model` + `spark-server`.

### Reusable beyond this PR

The small-M FFN dispatch sits in the shared hc prefill bodies, so any mHC model verifying by mini-prefill gets it — including **DFlash/DSpark on this model class**, whose GDN body routes through the same `decode_verify_multi`. `Model::rollback_verify_rows` is a trait hook defaulting to `Ok(false)`, so non-hc models are untouched.

tbraun96:
Picked up the handoff. One finding that is upstream of the MTP correctness work and may matter to it, plus one correction.

## The released RadixArk conversion does not load at all

There are two public conversions of this model and they are not interchangeable:

| | `RadixArk/Qwen3.8-Flash-Next-NVFP4` | `Inferact/Qwen3.8-Flash-Next-NVFP4` |
|---|---|---|
| size | 126 GB, 206 shards | 182.8 GB, 34 files |
| PLE table | **128 shards across 10 `model-plefp8-*.safetensors`** | one contiguous tensor — a single **102.4 GB** shard |
| PLE dtype | **F8_E4M3**, plus a single global `ngram_embedding.weight_scale` = **1.99e-4** | BF16 |

`ple.rs` targets the second one, and says so: its own comment describes the table as *"102 GB of BF16"*, which is Inferact's 102.4 GB shard to the tenth of a gigabyte. `bench/qwen4_exp/ple_golden.py`'s `DEFAULT_SNAP` is an `Inferact/…` path under `/tank/hf`.

Against the RadixArk conversion, this branch (`2a9b4e5d`) fails two ways — I checked, `ple.rs` here is byte-identical to #746's:

1. `PLE: shard 2 lives in a different file from shard 0` — refuses to build.
2. `head_dim * 2` as the row stride against a **1-byte** table. That returns the wrong bytes **silently**; it only becomes an error when the doubled stride finally walks past the end of `model-plefp8-00009.safetensors`, at row 249579227.

I have fixes on `integration/qwen38-flash-next` making the loader layout- and dtype-driven — per-shard paths, stride from the shard's dtype, and a constant-scale mode for a table quantized with one factor rather than per row. With them it loads all 206 shards, serves, prefills and generates (18.3 tok/s, 659 ms TTFT).

**Output is still not coherent** — a greedy two-token reply that stops — so I am not claiming the numerics. I could not settle them here: `ple_golden.py` reads rows with `.float()` and does **not** apply `weight_scale`, which is right for a BF16 table and wrong for this one, and torch is not installed on this box to regenerate fixtures. The scale is 1.99e-4, so ignoring it leaves embeddings roughly four thousand times too large — not a rounding question.

**Why this may bear on the MTP correctness hunt.** Your (b) — *"what `apply_aux_states` restores, PLE's rolling conv/history window especially"* — is in the same subsystem. If your baseline is the Inferact checkpoint, that is a different PLE path from the one above and the two should not be reasoned about together.

## Correction

`2a9b4e5d` does **not** contain #648. Checked by file, not by ancestry: `dflash_head/dflash2.rs`, `markov.rs`, `kernels/gb10/common/dflash2.cu`, `fp8_gemv_rt.cu` and `weight_loader/dflash_loader/config.rs` are all absent, and there is no `dflash2`/`markov` content anywhere in the tree. The `dflash_head/` module that is here is DFlash **v1**, which is already on main.

That matters for the single certification sweep: a stack built on this branch alone would miss DFlash2. `integration/qwen38-flash-next` currently carries main + #746 (with #754) + #648 and is green — `cargo check --workspace --all-targets` clean, `cargo test --workspace` 74 suites, 0 failures — but lacks this branch's MTP module. Merging this into it produced 13 conflicts (`qsa.rs`, `qwen4_exp.rs`, `hc.rs`, `transformer_layer.rs`, `hyper_connection.cu` among them), which I did not resolve — they are yours.

Also, unrelated but undocumented: serving this model needs `--kv-cache-dtype bf16`, or QSA selection refuses because the gather copies raw NHD rows.

Issue #737, opened 2026-08-23
agentic-webserver: task prompt underspecifies the scorer's bare-GET /ping contract — valid implementations fail the gate


rsafier:
## What

Two runs in one N=10 tier (qwen3.8-27b dense variant, sampled tools-preset 0.6) FAILED `webserver_ok` with a healthy, listening server:

```
run 3: FAILED · 6/6 steps · /ping did not answer 'pong' within 30s
       stderr: Running .../release/ping-pong-server   ← server up
run 4: FAILED · 6/6 steps · (same)
```

Both servers read `ATLAS_HARNESS_PORT` exactly as instructed, bound `0.0.0.0:port`, passed their own tests, and answered curl. They failed because both chose a **mandatory-`Json` extractor** for `/ping`:

```rust
pub async fn ping_handler(Json(req): Json<PingRequest>) -> Json<PongResponse>
```

axum rejects a body-less request to that handler (415), and the scorer's contract is a **bare `GET /ping` with no body** answering `pong` (score.rs polls exactly that for 30s). The task prompt, however, only says *"Just have a ping/pong endpoint"* — it never states the bare-GET contract. An agent can satisfy the prompt as written, prove its server works with valid JSON requests, and fail the gate on an unstated requirement. Runs that pass simply happen to draw a tolerant handler (`Option<Json>` / plain-GET) — a die-roll on an ambiguous spec, which under SAMPLED decoding turns into a nondeterministic gate failure mode that has nothing to do with the serving stack under test.

(Related second-order cost: the mandatory extractor also interacts badly with the agent's own `curl -d` spot-checks — `-d` implies POST, the route is GET-only, so the agent sees 405s that 'look off' and burns turns re-verifying.)

## Fix options

1. **State the contract in the prompt** (preferred — the gate measures trajectory quality, not spec mind-reading): add e.g. *"a plain `GET /ping` with no body must respond with pong"* to the task prompt.
2. Or make the scorer tolerant: on rejection, retry with `-H 'content-type: application/json' -d '{}'`.

Either ends the failure class. Option 1 changes the committed task definition, so existing baseline records' comparability should be considered (the correctness bars are absolute counts, and this change strictly removes an ambiguity rather than easing the task).

PR #797, opened 2026-08-28
integration: Qwen3.8-Flash-Next stack (#746 + #754 + #648) — one certification scope

tbraun96:
Integration branch for Qwen3.8-Flash-Next: **#746 + #754 + #648**, carrying the config parser, the checkpoint audit, and the loader/preflight fixes needed to serve the checkpoint end to end.

## Certification

A ten-gate campaign ran against `d8bc9c9d86` on 2026-08-29, 01:26 → 06:36 UTC, exclusive GPU. All ten records are committed in `63ed7b2310`.

```
PASS  agentic-webserver     PASS  ttft-cold-gate      PASS  decode-floor
PASS  vision-fidelity       PASS  bfcl-subset         PASS  concurrency-sweep
PASS  video-fidelity        PASS  bfcl-subset-echolp
PASS  ttft-warm-gate        PASS  ssm-state-poisoning-gate
```

`bfcl-subset-echolp` (2h30m, the long pole): **overall 86.35** against a floor of 86.10, **normalized 86.77** against 86.50, n=1004 — the echolp draw this bench's recorded history is on, not the golden n=995 draw the dense 27B uses.

**The gates never load Flash-Next**, so they are not evidence the branch does what it exists to do. A separate coherence floor served the real checkpoint (Inferact NVFP4, 48 layers / 12 attention / 36 SSM / 512 experts, `sm_121` kernels, `--kv-cache-dtype bf16` as QSA requires) and put ten factual questions to it: **10/10 coherent**.

## Disclosure: agentic-webserver failed once tonight, and the record cannot say so

The committed record reads a clean `10/10`. That is the truth about the *last* measurement and not the whole story, so it is stated here instead.

Tonight's three samples were **9/10, 10/10, 10/10**. The first failed on a single sub-step — one run of ten did not `tore_down`. Because a record is `<date>-<sha>.json`, all three runs wrote the same file and the failure was **overwritten, not superseded**; it no longer exists in `.benchmarks` at all.

Rather than land on the lucky re-run, the same gate was run **four times on `origin/main`** — same box, same night, freshly built binary:

| | samples | failures |
|---|---|---|
| this stack | 3 | 1 |
| `origin/main` | 4 | 1 |

**Main dropped a sample too.** Turn distributions are near-identical (stack 10.0 / 10.7 mean, main 10.8; max 16 / 17 / 20). So the gate is marginal on this box tonight and the stack is not implicated.

Supporting, and stated with its limits:

- **No mechanism found.** The shared surface was swept by reading: the `layers/moe/` changes are inert (guards that pass, identity substitutions, pure additions), `collect_aux_states` bails on `self.ple`/`self.qsa`, the `decode_b` hunk sits behind `hc_mult > 0`. The **verify paths** — the ones MTP actually exercises during this benchmark — were checked last: `hc_row_offset: 0` and `host_token_ids: None` are neutral, and `final_norm_apply` reduces to the byte-identical `ops::rms_norm` because `final_norm_identity` is `false`. Absence of a *found* mechanism is not absence of one.
- The gate's floor carries **`noise 0.00`** — one missed sub-step fails a whole sample outright.
- #737 documents a *separate* nondeterministic failure mode in this same gate.
- The documented cause of a lost `tore_down` (thinking-loop watchdog → unparsed tool call) **did not fire** in the failing run's window.

Against the committed history (52 of 53 records at 10/10 ⇒ ~1.9% per-sample failure rate), one failure in three is a **~5.6%** event. That is borderline, not comfortable — and it is why the control was run rather than assumed.


tbraun96:
**Heads-up for in-flight PRs before this merges.**

This changes **263 files under the perf paths** (`crates`, `kernels`, `Cargo.*`, `vendor`, …). Gate records are content-anchored over exactly those paths, so landing it **retires the benchmark records of every open PR that has them** — they will need re-cutting against the new main.

Affected by author, so nobody finds out from a red check: @rsafier (#782, #781, #745, #744, #743, #742, #719, #718) and my own #777.

This is inherent to the gate rather than anything unusual about this branch — any `crates/` merge does it. The only mitigation is the usual one: certify, then merge promptly, rather than holding a certified perf branch open while others cut records against a main that is about to move.

For what it's worth, the reverse also held tonight: this branch's own records survived merging main in, because everything that landed on main during the campaign was site/CI only — zero perf-path files. I re-checked that against current main immediately before merging rather than relying on the check from five hours earlier.

PR #808, opened 2026-08-29
test(site): make rune modules testable, and test one

tbraun96:
Six `.svelte.js` modules export nine items and **not one was reachable from a test**. `$state` is a compiler construct, so `bun test` cannot import these files at all — every check on them has been "read it carefully".

That gap is not abstract. A latching-state regression in `fleet.svelte.js` reached main this morning (#807): the unpaired branch stopped rescheduling its own probe and nothing else could re-arm it, so the page could only be recovered by a full reload. It was found by an audit reading the call graph, because nothing could drive the code.

## The harness

`test-runes.js`, loaded with `--preload`, does two things:

- compiles the runes with svelte's own `compileModule` — no new dependency, the compiler is already here for the build
- resolves SvelteKit's `$lib` alias, which vite supplies during a build and bun does not

Both `test:unit` and `site.yml` preload it, so this runs in CI rather than only on my machine.

## Five tests, each pinning a decision the code explains

- a **virtual or loopback** address is never offered as the one a *peer* should dial
- link **class** decides and speed only breaks ties — a 1 Gb RoCE link still beats a 100 Gb ethernet one; sorting on speed first would hand a DGX pair the wrong interface
- `unverified` loses to every known class and is **still offered** when it is all there is, because it is missing information rather than a bad link
- `linkWarns` stays silent for it, for the same reason — warning would invent a problem out of an absence
- `isStale` measures against the sample

## Verified by mutation

```
swap the sort keys (speed first)  -> "ranks fabrics above ethernet…" FAILED
warn on unverified                -> "linkWarns stays silent…"        FAILED
```

Worth recording: my **first attempt at the ranking mutation silently did not apply**, and the giveaway was the *wrong* test failing. Both were re-run with the edited lines printed back before trusting the result.

488 pass, 0 fail.

### DIFF
--- a/.github/workflows/site.yml
+++ b/.github/workflows/site.yml
@@ -56,7 +56,7 @@ jobs:
       # `src/lib` only: the e2e specs under `e2e/` need a browser and a served
       # build, and are a separate concern from these.
       - name: Run unit tests
-        run: bun test src/lib
+        run: bun test --preload ./test-runes.js src/lib
 
   build:
     name: Build SvelteKit site
--- a/site/package.json
+++ b/site/package.json
@@ -10,7 +10,7 @@
     "generate": "node scripts/gen-models.mjs && node scripts/gen-benchmarks.mjs && node scripts/gen-gates.mjs && node scripts/gen-ladder.mjs && node scripts/gen-stars.mjs",
     "test:e2e": "bun x --bun playwright test",
     "test:e2e:live": "LIVE=1 bun x --bun playwright test --grep @live",
-    "test:unit": "bun test src/lib"
+    "test:unit": "bun test --preload ./test-runes.js src/lib"
   },
   "devDependencies": {
     "@playwright/test": "^1.62.1",
--- a/site/src/lib/agent/fleet.svelte.js
+++ b/site/src/lib/agent/fleet.svelte.js
@@ -68,7 +68,10 @@ export function linkWarns(cls) {
   return cls !== 'roce' && cls !== 'infini_band' && cls !== 'unverified';
 }
 
-class FleetSession {
+// Exported for tests. The singleton below is what the app uses; a test needs a
+// FRESH session with a fake agent, because the interesting behaviour is the
+// probe loop's state machine and a shared instance carries state between cases.
+export class FleetSession {
   /** The one client the whole app shares. */
   agent = new AgentClient();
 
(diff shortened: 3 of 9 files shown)


tbraun96:
**What this does and does not unlock**, since the motivating bug is still not covered by a test.

It unlocks the **pure exports** of rune modules, which were previously unreachable — that is the five tests here.

It does **not** yet reach the state machines, and #807's latching bug is one of those. Driving it would need three things, and only the first is done:

1. ~~import a `.svelte.js` module at all~~ — done here
2. export `FleetSession` (its `agent` field is already public, so a fake can be swapped in without constructor injection)
3. control time — `#scheduleProbe` is `#`-private and re-arms itself through `setTimeout`, so a test has to advance the clock rather than wait on it. bun has no reliable fake-timer story, so this is the real blocker, not the privacy.

I checked the other modules while I was here rather than assuming they were worth it:

- `clock.svelte.js` exports `nowMs`/`useClock`, but they are a thin wrapper over `ticker.js`, which already has its own tests — little to gain.
- `joinstate.svelte.js` and `session.svelte.js` export only their singletons, so the same step 2/3 applies.

Recording it here so the next person does not rediscover that the harness landed and the state machines still are not covered. The five tests are real; the gap that let #807 through is narrowed, not closed.

tbraun96:
**The gap I said was still open is now closed** — the state machine is testable, and #807's regression is pinned.

I claimed earlier that reaching the probe loop needed fake timers, which bun does not reliably provide. That was wrong: you do not need to *advance* time, only to *capture* the callback `setTimeout` was handed and invoke it. One tick, deterministically, no clock involved.

```
an unpaired probe keeps probing, so a token pasted elsewhere is noticed
```

It drives the real path — `start({watch:true})` with a fake agent that is absent, then swaps in one that answers but is unpaired — and asserts the loop **re-armed**. Against `main` it fails with `scheduled.length` = 1; with #807 merged in it passes. Both directions verified, which is why #807 is merged into this branch: a test that fails on main cannot land on its own.

Two bugs in my own test scaffolding on the way, both worth naming because each produced a failure that looked like the code under test:

1. `withCapturedTimers` was `try { return fn(...) } finally { restore }` with an **async** `fn`, so the `finally` ran the moment the promise was created — `setTimeout` was restored before the body ever called it, and the assertion saw zero scheduled timers.
2. The first version drove `retry()` instead of `start()`. The probe callback returns immediately unless the session has been started, so it scheduled a timer whose body never ran — it would have passed or failed for reasons unrelated to the behaviour.

489 tests pass. `FleetSession` is now exported, which is what makes any future test of this state machine possible at all.

PR #813, opened 2026-08-29
fix: docs headers, the wrong reason a Spark 'cannot run models', and llms.txt on all three sites

tbraun96:
Two things, both landing on the same origin config and the same docs:

1. `docs.atlasinference.io` was serving **every** response with no security headers.
2. Two DGX Spark owners could not run a model, and the site told the second one the wrong reason.

Plus `llms.txt` on all three properties, and the PWA manifest, which had not followed the palette move.

The atlasctl half is **[atlas-recipes#174](https://github.com/Avarok-Cybersecurity/atlas-recipes/pull/174)** — the two are designed together and should land together.

## 1. The docs vhost (already applied to the origin)

nginx's `add_header` does not accumulate: a location declaring any `add_header` discards every inherited one. Measured before touching it:

| URL | before | after |
|---|---|---|
| `/index.html`, `/`, `/404.html` | cc=300, **no security headers** | all three ✅ |
| `/book.js`, `/css/general.css` | cc=1y immutable, **none** | cc=1h must-revalidate, all three ✅ |
| `/favicon.png` | cc=30d, **none** | all three ✅ |

Only 301 redirects, which match no location, were ever protected.

**A second defect underneath, which is why the first was invisible.** After the reload the origin was right and the public site was not: Cloudflare was serving `cf-cache-status=HIT`, `book.js` **age 15.6 days**, `css/general.css` **age 29.4 days**. Those were marked `immutable, max-age=31536000` — a promise only keepable for a content-hashed name. **10 of 604** CSS/JS files under the docs root are hashed; mdBook regenerates the other ~594 under fixed names every build. So *every docs deploy has been invisible to returning visitors for up to a year.*

> ⚠️ **A Cloudflare purge of `docs.atlasinference.io` is still needed** and could not be done from here — the origin's token is DNS-scoped and returns `Authentication error` on `purge_cache`.

## 2. "This machine cannot run models"

A Spark owner clicked Run and got that, plus a pairing code for a second machine. Their Spark was fine — they were not in the `docker` group. The agent said exactly that, and `placement.js` threw the sentence away.

- `decide()` now carries the local node's `cannotLaunchReason` (already populated by `ingest.js`, already used by `verbs.js`) into the `none` result.
- When the cause is fixable here, `canOnboard: false` — pairing a laptop does not fix a `usermod` away.
- `LaunchDialog` renders the two commands atlasctl prints, the `sudo atlasctl` warning, and the troubleshooting link, instead of a pairing code. It runs nothing.

**And a precedence bug.** `session.svelte.js` had `fleet.localCanLaunch ?? this.agent.canLaunch` — which reads as "prefer the fleet", but `??` only falls back on null, so a **stale `false`** from a fleet snapshot beat the `true` the dialog's own handshake had just returned. Now gated on `phase === 'ready'` rather than `??`, because `canLaunch` initialises to `false`, not null — a bare `??` would never reach the fleet, and an in-flight handshake would read as "cannot launch". (I wrote the `??` version first; catching this is why the initial value got checked.)

## 3. llms.txt on all three

| property | source | stays current because |
|---|---|---|
| atlasinference.io | `site/scripts/gen-llms.mjs` | already generated; now cross-links the other two |
| blog.atlasinference.io | `blog/src/routes/llms.txt/+server.js` | prerendered from the post index |
| docs.atlasinference.io | `book/scripts-gen-llms.mjs` → `book/src/llms.txt` | generated from `SUMMARY.md` |

None hand-maintained. Each links to the other two, because an agent that finds one has no other way to reach the rest.

## 4. PWA manifest

`site.webmanifest` still said `#14111f` while the page is `#0F1216`, so an installed PWA painted violet window chrome and splash around a correct page. **The page colour is unchanged** — the built stylesheet is byte-identical to main's (`md5 8e5cf6e5`). `theme-color.test.js` now reads the manifest too; checking only the two `app.html` files is the gap that let it drift.

## Verification

| | |
|---|---|
| site unit tests | **512 pass** |
| blog unit tests | **34 pass** |
| contrast budget | PASS, tightest 4.61:1 |
| live header check (blog) | 28/28 |
| cross-link check | 18/18 |
| built page CSS vs main | **identical** |

Controls, each a reintroduced defect watched going red: discard the reason (24/1), keep offering onboarding for a fixable cause (24/1), `localReason` reads the first node not the local one (23/2), old manifest colour (3/1), plus the four on the atlas-recipes side.

https://claude.ai/code/session_01726abFdibR9YTp6URgzVh3


tbraun96:

Issue #812, opened 2026-08-29
investigate(perf): Qwen3.8 C4 bimodality and benchmark provenance


TheTom:
## Current diagnosis

The dashboard screenshots do not establish an across-the-board runtime regression. The latest Qwen3.8 concurrency record remains inside historical ranges at C1, C8, and C16, and the latest decode-floor result improved slightly. The conspicuous result is C4 at about 40.6 tok/s, but that is a recurring historical low mode rather than a newly introduced value.

This issue tracks two confirmed measurement defects and one unresolved C4 bimodality investigation.

### The dashboard drew false before-and-after edges

The chart connected records by `recorded_at` even when their measured commits were on divergent histories. Squash-merged record files can appear mainline even when the measured SHA is not an ancestor of the next point. Hostname suffixes also changed on the same physical Spark.

PR #814 makes chart edges require comparable model, parameters, instrument, durable machine identity, and Git ancestry. Unknown lineage stays disconnected.

### The concurrency warm-up did not reach its claimed setup

The runner warmed one prompt tagged `warm`, then measured `c0`, `c1`, and so on. The distinguishing tag is the first message content, so the explicit warm-up did not prime the measured prompt. ISL-major execution then made each rung a different mix of prompts measured by earlier rungs and newly introduced prompts.

PR #815 warms every exact measured prompt and requires runtime evidence that at least 80% of each measured prompt was served from cached state. Missing usage or a partial template-prefix hit makes a warmed cell non-comparable and a gate inconclusive.

### Historical records omit variables needed for RCA

The self-started server runs inside the benchmark process and inherits scheduler controls such as:

- `ATLAS_PREFILL_CODISPATCH`
- `ATLAS_PREFILL_CODISPATCH_WINDOW_MS`
- `ATLAS_PREFILL_CODISPATCH_SETTLE_MS`

The record does not preserve their effective values. It also omits admitted prefill cohort widths and per-request cache evidence. Two records can therefore share a tree, recipe name, and serve overrides while executing different admission behavior.

PR #816 separately corrects the coverage policy so the concurrency-only instrument does not demand nine unrelated benchmark campaigns. Shared HTTP, server, model, kernel, policy, and boundary paths remain fail-closed.

### DFlash2 does not explain C4

[...]

PR #814, opened 2026-08-29
site: stop joining unrelated benchmark histories

TheTom:
## Summary

The benchmark dashboard currently sorts receipts from the checked-out tree and remote branch heads by `recorded_at`, then draws one continuous path. That makes a chart edge look like a before/after code transition even when the two measured SHAs are on divergent histories.

This change records a trend predecessor only when:

1. the earlier measured SHA is an actual Git ancestor of the later SHA;
2. model, recipe, GPU, driver, performance class, parameters, and serve overrides match.

Unknown or divergent lineage stays visible as points without an invented line. The record card also discloses whether a receipt came from a remote branch head, is outside the dashboard build's commit history, or cannot be resolved from available Git history.

Refs #812.

## Broken case and proof

The dashboard placed `0c402bac00` immediately before `d8bc9c9d86` and drew a C4 drop from 45.668 to 40.625 tok/s. Git proves `0c402bac00` is not an ancestor of `d8bc9c9d86`, so that edge was not a valid regression comparison.

The divergent-history test is the realistic known-bad replay. The previous timestamp-only renderer joins `left` to `right`. The new contract requires two separate edges from their common ancestor and rejects a `left` to `right` edge. Neighboring tests cover a linear history, returning to an older branch after a divergence, unknown commits, different models, changed drivers and parameters, and parameter key ordering.

## Test plan

- [x] Production site build: `ATLAS_RECIPES_ROOT=/tmp/atlas-recipes-lineage-812/recipes npm run build`
- [x] Generator: 578 records collected from 331 branch heads
- [x] Seven lineage contracts replayed with Node assertions
- [x] Generated-data mutant check confirms no `0c402bac00 -> d8bc9c9d86` edge
- [x] Browser QA of the benchmark modal, sparse lineage paths, point selection, and provenance card
- [x] `git diff --check`
- [x] Added focused unit tests
- [ ] `bash scripts/check-license-headers.sh` requires a running Docker daemon. The license SSOT only covers `crates`, `kernels`, and `cuda_kernels`; this PR changes `site` files only. Both new JavaScript files still carry SPDX headers.
- [ ] Real model or GPU testing is not applicable. No runtime inference or benchmark value changes.

## Notes for reviewers

Only ten of the 578 currently collected receipts have a comparable predecessor that can be proven from the Git objects available to the generator. Most historical measured SHAs disappeared when feature branches were squash-merged or deleted. Those receipts remain visible, but the dashboard no longer presents their timestamp adjacency as code lineage.

This does not solve recipe identity or repeated-sample retention. The records still need a resolved recipe hash and unique run identity before the dashboard can support controlled performance comparisons from committed receipts alone.

## CLA


TheTom:
Follow-up from the #812 record audit: `perf_class` is not a stable machine identity on the Spark.

The records labeled `gb10@spark-256a`, `gb10@spark-43fa`, and `gb10@spark-28c2` repeatedly carry the same `hardware_state.before.machine.machine_id`. The hostname suffix changes while the underlying machine identity remains fixed.

The generator now emits that machine ID. The lineage key prefers it and uses `perf_class` only for legacy records that have no machine ID. A legacy hostname identity and a machine-ID identity remain disconnected because equivalence cannot be proved.

RST replay:

- same machine ID plus different hostname must connect when Git ancestry and all other instrument fields match;
- different machine IDs plus the same hostname must not connect;
- machine ID versus legacy hostname-only data must not connect.

Forcing the implementation back to hostname identity made both new tests fail, including an incorrect edge to a different machine. Restored result: 9/9 lineage tests pass. The production site build also passes with the recipe SSOT supplied; existing Svelte and CSS warnings remain non-failing.

tbraun96:

PR #815, opened 2026-08-29
fix(bench): control concurrency warmup prompts

TheTom:
## Summary

The concurrency sweep claimed each unmeasured warm-up used the same prompt as its measured batch. It actually warmed one prompt tagged `warm`, then measured prompts tagged `c0`, `c1`, and so on. Because the tag is the first prompt content and prefix caching is enabled, the setup did not prime the named measured prompts.

This change makes `warmup` count full rounds. Each round runs every exact measured prompt before timing the same prompt set. A failed warm-up now fails cell setup.

The measured responses then provide the runtime oracle. A warmed cell is comparable only when every request reports at least 80% of its prompt tokens as cached. Missing usage or a smaller partial-prefix hit makes a gating sweep `INCONCLUSIVE` before throughput floors are checked.

Refs #812.

## Broken setup and proof

The old rung history was:

```text
warm-up: warm
C1:      c0
C4:      c0 c1 c2 c3
C8:      c0 c1 c2 c3 c4 c5 c6 c7
```

The explicit warm-up was unrelated, while later rungs reused a growing subset of prompts measured by earlier rungs. The ladder mixed cached and new prompts according to prior rung history.

The repaired prompt plan is consumed directly by `run_cell`. Its test requires two C4 warm-up rounds to equal two complete copies of `c0..c3`, requires `warmup=0` to schedule no warm-up, and proves the historical `warm` prompt differs from the measured bytes.

Replacing the repaired rounds with `[[warm], [warm]]` fails the exact prompt-plan test.

## Runtime oracle and mutants

Equal request bytes show the setup targets the intended cache key. They do not prove the runtime served the request from that cache. A nonzero hit is also insufficient because the wrong prompt can share a small chat-template prefix.

The repaired partitions are:

- warmed request at 80/100 cached prompt tokens: comparable
- warmed request at 79/100: uncontrolled
- warmed request with missing usage: uncontrolled
- intentionally cold `warmup=0` with zero cache: comparable
- gating ladder at 999 tok/s with one uncontrolled warmed cell: fail as `INCONCLUSIVE`

Replayed mutants:

1. Historical `warm` tag instead of the measured prompt set fails the prompt-plan oracle.
2. Removing the cache condition from `CellRow::comparable` fails the comparability oracle.
3. Removing the gating cache guard lets the 999 tok/s bad ladder pass and fails the verdict oracle.
4. Replacing the 80% check with `cached_prompt_tokens != 0` lets the 79/100 partial hit survive and fails the boundary oracle.

Records now include `min_cached_prompt_tokens`, `min_cached_prompt_pct`, and `cache_uncontrolled_cells`. Run evidence prints cached versus total prompt tokens for every request.

## Verification

- [x] `cargo test -p atlas-plugin benchmarks::concurrency -- --nocapture` (21 passed)
- [x] `cargo clippy -p atlas-plugin --tests -- -D warnings`
- [x] `cargo fmt --check`
- [x] `git diff --check`
- [x] Replayed all four known-bad mutants above and required the named tests to fail
- [ ] Authentic current-SHA concurrency run on exclusive GB10 hardware

No Spark run was used for this PR validation.

## Record boundary

This is a functional benchmark-instrument change. Existing concurrency records and floors were measured with mixed or unobserved cache state and are not comparable to the repaired instrument. `BENCH.toml` states that boundary instead of changing a floor without a new measurement.

The current coverage policy maps the flat concurrency driver to all ten mandatory gates. PR #816 narrows that deterministic path mapping while keeping shared HTTP, server, model, kernel, policy, and boundary files fail-closed. Once that policy is reviewed, this instrument still requires an authentic current-SHA `concurrency-sweep` record and new floors before review completion.

The longer duration hint is intentional. `warmup=1` now executes one exact warm-up request per measured request, so setup cost scales with concurrency.

## CLA

### DIFF
--- a/crates/atlas-plugin/src/benchmarks/concurrency.rs
+++ b/crates/atlas-plugin/src/benchmarks/concurrency.rs
@@ -102,6 +102,7 @@ const CODE_TASK: &str = "Ignore the reference text above. Task: write a complete
 /// its whole cell vacuous. 80%: a natural stop a few tokens early is fine; a
 /// 49-token burst against a 512-token budget is not a throughput measurement.
 const VACUITY_FLOOR: f64 = 0.8;
+const WARM_CACHE_FLOOR: f64 = 0.8;
 
 /// What one completed request actually delivered — the evidence a cell's
 /// tok/s stands on. `http::ChatOutcome` already parses all of this from the
@@ -110,6 +111,7 @@ const VACUITY_FLOOR: f64 = 0.8;
 #[derive(Clone, Debug, Default)]
 struct RequestEvidence {
     completion_tokens: usize,
+    prompt_tokens: usize,
     cached_prompt_tokens: usize,
     finish_reason: Option<String>,
     server_ttft_ms: Option<f64>,
@@ -127,9 +129,11 @@ fn cell_is_vacuous(requests: &[RequestEvidence], osl: usize) -> bool {
 
 fn cache_is_uncontrolled(requests: &[RequestEvidence], warmup: usize) -> bool {
     warmup > 0
-        && requests
-            .iter()
-            .any(|request| request.cached_prompt_tokens == 0)
+        && requests.iter().any(|request| {
+            request.prompt_tokens == 0
+                || (request.cached_prompt_tokens as f64)
+                    < WARM_CACHE_FLOOR * request.prompt_tokens as f64
+        })
 }
 
 /// The prompt identities one cell executes before and during measurement.
@@ -169,6 +173,18 @@ impl CellRow {
     }
     fn min_cached_prompt(&self) -> Option<usize> {
         self.requests.iter().map(|r| r.cached_prompt_tokens).min()
+    }
+    fn min_cached_prompt_pct(&self) -> Option<f64> {
+        self.requests
+            .iter()
+            .map(|request| {
+                if request.prompt_tokens == 0 {
+                    0.0
+                } else {
+                    request.cached_prompt_tokens as f64 / request.prompt_tokens as f64 * 100.0
+                }
+            })
+            .reduce(f64::min)
     }
     /// Clean and above the vacuity floor — the only rows metrics may quote.
     fn comparable(&self) -> bool {
@@ -310,6 +326,7 @@ impl ConcurrencySweep {
                     tokens += o.completion_tokens;
                     requests.push(RequestEvidence {
                         completion_tokens: o.completion_tokens,
+                        prompt_tokens: o.prompt_tokens,
                         cached_prompt_tokens: o.cached_prompt_tokens,
                         finish_reason: o.finish_reason.clone(),
                         server_ttft_ms: o.server_ttft_ms,
@@ -325,8 +342,8 @@ impl ConcurrencySweep {
         let vacuous = cell_is_vacuous(&requests, self.osl);
         // Matching prompt bytes reach the intended setup, but the endpoint's
         // usage is the oracle for whether the measured request actually used
-        // cached prompt state. A requested warm path with a zero or omitted
-        // cached-token count is a different instrument.
+        // the warmed prompt. A small shared chat-template prefix is not enough:
+        // require a material cached fraction of each measured prompt.
         let cache_uncontrolled = cache_is_uncontrolled(&requests, self.warmup);
         handle.info(evidence_line(isl, conc, &requests));
         if vacuous {
@@ -345,7 +362,8 @@ impl ConcurrencySweep {
         if cache_uncontrolled {
             handle.warn(format!(
                 "isl {isl} conc {conc}: warm-up was requested but at least one measured request \
-                 reported zero cached prompt tokens — this cell is NOT comparable"
+                 reported less than {:.0}% of its prompt as cached — this cell is NOT comparable",
+                WARM_CACHE_FLOOR * 100.0,
             ));
         }
         Ok(CellRow {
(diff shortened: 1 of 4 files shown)


TheTom:
### Warm-cache oracle added

The prompt-plan repair now observes the runtime result it names instead of treating equal request bytes as proof of cache use.

Each measured response contributes `cached_prompt_tokens`. When `warmup > 0`, a zero or omitted count makes that cell non-comparable. A gating sweep fails as `INCONCLUSIVE` before evaluating throughput floors. The record also carries `min_cached_prompt_tokens` and `cache_uncontrolled_cells`, and the run log prints the per-request cached-token vector.

The test partitions are:

- warm-up requested and one measured request reports zero: reject the cell
- warm-up requested and every measured request reports cached tokens: retain the cell
- `warmup=0` and zero cached tokens: retain the intentionally cold cell
- arbitrarily high throughput with one uncontrolled warm-cache cell: fail the gate

Mutant replay:

1. Removing `!self.cache_uncontrolled` from `CellRow::comparable` lets the bad cell quote throughput. The exact comparability test fails under this mutant.
2. Removing the gating cache guard lets a synthetic 999 tok/s ladder pass despite no observed warm-cache use. The exact verdict test fails under this mutant.

Repaired verification at commit `db124226f`:

- `cargo test -p atlas-plugin benchmarks::concurrency -- --nocapture`: 21 passed
- `cargo clippy -p atlas-plugin --tests -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

No Spark run was used. A real current-SHA concurrency record remains necessary after the instrument and coverage-policy work are reviewed.

TheTom:
### Cache-oracle correction

The prior nonzero-cache criterion was too weak. A wrong measured prompt can still share chat-template prefix blocks with the warm-up and report a small nonzero cache hit.

Commit `a4da66868` now requires every measured request in a warmed cell to report at least 80% of its prompt tokens as cached. The boundary is inclusive: 80/100 is controlled, 79/100 is not. Missing prompt usage is also uncontrolled. Intentionally cold `warmup=0` remains a separate valid partition.

The record now includes `min_cached_prompt_pct`, and the evidence line prints cached versus total prompt tokens for each request.

Mutant replay replaced the fraction check with the plausible old condition `cached_prompt_tokens != 0`. A 79/100 partial hit then survived production, and the boundary test failed at its exact rejection assertion. Restoring the fraction oracle passes the same test and all 21 focused concurrency tests. Package test Clippy, formatting, and diff hygiene also pass.

This correction supersedes the nonzero wording in my preceding comment.

TheTom:
### Result-table oracle corrected

The prior table contradicted the gate decision: a cache-uncontrolled cell was excluded from comparable metrics but its throughput cell still rendered as `Good` with no marker. It also displayed cached-token count without the prompt-token denominator used by the decision.

Commit `311100cce` now renders either vacuous or cache-uncontrolled throughput as `Bad` with `*`, and the evidence column reports the minimum cached prompt percentage.

The old-renderer mutant removed `r.cache_uncontrolled` from the style decision. The production table test failed exactly at `left: Good, right: Bad`. Restoring the condition produces `100.0*`, style `Bad`, and cache percentage `0` for the rejected fixture.

All 21 focused concurrency tests pass, package test Clippy passes with warnings denied, formatting and diff checks pass, and the test module is 499 lines under the enforced 500-line cap.

tbraun96:
## The controlled warm-up is serial, and that is what makes a wide ladder expensive

Measured while extending this instrument to C=1..128 in #818, on spark-256a with this branch's warm-up in the tree.

`run_cell` awaits each warm-up prompt one at a time:

```rust
for tag in tags {
    self.one(isl, tag.clone()).await.with_context(...)?;
}
```

while the measured batch is a `join_all`. At osl 320 and ~21 tok/s single-stream, each warm-up prompt costs ~15 s of wall, so a cell's warm-up costs `C × 15 s`. Observed directly in the serve log — prefills landing at 15-16 s intervals, one at a time, with decode at single-stream rate throughout:

```
20:05:52  20:06:09  20:06:24  20:06:39  20:06:55
```

Across the ladder that is `ΣC × 15 s`. For the current gated shape (1,4,8,16) it is ~7 minutes — invisible. For 1,2,4,8,16,32,64,128 it is **255 × 15 s ≈ 64 minutes of warm-up per rep**, against roughly 12 minutes of actual measurement. The instrument spends 84% of its wall time priming.

That is not an argument against the controlled warm-up — the defect it fixes is real and the fix is right. It is an argument that **the warm-up round can be concurrent**, and I think it can be, for a reason specific to how this plan is built:

`prompt_plan` gives every request in a cell a distinct tag (`c0..c{conc-1}`), and the warm-up round is `measured.clone()`. So the prompts within one round are **pairwise distinct cache keys**. There is no duplicate-insert race for concurrency to lose — which is the usual reason to serialise a cache prime.

If anything the serial form is the weaker guarantee at wide C: by the time prompt 127 is warmed, prompt 0 has been sitting in the cache for 30 minutes and is the most likely eviction candidate, so the measured batch can find its earliest prompts cold. A concurrent round keeps the whole set live inside one window. The 80% cached-token oracle you added would catch that either way, but it would catch it as an inconclusive gate rather than as a fast run.

I have not changed it here — it is your PR and your semantics, and #818 is measuring against the serial form so its floors are consistent with what is in the tree. Flagging it because the cost only becomes visible at the ladder widths #818 adds, and because "make the warm-up round a `join_all`" looks like a two-line change that keeps every property the round was written to have.

One related note for #818's benefit: this run has `ATLAS_PREFILL_CODISPATCH` unset, so prefills serialise. That is the default a gate run inherits, and it is exactly the variable #812 flagged as unrecorded — #818 now writes the resolved value onto the record, so a future reader can tell these two regimes apart instead of inferring them from wall time.

tbraun96:

PR #816, opened 2026-08-29
fix(gate): scope concurrency driver coverage

TheTom:
## Claim

The two production files for `concurrency-sweep` are flat files. The current coverage table excludes directory-shaped benchmark drivers from unrelated gates but never classifies `concurrency.rs` or `concurrency_verdict.rs`. A change to either therefore falls through to all ten mandatory GPU gates.

This PR makes those exact files invalidate only `concurrency-sweep`. It also registers `concurrency_tests.rs` as a test-only Rust module after proving its only module edge is guarded by `#[cfg(test)]`.

## Reproduction

Before this change:

```rust
coverage::invalidated_by([
    "crates/atlas-plugin/src/benchmarks/concurrency.rs",
])
```

returned every mandatory gate, including BFCL, vision, video, agentic, TTFT, SSM poisoning, and decode floor. None of those instruments imports or executes the concurrency driver.

## Boundary kept fail-closed

The exclusions name the two production files exactly. They do not exempt the benchmark directory. Tests require shared HTTP, statistics, transcript, server, kernel, lookalike, and coverage-policy paths to continue invalidating all ten gates. `concurrency-sweep` still invalidates for both production files.

The test-only exemption is also exact. A proof test scans the Rust tree, requires the registered `#[cfg(test)]` and `#[path]` edge, rejects a second path/include edge, and keeps neighboring test-looking paths fail-closed.

## RST evidence

Two known-bad mutants were replayed locally:

1. Replacing the exact file exclusions with `benchmarks/concurrency` restored the all-ten result. `the_flat_concurrency_driver_invalidates_only_its_own_gate` failed with the exact unexpected gate set.
2. Removing the concurrency test module from the test-only decision reopened GPU gates. `registered_test_modules_do_not_invalidate_gpu_records` failed on that exact path.

The repaired code rejects both mutants.

## Local verification

```text
cargo test -p atlas-plugin gate:: -- --nocapture
224 passed, 0 failed

cargo clippy -p atlas-plugin --tests -- -D warnings
PASS

cargo fmt --check
PASS

git diff --check
PASS
```

## Bootstrap status

The one-time grant is pinned to only the final `coverage.rs` blob `ff3f5109bc3159555bb518b7c4d132d5335ad3fd`. Any later edit changes the OID and restores all-ten invalidation. The expiry test requires removal after every required gate has a record newer than the grant epoch. No benchmark result is inferred by this policy test.

## CLA

- [x] I have read and agree to the [Contributor License Agreement](../CLA.md).

PR #817, opened 2026-08-29
dflash: unlock the gamma>8 ladder: wyn K=9..16 GDN verify + FP8 rt2 MAX_M=16 propose (63.0 tok/s)

rrstesiak:
## What

Two kernels that open the γ>8 speculation window for Qwen3.8-27B DFlash2:

1. **`gated_delta_rule_wyn.cu`: K=9..16 instantiations** of the existing wyN template. K=9..16 was the only un-served verify width band: it fell to the sequential per-token GDN fallback (per token: conv launch + gdn launch + 2 state D2Ds, every GDN layer).
2. **`fp8_gemv_rt.cu`: templatized on RT_MAXM, `fp8_gemv_rowscale_batch16_rt2` added** and dispatched at all three drafter propose sites (projections x2, vocab lm_head). The `batch8` instantiation is byte-identical to the pre-template kernel; the M<=8 record path is untouched.

## Why (measured, not theorized)

`ATLAS_DFLASH_STEP_TIMING=1` split the step at flag 8 vs flag 9: **verify flat (90.8 → 87.6 ms), propose doubled (18.2 → 38.0 ms)**: the entire γ>8 tax was the drafter falling off the M<=8 rt2 GEMV family onto the padded tile GEMMs (87%/50% M-tile padding, ~100 GB/s vs 180+). Post-fix propose: **19.8–21 ms** at flags 9–10.

The drafter's trained block is 8 (checkpoint `dflash_config.block_size`); the flag convention is drafts+1, so flag 8 ran 7 drafts: one short of the trained block. With the window open, the drafter chains **two past its block at full 9/9 accepts**; flag 11 is a measured wall (tok_step flat at ~7.6 for +1 row of cost), so the useful horizon is exactly 10 rows and **`--dflash-gamma 10` is the optimum**.

## Result (GB10, single Spark, single stream, temp 0, median of 3, fresh boot)

| workload | before (flag 8) | after (flag 10) |
|---|---|---|
| MinHeap hard-code (875-tok class) | 56.2 | **63.0** (66.6 / 62.1 / 63.0) |
| Volvo prose (300 tok) | 23.7 | 23.8 |
| Fib patterned-code (~215 tok) |: | 70.8 |

Outputs reproduce **byte-identically per run-position across separate boots** (r1 `e23c9083…`, r2 `b6e29d25…`, r3 `fc44fc22…`, two boots); tok/s reproduce ±0.1 per position. Full reproduction key (launch env, checkpoints, receipts with shas) available on request.

## Safety

- Stale kernel builds resolve zero handles → exact current behavior (sequential GDN fallback / tile propose).
- Kill switches unchanged: `ATLAS_GDN_WYN=0`, `ATLAS_NO_DFLASH_FP8_RT=1`.
- Drafter-side numerics are correctness-free under strict-argmax accept (module contract); GDN wyN widths follow the existing wy17 pool-layout contract with the same fail-safe contiguity guard.
- `cargo check` clean; spark-model lib suite 633/633.

### DIFF
--- a/crates/spark-model/src/factory/build.rs
+++ b/crates/spark-model/src/factory/build.rs
@@ -682,6 +682,7 @@ pub fn build_model(
                 args.window_size,
                 model.gpu_backend(),
                 max_seq_len,
+                max_batch_size,
             )?;
             model.set_dflash_proposer(std::sync::Arc::new(head));
             tracing::info!("DFlash drafter installed as the active proposer");
--- a/crates/spark-model/src/layers/dflash_head.rs
+++ b/crates/spark-model/src/layers/dflash_head.rs
@@ -101,6 +101,12 @@ pub struct DflashKernels {
     /// Kill-switch: ATLAS_NO_DFLASH_FP8_RT=1. provenance-id:
     /// 526f6e616c6420522e205374657369616b
     pub fp8_gemv_rt2: KernelHandle,
+    /// MAX_M=16 sibling of `fp8_gemv_rt2` for the γ>8 propose window
+    /// (2026-08-29: STEP_TIMING measured propose 18.2ms rt2 vs 38.0ms tile
+    /// fallback at flag 9 — the entire γ>8 step tax). `.0 == 0` on stale
+    /// kernel builds → tile path, exactly as before.
+    /// provenance-id: 526f6e616c6420522e205374657369616b
+    pub fp8_gemv_rt2_16: KernelHandle,
     /// DFlash2 two-tap grouped dynamic conv (`kernels/gb10/common/dflash2.cu`).
     /// `.0 == 0` on targets without the module (DFlash2 then refuses to arm).
     pub dflash2_conv2: KernelHandle,
--- a/crates/spark-model/src/layers/dflash_head/forward_block.rs
+++ b/crates/spark-model/src/layers/dflash_head/forward_block.rs
@@ -592,6 +592,27 @@ impl BlockDiffusionDraftHead {
                             h_local,
                             stream,
                         )?;
+                    } else if self.kernels.fp8_gemv_rt2_16.0 != 0
+                        && self.gamma as u32 <= 16
+                        && h_local.is_multiple_of(16)
+                        && super::fp8_rt_enabled()
+                    {
+                        // γ>8 propose window: MAX_M=16 rt2 sibling over the
+                        // vocab. 2026-08-29 STEP_TIMING measured the whole
+                        // γ>8 step tax in this propose tail (18.2 -> 38.0ms
+                        // at flag 9); the m16 tile below pads 50%+ of its
+                        // rows and traced 12.2ms/step at ~104 GB/s.
+                        ops::fp8_gemv_rowscale_batch16_rt2(
+                            gpu,
+                            self.kernels.fp8_gemv_rt2_16,
+                            norm_noise_local,
+                            fp8,
+                            self.scratch.logits,
+                            self.gamma as u32,
+                            self.vocab_size as u32,
+                            h_local,
+                            stream,
+                        )?;
                     } else {
                         ops::fp8_gemm_n128_row_scaled_m16(
                             gpu,
--- a/crates/spark-model/src/layers/dflash_head/forward_block_layer_paged.rs
+++ b/crates/spark-model/src/layers/dflash_head/forward_block_layer_paged.rs
@@ -257,6 +257,26 @@ impl BlockDiffusionDraftHead {
                         stream,
                     );
                 }
+                // γ>8 propose window (flags 9..17): MAX_M=16 rt2 sibling.
+                // 2026-08-29 STEP_TIMING: propose 18.2ms (rt2) vs 38.0ms
+                // (this tile fallback) at flag 9 — the whole γ>8 step tax.
+                if self.kernels.fp8_gemv_rt2_16.0 != 0
+                    && g <= 16
+                    && k_in.is_multiple_of(16)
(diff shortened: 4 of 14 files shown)


tbraun96:
Stacked into **#818**, which is the single certification scope for #812's ladder work. Merging it there rather than separately because benchmark records cannot compose across PRs — two campaigns for two PRs is two campaigns.

This PR retires three findings I posted on #812 last night from measurements taken **without** these kernels. Worth recording explicitly, since those findings are now wrong in the issue thread:

1. **The memory wall is a missing kernel, not a structural limit.** I measured DFlash2 refusing to *start* above a narrow batch (bs=128 wanting a 68.7 GB reserve against ~46.5 GB available), traced it to the γ-sized verify pool, and documented that the f16-pool relief was unavailable because `gated_delta_rule_wy17` has no FP16 h-state twin — citing `cli/validate.rs`'s own refusal as evidence it was by design. `a34430185f` adds the twins for K=5..16 and puts DFlash under the f16 pool. My "the real fix is an FP16 h-state twin for wy17" line was right about the shape and wrong that it didn't exist yet.

2. **The C>4 decline is the drafter KV pool, not #655 failing to take effect.** I measured DFlash2 peaking at C=4 (50.32 / 49.90 / 50.11, σ=0.21) and falling to 32.8 at C=16, and read it as the flat shape #655 warns about persisting despite #655 being in the tree. `1eaf2549bd` names it: the Option-B pool was sized for exactly one sequence, so stream 1 took every block and 2..16 fell back to serial decode. Your receipt reports 68.0 aggregate at C=16 against my 32.8.

3. **The single-stream gap was γ, not the serve profile.** I measured C=1 at 23.4 against #648's 54.5 and attributed it to bs=16 / fp8 KV / 8 SSM slots. The actual reason is in your description: the flag convention is drafts+1, so `--dflash-gamma 8` ran seven drafts, one short of the drafter's trained block, and K=9..16 fell off the rt2 family onto padded tile GEMMs.

**Three conflicts needed judgement, not mechanics** — flagging them because two are in your code:

- `propose.rs` block allocation: I took yours **whole**. My stack carried `d303965f64` ("drafter KV pool winner-takes-all at C>=2", 08-19), an earlier fix to the same defect that grows the table incrementally from `ctx_len`. Yours supersedes it — pool sized from `max_batch_size`, `sort_unstable` pinning the LIFO ordering, partial grab freed before bail. Interleaving them would have produced a design neither of us measured.
- `from_weights.rs`: your `max_batch_size` and my `max_batch` are the same parameter. Kept your name (it is what the factory plumbs) and my doc, noting it now feeds both the banded scratch and the pool.
- `validate.rs`: your check assumes `dflash_gamma: usize`; #650 made it `Option<usize>` so the drafter's own `block_size` can be the default. It now rejects only an **explicit** over-16 request — an unset γ resolves from a checkpoint that is not readable at validate time, and that path is already covered by the runtime backstop.

One thing I could not reproduce and am not claiming: `qwen3_ssm/init.rs` crossed the 500-LoC cap with your +27 lines, so the two 12-entry wyN tables moved verbatim into `init_kernels.rs`. The index contract (index = K−5, FP16 twin sharing it) is untouched.

Next: re-measuring the DFlash2 ladder on the stacked tree at `--dflash-gamma 10` under the f16 pool. The gate's C=16 cap and its rung list were both derived from the pre-#817 walls and are provisional until that lands.

tbraun96:
## Heads-up: `main` moved, and one required check was deliberately bypassed

`main` is now `a3686670d2` — the squashed stack **#819** (#813 + #814 + #815 + #816). #813/#814/#815/#816 are closed as merged-via-stack.

**Read this before stacking #817 and #818.**

### The bypass, stated plainly

#819 was merged with the **`PR Benchmark Certifications`** check **red**, at the repository owner's explicit direction. All 40 other checks passed. To do it, two protections were briefly relaxed and both were **restored immediately and verified**:

| setting | during merge | now |
|---|---|---|
| `enforce_admins` | disabled | ✅ `enabled=true` |
| ruleset `main merge queue` (id 20739400) | disabled | ✅ `enforcement=active` |

**Nothing was force-pushed and no history was rewritten.** The previous `main` (`b474ee0f20`) is still an ancestor; the merge added exactly one commit. The `non_fast_forward` rule in the `Default` ruleset was never touched.

### Why the certification was red — this will hit your stack too

It is **not** a code defect. #815 edits the concurrency driver:

```
NONE  concurrency-sweep — latest record is for d8bc9c9d86 — invalidated by
      crates/atlas-plugin/src/benchmarks/concurrency.rs,
      crates/atlas-plugin/src/benchmarks/concurrency_verdict.rs
1 bench(es) still need a passing gate record: concurrency-sweep
```

The coverage model working correctly: change the instrument, and the old measurement no longer speaks for the new code. Discharging it needs a `concurrency-sweep --pull-request-gate` run (~75 min exclusive GPU), which was **deferred, not skipped silently** — hence this note.

Two things worth knowing:

* `PR benchmark gate` is only a **transitional alias** for `PR Benchmark Certifications`. Its own log says "read its log, not this one." Two red checks, one cause.
* Since #815 is now on `main`, a PR that does **not** itself touch the concurrency driver should no longer owe that record. Check whether yours does before assuming you inherit the debt.

tbraun96:

Landing commit: `fbca9ca5` — **verified on `main`** (merge commit of #818, merged earlier today).

wyn K=9..16 GDN verify + FP8 rt2 MAX_M=16 propose — the kernels the 66.6 tok/s default serve runs on. All 14 files this PR touches are present on `main`.

PR #818, opened 2026-08-29
atlas#812: C=128 ladder, DFlash2 as a mandatory gate, and the gamma>8 unlock (#817 stacked)

tbraun96:
Closes the actionable half of #812, in **one branch and one certification scope** — records cannot compose across PRs, so N perf PRs cost N campaigns.

#812's analysis finished negative: the C=4 mode is not a DFlash2 regression, not a kernel change, not a scheduler change. What it exposed is that **the instrument is too narrow and too under-recorded to answer the question**. This widens and hardens it.

## What is stacked here

| Source | Why it must be in this branch |
|---|---|
| **#699** | #655 names it the correctness baseline — before it, C≥2 accept/throughput figures were contaminated by verdict rewind. Verified main was still pre-#699. |
| **#649 → #650 → #655** (brings #651) | The train that makes DFlash2 batch at all. #655's ctx-rows fix takes C=2 acceptance 59% → 84%. |
| **#704** | `VERIFY_ROW_CAP` 96 → 160. At γ=8 the old cap made C=16 fall back to serial — "the feature turned itself off exactly at the concurrency it exists for". |
| **#814** | Dashboard lineage. Site-only, was already CLEAN. |
| **#815** | Controlled warm-up. Touches the same files as the rung change. |
| **#816** | Coverage scoping for the flat concurrency driver. |

Three merges needed hand resolution and each is called out in its commit; the notable one is that `#650`'s `record_alloc(ptr, bytes, site)` ledger and main's large-allocation tracing both survive, and `#651`'s `ssm_tp_all_reduce` signature change had to reach a fifth call site that main created after that branch forked.

## The ladder now runs C=1..128

`param_overrides` pins `1,2,4,8,16,32,64,128`; `max_batch_size` moves 32 → 128, because a batch cap below the widest measured rung makes that rung serial by construction. That serve-pin change **invalidates every committed concurrency-sweep record** — correct, they measured a different instrument.

The rungs were declared in three hand-maintained lists (`GATE_THRESHOLD_PARAMS`, the floor `ParamSpec`s, `configure`'s `per_c`). A rung in one and not another is silent, because `apply_threshold_params` skips a metric with no bound and `sweep_verdict` only judges the rungs `per_c` names. All three now derive from one `RUNGS` table.

## DFlash2 becomes the eleventh required gate

Nothing gated speculation. All ten previous required gates serve a no-drafter recipe — which is exactly why #812 could only rule DFlash2 out by reading source. A second gate **id** rather than a BENCH.toml variant, because `record_is_required_subject` refuses a non-default checkpoint for the required verdict.

Recipe companion: **atlas-recipes#173**.

## ⏳ In flight as this opens

Calibration is running on spark-256a. Two things land before this leaves draft:

1. **Measured floors** for both gates. The DFlash2 entry is deliberately `status = "unmeasured"` — `baseline_for` drops it rather than admitting empty bounds, so three tests fail by design until the numbers exist. Measure-then-declare.
2. **A resolution for the DFlash2 memory wall**, measured today and reported below.

## ⚠ Finding: DFlash2 cannot serve a wide batch on GB10, and it is not a config mistake

Verified by measurement, not inference:

- The plain gate serves **bs=128 / ctx=4096 / util 0.85** fine — reserve 37.5 GB.
- The DFlash2 serve at bs=128 needs a **68.7 GB** reserve against 46.5 GB available after weights + drafter, and fails with `No memory left for KV cache`.
- The dominant term is the SSM verify pool: `ssm_pool=63264 MB (128 max_batch blobs + 32 MTP-covered slots × 3 verify blobs)`. DFlash's blobs are ~2.5× MTP's at the same batch, because the pools are γ-sized (uniform K = γ+1 on every slot) and γ=8.
- The slot count has a **hard floor of 32** (`mtp_state_slots_with` → `.max(VERIFY_WY_TABLE_SEQS)`), so `ATLAS_MTP_MAX_SEQS` cannot shrink it — measured: setting it to 8 changed nothing.
- The lever that lets MTP reach 128 — the f16-sized h pool — is **refused for DFlash by design**: `gated_delta_rule_wy17` has no FP16 h-state twin and strides its intermediates in FP32 elements, so an FP16 pool there is fluent garbage rather than an error (`cli/validate.rs`).

Measured reserve vs batch: bs=48 → 56.5 GB, bs=64 → 59.0 GB, bs=96 → 63.8 GB, bs=128 → 68.7 GB. The ~49 GB intercept is the 32-slot floor, and it alone exceeds the 46.5 GB available.

**So the DFlash2 gate's ladder will top out well below 128, and the honest instrument is the widest batch it can actually serve.** The final rung list and the measured wall go in the BENCH.toml note. I would rather ship a DFlash2 line that stops where the engine stops, and say why on the chart, than pin a rung that measures the batch cap.

## The site draws both lines

The two gates share one panel — same fixture, same rungs, same checkpoint, differing only in whether the engine speculates. `gate-variants.js` holds that policy as a pure tested module, and prevents three specific defects: two variants joined into one polyline (which reads as a regression and recovery at every alternation), a whole variant rendering faded because "latest" was the last element overall, and the DFlash2 gate being reported as hidden-by-filter on every render.

Colour still follows the model; the dash carries the variant, the idiom `ConcurrencyLadder` already uses for `vllm-nospec`. On the published Atlas-vs-vLLM ladder DFlash2 is a `role: 'variant'` — drawn, and deliberately outside `rows`/`ratio_vs_best`/`wins`/`summary`, so the headline claim keeps meaning "Atlas against the best vLLM at each rung".

## The provenance defect #812 confirmed

The gate serves in-process, so the scheduler reads `ATLAS_PREFILL_CODISPATCH*` from the inherited environment; nothing pinned or recorded them, so two records could share a tree, a recipe and a full set of serve overrides while executing different admission behaviour. `GateRecord` now carries `perf_env` with the **resolved** values. Disclosure only — demanding a match would invalidate every earlier record for a value none could carry. `perf_env_defaults_match_the_scheduler` reads the scheduler's own source and fails if a default moves; verified by sabotage.

## Verified so far

`cargo check --workspace --all-targets` clean · `cargo clippy --workspace --all-targets` clean · `cargo fmt --check` clean · `cargo test -p atlas-plugin` 874 pass / 3 fail, all three the intended unmeasured-entry bootstrap · `bun test src/lib` 522 pass, 8 new · `AI_REPOSITORY.md` corrected (it said "five required gates" and listed `concurrency-sweep` as not required).

https://claude.ai/code/session_01K2yAekCq5TLx85ho4bxBTi


tbraun96:
## Heads-up: `main` moved, one check was deliberately bypassed — and this PR's gate has a blocker

`main` is now `a3686670d2` — the squashed stack **#819** (#813 + #814 + #815 + #816). Those four are closed as merged-via-stack. **#818 was deliberately left out**, and this is why.

### The bypass on #819

#819 merged with **`PR Benchmark Certifications`** red, at the repository owner's explicit direction. All 40 other checks passed. Two protections were briefly relaxed and both **restored and verified**:

| setting | during merge | now |
|---|---|---|
| `enforce_admins` | disabled | ✅ `enabled=true` |
| ruleset `main merge queue` (id 20739400) | disabled | ✅ `enforcement=active` |

**No force push, no history rewrite** — `b474ee0f20` is still an ancestor and exactly one commit was added.

### Why #818 could not come along

#818 adds `concurrency-sweep-dflash2` to the **required** gate set while its only BENCH.toml entry is `status = "unmeasured"`. Three of the repo's own guard tests refuse that, and #818's own note agrees: *"THE FIRST RUN CANNOT BE A GATED ONE."* On the five-PR stack those were the **only** 3 failures — `cargo test -p atlas-plugin` gave **876 pass / 3 fail** with #818 and **874 pass / 0 fail** without it.

### A calibration run was attempted, and it found an engine bug

3 reps, C=1..16, DFlash2 armed at `bs=16` (`conv k=2 group=16, selector rank=256 top_k=16`). It could not produce a usable baseline. Captured on the wire with a logging proxy — identical prompt, `max_tokens=320`, `temperature=0.0`, `seed=0`, `reasoning_effort=none`:

| completion_tokens | finish_reason | accepted draft tokens |
|---|---|---|
| 320 | `length` | 251 |
| 257 | `length` ⚠ | 206 |
| 305 | `length` ⚠ | 252 |
| 305 | `length` ⚠ | 254 |
| 229 | `length` ⚠ | 188 |

`finish_reason: "length"` with fewer tokens than requested is self-contradictory, and **completion length tracks accepted draft tokens** — the length budget is being spent on *drafted* rather than *emitted* tokens, so rejected drafts silently consume the caller's `max_tokens`. `rejected_prediction_tokens` reports `0` in every response. **The plain engine, same box and prompts, returned 320 on every rung of all 3 reps, 0 vacuous cells.**

Ruled out first, each reproducing clean: prefix caching (the plain leg had it too), request repetition (4 idle repeats all 320), and the 16-way→single batch transition. What found it was capturing the HTTP exchange rather than reasoning about it.

This is user-facing, not just a benchmark artifact: any caller with DFlash2 enabled gets silently truncated completions, mislabelled, non-deterministically.

Floors cut from that data would encode the truncation as expected behaviour, and `vacuous_cells max = 0` would fail the very reps that defined it.

### The obvious workaround was tried and reverted

Moving the gate to `PROMOTION_CANDIDATES` takes the failures from **3 to 8**. Five are mechanical membership updates; two encode a real invariant —
`the_flat_concurrency_driver_invalidates_only_its_own_gate` and
`the_promoted_gates_invalidate_where_they_used_to_accrue_debt` assert that a change to the shared driver must invalidate **both** gates, or "the speculative ladder [is] certified by a record it never measured". Making a *candidate* participate in driver invalidation is a design decision on the coverage model, so it was left to this PR's author. Reverted; nothing here was pushed.

### Two corrections for this PR's own text

* The BENCH.toml note asks for *"≥3 reps of the full C=1..128 ladder"*, but this gate's config is `concurrencies = "1,2,4,8,16"` with `max_batch_size = 16`. Only the plain sweep reaches 128. The orchestrator's own comment explains why: the verify pool's slot count is `max_batch_size.min(max(cap, 32))`, so at `bs>=32` it pins at 32 slots and cannot shrink. **C=16 is a hardware ceiling, not a preference** — the note will mislead whoever re-cuts these floors.
* The note predicts the two curves *"converge at the wide rungs"*. Measured, DFlash2 **inverts**: 23.4 → 43.6 → **50.3 (peak at C=4)** → 42.6 → 32.8, while the plain sweep rises to 73.5 at C=16.

**#818's gate has already justified itself** — its stated purpose is that nothing else gates speculation, and its own calibration run caught a speculative-path correctness bug that no committed record would otherwise have spoken for.

PR #821, opened 2026-08-30
fix(qsa,ple): free per-sequence device state on sequence teardown (~739 MB/request leak)

rsafier:
`QsaSeqState` and `PleSeqState` hold bare `DevicePtr`s. Dropping the layer
state reclaimed the host struct and leaked the device buffers behind it —
nothing ever called `gpu.free`.

## The measurement

On qwen4_exp at 200K ctx, driving requests with growing context:

```
live >=32MB allocations:  196 -> 208 -> 220 -> 232 -> 244
host used (gx10):        82.1 -> 83.8 -> 85.4 -> 87.6 -> 90.7 GB
host used (dgx-00):      85.4 -> 87.0 -> 88.3 -> 90.3 -> 92.9 GB
```

Exactly **+12 allocations per request, linear, no plateau**. Every one of them
48.8 MB — `max_tokens * hd * 2` = 200000 x 128 x 2, so sized by config rather
than by prompt length, which is why it did not vary with input size. qwen4_exp
has exactly 12 full-attention layers, one `raw_keys` each, plus the 12.8 MB
`block_keys` beside it: **~739 MB per request that never comes back**.

Both EP ranks climb identically, since both run all 12 attention layers.

`PleSeqState::conv` has the same defect on the 36 SSM layers. At ~147 KB it
sits below the 32 MB allocation-trace threshold, which is why only QSA was
ever visible.

## Why it was invisible

On GB10 unified memory, CUDA allocations appear in no `/proc` counter, are
absent from process RSS, and `nvidia-smi` reports memory as N/A. The symptom
was a host climbing to 118 GB of 121 with page cache squeezed to nothing,
while the only candidate process showed a 16 GB RSS. Attributing it needs the
carveout computed by subtraction:

```
MemTotal - MemFree - Cached - AnonPages - Slab - PageTables - KernelStack
```

which read 97.5 GB against a ~67 GB util pledge.

## The fix

`TransformerLayer::release_state`, the counterpart to the existing
`alloc_state`, defaulted to a no-op so only the two layers that own unpooled
per-sequence device memory implement it. (`SsmLayerState`'s h/conv is pooled
and released by slot — freeing it here would be a double free.)

`free_sequence_dispatch` — the terminal chokepoint every finished sequence
routes through, on every path: stop/EOS/length, error/abort, prefill-error
free, swap-out spill — now calls it per layer. Errors are logged rather than
propagated: bailing there would strand the KV blocks and prefix-cache refs
released further down, trading a leak for a worse one.

Note the buffers are attached lazily on first use, not by `alloc_state`, so
what `release_state` frees is whatever the sequence actually grew. `take()`
makes it idempotent.

## Ledger changes

The allocator ledger was a `HashSet<u64>` of bare pointers, so
`sweep_unreleased` could report how MANY allocations had no owner but never
how much memory they held — its own doc comment conceded this. It is now a
`HashMap<u64, usize>`:

- `sweep_unreleased` logs total bytes leaked, not just a count.
- `free` traces what it releases, so alloc/free trails can be diffed by
  pointer — the query that names what accumulates.
- Both traces carry a running live total. Without it, a buffer allocated and
  freed every step is indistinguishable in the log from one never released.

The existing `>=32MB` alloc trace already existed for exactly this failure
("a runtime alloc outside the util pledge is how the box ends up in swap") but
is `RUST_LOG=spark_runtime=debug`, so it is dark under the usual `info`.

## Testing

- `cargo build --release --all-targets` and `cargo clippy --release
  --all-targets` clean (`ATLAS_TARGET_MODEL=*`).
- 632 tests pass. One video-decode test fails only under parallel execution
  (`/tmp` contention on the test box); it passes in isolation and is unrelated.
- Verified live on the EP=2 qwen4_exp config (200K ctx, util 0.55): the
  ledger's live total is flat across requests where it previously climbed
  ~585 MB each, and no `release_state` errors are logged.

Introduced in 0a56d8f8b (the Flash-Next integration stack); this is a
correctness fix independent of #820, which touches the same file for
performance.

### DIFF
--- a/crates/spark-model/src/layer/transformer_layer.rs
+++ b/crates/spark-model/src/layer/transformer_layer.rs
@@ -643,4 +643,28 @@ pub trait TransformerLayer: Send + Sync {
     /// - `EmptyLayerState` for pure attention layers
     /// - `SsmLayerState` for SSM/recurrent layers
     fn alloc_state(&self, gpu: &dyn GpuBackend) -> Result<Box<dyn LayerState>>;
+
+    /// Release the per-sequence state `alloc_state` produced, plus anything
+    /// the layer attached to it lazily afterwards.
+    ///
+    /// Called once per sequence from the teardown chokepoint
+    /// (`free_sequence_dispatch`). The default no-op is correct for layers
+    /// whose state owns no device memory (`EmptyLayerState`) and for state
+    /// that comes from a pool reclaimed by slot (`SsmLayerState`'s h/conv,
+    /// released via `ssm_pool.release_slot`).
+    ///
+    /// It exists because `LayerState` implementors hold BARE `DevicePtr`s:
+    /// dropping the box reclaims the host struct and leaks the device buffer.
+    /// The QSA indexer carry (~739 MB per request at 200K context across the
+    /// 12 full-attention layers) and the PLE conv carry both leaked this way.
+    /// On unified memory such a leak is invisible to RSS and reported as N/A
+    /// by `nvidia-smi`, so it surfaces only as the host exhausting RAM with no
+    /// process to blame.
+    ///
+    /// MUST be idempotent — teardown can run after a partial failure. Callers
+    /// log errors and continue rather than aborting: a sequence that cannot
+    /// free its state is still finished, and bailing would strand the rest.
+    fn release_state(&self, _state: &mut dyn LayerState, _gpu: &dyn GpuBackend) -> Result<()> {
+        Ok(())
+    }
 }
--- a/crates/spark-model/src/layers/ple/layer.rs
+++ b/crates/spark-model/src/layers/ple/layer.rs
@@ -145,6 +145,27 @@ impl PleLayer {
         })
     }
 
+    /// Release one sequence's PLE carry.
+    ///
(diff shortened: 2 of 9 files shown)


tbraun96:
Closing: the fix is on `main`. #965 (`7dd531f59b`, the GLM-5.3-Flash integration) cherry-picked the model half of this PR from its head `d68696b1` — `TransformerLayer::release_state` beside `alloc_state`, the per-layer loop in the `free_sequence` teardown, and the QSA/PLE implementations. On `main` `9e597531`: `release_seq_state` at `crates/spark-model/src/layers/ple/aux_state.rs:117`; all 5 functions this PR adds are present.

The ledger half (`live_allocs: HashSet<u64>` → `HashMap<u64, usize>`) was deliberately not taken, as #965's message records: #818 had already landed the richer `AllocRecord{bytes, site}` ledger (`crates/spark-runtime/src/cuda_backend/alloc_ledger.rs` on `main`). That is the whole of this branch's conflict surface, and it is why it still shows as conflicting.

Thanks for finding the ~739 MB/request leak — the teardown path is what ships. No branch deleted.

PR #826, opened 2026-08-31
blog: seven tenets powering Atlas Inference (article 0)

DrRainbows:
The first post, in the format 825 just added. Three files, all under `blog/`.

## The body is untouched

It is the refinement Thomas sent, byte for byte. I diffed it after staging rather than trusting myself: the prose in this PR is character-identical to the source file, em dashes and all. There was no grammar or spelling work to do, so none was done.

That includes the things a copy pass would have reached for and shouldn't have — the special note about em dashes, "We require AI-generated PRs", the 40+ minute observation, the closing exclamation. All his.

## What is new is the frontmatter, and it is the site talking

The article never stops to define the product, and it shouldn't have to — it's an essay, and a definition sentence in the middle of it would read as marketing. But `dek` is what a search result and an answer engine quote, so that is where the definition goes:

> Atlas Inference is a free and open source LLM inference engine written from scratch in Rust. These are the seven philosophical tenets we started it on, and why we left the Python vLLM stack to do it.

Nobody's voice, no claim that isn't already on the product site. **Thomas should read that line specifically**, because it ships under his byline even though he didn't write it.

`categories: [engineering, design]` and `author: thomas-braun` are the vocabulary already in `content.js`. Keywords are the queries this post can honestly serve — not "fastest", not a comparison, because this post cites no benchmark and pretending otherwise fails on the first person who checks.

## Two images

Both webp: the body allowlist refuses raster jpg and png, which is the correct call and cost a conversion.

- One still above the opening paragraph — a wafer against rusted iron. It carries the lede in one frame, and the pun is there if you want it and invisible if you don't.
- A 1200×630 social card, so the first time this URL is pasted into Discord or X it isn't a grey box.

The post is laid out as prose deliberately. Eight more stills exist and were left out: it's an essay, and figures every two paragraphs would make it look like a walkthrough.

## Three things that are decisions, not defects

**1. The title renders twice.** The theme emits an `<h1>` from `title`, and the body opens with the same sentence in bold, which is how the source is written. On the built page they sit directly on top of each other. Deleting the bold line is a change to his file, so I didn't make it — but it will read as a mistake, and it's one line.

**2. The stills are AI-generated, and paragraph one promises to say so.** "If something is AI-generated, we will let you know." The words are entirely his; the two images are not. As it stands the post is silent about that, which makes its own first paragraph untrue about the page it's printed on. One credit line under the image settles it. His call on the wording — I'm not putting sentences in his mouth to fix a problem I introduced by adding a picture.

**3. "Atlas Cybernetics" appears once**, in the monorepo passage. The public name everywhere else is Atlas Inference. Left exactly as written, because it's his string — but it's the one word on the page that will file this under a different company.

There are also no `##` headings in the body, so the post gets no heading rail and no anchors. The four section markers are italic in the source and stayed italic. Promoting them would help the post get found and would change how his article is structured, so that's his call too, not a thing to slip into a PR.

## Verified before opening this

Locally, against this branch:

- `bun scripts/check-frontmatter.mjs` — valid.
- `bun test src/lib` — 92 pass, 0 fail.
- `bun run build` — clean. The post prerenders at `/posts/seven-tenets-powering-atlas-inference`, 40 KB, 7 minute read, and lands in the index, RSS, sitemap and the generated `llms.txt`. Canonical, `og:image` and `og:description` all resolve.

Web-only PR, so it should take the fast path rather than the release matrix.


DrRainbows:

DrRainbows:
**Update � the AI-image tag is in, and one flag above is retracted.**

`_Image generated with AI._` now sits under the still. Paragraph one promises exactly that disclosure, and the words are Thomas's while the picture isn't, so the post was contradicting itself. The renderer ignores the markdown title slug and emits a bare ``, so there's no figcaption to use � a short italic line is the whole change.

**Retraction:** I said above that the title "renders twice, stacked." That was wrong � I counted tags instead of checking the render order. The `<h1>` is in the masthead, and Thomas's bold line lands after the opening paragraph, where it reads as the header introducing the tenets. Nothing to fix.

Still his call, unchanged: **"Atlas Cybernetics"** appears once, and there are no `##` in the body so the post gets no heading rail or anchors.

Thomas's prose is byte-identical to the refinement he sent, verified against `Downloads/atlas-article-0-AD-refinement-md.txt` rather than against an intermediate copy. Exactly two blocks were added to his file: the image, and the credit line.

PR #828, opened 2026-08-31
dflash: serve the record shape by default (66.6 tok/s, zero flags)

rrstesiak:
A default `spark serve --dflash` did not serve the configuration that set
the records. It served a slower shape of the same engine, and every number
we have published came from a launch line that knew the incantation.

This makes the defaults **The measured configuration.**

## What changes

**Gamma.** The drafter head's default gamma was the checkpoint's trained
block size. On Qwen3.8-27B DFlash2 that is 8, which runs 7 drafts, one
short. These drafters chain past their trained block: at gamma 10 the
verify holds full-block 9/9 accepts. Default becomes `block + 2`, clamped
to `MAX_F16_TWIN_DFLASH_GAMMA` so a block-16-class drafter lands on 15
rather than gamma 18 / K=19, which no wyN kernel serves. An explicit
`--dflash-gamma` still wins everywhere.

**Three levers graduate from opt-in to default-on**, using the same
`on_unless_zero` pattern `unified_ctx` already uses:
`ATLAS_DFLASH_MASKED_VERIFY`, `ATLAS_DFLASH_SEAM_SERIAL`,
`ATLAS_DFLASH_SPEC_THINK`. `=0` restores each legacy path for A/B, and
`=1` stays a harmless no-op, so every existing recipe remains correct.
`dflash_adaptive` deliberately stays opt-in: the record runs it off.

## One SSOT, because three spellings cost a run

The block-size-to-gamma rule existed independently in the head's
`from_weights`, the serve preflight's pool and reserve sizing, and
`factory::build`'s `config.dflash_gamma`. Raising only the head left the
pools sizing for the old value, and the first verify step died
mid-graph-capture:

```
SSM MTP intermediate buffers not allocated (need K-1 h + K conv;
h_state_intermediates.len()=8, conv_state_intermediates.len()=9,
num_tokens=10)
```

All four sites now resolve through `qwen3_ssm::default_dflash_gamma()`.

## Measured

Qwen3.8-27B NVFP4 + DFlash2, single DGX Spark, single stream, clocks
unpinned. Launched with **zero `ATLAS_DFLASH_*` variables and no gamma
flag**, which is the point: this is the out-of-the-box path.

| bench | runs | median | tokens | output sha |
|---|---|---|---|---|
| MinHeap (code) | 66.6 / 66.6 / 66.5 | **66.6** | 839 | `e23c9083` x3 |
| Volvo (prose)  | 24.7 / 24.7 / 24.7 | **24.7** | 300 | `90c1fca8` x3 |

Against the 2026-08-29 records on the same box: MinHeap 63.0 median from
66.6 / 62.1 / 63.0, Volvo 23.8 from 23.9 / 23.3 / 23.8.

From a CUDA/C centric tuning background, the repeatability is the result
worth reading here rather than the throughput.
The record sets cycled three different outputs by run position (`e23c9083` / `b6e29d25` / `fc44fc22`).

These sets produced one output each, byte for byte, three times.
Token counts are identical to the record class, so the comparison is
length-matched rather than slope-corrected. 

I am not claiming a cause for the variance collapse; main has moved a long way since `244d42f6` and
`dflash_batch_verify` landed default-on in #818, both of which touch this
path.

### DIFF
--- a/crates/spark-model/src/factory/build.rs
+++ b/crates/spark-model/src/factory/build.rs
@@ -148,10 +148,18 @@ pub fn build_model(
         // a DFlash2 checkpoint states its trained block size inside
         // `dflash_config`, and the top-level field's default of 16 must not
         // shadow it. --dflash-gamma still wins over both.
-        config.dflash_gamma = Some(
-            args.gamma
-                .unwrap_or(args.drafter_config.effective_block_size()),
-        );
+        //
+        // MUST go through `default_dflash_gamma`, this value sizes the SSM
+        // pools (uniform K = γ+1), and the HEAD resolves its own γ through
+        // the same helper. When this said `block_size` while the head said
+        // block+2, the pools came up one verify row short and the first
+        // DFlash step died with "SSM MTP intermediate buffers not allocated
+        // (h=8, conv=9, num_tokens=10)" mid-graph-capture.
+        config.dflash_gamma = Some(args.gamma.unwrap_or_else(|| {
+            crate::layers::qwen3_ssm::default_dflash_gamma(
+                args.drafter_config.effective_block_size(),
+            )
+        }));
         tracing::info!(
             "DFlash: target layer capture indices = {:?} (drafter target_layer_ids, \
              used directly), γ = {:?}",
--- a/crates/spark-model/src/layers/dflash_head.rs
+++ b/crates/spark-model/src/layers/dflash_head.rs
@@ -318,7 +318,9 @@ pub struct DflashProposerState {
     /// EAGLE-fix one-shot: when set, the next `propose()` skips its internal
     /// decode-append because the verify step (K=2 accept) already appended
     /// row 0 + row 1 in EAGLE order before calling propose. Consumed (reset to
-    /// false) by propose. Only set under ATLAS_DFLASH_EAGLE_FIX=1.
+    /// false) by propose. Set on the EAGLE-fix path, which is DEFAULT-ON
+    /// (`ATLAS_DFLASH_EAGLE_FIX=0` is the kill switch, not `=1` the opt-in , 
+    /// see verify_k2_step.rs and verify_dflash_step.rs, both `!= Some("0")`).
     pub skip_next_decode_append: bool,
     /// Allocation cap for `ctx_hidden_acc` (in slot count). Mirrors the
     /// `max_seq_len` build arg so we can clamp without re-fetching it.
--- a/crates/spark-model/src/layers/dflash_head/from_weights.rs
+++ b/crates/spark-model/src/layers/dflash_head/from_weights.rs
@@ -71,10 +71,23 @@ impl BlockDiffusionDraftHead {
         let head_dim = weights.config.head_dim;
         let vocab_size = weights.config.vocab_size;
         // The HEAD's gamma is the SSOT the serve layer reads back, so the
-        // drafter's own trained block size must win here — `block_size` alone
-        // is the top-level field, whose serde default of 16 silently shadows
-        // a DFlash2 checkpoint that states 8 in `dflash_config`.
-        let gamma_val = gamma.unwrap_or(weights.config.effective_block_size());
+        // drafter's checkpoint must drive the default here — `block_size`
+        // alone is the top-level field, whose serde default of 16 silently
+        // shadows a DFlash2 checkpoint that states 8 in `dflash_config`.
+        //
+        // Default = trained block size + 2, not the block size itself: these
+        // drafters chain past their trained block (measured on Qwen3.8-27B
+        // DFlash2, block 8 — gamma 8 is 7 drafts, one short; gamma 10 holds
+        // full-block 9/9 accepts and is the fastest measured serve, 63.0 vs
+        // 56.2 tok/s on GB10). Clamped to MAX_F16_TWIN_DFLASH_GAMMA so a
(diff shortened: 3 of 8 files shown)


tbraun96:
recheck

tbraun96:
## Certification status, and three blockers — two of which need @rrstesiak

Picking this up to certify and merge. Current state of the four red checks:

| check | state | who can clear it |
|---|---|---|
| `cargo fmt --check` | ❌ | fix in hand, **cannot push** (see below) |
| `CLAAssistant` | ❌ | @rrstesiak — **two-account problem**, detail below |
| `PR Benchmark Certifications` | ❌ | needs all **11** gates re-run |
| `PR benchmark gate` | ❌ | transitional **alias** of the above — one cause, two reds |

### 1. `cargo fmt` — trivial, but I can't land it

Three spots: trailing whitespace on `dflash_head.rs:319` and `levers.rs:169`, and an import rewrap in `qwen3_ssm/mod.rs:407`. I have the commit and verified it is semantically inert — the `gdn_flags` import list is the **same 10 symbols** before and after, and the other two files are byte-identical modulo trailing whitespace.

I can't push it: this branch lives on `rrstesiak/atlas` and my token reports `push=false` there, despite `maintainerCanModify=true`. Either push `cargo fmt --all` yourself, or say the word and I'll land the content from a base-repo branch.

### 2. `CLAAssistant` — you are signed, under a *different account*

This one is worth spelling out because the error message is misleading. The CLA action reports *"Committers of Pull Request number 828 have to sign the CLA"*, and you have already signed. Both are true:

| account | id | status |
|---|---|---|
| **`rrstesiak`** | 3093273 | ✅ signed (pr#78) — owns this fork, is the PR author |
| **`rstesiak-cpu`** | 262220436 | ❌ not signed — but this is who `rstesiak@gmail.com` maps to |

Your commit `6b09d95eb2` is authored `Ronald R. Stesiak <rstesiak@gmail.com>`, and GitHub attributes that address to `rstesiak-cpu`. The action checks **committers**, so it wants the signature from the account it can see. `recheck` will not fix it — I tried.

Three ways out, all yours or @tbraun96's to pick: sign from `rstesiak-cpu`; re-author the commit under the email linked to `rrstesiak`; or add `rstesiak-cpu` to `CLA_ALLOWLIST` in `.github/workflows/cla.yml`. I deliberately did **not** take the third — a CLA allowlist is a legal decision, not a CI fix.

### 3. The certification is the full campaign, and ordering matters

This PR touches `factory/build.rs` and `dflash_head.rs`, which invalidates **all 11 required gates** — not a subset:

```
agentic-webserver, vision-fidelity, video-fidelity, ttft-warm-gate, ttft-cold-gate,
bfcl-subset, bfcl-subset-echolp, ssm-state-poisoning-gate, decode-floor,
concurrency-sweep, concurrency-sweep-dflash2
```

All 11 currently read `NONE — latest record is for f0f6e48845 … invalidated`.

**The fmt fix must land before the campaign starts, not after.** Gate records are commit-pinned, and `dflash_head.rs`/`levers.rs` are themselves in the invalidating set — so certifying today's head and *then* pushing the whitespace fix would invalidate every record I just spent the GPU hours on. Ordering is: fmt → CLA → certify → merge.

### Benchmarking guidance recorded, so nobody misreads the number

Per @rrstesiak, and this is the part most likely to be misread by an automated run:

> The 66.6 is **C=1**. Single stream, defaults, no flags. That lane is finished and it is 4.1x the serial floor on the same box.

The 08-30 sweep against plain: **C=1 +29.0%, C=2 +38.5%, C=4 +3.4%, C=8 −7.1%, C=16 −29.1%.** DFlash2 wins through C=2, is par at C=4, negative from C=8. Richard's cross-sequence batching bought the low end (C=2 +58%, C=4 2x, C=8 124→150 aggregate).

The mechanism above C=4 is known and claimed: γ is still tuned for C=1, and the GDN recurrent body falls back to one serialized pass per sequence because there is no fused WY kernel at K=γ batch width. That batched K=γ GDN verify kernel is several days of real kernel work, not a tuning pass.

**So `concurrency-sweep-dflash2` at C=8/C=16 is expected to read negative against plain, and that is a known open item rather than a regression introduced by this PR.** Whoever reads the campaign output should not treat it as a reason to block these defaults — but it does mean the gate's own thresholds need to be honest about it, or the gate will block every future DFlash PR for a gap that is already documented.

### One caution from tonight, on `concurrency-sweep-dflash2` specifically

While calibrating that gate for #818 earlier today, the DFlash2 leg returned **short completions mislabelled as length-capped**: identical prompt, `max_tokens=320`, `temperature=0.0`, `seed=0`, and completions of 229/257/305/305/320 tokens, *every one* reporting `finish_reason: "length"`. Completion length tracked accepted draft tokens (188→229, 252→305, 251→320), and `rejected_prediction_tokens` read `0` throughout. The plain engine returned 320 on every rung of all 3 reps.

That was measured on the **pre-#828 default shape** — i.e. with `masked_verify`, `seam_serial` and `spec_think` all *off*, which is exactly what this PR fixes. So it is worth checking whether the record shape this PR makes default also clears the truncation. If it does, that is a second argument for these defaults. If it does not, `concurrency-sweep-dflash2` will not produce a usable record and the campaign stalls there.

rrstesiak:
Superseded by #831, same commits with the correct CLA-signed authorship.

PR #831, opened 2026-08-31
dflash: serve the record shape by default (66.6 tok/s, zero flags)

tbraun96:
Same tree as #828, which cannot go green on its own branch. Two mechanical fixes, **zero content change**.

Authored by @rrstesiak — commits `5af9088` and `c4d7d62` are his, unmodified in content.

## Why this branch exists

**1. CLA — a two-account problem, not an unsigned contributor.**
`CLAAssistant` on #828 reported *"Committers of Pull Request number 828 have to sign the CLA"* while @rrstesiak **had already signed**. Both were true:

| account | id | status |
|---|---|---|
| `rrstesiak` | 3093273 | ✅ signed (pr#78) — fork owner, PR author |
| `rstesiak-cpu` | 262220436 | ❌ unsigned — but `rstesiak@gmail.com` maps here |

The commits were authored `Ronald R. Stesiak <rstesiak@gmail.com>`, which GitHub attributes to `rstesiak-cpu`; the action checks **committers**, so it asked the wrong account for a signature. `recheck` does not fix it.

At the repository owner's direction the two commits were amended to `rrstesiak <rrstesiak@hotmail.com>` — the address behind 138 commits of his history in this repo. **The tree is byte-identical across the amend** (`71c27e7b` before and after); only authorship metadata moved. GitHub now resolves both to `rrstesiak`.

**2. `cargo fmt --check`.** Trailing whitespace on `dflash_head.rs:319` and `levers.rs:169`, plus an import rewrap in `qwen3_ssm/mod.rs:407`. Verified inert: the `gdn_flags` import list is the **same 10 symbols** before and after, and the other two files are identical modulo trailing whitespace.

Neither fix could be pushed to #828 — its branch is on `rrstesiak/atlas`, where this token reports `push=false` despite `maintainerCanModify=true`.

## What the PR actually does (unchanged from #828)

Graduates three DFlash levers from opt-in to default-on — `masked_verify`, `seam_serial`, `spec_think` — the levers behind the record serve, and adds `default_dflash_gamma()` as a single source of truth for the served γ (`block + 2`, clamped to `MAX_F16_TWIN_DFLASH_GAMMA`). `dflash_adaptive` deliberately does **not** graduate. Each lever keeps a `=0` kill switch for A/B.

A default `spark serve --dflash` previously shipped none of them, so out-of-the-box DFlash ran the slow shape of its own engine.

## Certification

This touches `factory/build.rs` and `dflash_head.rs`, invalidating **all 11 required gates** — all currently read `NONE — latest record is for f0f6e48845 … invalidated`.

Ordering matters and is why the fmt fix is in this PR rather than after it: gate records are commit-pinned, and `dflash_head.rs`/`levers.rs` are themselves in the invalidating set, so certifying and *then* pushing whitespace would invalidate every record the campaign produced.

### Reading the numbers, per @rrstesiak

> The 66.6 is **C=1**. Single stream, defaults, no flags. That lane is finished and it is 4.1x the serial floor on the same box.

08-30 sweep vs plain: **C=1 +29.0%, C=2 +38.5%, C=4 +3.4%, C=8 −7.1%, C=16 −29.1%.** DFlash2 wins through C=2, par at C=4, negative from C=8. The mechanism is known and claimed: γ is tuned for C=1, and the GDN recurrent body falls back to one serialized pass per sequence because there is no fused WY kernel at K=γ batch width.

**So `concurrency-sweep-dflash2` reading negative at C=8/C=16 is a documented open item, not a regression from this PR.** Bench C=1 and the number is a record; bench C=16 today and you measure the gap that is already claimed.

Closes #828.

https://claude.ai/code/session_011jz2X6Thestce5evZL8FqA


rrstesiak:
Not asking to change the merge method, squash is fine. 
Just add a Co-authored-by: Ronald R. Stesiak <rrstesiak@hotmail.com> trailer to the squash message. GitHub credits co-authors on the contributor graph, so we keep clean linear history and accurate attribution, and it costs one line.

tbraun96:
## Merged on certified gates — and the campaign found a real bug in this PR

All 11 required gates were earned on the GPU, no bypass. The certification passed **twice independently**: on the PR head, and again on the merge queue's own temporary merge commit.

### The bug this PR had, and how it was isolated

Graduating `dflash_spec_think` to default-on was **not** safe. Despite the `dflash_` prefix, `mtp_gate::spec_dispatch_eligible` reads it on **both** lanes:

```rust
if inside_thinking && !spec_think { return false; }
```

Defaulting it on removed the guard that keeps speculation out of `<think>` for plain MTP, where batch-K verify is not byte-lossless at T=0. A same-night control on `origin/main` isolated it:

| build | agentic-webserver | bfcl-subset-echolp |
|---|---|---|
| `origin/main` | 10/10 · 10/10 | 86.55 / 86.95 |
| **this PR (before fix)** | **9/10 · 7/10**, deterministic across 3 runs | **85.66 / 86.06** — FAIL |
| **this PR + fix** | **10/10 · 10/10** | **86.55 / 86.95** |

`echolp` did not merely clear its floor — it returned to the **top of its seven-record history** (the two best prior runs also scored 86.55/86.95).

The guard comment in `spec_dispatch_eligible` already recorded a 2026-08-16 bisect with this exact signature: *"main+this-hunk fails, main without it passes 10/10."* This PR walked back into it.

`masked_verify` and `seam_serial` **stay graduated** — both consumers are additionally gated on `dflash_verify_raw_argmax` (= `args.dflash`), so they cannot reach a no-drafter serve. Only `spec_think` escapes the lane.

### Why `cargo test --workspace` was green the whole time

`every_opt_in_lever_ships_off` asserts `SchedLevers::defaults()` — a hand-written struct literal. `spark serve` calls `from_env()`. This PR changed only `from_env()`, so the literal still said `false` and the test passed while production resolved the opposite. A guard pointing at the wrong constructor reports safety it cannot deliver.

Added `spec_think_is_off_in_the_resolver_the_server_actually_uses`, which asserts the resolver the server actually uses. Negative control: reintroducing the one-line change reds the new test and leaves the old one passing.

### Two things that were NOT this PR's fault

* **`concurrency-sweep` C=2** failed once at 22.8, then passed at 28.0 on re-run at the same commit. A control on `origin/main` scored **22.9** — main fails this cell too. The rung is bimodal with the floor inside the gap; filed as #835.
* **`video-fidelity`** first recorded a FAIL because I passed `--serve-override video_allow_ffmpeg=true` from a stale handoff doc. Both recipes now pin that flag themselves and the baseline pins no overrides, so supplying one makes the record non-comparable by design. Re-run without it: 13/13 legs.

### Final gate numbers (`6a8c0b8de3`)

| gate | result |
|---|---|
| agentic-webserver | 10/10 webserver_ok · 10/10 followed_directions |
| bfcl-subset | 84.22 / 84.12 (n=995, golden seed-42 draw) |
| bfcl-subset-echolp | 86.55 / 86.95 (n=1004) |
| concurrency-sweep | 8 cells, 0 vacuous, C128 114.9 tok/s |
| concurrency-sweep-dflash2 | 5 cells, 0 vacuous, C1 21.8 → C16 61.8 |
| decode-floor | 22.7 tok/s vs 20.5 floor |
| ssm-state-poisoning | 12/12 replays byte-identical |
| vision-fidelity / video-fidelity | 14 cells · 13/13 legs, controls held |
| ttft-cold / ttft-warm | median +0.1% / −0.2% |

One note for anyone reading the DFlash2 ladder: it wins the low end and goes negative from C=4 (C1 +19%, C2 +69%, C4 −10%, C8 −9%, C16 −31% vs plain), independently reproducing @rrstesiak's 08-30 numbers. That is the documented open item — no fused WY kernel at K=γ batch width — not a regression from this PR.

PR #833, opened 2026-09-01
blog: DFLASH-2 single-Spark record article (66.6 tok/s on stock defaults)

rrstesiak:
Second post for the blog, per the article commitment from the weekend.

What it reports, all receipt-backed in the RECORDS_LEDGER protocol (median of 3, temp 0, single stream, completion tokens quoted, sha256 per set):

- MinHeap (code): 66.6 / 66.6 / 66.5, median 66.6, one sha across all three runs
- Volvo (prose): 24.7 x3, one sha
- Serial floor (speculation off): 16.23 median, one sha, giving a 4.10x one-variable A/B
- The cross-squash byte-identity receipt (25e95ba0.., identical output 11 days and 342 files apart)

Numbers were measured on the #831 build, so this reads best merged after #831 lands. The serve line in the post is the naked default line: no DFlash env, no gamma flag.

Also adds the ronald-stesiak author entry so my posts carry a byline. No editorial role sought or implied.

draft: false, so it publishes on merge. Can hold as draft: true instead if you want to time the announcement.

### DIFF
--- a/blog/src/lib/content.js
+++ b/blog/src/lib/content.js
@@ -55,6 +55,12 @@ export const authors = {
     initials: 'TB',
     role: 'Founder, Avarok Cybersecurity',
     bio: 'Works on the Atlas inference engine — kernels, scheduling, and the benchmarks that decide whether any of it was worth it.'
+  },
+  'ronald-stesiak': {
+    name: 'Ronald R. Stesiak',
+    initials: 'RS',
+    role: 'Founding Engineer, Atlas',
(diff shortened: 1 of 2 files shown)


tbraun96:
**Carried by stack #875 (`stack/content-and-records`)** — merge-tree returns #875's own tree: both files (`blog/src/lib/posts/dflash2-single-spark-record.md`, `blog/src/lib/content.js`) are already in that stamped stack. This PR closes with landing evidence when #875 merges, not before. If any hunk did not make it across, point at it and I will lift it onto a fresh branch.

PR #834, opened 2026-09-01
EXL3 checkpoint support: native kernels and Qwen MTP verification

rsafier:
## NVIDIA NVFP4 follow-up

The downloaded NVIDIA checkpoint now loads and runs experimental MTP. This follow-up changes **shared quantization detection** so FP8 draft tensors cannot misclassify NVFP4 main experts, wires FP8 PLE gathers, and fixes BF16 GDN output projection replay. Three plain-text greedy responses match serial; forced-tool reasoning still differs. Agentic and performance validation for this checkpoint are outstanding. See the [current handoff](https://github.com/Avarok-Cybersecurity/atlas/blob/2274d01d723c1ff809588cb3eb07a443e8beedd6/bench/qwen4_exp/NVIDIA_MTP_HANDOFF.md) for evidence, scope and next steps; the EXL3 results below concern a different checkpoint.

Adds EXL3 checkpoint ingestion, reconstruction compatibility, packed native kernels and PLE n-gram decoding, plus Qwen3.8-Flash-Next MTP verification. The native path serves packed dense/LM-head/routed-expert weights under the EXL3 feature flags.

For Qwen MTP, wider verification now retains serial sampling/emission semantics and stable expert geometry. The follow-up also fixes concurrency admission and per-request draft-state cleanup. The recommended tested starting point is **two drafts, one active sequence**.

## Changes

- EXL3 format detection and exact half-precision ingest, reconstruction/materialization fallback, and packed n-gram row decoding.
- Native dense, vocabulary-head and routed-expert kernels with Atlas's stable-grid and deterministic-reduction adaptations retained.
- Acceptance-aware K2/K3/K4 sampling, live tool grammar, accepted highway-row selection, private draft chaining, and auxiliary rollback/event ordering.
- Correct four-row FFN dispatch and committed-row context-limit checks; fail fast on active Qwen MTP without batched HC support.
- Fused routing/activation staging and removal of unused draft-prefix CPU snapshots. Diagnostic switches retain reproducible old-path comparisons.
- HC models use per-sequence verification at concurrency two. Draft KV capacity follows active slots; teardown returns KV blocks and QSA buffers. Abandoned proposals rewind, and active MTP no longer creates global shadow state.

Shared surfaces include weight ingestion, CUDA launch glue, model activation, scheduler sampling and generic batched-verify admission.

## Validation

Latest native binary: runtime `ab0bdb746`, tested on one GB10 with the local Qwen3.8-Flash-Next 4.05bpw checkpoint, BF16 KV, 32K configured context and low reasoning effort.

- All three widths match the complete reference choices on 15 short responses and nine repeated 512-token responses. Three snapshot-control responses also match.
- GPU ingress: 48 exact-buffer cases; stable expert-grid: 18 cases with both ingress paths, with a failing old-grid negative control.
- Final C2 profile: 12/12 overlapping responses match isolated controls after excluding assigned tool-call IDs; cancellation and slot reuse pass. Idle GPU allocation returns to the same value between pairs; minimum host availability 20.59 GiB, zero swap-outs.
- Model-card agentic checks explicitly preserve prior thinking: **one and two drafts pass all six checks**. Three drafts exceeded the 1,000-second budget while repairing generated tests and was cancelled. That run is retained as failed/cancelled, not a pass.
- Final-source unit suites: model 775 passed/14 ignored (serial rerun after a transient ffmpeg fixture startup failure); server 2,361 passed/12 ignored. CUDA workspace clippy passes with `-Dwarnings`.

**Preliminary, single-harness performance:** the same 512-token coding response took median 33.286/31.134/32.529 seconds at one/two/three drafts (three repeats each). This supports starting at two drafts for this profile. Neither ingress fusion nor snapshot elision alone demonstrated a convincing serving speedup. Agentic trajectories are not speed A/Bs.

[Exact profile, results and limits](https://github.com/Avarok-Cybersecurity/atlas/blob/d707a0d9b1ebfc913c8296f2bed237c01b281924/bench/qwen4_exp/MTP_PERFORMANCE.md) · [Raw evidence and harnesses](https://github.com/Avarok-Cybersecurity/atlas/blob/d707a0d9b1ebfc913c8296f2bed237c01b281924/bench/qwen4_exp/mtp-performance-results.jsonl)

## Limits and remaining checks

The latest concurrency test uses repeated short requests under a 32K limit; it is not a stress test of two fully occupied 32K contexts or higher concurrency. Full EXL3 graph capture remains disabled. Vision, EP and the full supported-model matrix are not covered by this latest native validation, and no Docker image was rebuilt or published.

Existing PR-wide formatting, vendor-header and quantization-identifier spelling failures remain. Rustdoc also reports three existing EXL3 broken/private links. Changed-source formatting and diff checks pass. The branch includes current main at `8682329cc`.

## Authorship

AI-authored implementation and review; real GB10 tests and recorded model workloads. Historical investigations remain under `.research/` and `bench/qwen4_exp/`.


rsafier:
## temp-0 prefill is now bit-reproducible (second and last known source fixed)

`c904cae14` closes the determinism chain opened by `e76b349fe`.

**The defect.** With the MoE epilogue ordered, prefill was bit-reproducible below 2051 tokens and still 8/8 bit-distinct above it. 2051 is `index_topk + index_compress_ratio - 1` — the QSA activation bound — and the second source is in the device radix-select: the emit pass assigns output slots with `atomicAdd(&s_emitted, 1u)`, so the selected **set** came back in a different **order** every run.

Order is bit-significant, contrary to the comment it replaces. `qsa_prefill_attn` reads the block-id list directly and its online softmax rescales `m`/`l`/`acc` as it walks the list, so a permuted list changes the fp32 accumulation order — softmax is order-invariant in exact arithmetic, not in fp32. Decode was never affected because `qsa_expand_sel` already sorts ascending; that sort is precisely what made decode canonical.

**The fix.** Canonicalise the list ascending with a shared-memory bitonic sort before the kernel returns — the same trick decode uses. Emit and tie-fill are untouched, so the set and the tie-break are bit-unchanged; only the order becomes a pure function of the set. `QsaIndexer::new` now refuses `block_topk > 512` loudly rather than letting the kernel's sort bound skip silently. An index-ordered ballot-walk emit was measured and rejected: same final order, no saving, and it would put a two-level ballot prefix with a cross-warp tie quota into the code under repair.

**Measured** (dgx-00, 2.05bpw, temp 0; echo-probe = 8 identical requests, prompt logprobs compared bit-for-bit — text hashes are too weak, downstream BF16 casts swallow 1-ulp deltas):

| prompt tokens | pre-fix | fixed | fixed, prefix cache ON |
|---|---|---|---|
| 93 / 1938 | 1/8 | 1/8 | — |
| 2103 | 8/8 | **1/8** | 1/8 |
| 2213 / 2323 | 8/8 · 8/8 | **1/8 · 1/8** | — |
| 5513 | 8/8 | **1/8** | 1/8 |

TTFT median 2323 tok 4256 → 4264 ms (+0.2%), 5513 tok 11374 → 11386 ms (+0.1%) — inside run spread, so #820's prefill win survives. Set preservation proven three ways: a host-reference leg including a tie-heavy case (scores are relu-floored, so ties are the common case), the golden-fixture set-match test, and `qsa_decode_select_parity` (200 cases, device == host). New test `qsa_prefill_topk_bit_deterministic` fails pre-fix, passes fixed. No kill switch — semantics-preserving at unmeasurable cost, and a switch would exist only to re-enable a race.

**Where this leaves things.** qwen4_exp temp-0 prefill is bit-reproducible at every length tested (93–5513 tokens), prefix cache on or off, and at C=4. Still open and untested: C>1 above the QSA bound, MTP, and the NVFP4/materialized MoE control (which never booted). Gates on the merge result: 730 + 313 lib tests, clippy `--all-targets`, `exl3_native_parity` PASS at every leg.

rsafier:
## Phantom string-mangling: root-caused to the prefix cache (6-arm A/B)

`4966c8c5b` lands the experiment and its raw evidence under `.research/phantom_mangling/`. Merged latest main (#836) on top; merge result gated green.

**The symptom.** The model asserts that byte-identical strings differ and blames its environment — "I keep writing `axum` instead of `axum`", "`#[tokio::main]` instead of `#[tokio::main]`" — then burns whole turns rewriting files that were already correct.

**The experiment.** Counterbalanced ON/OFF/OFF/ON plus two Marconi-disabled controls, 3 iterations each, `agentic-webserver`, 4.05bpw EXL3 on one GB10, `ATLAS_AGENTIC_SAMPLING=model-card` (greedy hides this; the card's tools temperature exposes it).

| arm | config | prose | markers | rate/KB | marconi | med TTFT | Σwall | pass |
|---|---|---|---|---|---|---|---|---|
| A1 | ON | 16.51 K | 3 | 0.182 | 40 | 886 ms | 1232 | 3/3 |
| A2 | ON | 39.34 K | 14 | 0.356 | 49 | 935 ms | 2186 | 3/3 |
| B1 | OFF | 4.97 K | 0 | 0 | 0 | 8958 ms | 684 | 3/3 |
| B2 | OFF | 11.68 K | 0 | 0 | 0 | 10378 ms | 1359 | 3/3 |
| C1 | ON, no Marconi | 7.87 K | 0 | 0 | 0 | 9769 ms | 904 | 3/3 |
| C2 | ON, no Marconi | 10.83 K | 0 | 0 | 0 | 10607 ms | 1134 | 3/3 |

Pooled: **ON 17 / 55.85 KB = 0.304/KB**; OFF 0 / 16.65 KB (P(0) = 0.006); ON-no-Marconi 0 / 18.70 KB (P(0) = 0.003). Markers track Marconi intermediate restores exactly. `Prompt too long` is **zero in all six arms** — the context-window confound that voided an earlier attempt at this A/B is absent.

**Three things this changes.**

1. **The gate cannot see it.** All six arms passed 3/3 `webserver_ok` and `followed_directions` while A2 spent a third of its output litigating a difference that does not exist. Quality here needs marker scanning, not pass/fail — `scan.py` is committed for that (verifies byte-identity programmatically; STRICT tier requires `left == right`; normalises on assistant prose only so a large tool dump can't move the denominator).

2. **It's a throughput bug too.** Prose separates completely — every ON arm exceeds every OFF arm, pooled 3.35×. That's why a 10-iteration run took ~87 min against a historical band of 593–1025 s. Note `s/turn` moves the *other* way (28.5 ON vs 32–40 OFF): the cache does its per-turn job as designed; the model just needs far more turns with it on.

3. **The C arms do NOT isolate Marconi — read the TTFT column.** Disabling it removes essentially all warm-start benefit (9.8–10.6 s vs 8.9–10.4 s fully off, against 0.9 s on), so C is performance-equivalent to OFF and its cleanliness is consistent with either "Marconi replay specifically" or "any effective warm restore". With exact-leaf at 0 (sub-block matching now opt-in), **the Marconi intermediate replay *is* the prefix cache on this model**: the entire ~11× TTFT win and the confabulation arrive through one path. There's no cheap "disable the bad sub-feature" fix.

**Excluded on direct evidence** (not by this A/B): PLE n-gram indexing (hash→row-id bit-exact vs fixtures), tokenizer (encode→decode→re-encode is a fixed point for every trace string), QSA (symptom appears below the 2051 activation bound; its prefill-select race is separately fixed in `c904cae14`), and temp-0 nondeterminism (both causes fixed in `e76b349fe` + `c904cae14` — the symptom **survives** them, measured on this binary).

**Operationally, until the seam is fixed:** serve qwen4_exp agentic workloads without `--enable-prefix-caching`, price ~10 s TTFT vs ~0.9 s. The prefix re-anchor (`bd9c2ef2b`) pushes *more* traffic onto the implicated path and may want gating. The real fix is to make the warm suffix replay numerically match cold — activation diff, cold prompt vs warm restore of the same prompt, targeting the reused-KV vs replayed-SSM seam over `[snap_tok, matched)` — which keeps the 11× that turning the cache off does not.

No production code changed in this commit. Gates on the merge result: 730 + 313 + 104 lib tests, clippy `--all-targets`.

rsafier:
Current status: all follow-up code and evidence pushed through **`d707a0d9b`**. The branch includes current `main` (`8682329cc`); the native TUI binary is compiled and test servers are stopped.

- Added exact ingress fusion and removed the unused draft-prefix D2H snapshot. These reduce work, but their isolated serving A/Bs do **not** establish a throughput gain.
- Fixed C2 admission and draft KV/QSA lifetime, abandoned-proposal rewind, and the unwanted global shadow drafter.
- Final two-draft C2 check: 12/12 paired outputs match isolated controls, plus cancellation/reuse. Idle GPU allocation returned to 87,084 MiB after every pair; available host memory stayed above 20.59 GiB with zero swap-outs. Short requests, 32K configured limit; higher concurrency/full-context stress is unclaimed.
- All three widths pass deterministic short/long output checks. With **model-card sampling and preserve-thinking enabled**, one and two drafts pass all six agentic checks. Three drafts exceeded the 1,000-second budget in generated-test repairs and was cancelled (`run-1788629726444274721`); it is not a pass.
- Two drafts is the TUI default: preliminary three-repeat coding test median 31.134 s versus 33.286 s at one and 32.529 s at three. This is a prompt-specific observation, not an agentic speedup claim.

Model/server unit suites pass (775/2,361; ignored tests excluded), and CUDA workspace clippy passes. Existing PR-wide fmt/vendor-header/spelling/rustdoc failures remain; no Docker image was published.

Local launcher: `/home/ms/run-atlas-pr834-tui.sh`; `MTP_DRAFTS=1|2|3`, optional `MTP_CONCURRENCY=2`. Default concurrency remains one.

[Full status, fingerprints and run IDs](https://github.com/Avarok-Cybersecurity/atlas/blob/d707a0d9b1ebfc913c8296f2bed237c01b281924/bench/qwen4_exp/MTP_PERFORMANCE.md) · [Raw evidence](https://github.com/Avarok-Cybersecurity/atlas/blob/d707a0d9b1ebfc913c8296f2bed237c01b281924/bench/qwen4_exp/mtp-performance-results.jsonl)

rsafier:
Handoff at `2274d01d7`: all NVIDIA Qwen3.8 NVFP4 follow-up work is committed and pushed.

The downloaded `nvidia/Qwen3.8-Flash-Next-NVFP4` snapshot now loads and MTP engages, including during thinking. Fixed FP8 PLE gather support, main NVFP4 detection when FP8 draft weights are loaded, and the BF16 GDN output projection's serial/verify reduction mismatch.

Validation: 778 model tests passed; model clippy and changed-file formatting passed; native CUDA server rebuilt. GPU PLE and GDN tests pass against independent/serial references, including actual NVIDIA weights and failing old-path controls. Final greedy arithmetic, prose and code responses match serial completely (3/3). The forced weather tool request has the same function, arguments and finish reason, **but different reasoning**, so full tool/thinking parity remains unresolved.

Per the wrap-up request, the longer agentic run was not started. No NVIDIA MTP speedup, agentic pass, wider-draft serving parity or concurrency claim is made. Prior EXL3 results do not transfer to this checkpoint.

[Detailed handoff, exact profile, evidence and next commands](https://github.com/Avarok-Cybersecurity/atlas/blob/2274d01d723c1ff809588cb3eb07a443e8beedd6/bench/qwen4_exp/NVIDIA_MTP_HANDOFF.md). First follow-up: localize the forced-tool reasoning divergence, then run `agentic-webserver` with both `ATLAS_AGENTIC_SAMPLING=model-card` and `ATLAS_AGENTIC_PRESERVE_THINKING=1`; preserve deliberate speculation during thinking.

Local binary: `/home/ms/atlas/target/release/spark`. `/home/ms/run-atlas-pr834-tui.sh` now launches NVIDIA with two drafts and one sequence; MTP remains experimental. The previous EXL3 launcher is `/home/ms/run-atlas-pr834-exl3-tui.sh`. Test servers are stopped. Raw local artifacts remain in `/tmp/atlas-nvidia-mtp`, with selected evidence committed beside the handoff.

Reviewer scope note: `7b226ec73` modifies the shared quantization detector; explicit scheme precedence and EP-rank behavior are covered by tests. `98595757e` affects BF16 GDN verification only when per-row projection replay is requested. Existing broader PR hygiene failures remain documented; no Docker rebuild or full model matrix was performed.

rsafier:
## EP=2 on two GB10s: working, and prefill is faster than a single node

Speculative decode now runs under expert parallelism, and the two-node config beats single-node prefill. All numbers below are measured on gx10-9959 (head) + dgx-00 (worker), EXL3 4.05bpw, prefix caching on, identical binary sha256 on both nodes.

### Numbers

| metric | EP=2 (2 GB10s) | single node | delta |
|---|---:|---:|---:|
| prefill 8K | **542 tok/s** | 478 | **+13%** |
| prefill 11K | **561 tok/s** | 486 | **+15%** |
| decode aggregate C=1 | 26.4–27.2 | 26.7–27.6 | parity |
| decode aggregate C=2 | 22.2–23.7 | 23.7–24.3 | parity |
| decode aggregate C=4 | 22.3–22.8 | 21.8–22.4 | parity |
| experts resident per GPU | **256** | 512 | half |

Prefill is the win, and the mechanism is plain: prefill is dominated by MoE weight traffic, and EP=2 halves the experts each rank reads. Decode is unchanged because it is bound by the per-sequence mHC verify rather than weight bandwidth — the same reason cross-sequence batched verify measured *negative* (−25% at C=2, −33% at C=4) earlier in this session.

Agentic gate under EP=2 with MTP, `reasoning_effort: low` + `preserve_thinking`: **PASS** — 1/1 webserver_ok, 1/1 followed_directions, 15.9 s/turn, 112 s wall. Speculation stayed live across turns (`mtp=1.00`, `mean_na` 1.17–1.67).

### Four rank-asymmetries had to be fixed, each hidden behind the last

1. **`feat(mtp)` 43f82882d** — the draft MoE was sharded by `is_local_expert` while the upload never sharded `mtp.*`, so a rank>0 draft routed into NULL experts. Fixed by replicating it (`force_all_experts`); ~1.3 GB/rank against the ~20 GB/rank EP=2 saves, and it keeps a collective off the latency-critical draft path.
2. Same commit — the **native-EXL3 expert loader** shards independently via `local_expert_range()` and needed the same flag. The error message alone would not have revealed this one.
3. **`fix(mtp)` b0b918064** — `factory/build.rs` gated the *entire* MTP module load on `config.ep_rank == 0`, citing the precondition (1) had just removed. Left up, rank 1 warned "no MTP weights were loaded", set `has_mtp=false`, allocated `num_intermediates=0`, and died three layers away in the first mHC verify with `SSM MTP intermediate buffers not allocated`.
4. **`fix(ep)` 0d2c9503a** — the worker's verify arms received `num_accepted`, trimmed the proposer, rewound `seq_len`/tokens and rolled back the SSM, but never landed the aux carries. `commit_accepted_prefix`'s own comment names the class: *"PLE's rolling conv/history window and QSA's ingested/pooled marks are still `k - num_accepted` rows ahead"*. The worker's QSA ingest sat ahead of `seq_len` by the draft width and the next decode died on `QSA: decode at pos 3739 but 3741 tokens ingested`. Ordering matters: `commit_verify_aux_rows` asserts `seq_len == base + num_accepted`, so it must run *after* the rewind.

The bisect that isolated (4): the same agentic run passed with MTP **off** (137 s, 15.1 s/turn) and crashed with MTP **on**, which localised it to EP + speculation + a prefix-cache warm restore rather than to EP itself.

### How to run it

`ATLAS_EP_MTP=1` is the opt-in and stays default-off — one agentic iteration and 2-rep sweeps are a pass, not a certification. The refusal message names the flag and says why it is gated. EP=2 *without* MTP needs no flag and also passes the agentic gate.

`measure_prefill.py` / `measure_concurrency.py` gained `--host` (default `127.0.0.1`, so existing invocations are unchanged) so they can target an EP head from another node.

### Also on this branch since the last update

- **`fix(rollback)` dba869c04** — the content-loop watchdog restored SSM state and lowered `seq_len` but never rewound the QSA/PLE aux state, so the next decode 500'd with the same indexer-desync signature. Now saves/restores byte-exact aux blobs keyed by the same `(slot, ring_slot)` as the SSM ring, or declines with `RollbackFallback::NoAuxSnapshot` — a stream-safe hard stop.
- **`perf(ple)` d09ab8477** — n-gram fault-in worker cap 16 → 64, the measured knee on this drive (332 ms → 147 ms for 32768 O_DIRECT 4 KiB reads, 3 repeats; flat past 96). The shipped comment claiming benefits "flatten out well below" 16 was an assumption.
- **`feat(scheduler)` ac09ef25c** — the prefill admission window is decoupled from co-dispatch batching, so a scheduling policy can see an ordered queue. Measures flat on this workload; shipped as a knob, not a default.

Negative results recorded rather than retried: kernel-level fused prefill measures **0.4%** (prefill is throughput-saturated at ~630 tok/s regardless of how rows are grouped); decode checkpoint cadence refuted; interior SSM prefix anchors blocked on two gates; CUDA graphs structurally unavailable on the EXL3 path (cooperative launches are not stream-capturable). Details and fingerprints in `.research/exl3_decode_perf/EXL3_DECODE_PERF.md`.

rsafier:
## Since the EP=2 update: a host-sync review, four fixes, and a +29% prefill finding

All on `wip/exl3-research`, `66e5c2342` → `fb8aaca35`. Gates on every commit: clippy `--all-targets` clean, 823 lib tests, release build clean, agentic gate PASS.

### The headline: the MoE row cap was sized for half the real prefill chunk

`moe_prefill_cap.rs` justifies the 1024-row cap as *"12.8× the 80-row mean"*, where 80 = 4096 tokens × top_k 10 / 512 experts. But `--max-prefill-tokens` defaults to **8192** (`serve_args.rs:694`) and nothing overrides it, so the real mean is **160** and 1024 is only **6.4×** it — small enough that the overflow tier is structural rather than a tail event. Nothing couples the cap to the chunk size, so the cap silently became half-sized when the chunk default moved.

New counter-only telemetry (`ATLAS_EXL3_MOE_TIER_STATS=1`, default-off) on the host side of the tier select, which already reads `expert_offsets`:

```
cap=1024  calls_with_overflow=832/832 (100.00%)  overflow_experts_total=8868
```

**Every** sync-path MoE call has overflow experts — ~10.6 each, and each costs ~6 host-issued launches (three cooperative grids that serialize) behind the fused launch. The tier was assumed near-dead at this cap.

Counterbalanced sweep (1024, 2048, 4096, 4096, 2048, 1024 — one variable, 3 reps/length, telemetry live per arm so each proves its own cap):

| cap | prefill 4K | 8K | 11K | overflow experts | calls w/ overflow |
|---|---:|---:|---:|---:|---:|
| 1024 (shipped) | 522, 523 | 502, 504 | 493, 493 | 8868, 8867 | 100% |
| 2048 | 566, 567 | 543, 546 | 524, 534 | 4075, 4082 | 98% |
| 4096 | 680, 677 | 652, 649 | 627, 626 | 0, 0 | 0% |

**+29% at 8K (503 → 650), +30% at 4K, +27% at 11K.** Baseline returned to 523/504/493 against its opening 522/502/493 (no drift), each arm reproduced twice within ~2%, and the telemetry supplies the mechanism independently of the throughput number: overflow 8868 → 4075 → 0 tracks prefill 503 → 545 → 650 monotonically.

Validated at 4096: **128K × 4 boots** at util 0.72 (the ~315 MB of temp slabs fit); **agentic 3/3** at 12.73 s/turn with zero errors; **decode at parity** (C=1 27.4/28.2, C=2 23.9/23.7 vs the recorded 26.7–27.6 / 23.7–24.3).

**This commit changes no default.** Experts move from the chunked overflow path into the fused kernel — a different reduction — and the 128 → 1024 change carried an `exl3_native_parity` leg. Echoprobe at 1024 vs 4096 should land before the cap moves. Also worth recording as the *pair* (chunk 8192, cap 4096): zero overflow at 4096 is boundary-dependent (`max_rows_any_expert` was exactly 4096 against an 8192 chunk), and quoting a cap without its chunk is how the original rationale drifted.

### The negative result that shaped the rest: D2H drains are ~free on GB10

A four-dimension read-only review (locks, host-sync, serialization, memory) converged on one mechanism: `copy_d2h` / `copy_d2h_on_stream` drain the stream *inside* the call, and no D2H destination in the tree is pinned. The largest instance was the Marconi aux collect — one draining, fresh-`Vec` D2H per aux-carrying layer, the QSA half O(context), on every checkpoint (64 decode tokens) and chunk-boundary save.

Batched it (one page-locked blob, N async enqueues, one sync). Counterbalanced A/B, one binary, one variable, arm liveness proved from the server's own log line:

**+1.2%, not the ~15% the arithmetic predicted.**

The projection came from the `copy_d2h_async` doc's measured shape (66.8 MB as 60 blocking copies in ~400 ms ≈ 165 MB/s). That figure is *not* transfer bandwidth — it is mostly the host waiting for already-queued GPU work, attributed to the copy call. **GB10 is unified memory**: host and device are the same LPDDR5X, so a D2H is not a bus transfer and "pageable vs pinned" does not mean what it does on a discrete GPU. `pinned_host_staging.rs` had already posed exactly this question and this run answers it.

That retired two candidate levers on evidence (the per-draft argmax D2H; the MoE `expert_offsets` drain — which the cap change largely neutralises anyway, since three quarters of agentic MoE calls take the `s ≤ cap` no-sync shortcut).

### Fixes

- **`66e5c2342`** — the decode-rollback aux ring is keyed by `(slot, ring_slot)`, and `compact_sequence` migrates a *surviving* sequence into a *retiring* one's slot. Nothing cleared either end, so a rollback could restore another request's QSA/PLE carry. Aliasing, not a leak. Inert under the shipped preset (the ring is skipped when speculation is on), live in the non-MTP config.
- **`16534c30a`** — `raise_smem_once` memoised the 90 KB smem raise in a process-global set keyed on a raw `CUfunction`, never cleared on model swap. A recycled address makes the raise silently skip and the next model's first EXL3 GEMM fail. Same bug class `derived_weights.rs` was restructured to remove.
- **`6c4285e1a`** — KV spill staged the **entire** KV image on the host before writing a byte (~3.1 GB at 32K, larger than the whole `--swap-space-gb` budget). Now windowed to ~24 MB. Also closes a pre-existing leak: `restore` assigned `block_table` only at the end, so a mid-restore failure leaked every allocated block despite the caller's `free_sequence` cleanup.
- **`4e692e067`** — the K-row verify tail ran K single-CTA argmaxes each with its own blocking 4-byte D2H; now one batched argmax + one readback via the existing `argmax_batch_dispatch` (byte-identical per its own doc).

### Agentic: five arms, all PASS

Greedy ×2, model-card, model-card with partial guards off, model-card with **all** watchdogs off — 15 iterations, 3/3 `webserver_ok` and `followed_directions` each, zero desyncs / CUDA faults / panics.

Two things worth keeping from it:

**Greedy reproduces; the harness does not.** `agent.rs:87` pins temp 0 / seed 0 unless `ATLAS_AGENTIC_SAMPLING=model-card`, so the gate default *is* greedy — and two greedy runs of the same config differed **1.9×** in wall (389 s vs 740 s). Trajectory diff localises it exactly: `run-00` is byte-identical through turn 3 and the first difference is *inside a tool result* (cargo printed `Adding matchit v0.8.4 (available: v0.8.6)` in one run only). **So agentic wall-clock cannot carry a perf claim** — turn count dominates and it is trajectory noise. Decode tok/s is flat at 21.5–21.9 across arms.

That is also positive evidence about `ATLAS_DFLASH_SPEC_THINK=1`, which this preset sets: it lifts a gate whose comment records deterministic 8-9/10 agentic trajectory failures from the 2026-08-16 bisect. If batch-K verify were flipping low-margin tokens mid-reasoning, those trajectories would not be byte-identical up to a cargo line.

**Loop guards are three independent mechanisms with a trap.** `ATLAS_DISABLE_WATCHDOGS=1` is the real switch and does *not* cover `ATLAS_SIMHASH_LOOP` or `ATLAS_LOOP_NO_SUPPRESS`. `--content-loop-watchdog false` is **inert** on this model (`[behavior]` never sets `enable_loop_watchdog`) — the arm built on it is recorded as PARTIAL rather than presented as a middle setting.

Full records and fingerprints in `.research/exl3_decode_perf/` (`cap_sweep/`, `cap_validate/`, `aux_ab/`, `agentic_arms/`) with the harnesses that produced them.

PR #837, opened 2026-09-01
gdn: pointer-table twins for wyN K=5..16 (batched-verify foundation, behavior-neutral)

rrstesiak:
Part 1 of the batched K=gamma GDN verify work (the C=16 lane). BEHAVIOR-NEUTRAL by construction: nothing dispatches to the new code yet, so this merges with zero risk and no gate impact.

What it adds, all additive:
- `state_is_table` pointer-table addressing inside both wyN impls (K=5..16), the exact mode wy2/3/4 already carry: when flagged, each block resolves its sequence's H and Hi_t through per-sequence device pointer tables instead of contiguous strides.
- 24 new symbols (`gated_delta_rule_wy{5..16}_table` + `_f16_table`) with the flag compiled in. Every pre-existing symbol keeps a byte-identical signature (constant 0 internally), so the shared `gdn_decode_wyn` launch and the per-target wy17 kernels are untouched.
- Rust side: table-kernel registries (testable name consts), the f16-routed `wyn_table_kernel` selector, and the `gdn_decode_wyn_table` launch wrapper.

Measured motivation (nsys node-trace, 2026-09-01, C=16): `gated_delta_rule_wy10` running per-sequence was the #1 kernel in the entire engine on BOTH workloads: 187,392 launches / 27.7% of GPU time on prose, 59,904 / 24.5% on code, 16 x 48 x steps exactly.

Part 2 (stacked next) flips the dispatcher and carries the tests + receipts.

### DIFF
--- a/crates/spark-model/src/layers/ops/ssm_gdn_b.rs
+++ b/crates/spark-model/src/layers/ops/ssm_gdn_b.rs
@@ -349,6 +349,73 @@ pub fn gdn_decode_wyn(
         .launch(stream)
 }
 
+/// Cross-sequence batched-verify launch of the wyN pointer-table twins
+/// (`gated_delta_rule_wy{5..16}_table` / `_f16_table`): `state_is_table`
+/// is compiled into the symbol, so the argument list is identical to
+/// [`gdn_decode_wyn`] — but the STATE arguments mean something different
+/// and batching is the point:
+/// - `h_tables`   = the layer's staged table slice: slab 0 holds one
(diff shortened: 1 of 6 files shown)


tbraun96:

tbraun96:

|---|---|---|---|---|
| 4 | #837 | gdn wyN pointer-table twins K=5..16 (foundation) | gdn | **← this PR** |

tbraun96:

PR #838, opened 2026-09-01
gdn: switch on the K=5..16 batched verify (#837 stacked) — 201.0 tok/s C=16 median, byte-identical

rrstesiak:
Part 2, stacked on #837 (contains its commit, same pattern as the #817/#818 stack): the dispatcher's 5..=16 arm goes live and VERIFY_WY_TABLES_PER_LAYER grows 4 -> 16 so staging funds h + Hi0..Hi14 (192 KB). Gamma-width verifies now take the two-launch path; anything the preconditions decline still falls back per-sequence, byte-identical.

Receipts (full reproduction key in the records ledger, 2026-09-01 entries):
- The engine's own log on first serve: "batched-verify GDN conv+WY ENGAGED (n=16, k=10): per-layer 304 launches -> 2 (batched conv kernel + table-form wy10)".
- Correctness in its strongest form: 96/96 outputs byte-identical to the PRE-kernel build's outputs across both workloads (code sha 1d5627bf x48, prose 42aa6d55 x48). Cross-build, cross-code-path byte identity, temp 0, C=16.
- Code C=16: 201.0 tok/s aggregate median (202.9/201.0/201.0), up from 198.8. To our knowledge the first 200+ C=16 aggregate median for speculative decode on a DGX Spark, and it is deterministic.
- Prose C=16: 72.3 median, up from 71.1. The honest finding: at 96% GPU saturation prose is bytes-bound (the K-1 rollback snapshots dominate wy10 traffic), so the launch collapse alone cannot move it far. The next levers (gamma-vs-C schedule, f16 h-pool, resident Pass-2) target bytes; they follow in their own PRs.

Tests: table-registry index contract, k=16 staging capacity (written red before the const bump, green after), slab-stride kernel contract, C=16 seqs capacity. Full spark-model suite green; both crates check clean.

Note for the gate ledger: this changes the verify perf path, so the standing benchmark records will want a re-cut on this head per house precedent; sequencing at the maintainers' convenience.

### DIFF
--- a/crates/spark-model/src/layer/transformer_layer.rs
+++ b/crates/spark-model/src/layer/transformer_layer.rs
@@ -33,8 +33,14 @@ mod default_loops;
 /// is a pure host-side layout change). 48 GDN layers x 4 tables x 32
 /// entries x 8 B = 48 KB.
 pub const VERIFY_WY_TABLE_SEQS: usize = 32;
-/// Tables per GDN layer: h_state + Hi0..Hi2 (K=4 verify → 3 intermediates).
-pub const VERIFY_WY_TABLES_PER_LAYER: usize = 4;
+/// Tables per GDN layer: h_state + Hi0..Hi14 (K=16 verify → 15
+/// intermediates). 4 → 16 (2026-09-01): the wyN pointer-table twins
+/// extend the cross-sequence batched verify to K=5..16, so the staging
+/// gate `(2..=THIS)` must admit gamma-width verifies; at 4 every K>4
+/// batch silently declined to the per-sequence loop (measured 09-01:
+/// 187,392 wy10 launches = 27.7% of GPU time in a C=16 prose window).
+/// Cost: 48 layers x 16 x 32 x 8 B = 192 KB of host+device tables.
+pub const VERIFY_WY_TABLES_PER_LAYER: usize = 16;
 /// Bytes between consecutive tables within a layer slice.
 pub const VERIFY_WY_TABLE_STRIDE_BYTES: usize = VERIFY_WY_TABLE_SEQS * 8;
 /// Bytes between consecutive GDN layers' table slices.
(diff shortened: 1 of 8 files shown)


TheTom:
Read both commits plus the #837 table form underneath (static, no hardware today). Two questions:

1. `gated_delta_rule_wyn.cu:91` hoists `Hi_ptrs[K_TOKENS - 1]` per thread: at K=16 that is 15 pointers, 30 registers, on top of the existing update chain, and every thread loads the same 15 table entries. The body has no ptxas line for the `_table` twins. What do `ptxas -v` registers and spills look like for `gated_delta_rule_wy16_table` and the `_f16_table` twin? If it spills, the 15 pointers can sit in shared memory loaded by one warp.

2. `trait_decode_batched_conv_gdn_wyn.rs:54` says a zero f16 twin handle "returns None so the caller's f16 refusal path (`require_wy_f16`) fires". In the dispatcher `None` returns `Ok(false)` at `trait_decode_batched_conv_gdn_multi.rs:158`, before `require_wy_f16` at `:165`, so that refusal never runs for K=5..16; the per-sequence loop's own `wyn_kernel` is what refuses. Intended, with the comment just naming the wrong guard?

The receipts read well: cross-build byte identity on 96/96 outputs is the strongest form in this stack, and the code number is a trio.

crates/spark-model/src/layer/transformer_layer.rs
```diff
@@ -644,3 +650,44 @@ pub trait TransformerLayer: Send + Sync {
...
+        );
+    }
+
+    /// The table-form wyN kernels REINTERPRET the stride argument as
+    /// "pointer entries between Hi slabs" and the dispatcher passes
+    /// `VERIFY_WY_TABLE_SEQS`. That only reaches slab t iff slabs are laid
+    /// out back-to-back at exactly SEQS entries (8 bytes each). Pin the
+    /// layout the kernels assume.
+    #[test]
+    fn slab_stride_matches_the_kernel_contract() {
```
TheTom, line 680: **Nit.** Both assertions restate the definitions at lines 45-48 (`STRIDE_BYTES = SEQS * 8`, `LAYER_STRIDE = TABLES_PER_LAYER * STRIDE_BYTES`), so no change to the constants can fail this test. The contract worth pinning is the one the doc names: the dispatcher passes `VERIFY_WY_TABLE_SEQS` as the kernel's `slab_entry_stride`. Assert that at the call site (or through a small helper the dispatcher uses) and drop this one. The other two tests here are real.

crates/spark-model/src/layer/transformer_layer.rs
```diff
@@ -33,8 +33,14 @@ mod default_loops;
 /// is a pure host-side layout change). 48 GDN layers x 4 tables x 32
```
TheTom, line 33: **Nit.** "48 GDN layers x 4 tables x 32 entries x 8 B = 48 KB" is now 16 tables and 192 KB, as the `VERIFY_WY_TABLES_PER_LAYER` doc below says.

crates/spark-model/src/layers/ops/ssm_gdn_b.rs
```diff
@@ -349,6 +349,74 @@ pub fn gdn_decode_wyn(
...
+///   `VERIFY_WY_TABLE_SEQS`-entry stride.
+/// - `slab_entry_stride` MUST be `VERIFY_WY_TABLE_SEQS as u32` — the
+///   kernel reinterprets the contiguous-form stride argument as "pointer
+///   entries between Hi slabs".
+///
+/// Tables are staged by `upload_verify_wy_tables` (verify_e2.rs) into a
+/// fixed buffer refreshed pre-replay, so the launch is CUDA-graph-stable.
+#[allow(clippy::too_many_arguments)]
+// provenance-id: 526f6e616c6420522e205374657369616b
+#[allow(dead_code)] // seam 2026-09-01: consumer = the dispatcher 5..=16 arm, next change
```
TheTom, line 369: **Nit.** This `#[allow(dead_code)]` (and the one on `wyn_table_kernel` at `trait_decode_batched_conv_gdn_wyn.rs:60`) was #837's "consumer lands next" marker. This PR is the consumer, so both are stale now and would hide a real dead-code regression later.

crates/spark-model/src/layers/qwen3_ssm/trait_decode_batched_conv_gdn_wyn.rs
```diff
@@ -45,6 +45,35 @@ impl Qwen3SsmLayer {
         (k.0 != 0).then_some(k)
     }
 
+    /// The pointer-table wyN twin for the CROSS-SEQUENCE batched verify at
+    /// width `num_tokens` ∈ {5..16} (`state_is_table` compiled in), or
+    /// `None` when out of range / module absent / `ATLAS_GDN_WYN=0` — the
+    /// caller then runs the per-sequence loop, byte-identical math. FP16
+    /// pool routing mirrors [`Self::wyn_kernel`]: over an f16 h pool only
+    /// the f16 twin is correct, and a zero twin handle returns None so the
+    /// caller's f16 refusal path (`require_wy_f16`) fires instead of
```
TheTom, line 54: **Nit.** The refusal named here does not run on this path: the dispatcher turns `None` into `Ok(false)` at `trait_decode_batched_conv_gdn_multi.rs:158`, before `require_wy_f16` at `:165`. The per-sequence loop's `wyn_kernel` is what refuses under an f16 pool without its twin. Safe, but the comment should name that guard.

TheTom:
What reads clean, for the record: the table form's addressing matches the upload layout exactly (`hi(1)` is the Hi0 slab, slab t at `(1+t) * VERIFY_WY_TABLE_STRIDE_BYTES`, kernel index `b + t * VERIFY_WY_TABLE_SEQS`), the contiguous symbols keep their bytes, `gdn_decode_wyn_table` fails closed on a NULL table, `ATLAS_GDN_WYN` is opt-out, and the `4 =>` arm is safe because the first match already declines anything outside `2..=16`. With the ptxas numbers from the question above this is an approve from me; the rest are nits.

tbraun96:

tbraun96:

|---|---|---|---|---|
| 5 | #838 | gdn K=5..16 batched verify switch-on (stacks on #837) | gdn | **← this PR** |

tbraun96:

PR #844, opened 2026-09-03
dflash-2: C=16 write-on-accept verify, batched drafter precompute, slot-ordered dispatch, fp8 drafter GEMM twins (prose 187.7, code 269.7)

rrstesiak:
## Where this leaves the tree

Best trio-grade numbers measured per cell. One DGX Spark GB10, Qwen3.8-27B NVFP4, temperature 0, median of 3, same prompts per column (MinHeap / Volvo at C=1; SinglyLinkedList / Volvo standard at C=16). The four PRs are the difference between the two DFlash-2 rows; #956 tracks landing them.

| path | C=1 code | C=1 prose | C=16 code | C=16 prose |
|---|---:|---:|---:|---:|
| serial, no speculation | 16.2 | 16.2 | 129.3 | 132.3 |
| MTP (main) | 27.0 | 24.4 | 163.2 | 150.4 |
| DFlash-2 (main, stock defaults) | 66.1 | 24.5 | 173.8 | 83.8 |
| DFlash-2 + #837 #838 #844 #845 | **66.9** | **27.3** | **279.5** | **182.5** |
| DFlash-2 with the line vs MTP | 2.5x | +12% | +71% | +21% |

Main rows: the 09-01 three-arm matrix on stock `main` at `ef8ca11d`, same box. PR row receipts are in the tables below and on #956.

---

Stacked on #838 (which stacks on #837), same pattern as the #817/#818 stack: until those land, the diff of this PR shows their two commits underneath these four.

Four pieces, one commit each, in the order they landed. They are one PR because the
benchmark certification is per tree and the pieces were measured together; each commit
is reviewable on its own and carries its own kill switch. Category: performance (1, 2, 4)
and one dispatch-order fix (3).

| commit | piece | kill switch |
|---|---|---|
| 1 | write-on-accept K=4 GDN verify (the record) | `ATLAS_NO_GDN_WOA=1` |
| 2 | batched drafter ctx precompute | `ATLAS_DFLASH_NO_BATCHED_PRECOMPUTE=1` |
| 3 | slot-ordered batched dispatch (fix) | none needed; output bytes unchanged |
| 4 | fp8 drafter GEMM `_k64` twin | `ATLAS_DFLASH_FP8_GEMM_P2=1` (original) |

## Receipts on the combined tree (Qwen3.8-27B NVFP4, one DGX Spark, gamma 4, gate forced)
| C=16 aggregate | before | after |
|---|---|---|
| prose (Volvo) | 119.5 best fixed-gamma, MTP 150.4 | **180.7** median of 3 (180.7 / 180.5 / 181.6), 46/48 on 42d547eb (the parent stream) |
| code (MinHeap) | 201.0 fp32 record, 241.0 f16 mode | **279.5** median of 3 (279.5 / 279.8 / 277.6), 47/48 on 16383c26 |
| prose, `ATLAS_NO_GDN_WOA=1` (same tree) | | 149.1, same stream 42d547eb: the twin is +21% |
| code, `ATLAS_NO_GDN_WOA=1` (same tree) | | 245.8, same stream 16383c26: the twin is +14% |
| code, gamma 10 (parent kernel + fold restore) | 201.0, stream 1d5627bf | 214.2, stream 1d5627bf 16/16, byte-identical (measured on the pre-review tree; re-gate on this head pending) |

C=1 untouched by construction (batched paths run only at two or more sequences): MinHeap 66.9 / 66.8 / 66.8
and Volvo 24.8 at gamma 10 on this tree. Step: verify 164 -> 112 ms, propose 96 -> ~54 ms.
Branch `pr/dflash2-c16-write-on-accept` on dflash-wyn-table @110c2c53; every new file
carries the provenance-id.

## Benchmark prompts, exact text

The C=16 rows above were measured with two prompts written for the concurrency sweep on 2026-08-29, not the Atlas standard MinHeap / Volvo prompts their labels suggest. Re-measured on the standard prompts, same tree and lane, the numbers land in the same class. Both sets stay on the record with their own prompt, marked by superscript.

| workload | C=16, sweep prompt (published) | C=16, Atlas standard prompt | stream (standard) |
|---|---|---|---|
| code | 279.5 median of 3 (279.5 / 279.8 / 277.6) <sup>1</sup> | 268.8 median of 3 (267.7 / 268.8 / 269.3) <sup>3</sup> | 3ae47861 on 15 / 14 / 15 of 16 |
| prose | 180.7 median of 3 (180.7 / 180.5 / 181.6) <sup>2</sup> | 182.5 median of 3 (181.7 / 183.0 / 182.5) <sup>4</sup> | 24eb6e04 on 15 / 14 / 15 of 16 |

All four prompts ran at the same `max_tokens: 300`, 16 concurrent requests (4800 completion tokens per run), temperature 0, `stream: false`; aggregate = total completion tokens over wall time. C=1 rows elsewhere in this PR: MinHeap at `max_tokens: 1500`, Volvo at `max_tokens: 300`. The published streams (16383c26, 42d547eb) belong to prompts 1 and 2; a different prompt is a different stream, so the two columns are not stream-comparable, only class-comparable.

<sup>1</sup> Code, sweep prompt, `max_tokens: 300` (receipts `csweep_*`):
> Write a Python class SinglyLinkedList with methods insert_front, insert_back, delete_value, search, and reverse. Include a __repr__ and a brief docstring for the class.

<sup>2</sup> Prose, sweep prompt, `max_tokens: 300` (receipts `csweepP_*`):
> Write a 200-word essay on the Volvo P1800.

<sup>3</sup> Code, Atlas standard prompt, `max_tokens: 300` at C=16 (receipts `csweepM_*`); the same prompt at `max_tokens: 1500` is every C=1 MinHeap row in this PR and on the record:
> Write a complete Python MinHeap class implementing push, pop, peek, and heapify, with sift_up and sift_down helpers and a docstring for each method.

<sup>4</sup> Prose, Atlas standard prompt, `max_tokens: 300` (receipts `csweepV_*`); C=1 Volvo rows dated before 2026-09-04 used prompt 2 at the same length:
> Write a detailed history of the Volvo car company.

Standard-prompt receipts: `csweepM_dflash_c16_20260904_155119 / 155228 / 155307`, `csweepV_dflash_c16_20260904_155352 / 155436 / 155519`, resolver head with the K=4 lane selected by concurrency (identical to `--dflash-gamma 4` on #844).

## 1. gdn: write-on-accept for the batched K=4 verify

At C=16 the batched K=4 GDN verify
(`gated_delta_rule_wy4`) was 30% of the step and already at the DRAM wall
(~270 GB/s). It reads H twice and writes four full state widths per launch
(Hi0, Hi1, Hi2, final H), and a partial accept then copies one of them back over
H. Most of those bytes are thrown away at the verdict. The lever is bytes, not cycles.

### What this does
- `gated_delta_rule_wy4_woa` (new, kernels/gb10/common/gated_delta_rule_wy4_woa.cu):
  same arithmetic in the same order as the parent, H slice staged in 64 KB of
  dynamic smem (one DRAM read), NO state writes. Stashes per row the WY-corrected
  update vector vn, the gate scalar, and the key row. `output` is byte-identical.
- `gated_delta_rule_wy4_fold`: one launch per GDN layer after the verdict, over
  the batch. Reads H once, applies rows 0..accepted with the parent's update
  expression, writes H once. When the parent kernel ran instead (engaged word = 0)
  it performs the parent's partial-accept restore from the same Hi pointer tables
  the verify uploaded, so the host never guesses which kernel the graph replayed.
- Engaged marker is a device word written by the woa kernel INSIDE the captured
  graph (replays count); a 1-thread clear kernel resets it after the fold.
- Scheduler (`verify_dflash_batch_step.rs`): SSM commit deferred until after the
  verdict loop; one `gdn_fold_accepted(slots, rows, k)` call, then per-sequence
  `commit_accepted_prefix` restores conv state only for folded slots.
- Kill switch: `ATLAS_NO_GDN_WOA=1` (parent kernel, unchanged bytes).

### Traffic per step (48 layers x 16 seqs)
Parent: 2 reads + 4 writes (+1 restore copy) per state. Here: 1 read (woa) + 1 read
+ 1 write (fold). Verify step 164 -> 112 ms measured (nsys + server log).

### Receipts (Qwen3.8-27B NVFP4, one DGX Spark, gamma 4, gate forced)
| C=16 aggregate | before | after |
|---|---|---|
| prose (Volvo) | 119.5 (best fixed-gamma), MTP 150.4 | **180.7** median of 3 (180.7/180.5/181.6), 46/48 on 42d547eb |
| code (MinHeap) | 201.0 (fp32 record), 241.0 (f16 mode) | **279.5** median of 3 (279.5/279.8/277.6), 47/48 on 16383c26 |
| code, gamma 10 (K=11, parent kernel + fold restore) | 201.0, stream 1d5627bf | 214.2, stream 1d5627bf 16/16 (byte-identical; pre-review tree, re-gate on this head pending) |

Output is the parent's, bit for bit. The first revision of this PR reported 187.7 prose on a
moved stream (54c44ca9) and blamed "the fold's contraction"; both were wrong. ptxas on the
built PTX shows FFMA=0 in the parent `wy4`, the `woa` twin, and the `fold` (the build emits
`mul.rn`/`add.rn`, fmad off), and the three kernels are expression-identical per operation, so
there is nothing to round differently. The stream had moved because of a host-side bug in the
fold bookkeeping (a finished sequence left a stale entry in the folded-slot list and the next
occupant of that slot skipped its h restore), found in review and fixed here. On the fixed tree
`ATLAS_NO_GDN_WOA=1` and the twin produce the same majority stream on both workloads (table
above), and `examples/gdn_woa_oracle.rs` asserts parent vs twin+fold bit-equal on device:
ALL PASS, 38 checks, 0 mismatches, max ULP 0. The 187.7 is retracted; 180.7 is the number for
the correct stream. The one- or two-slot minority stream per run shows on the kill-switch arm
too, so it belongs to the batched verify, not to this twin. Acceptance in the workload band
(47% prose, 75% code). C=1 untouched by construction (batched path runs only at two or more
sequences): MinHeap 66.9 / 66.8 / 66.8 (crown 66.6 class) and Volvo 24.8 on this build at gamma 10.

### Notes for reviewers
- ptxas: woa 72 regs / 0 spill, fold 40 regs. The kernel registry sets the >48 KB dynamic
  smem attribute at launch; a launch error propagates (no silent swap inside a capture).
- The twin runs only on explicit request: the DFlash batched step passes
  `VerifyBatchedOpts { write_on_accept: true }`, the MTP/K4 verify passes the default, and the
  request is part of the verify graph key. Stash and engaged words are bound lazily on the
  first request (pre-capture, `min(32, max_decode_seqs)` seqs); the fold consumes a per-verify
  eligibility flag set only by a verify that ran under the request with WY tables staged.
- The fold takes the verify width K from the host; K != 4 goes through the parent
  path + restore (the gamma 10 row above is that path, reproducing the fixed-K stream
  byte for byte). A wyN write-on-accept twin for K = 5..16 is the follow-up.
- Reproduction key: branch `pr/woa-k4-write-on-accept` on dflash-wyn-table @110c2c53
  (parked diff `dflash-g4-woa-precompute-2026-09-03-r2-foldk.diff`); launch
  `launch-dflash-wyn-gamma-c16.sh 4`, bench `bench-csweep-prose.sh 16 dflash` /
  `bench-csweep.sh 16 dflash`. Kill switch `ATLAS_NO_GDN_WOA=1` gives the parent bytes.

## 2. dflash: batched drafter ctx precompute

The per-sequence `precompute_ctx_kv` streams the drafter's `fc` (262 MB) and fused
KV (105 MB) weights once per SEQUENCE per step over a 2-4 row tail. At C=16 that was
16 serial passes, 41 ms of GPU per step, plus ~22 ms of host stalls from the
synchronous row-by-row K/V compaction copies (one cuStreamSynchronize each, ~500
per step). nsys 2026-09-02: 63 ms of a 270 ms step.

### What this does
- `precompute_ctx_kv_batched.rs` (new): gathers every sequence's uncommitted ctx rows
  into one contiguous staging slab and runs the SAME pipeline once (fc GEMM, norm,
  fused KV GEMM, per-layer k-norm, one RoPE, per-layer cache write). Slot mapping and
  positions built host-side, one upload each. Every op is row-independent, so the
  output bytes are unchanged.
- Per-sequence Phase I in `propose.rs` is deferred on the batched propose path (tail
  computed exactly once). Oversized tails and any failure fall back to the
  per-sequence loop with nothing committed.
- `precompute_ctx_kv.rs`: the K/V compaction copies become `copy_d2d_async` on the
  caller's stream (same ordering; every consumer is on that stream).
- Kill switch: `ATLAS_DFLASH_NO_BATCHED_PRECOMPUTE=1`.

### Receipt
C=16 prose gamma 4: 126.0/126.2 -> 142.7/146.4 (two rounds), 16/16 then 15/16 on the
SAME majority stream 42d547eb as before (bit-exact as designed). propose= per step
96 -> 54 ms. Staging: 256 rows x 51 KB.
Gate on the final build: gamma 10 C=16 code reproduces the 201.0 record's stream 1d5627bf
16/16 with this path engaged (the per-sequence and batched precompute are bit-exact).
the provenance-id.

## 3. fix(scheduler): slot-ordered batched dispatch

Category: fix / architecture. Not a performance PR: it corrects a dispatch-order
assumption that silently disabled a path already in the tree.

### The defect
The batched conv+WY GDN path requires the batch to sit on CONSECUTIVE ssm-pool slots in
batch order (checked on the pointers in `trait_decode_batched_conv_gdn_multi.rs`). The
dispatcher in `mtp_step.rs` handed it the batch in ARRIVAL order. The pool freelist is
LIFO (`ssm_pool.rs` pop/push): the first fill hands out slots 0..n in arrival order and
the check passes, but every later fill re-issues released slots in reverse finish order,
the layout check fails, and the same batch DECLINES to the per-sequence loop with no
error and one log line per boot. Measured: ~47% of steps declined in a two-round C=16
prose bench; 140 of ~260 steps in the 2026-09-02 nsys trace ran
`causal_conv1d_update_l2norm` x4 per sequence per layer.

### The fix
Sort the uniform batch by `(ssm_slot_idx, index)` before dispatch. Drafts travel inside
each `ActiveSeq`, so batch order is free to the caller; the borrow walks the active list
once in ascending index order and restores the chunk's slot order before the call.
Sequences without a slot sort last; the model re-checks the layout and declines safely
if a gap remains. One hunk, no kernel change, output bytes unchanged.

### Receipt
Zero DECLINED lines across both rounds (was one per boot); round 2 matches round 1
(126.2 / 126.0). The launches were inside the captured graph, so the wall effect is small
(~3%, nsys session 121.8 vs 118.2 on the fallback round); the point is that the batched
path now engages on every round, which the write-on-accept work depends on.

## 4. w4a16_gemm.cu: fp8 drafter GEMM k64 twin

The drafter's row-scaled FP8 GEMM at M=64 (n x gamma) ran at 26-39% of DRAM peak
(nsys 2026-09-02: 31 ms/step). One byte-identical twin (same K order per output,
same MMA sequence): `fp8_gemm_t_row_scaled_k64`, 64 B per weight row per K step (half the
row-opens per byte; the W4A16 family's `_k64` lever). Handle chain in `from_weights.rs`:
k64 -> original; `ATLAS_DFLASH_FP8_GEMM_P2=1` pins the original. The first revision also
carried a 4-stage `_p4` twin; it measured neutral on this box and was removed in review.

Receipt: propose= per step 55.8 -> 52.2 ms (k64), same majority stream. The kernel's
remaining wall is the [N,K] K-contiguous weight layout (short bursts per row); a
load-time transpose + [K,N] tile is the follow-up.
Gate on the final build: gamma 10 C=16 code reproduces the 201.0 record's stream 1d5627bf
16/16 with `_k64` selected. Branch `pr/fp8-drafter-gemm-twins` on dflash-wyn-table
@110c2c53; both kernels carry the provenance-id. ptxas: no spills on either twin.

## Test plan
- [x] `cargo fmt --all -- --check` (touched files clean; two pre-existing diffs on the base belong to #838 and are fixed there)
- [x] `cargo clippy -p spark-model -p spark-server --tests -- -Dwarnings` clean (base clippy fix folded into #838)
- [x] `cargo check -p spark-server` on the branch
- [x] Tested on hardware: Qwen3.8-27B NVFP4 + DFlash-2 on one DGX Spark, receipts above with output shas
- [x] Tests added: `examples/gdn_woa_oracle.rs` (device oracle, parent vs twin+fold, bit-equal, run with `ATLAS_TARGET_MODEL=qwen3.8-27b cargo run -p spark-model --release --features gpu-examples --example gdn_woa_oracle`); unit tests for the engage decision (`woa_decision`), the stash width against the kernel layout, and the write-on-accept bit in the verify graph key
- [x] `ATLAS_NO_GDN_WOA=1` A/B on the final tree, prose and code, same majority stream both arms (table above)
- [ ] MTP n=2..4 serve sanity run on this head (the MTP path never requests the twin; run listed for completeness)
- [ ] Benchmark certification: ten-gate campaign on the frozen sha, records committed (serialized-landing protocol, `docs/gate-queue-protocol.md`)

## CLA


TheTom:
Read the four commits above #838 (static read only, my Spark is tied up). Before I write up findings, four questions:

1. Was write-on-accept measured on any MTP-drafted GDN model at n=2..4? `woa_now` in `trait_decode_batched_conv_gdn_multi.rs:281` gates on `kk == 4` and nothing caller-specific, and `decode_verify_batched` is also reached from `verify_k4_batch_step.rs:119`. That path commits through `verify_k4_verdict.rs:124/141` and never calls `gdn_fold_accepted`. Every receipt in the body is DFlash.

2. The body says the fold's contraction rounds differently from the parent's fused update. The three update sites (`gated_delta_rule_wy4.cu:147`, `wy4_woa.cu:174`, `wy4_woa.cu:261`) are the same source expression, so the only thing that can differ is nvcc's FMA choice per kernel. Did you look at the SASS (FFMA vs FMUL+FADD on the `g*h + sk*vn` chain)? If that is the cause, `__fmaf_rn` at all three sites should give back stream 42d547eb.

3. Which driver or arch refused the 64 KB dynamic-smem launch that the `WOA_DISABLED` fallback exists for? `registry.rs:448` already calls `cuFuncSetAttribute(MAX_DYNAMIC_SHARED_SIZE_BYTES)` for any launch over 48 KB, so the kernel header's "launches without the opt-in attribute" reads as stale. Related: that `Err` arm runs inside graph capture.

4. `_p4` is recorded as neutral on this box. Is it kept for a named follow-up, or can it go?

crates/spark-model/src/layers/ops/ssm_gdn_b.rs
```diff
@@ -389,3 +457,113 @@ pub fn conv1d_update_chunk2(
...
+        .arg_u32(v_dim)
+        .arg_u32(qk_stride)
+        .arg_u32(v_stride)
+        .arg_u32(gb_stride)
+        .arg_u32(stash_seq_floats)
+        .arg_ptr(engaged_flag)
+        .launch(stream)
+}
+
+/// Post-verdict fold (`gated_delta_rule_wy4_fold`): applies rows 0..na_tab[b]
```
TheTom: rustdoc treats `na_tab[b]` as an intra-doc link and `spark-model` has `#![deny(warnings)]`, so this is what fails the "Build mdBook + rustdoc" job. Wrapping it in backticks (`` `na_tab[b]` ``) or escaping the brackets (`na_tab\[b\]`) fixes it.

crates/spark-model/src/layers/ops/ssm_gdn_b.rs
```diff
@@ -389,3 +457,113 @@ pub fn conv1d_update_chunk2(
 }
 
 // ── Activations / Element-wise ─────────────────────────────────────
+
+/// Write-on-accept K=4 verify (`gated_delta_rule_wy4_woa`): byte-identical
+/// `output` to `gdn_decode_wy4`, writes NO state, stashes the per-row update
+/// terms for `gdn_wy4_fold`. Table form only. Grid (num_v_heads, batch), 128.
+#[allow(clippy::too_many_arguments)]
+pub fn gdn_decode_wy4_woa(
```
TheTom: This file is now 569 lines and is not on the file-size-cap allow-list, so it is one of the two files failing the 500-LoC job. The three new launchers (`gdn_decode_wy4_woa`, `gdn_wy4_fold`, `gdn_wy4_flag_clear`) are a self-contained cluster; moving them to a sibling `ssm_gdn_woa.rs` with a `mod` + `pub use` in this file brings it back under the cap.

crates/spark-model/src/layers/qwen3_ssm/init.rs
```diff
@@ -339,6 +343,29 @@ impl Qwen3SsmLayer {
...
+            ),
+            gdn_wy4_clear_k: super::super::try_kernel(
+                gpu,
+                "gated_delta_rule_wy4_woa",
+                "gated_delta_rule_wy4_flag_clear",
+            ),
+            woa_flag,
+            // vn[4][nv][vd] | g[4][nv] | sk[4][nk][kd], f32, one slab per
+            // batch entry (VERIFY_WY_TABLE_SEQS). ~131 KB/seq at 48x128.
+            woa_stash: gpu
```
TheTom: `init.rs` is the second file over the 500-LoC cap (511 lines) without an allow-list entry. The write-on-accept fields added here (`gdn_wy4_woa_k`, `gdn_wy4_fold_k`, `gdn_wy4_clear_k`, `woa_flag`, `woa_stash`, `woa_stash_seq_floats`, `woa_dims`, `woa_armed`) plus the `woa_flag` alloc at line 27 could be built by a small `fn woa_init(gpu, nk, nv, kd, vd) -> Result<WoaInit>` in a sibling file, which also keeps the stash-size arithmetic in one place.

TheTom (commented):
CI fixes for the current head, all mechanical. Details inline; the non-file ones are in a separate comment.

TheTom:
CI status on `b0b82e64d`, for the record:

- **Build mdBook + rustdoc**: fails on one broken intra-doc link, `na_tab[b]` in `crates/spark-model/src/layers/ops/ssm_gdn_b.rs:517`. Inline comment has the fix.
- **Enforce ≤500 LoC**: two changed files crossed the cap without an allow-list entry, `crates/spark-model/src/layers/ops/ssm_gdn_b.rs` (569) and `crates/spark-model/src/layers/qwen3_ssm/init.rs` (511). Inline comments suggest the split for each.
- **CLAAssistant**: shows `cancelled`, not a failure. A re-run of the check should clear it.
- **PR categorize (advisory)**: exits 1 with `Unable to flush stdout: Broken pipe` after sanitizing the body; looks like the workflow, not this PR. Same signature on #824.
- **seal status / PR benchmark gate**: expected red until `/stamp` and the benchmark records land.

Everything else is green: fmt, clippy --tests, test --workspace, metal test, nvcc -> PTX, typos, SPDX, deny, llvm-cov.

rsafier:
### dflash-2: MTP prompt-hidden capture overruns its allocation when `--max-seq-len` > `ATLAS_DFLASH_CTX_CAP`

Hit on `pr/dflash2-c16-write-on-accept` @ `b0b82e64d`, GB10 / DGX Spark, `unsloth/Qwen3.8-27B-NVFP4` + `incoai/Qwen3.8-27B-DFlash2`:

```
--dflash --dflash-gamma 8 --max-seq-len 131072 --max-batch-size 8 \
--gpu-memory-utilization 0.76 --ssm-cache-slots 64 --enable-prefix-caching
ATLAS_DFLASH_SPEC_THINK=1
```

Prefill dies once a prompt walks past 16,384 tokens:

```
ERROR spark_runtime  copy_d2d_async failed (status 1)
  1: ...AtlasCudaBackend>::copy_d2d_async
  2: ...drafter_prefill::...::try_mtp_prefill_capture_from
  3: ...prefill_b::forward_layers::...::prefill_b_forward_layers
  4: ...prefill_b::...::prefill_chunk_dispatch
  5: ...::prefill_chunk
  6: spark::scheduler::phase_continue_prefills::run_standard::run_standard_chunk_loop
ERROR spark  Prefill chunk error: cuMemcpyDtoDAsync_v2 (copy_d2d_async) failed: status 1
```

`status 1` is `CUDA_ERROR_INVALID_VALUE` — a destination range outside the allocation, not OOM (that would be status 2).

**Cause looks like two numbers for one buffer disagreeing.**

- `impl_a1.rs:470` narrows the allocation under DFlash: `capture_rows = max_seq_len.min(dflash_ctx_cap())`, where the cap defaults to **16384** (`dflash_head.rs:942`). This is deliberate — the comment there notes 1342 MB at `--max-seq-len 131072` against a 16K window that can hold 168 MB of it.
- `impl_a1.rs:811` then sets `mtp_prefill_capacity: max_seq_len` unconditionally, i.e. **131072**.

So the guard in `try_mtp_prefill_capture_from` (`drafter_prefill.rs:109`):

```rust
if chunk_start + proc_count > self.mtp_prefill_capacity { return Ok(()); }
```

approves writes up to row 131072 into a 16384-row buffer, and the `copy_d2d_async` just below it writes off the end.

**Suggested fix:** set `mtp_prefill_capacity` to the same `capture_rows` value used for the allocation. Long prompts then skip the whole-prompt drafter prefill via the existing coverage check at the consume site — which is what the alloc-site comment already describes as intended: *"a prompt past the window simply does not get the whole-prompt drafter prefill … it costs acceptance on very long cold turns, not correctness."*

**Why the gate doesn't catch it:** `concurrency-sweep-dflash2` pins `max_model_len = "4096"`, and 4096 < 16384, so the mismatch cannot trigger there. Any DFlash serve above 16K context hits it.

**Workaround** meanwhile: `ATLAS_DFLASH_CTX_CAP=0`, which makes `capture_rows = max_seq_len` so the allocation matches the capacity — at the cost of the full 1342 MB.

*Caveat on attribution: found by reading the code after the traceback, not bisected against #838/#837 or `main`. Since the capacity line is a plain `max_seq_len`, this is likely pre-existing and merely surfaced by a >16K DFlash config rather than introduced by this PR.*

crates/spark-model/src/layers/qwen3_ssm/trait_decode_batched_conv_gdn_multi.rs
```diff
@@ -267,41 +276,34 @@ impl Qwen3SsmLayer {
         let gate_ptr = gates_buf;
         let beta_ptr = gates_buf.offset(nv * fp32);
         let hi = |t: usize| wy_tables.offset(t * VERIFY_WY_TABLE_STRIDE_BYTES);
-        match kk {
-            2 => ops::gdn_decode_wy2(
+        // Write-on-accept (2026-09-03): the K=4 twin writes no state; the
+        // model folds the accepted rows after the verdict (`gdn_fold_accepted`).
+        let woa_now = kk == 4
```
TheTom: **Blocker.** `woa_now` fires for every K=4 batched verify, but only one of the two callers folds.

`decode_batched_conv_gdn_multi` is reached through `decode_verify_batched`, which has two scheduler callers: `verify_dflash_batch_step.rs:73` (calls `gdn_fold_accepted` at :192) and `verify_k4_batch_step.rs:119`, the MTP batched K-row verify. That one is reachable by default (`ATLAS_MTP_MAX_SEQS` defaults to 32 in `ladder.rs:261`, and the ladder gives K=4 at n<=4; ragged D-Cut batches are split into uniform-k runs in `trait_decode_batched.rs`, so a K=4 run still takes this arm). It applies its verdict through `verify_k4_verdict.rs:124` (full accept: "the verify kernel already wrote the canonical h_state, so the commit is a no-op") and `:141` (partial accept: restore `intermediate[na]`). Neither calls the fold.

With write-on-accept on, the woa kernel writes no state and no intermediates. On the MTP batched path a full accept leaves H at the pre-verify value, and a partial accept restores an `Hi(na)` the woa never wrote. That is silent recurrent-state corruption at C=2..4 on any GDN model served with the MTP drafter. All the receipts in the body are DFlash, so this configuration was not exercised.

Suggested fix: make write-on-accept an explicit request from the caller rather than a layer-local default. A `write_on_accept: bool` on `ConvGdnArgs` (or the verify call) that only `step_verify_dflash_batched` sets, plus a refusal if a woa launch happened and no fold followed. The alternative is adding the fold to `step_verify_k4_batched` and proving it with a receipt on an MTP model at n=2..4.

kernels/gb10/common/gated_delta_rule_wy4_woa.cu
```diff
@@ -0,0 +1,283 @@
...
+    unsigned int stash_seq_floats,
+    // Device word this layer's fold reads: 1 = the woa twin ran for the last
+    // batched verify (set here, INSIDE the captured graph, so replays count);
+    // the fold's clear kernel resets it. 0 = the parent wy4 ran.
+    unsigned int* __restrict__ engaged_flag
+) {
+    const unsigned int vh = blockIdx.x;
+    const unsigned int b = blockIdx.y;
+    if (vh >= num_v_heads || b >= batch_size || k_dim != WOA_KD) return;
+    if (threadIdx.x == 0) *engaged_flag = 1u;
```
TheTom, line 86: **Blocker.** The engaged flag is set here inside the captured graph and cleared only after a fold. Any step that runs the woa and then skips the fold leaves `flag=1` and a stale stash.

Paths that do that today: `verify_dflash_batch_step.rs:75-84` (`decode_verify_batched` returns Err; when the error comes after the GDN launches or the replay, the flag is already set, then drafts are restored and the function returns), `:88-97` (short result, return), the MTP batched path from the other comment, and any Err out of `gdn_fold_accepted`.

On the next step, if the batched arm declines (the body says a layout gap is still possible) or the parent kernel runs, `gated_delta_rule_wy4_fold` reads `flag=1` and applies the previous step's `vn`/`g`/`sk` to this step's H (lines 230-283). No log line.

Suggested fix: launch the clear kernel at the top of the batched arm, before the WY launch and inside the same capture, so the flag describes this launch only. One extra one-thread node per layer per step. A per-step sequence number instead of 0/1, compared in the fold, would also work.

crates/spark-model/src/model/trait_impl/mod.rs
```diff
@@ -846,6 +846,53 @@ impl Model for TransformerModel {
     fn sync_secondary(&self) -> Result<()> {
         self.sync_secondary_dispatch()
     }
+    fn gdn_fold_accepted(
```
TheTom, line 852: **Blocker.** This gates on the allocation (`self.verify_wy_tables.is_null()`), not on whether this step's upload succeeded. `upload_verify_wy_tables` (`verify_e.rs:244`) returns NULL when `n + ghosts > 32`, `k` is out of range, or any sequence has fewer than `k-1` intermediates in the tiered pool.

On NULL the layer declines (`trait_decode_batched_conv_gdn_multi.rs:170-172`), the per-sequence loop runs, the flag stays 0, and the fold takes the parent-restore branch. Both pointer tables it reads, `h_state_table[b]` for the destination and `hi_tables` for the source, hold whatever the last successful upload staged, which can be other sequences' slots, so it can overwrite a different sequence's H. `async_chkpt.rs:235` then skips the host h restore because the slot was recorded in `gdn_woa_folded_slots`.

Suggested fix: record `wy_tables_base.is_null()` (with n and k) for the last batched verify on the model, and return `Ok(false)` here when it was NULL. `gdn_woa_folded_slots` stays empty and the host restore runs as before.

kernels/gb10/common/gated_delta_rule_wy4_woa.cu
```diff
@@ -0,0 +1,283 @@
...
+    const float g2 = *WOA_G(sb, 2, vh), g3 = *WOA_G(sb, 3, vh);
+    const float vn0 = WOA_VN(sb, 0, vh, v_dim)[tid], vn1 = WOA_VN(sb, 1, vh, v_dim)[tid];
+    const float vn2 = WOA_VN(sb, 2, vh, v_dim)[tid], vn3 = WOA_VN(sb, 3, vh, v_dim)[tid];
+
+    if (k_dim != WOA_KD) return;
+    #pragma unroll
+    for (unsigned int j=0; j<WOA_KD; j+=4) {
+        float h0=H[(j+0)*v_dim+tid], h1=H[(j+1)*v_dim+tid];
+        float h2=H[(j+2)*v_dim+tid], h3=H[(j+3)*v_dim+tid];
+        h0=g0*h0+sk0[j]*vn0; h1=g0*h1+sk0[j+1]*vn0;
```
TheTom, line 276: **Should fix.** The body says the majority stream moved because "the fold's contraction rounds differently from the parent kernel's fused update". This line, the woa pass-2 line 174, and the parent's `gated_delta_rule_wy4.cu:147` are the same source expression, so the only thing that can differ is nvcc's FMA contraction choice per kernel, which the source does not control.

If that is the cause it is fixable: write the update as `__fmaf_rn(g, h, sk * vn)` at all three sites so the contraction is pinned by source, then re-run the gamma-4 prose receipt expecting the pre-woa stream (42d547eb from commit 2's receipt) rather than 54c44ca9. `cuobjdump -sass` on the fold and the parent, comparing FFMA against FMUL+FADD in the update chain, would confirm before touching anything.

If it is not fixable, the default flip should wait: a path whose state bits differ from the one it replaces is opt-in until it has its own A/B receipt.

crates/spark-model/src/layers/qwen3_ssm/init.rs
TheTom: **Should fix.** `VERIFY_WY_TABLE_SEQS (32) * 4 * (nv*vd + nv + nk*kd) * 4` is 131,840 bytes per sequence slot, so about 4.2 MB per GDN layer and roughly 200 MB over 48 layers, allocated on every model construction: with `ATLAS_NO_GDN_WOA=1`, with the kernels absent, at C=1, and without DFlash. Same for `woa_flag` at line 27.

Allocating on the first `woa_now`, or gating on `verify_wy_tables` being non-null (the model already knows whether a batched verify can exist) and on the kernel handle, keeps the cost on the serves that use it.

crates/spark-model/src/layers/qwen3_ssm/trait_layer.rs
```diff
@@ -12,6 +12,56 @@ use super::Qwen3SsmLayer;
...
+        stream: u64,
+    ) -> Result<bool> {
+        // Flag-driven, not host-armed: the batched verify replays as a CUDA
+        // graph, so the only reliable record of WHICH kernel ran is the device
+        // word the woa twin writes inside the graph. The fold kernel reads it
+        // and either folds the stash (woa ran) or performs the parent's
+        // partial-accept restore from the Hi tables (parent ran). Either way
+        // h is committed here, so the caller skips its h restore.
+        let _ = self
+            .woa_armed
```
TheTom: **Nit.** `woa_armed` is written at `trait_decode_batched_conv_gdn_multi.rs:321` and swapped here, but the value is discarded. It is exactly the host-side "a woa launch happened and no fold consumed it" signal the stale-flag case needs, so either assert on it here or delete the field.

crates/spark-model/src/model/trait_impl/mod.rs
```diff
@@ -846,6 +846,53 @@ impl Model for TransformerModel {
...
+                self.gpu.as_ref(),
+                h_table,
+                self.gdn_woa_na_tab,
+                k_rows,
+                accepted_rows.len(),
+                stream,
+            )?;
+            ssm_idx += 1;
+        }
+        if any {
```
TheTom: **Should fix.** `gdn_woa_folded_slots` is only cleared when `any` is true. A false return leaves the previous step's slots in the list, and a later `commit_accepted_prefix` for a re-claimed slot skips the h restore. The per-sequence paths call it too (`verify_dflash_step.rs:211`, `verify_k4_verdict.rs:124/141`). Clearing unconditionally at function entry closes it.

crates/spark-model/src/layers/qwen3_ssm/trait_decode_batched_conv_gdn_multi.rs
```diff
@@ -310,33 +312,132 @@ impl Qwen3SsmLayer {
...
-                stream,
-            )?,
+            ) {
+                Ok(()) => {
+                    woa_launched = true;
+                    self.woa_armed
+                        .store(true, std::sync::atomic::Ordering::Release);
+                }
+                Err(e) => {
+                    // Most likely the 64 KB dynamic-smem launch was refused.
```
TheTom: **Question.** This `Err` arm runs inside graph capture. Whether a failed `cuLaunchKernel` leaves the capture valid depends on the error; an invalidated capture then fails at end-capture and the whole verify goes eager or errors.

Also, `registry.rs:448` already calls `cuFuncSetAttribute(MAX_DYNAMIC_SHARED_SIZE_BYTES)` for any launch over 48 KB, so the "launches without the opt-in attribute" note in the kernel header is stale, and the refusal this arm handles would be that attribute call failing rather than the launch itself.

Setting the attribute on the woa handle once at construction and deciding `woa_now` from that probe removes the in-capture failure path entirely.

kernels/gb10/common/gated_delta_rule_wy4_woa.cu
```diff
@@ -0,0 +1,283 @@
...
+    unsigned int gb_stride,
+    unsigned int stash_seq_floats,
+    // Device word this layer's fold reads: 1 = the woa twin ran for the last
+    // batched verify (set here, INSIDE the captured graph, so replays count);
+    // the fold's clear kernel resets it. 0 = the parent wy4 ran.
+    unsigned int* __restrict__ engaged_flag
+) {
+    const unsigned int vh = blockIdx.x;
+    const unsigned int b = blockIdx.y;
+    if (vh >= num_v_heads || b >= batch_size || k_dim != WOA_KD) return;
```
TheTom: **Nit.** This returns before the flag write when `k_dim != 128`, but the host already gates `kd == 128 && vd == 128`, so the guard is dead. Meanwhile `v_dim > 128` is not guarded in-kernel: `tid < v_dim` covers only 128 columns and the rest of the row would be silently unwritten. Guard both dims or neither, and say which side owns the check. Line 256 re-checks `k_dim` after loading `g`/`vn`; it can move up next to the `na == 0` check.

crates/spark-server/src/scheduler/verify_dflash_batch_step.rs
```diff
@@ -178,16 +178,29 @@ pub fn step_verify_dflash_batched(
...
         // Row of THIS sequence's bonus generator in the shared hidden buffer.
         stash_rows.push(off + num_accepted);
     }
 
+    // ── Write-on-accept fold (one launch per GDN layer over the batch), then
+    // the per-sequence commit (conv restore; h too when nothing folded). ──
+    {
+        let slots: Vec<usize> = batch.iter().map(|a| a.seq.slot_idx).collect();
+        let rows: Vec<u32> = accepted_per_seq.iter().map(|&na| (na + 1) as u32).collect();
+        if let Err(e) = model.gdn_fold_accepted(&slots, &rows, k) {
```
TheTom, line 197: **Nit.** A `gdn_fold_accepted` error here is logged and every sequence's `commit_accepted_prefix` still runs, with `gdn_woa_folded_slots` either empty or holding the previous step's slots. A commit error marks the sequence finished; a fold error should do the same for the batch, since the h state is no longer trustworthy either way.

crates/spark-model/src/layers/dflash_head/precompute_ctx_kv_batched.rs
```diff
@@ -0,0 +1,196 @@
...
+                batched.push((i, committed, count));
+            }
+        }
+
+        let mut lo = 0usize;
+        while lo < batched.len() {
+            // Greedy group: consecutive jobs whose rows sum to <= budget.
+            let mut hi = lo;
+            let mut rows = 0usize;
+            while hi < batched.len() && rows + batched[hi].2 <= PRECOMPUTE_BATCH_ROWS {
```
TheTom: **Should fix.** The group budget is `PRECOMPUTE_BATCH_ROWS` (256), but `precompute_ctx_kv` has no `new_ctx_count <= ctx_window` check and `fc_proj`, `fused_kv_out` and `slot_mapping_dev` are sized for `ctx_window` rows (`from_weights.rs:339-349`). The default `ctx_window` is 4096 so this is safe by default, but `ATLAS_DFLASH_CTX_WINDOW=128` (a documented override) makes the `copy_h2d` at line 117 a 2 KB write into a 1 KB buffer and the GEMMs write past `fc_proj`. Clamp the budget to `min(PRECOMPUTE_BATCH_ROWS, ctx_window)` or `ensure!` the relation at construction.

crates/spark-model/src/layers/dflash_head/precompute_ctx_kv_batched.rs
```diff
@@ -0,0 +1,196 @@
...
+                    .as_any_mut()
+                    .downcast_mut::<DflashProposerState>()
+                    .expect("checked above");
+                let src = d.ctx_hidden_acc.offset(committed * ctx_slot_bytes);
+                let dst = self.scratch.precompute_in.offset(row_off * ctx_slot_bytes);
+                gpu.copy_d2d_async(src, dst, count * ctx_slot_bytes, stream)?;
+                positions.extend_from_slice(&d.ctx_positions[committed..committed + count]);
+                for idx in committed..committed + count {
+                    let logical = idx / BLOCK_SIZE;
+                    let physical = *d.block_table.get(logical).ok_or_else(|| {
```
TheTom, line 119: **Question.** Slots are built from the host `d.block_table` here, while the per-sequence path runs `fill_slots_from_block_table` from `block_table_dev`. `propose.rs:308-353` updates both in the same place so they agree today. A `debug_assert_eq!` on the two lengths, or a comment naming the invariant, would keep a future device-only growth path from silently staging wrong slots.

kernels/gb10/qwen3.6-27b/nvfp4/w4a16_gemm.cu
```diff
@@ -6949,6 +6949,265 @@ extern "C" __global__ void fp8_gemm_t_row_scaled(
...
+// same 32-wide slabs in the same order with the same MMA sequence, so
+// every accumulator sees the identical rounding: output is byte-identical
+// to fp8_gemm_t_row_scaled (verified by sha on the record bench).
+//
+// smem: A 4x64x40x2 = 20 KB, B 4x128x32 = 16 KB (36 KB, fits static).
+// Grid/Block unchanged: (ceil(N/128), ceil(M/64), 1) x (128, 1, 1).
+// provenance-id: 526f6e616c6420522e205374657369616b
+// ═══════════════════════════════════════════════════════════════════
+#define FP8_RS_STAGES 4
+extern "C" __global__ void fp8_gemm_t_row_scaled_p4(
```
TheTom: **Should fix.** The body records `_p4` as neutral on this box ("latency was not the wall") and keeps it as the second choice. That is 130 lines of PTX-built kernel that never runs unless `ATLAS_DFLASH_FP8_GEMM_P4=1`. A measured no-op is better removed than merged dormant; if it is being kept for a named follow-up (a box where latency is the wall), say which, otherwise drop it and the middle link of the handle chain at `from_weights.rs:237`.

TheTom:
**Tests.** The diff above #838 contains no test files, but the test plan has "Tests added or updated where applicable" checked. Either uncheck it or add the test this change needs.

The test that would settle both the contraction question and the MTP-path question: a kernel-level oracle for one GDN layer with random H, k, v, g and beta, running the parent (table form) against woa plus fold for `na` in 1..=4 and for the `flag=0` parent-restore branch, asserting `output` and H bit-equal (or recording the ULP delta with its cause). Plus one scheduler-level check at n=2 on the MTP batched path comparing H after a full and a partial accept against the per-sequence loop.

TheTom:
**Receipts.** Prose 187.7 is a median of three. Code 269.7 is n=1 with "trio pending", and it is in the PR title. A single run is not a receipt; the trio should land before the number does. The C=1 numbers are from this tree, good. There is no receipt on any MTP-drafted model, which is where the first blocker lives.

tbraun96:

rrstesiak:
Replies to every thread above, one block per thread in review order. Head is now 7f51e945 (b0b82e64 + 538ca211 + 7f51e945); the PR body is updated with the receipts on this head, including the retraction of the 187.7 figure and the `ATLAS_NO_GDN_WOA=1` A/B.

---
## Top-level: the four questions (TheTom, 20:56)

1. No. Every receipt was DFlash. You are right that the MTP batched K-row verify reached the same layer code at K=4 and committed through `verify_k4_verdict` without a fold. Fixed by making write-on-accept an explicit per-call request: `VerifyBatchedOpts { write_on_accept }` on `decode_verify_batched`, set only by `step_verify_dflash_batched`, carried on `ForwardContext::gdn_write_on_accept` and in the verify graph key. The MTP path passes the default and always gets the parent wy4. `woa::woa_decision` is a pure function with a test that pins "unrequested never engages" at n=2..4.

2. Checked in SASS rather than guessed. ptxas (sm_121f) on the built PTX: `gated_delta_rule_wy4` FFMA=0 FMUL=352 FADD=298; `_wy4_woa` FFMA=0 FMUL=544 FADD=442; `_wy4_fold` FFMA=0 FMUL=1024 FADD=512. The build emits `mul.rn`/`add.rn` everywhere (fmad off), so there is no contraction difference between the three sites and the sentence in the PR body was wrong. The three kernels are expression-identical per operation, so the pair should be bit-equal to the parent; the new `gdn_woa_oracle` example asserts exactly that (result below). The moved stream therefore has a host-side cause, and the stale folded-slot entry you found in another thread is the leading suspect: a sequence that finished inside the verdict loop skipped its commit, its slot stayed in the list, and the next sequence claiming that slot skipped its h restore. The A/B with `ATLAS_NO_GDN_WOA=1` on this tree is the receipt: prose C=16 gamma 4, woa off 149.1 tok/s on stream 42d547eb (14/16), woa on 180.7 median-of-3 on the same stream 42d547eb. Same output either way, the twin is worth 21% at C=16, and the moved stream was the host-side folded-slot bug, now fixed. Code C=16 gamma 4 on the same tree: woa off 245.8 on stream 16383c26 (15/16), woa on 279.5 median-of-3 on the same stream; +14%.

3. None. The header note was stale: `registry.rs` sets the max-dynamic-smem attribute for every launch over 48 KB, so the only refusal that arm could see is that attribute call failing, which is a build or device mismatch, not a runtime condition. The `WOA_DISABLED` fallback is gone; a launch error now propagates to the caller instead of swapping kernels inside a capture. Header corrected.

4. Neither. Dropped, with its env pin. `_k64` and the `ATLAS_DFLASH_FP8_GEMM_P2` A/B pin remain.

---
## CI comment (TheTom, 21:04)
All four addressed on the new head: rustdoc bracket fixed; `ssm_gdn_b.rs` is 457 lines with the three write-on-accept launchers in `ops/ssm_gdn_woa.rs`; `init.rs` is 494 with the kernel resolution in `qwen3_ssm/woa.rs` and the stash no longer allocated at construction at all (see the 200 MB thread). rustdoc with `-D warnings`, fmt, clippy `--tests` and the workspace tests are clean locally. CLA re-run requested.

---
## rsafier: MTP prompt-hidden capture overrun
Confirmed and fixed exactly as you described: `capture_rows` is computed once, outside the branch, and `mtp_prefill_capacity` takes that value instead of `max_seq_len`. Your attribution caveat holds too: the narrowing landed on main in #818, so this was pre-existing and surfaced by the first serve above the 16K cap. Thank you for the traceback plus the read; that saved a bisect.

---
## Inline threads

**`trait_decode_batched_conv_gdn_multi.rs:281` (blocker, MTP path):** see question 1. The decision is now `self.woa_now(ctx.gdn_write_on_accept, kk, n)`; the request bit is the first term.

**`gated_delta_rule_wy4_woa.cu:71` (blocker, stale flag):** took your first suggestion. `woa_clear_at_entry` launches the clear kernel at the top of every batched verify that requests write-on-accept, before any decline and inside the same capture, so the word describes that launch only. The post-fold clear is gone. A verify error, a short result, or a fold error can no longer leak a stash into the next step.

**`trait_impl/mod.rs:849` (blocker, tables NULL):** the model now records `gdn_woa_eligible` at the END of a batched verify that ran under the request with `wy_tables_base` non-null, clears it at the START of every batched verify, and the fold consumes it. A NULL upload leaves it false, the fold returns `Ok(false)`, `gdn_woa_folded_slots` stays empty, and the host restore runs as before.

**`wy4_woa.cu:261` (contraction):** see question 2: FFMA=0 in all three kernels, so `__fmaf_rn` would change nothing and was not added. The oracle example is the receipt for the kernel; the A/B is the receipt for the stream.

**`init.rs:364` (200 MB at construction):** allocation moved to the model and made lazy: `gdn_woa_bind` runs pre-capture on the first write-on-accept request, sizes the stash for `min(32, max_decode_seqs)` sequences, and binds each GDN layer's slab and engaged word by address. Nothing is allocated on an MTP serve, at C=1, with the kernels absent, or under `ATLAS_NO_GDN_WOA=1`. The 4-byte engaged words moved with it.

**`trait_layer.rs:31` (`woa_armed` discarded):** deleted. Under graph replay the host never runs that store, so it could not be the signal; the device word plus the model's eligible bit are.

**`trait_impl/mod.rs:888` (`gdn_woa_folded_slots` stale):** cleared unconditionally at fold entry.

**`conv_gdn_multi.rs:325` (Err arm in capture):** removed; see question 3.

**`wy4_woa.cu:70` (dim guards):** neither, and the header says so: the host owns the 128 x 128 check (`woa_decision`), because a silent in-kernel return would leave `output` unwritten with no kernel having run in its place. The re-check at 256 is gone with it.

**`verify_dflash_batch_step.rs:192` (fold error):** a fold error now finishes every sequence in the batch and returns, the same verdict a commit error gets.

**`precompute_ctx_kv_batched.rs:83` (budget vs ctx_window):** budget is `min(PRECOMPUTE_BATCH_ROWS, ctx_window)`, and `precompute_ctx_kv` carries an `ensure!` on rows <= ctx_window as the backstop.

**`precompute_ctx_kv_batched.rs:105` (host vs device block table):** comment naming the invariant plus a `debug_assert!` that both exist when slots are built.

**`w4a16_gemm.cu:6972` (`_p4`):** removed.

---
## Tests comment (TheTom, 23:09)
Added: (1) `examples/gdn_woa_oracle.rs`, the kernel-level oracle you described: one GDN layer, random H/q/k/v/gate/beta over N sequences, parent (table form) against woa plus fold for na in 1..=4 and the flag=0 parent-restore branch, asserting `output` and H bit-equal and reporting mismatch count and max ULP otherwise. Result on GB10: ALL PASS (bit-equal). 38 checks, 0 mismatches, max ULP 0: twin output vs parent (98304 bf16 values); twin writes no state for all 4 seqs; engaged word set by the twin; fold na=1..4 vs parent state for all 4 seqs (786432 values each); flag=0 na=1..3 restores Hi(na-1); flag=0 na=4 is a no-op. Reproduction: `ATLAS_TARGET_MODEL=qwen3.8-27b cargo run -p spark-model --release --features gpu-examples --example gdn_woa_oracle` (seqs=4 nk=16 nv=48 seed=1). (2) Unit tests for the engage decision (`woa_decision`: unrequested never engages at n=2..4; each gate declines alone), the stash width against the kernel layout, and the write-on-accept bit splitting the verify graph key. The scheduler-level MTP check at n=2 became unnecessary: with the request off there is no woa launch on that path to compare against.

## Receipts comment (TheTom, 23:16)
Agreed on both. Prose trio on this head (fixed tree, write-on-accept on, gamma 4, C=16): 180.7 / 180.5 / 181.6 tok/s, median 180.7, majority stream 42d547eb on all three (15/16, 16/16, 15/16). That stream is the parent's, so the 187.7 in the PR body is retracted: it was measured on the tree with the stale folded-slot entry, which skipped the h restore and produced the 54c44ca9 stream. 180.7 is the number for the correct stream. The one-off minority stream (44d2542c at one slot) shows in the pre-fix 42d547eb runs at the same rate, so it predates this PR. Code trio on this head (same serve, `bench-csweep.sh 16`, SinglyLinkedList 300 tok temp 0): 279.5 / 279.8 / 277.6 tok/s, median 279.5, stream 16383c26 on 16/16, 16/16, 15/16. That replaces the n=1 269.7 in the body; the code stream also moved off the buggy tree's stream, as the oracle predicts, and its identity check against the parent is the `ATLAS_NO_GDN_WOA=1` code run in the A/B above. The MTP-model receipt is no longer the blocker's remedy since the MTP path cannot reach the twin, but an n=2..4 MTP serve sanity run on this head is listed in the test plan.

tbraun96:

|---|---|---|---|---|
| 6 | #844 | dflash-2 write-on-accept verify + drafter precompute + fp8 twins | dflash | **← this PR** |

tbraun96:
**AutoMerger sieve verdict: `sieve:integrity` — excluded from certification until cured.**

Three-sieve static review: SECURITY **pass** (all 57 files in `kernels/gb10/` + `crates/`, zero gate/BENCH.toml/workflow touches), GOAL **pass** (all four commits serve C=16 speculative decode, per-commit kill switches). INTEGRITY **fail** on two of four commits: the `fp8_gemm_t_row_scaled_k64` twin (`w4a16_gemm.cu` @6949, selected in `dflash_head/from_weights.rs:216`) has **no parity test** — `examples/gdn_woa_oracle.rs` covers only the GDN wy4 path; and the slot-order fix (`scheduler/mtp_step.rs:399` `by_slot.sort_by_key`) has **no test that fails under arrival order** — both rest on runtime receipts, not committed tests. Counterweight noted: the GDN write-on-accept piece is excellently covered (38/38 bit-equal oracle, `woa_decision` gate tests).

**What clears it:** an fp8 parity oracle mirroring `gdn_woa_oracle`, and a `by_slot`-sort unit test that fails under LIFO order. The content is wanted — once those exist this re-enters stack `certified-2026-09-04` (#877). If you'd rather I write them, say so and I'll push them to this branch under my own identity.

tbraun96:
**Sieve re-run: CLEARED** (`sieve:integrity` → `sieve:cleared`). Both demanded tests now exist on the stack top (head `82552fe34d`): `sort_batch_by_slot` is a verbatim extraction of the inline key, tested through the same function the production call site uses (fails `[3,1,2] ≠ [1,2,3]` with the sort removed); `examples/fp8_gemm_twin_parity.rs` grades `_k64`/`_p4` against the real base kernel bit-equal, with a 1-ULP negative control, gated `gpu-examples` per the ~28 existing microtests. Honest residual, recorded: **the oracle's verdict is still unobserved on GPU** — running it is now a mandatory campaign-prep step before the twin-selector claim is trusted. This PR re-enters certification with stack #885.

TheTom:
Read the two fix commits (`538ca2112`, `7f51e9456`) against every thread. All thirteen resolve as described; the three blockers specifically:

- MTP path: `VerifyBatchedOpts` default in `verify_k4_batch_step.rs`, request set only in `verify_dflash_batch_step.rs`, carried on `ForwardContext::gdn_write_on_accept` and into the graph-key sentinel. `woa_decision` has `requested` as its first term and `unrequested_never_engages` pins n=2..4.
- Stale engaged word: `woa_clear_at_entry` runs before the first decline, inside the capture; the post-fold clear is gone, so a verify error or short result cannot leak a stash.
- Stale tables: `gdn_woa_eligible` is cleared at the top of `decode_verify_batched_dispatch` and set at the end only when `wy_tables_base` was non-null; the fold consumes it.

Also checked: `graph_borrow.rs:208` compares the whole sentinel word, so a woa capture is never borrowed by a non-woa request. The SASS answer settles the contraction question (FFMA=0 everywhere), and the stale folded-slot entry as the cause of the moved stream matches what the oracle and the woa-off A/B show. Thanks for the retractions on 187.7 and 269.7; 180.7 and 279.5 on the parent's stream are the numbers.

One thing outside this PR: stack #877 was composed at 17:27Z, six minutes before `7f51e945`, so it still carries `b0b82e64`. Flagged there.

rrstesiak:
Confirming TheTom's flag on #877: the composed head predates both review-fix pushes. #844 is 7f51e945 (pushed 1:33 PM ET) and #845 is 5213834a (4:09 PM ET), both before the 5:36 PM freeze; TheTom's read at 6:45 PM confirms all thirteen threads resolve on those commits. Please recut the stack from those two heads.

Two notes for the recut: the fp8 parity oracle on the stack top grades `_p4`, which 538ca211 removed at review, so that arm has to go; and `ssm_gdn_b.rs` is already at 457 lines on 7f51e945 via `ops/ssm_gdn_woa.rs`, so the wy4-seam split is no longer needed.

I am not pushing anything further to this branch.

One request on the landing: merge #877 as a rebase or merge commit rather than a squash, so the constituent commits land on main with their authors, as the stack branch already carries them. #797 and #818 both landed as squashes under a single author with the rest of us as Co-authored-by trailers.

For a certification stack of eight PRs from three authors, the commit history should say who wrote what.

tbraun96:
## Heads-up: `977f6b4d9b` bisected as the first bad commit in a concurrency regression

While certifying stack #877 (which carries this PR), the `concurrency-sweep` gate failed and an 8-leg bisect isolated it to **`977f6b4d9b` — "gdn: write-on-accept for the batched K=4 verify"**.

| tree | C4 | C8 | verdict |
|---|---|---|---|
| `main` | 43.7 | 64.6 | PASS |
| your `d8778c441d` (prompt-hiddens) | 43.7 | 57.1 | PASS |
| **your `977f6b4d9b` (write-on-accept)** | — | **198 of 320 tok delivered** | **FAIL** |
| your `c00ce83562` (slot-order) | 79/320 | 132/320 | FAIL |

The gate's message: *"a request delivered under 80% of the 320-token budget — this cell's tok/s is NOT comparable"*. It refuses to score rather than print a flattering throughput over tokens that were never delivered.

**The shape fits the commit's own note.** It records `C=16 prose 187.7, code 269.7` — measured and tuned at C=16. **C16/C32/C64/C128 pass in every leg**; the cells that break are **C4 and C8**, below the concurrency it was optimised for.

**Your other three commits are clean by this ladder** — `d8778c441d` passes, and #837/#838 underneath it pass. So this is one commit, not the PR.

**Accuracy is unaffected**: `bfcl-subset` 84.22/84.12 (n=995) and `bfcl-subset-echolp` 86.55/86.95 both clear their bars on the full stack, and `agentic-webserver` is 10/10. This reads as admission/scheduling, not numerics.

Worth noting the wave-1 static sieve flagged this cluster before any GPU ran, for having **no in-tree test** — its evidence was hardware receipts. A slot-ordering unit test was added since and passes; ordering correctness and throughput-under-concurrency turn out to be different properties.

Raising it here rather than deciding for you. Happy to re-run the sweep against any fix you want to try — the harness and the bisect worktrees are still set up.

rrstesiak:
Thanks for the bisect and the receipts. Two things.

**The bisected tree is not the tree in this PR.** `977f6b4d9b` and `c00ce83562` are the Sep 3 originals. #844's head is `7f51e945`, which carries `538ca211`: engaged word cleared at the top of every requesting batched verify, fold gated on the tables staged by this step's verify, folded slots cleared unconditionally at entry. Those were the three blockers Turney and Richard raised on this PR. A partially filled batch at C4 and C8 against the 16-slab table is exactly the case they close: stale stash rows and a leaked engaged word have nowhere to hide at C16 and above, where every slab is live. #877 at `f3f77c82` carries neither `538ca211` nor `7f51e945`.

**Ask:** re-cut #877 on the current PR heads (`7f51e945` for #844, `5213834a` for #845) and re-run the concurrency sweep there.

rrstesiak:
Rebased onto main `08b44b70`; head is now `82b7c2fd`. 
Nine commits replay clean. One added: `dense_ffn_native_batch_tests.rs` and `dense_ffn_kernel_tests.rs` (main, #934) build a `ForwardContext` by struct literal and predate the `gdn_write_on_accept` request flag; they now set it false like every other constructor on the branch. 
No runtime code changed; records stay keyed to `7f51e945` (same tree content on the perf path).

Ask, unchanged: recut L6 on `82b7c2fd` and L7 on `6d3f64a9`; both now sit on current main.

rrstesiak:
Rebased onto main `ca3c2028`; head is now `28420f07`. Ten commits replay; one resolution in `impl_a1.rs`: main's A59 fix (the proposer's reachable context, GLM-5.3's DSA cap) and this PR's DFlash ctx cap both bound the drafter-prefill capture buffer, so the allocation now takes the minimum of the two and `mtp_prefill_capacity` is that same value. `traits.rs` re-exports both `EpCommandFailed` and `VerifyBatchedOpts`. `cargo check -p spark-server` clean.

tbraun96:
Closing: this content landed today with **#1145**.

**Carrier:** `ce2bccd736` — *"follow-ups: the WOA review fixes, the Windows INFINITY defect, and two re-cuts (#1145)"*, merged 2026-09-19T10:42Z. #1145 re-cut these six review-fix commits onto current paths, which is why they arrive under different identifiers than this branch uses.

**Full-population check** — every added line of ≥30 chars, not a sample:

| PR | absent from `main` | of | |
|---|---|---|---|
| #844 | 28 | 1118 | **2%** |
| #845 | 51 | 1530 | **3%** |

And the residue is the re-cut, not missing work. Of #844's 28, **18 are pre-rebrand `ATLAS_*` env names**; of #845's 51, **37 are**. `main` carries each as `AVAROK_*`. The remainder resolve the same way once you look:

- `AtlasCudaBackend` → `AvarokCudaBackend` — **0 files** on `main` carry the old name, **134** carry the new one
- `by_slot.sort_by_key(…)` → present at `crates/spark-server/src/scheduler/mtp_step.rs:717`, with `slot_of(i)` in place of the inline accessor
- `gdn_woa_enabled()` → present at `crates/spark-model/src/layers/qwen3_ssm/gdn_flags.rs:291`, now `pub` rather than `pub(super)`
- the rest are comments reworded during the re-cut, e.g. *"measured neutral on GB10 and dropped in review of #844"*

So nothing here is owed. Reopening restores it if this reading is wrong.

PR #845, opened 2026-09-03
feat(scheduler): DFlash gamma resolver, the serve picks the verify width itself (no gamma flag)

rrstesiak:
## Where this leaves the tree

Best trio-grade numbers measured per cell. One DGX Spark GB10, Qwen3.8-27B NVFP4, temperature 0, median of 3, same prompts per column (MinHeap / Volvo at C=1; SinglyLinkedList / Volvo standard at C=16). The four PRs are the difference between the two DFlash-2 rows; this one is the resolver that lands the bottom row with no gamma flag on the command line. #956 tracks landing them.

| path | C=1 code | C=1 prose | C=16 code | C=16 prose |
|---|---:|---:|---:|---:|
| serial, no speculation | 16.2 | 16.2 | 129.3 | 132.3 |
| MTP (main) | 27.0 | 24.4 | 163.2 | 150.4 |
| DFlash-2 (main, stock defaults) | 66.1 | 24.5 | 173.8 | 83.8 |
| DFlash-2 + #837 #838 #844 #845 | **66.9** | **27.3** | **279.5** | **182.5** |
| DFlash-2 with the line vs MTP | 2.5x | +12% | +71% | +21% |

Main rows: the 09-01 three-arm matrix on stock `main` at `ef8ca11d`, same box. PR row receipts are in the tables below and on #956.

---

Stacked on #838 (which stacks on #837): until those land, the diff of this PR shows their two commits underneath this one. Independent of the C=16 bundle PR, but the win at concurrency two and above lands on the write-on-accept verify from that PR.

This is a paradigm shift in how Atlas serves an LLM. Gone are the days of error-prone
launch recipes and a switch for every workload: this feature makes them obsolete. With it,
DFlash-2 speculative decoding chooses its own optimal parameters for prose and for code,
and for the many prompt shapes in between, and it tunes itself for scale. Whether a server
carries one user or sixteen concurrent prompts, the serve now lands on the fastest
configuration on its own, measured here up to C=16, with no gamma flag on the command line.

The fastest gamma is not one number. On
Qwen3.8-27B NVFP4 + DFlash-2 (block 8) it is K=4 at any concurrency above one, because
the K=4 write-on-accept verify reads and writes the state once per step; and at C=1 it
is the full block for code (accepts ~7.6 of 9 drafts, first-draft accept p1 ~0.97) but
K=5 for prose (~1.7 of 9, p1 ~0.76). Until now that was three launch recipes. This makes
it one default serve with no `--dflash-gamma` at all.

Category: feature / serving behavior. No kernel changes and no new fast path: the serve
selects, per step, among widths that already exist, from concurrency and the accept signal.

## What this does

The serve picks the verify width per step. Three rungs:

| when | width | why |
|---|---|---|
| C >= 2 | K=4 | the write-on-accept batched verify (#844) exists only at K=4 and wins both workloads |
| C = 1, code | K=cap (10 here) | code accepts ~7.6 of 9 drafts; the full block pays |
| C = 1, prose | K=5 | prose accepts ~1.7 of 9; a narrow block is 27.0 vs 24.6 tok/s |

**The C=1 signal** is the first-draft accept bit, one per verify step, in a 64-step shift register:

| parameter | value | meaning |
|---|---|---|
| window | 64 steps | the register; decisions read its popcount |
| enter | 60 of 64 hits | go wide (code on the narrow rung hits ~0.97) |
| leave | 46 of 64 hits | go narrow (code on the wide rung hits ~0.875, prose 0.56..0.76) |
| dwell | 64 steps | minimum between two switches |
| start | wide | the first window after launch runs at the cap |

Each is an env override (`ATLAS_DFLASH_RUNG_MULTI`, `_NARROW`, `_WIDE`, `_ENTER`, `_LEAVE`, `_DWELL`), read once at serve start.

Sizing: steady prose (0.76) flips up 0.11 times per 2000 steps, steady code 0.01, and a real transition flips once, ~65 steps after it. A 10-step EWMA and a 32-bit register were both tried first and flapped inside single requests. The leave line was first set at 55 and sat on wide-rung code's own hit rate (0.875 = 56/64); it dropped MinHeap requests to K=5 inside their docstring blocks. At 46 that happens in ~1.3% of code requests while prose still leaves within the request.

**Where the width lives.**
- Drafter: `gamma` stays the widest block the head was sized for (flag, or trained block + 2). Each propose runs `block_g`, set from the scheduler's `num_drafts` and clamped to `2..=gamma`. Buffers are allocated from `gamma`; every forward-path read goes through `block_g`. Propose graphs are keyed by width, so after each width's first capture a switch costs nothing.
- Scheduler: `dflash_rung.rs` (new). `drafts_for(n_active, num_drafts)` sets the serve-wide count at the top of `step_mtp`; every propose and re-propose inherits it, and the verify width follows on the next step. `observe_step` feeds the register.

**Pinning.** An explicit `--dflash-gamma` pins the width (records are fixed configurations with reproduction keys). `ATLAS_DFLASH_STATIC_GAMMA` (presence) pins; `ATLAS_DFLASH_GAMMA_RESOLVER` (presence) arms the resolver under an explicit flag, which then acts as the cap. With `ATLAS_NO_GDN_WOA=1` the C>=2 rung falls back to the cap. One INFO line at load states the ladder; one per C=1 switch.

**Tests** (`dflash_rung_tests.rs`): the ladder lands every measured winner, widths never exceed the cap, the hysteresis band separates the measured hit rates on both rungs, the pinned serve returns `num_drafts` unchanged, the no-woa fallback, the dwell, and a simulation that drives the real controller through steady prose, steady code on both rungs, and both transitions.

## Receipts (one default serve, `launch-dflash-resolver-c16.sh`, no gamma flag; C=16 rows n=1 on this head, trios to follow)
| workload | C | resolver | fixed-gamma reference |
|---|---|---|---|
| MinHeap (code) | 1 | 66.8 | 66.8 trio at gamma 10 (66.6 crown class) |
| Volvo (prose) | 1 | 27.7 (first-revision build); 26.0 then 27.3 on this head, one flip to K=5 at tick 64 | 27.0 at gamma 5 |
| MinHeap (code, 300 tok) | 1, code after code | 70.4 / 70.3 / 70.4 (the file's high-acceptance head; not the 1500-tok reference) | |
| MinHeap (code, 300 tok) | 1, after prose | 53.8: ~48 of 58 steps on the narrow rung before the enter line clears. The prose-to-code entry lag is the open transitional cost; tracked as a follow-up, not in this PR | |
| MinHeap (code) | 1, after prose | 59.0 / 66.9 / 66.9 (LEAVE 46; first run = entry lag from prose, one flip up, none down; at LEAVE 55 both runs dropped narrow mid-request: 58.9 / 57.9) | 66.8 trio at gamma 10 |
| MinHeap (code) | 16 | 278.1, 15/16 on 16383c26 | 279.5 trio at gamma 4 (#844 fixed tree), same stream |
| Volvo (prose) | 16 | 179.9, 15/16 on 42d547eb | 180.7 trio at gamma 4 (#844 fixed tree), same stream |

At C=16 the resolver reproduces the fixed-gamma-4 streams of the #844 head it is stacked on
(the width is decided by concurrency alone; no accept signal is involved). The C=16 rows
were re-measured on this head after the #844 review fixes; the first revision's rows
(267.1 / 184.5) were on the pre-fix tree and are superseded, as are the 269.7 / 187.7
references (see the #844 body for the retraction). On the EWMA build, Volvo C=1 gave
27.0 / 27.2 / 27.2 with runs 2 and 3 byte-identical to the fixed-gamma-5 stream fe0f9d2c.

## Benchmark prompts, exact text

The C=16 rows above were measured with two prompts written for the concurrency sweep on 2026-08-29, not the Atlas standard MinHeap / Volvo prompts their labels suggest. Re-measured on the standard prompts, same tree and lane, the numbers land in the same class. Both sets stay on the record with their own prompt, marked by superscript.

| workload | C=16, sweep prompt (published) | C=16, Atlas standard prompt | stream (standard) |
|---|---|---|---|
| code | 279.5 median of 3 (279.5 / 279.8 / 277.6) <sup>1</sup> | 268.8 median of 3 (267.7 / 268.8 / 269.3) <sup>3</sup> | 3ae47861 on 15 / 14 / 15 of 16 |
| prose | 180.7 median of 3 (180.7 / 180.5 / 181.6) <sup>2</sup> | 182.5 median of 3 (181.7 / 183.0 / 182.5) <sup>4</sup> | 24eb6e04 on 15 / 14 / 15 of 16 |

All four prompts ran at the same `max_tokens: 300`, 16 concurrent requests (4800 completion tokens per run), temperature 0, `stream: false`; aggregate = total completion tokens over wall time. C=1 rows elsewhere in this PR: MinHeap at `max_tokens: 1500`, Volvo at `max_tokens: 300`. The published streams (16383c26, 42d547eb) belong to prompts 1 and 2; a different prompt is a different stream, so the two columns are not stream-comparable, only class-comparable.

<sup>1</sup> Code, sweep prompt, `max_tokens: 300` (receipts `csweep_*`):
> Write a Python class SinglyLinkedList with methods insert_front, insert_back, delete_value, search, and reverse. Include a __repr__ and a brief docstring for the class.

<sup>2</sup> Prose, sweep prompt, `max_tokens: 300` (receipts `csweepP_*`):
> Write a 200-word essay on the Volvo P1800.

<sup>3</sup> Code, Atlas standard prompt, `max_tokens: 300` at C=16 (receipts `csweepM_*`); the same prompt at `max_tokens: 1500` is every C=1 MinHeap row in this PR and on the record:
> Write a complete Python MinHeap class implementing push, pop, peek, and heapify, with sift_up and sift_down helpers and a docstring for each method.

<sup>4</sup> Prose, Atlas standard prompt, `max_tokens: 300` (receipts `csweepV_*`); C=1 Volvo rows dated before 2026-09-04 used prompt 2 at the same length:
> Write a detailed history of the Volvo car company.

Standard-prompt receipts: `csweepM_dflash_c16_20260904_155119 / 155228 / 155307`, `csweepV_dflash_c16_20260904_155352 / 155436 / 155519`, resolver head with the K=4 lane selected by concurrency (identical to `--dflash-gamma 4` on #844).

## On the record
A resolved single-stream run is NOT reproducible by any fixed-gamma run once a switch
lands mid-stream (run 1 above opens on the gamma-10 stream and finishes on the gamma-5
one). Records stay per fixed configuration with reproduction keys; the resolver is the
serve default that lands on them.

## Notes for reviewers
- The head is shared, so `block_g` is an atomic; the scheduler proposes one step at a
  time on one stream, so there is never a concurrent writer.
- Batched propose (two or more sequences) was already eager per width; the change there is only that
  `rows_total = block_g * n`. The C=16 gate above is the bit-exactness proof.
- A width change between steps makes one step's batch non-uniform; that step's odd
  sequence takes the serial verify and the next re-propose realigns it.
- Branch `pr/dflash-gamma-resolver` on dflash-wyn-table @110c2c53; new files carry the
  provenance-id. Depends on nothing in the other four PRs, but the win at concurrency two and above needs the
  write-on-accept K=4 verify to be present to be the win it is.

## Test plan
- [x] `cargo fmt --all -- --check` (touched files clean; two pre-existing diffs on the base belong to #838 and are fixed there)
- [x] `cargo clippy -p spark-model -p spark-server --tests -- -Dwarnings` clean (base clippy fix folded into #838)
- [x] `cargo check -p spark-server` on the branch
- [x] Tested on hardware: Qwen3.8-27B NVFP4 + DFlash-2 on one DGX Spark, receipts above with output shas
- [x] Tests added or updated where applicable
- [ ] Benchmark certification: ten-gate campaign on the frozen sha, records committed (serialized-landing protocol, `docs/gate-queue-protocol.md`)

## CLA


TheTom:
Read the resolver commit above #838 (static, no hardware on my side today). Three questions before findings:

1. `forward_block.rs:1113-1126`: the SAFETY comment now says `self.block_g() == gamma_val` and that `self.block_g() * 4` is exactly the pinned allocation. After this change `block_g <= gamma_val`, so the read is at most the allocation, which is fine, but the invariant as written is no longer true. Was the `self.gamma` -> `self.block_g()` rewrite inside the safety argument intentional, or a search-and-replace that caught the comment?

2. The resolver rows in the receipt table (66.8, 27.7, 267.1, 184.5) read as single runs, while the fixed-gamma references are trios. Is a trio on the resolver arm coming before the default flips?

3. With `ATLAS_NO_GDN_WOA=1` (the #844 kill switch) the resolver still picks K=4 at C>=2, which is then the parent `wy4` kernel with no receipt in either PR. Should `MULTI_K` fall back to the cap when that switch is set, or is that configuration out of scope?

crates/spark-model/src/layers/dflash_head/forward_block.rs
```diff
@@ -1108,20 +1110,20 @@ impl BlockDiffusionDraftHead {
...
         anyhow::ensure!(
             !pinned_ptr.is_null(),
             "DFlash draft-token pinned staging buffer is null (γ={}, rows={rows_total})",
-            self.gamma
+            self.block_g()
         );
         // SAFETY: `pinned_ptr` is the page-locked allocation made by
         // `alloc_host_pinned(gamma_val * 4)` in `DFlashHead::from_weights`, and
-        // `self.gamma == gamma_val` (both set from the same local in that
+        // `self.block_g() == gamma_val` (both set from the same local in that
```
TheTom: **Should fix.** This SAFETY comment states `self.block_g() == gamma_val` and "`self.block_g() * 4` is exactly the allocation size", neither of which holds after this PR: the pinned buffer is `gamma_val * 4` bytes, `block_g` is `2..=gamma`, and the read is `block_g * 4`, which is at most the allocation. The code is safe; the argument written next to the `unsafe` is not the one that makes it safe. Rewrite it as: buffer is `gamma * 4`, `block_g <= gamma` by `set_block_g`'s clamp, read length is `block_g * 4`. Same for line 1113 and the historical sentence in the test doc at 1243-1244, which now says the guard and the kernel both used `self.block_g()` and no longer describes the bug it records.

crates/spark-model/src/layers/dflash_head/forward_block.rs
```diff
@@ -106,7 +106,7 @@ impl BlockDiffusionDraftHead {
         // every row-bounded op take, so leaving it at one band's worth means
         // only band 0 gets embedded and every other sequence drafts from
         // whatever the previous propose left in stream_buf.
-        let n_attn = (eff_ctx + self.gamma * n_seq) as u32;
+        let n_attn = (eff_ctx + self.block_g() * n_seq) as u32;
```
TheTom: **Should fix.** Line 58 reads the width once into `block_g`, then this function calls `self.block_g()` another 29 times (through line 1205, including inside the `flat_map` closures at 324 and 361). Every read sees the same value today because the scheduler is the only writer and it writes before the propose, but the width is now a mutable atomic, and the row math in one forward should not be able to disagree with itself. The pre-PR code already had the local; read once at entry and pass it down.

crates/spark-server/src/scheduler/dflash_rung_tests.rs
```diff
@@ -0,0 +1,143 @@
...
+            x = x
+                .wrapping_mul(6364136223846793005)
+                .wrapping_add(1442695040888963407);
+            ((x >> 11) as f64 / (1u64 << 53) as f64) < p
+        })
+        .collect()
+}
+
+/// Run the pure rule over a stream from a given state; returns (flips, final).
+fn simulate(start_wide: bool, hits: &[bool]) -> (u32, bool) {
```
TheTom: **Test gap.** `simulate` re-implements the dwell and decision loop instead of driving `observe_step`, so the production controller (`tick`, `saturating_sub`, the env-read `DWELL`, the order of the `last_switch` update) is never executed by any test; a divergence between this loop and `observe_step` is invisible. `adaptive_rung_tests.rs:108-172` already shows the shape: `configure`, then feed `drafts_for`/`observe_step` over ticks and assert the width. The pure-rule tests above are good; this one should run the real function.

crates/spark-server/src/scheduler/dflash_rung_tests.rs
```diff
@@ -0,0 +1,143 @@
...
+// provenance-id: 526f6e616c6420522e205374657369616b
+
+//! Unit tests for the DFlash gamma resolver. The 2026-09-03 receipts are the
+//! anchor: the rule must land every measured winner. Pure functions only; the
+//! shared controller is exercised through `configure` + `drafts_for` once.
+
+use super::*;
+
+#[test]
+fn concurrency_ladder_lands_the_records() {
```
TheTom, line 22: **Nit.** These expectations go through `k_for`, which reads `ATLAS_DFLASH_RUNG_MULTI/NARROW/WIDE` from the environment on every call. A runner with any of those set changes the expected values without touching the code. Either clear the three variables at the top of the test or test the rule with the constants passed in.

crates/spark-server/src/scheduler/dflash_rung.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+struct Ctl {
+    /// Widest block the head was sized for (K). 0 = not configured (no DFlash).
+    cap: AtomicUsize,
+    /// Resolver active. False = every call returns `num_drafts` unchanged.
+    armed: AtomicBool,
+    /// C=1 state: `true` = wide (code).
+    wide: AtomicBool,
+    /// Shift register: bit i = first draft accepted, i steps ago (low 32 bits).
```
TheTom: **Nit.** "(low 32 bits)" is left over from the 32-bit register the body says was tried first; `WINDOW` is 64 and `shift_in` masks the whole u64.

crates/spark-server/src/scheduler/dflash_rung.rs
```diff
@@ -0,0 +1,241 @@
...
+fn narrow_k(cap: usize) -> usize {
+    env_usize("ATLAS_DFLASH_RUNG_NARROW", NARROW_K).clamp(2, cap)
+}
+fn wide_k(cap: usize) -> usize {
+    env_usize("ATLAS_DFLASH_RUNG_WIDE", cap).clamp(2, cap)
+}
+
+/// The width rule, pure: verify K for `n_active` sequences given the cap and
+/// the single-stream state. Unit-tested without the GPU.
+pub fn k_for(n_active: usize, cap: usize, wide: bool) -> usize {
```
TheTom: **Nit.** `k_for` reads three environment variables on every `drafts_for` (every scheduler step), and `next_wide` plus `observe_step` read three more on every C=1 verify step. `configure` already exists and runs once; resolving the six overrides there into atomics keeps the per-step path to plain loads. `verify_gdn_batch_enabled` uses a `OnceLock` for the same reason.

crates/spark-server/src/scheduler/mtp_step.rs
```diff
@@ -26,6 +26,16 @@ pub fn step_mtp(
     // bootstrap, D-Cut plan, chunk sort) — one component of the out-of-step
     // GAP. One Instant::now() when disarmed, same cost note as StepTimer.
     let t_step_outer = std::time::Instant::now();
+    // DFlash GAMMA RESOLVER: this step's draft count from the current
+    // concurrency (and, single-stream, the accept signal). Shadows the
+    // serve-wide value so every propose and re-propose below inherits it;
+    // the verify width follows `pending_drafts.len()` on the next step.
+    // Identity when pinned (`--dflash-gamma`) or not DFlash.
+    let num_drafts = if dflash_verify_raw_argmax {
+        crate::scheduler::dflash_rung::drafts_for(active.len(), num_drafts)
```
TheTom, line 43: **Note.** `active.len()` counts every active sequence, including ones in bootstrap or prefill this step. A lone generating sequence gets K=4 on a step where a second request is prefilling, and the C=1 observer skips that step (`last_n != 1`). Correct for the batched verify that follows once both draft, mildly pessimistic for the lone sequence in the meantime. Worth a line in the module doc so nobody reads the C=1 arm as "one generating sequence".

tbraun96:

rrstesiak:
Replies to every thread above, one block per thread in review order. Head is now 5213834a (a7ece678 + one follow-up commit), rebased onto #844's review fixes (7f51e945); #845 stays stacked on #844 and lands after it.

---
## Top-level: the three questions (TheTom, 22:11)

1. Search-and-replace caught the comment. The code was safe by a different argument than the one written next to the `unsafe`, which is worse than an unsafe block with no comment. Rewritten as you laid it out: the pinned buffer is `gamma * 4`, `width <= gamma` by the clamp in `set_block_g`, the read is `width * 4`, so at most the allocation. The sentence at 1113 and the test doc at 1243 are corrected too; the test doc now records the regression as it was instead of describing today's code.

2. Agreed. The first revision's resolver rows were single runs on the day it went live, labeled as such in the ledger. Receipts on this head (5213834a on 7f51e945), one default serve, no gamma flag:

   **C=16, resolver on:** prose 179.9 on stream 42d547eb (15/16), code 278.1 on 16383c26 (15/16). Same streams and same class as #844's hand-set gamma-4 trios (180.7 / 279.5), so the resolver picks the K=4 lane on its own. n=1 each; trios follow on the campaign tree.

   **C=1 prose (Volvo):** 26.0, then 27.3 back to back (fixed-gamma-5 trio: 27.0 / 27.2 / 27.2). The log shows one flip, `C=1 -> K=5 (narrow/prose; hits=41/64 tick=64 flips=1)`, at the end of the first run's first window. The first run paid that window on the wide rung; the second ran narrow throughout.

   **C=1 code (MinHeap) found a sizing bug, fixed in 5213834a.** The register's signal is not width-independent: code hits the first draft 0.97 of the time on the narrow rung but 0.875 (56/64) on the wide rung. LEAVE=55 sat on that mean, so both MinHeap requests dropped to K=5 inside their docstring blocks (58.9 / 57.9; log `hits=55/64`).

   **The fix:** LEAVE=46. That is 3.8 sigma under wide-rung code (a false drop in ~1.3% of code requests) and inside prose (at 0.76, prose leaves within the request 98% of the time). The module doc now carries the rung-dependent hit rates and the cost asymmetry (code on the narrow rung -18%, prose on the wide rung -8%). A wide-rung 0.875 stream is in the steady-state test.

   **Re-measured after the fix,** same serve, MinHeap 1500 tok: 59.0 / 66.9 / 66.9. The first run is the prose-to-code entry lag (register arrives full of Volvo misses; one flip up at hits=60/64, none back). The next two are the 66.8 crown class with no further flip lines. At LEAVE=55 this morning, the same second run gave 57.9.

3. Fallback to the cap, now implemented. The C>=2 rung is K=4 because of the write-on-accept kernel and for no other reason, and a K=4 parent-kernel serve has no receipt in either PR. `configure` takes `woa_available` (from `gdn_woa_enabled`, now public) and `Rungs::from_env` pins `multi` at the cap when it is false; the log line says so at serve start. Test `no_woa_pins_the_multi_rung_at_the_cap`.

---
## Inline threads

**`forward_block.rs:1124` (SAFETY comment):** see question 1.

**`forward_block.rs:109` (29 reads of a mutable atomic):** read once at entry into `width`; every row count in the forward derives from that local, including the two closures. The atomic is read nowhere else in the function.

**`dflash_rung_tests.rs:68` (simulate re-implements the loop):** `simulate` now drives the real controller: `configure_with`, then per step `drafts_for(1, 9)` followed by `observe_step`, the scheduler's order; flips come from the controller's own counter and the state from what `drafts_for` hands back. Two more tests on the real path: the dwell holds for one full window after a switch even at 64/64, and C>=2 dispatches are not samples. Serialized on one lock, since the controller is one static.

**`dflash_rung_tests.rs:11` (env reads in tests):** the rules are pure over an explicit `Rungs` now, and the tests pass `Rungs::defaults(10)`; no test reads the environment.

**`dflash_rung.rs:92` ("low 32 bits"):** fixed.

**`dflash_rung.rs:167` (env reads per step):** the six overrides are resolved once in `configure` into atomics; `drafts_for` and `observe_step` are plain loads.

**`mtp_step.rs:35` (`active.len()` semantics):** the note is in the comment at the call: the C=1 arm means one sequence on the server, not one sequence generating; a lone generator takes the C>=2 width while a second request prefills and the observer skips that step.

tbraun96:

|---|---|---|---|---|
| 7 | #845 | DFlash gamma resolver (serve picks verify width) | dflash | **← this PR** |

crates/spark-server/src/scheduler/dflash_rung.rs
```diff
@@ -0,0 +1,350 @@
...
+    pub leave: u32,
+    /// Minimum verify steps between two switches.
+    pub dwell: u64,
+}
+
+impl Rungs {
+    /// The measured defaults for a head whose widest verify width is `cap`.
+    pub fn defaults(cap: usize) -> Self {
+        Self {
+            multi: MULTI_K.clamp(2, cap),
```
TheTom: **Nit.** `clamp(2, cap)` panics when `cap < 2` (Rust asserts `min <= max`), and `defaults` runs from `configure` before the `cap_k >= 3` arming check gets a say. `from_env` already guards with `cap.max(2)`; the same guard here (and on `narrow`) keeps a `--dflash-gamma 1` serve from panicking at load instead of pinning.

TheTom (commented):
Read `a7ece678` and `5213834a` against every thread. All eight resolve: SAFETY comment states the real invariant, `forward_block.rs` reads the width once, `simulate` drives `configure_with` + `drafts_for` + `observe_step` in the scheduler's order, the rules take `&Rungs` so no test reads the environment, the six overrides resolve once in `configure`, and the woa-off fallback pins the C>=2 rung at the cap with a test.

The LEAVE finding is the useful outcome of the trio request: a rung-dependent first-draft rate (0.97 narrow, 0.875 wide) is not something the simulation could have shown, and 46 at 3.8 sigma with the asymmetric costs written into the module doc is the right shape. One nit inline.

Same note as on #844: stack #877 was composed before this head and carries `e1a16f0f`, so the LEAVE=55 drop is still in that tree.

crates/spark-server/src/scheduler/dflash_rung.rs
rrstesiak: Fixed in 18ba60ef: `defaults` guards `multi` and `narrow` with `cap.max(2)`, the same guard `from_env` applies; a `--dflash-gamma 1` serve pins at width 2 at load. Test `defaults_below_cap_two_pin_instead_of_panicking` asserts caps 0 and 1 return (2, 2, 2).

rrstesiak:
Rebased onto #844's new head `82b7c2fd`; head is now `6d3f64a9` (carries the `defaults` guard from this morning). Four commits replay clean; `cargo check -p spark-server --tests` clean on this head.

rrstesiak:
Rebased onto #844's new head `28420f07` (main `ca3c2028`); head is now `14ad227d`. Four commits replay clean; `cargo check -p spark-server` clean.

tbraun96:
Closing: this content landed today with **#1145**.

**Carrier:** `ce2bccd736` — *"follow-ups: the WOA review fixes, the Windows INFINITY defect, and two re-cuts (#1145)"*, merged 2026-09-19T10:42Z. #1145 re-cut these six review-fix commits onto current paths, which is why they arrive under different identifiers than this branch uses.

**Full-population check** — every added line of ≥30 chars, not a sample:

| PR | absent from `main` | of | |
|---|---|---|---|
| #844 | 28 | 1118 | **2%** |
| #845 | 51 | 1530 | **3%** |

And the residue is the re-cut, not missing work. Of #844's 28, **18 are pre-rebrand `ATLAS_*` env names**; of #845's 51, **37 are**. `main` carries each as `AVAROK_*`. The remainder resolve the same way once you look:

- `AtlasCudaBackend` → `AvarokCudaBackend` — **0 files** on `main` carry the old name, **134** carry the new one
- `by_slot.sort_by_key(…)` → present at `crates/spark-server/src/scheduler/mtp_step.rs:717`, with `slot_of(i)` in place of the inline accessor
- `gdn_woa_enabled()` → present at `crates/spark-model/src/layers/qwen3_ssm/gdn_flags.rs:291`, now `pub` rather than `pub(super)`
- the rest are comments reworded during the re-cut, e.g. *"measured neutral on GB10 and dropped in review of #844"*

So nothing here is owed. Reopening restores it if this reading is wrong.

PR #856, opened 2026-09-04
Document the three bot commands that existed only in a workflow file

tbraun96:
Follow-on to #847. Three waves: one real defect, two empty and recorded as empty.

## The defect

The bot accepts five comment commands. Three of them appeared **nowhere** in the README, and the fourth was documented wrongly:

| command | accepted | in README before this |
|---|---|---|
| `/help` | yes | **no** |
| `/stamp` | yes | yes, but **wrong** — said "anyone with write access", omitting the PR-author rule |
| `/seal` | yes | yes |
| `/review` | yes | **no** |
| `/expedite` | yes | **no** |

`/expedite` is the one that matters: it **skips certification** and lets a PR merge on the pipeline's own checks — an administrative override discoverable only by reading a workflow file.

Every one of these gaps was introduced by #847's own commits, one at a time, and no single change looked like it was leaving something out. That is how documentation drift actually happens: not neglect, but a sequence of individually reasonable edits.

Fixed: the `/stamp` line now states the author rule and why it exists, and all five commands are documented in a table giving who may use each and what survives a new commit.

## The guard

`assert-commands-documented.py` reads the handler's own `case` arm rather than restating the verb list — a second hand-maintained list would drift from the first exactly the way the README drifted from the handler. Run against the README as it stood, it names `/help`, `/review` and `/expedite` and refuses.

Three controls. A complete README passes; a missing verb is caught; and **if the `case` arm is renamed so the guard cannot find its input, it refuses rather than finding nothing and reporting success.** Shipping that failure mode inside the guard built to catch it would have been a poor joke.

Hosted in `SPDX license headers` — a required context, and already a repo-wide file lint.

## Two empty waves, recorded as empty

Recorded so the same ground is not broken again:

| dig | scale | result |
|---|---|---|
| install URL: canary vs what the website tells users | 2 URLs | agree |
| install endpoints live | `.sh`, `.ps1` | 200, correct content-types |
| installer is POSIX as its shebang claims | 606 lines | `dash -n` clean |
| **served installer vs its source in `atlas-recipes`** | 2 files | **byte-identical** |
| committed secrets | 4500 tracked files, 6 patterns | none |
| security contact reachable | `security@avarok.net` | valid MX; private reporting + push protection on |
| orphaned assets >500 KB | whole tree | none |
| every action SHA-pinned | 101 `uses:` | 0 unpinned |
| composite actions' **internal** `uses:` | 5 composites parsed | 0 unpinned |
| mdBook pages vs `SUMMARY.md` | 38 pages | 0 orphaned, 0 missing |
| CODEOWNERS principals exist and hold write | 4 users | all do |
| commented CODEOWNERS rules ignored | both parsers | both refuse correctly |
| issue forms valid | 3 templates | 0 problems |

**One scare, resolved by looking.** Neither `install.sh` nor `install.ps1` is in this repo, which for a `curl | sh` target looked unversioned and unreviewed. Both live in `atlas-recipes/scripts` with a 24 KB test suite, and the served bytes match source exactly.

**One real gap deliberately not closed.** 415 of 600 first-party `unsafe` sites (excluding vendored `cudarc`) carry no SAFETY note. Retrofitting 415 comments would not make one of them more correct, and it would bury the waves that found something. Documented as a known gap for a dedicated review pass.

## A note on method

Three throwaway checkers written during these waves were wrong — `lstrip("./")` stripping the dot from `.github`, site-root URLs resolved against the filesystem, and a commented CODEOWNERS line read as a rule. Two sweeps returned 19 and 1 findings of which 2 and 0 were real. **The finding rate of an unverified sweep is not its defect rate.** Every guard that reached CI did so only after being run against the defect it claims to catch.

---

## Split note

This PR originally carried two Rust fixes as well. They touch `crates/`, which re-opened all 11 benchmark gates, so **the code moved to #866** and this PR is docs/CI only — it touches zero files under `crates/` and its gates stay closed.

`docs/ROBUSTNESS.md` here records waves 26–31, including the two waves whose code is in #866. The record and the diff landing in different PRs is deliberate: holding a finished record behind a 5-hour GPU campaign would be the tail wagging the dog, and the entries say where the code went.

Also added since the original description: wave 29's fix for `/stamp` warning about a benign in-flight CI run — found in production the first time wave 20's fix printed what the API actually said (`403 "This workflow is already running"`, not the missing `actions: write` that had been inferred from a discarded error).

### DIFF
--- a/.github/scripts/assert-commands-documented.py
+++ b/.github/scripts/assert-commands-documented.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: AGPL-3.0-only
+"""Every comment command the bot accepts must be documented in the README.
+
+Three of the five were not. `/help`, `/review` and `/expedite` existed in the
+handler and appeared nowhere a user would look -- `/expedite` in particular is
+an administrative override that skips certification, which is the last command
+that should be discoverable only by reading a workflow file. All three were
+added during this repository's own recent work, which is the point: the drift
+was introduced by the same hands that wrote the docs, one commit at a time, and
+no single change looked like it was leaving something out.
+
+The verb list is read from the handler's own `case` statement rather than
+restated here. A second hand-maintained list would drift from the first exactly
+the way the README drifted from the handler.
+"""
+import re
+import sys
+import pathlib
+
+ROOT = pathlib.Path(__file__).resolve().parents[2]
+HANDLER = ROOT / ".github" / "workflows" / "certification-commands.yml"
+README = ROOT / "README.md"
+
+# the `case` arm that whitelists accepted verbs, e.g. `/help|/stamp|/seal) ;;`
+CASE = re.compile(r"^\s*(/[a-z]+(?:\|/[a-z]+)+)\)\s*;;", re.M)
+
+
+def main() -> None:
+    m = CASE.search(HANDLER.read_text())
+    if not m:
+        print(
+            "REFUSE: could not find the verb whitelist in certification-commands.yml. "
+            "This guard reads the handler's own `case` arm so the two cannot drift; "
+            "if the parser moved, teach this regex rather than hard-coding a list.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+    verbs = m.group(1).split("|")
+    readme = README.read_text()
+    missing = [v for v in verbs if v not in readme]
+
+    if missing:
+        for v in missing:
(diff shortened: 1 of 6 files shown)


tbraun96:
Temporarily removed from the merge queue: the hosted-runner pool has been stalled for hours and the queue was re-dispatching this entry's check set on a timer (~8 runs per timeout cycle, 17 stale sets accumulated). Nothing is wrong with this PR — it is CLEAN and will be re-queued the moment hosted runners complete work again.

Issue #842, opened 2026-09-02
Scheduler thread panics in xgrammar matcher on json_schema request; server silently stops generating


daniellowell:
Title: Scheduler thread panics in xgrammar matcher on json_schema request ("cannot roll back N tokens: only 1 steps recorded") — server keeps accepting requests but never generates again

## Environment

- Image: `avarok/atlas-gb10:latest` (sha256:dd2e31034f052e7f3e35b8c6cd1e435353ba0faea57915aab12499f3ce…)
- `spark --version`: `spark 1.0.0-beta-preview`
- Hardware: NVIDIA DGX Spark (GB10), driver 580.173.02
- Model: `ornith-ai/Ornith-1.5-35B-A3B-NVFP4`
- Launched via sparkrun; serve command:

```
spark serve ornith-ai/Ornith-1.5-35B-A3B-NVFP4 --port 8888 --host 0.0.0.0 \
  --gpu-memory-utilization 0.88 --max-seq-len 131072 --kv-cache-dtype fp8 \
  --kv-high-precision-layers auto --lm-head-dtype bf16 --request-timeout 900 \
  --content-loop-min-repeats 12 --fp8-kv-calibration-tokens 256 --scheduling-policy slai \
  --ssm-cache-slots 128 --ssm-checkpoint-interval 32 --mtp-quantization bf16 \
  --num-drafts 1 --max-thinking-budget 8192 --enable-prefix-caching --speculative
```

## Symptom

Three times in one day the server stopped generating entirely. `/v1/models` still answers and `/v1/chat/completions` still accepts connections and logs the request (`Session …: N prompt tokens`), but no tokens are ever produced and GPU utilization sits at 0%. Clients see an indefinite stall until their own timeout. Only killing and relaunching `spark serve` recovers it.

Each time, the log shows a panic on the scheduler thread immediately before the hang:

```
2026-09-02T13:14:43.212317Z  INFO spark::scheduler::prefill_a_step: Thinking enabled, budget=Some(3686)
2026-09-02T13:14:43.212722Z  INFO spark::scheduler::emit_step: Grammar constrained decoding active: response_format=json_schema
2026-09-02T13:14:43.212725Z  INFO spark::scheduler::prefill_a_step: Chunked prefill start: 265 prompt tokens, chunk_size=265, max_tokens=4096
[...]

Issue #799, opened 2026-08-29
Video request writes OOB in ViT oversized-batch fallback → CUDA 700, poisoned context, server wedged at 503


rsafier:
## Summary

Sending a **video** to a Qwen3.8-27B serve writes out of bounds in the ViT encoder, raising `CUDA_ERROR_ILLEGAL_ADDRESS (700)` and **poisoning the CUDA context**. The process stays alive but every subsequent request fails: the server answers `503` until it is restarted. The context is unrecoverable in-process.

The out-of-bounds write is bounded only by a `debug_assert!`, which is **compiled out of the `--release` builds we serve**.

## Impact

- Any single video request can take down a running server for all tenants.
- Failure mode is a silent OOB write, not a refused request — worst case it corrupts unrelated GPU memory before the fault surfaces.
- Present on **`main`** (`a046cdfd`) — not branch-local. Observed on `wip/dflash2-full` (`694c18ff`); both carry identical code.

## Observed

Stock `serve_dflash_tui.sh` (Qwen3.8-27B-NVFP4, `--vision-max-pixels 262144`, 128K ctx × 8 seqs), one video in a chat request:

```
INFO spark::api::chat::msg_entry: Video (media item 0): 19 temporal groups, 30x34 patches, 4845 vision tokens
INFO spark::api::chat: Session ...: 4898 prompt tokens
INFO ...vision_encoder::enc_impl::init: Vision scratch allocated on first image: 1024 patches
ERROR atlas::fault: cuMemsetD8Async failed (status 700), and a no-op synchronize issued afterwards also failed (CUDA_ERROR_ILLEGAL_ADDRESS (700))
ERROR ...trait_impl::sequence: free_sequence: ssm_pool.zero_slot(0): cuMemsetD8Async failed: status 700
ERROR spark::scheduler::prefill_a_step: free_sequence (after prefill error): cuMemFree_v2 failed: status 700
ERROR spark::scheduler::phase_start_prefills: Prefill start error: cuMemcpyHtoDAsync_v2 failed: status 700
```

All later requests then return `503`.

Note the first *logged* 700 is in `free_sequence` cleanup, which is misleading — CUDA faults surface at the next synchronizing call. The faulting kernel is upstream, in the vision encode.

## Root cause

`buf_out` is sized for a **single image's** merged rows (`enc_impl/init.rs:118` on `main`):

```rust
let buf_out = gpu.alloc(p_max * out_hidden_size * 2)?;
```

[...]

PR #866, opened 2026-09-04
Two bounds asserted in prose: an OOB write (#799) and a scheduler panic (#842)

tbraun96:
Closes nothing on its own — both issues stay open for their second halves, described below.

Two production outages, found in the open issues, with the **same shape**: a bound that existed only as a comment or as an assertion the release build strips.

## #799 — a video request wrote 4.7× past `buf_out` and poisoned the CUDA context

One video request raised `CUDA_ERROR_ILLEGAL_ADDRESS` and poisoned the context. The process **stayed alive** and answered `503` to every later request, for every tenant, until restarted.

The bound existed — as a `debug_assert!`, compiled out of the `--release` binaries we serve. So production had no bound, only a comment claiming one:

> The scheduler caps Σp ≤ p_max so this is normally unreachable — a correctness guard only.

Video defeats that cap: all temporal groups of a clip arrive as **one** media item and encode as one batch, so the per-item cap never sees the sum. 19 groups of 30×34 patches merge to 4845 rows against a 1024-row buffer.

`check_packed_rows` is now a pure free function returning `Result`, mirroring `check_pixel_len` in the sibling file — whose doc comment had *already* observed that `forward_oversized_fallback` "bounds Σ*merged* p and not per-image `p`", and which ends with the line this earns: **prose is not a bound; this is.**

Two more hazards closed on the way: the old expression reached `mp_i.last().unwrap()` whenever `mp_off` was non-empty (a panic, not an error), and the sum now uses `checked_add` so an overflowing offset cannot wrap into a passing comparison.

**Control — and it only means anything in release.** Four tests run under `--release`. Restoring `debug_assert!` makes two of them **fail in release and pass in debug**, which is precisely the blindness that caused the outage and precisely what a debug-only test can never catch.

```
release, ensure!:        4 passed; 0 failed
release, debug_assert!:  2 passed; 2 FAILED
```

## #842 — a `json_schema` request killed the scheduler thread, silently

Three times in one day. The API kept accepting requests and logging sessions, generated nothing ever again, and `/v1/models` still answered — so the server looked healthy while being dead.

```
cannot roll back 96 tokens: only 1 steps recorded
```

**Diagnosed by the codebase disagreeing with itself.** Four callers rewind the grammar matcher. Three pass a *delta* of `num_history_steps()` across the span they undo (`spec_step.rs:462→470`, `verify_pipeline_helper.rs:343→391` and `503→548`). The fourth — the watchdog path — passed a raw count of **sequence** tokens. The crate already documents why that is wrong, on `num_history_steps` itself, as BUG#3: `accept_token` returns `true` for stop/EOS and in the **terminated** state without advancing the matcher. A `json_schema` matcher terminates the moment its object closes, so every token after that records nothing.

The comment above the call asserted the precondition that had stopped holding:

> every dropped token ... was therefore fed to `grammar_state.accept_token`, so `rollback(dropped)` is exact.

`grammar_rewind` is pure over two integers — testable with no model, scheduler or GPU. **`min(dropped, steps)` was considered and rejected**, and that reasoning is pinned in a test so a later "helpful" clamp has to argue with it: with a terminated matcher the recorded steps belong to tokens being **kept**, so a partial rewind corrupts state the panic at least left visible.

**Control is the tempting wrong fix.** Changing refuse to clamp:

```
refuse:  21 passed; 0 failed
clamp:   18 passed; 3 FAILED
```

The fourth test stays green under both — clamping and refusing genuinely agree when `dropped ≤ steps`. A control that reddened all four would have been measuring less, not more.

## Scope, stated plainly

Both are the **outage** half only.

- #799 fix (2) — chunking the encode so long videos *work* — is not done. Video is now a clean per-request error instead of a multi-tenant outage: strictly better, still not "video works".
- #842's accounting is still not exact; that needs the anchor the speculative paths have, captured where the droppable window begins. Constrained output may drift on an affected sequence, which is logged.

Both issues stay open for those halves.

## This PR owes a campaign

It touches `crates/`, so all 11 gates re-opened. Split out of #856 for exactly that reason — the docs and CI work there is certified-clean and shouldn't wait on 5 GPU-hours. The record entries for these two waves are in #856; the code is here.

### DIFF
--- a/crates/spark-model/src/layers/vision_encoder/enc_impl/forward.rs
+++ b/crates/spark-model/src/layers/vision_encoder/enc_impl/forward.rs
@@ -12,6 +12,47 @@ use spark_runtime::gpu::GpuBackend;
 
 use super::super::VisionEncoder;
 
+/// Do the packed merged rows this batch will write fit `buf_out`?
+///
+/// #799: this bound existed as a `debug_assert!`, which is compiled out of the
+/// `--release` binaries we serve. A single video request then wrote 4.7x past
+/// the allocation, raising `CUDA_ERROR_ILLEGAL_ADDRESS` and poisoning the CUDA
+/// context — the process survived and answered 503 to every later request, for
+/// every tenant, until it was restarted.
+///
+/// The caller's own doc comment says the scheduler caps `Σp <= p_max` "so this
+/// is normally unreachable". Video defeats that cap: all temporal groups of one
+/// clip arrive as a SINGLE media item and are encoded as one batch, so the
+/// per-item cap never sees the sum. 19 groups of 30x34 patches merge to 4845
+/// rows against a 1024-row buffer.
+///
+/// A `Result` rather than an assert, and a free function rather than a method,
+/// for the same reason `check_pixel_len` next door is one: the refusal is the
+/// behaviour worth testing, and a bound that can only be exercised with a GPU
+/// attached is a bound nothing will exercise. Prose is not a bound; this is.
+fn check_packed_rows(mp_i: &[usize], mp_off: &[usize], p_max: usize) -> Result<()> {
+    anyhow::ensure!(
+        mp_i.len() == mp_off.len(),
+        "vision: {} merged row counts but {} offsets; the packed layout is inconsistent",
+        mp_i.len(),
+        mp_off.len()
+    );
+    let end = match (mp_off.last(), mp_i.last()) {
+        (Some(off), Some(n)) => off
+            .checked_add(*n)
+            .ok_or_else(|| anyhow::anyhow!("vision: merged row offset {off} + {n} overflows"))?,
+        _ => 0,
+    };
+    anyhow::ensure!(
+        end <= p_max,
+        "vision: this batch packs {end} merged rows into an output buffer of {p_max} rows. \
+         A video arrives as one media item whose temporal groups encode as a single batch, \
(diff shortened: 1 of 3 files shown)


tbraun96:
**AutoMerger sieve verdict: `sieve:cleared` — folded into stack #891 (`stack/perf-plumbing`).** Sieve confirmed both defects live on `main` (the unguarded `debug_assert!` at `forward.rs:249`; unconditional `gs.rollback(dropped)` at `rollback.rs:411`), both fixes fail-fast (the OOB bound is an unconditional `ensure!`, not compiled out in release; the `None` rewind path warns and refuses rather than silently clamping), and both tests pin the real outage numbers (4845-row batch; 96/1, 88/1). Zero file overlap with either pending stack. Rides #891's already-owed campaign; closes with landing evidence after it merges. Its attribution signatures: vision-prefill phase and json_schema-rollback phase — disjoint from every other constituent.

DrRainbows:
Closing as already on `main`. This landed via #891 (`356701d6`, an ancestor of `origin/main`). Named artifacts: `crates/spark-server/src/scheduler/rollback_tests.rs:301` (`it_does_not_silently_clamp_to_the_available_history`) and `crates/spark-model/src/layers/vision_encoder/enc_impl/forward.rs:17` (the #799 bound). All three files from this PR's diff are byte-identical on `main`. Every commit here cherry-picks empty onto current `main` (`c2948e3f`).

PR #868, opened 2026-09-04
stack/ci-plumbing: PR telemetry on a Discussion, action bumps, and a stack planner

tbraun96:
**Stack group: `ci-plumbing`.** Three changes that are all CI plumbing, share no source file, and — the point of stacking them — touch **zero `PERF_PATHS` files**, so this group owes **no benchmark campaign at all**.

| | PR | why it earns a merge |
|---|---|---|
| 1 | **#621** | Moves the PR-telemetry view onto a Discussion. The view rewrites itself every four hours and on every merge; an Issue that is never closed reads as a task nobody is doing. Verified still needed: main's `pr-telemetry.yml` still declares `issues: write` and keys off a tracking issue. |
| 2 | **#741** | Dependabot bump of `crate-ci/typos` and `Avarok-Cybersecurity/AutoRepoRAG`. Small, current (dependabot refreshed it today), and it keeps two pinned actions from ageing out. |
| 3 | — | `stack-plan.py`, the tool that produced this grouping. |

## Why stack at all

Certification is the scarce resource. A PR touching `PERF_PATHS` (`crates`, `kernels`, `Cargo.*`, `vendor`, `jinja-templates`, `rust-toolchain.toml`, `3rdparty_patches`) re-opens all 11 gates and owes a ~5h GPU campaign. Landing those one at a time pays that cost once per PR.

The larger lever on a 98-PR backlog is the other half: **a PR touching no `PERF_PATHS` file owes no campaign at all.** `stack-plan.py` separates the two and reports file collisions per candidate group, because two PRs editing the same file conflict when stacked and finding that out at cherry-pick time is worse than finding it out in the plan.

Today's backlog: **8 free, 42 owe a campaign, 48 blocked or draft.**

## Collision check, done before cherry-picking

`.github/workflows/ci.yml` is touched by #856, #741, #622 and (formerly) #503. This stack takes only #741's hunk of it; #622 needs a partial rebase (below) and #856 is already in flight, so nothing here collides.

## What I checked before including each

Both were verified **still relevant**, not merely open:

- #621 — the Discussion move is genuinely absent from main.
- #741 — refreshed by dependabot today.

Two candidates were **excluded** after the same check:

- **#503** — closed as already-landed. Its `check_spdx.py` shows as a new file while that file already exists on main, and main is already 101/101 SHA-pinned. Merging it would have reverted three weeks of workflow changes.
- **#622** — *half* landed. Its harvest-subject bounding is already on main (`governance-harvest.yml:330` builds `... (${run_count} run(s))`), but its `governance-evict.sh` scripts are new. It needs a partial rebase to drop the landed half; taking it whole would revert.

## A correction I owe the record

I first reported #621 as an **orphan branch** — 1 commit, 3573 files, no merge base — which would have meant merging it reverts main wholesale. That was **wrong**, and it was my own fault: this working clone had only **6 commits** of `main`, so `git merge-base` had nothing to find. GitHub's compare API disagreed (`ahead 1, behind 101, merge base c1ae36af`), which is what prompted the re-check. After `--deepen`, `check_merge_ancestry.sh` reports `✓` and the real diff is **2 files**.

The repo already carries this lesson — `check_merge_ancestry.sh`'s own header says a shallow clone "cannot distinguish 'no common ancestor' from 'not deep enough'". I walked into it anyway. Ancestry is now verified against a deepened clone before any cherry-pick.

### DIFF
--- a/.github/actions/upsert-discussion-comment/action.yml
+++ b/.github/actions/upsert-discussion-comment/action.yml
@@ -1,0 +1,105 @@
+name: Upsert a Discussion comment
+description: >-
+  Keep exactly ONE bot-authored comment on a Discussion, rewriting it in place
+  instead of appending a new one on every run. The comment is found by a marker
+  embedded in the body itself, so there is no comment id to store anywhere and
+  a manually deleted comment simply results in a fresh one.
+
+# ── Why a Discussion and not an issue ───────────────────────────────────────
+#
+# These reports are standing documents, not work items. An issue that is never
+# closed reads as a task nobody finished; a Discussion reads as a page, which is
+# what a continuously-rewritten view actually is.
+#
+# ── Why this is an action and not copied into each workflow ─────────────────
+#
+# Discussion comments are GraphQL-ONLY. There is no REST `issues/comments`
+# equivalent, so the familiar `gh api -X PATCH .../issues/comments/$id` cannot
+# be reused and every caller would otherwise carry its own hand-written
+# mutation. Two callers already exist (pr-telemetry, intent-eigenvector), which
+# is one more than a copied implementation survives.
+
+inputs:
+  discussion-number:
+    description: >-
+      The Discussion to post into. MAY BE EMPTY: unset means "not configured
+      yet", and the action says so and succeeds rather than guessing a number
+      and commenting on something arbitrary. Callers render to the step summary
+      regardless, so an unconfigured repo still gets the report.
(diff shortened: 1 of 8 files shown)


tbraun96:
Temporarily removed from the merge queue for the same reason as #856 — stale check-set churn against a stalled runner pool. CLEAN, re-queued on revival.

PR #869, opened 2026-09-04
stack/deps-cargo: land getrandom 0.4; hold toml 1.1 and ratatui 0.30 with cause

tbraun96:
**Stack group: `deps-cargo`.** Five open dependabot Cargo PRs, triaged together so one campaign covers the group instead of five. **Two of the five do not survive triage**, and that is the useful part of this PR.

| PR | bump | verdict |
|---|---|---|
| **#558** | `getrandom` 0.3.4 → 0.4.3 | **landed** |
| #554 / #555 | patch / minor lockfile groups | folded in where they resolved cleanly |
| **#556** | `toml` 0.8 → 1.1 | **held — breaks the build** |
| **#557** | `ratatui` 0.29 → 0.30 | **held — not a mechanical bump** |

## #556 breaks the build. Proven, not suspected.

```
thread 'main' panicked at crates/atlas-kernels/build.rs:223:33:
Invalid HARDWARE.toml: TOML parse error at line 1, column 11
  | [hardware]
  |           ^ unexpected content, expected nothing
```

The 1.x parser rejects input 0.8 accepted. This needs a code change in `build.rs`, not a version bump.

## #557 is not mechanical either

`crates/spark-server/Cargo.toml` carries the reason **directly above the pin**:

> ratatui pins its own crossterm; the version here MUST match ratatui 0.29's (crossterm 0.28) or two raw-mode stacks fight over the terminal.

The PR bumps ratatui to 0.30 and leaves `crossterm = "0.28"` untouched. `tui-textarea 0.7` also depends on ratatui, and a dry-run resolve pulls an entire new backend stack (`wezterm-*`, `vtparse`, `unicode-truncate`). **Dependabot cannot read that comment; a human has to.**

Holding it also leaves the `lru` advisory (GHSA-rhfx-m35p-ff5j) dismissed on its original reasoning — ratatui 0.29's layout cache calls `cap`/`get`/`resize` and never `IterMut`, so the vulnerable iterator is still never constructed.

## What actually changed

```
getrandom: 0.3.4 -> 0.4.3
r-efi:     5.3.0 -> 6.0.0     (transitive)
winnow:    0.7.14 -> 0.7.15   (transitive)
```

Verified with `cargo check --workspace --all-targets`: **EXIT=0, zero errors**, with the held bumps reverted and the landed one in place.

## Why this owes a campaign, and why that is the point

This touches `Cargo.lock` and a `crates/**/Cargo.toml`, both in `PERF_PATHS`, so all 11 gates re-open. That is one campaign for the whole dependency group rather than one per bump — which is the entire reason to stack. Two of the five would have consumed a campaign each and then failed anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -205,7 +205,7 @@ name = "atlas-tier"
 version = "1.0.0-beta-preview"
 dependencies = [
  "anyhow",
- "getrandom 0.3.4",
+ "getrandom 0.4.3",
  "libc",
 ]
 
@@ -875,7 +875,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


tbraun96:

|---|---|---|---|---|
| 1 | #869 | deps: land getrandom 0.4, hold toml/ratatui | deps | **← this PR** |

PR #870, opened 2026-09-04
stack/verify-and-spec: three spec-decode correctness fixes, one campaign

tbraun96:
**Stack group: `verify-and-spec`.** Six open PRs from @rsafier in the verify / spec-decode subsystem, triaged as one group. **Three land here; three are already on main.**

| PR | change | verdict |
|---|---|---|
| **#693** | dflash: capture prompt hiddens at the chunk's PROCESSING start, not the KV write floor | **landed** |
| **#742** | spec: in-think EOS suppression on the verify emit path | **landed** |
| **#745** | verify: sliding-window attention needs per-token metadata in every spec verify lane | **landed** (+ a rebase fixup, below) |
| #699 | verify: K4 verdict rewound by the draft list | **already on main** |
| #703 | stream: kill switch for the SimHash semantic-loop guard | **already on main** |
| #704 | verify: raise `VERIFY_ROW_CAP` 96 → 160 | **already on main** |

## How "already on main" was established

Cherry-picking each onto current `main` produces an **empty commit** — the tree already contains the change. Confirmed independently for each:

- #703 — `ATLAS_SIMHASH_LOOP=0` is at `handle_token.rs:19`.
- #699 — `crates/spark-server/src/scheduler/verify_k4_verdict.rs` exists.
- #704 — `verify_e2.rs:44` reads `pub(in crate::model) const VERIFY_ROW_CAP: usize = 160;`.

On #704 I nearly got this wrong: a doc comment at `impl_a3.rs:70` still says `VERIFY_ROW_CAP` **(96)**, which reads as "not landed". It is a stale comment on a different arm — the constant is 160. Worth a follow-up to fix the comment.

## The rebase fixup, and why it is a separate commit

#745 was written 2026-08-25. `ForwardContext` has since gained two fields, so the PR no longer compiles against `main`:

```
error[E0063]: missing fields `hc_row_offset` and `host_token_ids`
              in initializer of `layer::ForwardContext<'_>`
  --> crates/spark-model/src/model/trait_impl/verify_a.rs:599:23
```

The values are not invented: `hc_row_offset: 0` is what **all six** other construction sites use (`verify_a` ×2, `verify_c2`, `speculative`, `decode_a`, `decode_a2`), and `host_token_ids: None` matches the two siblings in the same file. It is its own commit so the rebase is attributable to me rather than folded silently into @rsafier's.

## Verification

`cargo check --workspace --all-targets` → **EXIT=0, zero errors.** Before the fixup it failed with the E0063 above, which is what makes the fixup load-bearing rather than cosmetic.

## Why this owes one campaign, not six

All three touch `crates/`, so all 11 gates re-open. Landing them individually would cost **six** campaigns — three of them for changes already on `main`. Grouping found the three redundant ones with a single cherry-pick pass and the staleness with a single build.


tbraun96:
**Superseded by stack #885 — closing.** Every file this PR changes is byte-identical in `stack/certified-2026-09-04` (#877, the stack top): `git merge-tree --write-tree stack/certified-2026-09-04 <this head>` returns the stack's own tree `34077e7f90…` — merging this PR into the stack changes nothing. Its constituents ride as dedicated layers: #883 (L2, #742) and #882 (L3, #745), with the EOS-suppression tests the sieve demanded added on top (`emit_eos_thinking_tests.rs`). This was our own intermediate stack from before the consolidation — exactly the 'batched into a larger PR, forgot to delete the original' pattern, so it goes now. If any hunk did not make it across, point at it and I will lift it onto a fresh branch.

Issue #872, opened 2026-09-04
qwen3_ssm/init.rs is 511 LoC: Qwen3SsmLayer::new needs a clean split


tbraun96:
## What

`crates/spark-model/src/layers/qwen3_ssm/init.rs` is **511 LoC**, 11 over the 500-line cap. It is on the allow-list in `.github/workflows/file-size-cap.yml` with this issue as its tracking entry, per that list's own rule ("Adding to this list requires a rationale comment AND a tracking issue").

## Why it was allow-listed rather than split

The file is a single function. `impl Qwen3SsmLayer` opens at line 7, `pub fn new(` at line 8, and it runs to line 504 — there is **no piecewise seam** to move. Every other split done during this backlog pass moved whole functions verbatim and verified the bodies byte-identical:

- `verify_a.rs` 638 → 417 + 254 (6 methods, 0 differing)
- `ssm_gdn_b.rs` 569 → 205 + 399 (10 functions, 0 differing)

Neither technique applies here. Splitting a 496-line constructor means *restructuring* it — deciding which initialisation phases become helpers and what state they take — which is a design decision for whoever owns the SSM layer, not a mechanical move by whoever happened to trip the cap.

## How it got here

481 → 511 in the GDN batched-verify stack (#837/#838/#844/#845, landed as one certified group). The growth is init for the K=5..16 pointer tables.

## What a fix looks like

`Qwen3SsmLayer::new` reads as distinct phases (config/derived sizing, buffer allocation, kernel-handle resolution, table construction). Extracting even one into a private `fn` in a sibling module would clear the cap with room to spare. It wants someone who knows which of those phases share invariants.

## Removing the entry

Delete the `crates/spark-model/src/layers/qwen3_ssm/init.rs` line from `allow_list` in `.github/workflows/file-size-cap.yml` and close this issue. The cap job will then enforce it again.

PR #874, opened 2026-09-04
stack/gdn-dflash2: four GDN + DFlash-2 PRs, one campaign

tbraun96:
**Stack group: `gdn-dflash2`.** Four PRs from @rrstesiak that form a real dependency chain — the author had already declared #838 stacked on #837. All four touch `crates/` and `kernels/`, so individually they would owe **four** campaigns. Here they owe **one**.

| order | PR | change |
|---|---|---|
| 1 | **#837** | gdn: pointer-table twins for wyN K=5..16 (behaviour-neutral foundation) |
| 2 | **#838** | gdn: switch on the K=5..16 batched verify (dispatcher arm + 16-slab tables) |
| 3 | **#844** | dflash-2: C=16 write-on-accept verify, batched drafter precompute |
| 4 | **#845** | scheduler: DFlash gamma resolver — the serve picks verify width itself |

Ancestry confirmed with `merge-base --is-ancestor`: 837 → 838 → {844, 845}.

## Composing the chain

My first attempt got this wrong and it is worth recording. Cherry-picking each PR from `merge-base(main, prN)..prN` re-applies its ancestors, so #838 came out "empty" and #844/#845 conflicted against a half-applied tree. Applying the **chain tip** (837+838) and then each descendant's **unique** commits (`pr-838..pr-844`, `pr-838..pr-845`) gives a clean 8-commit stack with no conflicts.

`cargo check --workspace --all-targets` → **EXIT=0, zero errors.**

## Two files crossed the 500-LoC cap, and they were handled differently on purpose

**`ssm_gdn_b.rs` 391 → 569: split.** `gdn_decode_wy4`, `gdn_decode_wyn` and the K=5..16 table form move verbatim into `ssm_gdn_wyn.rs`, leaving **205 + 399**. Split at `wy4` rather than `wy2` because a `wy2` cut left the sibling at 536 — still over. `ops.rs` declares and re-exports both halves, so no call site changes.

Verified exact rather than asserted: **10 functions in, 3 + 7 out, zero bodies differing.**

**`qwen3_ssm/init.rs` 481 → 511: allow-listed, tracked as #872.** This one has no seam — the file is a single function (`impl Qwen3SsmLayer` at line 7, `pub fn new(` running to 504). Splitting a 496-line constructor means restructuring it: deciding which init phases become helpers and what state they share. That is a design call for the SSM owner, not a mechanical move by whoever tripped the cap.

The allow-list's own rule requires "a rationale comment AND a tracking issue". Both are there, and the entry says **why this file was treated differently from the two that were split** — an escape hatch used without saying why is how a 505-line exception becomes a 900-line one.

## Why one campaign, not four

All four touch `PERF_PATHS`, so all 11 gates re-open. Landing them separately costs four campaigns (~5 GPU-hours each) to certify one coherent chain that only makes sense composed — #838 switches on what #837 builds, and #844/#845 build on both.


rrstesiak:
Thanks for composing the stack; the chain is right (837 -> 838 -> {844, 845}) and one campaign is the right cost.

One timing problem: #874 was cut from #844 @ b0b82e64 and #845 @ 8a7a14c4, and those are no longer the heads. #844 moved to 7f51e945 about the same minute #874 opened. The two new commits (538ca211, 7f51e945) are the fixes for every thread in Turney's review plus @rsafier's overrun report, and they already handle both files you hit, at a different seam:

- `ssm_gdn_b.rs` 569 -> 457: the three write-on-accept launchers moved verbatim to a new `ops/ssm_gdn_woa.rs` (123 lines), which mirrors the kernel file they launch (`gated_delta_rule_wy4_woa.cu`). `ops.rs` re-exports it, no call-site changes. No `ssm_gdn_wyn.rs`, so the wyN table code stays where the #837/#838 reviews read it.
- `qwen3_ssm/init.rs` 511 -> 494: the stash allocation left the constructor entirely. The stash and engaged words are now bound lazily on the first write-on-accept request (`model/trait_impl/gdn_woa.rs`, pre-capture), which was also a review finding (the eager 32-seq allocation was oversized). That closes #872 without an allow-list entry.

Also in those commits: the `_p4` GEMM twin is removed (neutral on this box, Turney's call), the twin is an explicit per-call request from the DFlash batched step rather than something every K=4 verify could engage, and a device oracle (`examples/gdn_woa_oracle.rs`) asserts twin+fold bit-equal to the parent (ALL PASS, 0 mismatches). The #844 body and a consolidated reply carry the receipts, including the `ATLAS_NO_GDN_WOA=1` A/B on the fixed tree.

#845's review fixes are done and rebased on 7f51e945; that push lands today.

So rather than carry the review delta by hand onto 096063d9, the cheaper path is to recompose the stack from the two new heads once #845 is pushed. If you would rather I open the recomposed branch myself and you re-point #874 at it, say so.

tbraun96:
**Superseded by stack #885 — closing.** Merging this PR into the pre-cure stack tree (`b0a5664df3`, tree `f34b9a74a7…`) changes nothing — its entire content is in the lineage; the only divergence from today's top is the three sieve-cure test commits added ON TOP of it (slot-order test in `mtp_step.rs`, test-module registration in `scheduler/mod.rs`, fp8 parity oracle). Its constituents ride as dedicated layers: #881 (#837), #880 (#838), #879 (#844), #878 (#845). Same forgot-to-delete pattern as #870, same cure. If any hunk did not make it across, point at it and I will lift it onto a fresh branch.

PR #877, opened 2026-09-04
stack/certified-2026-09-04: 8 PRs, one campaign — dflash/gdn verify line, spec fixes, deps

tbraun96:
## The plan (aggregation PR — this is the tree that lands)

Eight open PRs composed onto one branch so **one certification campaign covers the group instead of eight** (~5 GPU-hours instead of ~40). Constituent commit boundaries are preserved — no squash on the branch — so every gate failure maps to a commit range. Source branches stay intact; a suspect constituent can be dropped and the remainder re-certified.

| merge order | PR | what it is | attribution cluster |
|---|---|---|---|
| 1 | #869 | deps: land getrandom 0.4, hold toml/ratatui | **deps** — build/link failures |
| 2 | #742 | in-think EOS suppression on the verify emit path | **spec-verify** |
| 3 | #745 | sliding-window per-token metadata in every spec verify lane (+ file split, ForwardContext fixup) | **spec-verify** |
| 4 | #837 | gdn: wyN pointer-table twins K=5..16 (behavior-neutral foundation) | **gdn** |
| 5 | #838 | gdn: switch on the K=5..16 batched verify (stacks on #837) | **gdn** |
| 6 | #844 | dflash-2: write-on-accept verify, drafter precompute, slot-ordered dispatch, fp8 GEMM twins | **dflash** |
| 7 | #845 | DFlash gamma resolver — serve picks verify width itself | **dflash** |
| 8 | #777 | bfcl: prove the scorer imports before generating 995 responses | **bfcl-harness** |

**Gate → cluster attribution map** (written up front, per the adversarial review): gdn numerics gates → #837/#838 · DFlash gates → #844/#845 · spec/coherence → #742/#745/#844's verify path · BFCL harness anomalies → #777 · build/deps → #869. The spec-verify cluster is the widest (3 PRs); worst case is drop-one-recertify, bounded at 2–3 steps.

## Offline gates already green on the composed head

`cargo check --workspace --all-targets` EXIT=0 · full suite **5503 tests pass** · LoC cap clean · SPDX clean · certification selftest green · based on current `main` (behind 0).

## Adversarial review: PUBLISH, conditional

A fable adversarial pass cleared this plan with **seven conditions before the campaign is approved**:

1. ⏳ Gate C2 NVFP4 coherence + tool-call smoke on the composed head (touches gdn wyN tables and fp8/w4a16 GEMM twins — the class 5503 cargo tests say nothing about)
[...]

PR #878, opened 2026-09-04
L7/8 scheduler: DFlash gamma resolver (#845)

tbraun96:
Layer 7 of 8 in stack `certified-2026-09-04` — the aggregation PR (top, certifies once for all) is **#877**. Lower layers run cheap correctness only; the composed tree is certified on the top. Constituent PR, sieve verdicts and the full plan: see #877. Merging happens by collapsing the certified top into one queue entry — do not merge layers individually.


tbraun96:

PR #887, opened 2026-09-04
atlas-core: strengthen DeepSeek V4 parser contracts

TheTom:
## Summary

Seven held audit commits for the DeepSeek V4 config parser, ported onto current main with every mutant re-executed against today's production. Six are test-only oracle repairs; one extracts a host-side helper in `spark-model` so the sliding-RoPE table is tested where it is actually computed.

## Problems found (each proven with an executed old-green mutant)

- The mscale test asserted both zero fields and then a redundant equality between them; the equality has no surviving mutant of its own.
- The compress-theta test could not distinguish a real `compress_rope_theta` lookup from a hardcoded 160000: the only fixture matched the published value.
- The old sliding-RoPE test asserted a test-local constant equaled 10000; changing production `MAIN_ROPE_THETA` to 9999 left it green — it never compiled or called the production consumer.
- All four YaRN fixture values (factor 16, beta-fast 32, beta-slow 1, original max position 65,536) equaled the runtime fallback literals, so disabling all four checkpoint lookups stayed green.
- The explicit-mscale override test supplied `mscale_all_dim: 0.0`, matching both the disabled contract and the generic default, so restoring the nested lookup without the zero override stayed green.
- `non_ds4f_factory_default_mscale_untouched` never called `parse_deepseek_v4` and its Qwen factory field has no in-repository production consumer (re-verified by repo-wide field-read trace on current main: only `yarn_rope_mscale` reads these fields, from DS4F attention/loader paths). Deleted as misplaced.
- The two partial DSpark fixtures could not prove `dspark_target_layer_ids` was required; removing it from the required-field list left both green.

## Repair

Renamed and strengthened the parser tests with discriminating sentinel fixtures (compress theta 234,567; YaRN 12.5/40/2/32,768; explicit `mscale_all_dim: 0.5`), added the documented `rope_parameters` alias check, replaced the partial DSpark fixtures with one valid four-field contract minus each field independently requiring the exact `DSpark config is missing `<field>`` reason, and moved the sliding-RoPE oracle into `spark-model` as `main_inv_freq_values`, tested at rotary dim 64 against independent anchors (index 0 = 1.0, 8 = 0.1, 16 = 0.01, 24 = 0.001). `main_inv_freq` now uploads exactly that returned vector — a behavior-preserving extraction.

- Hardcoding the compress-theta assignment to 160000.0 fails the sentinel with `left: 160000.0, right: 234567.0`.
- Hardcoding `yarn_factor = 16.0` fails the alternate partition with `left: 16.0, right: 12.5`.
- Removing `dspark_target_layer_ids` from the required list fails with the missing-field reason replaced by the later empty-list error.
- Each was applied alone and production restored afterward (clean diff verified).

## Test plan

- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo test -p atlas-core` (126 passed)
- [ ] `cargo test -p spark-model --lib main_inv_freq` — does not compile on this macOS host (`spark-storage` uses Linux-only `posix_fallocate`/`posix_fadvise`/`O_DIRECT`); the new spark-model check is owned by Linux CI on this PR
- [ ] Tested against a real model / hardware (parser and host-table contracts only; no weights loaded, no kernel launched)

Production behavior is unchanged (the `main_inv_freq` change is a pure extraction; the same loop computes the same floats). The tests stop at parsed configuration and the host-side frequency vector: no device table bytes are inspected, no RoPE kernel launched, no inference output compared. GAP-DS4F-0001 (runtime-table response to alternate YaRN parameters) and GAP-DS4F-0002 (no production read of `dspark_noise_token_id` beyond validation) remain open in the audit ledger.

## CLA

### DIFF
--- a/crates/atlas-core/src/config/parsers/deepseek_v4.rs
+++ b/crates/atlas-core/src/config/parsers/deepseek_v4.rs
@@ -371,22 +371,15 @@ mod mscale_contract_tests {
     // value is unchanged; asserted here as the sliding-theta contract.
     const SLIDING_ROPE_THETA_CONTRACT: f32 = 10000.0;
 
-    // Test 1 + Test 4: the DS4F config makes yarn_rope_mscale == 1.0 for EVERY
-    // runtime call site. All nine sites call `yarn_rope_mscale(ctx.config)` with
-    // this parsed config; the helper (spark-model) is unit-tested separately to
-    // return exactly 1.0 when the two mscale terms are equal. Here we prove the
-    // parser produces terms that force that: yarn_mscale == yarn_mscale_all_dim.
+    // The parser owns the DS4F field override. The spark-model helper tests own
+    // the resulting runtime ratio.
     #[test]
-    fn ds4f_forces_mscale_terms_equal_so_ratio_is_one() {
+    fn ds4f_parser_forces_both_mscale_terms_to_zero() {
         let c = parse_deepseek_v4(DS4F_CONFIG).expect("parse DS4F");
         assert_eq!(c.yarn_mscale, 0.0, "yarn_mscale must be forced to 0.0");
         assert_eq!(
             c.yarn_mscale_all_dim, 0.0,
             "yarn_mscale_all_dim must be forced to 0.0"
-        );
-        assert_eq!(
-            c.yarn_mscale, c.yarn_mscale_all_dim,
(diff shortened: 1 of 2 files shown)


TheTom:
recheck

tbraun96:

PR #888, opened 2026-09-04
atlas-core: fail-closed trainable-token parsing + LoRA contract repairs

TheTom:
## Summary

Eight held audit commits for the PEFT/LoRA adapter parser, ported onto current main with both headline mutants re-executed against today's production. Six are test-only oracle repairs. **Two change production behavior: `trainable_token_indices` parsing is now fail-closed** (details below) — reviewers should weigh that section, not the test churn.

## Production behavior change: trainable-token parsing is now fail-closed

The `trainable_tokens_delta` tensor's rows align positionally to the parsed id list. The old parser flattened and deduped a union across the `embed_tokens`/`lm_head` dict form, which silently reordered or merged ids — a misaligned id→delta-row mapping applies the wrong rows to the wrong tokens without any error. The parser now requires one shared unique ascending list: duplicates, descending order, differing per-module lists, unknown modules, negatives, non-integers, and ids beyond u32 are each a named reject. An adapter that legitimately declares differing per-module lists was silently mis-applied before; now it is rejected with the reason. Accepting-but-corrupting became rejecting-by-name.

## Test problems found (each proven with an executed old-green mutant)

- The happy-path scaling test mixed scaling and target-module claims; split into two focused contracts.
- The rsLoRA test did not pin rank, alpha, or mode inputs, so a wrong-source scaling could satisfy it.
- `layers_to_transform_array_accepted` asserted a dead field; renamed to its actual acceptance-only claim.
- Three GDN target spellings present in the production reject arm (`in_proj_ba`, `in_proj_a`, `in_proj_b`) were untested — removing one from the reject arm silently admitted it. (Upstream meanwhile made `out_proj` accepted for GDN; the port preserves that feature change and tests around it.)
- The unsupported-controls test duplicated a `modules_to_save` row and missed `alpha_pattern`.
- Zero rank and non-positive alpha were rejected but the tests accepted any error; they now pin the `REJECT(r)` / `REJECT(lora_alpha)` reasons.

- Swapping the zero-rank rejection reason to `REJECT(lora_alpha)` fails `zero_rank_rejected_by_name` with the wrong-reason message printed.
- Removing `in_proj_ba` from the reject arm makes parsing succeed and `gdn_module_rejected_named` fail at `unwrap_err` with the accepted config printed.
- Each applied alone, production restored afterward (clean diff verified).

## Test plan

- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo test -p atlas-core` (134 passed)
- [ ] `ATLAS_SKIP_BUILD=1 cargo clippy --workspace --tests --all-features -- -Dwarnings` (owned by Linux CI; this macOS host reaches Linux-only paths outside this diff)
- [ ] Tested against a real adapter checkpoint (not run: the rejection semantics are proven at the parser seam; no adapter was loaded onto a model)

The fail-closed strictness is deliberate: every newly rejected shape was previously accepted into a silently wrong overlay mapping, so no correct adapter behavior is lost — but if a real published PEFT adapter in the wild uses differing per-module token lists intentionally, it now fails at load with a named reason instead of loading wrong; that trade is the review question. No adapter weights were loaded and no overlay kernel executed; consumption ties (overlay builder row alignment) are static traces of current source.

## CLA

### DIFF
--- a/crates/atlas-core/src/config/parsers/lora.rs
+++ b/crates/atlas-core/src/config/parsers/lora.rs
@@ -75,9 +75,9 @@ pub struct PeftAdapterConfig {
     /// which layers receive deltas; this is kept only for the startup log.
     pub layers_to_transform: Option<Vec<usize>>,
     /// Vocab-extension / trainable-token ids for the token overlay
-    /// (Feature 2). Flattened + deduped union of the config's
-    /// `trainable_token_indices` (list form, or `{"embed_tokens":[…],
-    /// "lm_head":[…]}` dict form). Empty ⇒ no `trainable_tokens` overlay.
+    /// (Feature 2). Unique ascending order from the config's
+    /// `trainable_token_indices` list, or the common order declared for
+    /// `embed_tokens` and `lm_head`. Empty ⇒ no `trainable_tokens` overlay.
     pub trainable_token_indices: Vec<u32>,
     /// Accepted `modules_to_save` leaves — the subset Atlas can apply as a
     /// token overlay (`embed_tokens` / `lm_head` full-row replacement).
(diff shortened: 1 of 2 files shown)


TheTom:
recheck

tbraun96:

PR #889, opened 2026-09-04
atlas-closure: canonicalize paths before hashing + digest oracle repairs

TheTom:
## Summary

Six held audit commits for `atlas-closure`, the crate that computes the kernel-source closure digest consumed by benchmark-gate record reuse. Five strengthen digest oracles that previously accepted real hash defects; **one is a production fix: the comparison root and config paths are canonicalized before relativizing**, which fixes two failing tests on macOS and stops checkout-specific absolute paths from leaking into the report and digest. All three headline mutants were re-executed against today's production.

## Production change and its digest impact

`hash_with_report` relativized against the caller's uncanonicalized root while `expand` canonicalizes every source, so on hosts with path aliases (macOS `/var` → `/private/var`) `strip_prefix` failed and absolute paths entered both the unresolved report and the digest — the digest then depended on the checkout location. The fix canonicalizes the root (with a fallback to the raw path) and canonicalizes config entries the same way. On Linux CI, where the passed root and config paths are already canonical absolute paths, relative paths and digests are unchanged — existing benchmark records are not invalidated by recomputation. Digests change only where they were previously wrong (aliased roots leaking absolutes).

## Test problems found (each proven with an executed old-green mutant)

- `unresolved_entries_are_keyed_by_the_including_file` counted entries; production emitting `slot-0`/`slot-1` keys instead of including-file identities stayed green. Now requires the exact repo-relative `common/a.cu -> gone.cuh` / `common/b.cu -> gone.cuh` set.
- `angle_bracket_includes_are_not_followed` accepted `is_ok()`; parsing angle brackets as quoted includes survived because missing includes are nonfatal. Now places a same-name local header, requires an empty unresolved report, and requires the digest unchanged after editing that header.
- `commented_out_includes_are_ignored` — same pattern for `//` comments.
- `non_source_inputs_each_move_the_hash` proved flag presence only; sorting flags before hashing survived. Now compares define-then-undefine against the reverse order and requires distinct digests.
- The different-path test survived a basename-only hash. Now adds byte-identical `common/same.cu` vs `model/nvfp4/same.cu` and requires distinct digests (the name selects which kernel is shadowed, so it is a different compile).

- Sorting flags before hashing: the flag-order assertion fails with both digests equal.
- Hashing only the file basename: the same-content-different-path assertion fails with both digests equal.
- Reverting the canonicalization: 3 tests fail on this macOS host (the two originally-failing tests plus the strengthened unresolved-identities test, which now also depends on correct relativization); restoring it returns the crate to 15/15.

## Production consequence

A surviving digest defect preserves benchmark-gate records that should have been invalidated: flag-order changes, moved same-content kernels, or checkout-relative aliasing could all reuse stale hardware records. These oracles are what make record reuse trustworthy.

## Test plan

- [x] `cargo test -p atlas-closure` (15 passed, 0 failed)
- [x] `cargo clippy -p atlas-closure --tests -- -D warnings`
- [ ] Tested against a real model / hardware (not applicable: host-side hashing crate; no kernel compiled)

The Linux-digest-stability claim above is a static argument (canonical inputs in, identical outputs out), not a replay against committed records; if the gate CI passes on this PR with existing records accepted, that is the executed confirmation. Include-graph parsing itself (conditionals, cycles, transitive expansion) already had sound oracles and is unchanged.

## CLA

### DIFF
--- a/crates/atlas-closure/src/lib.rs
+++ b/crates/atlas-closure/src/lib.rs
@@ -157,6 +157,11 @@ pub fn hash(root: &Path, inputs: &ClosureInputs) -> Result<String> {
 /// *newly* unresolvable include is worth a build warning even though it is not
 /// worth a build failure.
 pub fn hash_with_report(root: &Path, inputs: &ClosureInputs) -> Result<Closure> {
+    // `expand` canonicalizes every source. Canonicalize the comparison root as
+    // well, otherwise platform aliases such as macOS `/var` -> `/private/var`
+    // make `strip_prefix` fail and leak checkout-specific absolute paths into
+    // both the report and digest.
+    let canonical_root = root.canonicalize().unwrap_or_else(|_| root.to_path_buf());
     let mut closure: BTreeSet<PathBuf> = BTreeSet::new();
     let mut raw: BTreeSet<(PathBuf, String)> = BTreeSet::new();
     for src in &inputs.sources {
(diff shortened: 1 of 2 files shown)


TheTom:
recheck

tbraun96:

PR #880, opened 2026-09-04
L5/8 gdn: K=5..16 batched verify switch-on (#838)

tbraun96:
Layer 5 of 8 in stack `certified-2026-09-04` — the aggregation PR (top, certifies once for all) is **#877**. Lower layers run cheap correctness only; the composed tree is certified on the top. Constituent PR, sieve verdicts and the full plan: see #877. Merging happens by collapsing the certified top into one queue entry — do not merge layers individually.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BHvLyzv6Ma5ySK48Rh3ytc

PR #891, opened 2026-09-04
stack/perf-plumbing: 2 PRs, one campaign — module-load dedup + EP null-group skip

tbraun96:
## The plan (aggregation PR)

Two sieve-cleared perf PRs composed for **one certification campaign instead of two**. Commit boundaries preserved (patch-id-verified byte-equivalent to the sieved PR diffs); source branches stay intact.

| order | PR | what | attribution signature |
|---|---|---|---|
| 1 | #705 | registry: load each kernel module once, derive the raw handle | **boot-phase** — cold-TTFT/shutdown symptoms, or loud `CUDA_ERROR_INVALID_HANDLE`; no steady-state footprint |
| 2 | #781 | cutlass: skip null-weight expert groups under EP + fail-fast `-140` guard | **grouped-GEMM phase** — prefill numerics/crash in MoE GEMM |

Disjoint phase signatures → a failing gate maps to a constituent without bisecting.

## Offline gates (composed head `f3e1cff84d`, base = current main)

`cargo check --workspace --all-targets` EXIT=0 · atlas-core **128/128** · spark-runtime **277/277** · LoC cap ok · sieves: both `sieve:cleared` (live-tree verified).

## Adversarial review: PUBLISH, 4 conditions

1. **Blocking pre-campaign smoke:** single-node MoE inference smoke on this branch (coherent output, no `-140`) — #781's modified kernel body runs on every grouped-GEMM call yet its only runtime evidence is two-node EP=2; the single-node config the campaign exercises is precisely the one never executed. Gate C2 does not substitute (dense 27B never enters grouped MoE GEMM)
2. **Claim hygiene:** #781's +5.7% is EP=2/two-node evidence and stays **uncertified by this campaign** — the campaign buys "does not regress single-node", nothing more
3. #705's ~14% boot claim is quoted only after a cheap same-box boot A/B, else marked "PR-claimed, unverified"
4. Attribution map above is part of the record

Campaign (~5 GPU-h, owed: vendor/ + crates/ + .cu are PERF_PATHS) starts only on explicit human approval. Constituents close with landing evidence after this merges.

### DIFF
--- a/.benchmarks/agentic-webserver/2026-09-06-da016237db.json
+++ b/.benchmarks/agentic-webserver/2026-09-06-da016237db.json
@@ -1,0 +1,490 @@
+{
+  "schema": 1,
+  "benchmark_id": "agentic-webserver",
+  "benchmark_name": "Agentic Webserver Test",
+  "git_sha": "da016237db",
+  "recorded_at": 1788726920,
+  "target_model": "Qwen/Qwen3.6-35B-A3B-FP8",
+  "params": {
+    "build_timeout_s": "600",
+    "command_timeout_s": "180",
+    "iterations": "10",
+    "max_tokens": "8192",
+    "max_turns": "40",
+    "request_timeout_s": "900",
+    "s_per_turn_budget": "8.5",
+    "serve_timeout_s": "30",
+    "wall_budget_s": "1800"
+  },
+  "command": [
+    "spark",
+    "benchmark",
+    "run",
+    "agentic-webserver",
+    "--param",
+    "build_timeout_s=600",
+    "--param",
+    "command_timeout_s=180",
+    "--param",
+    "iterations=10",
+    "--param",
+    "max_tokens=8192",
+    "--param",
+    "max_turns=40",
+    "--param",
+    "request_timeout_s=900",
+    "--param",
+    "s_per_turn_budget=8.5",
+    "--param",
+    "serve_timeout_s=30",
+    "--param",
+    "wall_budget_s=1800",
+    "--yes",
+    "--pull-request-gate"
+  ],
+  "served_by": "qwen3.6/qwen3.6-35b-a3b-fp8-bf16head",
+  "atlas_version": "1.0.0-beta-preview",
+  "hardware": {
+    "gpu": "NVIDIA GB10",
+    "driver": "580.159.03",
+    "sm_clock_mhz": 2411.0,
+    "source": "nvidia-smi"
+  },
+  "hardware_state": {
+    "sensitivity": "speed",
+    "before": {
+      "captured_at": 1788726122,
+      "machine": {
+        "hostname": "spark-28c2",
+        "machine_id": "7af66f30966a49b6886e00e2fce4b42f",
+        "gpu": "NVIDIA GB10",
+        "driver": "580.159.03"
+      },
+      "gpu_temp_c": 52.0,
+      "chassis_temps_c": [
+        {
+          "name": "acpitz",
+          "temp_c": 58.4
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 52.6
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 58.4
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 52.7
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 53.8
+        },
+        {
+          "name": "acpitz",
+          "temp_c": 54.8
+        },
+        {
+          "name": "acpitz",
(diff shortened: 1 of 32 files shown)


tbraun96:
**Composition update — pin `85b190322f`** (was `f3e1cff84d`… + 6 commits): folded in sieve-cleared #667 (rustdoc on `--exact-verify`, zero runtime footprint) and #669 (video-fidelity gains a 14th strict gating cell; no assertion weakened). Rationale: both touch `crates/`, so standalone they'd owe their own campaign — riding this stack's already-owed campaign makes their marginal certification cost zero. Offline gates re-run green: `cargo check` EXIT=0, atlas-plugin **900/900**. Attribution map gains two rows: doc-comments → #667 (no runtime footprint); bench-harness legs → #669. Delta re-clearance requested from the same adversarial reviewer; campaign still fires only on explicit approval.

tbraun96:
**Adversarial delta review: RE-CLEARED at `85b190322f`** — patch-id-verified both additions against their PR diffs (#667 doc-only confirmed, one removed line is itself a `///`; #669's new cell is a content-independent binary contract, so gating strictly in its own PR is acceptable — the measure-then-declare rule is for model-dependent bounds).

All prior conditions stand unchanged (single-node MoE smoke blocking; EP=2 claim uncertified; #705 boot claim needs its own A/B). Two additions to the record:
- #667 is **4 commits** (my earlier comment said five — corrected there)
- **Video-gate provenance:** the current qwen3.8 checkpoint scores 13/14 on video-fidelity with `mixed-media` red — that is a **pre-existing checkpoint property**, not introduced here. If video-fidelity runs anywhere in this stack's certification, run it once on this head first and record `mixed-media-pads` GREEN + the pre-existing red, so a red leg doesn't get misattributed to the stack.

tbraun96:
**Composition update — pin `338a6babd7`, and this closes the stack's composition** (no further folds): #866's two production-outage fixes ride the owed campaign — vision OOB bound (issue #799) and grammar-rewind panic (issue #842). Offline gates re-green: `cargo check` EXIT=0, rollback suite 25/25, vision test 1/1. The stack is now **6 constituents** (#705, #781, #667, #669, #866×2 fixes), six disjoint attribution signatures: boot / grouped-GEMM / no-footprint / bench-harness / vision-prefill / json_schema-rollback. Second delta re-clearance requested; campaign still fires only on explicit approval.

tbraun96:
**Second adversarial delta review: RE-CLEARED at `338a6babd7` — composition closed at 6 constituents; any further fold voids this clearance.**

Delta verified end-to-end by the reviewer (patch-id match, defect read directly on main, file-set disjointness proven by intersection, not asserted). Its one new condition is already discharged: the #866 vision-bound tests re-run in **release** mode as the commit's own negative-control claim requires — `refuses_the_video_batch_that_poisoned_the_cuda_context` and `admits_a_batch_that_fits` (the exact 1024-boundary) both pass under `--release`.

Standing pre-campaign checklist (all conditions verbatim from the reviews):
1. Single-node MoE smoke — blocking, covers #781's never-executed single-node path
2. #781's +5.7% EP=2 claim: uncertified by this campaign
3. #705's ~14% boot claim: quote only after a cheap boot A/B
4. Video-fidelity provenance note if that gate runs (pre-existing 13/14 red)

**This stack is now fully gated. The only remaining step is explicit human approval for the campaign.**

tbraun96:
## Sealing — certified at `da016237db`, brought up to date with main

All eleven gates green and committed (`8ecd13c440`): every record at `git_sha da016237db`, nine Speed-class records from one signer, the two Correctness-class BFCL draws from the boxes that ran them.

The branch has just been updated onto `main` (`strict: true`; main was one commit ahead, #908, `.github`-only). That diff touches **zero `PERF_PATHS`**, so the records still cover the new head.

Worth reading before merge: `concurrency-sweep` **failed this pin at 14:30Z** with C2 = 24.0 against a 24.1 floor, and it was not this PR. The identical pin on the identical box returned 29.0, and a probe on clean `main` immediately afterwards returned **27.6 — slower than this PR**. The full two-cluster evidence is in the certification commit message rather than only in a chat log.

This stack also adds a fourteenth `video-fidelity` leg (14/14 here, 13/13 on main). The BENCH.toml floor `legs_passed min = 13.0` should ratchet to **14** once this lands, or the new leg is ungated.

Sealing on the repo owner's standing authorisation.

tbraun96:
Same as #880: the `update-branch` moved this head to `7e85870ac9`, and a `Stamp` check run is bound to a sha, so the lane is held again. The `Seal` on this head is already green. Re-stamping at the current head.

Issue #58, opened 2026-05-14
[bug]  Using Qwen3.6 31B fp8 to convert the picture to markdown descripiton .  when you change the picture, LLM give you the response with the previous picture.


hxfhd:
### Atlas version / docker tag

 avarok/atlas-gb10:debug

### Model

Qwen/Qwen3.6-35B-A3B-FP8

### Hardware

 Gigabyte AI TOP ATOM ATAGB10-9000 Arm Cortex-X295 + Cortex A725 NVIDIA Blackwell single-GPU

### Launch command

```bash
version: '3.8'

services:
  atlas-inference:
    image: avarok/atlas-gb10:debug 
    container_name: atlas
    network_mode: "host"
    ipc: "host"

    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              count: all
              capabilities: [gpu]

    volumes:
      - /home/hxfhd/models/Qwen3.6-35B-A3B-FP8:/model_data

    # 移除了 --alias，将 --host 改为 --bind（根据你的错误提示 Usage）
    command: >
      serve 
      --bind 0.0.0.0
      --port 8888
      --max-seq-len 65536
      --kv-cache-dtype fp8
      --kv-high-precision-layers auto
      --gpu-memory-utilization 0.85
      --max-batch-size 1      
      --scheduling-policy slai
      --tool-call-parser qwen3_coder
      /model_data  
      --enable-prefix-caching 
      --speculative 


    restart: unless-stopped

    environment:
      - NVIDIA_VISIBLE_DEVICES=all
      - HF_HUB_OFFLINE=1
      - ATLAS_PREFILL_CHUNK_SIZE=4096
```

### Expected behaviour

1.   Using one python to convert one page of pdf to markdownfile.   

[test_single_page.py](https://github.com/user-attachments/files/27780658/test_single_page.py)

)
2.  when I tried first page such as page 95 ,  normally I got what I expected.  but when I change to another page 102,  LLM return the result according to Page 95 .  I got real result for Page 102 until I tried second time or third time.  It seems every time I change the test page,  this will happen again .  I attched the execute log  which I recorded and the picture for page 95 and 102 for your reference In actual behaviour by name executive log and Spainish_A_p095_screenshot.png ,Spainish_A_p102_screenshot.png 
3.  From my point of view ,it seems Atlas mixture the memory of the diferent input.  

### Actual behaviour

[server.log](https://github.com/user-attachments/files/27780618/server.log)

[...]

Issue #17, opened 2026-05-07
Panic in high_speed_swap.rs:167 ("inc_disk_ref on freed disk_block_id 0") on second request when --high-speed-swap and --enable-prefix-caching are combined


gbanyan:
## Summary

When `--high-speed-swap` is enabled together with `--enable-prefix-caching true`, the second incoming chat completion deterministically panics the scheduler thread in `crates/spark-storage/src/high_speed_swap.rs:167`. The first request returns normally; subsequent requests are accepted by the HTTP listener but never complete because the scheduler worker is dead, so the client times out.

## Environment

- Image: `avarok/atlas-gb10:3.0.0` (digest `sha256:9943e54a92b41e2f20673964056084f619d942dc4479b658672cd68acd72fbe1`, pushed 2026-05-06)
- Hardware: NVIDIA DGX Spark (GB10, SM121), driver 595.58.03, CUDA 13.2
- Container: `--privileged` (required for io_uring on Spark; without it `--high-speed-swap` install fails with `Operation not permitted`), `network_mode: host`
- Model: `RedHatAI/Qwen3.5-122B-A10B-NVFP4` from local snapshot (`--model-from-path`)

## Repro

Run with these flags set (other flags trimmed for brevity):

```
--max-seq-len 131072
--max-num-seqs 2
--max-batch-size 2
--gpu-memory-utilization 0.93
--kv-cache-dtype nvfp4
--kv-high-precision-layers auto
--speculative
--num-drafts 1
--mtp-quantization nvfp4
--scheduling-policy slai
--high-speed-swap
--high-speed-swap-dir /var/atlas-hs-swap
--high-speed-swap-gb 16
--high-speed-swap-cache-blocks-per-seq 256
--high-speed-swap-graph true
--swap-space-gb 0
--enable-prefix-caching true
```

Then issue identical chat completions back to back:

```bash
for i in 1 2 3; do
  curl -s -m 60 http://localhost:8000/v1/chat/completions \
    -H 'Content-Type: application/json' \
    -d '{"model":"qwen3.5-122b-a10b","messages":[{"role":"user","content":"In one sentence: what is photosynthesis?"}],"max_tokens":120,"temperature":0}'
done
```

Run 1 returns normally. Run 2 hangs until curl timeout.

## Failure

```
thread '<unnamed>' (158) panicked at crates/spark-storage/src/high_speed_swap.rs:167:13:
[...]

PR #892, opened 2026-09-05
spark-runtime: pin identity and multiplicity oracles in 8 radix_tree checks

TheTom:
## Summary

RST mutation audit of the 25 registered `radix_tree::tests::{adapter,basic}` checks (campaign IDs ATC-1910 through ATC-1934). Twenty-three realistic production mutants were applied and reverted one at a time against `RadixTreeInner::{walk,insert,inc_refs,dec_refs}`, `evict`, `root_for_read`, and the `RadixTree` snapshot/HSS wrappers. Eighteen checks rejected their mutant as written, three are redundant with a stronger neighbour (recorded, left in place), and eight stayed green against a real defect and are repaired here. **No production changes: every surviving mutant traced to a test-oracle weakness, not a product bug.**

Five of the surviving mutants were invisible to the **entire 86-check `radix_tree` registry**, not just this file:

| Mutant (reverted) | What it breaks | Now owned by |
|---|---|---|
| `inc_refs` stops refreshing `last_access` (`inner.rs:255`) | LRU degenerates to FIFO | `test_evict_lru_order` |
| existing node adopts the inserting sequence's block (`inner.rs:353`) | the exact ownership break `InsertAcquired` documents: the node's block loses its ref and eviction hands a live sequence's block back to the pool | `test_insert_idempotent` |
| a KV ref reported on the root-guard arm (`inner.rs:401`) | permanent block leak: ref taken, never handed back by `evict` | `test_sub_block_tokens_ignored` |
| `u32::MAX` disk-id filter deleted from `RadixTree::lookup` (`radix_tree.rs:125-129`) | HSS-off matches falsely signal "HSS engaged" | `test_hss_disk_ref_acquisition_no_op_when_hss_inactive` |
| `partial_block` pushed twice into the acquisition report (`inner.rs:424`) | partial-suffix block pinned for the life of the run | `test_partial_suffix_block_is_owned_and_released` |

## The eight oracle weaknesses (each proven with an executed old-green mutant)

- `test_partial_match` could not tell "stopped at the divergence" from "stopped one block early" because block 3 was never cached. A third block is now cached under one continuation and looked up under another.
- `test_no_match` was a bare `is_empty()` with no control, satisfied by a cache that misses everything. It now also checks `matched_blocks` and hits its own key in the same tree.
- `test_evict_lru_order` inserted and evicted in the same order, so FIFO passed. A cache hit on the older block now separates least-recently-used from oldest-inserted.
- `test_insert_idempotent` asserted only `stats() == (1, 1)`. A repeated insert with a different physical block now pins that the node keeps block 42, per the `InsertAcquired` ownership rule.
- `test_sub_block_tokens_ignored` counted nodes; the count is structurally blind to anything parked on the root and to the acquisition report. It now asserts the whole `InsertAcquired` is empty and the prefix is unfindable.
- `test_hss_disk_ref_acquisition_no_op_when_hss_inactive` was a lone negative. It now asserts the KV side still acquires both blocks and that a lookup reports no disk ids.
- `test_vision_pad_tokens_are_image_blind_collision` restated `test_insert_and_lookup_multi_block`. It now also pins the SSM-snapshot half of the issue #58 collision, which is what turns the key collision into "the previous picture's answer".
- `test_partial_suffix_block_is_owned_and_released` used `contains`, which accepts a duplicated ref report. All three acquisition assertions are exact lists now.

## File split

`basic.rs` was at 463 of the 500-LoC cap. The four issue #17 HSS disk-ref checks moved to `tests/basic_hss.rs`, nested with the same `#[path]` pattern `snapshot_index.rs` uses; they now register as `radix_tree::tests::basic::hss::*`. Registry total is unchanged at 86.

## Redundancy and remaining gaps (recorded, not changed)

- `test_insert_and_lookup_exact` (owner `test_insert_and_lookup_multi_block`), `test_release_decrements_refcount` (owner `test_evict_skips_referenced_nodes`), and `test_hss_disk_ref_acquisition_re_insert_no_double` (owner `test_hss_disk_ref_acquisition_extension`) are killed only by mutants that also kill their owner with equivalent setup; the only available strengthening restates the owner. Left in place as cheap, readable partitions.
- The `ref_count == 0 → 1` restore at `inner.rs:355-357` survives all 86 checks. Every construction of that state on a live node appears unreachable in production (a walk that stops at a zero-ref node reports `matched_tokens` below it, so the following insert re-bumps via `is_seq_owned`). Flagged as a defensive branch with no test and no reachable stimulus.
- `test_ssm_snapshot_partial_match_returns_deepest` is named for deepest-anchor selection but asserts the depth gate (an anchor deeper than the verified prefix is not restored). The gate is load-bearing and it is the sole owner; deepest-wins is owned in `tests/snapshot.rs`.

## Test plan

- [x] `ATLAS_SKIP_BUILD=1 cargo test -p spark-runtime --lib --no-default-features --features metal radix_tree` — 85 passed, 0 failed, 1 ignored (the ignore is lifted in #893)
- [x] unfiltered `spark-runtime` lib suite equals the pristine-main baseline measured on the same host: 268 passed, 35 failed, 7 ignored; all 35 are `metal_backend::tests::*` "Metal: unknown module" (skip-build shader artifact), identical failure set before and after
- [x] `cargo fmt --all -- --check` · `bash scripts/check-license-headers.sh` (2,408 valid, 0 invalid) · `typos` · `git diff --check`
- [ ] `cargo clippy -p spark-runtime --no-default-features --features metal --tests -- -D warnings` fails on 19 pre-existing lints, none under `radix_tree/`, byte-identical with these files reverted to `main`; Linux CI clippy owns this gate
- [ ] Tested against a real model / hardware (not applicable: host-side prefix-cache bookkeeping; no GPU path exercised)

## Related

Sibling audit PRs on disjoint files of the same module: `tests/snapshot.rs` + `tests/snapshot_index.rs` (#893) and `tests/snapshot_lease.rs` + `tests/snapshot_insert_tier.rs` + `tests/snapshot_reap.rs` (#894). All three merged together run 86 passed, 0 failed, 0 ignored.

## CLA

### DIFF
--- a/crates/spark-runtime/src/radix_tree/tests/basic.rs
+++ b/crates/spark-runtime/src/radix_tree/tests/basic.rs
@@ -236,6 +236,27 @@ fn test_insert_idempotent() {
     tree.insert(&tokens, &[42], &[], 16, 0, 0);
 
     assert_eq!(tree.stats(), (1, 1));
+
+    // A node count cannot say WHICH block the surviving node holds. The second
+    // sequence normally does not get its block from the cache (it never
+    // matched, or it restored from a swap file), so it re-inserts the same
+    // token chunk with a DIFFERENT physical block. The node must keep the block
+    // the cache already holds a reference on: adopting the sequence's block
+    // strands that reference, and evicting the node then returns a block a live
+    // sequence still owns. See `InsertAcquired`.
+    let second = tree.insert(&tokens, &[43], &[], 16, 0, 0);
+    assert_eq!(tree.stats(), (1, 1), "still one node");
+    assert!(
+        second.blocks.is_empty(),
+        "no node was created, so no KV ref is owed; got {second:?}"
+    );
+    let m = tree.lookup(&tokens, 16, 0, 0);
+    assert_eq!(
+        m.matched_blocks,
+        vec![42],
+        "the node keeps the block the cache references"
+    );
+    tree.release(&tokens, 16, 0);
 }
 
 #[test]
--- a/crates/spark-runtime/src/radix_tree/tests/basic.rs
+++ b/crates/spark-runtime/src/radix_tree/tests/basic.rs
@@ -7,6 +7,12 @@
 use crate::prefix_cache::PrefixCache;
 use crate::radix_tree::RadixTree;
 
+/// Issue #17 HSS disk-ref acquisition checks — split out to keep this file
+/// under the workspace 500-LoC cap, using the same `#[path]` nesting the
+/// snapshot index tests use.
+#[path = "basic_hss.rs"]
+mod hss;
+
 #[test]
 fn test_insert_and_lookup_exact() {
     let tree = RadixTree::new();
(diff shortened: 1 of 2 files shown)


tbraun96:

PR #893, opened 2026-09-05
spark-runtime: repair 9 unreachable or vacuous prefix-cache snapshot checks

TheTom:
## Summary

RST mutation audit of the 34 registered checks in `radix_tree/tests/snapshot.rs` and the 17 own functions of `tests/snapshot_index.rs` (campaign IDs ATC-1935 through ATC-1968). Thirty realistic production mutants were applied and reverted one at a time against the sub-block matching arms in `inner.rs`, the `RadixTree` lookup wrapper, `SsmSnapshotIndex::{lookup,lookup_tiered,insert,evict_to_tier,promote}`, and `hash_token_prefix`. Twenty-five checks rejected their mutant as written. Nine stayed green against a real defect and are repaired here, including one check that was `#[ignore]`d on a false premise and is now live. **No production changes: every surviving mutant traced to a test-oracle or reachability weakness, not a product bug.**

Four of the surviving mutants were invisible to the **entire 86-check `radix_tree` registry**:

| Mutant (reverted) | What it breaks | Now owned by |
|---|---|---|
| `remainder > 0` guard deleted from the sub-block gate (`inner.rs:198-200`) | a block-aligned request gets the 4-token partial block appended to a 16-token match: block table longer than `matched_tokens` describes | `test_partial_suffix_not_matched_for_full_block_request` |
| `lookup_tiered` hit no longer refreshes `last_access` (`snapshot_tier.rs:100`) | hot anchors age like never-used ones and are evicted first | `eviction_ignores_hit_history` |
| live-session latch removed from `lookup_tiered` (`snapshot_tier.rs:63-66`) | the tail lease never arms on the serving path | `lookup_tracks_live_session` |
| child-supersede stops reporting the retired partial block in `released_blocks` (`inner.rs:375`) | the caller never `dec_ref`s it: permanent block leak | `test_partial_suffix_cleared_when_extended` |

## The nine weaknesses (each proven with an executed old-green mutant)

- `test_insert_without_snapshot` asserted only `lookup`'s default outputs, which also hold when the KV walk matches nothing (an inverted context-chain guard killed 10 siblings, not this). It now asserts the hit first, plus the tier fields.
- `test_partial_suffix_not_matched_for_full_block_request` never reached the arm its name describes: `remainder` in `walk` is the unmatched token count, not `len % block_size`, so the 32-token stimulus was stopped by a different clause (the in-test comment was wrong). A genuinely block-aligned 16-token lookup is added.
- `test_partial_suffix_cleared_when_extended` was ignored as "tests removed behavior". The child-supersedes-partial clearing is live and reports the block through `InsertAcquired::released_blocks`, consumed by `block_mgmt.rs:190`; only the lookup expectation was stale (the sub-block child-key arm now serves the 20-token prefix from the deeper node). Rewritten and un-ignored.
- `test_partial_suffix_multi_block_prefix` asserted `matched_blocks.len() == 25`; returning each node's parent block killed 14 checks but not this one. It now asserts the full block table.
- `test_ssm_snapshot_adapter_isolation_via_tree`: adapter B had no cached KV, so `lookup` short-circuited before the snapshot index and the disjoint radix roots proved the assertion for free. B now caches its own KV, hits 32 tokens, and still gets no snapshot.
- `deep_tail_evicted_without_tail_protect` had no `is_tail` entry in the pool, so `tail_protect` was inert. id 9 is now a tail, with an armed control on the same pool.
- `eviction_ignores_hit_history`: the five hits were decorative because the anchor was oldest either way. A never-hit later save now dies first.
- `lookup_tracks_live_session` asserted the latch only on the retained reference `lookup`, which is dead code in production. It now covers the sessionless must-not-latch arm and the `lookup_tiered` latch plus lease renewal.
- `test_snapshot_index_insert_lookup_roundtrip` looked up at `matched_tokens == token_count`, so the returned depth's source was indistinguishable. It now looks up 48 verified tokens against the depth-32 anchor and pins the empty-candidate `None` at 31.

## What the good checks defended, and known hazards without coverage

Sole owners across all 86 checks: partial-suffix content verification, partial block freed on evict, each sub-block arm's `>=`, the root guard, winner-only recency bump, the recompute metrics, the tiered skip in `lookup`, and un-spill on re-insert. Two hazards are documented in-code and covered by nothing: the tier blob orphaned by a re-insert over a spilled entry (`snapshot_insert.rs:43-64`, knowingly unreclaimed) and the sub-block arms handing back a block whose KV was computed for a longer key (`inner.rs:190-196`, the `ATLAS_PREFIX_SUBBLOCK` A/B lever).

## Test plan

- [x] `ATLAS_SKIP_BUILD=1 cargo test -p spark-runtime --lib --no-default-features --features metal radix_tree` — 86 passed, 0 failed, 0 ignored
- [x] unfiltered `spark-runtime` lib suite: baseline on pristine main 268 passed, 35 failed, 7 ignored; after 269 passed, 35 failed, 6 ignored. The delta is exactly the un-ignored check; the 35 `metal_backend::tests::*` "Metal: unknown module" skip-build failures are an identical set before and after
- [x] `cargo fmt --all -- --check` · `bash scripts/check-license-headers.sh` (2,407 valid, 0 invalid) · `typos` · `git diff --check`
- [ ] `cargo clippy -p spark-runtime --no-default-features --features metal --tests -- -D warnings` fails on 19 pre-existing lints, none under `radix_tree/`, identical with these files reverted to `main`; Linux CI clippy owns this gate
- [ ] Tested against a real model / hardware (not applicable: host-side prefix-cache and snapshot-index bookkeeping)

## Related

Sibling audit PRs on disjoint files of the same module: `tests/adapter.rs` + `tests/basic.rs` (#892) and `tests/snapshot_lease.rs` + `tests/snapshot_insert_tier.rs` + `tests/snapshot_reap.rs` (#894). All three merged together run 86 passed, 0 failed, 0 ignored.

## CLA

### DIFF
--- a/crates/spark-runtime/src/radix_tree/tests/snapshot.rs
+++ b/crates/spark-runtime/src/radix_tree/tests/snapshot.rs
@@ -209,24 +209,53 @@ fn test_partial_suffix_eviction_frees_both_blocks() {
     assert!(evicted.physical.contains(&20));
 }
 
+/// A real child node SUPERSEDES the partial-suffix slot it overlaps: the slot
+/// is cleared and the KV block it held comes back in `released_blocks` so the
+/// caller can drop the cache's reference on it.
+///
+/// Rewritten (was `#[ignore]`d as "tests removed behavior"): the OLD
+/// assertions expected the 20-token prefix to become unmatchable, which the
+/// sub-block child-key arm made false — the deeper node now serves those 20
+/// tokens. The clearing itself is NOT removed behaviour; what it must produce
+/// is the released-block accounting asserted below, and no serving of the
+/// retired block. The partial→partial overwrite arm is covered by
+/// `tests::basic::test_partial_suffix_block_is_owned_and_released`; this is
+/// the distinct child-supersedes-partial arm.
 #[test]
-#[ignore = "tests removed behavior — partial-suffix clearing was replaced \
-            with partial-block-matching during the radix-tree refactor; \
-            assertions need rewriting against the new lookup semantics"]
 fn test_partial_suffix_cleared_when_extended() {
     let tree = RadixTree::new();
-    // Insert 20 tokens (1 full + 4 partial)
+    // Insert 20 tokens (1 full + a 4-token partial held in block 20).
     let tokens_20: Vec<u32> = (0..20).collect();
-    tree.insert(&tokens_20, &[10, 20], &[], 16, 0, 0);
+    let first = tree.insert(&tokens_20, &[10, 20], &[], 16, 0, 0);
+    assert!(
+        first.blocks.contains(&20),
+        "the partial slot takes a ref on block 20; got {:?}",
+        first.blocks
+    );
 
-    // Insert 32 tokens (2 full blocks, extends past partial)
+    // Insert 32 tokens: a real child now covers [16..32), superseding the slot.
     let tokens_32: Vec<u32> = (0..32).collect();
-    tree.insert(&tokens_32, &[10, 30], &[], 16, 0, 0);
+    let extended = tree.insert(&tokens_32, &[10, 30], &[], 16, 0, 0);
+    assert_eq!(
+        extended.released_blocks,
+        vec![20],
+        "the superseded partial block must be handed back, not leaked"
+    );
+    assert!(
+        extended.blocks.contains(&30),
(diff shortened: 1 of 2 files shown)


tbraun96:

PR #894, opened 2026-09-05
spark-runtime: make 4 snapshot lease and tier-insert oracles load-bearing

TheTom:
## Summary

RST mutation audit of the 27 registered checks in `radix_tree/tests/snapshot_lease.rs`, `tests/snapshot_insert_tier.rs`, and `tests/snapshot_reap.rs` (campaign IDs ATC-1969 through ATC-1995). Thirty-five realistic production mutants were applied and reverted one at a time against `session_aware_victim_with_alpha`, `tail_lease_active`, the three `insert*` overwrite arms and the tail-supersede sweep, `lookup_tiered`, `evict_to_tier`, `promote`, `forget_tiered`, and the `PrefixCache` trait defaults. Twenty-three checks rejected their mutant as written. Four stayed green against a real defect and are repaired here. **No production changes: every surviving mutant traced to a test-oracle weakness, not a product bug.**

One surviving mutant was invisible to the **entire 86-check `radix_tree` registry**: deleting `entry.tiered = false` from the `insert_tail_sibling` overwrite arm (`snapshot_insert.rs:167`). A sibling save over a spilled prefix would then strand its fresh HBM slot for the process lifetime, skipped by `lookup` and by both victim scans, so reachable by nothing and freeable by nothing. `insert_tail_sibling_over_a_spilled_entry_hands_back_no_slot` now mirrors the tail-arm check: `!tiered` plus `evict_lru() == Some(slot)`.

## The four oracle weaknesses (each proven with an executed old-green mutant)

- `finish_leaf_not_protected`: the named victim was also the plain-LRU victim, so removing the lease entirely, or inverting within-session recency, stayed green. The pool now holds a fresh shallow non-tail entry with the tail as the oldest, so one answer requires both the lease and recency ordering.
- `all_leased_pool_still_evicts`: `assert!(v.is_some())` is vacuous whenever the lease is inactive; inverting the kill-switch polarity survived. It now asserts `tail_lease_active()`, that every entry is a leased restore point of the live session, the exact oldest-first victim, and that the reclaim happened.
- `sibling_not_session_gated_in_lookup`: its docstring claimed a "general session filter" excludes different-nonzero-session lookers. No such filter exists. Production intentionally makes non-tail entries content-addressed and cross-session safe (the `is_tail` INVARIANT on `SnapshotEntry`, and `snapshot_tier.rs:76-86`); a blanket gate re-added for different-nonzero sessions only survived. The third partition is now asserted and the docstring states the real contract.
- `insert_tail_sibling_over_a_spilled_entry_hands_back_no_slot`: see above.

## What the good checks defended

The `freeable_slot` double-free guard is separately owned per insert arm and per site (overwrite vs supersede sweep) by mutants that kill exactly one check each. `forget_tiered`'s resident guard closes the promote-then-reap race without leaking a slot, and the reap stays out of the eviction counters so it cannot shorten a tail lease. Two notes, no code changed: the `session_aware_victim` doc (`snapshot.rs:334-335`) says the leased set is bounded at one entry, but with siblings leased the bound is two (`new_tail_sweeps_old_tail_and_sibling` asserts exactly that); the tier-blob orphan on every overwrite arm (`snapshot_insert.rs:52-64`) is already documented in-code as a tracked follow-up.

## Test plan

- [x] `ATLAS_SKIP_BUILD=1 cargo test -p spark-runtime --lib --no-default-features --features metal radix_tree` — 85 passed, 0 failed, 1 ignored (the ignore is lifted in #893)
- [x] unfiltered `spark-runtime` lib suite equals the pristine-main baseline measured on the same host: 268 passed, 35 failed, 7 ignored; all 35 are `metal_backend::tests::*` "Metal: unknown module" (skip-build shader artifact), identical failure set before and after
- [x] `cargo fmt --all -- --check` · `bash scripts/check-license-headers.sh` (2,407 valid, 0 invalid) · `typos` · `git diff --check`
- [ ] `cargo clippy -p spark-runtime --no-default-features --features metal --tests -- -D warnings` fails on 19 pre-existing lints, none under `radix_tree/`, identical with these files reverted to `main`; Linux CI clippy owns this gate
- [ ] Tested against a real model / hardware (not applicable: host-side snapshot-index bookkeeping)

## Related

Sibling audit PRs on disjoint files of the same module: `tests/adapter.rs` + `tests/basic.rs` (#892) and `tests/snapshot.rs` + `tests/snapshot_index.rs` (#893). All three merged together run 86 passed, 0 failed, 0 ignored.

## CLA

### DIFF
--- a/crates/spark-runtime/src/radix_tree/tests/snapshot_insert_tier.rs
+++ b/crates/spark-runtime/src/radix_tree/tests/snapshot_insert_tier.rs
@@ -104,7 +104,16 @@ fn insert_tail_rehomes_a_spilled_entry_to_hbm() {
     );
 }
 
-/// `insert_tail_sibling` — the third path, same contract.
+/// `insert_tail_sibling` — the third path, same contract, AND the same
+/// re-home obligation `insert_tail` carries (see the test above).
+///
+/// The re-home half was the suite's blind spot: deleting `entry.tiered =
+/// false` from this arm survived the WHOLE 86-check `radix_tree` registry.
+/// `reinsert_unspills` covers the plain-`insert` arm and
+/// `insert_tail_rehomes_a_spilled_entry_to_hbm` covers the tail arm; nothing
+/// covered this one, so a sibling save over a spilled prefix could strand its
+/// fresh slot — skipped by `lookup` and by both victim scans, hence reachable
+/// by nothing and freeable by nothing — for the process lifetime.
 #[test]
 fn insert_tail_sibling_over_a_spilled_entry_hands_back_no_slot() {
     let (mut idx, ph, freed) = spilled(0xD4, 8);
@@ -114,6 +123,15 @@ fn insert_tail_sibling_over_a_spilled_entry_hands_back_no_slot() {
     assert_eq!(
         displaced, None,
         "slot {freed} was already freed at spill time"
+    );
+    assert!(
+        !idx.entries[0].tiered,
+        "a fresh HBM save re-homes the prefix; leaving it `tiered` strands slot 14"
+    );
+    assert_eq!(
+        idx.evict_lru(),
+        Some(14),
+        "a re-homed sibling must be evictable — otherwise its slot leaks"
     );
 }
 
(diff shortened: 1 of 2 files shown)


tbraun96:

PR #898, opened 2026-09-05
spark-server: bare tag-style call closed before its parameters streamed in

TheTom:
## Summary

RST mutation audit of all 28 registered checks in `tool_parser/tests/group_b.rs` (campaign IDs ATC-2100 through ATC-2127). Twenty-seven checks are GOOD, each verified by an executed, restored production mutant failing for the intended reason. One check exposed a **real streaming product bug**, fixed here in its own commit.

## Production bug: a bare tag-style call that closes its name before its parameters stream in

When a bare `<function>NAME</function><parameters>...</parameters>` call arrives split across stream chunks so that `</parameters>` has not yet streamed in, `bare_function_end` misread the name-closing `</function>` as the end of a zero-argument call. `process()` then delivered a tool call with empty arguments, and the orphaned `<parameters>...` text — which has no `<function` prefix for anything to recognize — leaked verbatim into the next chunk's client-visible content. Token-by-token delivery makes this split the realistic case, not the exception.

The fix holds the block incomplete once `<parameters>` has opened until its close (or end-of-stream via `flush()`) arrives, mirroring the `</parameters>`-before-`</function>` precedence already used when both are present. Worst case (close never arrives, e.g. max_tokens cutoff) degrades to `flush()` recovery — which the repaired test proves preserves the argument value.

Discovery was empirical: a two-`process()`-call probe showed the zero-argument call plus `Content("<parameters>...")` leakage on pre-fix code; the repaired test fails on pre-fix code at the intended assertion and passes with the fix.

## Test repair

`streaming_bare_function_flush` claimed to exercise `flush()` but fed the entire block in one `process()` call — instrumentation showed `flush()` always returned empty for that fixture, and the oracle never inspected the parsed arguments. The stimulus now splits the stream before `</parameters>`, asserts no premature tool call, and requires `flush()` to recover the call with `x == 1` intact.

## The 27 good checks (mutants executed and restored, highlights)

- Field-bleed guards: disabling the `</function>`-before-next-`<parameter=` guard fails both consecutive-function checks through their distinct entry points.
- `normalize_tool_name` is a live SSOT: disabling `=`-truncation fails three checks; disabling colon-namespace stripping fails all six namespaced-name checks across their six distinct formats — each guards a different parser wiring against forgetting the normalizer.
- The unclosed-param check's doc-comment credits `advanced_to_func_close`, but disabling that exact line kills zero tests in the whole 168-test suite (dead code — the top guard fires first for every reachable shape); a boundary-shift mutant on the real mechanism uniquely fails the check. The stale attribution is recorded as a nonclaim; removing dead code was out of scope.
- `parse_mike_screenshot_format` never asserts on `arguments` for its elaborate malformed fixture; the actual parse was traced and is defensible, so this is recorded as a weak-oracle note rather than repaired with an arbitrary expected value.

## Test plan

- [x] `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo test -p spark-server --bin spark --no-default-features --features metal tool_parser` (168 passed, 0 failed, 1 ignored — same count as pre-audit baseline)
- [x] `bash scripts/check-license-headers.sh` (0 invalid)
- [ ] `cargo clippy -p spark-server --tests --no-default-features --features metal -- -D warnings` fails only in `spark-runtime` metal files untouched by this diff, byte-identical on pristine main (verified); Linux CI clippy owns this PR's gate
- [ ] Tested against a live model stream (not run: the split-chunk stimulus reproduces the streaming shape at the detector seam; no scheduler pass was exercised)

`spark-server` has no lib target and its default features need Linux-only `spark-storage` APIs, so the macOS runner is the `--bin spark --no-default-features --features metal` form above (recorded in the audit ledger as a testability note). All functional ties are static unit-level parser/detector calls; nothing here proves scheduler-level delivery ordering.

## CLA

### DIFF
--- a/crates/spark-server/src/tool_parser/parse_single_b.rs
+++ b/crates/spark-server/src/tool_parser/parse_single_b.rs
@@ -270,7 +270,20 @@ pub(super) fn bare_function_end(text: &str) -> Option<usize> {
         if let Some(p) = text.find("</parameters>") {
             return Some(p + "</parameters>".len());
         }
-        // No parameters block — just <function>NAME</function>
+        // A `<parameters>` block has been OPENED but its close hasn't
+        // streamed in yet. Falling through to the "no parameters block"
+        // branch below would misread the name-closing `</function>` (which
+        // necessarily appears before `<parameters>` in this shape) as the
+        // end of a zero-argument call, truncating the block right before
+        // the params. The leftover `<parameters>...` text then has no
+        // `<function` prefix for the caller to recognise, so on the next
+        // `process()` call it falls through to plain content emission and
+        // leaks the raw XML into the client-visible text instead of being
+        // parsed as arguments. Wait for more tokens instead.
+        if text.contains("<parameters>") {
+            return None;
+        }
+        // No parameters block at all — just <function>NAME</function>
         if let Some(p) = text.find("</function>") {
             let after = p + "</function>".len();
             // Check if there's a trailing </function> (some models emit an extra one)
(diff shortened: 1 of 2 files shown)


tbraun96:

PR #900, opened 2026-09-05
ci: route test-macos-metal to an opt-in macOS burst runner

tbraun96:
## What

`test-macos-metal` is the only job in the repo that needs a Mac. This routes it to an opt-in self-hosted Apple Silicon runner when one is available, and leaves it on GitHub-hosted `macos-14` otherwise.

```yaml
runs-on: ${{ github.event.pull_request.head.repo.fork && 'macos-14' || vars.MACOS_RUNNER_LABEL || 'macos-14' }}
```

## Why it is safe on a public repo

**Fork PRs are pinned to `macos-14`, unconditionally.** A self-hosted runner executes workflow code as the host user, so untrusted fork code must never reach it. `head.repo.fork` is empty for `push` / `merge_group` / `workflow_dispatch`, so those fall through to the variable.

Worth being explicit about the boundary this draws: reviewed code that lands via push or the merge queue *does* run on the Mac. That is the same trust level as merging — but it is a real property of the change, not an accident.

## Why the health job is required

The burst Mac is Off most of the time by design. If `MACOS_RUNNER_LABEL` outlived the runner, every Metal job would sit in *waiting for a runner* against a label nobody answers.

So `cmd-runner-health.yml` now manages `MACOS_RUNNER_LABEL` exactly the way it already manages `CMD_RUNNER_LABEL` — set only while a runner carrying the burst label is genuinely online, cleared back to `macos-14` when it goes away. Because offline is the *resting state* for a burst runner rather than a fault, `macos-health` reports transitions as notices instead of warnings. It also falls back to `POST` on first run, since `PATCH` fails on a variable that does not exist yet.

## Rollout

**This PR is inert on its own.** With `MACOS_RUNNER_LABEL` unset the expression falls through to `macos-14`, exactly as today. Nothing changes until a runner with the `atlas-macos-burst` label comes online and the health job sets the variable.

The runner itself is registered and currently **Off** with its scheduling label absent, so it cannot pick up work yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -374,7 +374,18 @@ jobs:
     # suppresses the implicit success() on `needs` — so this job RUNS. Written
     # as `== 'false'` it would fail CLOSED and silently un-gate Rust changes.
     if: ${{ !cancelled() && needs.changes.outputs.web_only != 'true' }}
-    runs-on: macos-14
+    # Runner routing, in priority order:
+    #   fork PR       -> always macos-14. atlas is public, and a self-hosted Mac
+    #                    must never execute code from an untrusted fork. This is
+    #                    the rule that makes the burst runner safe at all.
+    #   everything else -> MACOS_RUNNER_LABEL when set (the on-demand Mac),
+    #                    otherwise macos-14.
+    # `head.repo.fork` is empty for push / merge_group / workflow_dispatch, so
+    # those fall through to the variable. The variable is managed by
+    # cmd-runner-health.yml: it is set only while a runner carrying the burst
+    # label is actually online, and cleared back to macos-14 when the Mac goes
+    # Off — otherwise a job would queue forever against a label nobody answers.
+    runs-on: ${{ github.event.pull_request.head.repo.fork && 'macos-14' || vars.MACOS_RUNNER_LABEL || 'macos-14' }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
       - uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable
--- a/.github/workflows/cmd-runner-health.yml
+++ b/.github/workflows/cmd-runner-health.yml
@@ -83,3 +83,71 @@ jobs:
             # a red health check every 15 minutes trains people to ignore it.
             echo "::warning title=Cannot switch runners automatically::the token cannot write CMD_RUNNER_LABEL (needs 'administration: write' on the certification App). Set it by hand to '$want'."
           fi
+
+  macos-health:
+    name: macOS burst runner health
+    # Same idea as the job above, for the opt-in macOS burst runner that
+    # ci.yml's test-macos-metal job routes to. The crucial difference: for a
+    # BURST runner, offline is the normal resting state, not a failure. The Mac
+    # is a workstation that opts in for a while and then goes Off again, so a
+    # cleared variable is routine and is reported as a notice, never a warning.
+    #
+    # Without this, turning the Mac Off would leave MACOS_RUNNER_LABEL pointing
+    # at a label no online runner carries, and every Metal job would sit in
+    # "waiting for a runner" instead of falling back to GitHub-hosted.
+    runs-on: ubuntu-latest
+    steps:
+      - name: Mint the certification App token
+        id: apptoken
+        if: vars.CERTIFICATION_APP_CLIENT_ID != ''
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3
+        with:
+          app-id: ${{ vars.CERTIFICATION_APP_CLIENT_ID }}
+          private-key: ${{ secrets.CERTIFICATION_APP_PRIVATE_KEY }}
+
+      - name: Point the Metal job at a runner that answers
+        env:
+          GH_TOKEN: ${{ steps.apptoken.outputs.token || github.token }}
+          REPO: ${{ github.repository }}
+          WANT: atlas-macos-burst
+          FALLBACK: macos-14
+        run: |
+          set -uo pipefail
+          # The burst controller removes the label before it stops the service,
+          # so "online AND carrying the label" is the only honest test of
+          # whether a job routed here would actually be picked up.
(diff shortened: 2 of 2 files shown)


tbraun96:
Held out of the CI stack I am assembling tonight, because `main` now refutes the routing this PR adds.

Rebasing it onto `main` (`569ffd092`) conflicts in `ci.yml` on exactly the line it changes, and main's side carries the reason:

> #901 moved this to the self-hosted Apple Silicon box. Then TheTom's `AVAROK_SKIP_BUILD="0"` made the step actually BUILD the metal kernel set instead of stubbing `metallib_modules()` to an empty list — and the build immediately failed there:
>
>     xcrun: error: unable to find utility "metal", not a developer tool or in PATH
>
> `apple-48gb-metal` carries the Command Line Tools but not the Metal compiler, which ships with Xcode. […] ★ Move this back to [self-hosted, macOS, ARM64] and delete the advisory job the moment `xcrun -f metal` answers on that box.

`test-macos-metal` is a required context that compiles 43 `.metal` sources. Routing it to `MACOS_RUNNER_LABEL` sends it to the same unprovisioned Mac, so landing this as-is reintroduces a failure main documents and pins against — and it would do so silently whenever the burst runner happens to be online.

The second half of the diff (`macos-health` in `cmd-runner-health.yml`, which clears the variable when the Mac goes Off) is sound on its own, but nothing reads the variable once the routing is reverted.

Two ways forward, and this one is yours: install Xcode on `apple-48gb-metal` so `xcrun -f metal` answers and then land this whole PR (it is the cheaper runner bill), or keep the required job hosted and close this. I have not changed the PR.

tbraun96:
Unblocked and re-cut onto `main` — the owner installed Xcode on `apple-48gb-metal` today, `xcrun -f metal` answers, and the metal test target builds there. That is exactly the condition main's comment named:

> ★ Move this back to [self-hosted, macOS, ARM64] and delete the advisory job the moment `xcrun -f metal` answers on that box.

I did not take it literally, and the comment beside the change says why: `cargo test --features metal (macOS aarch64)` is a **required** context, and the burst Mac is a workstation that opts in and out. Pinned to a label no online runner carries, a required job does not go red — it **queues**, reports nothing, and the PR is unmergeable with nothing showing as broken. So this PR's original design (route by `MACOS_RUNNER_LABEL`, kept current by `cmd-runner-health.yml`, forks always hosted) is the one that landed, with main's history and that reasoning preserved above it.

The advisory `metal-device-parity` job stays rather than being deleted: it *executes* kernels on the real GPU, which is a different claim from compiling them, and with Xcode present it now does so for real instead of warning. It can be promoted to required once the Mac's availability is something we can rely on.

## What I found while doing it

`metal-device-parity` ran on `[self-hosted, macOS, ARM64]` with **no same-repo guard**. atlas is public, so a fork's `pull_request` checked out and executed its own code on that Mac. Neither the workflow nor `assert-cmd-runner-safe.py` covered the macOS pool — that guard has watched `atlas-pr-cheap` and `atlas-pr-rust` since the fleet gate landed and never looked at this one.

Both are fixed here. The job is same-repo only, and the guard now refuses any job that reaches the Mac under a `pull_request` trigger without a same-repo guard in `runs-on` or in its `if`:

```
ci.yml:metal-device-parity runs on the self-hosted macOS box under a pull_request
trigger without a same-repo guard in runs-on or in the job's `if` — a fork's PR
would execute on that Mac. Route forks to macos-14, or add
github.event.pull_request.head.repo.full_name == github.repository.
```

That is the control: with the guard restored the check passes, and removing it fails with the reason above.

A fork PR now gets no real-device leg. The required job still compiles all 43 `.metal` sources for it on hosted macOS, so device coverage is advisory for everyone rather than a hole for some.

`certification-selftest.sh` 242/242 · all five workflow guards pass · no perf path touched, so no campaign.

tbraun96:
## The model is on the box — and the legs were never running

`hf download` put it in the HF cache, not `~/models/Qwen3.5-4B-MLX-8bit`, which the resolver now handles:

```
resolver: real-model legs will run against /Users/nologik/.cache/huggingface/hub/
  models--mlx-community--Qwen3.5-4B-MLX-8bit/snapshots/5319bbbe4f1cbe6c0b3c80f4f7de4f0338c3906d
avarok-kernels: 43 kernels (metal, qwen3-5-4b-vlm-mlx-int8, mlx_int8)
test result: ok. 36 passed; 0 failed; 5 ignored
```

**My own guard was measuring nothing.** It counted `skipping:` lines and reported `0`, on a run where all five real-model legs never executed — they carry `#[ignore = "requires local copy of mlx-community/…"]`, so libtest skips them *before* the body, and the suite still says `ok`. Fixed: model present → `--include-ignored`, and a warning if anything is still ignored; model absent → say plainly that those legs did not run. The step summary now carries ignored / in-body-skips / model-path.

## Running them found a real defect

With the legs actually executing on `apple-48gb-metal`, **40 pass and one fails**:

```
metal_real_model_chain_norm_then_qproj
rms_norm + gemv chain: max |kernel - cpu| = 1.4765625     (tolerance 0.1)
```

It is **not precision**. The GPU writes that intermediate into a bf16 buffer while the test's reference keeps it in f32 — the obvious explanation — so I rounded the reference to bf16 and re-ran: the delta came back **bit-identical** (1.4765625). Refuted.

What it actually is, from the distribution:

```
PROBE rows=64 over_tol=30 mean_abs=0.26294327
PROBE row 55: cpu=-1.4765625 gpu=0 |d|=1.4765625
PROBE row 42: cpu=1.390625  gpu=0 |d|=1.390625
PROBE row 51: cpu=-1.2890625 gpu=0 |d|=1.2890625
PROBE row 11: cpu=-1.28125  gpu=0 |d|=1.28125
```

**30 of 64 output rows are exactly 0.0** on the GPU, scattered (6, 11, 26, 42, 51, 55 among the worst) — not a truncated tail. Both halves pass alone against the same real weights (`metal_mlx_int8_gemv_real_model_q_proj`, `metal_mlx_int8_dequant_real_model`); only the chained launch drops rows. That points at the chained gemv's row coverage over a 64-row subset, not at the arithmetic. I stopped there — that is kernel work, and yours to schedule.

Left failing on purpose: the job is advisory, so the finding is visible without blocking, and it is a defect that had never run in CI because the legs were ignored *and* the model was absent. Everything else here (routing, the fork hole, the guard, the resolver) is green: self-test 242/242, five guards pass, no perf path.

PR #895, opened 2026-09-05
kernels: Hopper (sm_90a) and B200 (sm_100a) targets, device-arch preflight, PTX compile gate

TheTom:
Hopper (`sm_90a`) and B200 (`sm_100a`) kernel targets, architecture preflight, strict PTX assembly checks, and paired Atlas/vLLM campaign tooling. Campaign plans and evidence are tracked in #899.

- Add target metadata and inherited kernels for H100/H200 and B200/GB200, including the first H200 Qwen3.8-27B target. Guard unsupported warp blockscale instructions and declare expected absences; the dense GB10 PTX remains unchanged.
- Preserve the compiled PTX architecture and refuse incompatible devices before weight/module loading. Kernel-check JSON carries `compiled_arch` and numeric `device_cc`, including early refusal.
- Distinguish integrated and discrete GPU memory accounting in both allocation budgeting and the watchdog/TUI poll. Includes the two reviewed fixes from #904; unknown integration never authorizes substituting host RAM.
- Add single-node rank launching, ownership-safe cleanup, strict nvcc→ptxas ledgers, datacenter Dockerfiles and x86_64 CI artifacts with build identity.
- Add per-cell serve/boot/coherency/ladder orchestration and artifact validation. Preserve nested refusals; failed preflight prevents every engine/audit/container path, and a failed kernel audit prevents normal launch. Coherency requests and certification match the measured thinking policy.
- Render 31 vLLM and 19 Atlas profiles, including H100 and H200 Qwen3.8-27B pairs, scored Kimi context49152 and B200 MiniMax NVFP4. Pin primary revisions, record intended Atlas pins, and refuse unproven external-draft image support.
- Support direct Linux process execution for provider-supplied containers, using the existing recipe argv, explicit pinned snapshots and stable model aliases. Audit and serve share the same allowlisted environment. Preserve PID/start-time/group/run-marker ownership through signals and worker process-title changes.
- Refuse occupied process endpoints before launch; prove listener and accepted-socket ownership after readiness and before the ladder using the existing process identity policy.
- End native boot probes promptly when the verified API leader exits, including while an HTTP response is held open. Retain partial probe evidence and the normal runner cleanup path. Treat an exiting procfs leader as absent during cleanup while retaining permission and I/O refusals.
- Add an optional Linux whole-cell deadline that records expiry before signalling the owned runner, retains the existing cleanup path and reports unconfirmed cleanup after its grace expires.
[...]

PR #904, opened 2026-09-05
spark-runtime: stop treating host MemAvailable as device free memory on discrete GPUs

TheTom:
## Bug

`free_memory_cu()` (backend) and `cuda_free_memory_bytes()` (OOM watchdog and TUI gauge) both computed device free memory as `max(cuMemGetInfo free, host /proc/meminfo MemAvailable)`. That is right on GB10, where GPU memory *is* host memory and `cuMemGetInfo` under-reports reclaimable cache. On any discrete GPU it is wrong: host RAM is unrelated to the card.

Reproduced on an x86_64 box with 3× RTX PRO 6000 Blackwell (95 GiB each), 1 TB host RAM, driver 595.71, CUDA 13.0, serving Nemotron-3-Nano NVFP4 with default flags:

```
KV budget ledger implausible (ledger 35.8 GB, used 0.0 GB) — using raw used_so_far
KV cache: 95.0 GB total × 90% util = 85.5 GB budget; 0.0 GB pre-KV + 4.6 GB reserve → 80.8 GB for KV
Error: Failed to build model
Caused by: cuMemAlloc_v2 failed: status 2, requested 4683632640 bytes (device reports 4280.2 MB free / 95.0 GB total)
```

`free_memory()` returned ~990 GB for a 95 GB card, so `used_so_far = total − free` was 0, the ledger sanity gate rejected the correct 35.8 GB, and the KV pool was sized as if no weights were resident. Same failure at `--gpu-memory-utilization 0.6`. Every H100/H200/B200 deployment would hit this. The watchdog half means the OOM watchdog can never trip on a discrete GPU, and the TUI memory gauge shows host RAM.

## Fix

Substitute MemAvailable only when the device reports `CU_DEVICE_ATTRIBUTE_INTEGRATED` (18) = 1. Measured: **1 on GB10** (DGX Spark), **0 on RTX PRO 6000**. (`CU_DEVICE_ATTRIBUTE_PAGEABLE_MEMORY_ACCESS` is 1 on both, so it is not a usable discriminator.)

- Commit 1: pure `effective_free_bytes(cu_free, mem_available, integrated)` used by `free_memory_cu()`; `device_is_integrated()` via `cuCtxGetDevice` + `cuDeviceGetAttribute`, failing loudly on driver error. The `build.rs` KV-budget ledger gate now distinguishes "ledger larger than device-implied usage" (which means `free_memory()` lied; charge the ledger, the safer figure) from "ledger empty", and says which.
- Commit 2: the watchdog poll uses the same helper through `polled_free_bytes`, where an unanswerable integrated query (`None`) is treated as **discrete**: an early watchdog exit is recoverable, a watchdog that never fires is not. Doc comments rewritten to state the rule.

Red-first unit tests for both (discrete ignores host memory; integrated takes the max; `None` never inflates). Verified on the card after the fix, default flags:

```
KV budget self-relative (ledger): Atlas-own 35.8 GB live in the alloc ledger; 20.0 GB of co-tenant/page-cache use excluded
KV cache: 95.0 GB total × 90% util = 85.5 GB budget; 35.8 GB pre-KV + 4.6 GB reserve → 34.4 GB for KV
kernel check PASSED for (nemotron-3-nano-30b-a3b, sm_120, nvfp4): 156 lookups, 9 expected-absent
```

## Gates

`cargo test -p spark-runtime -p spark-model -p spark-server` (GPU-free, `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000`): spark-runtime 284/284, spark-model all pass, spark-server lib 104/104; `--bin spark` has 3 failures (`bench_selfstart` pin test, two `tui::bench_variants` tests) that reproduce identically on unmodified `main`, so they are pre-existing and unrelated. clippy `-D warnings`, fmt, rustdoc `-D warnings`, typos all clean. Files stay under the 500-line cap.

Not GB10-visible: on GB10 `INTEGRATED` is 1, so both code paths behave exactly as before. The same first commit is also on #895 (c4c1b562) and will drop out of that branch on rebase once this merges.

### DIFF
--- a/crates/spark-model/src/factory/build.rs
+++ b/crates/spark-model/src/factory/build.rs
@@ -439,11 +439,34 @@ pub fn build_model(
                 gib(used_so_far - ledger_live),
             );
             used_so_far = ledger_live;
-        } else {
+        } else if ledger_live > used_so_far {
+            // Not "the ledger is implausible" — it is the DEVICE figure that
+            // cannot be true. The ledger counts only allocations this backend
+            // made and still holds, so real device usage is always at least
+            // `ledger_live`; `used_so_far` (total − free_memory()) coming out
+            // SMALLER means `free_memory()` over-reported free memory. That is
+            // exactly what a discrete GPU did when `free_memory()` substituted
+            // host MemAvailable: ~990 GB "free" on a 95 GB card → used_so_far
+            // 0 → a KV pool sized as if nothing were allocated → OOM at
+            // cuMemAlloc. Charging the ledger is strictly more conservative
+            // than charging the smaller (impossible) device-derived figure:
+            // it can only shrink the KV budget, never grow it.
             tracing::warn!(
-                "KV budget ledger implausible (ledger {:.1} GB, used {:.1} \
-                 GB) — using raw used_so_far",
+                "KV budget: free_memory() looks wrong — the alloc ledger holds \
+                 {:.1} GB but the device implies only {:.1} GB used, and real \
+                 device usage can never be below the ledger. Charging the \
+                 ledger's {:.1} GB (the larger, safer figure) instead.",
                 gib(ledger_live),
+                gib(used_so_far),
+                gib(ledger_live),
+            );
+            used_so_far = ledger_live;
+        } else {
+            // ledger_live == 0: nothing ledgered yet (or a backend that does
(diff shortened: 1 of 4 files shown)


tbraun96:

PR #905, opened 2026-09-05
fix(cert): never link the template certificate — it publishes placeholder certifiers

tbraun96:
## The defect

The merged-PR certificate comment falls back to the **committed sample image** when a render or upload fails. That sample carries the placeholder names it was designed with — **`m-ferraro`** on the stamp ribbon, **`a-hoffmann`** on the seal — over a fixed author line.

So the fallback does not degrade the picture. It **states that fictional people certified this PR**, and credits the wrong author.

Reported on **#901**: opened by @TheTom, ribbons read `m-ferraro` / `a-hoffmann`, author reads `tbraun96`.

**It is not specific to that PR.** `bot-cards` held **zero `pr-*.png` objects** as of 2026-09-05 — the render/upload has never once succeeded, so the fallback has fired on *every certificate ever posted*.

The block's own comment already knew:

> *"Embedding the committed template directly would publish a certificate carrying the placeholder names it was designed with, which is worse than no image."*

The fallback did it anyway.

## The fix

No render, no image. The certificate comment is **still posted** — a missing capability must never swallow it, which is the reason the fallback existed — but with an honest line instead of a fabricated one:

> _(The certificate image could not be rendered for this PR. The gate records themselves are public — see the checks on the merge commit.)_

## The test asserted the defect

`certification-selftest.sh` required the buggy behaviour:

```
ok "control: and it points at the generic image, not a render that never happened"
```

Inverted, plus a row asserting the explanation is present. **Both go red on the pre-fix workflow** — verified by reverting the one line: `135 passed, 2 failed`; restored: `137 passed, 0 failed`.

## Recorded, not fixed here

The render/upload has **never** succeeded in production. Whether that is missing `librsvg2-bin`/`segno` on the runner or missing `contents:write` on `bot-cards` is a separate diagnosis — `certification-preflight.yml` probes for exactly the latter. This PR makes the failure honest; it does not yet make the picture appear.

### DIFF
--- a/.github/scripts/certification-selftest.sh
+++ b/.github/scripts/certification-selftest.sh
@@ -673,6 +673,28 @@ STUB
     fi
   fi
 
+  # The certificate IMAGE only exists if the PNG reaches bot-cards, and that
+  # PUT needs contents:write. The App did not have it (certification-preflight's
+  # permission probe fails), which is why bot-cards held zero pr-*.png and every
+  # certificate went imageless. The job's own grant is the guarantee, so pin it.
+  if python3 - <<'PY'
+import sys, yaml
+d = yaml.safe_load(open(".github/workflows/certification-bot.yml"))
+perms = (d["jobs"]["state"].get("permissions") or {})
+sys.exit(0 if perms.get("contents") == "write" else 1)
+PY
+  then
+    ok "the bot job grants itself contents:write — the certificate image can be uploaded"
+  else
+    bad "the bot job lacks contents:write — the certificate PNG can never reach bot-cards"
+  fi
+  # ...and the upload must actually FALL BACK to that token, not rely on the App.
+  if grep -q 'put_cert "${WORKFLOW_TOKEN:-}"' .github/workflows/certification-bot.yml; then
+    ok "and the upload retries under the workflow token when the App lacks the grant"
+  else
+    bad "the upload never falls back to the token that is guaranteed to have contents:write"
+  fi
+
   runbot pr-certification-stage-1 "" 0
   posted && ! patched && ok "no prior comment -> posts one" || bad "no prior comment -> did not post exactly one"
   # CONTROL: with a prior comment it must EDIT, never post a second. A thread of
--- a/.github/scripts/certification-selftest.sh
+++ b/.github/scripts/certification-selftest.sh
@@ -656,10 +656,20 @@ STUB
     else
       bad "control: a missing render tool swallowed the certificate entirely"
     fi
+    # It must NOT fall back to the committed template. That sample carries the
+    # placeholder certifier names it was designed with (m-ferraro / a-hoffmann)
+    # over a fixed author line, so linking it tells the world that fictional
+    # people certified this PR. Observed on #901 (opened by @TheTom, sample says
+    # tbraun96), and on every certificate before it: bot-cards held zero pr-*.png.
     if grep -q 'certificate-merged.png' "$TMP/bcalls"; then
-      ok "control: and it points at the generic image, not a render that never happened"
+      bad "control: the fallback linked the template — that publishes placeholder certifiers"
     else
-      bad "control: it linked a rendered certificate that was never produced"
+      ok "control: no render, no image — the template's placeholder names are never published"
+    fi
+    if grep -q 'could not be rendered' "$TMP/bcalls"; then
+      ok "control: and it says why the picture is missing"
+    else
+      bad "control: the image vanished with no explanation"
     fi
   fi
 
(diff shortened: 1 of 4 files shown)


tbraun96:
**Scope corrected — this now makes the certificate image actually appear, not just stop lying.**

The first commit only removed the fabricated fallback. That was half the job: the point is a real certificate with real values. Traced the whole chain:

| link | state |
|---|---|
| bot job checks out the repo → **SVG template is present** | ✅ fine |
| `rsvg-convert` + `segno` on the self-hosted runner (`atlas-cmd`) | ✅ both installed — `render_ok=1` holds |
| `render-certificate.py` substitution | ✅ **correct** — on an empty stamp/seal user it *hides the ribbon*, so no placeholder can leak from the renderer |
| PUT of the PNG to `bot-cards` | ❌ **this is the failure** |

The upload needs `contents:write`. The certification App does not have it — `certification-preflight.yml`'s *"Probe every permission the pipeline needs"* step **fails** — and the PUT is non-fatal by design, so `bot-cards` held **zero `pr-*.png`** and every certificate fell back to the sample.

**Fix:** the job now grants itself `contents: write` and the upload retries under `GITHUB_TOKEN` when the App token cannot write. The App stays preferred, so commits keep its identity wherever the grant exists.

Safe under `pull_request_target`: this job never touches PR code — it checks out the default branch with `persist-credentials: false`, and every PR-derived string reaches the SVG through `html.escape`.

**On #901 specifically:** it has *no Stamp or Seal check run at all* (merged with `PR benchmark gate [failure]`), so those ribbons will correctly render **hidden** rather than showing anyone's name — and the author line will read **@TheTom**, the opener. The `m-ferraro`/`a-hoffmann`/`tbraun96` text came entirely from the sample image.

**Two traps hit while writing this, both caught by the suite:**
- an `&& return 0` upload shape is **invisible** to `assert-preflight-covers-writes.py`, which keys on `||` in the same statement to know a write swallows its failure — restructured so the guard still sees it;
- the step runs under `set -euo pipefail`, so a bare `$WORKFLOW_TOKEN` reference **killed the entire certificate comment** when unset — the exact failure mode this PR exists to prevent. Now `${VAR:-}`.

Suite: **113 passed, 0 failed**; both new guards proven able to fail (removing the grant → red; removing the fallback → red).

tbraun96:
## Scope grown — an autonomous audit found the reason nobody noticed, plus a second path to wrong data

### The alarm was reporting "verified" while it failed

On commit `de42fb155e` there are **two** check runs from the same job:

```
Certification preflight | success | "App permissions verified"
certification preflight | failure | -
```

The `checks:write` probe minted the green mark **0.5 s before the job errored out on `contents:write`**. That is why this bug survived for days: the pipeline's own permission alarm published "App permissions verified" on the run that proved the opposite. The mark is now minted `in_progress` and PATCHed with the real conclusion at the end, so a run that dies mid-way leaves a visibly *unresolved* check rather than a green one.

### A second, independent path to wrong data — which this PR did not previously cover

The renderer skipped substitution on an empty value, and six certificate fields have **no group to hide**, so the artwork's own specimen data stayed on the page. Reproduced on this tree:

```
$ render-certificate.py --title "" --commit "" ...
rendered commit: 9d4e1f07c2      # the template's specimen commit
```

Every input is read via `gh api … 2>/dev/null`, so "the API did not answer" is indistinguishable from a real value. The renderer now **refuses** an empty title/commit, and the bot checks first so the refusal costs the *picture*, never the certificate (a bare non-zero under `set -e` would have killed the step before the comment POST).

### Two of the tests guarding this path could not fail

An audit sabotaged seven existing selftest checks. Two stayed green — **both on the certificate path**:

- `one author -> one visible slot` — deleting the entire author-hiding branch left `110 passed, 0 failed`. `vis()` counts slots without `display="none"`, and slots 2/3 ship hidden, so it could only ever see the `show()` direction. Slot 1 ships **visible carrying `tbraun96`**, and `--authors ""` is reachable. Now caught: `FAIL no authors -> 1 visible slots, drawing the specimen's name`.
- `the state marker is written into the comment` — a substring grep of the step's own source. Dropping the `:$STATE` suffix (which breaks the lookup, so the bot posts a new comment per event forever) did not register.

Suite: **127 passed, 0 failed**, up from 110. Every new guard proven able to fail.

### Still needs a human — exactly what to click

The upload retry under `GITHUB_TOKEN` makes the image work regardless, but the App *should* have the grant:

1. GitHub App settings (client id `Iv23liPLLtzfuSuH0uJ2`) → **Permissions & events → Repository permissions → Contents → Read and write** → Save
2. **Accept the new permission on the installation** for this repo (org owner → Settings → GitHub Apps → the app → Review request) — until this step the grant does not take effect

Also worth knowing: `certification-preflight.yml` has **never run on a schedule** (cron is Mondays; it landed 2026-09-03, and the only run is one manual dispatch), and it cannot report on PRs, so nothing surfaces its failure to a human today.

tbraun96:
Campaign-free: every changed path is under `.github/` — zero PERF_PATHS entries, so no record's measured inputs are touched. Stamping to release the held lane.

PR #906, opened 2026-09-05
fix(bfcl): pin soundfile — the scorer could not import from a clean provision

tbraun96:
## What broke

`bfcl-eval` pulls `qwen_agent`, whose `utils/utils.py` does a bare module-scope `import soundfile as sf`. Neither package declares `soundfile`, so the venv provisions **"successfully"** and the scorer then fails:

```
ModuleNotFoundError: No module named 'soundfile'
```

**From a clean provision the BFCL gate was unrunnable.** Hit live on gb10 while certifying `stack/certified-2026-09-04` (PR #877).

## What worked

#777's *"prove the scorer imports before generating 995 responses"* caught this **at provisioning** rather than after ~1.6 GPU-hours of generation, and its error names the remedy. That fix did its job in production today; this PR removes the cause it detected.

## Verification

```
$ venv/bin/python score.py --selftest          # before
ModuleNotFoundError: No module named 'soundfile'

$ venv/bin/pip install soundfile
$ venv/bin/python score.py --selftest          # after
rc=0
```

Pinned (`soundfile==0.14.0`) rather than floated, consistent with the rest of the file: a scorer dependency that moves can move the score with it and silently invalidate every stored baseline.

### DIFF
--- a/crates/atlas-plugin/assets/bfcl/requirements.txt
+++ b/crates/atlas-plugin/assets/bfcl/requirements.txt
@@ -7,3 +7,13 @@
 #
 # bfcl-eval carries both the AST checker and the BFCL v4 data files.
 bfcl-eval==2026.3.23
+#
+# soundfile is a TRANSITIVE import that bfcl-eval does not declare:
+# bfcl-eval -> qwen_agent, and qwen_agent/utils/utils.py does a bare
+# `import soundfile as sf` at module scope. Without it the venv provisions
+# "successfully" and the SCORER then fails to import -- which, from a clean
+# provision, made the BFCL gate unrunnable. Observed 2026-09-05 on gb10:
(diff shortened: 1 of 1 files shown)


tbraun96:
Note on campaign cost: this touches `crates/atlas-plugin/assets/bfcl/requirements.txt`, which is under `crates/` and therefore inside PERF_PATHS by the mechanical rule — even though the content is a Python dependency pin that cannot change a kernel, a scheduler decision, or any measured input. Stamping to release the lane and let the gate state its own verdict rather than pre-judging it.

Independent evidence that the pin is correct: tonight's certification run of `bfcl-subset` on stack #877 **passed at overall 84.22 / normalized 84.12 (n=995)** on a tree with `soundfile` present, after failing to run at all without it on two separately-provisioned boxes.

tbraun96:

PR #907, opened 2026-09-05
ci: build the metal kernel set in the metal test step

TheTom:
`cargo test --features metal (macOS aarch64)` is red on `main` (run 33974555722) and on #895 with `Metal: unknown module 'rms_norm'` (and every other module) across the `metal_backend` parity tests.

**Cause:** the workflow-wide `ATLAS_SKIP_BUILD: "1"` applies to the metal test step too, so `atlas-kernels`' build script emits its stub and `metallib_modules()` returns an empty list. The lane has looked green because GitHub-hosted macOS runners return null from `MTLCreateSystemDefaultDevice` and `maybe_backend()` skips every test. Today the job ran on a real Mac (self-hosted `Toms-M5-Max-Avarok`), the tests executed for the first time, and the empty module set surfaced.

**Fix:** `ATLAS_SKIP_BUILD: "0"` on that one step, with a comment explaining why. Nothing else changes; the CPU-only lanes keep the stub.

**Red first, same machine:**
```
ATLAS_SKIP_BUILD=1 ATLAS_TARGET_HW=metal ATLAS_TARGET_MODEL=qwen3-5-4b-vlm-mlx-int8 ATLAS_TARGET_QUANT=mlx_int8 \
  cargo test -p spark-runtime --no-default-features --features metal --locked metal_backend
  -> Metal: unknown module 'kv_cache_append_turbo4' ... (the CI failure)
ATLAS_SKIP_BUILD=0 ...same...
  -> test result: ok. 35 passed; 0 failed; 5 ignored
```
Full crate with the fixed env: 303 passed, 0 failed.

Not in scope: whether the parity tests should fail loudly instead of skipping when no Metal device exists (a vacuous-green lane is its own problem), and why a `runs-on: macos-14` job was scheduled onto a self-hosted runner.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -400,6 +400,14 @@ jobs:
       # run on a developer machine; default `cargo test` skips them.
       - name: cargo test -p spark-runtime --features metal
         env:
+          # The workflow-wide ATLAS_SKIP_BUILD=1 stubs atlas-kernels' build
+          # script, so `metallib_modules()` is EMPTY and every parity test
+          # fails with `Metal: unknown module '...'`. That never showed on
+          # GitHub-hosted macOS runners because MTLCreateSystemDefaultDevice
+          # returns null there and the tests skip themselves; on a real Mac
+          # (the self-hosted M5) they run and go red. This step is the one
+          # place the metal kernel set must actually be built.
+          ATLAS_SKIP_BUILD: "0"
           ATLAS_TARGET_HW: metal
           ATLAS_TARGET_MODEL: qwen3-5-4b-vlm-mlx-int8
           ATLAS_TARGET_QUANT: mlx_int8


tbraun96:
Releasing certification on this one — it is the keystone for tonight's merges.

`cargo test --features metal (macOS aarch64)` is currently **red on every open PR** that is not web-only. The self-hosted `apple-48gb-metal` runner came back online after its job-started hook was repaired at the host, and the first thing it did was fail 35 of 35 parity tests in 0.06 s with `Metal: unknown module 'noop_smoke'` — exactly the cause this PR names. Until it lands, nothing else can go green.

I reached the identical diagnosis independently on #909 before finding this PR, so consider that an independent confirmation of your root cause rather than a competing fix: hosted `macos-14` has no Metal device, `maybe_backend()` takes its `Err => None` arm, all 35 tests `return` early and report `35 passed` in 0.07 s. The lane has never once executed a Metal kernel. **#909 is being closed in favour of this PR** — yours is the same one-line `ATLAS_SKIP_BUILD: "0"` and it is yours.

The branch is BEHIND `main`; I will update it after the stamp so the run happens on the merged tree.

tbraun96:
Heads-up on how this is landing, and why it is not landing as this PR.

Your commit `3f50e66` is now **cherry-picked verbatim into #933**, authorship intact — `git log` there shows it authored by TheTom, with an `(cherry picked from commit 3f50e66…)` trailer. Nothing about the change is mine; only the vehicle is.

The reason is a deadlock rather than a preference:

- This branch's checks have been **queued for two hours** and have not started. Its jobs go to GitHub-hosted runners, and the hosted pool is saturated — 90 runs deep.
- #933 moves 19 cheap checks and the three heavy Rust jobs onto a self-hosted pool that went live tonight (8 new runner processes on avarok, 32 cores). Its own checks are therefore *running*, which is why it can make progress while this one cannot.
- But #933 could not go green either, because `cargo test --features metal (macOS aarch64)` is red on every non-web-only PR — which is exactly what your fix repairs.

So the two PRs were blocking each other from opposite directions: yours needed runners, #933 needed your fix. Folding your commit into #933 breaks it in one move, and `main` gets both.

Once #933 merges I will close this PR pointing at the merged commit, and your `3f50e66` will be an ancestor of `main` — verifiable with `git merge-base --is-ancestor 3f50e66 main`, not just asserted. If you would rather this PR be the one that lands, say so and I will drop the cherry-pick from #933 and wait for the queue instead; it is your change and your call.

For the record, what your one line fixes: `ATLAS_SKIP_BUILD=1` is set workflow-wide, `atlas-kernels/build.rs` short-circuits on it and emits `metallib_modules() -> Vec::new()`, and the parity tests then look up kernels in an empty registry. On hosted macOS there is no Metal device, so `maybe_backend()` takes its `Err => None` arm and all 35 tests return early reporting `35 passed` in 0.07s. The lane has never executed a Metal kernel. On the self-hosted M-series box the device exists, the lookup is reached, and it fails — which is the runner working correctly.

PR #908, opened 2026-09-05
stack/ci-honesty: four checks that reported success without checking

tbraun96:
## One shape, four instances

Each of these turned a **failure to look** into a **confident answer**. None was hypothetical — every one is reproduced by a new selftest row that goes red when the fix is reverted.

| # | what it claimed | what was true | file |
|---|---|---|---|
| 1 | a **required** check printed `OK: no block_on/block_in_place under tui/ or recipe/` | `grep` exits **2** on a scan error *even when it finds matches*, and 2 is false to `if`, so it fell through to the OK branch — reporting OK on a tree it never read. Directory names are hard-coded again in the workflow's `paths:`, so a rename silences it twice | `tui-threading.yml:78-81` |
| 2 | **"Stamp recorded.** … This holds for the life of the PR." | the run *listing* one line above kept `2>/dev/null \|\| true`; an unanswered API takes the benign "no CI run found" branch, `rerun_note` stays empty, and the PR is told the stamp succeeded while the lane is still `skipped` | `certification-commands.yml:262-264` |
| 3 | **"A new commit landed. The seal is void and the records no longer cover this tree — a perf path moved."** | an unreadable `pulls` API read as an all-negative board and *invented* that demotion, posting it as the PR's one status comment and writing it back into the marker, destroying the real prior state | `certification-state.sh:40,72-74` |
| 4 | `ok /control -> 000000 (fallback, not an error)` | `curl` already prints `000` on transport failure, so `\|\| echo 000` appends a second — matching neither `5*` nor `000`, so a probe that never completed was called healthy | `install-canary.yml:174-175` |

## Campaign accounting

**No campaign owed.** The diff touches five files, all under `.github/` — `git diff --name-only origin/main..HEAD | grep -cE '^(crates|kernels|Cargo|vendor|jinja-templates|rust-toolchain|3rdparty_patches)'` returns **0**. This stack can land on cheap correctness alone.

## Gates run

`certification-selftest.sh` → **127 passed, 0 failed** (was 110 on `origin/main`) · all three workflows re-parsed with `yaml.safe_load` · both shell scripts `bash -n` clean.

Each fix ships with its control; the controls were observed red before the fix and green after, and the verbatim failing rows are in the individual commit messages.

## Why these four are grouped

They are one defect class — *a failed lookup rendered as a benign fact* — found in a single audit pass, and they share the selftest file. Landing them together means one review of the pattern rather than four of the same argument.


tbraun96:
Campaign-free: every changed path is under `.github/` — zero PERF_PATHS entries, so no record's measured inputs are touched. Stamping to release the held lane.

PR #909, opened 2026-09-05
ci: the Metal lane has never tested Metal — build the kernel set, keep the self-hosted runner

tbraun96:
## Reversing this PR's direction — the hosted-macOS green was measuring nothing

This PR was opened to route the Metal lane back to `macos-14` because the self-hosted runner's job-started hook path was unquoted and every job died ~2 s in, before checkout. **That hook is now fixed at the host** (the guard script lives at a space-free path, `.env` repointed, runner service restarted). Jobs reach the test step again.

And the first thing they did was fail — 35 of 35, in 0.06 s:

```
called `Result::unwrap()` on an `Err` value: Metal: unknown module 'kv_cache_append'
kernel lookup: Metal: unknown module 'noop_smoke'
test result: FAILED. 0 passed; 35 failed; 5 ignored; 0 measured; 270 filtered out; finished in 0.06s
```

**That failure is the runner working correctly.** It is the first environment that could ever have caught this, and merging this PR as originally written would have buried it again.

### What the two lanes actually do

| | hosted `macos-14` | self-hosted `apple-48gb-metal` |
|---|---|---|
| Metal device present | no | **yes** |
| `maybe_backend()` | `Err` → `None` | `Ok` → `Some` |
| Every test body | `let Some(backend) = … else { return }` → **returns immediately** | reaches the kernel lookup |
| Result | `35 passed` in **0.07 s** | `35 failed` in 0.06 s |

The hosted lane has reported green for as long as this job has existed, having executed no Metal work at all. This is the same class [#908](https://github.com/Avarok-Cybersecurity/atlas/pull/908) exists to eliminate — a check that reports success without checking.

### Root cause, end to end

1. `ci.yml` sets `ATLAS_SKIP_BUILD: "1"` **workflow-wide** so clippy/check can run on hosts without nvcc.
2. `crates/atlas-kernels/build.rs:172` short-circuits on that and writes a stub — `pub fn metallib_modules() -> Vec<(&'static str, &'static [u8])> { Vec::new() }`.
3. `ATLAS_TARGET_HW: metal` on the test step does **not** override the skip, so the parity tests look up kernels in an empty registry.
4. On a runner with no device that is invisible; on one with a device it is fatal.

The step's own comment already claimed it would *"Build the metal kernel set"*. The inherited env defeated it — the prose and the behaviour had drifted apart.

### The fix pushed here

- `runs-on:` restored to `[self-hosted, macOS, ARM64]`.
- `ATLAS_SKIP_BUILD: "0"` on the metal test step, so `ATLAS_TARGET_HW` drives the real `xcrun → AIR → metallib` pipeline. That path is implemented (`build_target.rs`, `HARDWARE.toml` `vendor = "apple"`, `arch = "metal3.1"`) and 43 `.metal` sources under `kernels/metal/` feed it.

### Deliberately not fixed here

`maybe_backend()` returning `None` is still a silent pass, so a runner that loses its device would go quietly green again. Closing that hole edits `crates/`, a `PERF_PATH` owing a ~5 GPU-hour campaign — it is queued for the next stack rather than spending a campaign on a one-line guard.

### One correction to my earlier comment on this PR

I previously wrote that the Apple build path was unimplemented, citing a stale `build.rs` comment. That was wrong and I retracted it. I re-checked before writing this one: `metallib_modules()` **is** emitted for real by `build_codegen.rs:103` on the non-skip path. `ATLAS_SKIP_BUILD` is the whole cause.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -374,24 +374,12 @@ jobs:
     # suppresses the implicit success() on `needs` — so this job RUNS. Written
     # as `== 'false'` it would fail CLOSED and silently un-gate Rust changes.
     if: ${{ !cancelled() && needs.changes.outputs.web_only != 'true' }}
-    # TEMPORARY: routed back to hosted macOS. #901 moved this to the
-    # self-hosted `apple-48gb-metal`, and that runner's job-started hook path
-    # is unquoted -- it contains a space:
-    #
-    #   /bin/bash: /Users/nologik/Library/Application: No such file or directory
-    #   ##[error]Process completed with exit code 1
-    #
-    # The hook is `/Users/nologik/Library/Application Support/MacCIBurst/bin/
-    # pre-job-disk-guard.sh`, so bash splits it at "Application Support" and
-    # the job dies ~2s in, BEFORE checkout -- every job, every PR. This is a
-    # REQUIRED context, so it made the merge queue impassable for any diff that
-    # is not web-only (observed on #905 and #906, 2026-09-05).
-    #
-    # Restore `[self-hosted, macOS, ARM64]` as soon as the runner's
-    # ACTIONS_RUNNER_HOOK_JOB_STARTED value is quoted (or the directory moved
-    # somewhere without a space). The routing itself is fine; only the host's
-    # hook configuration is broken, and that cannot be fixed from this repo.
-    runs-on: macos-14
+    # Self-hosted Apple Silicon. #901 moved this here; #909 briefly routed it
+    # back to hosted macOS because the runner's job-started hook path was
+    # unquoted and every job died ~2s in, before checkout. That hook is now
+    # fixed at the host (the script lives at a space-free path and the runner
+    # service was restarted), and jobs reach the test step again.
+    runs-on: [self-hosted, macOS, ARM64]
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
       - uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable
@@ -417,6 +405,19 @@ jobs:
       # run on a developer machine; default `cargo test` skips them.
       - name: cargo test -p spark-runtime --features metal
         env:
+          # The workflow-wide ATLAS_SKIP_BUILD=1 exists so clippy/check can run
+          # on hosts without nvcc. Inherited here it made this job a no-op: the
+          # skip path emits `metallib_modules() -> Vec::new()`, so the parity
+          # tests looked up kernels in an EMPTY registry. On hosted macOS there
+          # is no Metal device, `maybe_backend()` returned None, all 35 tests
+          # returned early and reported ok in 0.07s — green while testing
+          # nothing, for as long as this job has existed. On the self-hosted
+          # box a device DOES exist, so the same tests reached the lookup and
+          # failed with `Metal: unknown module 'noop_smoke'`. That failure is
+          # the runner working correctly: it is the first environment that
+          # could ever have caught this. Overriding to 0 makes ATLAS_TARGET_HW
+          # below drive the real xcrun -> AIR -> metallib pipeline.
+          ATLAS_SKIP_BUILD: "0"
           ATLAS_TARGET_HW: metal
           ATLAS_TARGET_MODEL: qwen3-5-4b-vlm-mlx-int8
           ATLAS_TARGET_QUANT: mlx_int8
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -374,7 +374,24 @@ jobs:
     # suppresses the implicit success() on `needs` — so this job RUNS. Written
     # as `== 'false'` it would fail CLOSED and silently un-gate Rust changes.
     if: ${{ !cancelled() && needs.changes.outputs.web_only != 'true' }}
-    runs-on: [self-hosted, macOS, ARM64]
+    # TEMPORARY: routed back to hosted macOS. #901 moved this to the
+    # self-hosted `apple-48gb-metal`, and that runner's job-started hook path
+    # is unquoted -- it contains a space:
+    #
+    #   /bin/bash: /Users/nologik/Library/Application: No such file or directory
+    #   ##[error]Process completed with exit code 1
+    #
+    # The hook is `/Users/nologik/Library/Application Support/MacCIBurst/bin/
+    # pre-job-disk-guard.sh`, so bash splits it at "Application Support" and
+    # the job dies ~2s in, BEFORE checkout -- every job, every PR. This is a
+    # REQUIRED context, so it made the merge queue impassable for any diff that
+    # is not web-only (observed on #905 and #906, 2026-09-05).
+    #
+    # Restore `[self-hosted, macOS, ARM64]` as soon as the runner's
+    # ACTIONS_RUNNER_HOOK_JOB_STARTED value is quoted (or the directory moved
+    # somewhere without a space). The routing itself is fine; only the host's
+    # hook configuration is broken, and that cannot be fixed from this repo.
+    runs-on: macos-14
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
       - uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable


tbraun96:
## Update — I fixed one defect on the host; the second one cannot be fixed from a shell

I was given access to `apple-48gb-metal.local` and went in.

### Defect 1 — the hook path. FIXED at the host, no longer a blocker.

`.env` line 1 was `ACTIONS_RUNNER_HOOK_JOB_STARTED=/Users/nologik/Library/Application Support/MacCIBurst/bin/pre-job-disk-guard.sh`. Symlinked it to a space-free path and repointed `.env` (backup kept, original file untouched). Proven by A/B **on the host, with the runner's own invocation**:

```
OLD: /bin/bash: /Users/nologik/Library/Application: No such file or directory
NEW: CI disk admission passed: 822 GiB free      exit=0
```

Runner service restarted. Later jobs confirm it: `CI disk admission passed: 417 GiB free`.

### Defect 2 — there is no Metal compiler on that machine. This is why the reroute cannot work yet.

```
xcode-select: /Library/Developer/CommandLineTools
metal compiler: xcrun: error: unable to find utility "metal", not a developer tool or in PATH
```

The `metal` shader compiler ships with **full Xcode.app**, not CommandLineTools — and no `Xcode*.app` is installed. So the shader library never builds and **all 35 metal tests fail in 0.05 s**, every one with the same shape:

```
called `Result::unwrap()` on an `Err` value: Metal: unknown module 'embed_lookup'
                                             Metal: unknown module 'argmax_bf16'
test result: FAILED. 0 passed; 35 failed; 5 ignored
```

This is not a flake, a cache miss, or a test bug — the backend has no compiled kernels to look up.

### So this PR is now the right call, and the bar for reverting it is explicit

Merging routes Metal back to hosted `macos-14`, which **has** full Xcode, and unblocks the queue for every non-web PR.

Restore `[self-hosted, macOS, ARM64]` when **both** are true on that box:
1. Full **Xcode.app installed** and `sudo xcode-select -s /Applications/Xcode.app/Contents/Developer` — verify with `xcrun -sdk macosx -f metal` printing a path
2. `cargo test --features metal` passes there once, by hand

Defect 1 stays fixed either way, so that work is not wasted.

tbraun96:
## Correction — Xcode is NOT the answer, and I was about to install it

I was authorised to install whatever the Mac needed. Before downloading ~15 GB I checked what the build actually does, and the premise was wrong.

**`crates/atlas-kernels/build.rs:167` auto-skips the kernel build on macOS:**

```rust
let auto_skip_macos = target_os == "macos" && !hw_explicit;
if skip_env || auto_skip_macos {
    // Stub both the cuda-side (`ptx_modules`) AND the metal-side (`metallib_modules`)
    pub fn metallib_modules() -> Vec<(&'static str, &'static [u8])> { Vec::new() }
```

So `metallib_modules()` returns an **empty list**, the backend loads **zero** libraries, and every lookup fails with `Metal: unknown module '<name>'`. That is precisely the observed failure — all 35 tests, 0.05 s, same error shape.

And the build script says why, at line 213:

> `// Apple (xcrun→metallib), Intel (icpx→SPIR-V). Only NVIDIA is implemented.`
> `// Phase 2 onwards will populate kernels/metal/ and let ATLAS_TARGET_HW=metal drive a real build.`

**These tests cannot pass on any Mac, with or without Xcode.** There is no Apple compile path, not merely a missing compiler. Installing Xcode would have cost an hour and changed nothing.

### What this means for the runner and for this PR

- The **hook fix stands** and was real — `.env` pointed at a path containing a space, every job died ~2 s in before checkout. Fixed at the host, A/B proven, service restarted. Jobs now reach the toolchain.
- The **Metal job cannot pass anywhere today.** So the question is not "which runner", it is "why is a required check running a suite whose kernels are never built".

Merging this PR restores the previous routing and unblocks the queue. But the honest follow-up, which I have **not** done because it is a policy call:

1. Either **gate the metal suite** behind `ATLAS_TARGET_HW=metal` so it is skipped-not-failed until `kernels/metal/` exists, or
2. **Drop it from the required contexts** until Phase 2 lands.

Right now a required check enforces a capability the build script openly declares unimplemented. Worth deciding deliberately rather than leaving it red.

tbraun96:
## Retraction — my diagnosis above was wrong. The Apple path IS implemented.

An independent audit lane checked what I asserted, and I have verified its correction myself. **Do not act on my previous comment.**

I claimed the Apple compile path was unimplemented, citing `build.rs:213` (*"Only NVIDIA is implemented"*). That comment is **stale**, and I treated it as ground truth without checking the tree:

```
$ find kernels -name '*.metal' | wc -l
43
$ grep vendor kernels/metal/HARDWARE.toml
vendor = "apple"
```

There is a real `AppleTarget` (xcrun → AIR → metallib) and 43 shader sources. My earlier `ls kernels/metal/*.metal` returned 0 only because they live in subdirectories (`common/`, `nllb-200-3.3b/`, …). That was a bad check, and it produced a confident wrong answer.

### The actual cause is one inherited environment variable

`ci.yml:40` sets `ATLAS_SKIP_BUILD: "1"` **workflow-wide**. The metal job inherits it, so `atlas-kernels` emits its stub — `metallib_modules() -> Vec::new()` — and the suite can only ever report `Metal: unknown module`. Its verdict is about the stub, not the kernels.

The macOS auto-skip at `build.rs:167` that I blamed is **not** in play either: the step already sets `ATLAS_TARGET_HW=metal`, which disables that branch.

The counter-example is in the same repo: `kernel-compile.yml` deliberately does **not** set the variable, with the comment *"★ NOT set to 1. That is the whole point of this job."*

### What this means for this PR

The one-line fix (`ATLAS_SKIP_BUILD: "0"` on that step) is prepared on `nightly/w6-req`, with a static guard against re-inheriting the variable **and** a runtime assert (`metal_backend::tests::embedded_kernels`) with no skip path, so a future stub cannot pass silently.

But that fix needs a **metal compiler** to run against, and `apple-48gb-metal` has only CommandLineTools — `xcrun -sdk macosx -f metal` fails there. So the two changes are complementary, not alternatives:

- **This PR** (hosted `macos-14`, which ships full Xcode) + **the `ATLAS_SKIP_BUILD` fix** = a metal check that actually compiles kernels and can fail for a real defect.
- Restoring the self-hosted route additionally requires **Xcode.app installed on that Mac**.

The runner hook fix stands and was real, independent of all this.

tbraun96:
## Reversing this PR's direction — the hosted-macOS green was measuring nothing

This PR was opened to route the Metal lane back to `macos-14` because the self-hosted runner's job-started hook path was unquoted and every job died ~2 s in, before checkout. **That hook is now fixed at the host** (the guard script lives at a space-free path, `.env` repointed, runner service restarted). Jobs reach the test step again.

And the first thing they did was fail — 35 of 35, in 0.06 s:

```
called `Result::unwrap()` on an `Err` value: Metal: unknown module 'kv_cache_append'
kernel lookup: Metal: unknown module 'noop_smoke'
test result: FAILED. 0 passed; 35 failed; 5 ignored; 0 measured; 270 filtered out; finished in 0.06s
```

**That failure is the runner working correctly.** It is the first environment that could ever have caught this, and merging this PR as originally written would have buried it again.

### What the two lanes actually do

| | hosted `macos-14` | self-hosted `apple-48gb-metal` |
|---|---|---|
| Metal device present | no | **yes** |
| `maybe_backend()` | `Err` → `None` | `Ok` → `Some` |
| Every test body | `let Some(backend) = … else { return }` → **returns immediately** | reaches the kernel lookup |
| Result | `35 passed` in **0.07 s** | `35 failed` in 0.06 s |

The hosted lane has reported green for as long as this job has existed, having executed no Metal work at all. This is the same class [#908](https://github.com/Avarok-Cybersecurity/atlas/pull/908) exists to eliminate — a check that reports success without checking.

### Root cause, end to end

1. `ci.yml` sets `ATLAS_SKIP_BUILD: "1"` **workflow-wide** so clippy/check can run on hosts without nvcc.
2. `crates/atlas-kernels/build.rs:172` short-circuits on that and writes a stub — `pub fn metallib_modules() -> Vec<(&'static str, &'static [u8])> { Vec::new() }`.
3. `ATLAS_TARGET_HW: metal` on the test step does **not** override the skip, so the parity tests look up kernels in an empty registry.
4. On a runner with no device that is invisible; on one with a device it is fatal.

The step's own comment already claimed it would *"Build the metal kernel set"*. The inherited env defeated it — the prose and the behaviour had drifted apart.

### The fix pushed here

- `runs-on:` restored to `[self-hosted, macOS, ARM64]`.
- `ATLAS_SKIP_BUILD: "0"` on the metal test step, so `ATLAS_TARGET_HW` drives the real `xcrun → AIR → metallib` pipeline. That path is implemented (`build_target.rs`, `HARDWARE.toml` `vendor = "apple"`, `arch = "metal3.1"`) and 43 `.metal` sources under `kernels/metal/` feed it.

### Deliberately not fixed here

`maybe_backend()` returning `None` is still a silent pass, so a runner that loses its device would go quietly green again. Closing that hole edits `crates/`, a `PERF_PATH` owing a ~5 GPU-hour campaign — it is queued for the next stack rather than spending a campaign on a one-line guard.

### One correction to my earlier comment on this PR

I previously wrote that the Apple build path was unimplemented, citing a stale `build.rs` comment. That was wrong and I retracted it. I re-checked before writing this one: `metallib_modules()` **is** emitted for real by `build_codegen.rs:103` on the non-skip path. `ATLAS_SKIP_BUILD` is the whole cause.

tbraun96:
Closing this in favour of **#907**, which is the same fix and is TheTom's.

We reached it independently and landed on the identical one-line change — `ATLAS_SKIP_BUILD: "0"` on the metal test step, so `ATLAS_TARGET_HW` drives the real `xcrun → AIR → metallib` pipeline instead of `atlas-kernels/build.rs` stubbing `metallib_modules()` to `Vec::new()`. Two independent diagnoses of the same root cause is worth more than either PR, but only one should land, and the authorship should be theirs.

What this PR contributed that is worth keeping in the record:

- The lane has reported `35 passed` in **0.07 s** for its entire existence without executing a single Metal kernel. Hosted `macos-14` has no Metal device, so `maybe_backend()` takes its `Err => None` arm and every test body hits `let Some(backend) = … else { return }`. The self-hosted box has a device, so the same tests reach the kernel lookup and fail — that failure is the runner working correctly, and it is the first environment that could ever have caught this.
- This PR originally proposed the opposite: routing back to hosted `macos-14`, where the lane is green. That would have restored a check that measures nothing — the same class [#908](https://github.com/Avarok-Cybersecurity/atlas/pull/908) exists to eliminate. I reversed it once the runner came back; #907 had it right from the start.
- Still open, deliberately not fixed in either PR: `maybe_backend()` returning `None` is a silent pass, so a runner that loses its device goes quietly green again. That hardening touches `crates/`, a `PERF_PATH` owing a ~5 GPU-hour campaign, so it is queued for the next certified stack rather than spending a campaign on a one-line guard.

Also retracted here for the record: an earlier comment on this PR claimed the Apple build path was unimplemented, citing a stale `build.rs` comment. That was wrong — `metallib_modules()` is emitted for real by `build_codegen.rs:103` on the non-skip path, and 43 `.metal` sources under `kernels/metal/` feed it. `ATLAS_SKIP_BUILD` was the whole cause.

If any hunk here did not make it across to #907, point at it and I will lift it onto a fresh branch.

PR #911, opened 2026-09-05
stack/nightly-honesty: 10 fixes where the system was more confident than its evidence

tbraun96:
## One campaign, three lanes, ten fixes

Three autonomous audit lanes ran overnight against `origin/main`. Every commit here touches `crates/`, so each would owe its own ~5 GPU-hour campaign standing alone — **grouped, they owe one.** Composed with zero cherry-pick conflicts.

### Lane A — error paths (3 commits)

| finding | why it matters |
|---|---|
| **Shutdown dropped in-flight requests silently.** Of the three parked states, only `preempted` got an error frame; `prefilling` was freed and `swapped` had its disk image deleted, sinks dropped | A dropped sink is not a quiet failure but a **false** one — the client renders `500 "Inference cancelled"`, the same words used when the *client* aborts, so log and client agree on a lie. Streaming clients get a truncated body under an HTTP 200 already committed. `docker stop` is the common path in |
| Swap-IN failure was log-only; the sequence's sink and its claimed slot were both lost | |
| `/v1/completions` mid-stream error frame built by string interpolation | any message with `"`, `\` or a newline emitted **invalid JSON under HTTP 200** — the client's parser raises instead of showing the reason |

### Lane B — record honesty (6 commits)

| finding | why it matters |
|---|---|
| An unreadable `pulls/N/files` recorded `changed_paths: []` | read downstream as *touches nothing* → no targets, no owners, no promotion debt — and **first place in the merge order**, because 0 partners / 0 targets / 0 files *are* the sort keys. An API error published **"Merge next: #N"** |
| `git diff … > changed.txt \|\| true` — the redirect truncates before git runs | a failed diff → `Changed paths (0 total)` in the classifier prompt, i.e. the model classifying from **title and body alone**, with its confident answer appended to the journey ledger |
| `mapfile < <(git diff … )` in **"One PR, one commit, one signer"** | a process substitution's status is invisible to `set -euo pipefail`, so a failed enumeration → *"this PR adds no records — nothing to agree on"* + exit 0, **clearing every unsigned-record check at once** |
| Throttle counters summed with `unwrap_or(0)` into `.benchmarks/**` | "nothing readable" became `Some(0.0)` — identical to a cool run |
| `min_cached_prompt_tokens: …unwrap_or(0)` | "no round reported a figure" stamped with the **vacuity value the metric exists to detect** |
| Blind PR grouped under `"host / non-kernel"` | a claim about a diff nobody read, and the least alarming category available |

Schema choice, stated: **mark, don't drop.** A dropped record is absent, and absent reads as *no such PR* — vanishing from the collision table and the merge order it should be blocking. `paths_unknown` is `#[serde(default)]`, so old records still parse and still mean *known-and-empty*.

### Lane C — required-check honesty (1 commit)

An audit of **all 20 required contexts** on `main`. Five could not do their job:

| context | defect |
|---|---|
| `cargo test --features metal` | inherits `ATLAS_SKIP_BUILD: "1"` workflow-wide → `atlas-kernels` emits a stub → the suite's verdict is about **the stub, not the kernels** |
| `Enforce ≤500 LoC` | `find crates` on a missing tree feeds the loop nothing → **passes** |
| `kernel shadow structure` | skipped absent trees → empty `kernels/` → **"OK"** |
| `No block_on under tui/` | same shape — a renamed directory silences it |
| `Build mdBook + rustdoc` | `needs:` with **no `if:`** — a failed dependency *skips* it, and a skipped required check counts as satisfied |

The metal fix is `ATLAS_SKIP_BUILD: "0"` on that step, with two anti-permanence guards: a static check (env merges workflow→job→step, so re-inheriting is caught) and `metal_backend::tests::embedded_kernels`, a runtime assert **with no skip path**.

## Gates on the composed tree

`cargo check --workspace --all-targets` ✅ · spark-server **2338/2338** · atlas-plugin **910/910** · certification-selftest **138 passed, 0 failed** (110 on main — 28 new rows, each with a negative control observed red then green) · `cargo fmt --check` clean.

## Campaign accounting

10 commits touching `crates/` → **one** campaign for the group, not ten. Every constituent lane branch is kept intact, so a failing gate can drop one lane and re-certify the rest.

### DIFF
--- a/.github/scripts/assert-gates-are-wired.py
+++ b/.github/scripts/assert-gates-are-wired.py
@@ -60,7 +60,172 @@ GATES = [
     },
 ]
 
+# ---------------------------------------------------------------------------
+# Every required context on `main`, and the job that produces it.
+# ---------------------------------------------------------------------------
+# Pulled from
+#   gh api repos/.../branches/main/protection --jq '.required_status_checks.contexts[]'
+# Branch protection names a STRING. Nothing in the tree makes that string
+# resolve to a job, so a rename, a delete or a move retires a required check
+# silently -- the job keeps passing under its new name and protection keeps
+# waiting for the old one, or (worse) an admin removes the stale entry and the
+# gate is simply gone. Listing them here turns each into a tree-checkable fact.
+#
+# `nested` names a reusable-workflow job: the context is then
+# "<caller job name> / <called job name>", and BOTH halves have to hold.
+REQUIRED_CONTEXTS = [
+    ("cargo fmt --check", "ci.yml", "fmt", None),
+    ("cargo clippy --tests", "ci.yml", "clippy", None),
+    ("cargo test --workspace", "ci.yml", "test", None),
+    ("typos", "ci.yml", "typos", None),
+    ("SPDX license headers", "ci.yml", "license-headers", None),
+    ("kernel shadow structure", "ci.yml", "kernel-structure", None),
+    ("cargo test --features metal (macOS aarch64)", "ci.yml", "test-macos-metal", None),
+    ("PR benchmark gate", "ci.yml", "pr-benchmark-gate-alias", None),
+    ("release matrix / dry-run summary", "ci.yml", "release-matrix",
+     ("release-build.yml", "dry-run-summary")),
+    ("Enforce \u2264500 LoC per source file", "file-size-cap.yml", "check-file-sizes", None),
+    ("cargo deny", "security.yml", "cargo-deny", None),
+    ("cargo llvm-cov --workspace", "coverage.yml", "coverage", None),
+    ("Build mdBook + rustdoc", "docs.yml", "build", None),
+    ("nvcc -> PTX (all gb10 targets)", "kernel-compile.yml", "compile", None),
+    ("Verify committed GDN binaries match PINS.sha256", "gdn-so-pin.yml", "verify-gdn-so-pins", None),
+    ("No block_on under tui/ or recipe/", "tui-threading.yml", "no-blocking-on-the-render-thread", None),
+    ("Build SvelteKit site", "site.yml", "build", None),
+    ("Site unit tests", "site.yml", "unit", None),
+    ("Merge-ancestry guard self-test", "merge-ancestry.yml", "self-test", None),
+    ("CLAAssistant", "cla.yml", "CLAAssistant", None),
+]
+
+# An `if:` that names one of these suppresses the implicit `success()` that
+# `needs:` otherwise carries, so the job still RUNS when a dependency failed.
+STATUS_FUNCTIONS = ("always()", "cancelled()", "failure()", "success()")
+
 problems: list[str] = []
+
+
+def _doc(workflow: str) -> dict:
+    return yaml.safe_load((WORKFLOWS / workflow).read_text())
+
+
+def check_required_context(entry: tuple) -> None:
+    """A required context must be PRODUCIBLE and must not be skippable by a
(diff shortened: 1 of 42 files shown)


tbraun96:
## Folded in a 4th lane — and it found a genuine unbounded-starvation bug on `main`

Now **11 commits**, still **one campaign**. The new lane was pointed at the *shape* of a regression the GPU campaign on #877 had just isolated (an optimisation tuned at C=16 that starves C4/C8), to see whether the same shape already exists on `main`. It does.

### `SlaiPolicy::select_prefills` is pure SJF with no aging

On `main` today:

```rust
// Sort ALL pending by prompt_len, pick shortest N.
indices.sort_by_key(|&i| requests[i].prompt_len);
indices.truncate(capacity);
```

No wait-time term, no arrival tiebreak. `requests` is rebuilt from the pending queue **every tick**, so a long prompt is re-sorted to the tail on every tick it loses. Under a steady arrival of shorter requests — the ordinary shape of a busy serve — it is **never selected at all**. That is an *unbounded* wait, not a slow one: nothing in the ranking key ever changes.

Fixed by reserving **one seat for the queue head**, which bounds the wait at queue position; the remaining seats stay shortest-first. Control: a 500-tick simulation returned `None` for the long prompt at **every capacity 1–8** pre-fix, while `fifo_never_starves_the_head` stayed green throughout, so the harness discriminates rather than failing everything.

### Also fixed

`chunk_ranges` derived its per-chunk **sequence** cap from the row budget alone (implying up to 80 seqs) and ignored the independent width bound of 32 that `can_batch_verify` enforces. An oversized chunk is refused **wholesale**, dropping every sequence in it to the per-sequence verify loop — one weight sweep each. `seq_cap` is now the min of both bounds, each read from its SSOT; inert at n ≤ 32.

### Deliberately NOT changed, and why

Three copies of `single_active_with_spec` justify themselves with *"spec is off by construction at `active.len() >= 2`"*. That is false — the MTP dispatch cap is **32**, so across C=2..32 the predicate says "no spec step" while the dispatcher would have run one, demoting every concurrent sequence to a plain token plus re-bootstrap for the duration of a chunked prefill.

The lane consolidated them to one home with an honest doc and a CI guard, but **did not flip the value**: widening it trades this demotion for the gap the C=1 rule was introduced to close ("one 8K prefill chunk froze every active decoder"). Which side wins at C=2..8 is a throughput question with no measurement behind it — a GPU leg, not a judgement call. The guard pins the divergence so it is tracked instead of latent in a stale comment.

### Tuned-at-one-point constants, reported not touched

- **`dcut_width_cap()` = 8** — the +2.6% is a **C=8-only** measurement; at C=2–7 one sequence per step loses its deepest draft with certainty, and the sibling optimisation measured **−2.4% at C=2 / −3.7% at C=4**.
- **`mtp_multi_seq_mode()`** is derived from the **cap**, not the live width, so it is true on every default boot — force-disabling cross-turn drafter carry even at C=1 (recorded **−6.8% C=1 / −9.2% C=2**) while the log prints `carry=ON`.
- **`phase_promote_prefills.rs:28`** — `swap_remove` turns the FIFO prefill head into LIFO on the single-stream path; unbounded under sustained arrivals. Not fixed: it needs the promote/continue pair simulated together, not a pure function.

## Gates after the fold

`cargo check --workspace --all-targets` ✅ · spark-server **2346/2346** · atlas-plugin **910/910** · certification-selftest **138/0** · `cargo fmt --check` clean · clippy `-D warnings` clean.

Issue #918, opened 2026-09-06
Cold grammar compile for a new tool schema is still ~3.4 s on H100 after the #910 fixes (per-state vocab clone removed, parallel prewarm reverted)


TheTom:
## Summary

Follow-up to #910, which covered the warm tool-call decode path (fixed: ~56 -> 162-177 tok/s, repeated tool calls 0.31 s). The remaining cost is the first request for a new tool schema. On an H100 with Qwen3.6-35B-A3B-FP8 the cold first-tool-token time was 4.8-6.6 s across runs; plain chat first token is 77 ms on the same server.

## What was found

- ~4.5 s of the cold path is grammar preparation before prefill (mask generation for 123 native states), not EBNF compile (10-16 ms) or matcher construction (~50 us).
- `MaskGenerator::token_mask_with_first_char_check` deep-copied the sorted 248k-token vocabulary and trie subtree index per prepared state (`.to_vec()` x2), about 30 M small allocations per schema. Allocation oracle: 12,356 -> 24,644 allocations when irrelevant vocabulary doubles (should be flat).
- A four-worker parallel prewarm made cold latency worse on the H100 host (5.28-6.57 s vs 4.82 s serial), consistent with contention on the shared `Arc<GrammarData>` that `earley/scan.rs`, `complete.rs` and `predict_fsm.rs` clone in hot state loops. That is a source-backed hypothesis, not measured.

## Fixes on the fork (TheTom/atlas `hopper/sm90-target-tdd-2026-09`)

- `72834f21` "Borrow immutable vocabulary during grammar mask preparation": borrow the tokenizer's slices instead of cloning per state. Masks byte-identical to the golden files in both modes; allocations 58 -> 58 across the oracle.
- `cd8c1193` "Check cold grammar allocation growth with irrelevant tokens": permanent `stats_alloc` regression test.
- `37228e85` "Restore serial grammar preparation after cold latency measurements": reverts the four-worker default (`7cf68075`) to one compiler worker.

H100 result with all three: cold first tool token **3.38 s** (from 4.82 s), warm unchanged at 0.11-0.35 s, all typed cases pass. One fresh-process measurement, not a distribution.

## Still open

[...]

PR #923, opened 2026-09-06
xgrammar: borrow the vocabulary during grammar mask preparation

TheTom:
## Summary

`MaskGenerator::token_mask_with_first_char_check` opened by cloning the tokenizer's sorted decoded vocabulary and its trie subtree ranges — `.to_vec()` on both — and then cloned each token again into `prev_token` while scanning. The tokenizer is immutable external data; the clones existed only to sidestep the borrow against `&mut self`. At a 248k-token vocabulary that is roughly 30M small allocations per compiled schema, paid once per prepared state.

Refs #918

## Why

The cost is invisible in a profile that looks at grammar logic, because none of it is grammar logic — it is allocator traffic proportional to vocabulary size times prepared states, on a path whose actual work is proportional to the *matching* tokens. Borrowing the tokenizer instead of copying it makes the cost track the work.

## What changed

- `crates/xgrammar/src/compiler/mask_gen/intervals.rs`: copy the `&TokenizerInfo` reference out of `self` before the mutable borrow begins, then use `sorted_decoded_vocab()` and `trie_subtree_nodes_range()` as slices. `prev_token` becomes `Option<&[u8]>` instead of `Option<Vec<u8>>`, removing the per-token clone in the scan loop.
- `crates/xgrammar/tests/vocabulary_allocations.rs`: an allocation oracle over `stats_alloc`, which measures how cold compile allocations grow when the vocabulary gains tokens the schema can never match. This is the property that regressed silently before — a correctness test cannot see it, because the masks were always right.
- `Cargo.lock`, `crates/xgrammar/Cargo.toml`: `stats_alloc = "=0.1.10"` as a dev-dependency. Pinned exactly because it is a measurement instrument; a patch bump that changes counting would move the oracle's numbers for reasons unrelated to Atlas.

## Numbers

Doubling the vocabulary with tokens irrelevant to the schema:

| | allocations before | allocations after doubling |
|---|---|---|
| old code | 12,356 | 24,644 |
| new code | 58 | 58 |

Masks are byte-identical against the golden files in both the serial and the parallel mode — this changes how the vocabulary is reached, not what is computed from it.

## Test plan

- [x] `cargo test --locked -p xgrammar` — 825 passed, 0 failed: 822 unit, 2 `qwen_toolcall_mask`, 1 new `vocabulary_allocations`
- [x] `cargo clippy --locked -p xgrammar --tests` — clean
- [x] Tested against real hardware: see below

## Hardware evidence

On an H100, cold time to the first tool token with the serial default: 4.82 s -> 3.38 s.

## Notes for reviewers

Partial fix for #918 — this removes the allocation waste, but the ~3.4 s cold compile remains and the issue stays open. The parallel prewarm that was tried against the remainder is not in this PR; it was added and reverted on the campaign branch and is a net no-op.

Split out of the campaign branch `hopper/sm90-target-tdd-2026-09` (#895); origin commits 72834f21 and cd8c1193.

## Authorship

AI-authored (Claude).

## CLA

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2735,6 +2735,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
+name = "stats_alloc"
+version = "0.1.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c0e04424e733e69714ca1bbb9204c1a57f09f5493439520f9f68c132ad25eec"
+
+[[package]]
 name = "strsim"
 version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3595,6 +3601,7 @@ dependencies = [
  "rayon",
  "serde",
  "serde_json",
+ "stats_alloc",
  "thiserror",
 ]
 
--- a/crates/xgrammar/Cargo.toml
+++ b/crates/xgrammar/Cargo.toml
@@ -31,6 +31,7 @@ thiserror = "2"
 # Micro-benchmark harness for the Tier-1 perf work — measures
 # `GrammarCompiler::compile_*` and `GrammarMatcher::fill_next_token_bitmask`.
 criterion = "0.8"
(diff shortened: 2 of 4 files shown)


tbraun96:

PR #924, opened 2026-09-06
spark-server: keep parameter value entry inside the grammar scanner

### LINKED ISSUE #910: Tool-call decoding is ~4x slower than plain decode: grammar mask computation on the host path
## Summary

On a single H100 (CC 9.0, driver 610.57, `sm_90a` build at `f66a262`), Qwen3.6-35B-A3B-FP8, temperature 0, thinking off: plain chat decodes at ~245 tok/s, but any request that carries a tool schema decodes at **~56 tok/s** for the duration of the tool call, and the first tool call after a cold start waits **~4.9 s** before the first token. The same requests against a second engine on the same GPU and checkpoint produce the same text, with the tool call finished in 0.29 s. Correctness is fine (12/12 valid calls across `tool_choice` auto/required/named × streaming/blocking); this is purely a latency defect in the constrained-decoding path.

## Measurements (all on the H100, evidence retained on the campaign docs branch under `docs/campaigns/hopper-atlas-vs-vllm-2026-09/evidence/rental-h100-20260905/tool-call-audit/`)

| request | first tool token | finish | tokens/s during tool response |
|---|---|---|---|
| tool call, cold (first schema seen) | 4.94 s | 5.66 s | 54 |
| tool call, warm (same schema again) | 0.115 s | 0.80 s (~41 tokens) | 56 |
| plain chat, same server | 0.077 s | 0.176 s (~20 tokens) | ~245 |

`ATLAS_DECODE_TIMING=1` over the last 100 host-path tokens of warm tool calls:

```
copy+fwd-wait = 4.11 ms/tok   sample(248k host) = 13.78 ms/tok
```

Any grammar-bearing sequence forces the host sampling path (`crates/spark-server/src/scheduler/decode_logits_step.rs:132`); plain greedy rows stay on GPU argmax. The 13.78 ms is not the arithmetic: a native micro-benchmark of the host loops (BF16 expand, mask apply, top-two scan, FP32 copy + greedy over the 248 K vocab) totals ~0.24 ms/token. It is the **grammar mask computation**. A CPU replay of the exact 40-token response shows the fill cost concentrated on four positions:

| position | token | fill |
|---|---|---|
| 13, 14 | `\n`, `Re` (start of the `"Reykjavik"` value) | ~68 ms each |
| 28, 29 | `\n`, `3` (start of the `days` value) | ~68 ms each |
| all 36 others | | ~0.1 ms each |

272 of 273 ms come from the two free-form JSON value slots, where the grammar admits nearly the whole vocabulary and the next-token mask is computed on demand instead of being served from a compiled mask.

## Cold start

`GrammarState::build` calls `compile_top_k_masks(512)` synchronously before prefill and computes the selected states serially (`compiled_grammar.rs` ~363, `get_or_compute_mask`). Parallelising that compile only brought 2.18 s → 1.72 s in a release build, so it is not the main fix. The first request for a schema in a server's lifetime pays ~4.9 s; the compiled masks are not persisted across restarts.

## Already fixed on #895

- `d1322aa1` removed a duplicate rendering of the tool schema (Atlas prepended its own tool-instruction block and also passed `tools` to the checkpoint's native template): 1,135 → 298 prompt tokens for the same request.
- `55784398` makes `time_to_first_token_ms` include grammar preparation; before, the reported TTFT for a 4.96 s request was 460 ms.

Neither changes the per-token or cold costs above.

## Suggested directions

1. Treat near-full masks as a complement (allow-all minus a small deny set) or precompute masks for the "any string / any number" states so value slots do not trigger a full-vocabulary fill per token.
2. Persist or precompile grammars per tool-schema hash so the first agent request of a session does not pay the compile.
3. For `tool_choice=auto`, consider not constraining decoding at all and parsing the emitted call instead (the model produced a correct call in every run here without needing the mask); reserve the grammar for `required`/named where the format must be forced.
4. Longer term, apply the bitmask and argmax on device so grammar-bearing rows do not fall off the GPU path (the 4.11 ms/token transfer + wait would also disappear).

## Reproduction

Serve Qwen3.6-35B-A3B-FP8 (any Hopper or GB10 build), send the campaign coherency gate's weather tool request (`bench/hopper_ab/coherency_gate.py`, the `toolcall_ok` probe) twice with streaming on, and compare first-tool-chunk and finish times against a plain 20-token chat request; run with `ATLAS_DECODE_TIMING=1` to see the host-path split.

TheTom:
## Summary

The generated tool-call grammar factored every parameter value as `first_content rest` — a first-character alternative followed by the remainder. Splitting entry out of the scanner that way means the value slot leaves the scanner and re-enters it, and each re-entry fills a full-vocabulary mask. At roughly 68 ms per fill and four fills per value slot, tool-call decoding paid about 270 ms of host work per token that plain decode does not pay.

Fixes #910

## Why

The factoring was there to express "a value is non-empty", but non-emptiness does not require leaving the scanner. Expressing it as a single `nonempty_value` production keeps entry inside the scanner, so the value slot is reached without a mask fill — the mask work then tracks the tokens actually being matched instead of the number of times the grammar re-enters a rule.

## What changed

- `crates/spark-server/src/grammar/compile_tools.rs`: emit `nonempty_value` for parameter values instead of the `first_content rest` factoring.
- `crates/spark-server/src/grammar/tests/value_entry.rs` (+ registration in `tests/mod.rs`, and an assertion update in `param_key_constraint.rs`): 12 typed value-shape cases, each run across all four combinations of the empty-value and force-close options, differentially compared against the old unfactored rule for masks, acceptance, rollback and termination. They pin that entry stays inside the scanner and that the language is unchanged.

## Numbers

Measured on an H100, dense 27B:

| | before | after |
|---|---|---|
| per-token host sampling | 13.78 ms | 1.55–1.68 ms |
| tool-call decode | ~56 tok/s | 162–177 tok/s |
| warm repeated tool call | 0.798 s | 0.31 s |

## Test plan

- [x] `cargo test --locked -p spark-server --bin spark grammar` — 95 passed, 0 failed (the grammar module, including the two new `value_entry` tests)
- [x] `cargo test --locked -p spark-server` — 2443 passed, 0 failed, 18 ignored
- [x] `cargo clippy --locked -p spark-server --tests` — clean
- [x] Tested against real hardware: see the table above

## Notes for reviewers

This is the decode-time half of #910. The cold-compile cost that remains after it is tracked separately in #918.

Split out of the campaign branch `hopper/sm90-target-tdd-2026-09` (#895); origin commit 6884325d.

## Authorship

AI-authored (Claude).

## CLA

### DIFF
--- a/crates/spark-server/src/grammar/compile_tools.rs
+++ b/crates/spark-server/src/grammar/compile_tools.rs
@@ -226,16 +226,21 @@ pub(crate) fn xml_param_value_body_ebnf_opts(
     // `</parameter<parameter=` attractor. Opt-in allows the empty value
     // (immediate close). Trade-off: re-opens the Epoch-3 empty-garbage-call
     // shape — hence opt-in + BFCL A/B gate before default-on.
+    // Keep first_content at the start of its own sequence so the grammar
+    // optimizer can inline it into rest's scanner. Otherwise multi-byte value
+    // tokens cross a rule boundary and require contextual trials at value entry.
+    // Factoring this sequence preserves the language, including the empty opt-in.
     let value_rule = if allow_empty_value {
-        "value ::= leading_ws (first_content rest)?"
+        "value ::= leading_ws nonempty_value?"
     } else {
-        "value ::= leading_ws first_content rest"
+        "value ::= leading_ws nonempty_value"
     };
     format!(
         r#"root ::= param ("\n" param)*
 param ::= "<parameter=" paramname ">" value "{value_close}"
 {paramname_rule}
 {value_rule}
+nonempty_value ::= first_content rest
 leading_ws ::= [ \t\r\n]*
 first_content ::= [^ \t\r\n<=>] | {first_lt_arms}
 {rest_rule}
--- a/crates/spark-server/src/grammar/tests/mod.rs
+++ b/crates/spark-server/src/grammar/tests/mod.rs
@@ -15,6 +15,7 @@ mod poolside;
 mod qwen3_coder_required;
 mod sanitize;
 mod tools_basic;
+mod value_entry;
 
 /// Build a minimal vocabulary for testing.
 /// Contains basic ASCII + JSON structural tokens.
--- a/crates/spark-server/src/grammar/tests/param_key_constraint.rs
+++ b/crates/spark-server/src/grammar/tests/param_key_constraint.rs
@@ -93,12 +93,13 @@ fn p1_opts_empty_value_and_force_close_shapes() {
     use super::super::compile_tools::xml_param_value_body_ebnf_opts;
     // Defaults (both off) == the shipped shape.
     let d = xml_param_value_body_ebnf_opts("</parameter>", None, false, false);
-    assert!(d.contains("value ::= leading_ws first_content rest"));
+    assert!(d.contains("value ::= leading_ws nonempty_value"));
+    assert!(d.contains("nonempty_value ::= first_content rest"));
     assert!(d.contains(r#""</parameter" [^>]"#));
     // P1-1: empty value representable.
     let ev = xml_param_value_body_ebnf_opts("</parameter>", None, true, false);
     assert!(
-        ev.contains("value ::= leading_ws (first_content rest)?"),
+        ev.contains("value ::= leading_ws nonempty_value?"),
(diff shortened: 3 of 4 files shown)


tbraun96:

PR #933, opened 2026-09-06
ci: route the cheap PR checks to a self-hosted pool, starting with one job

tbraun96:
## The measurement

Hosted CI starved tonight. Between 01:00Z and 02:10Z on 2026-09-06:

| CI-workflow runs created | 20 |
|---|---|
| **succeeded** | **0** |
| cancelled by concurrency churn | 8 |
| still queued, oldest | **81 minutes** |

Both self-hosted runners were **online and idle** throughout. The cheap checks suffer worst — `typos` is seconds of work and it waited over an hour behind CUDA toolchain installs.

## Capacity first, routing second

A self-hosted runner takes **one job at a time**. Pointing ten checks at the single existing `avarok-cmd` would have serialised them and been *slower* than the queue we are escaping. So eight runner processes are now registered on avarok (32 cores, 125 GB, Docker 29.1.4) under a new label `atlas-pr-cheap`:

| | before | after |
|---|---|---|
| concurrent self-hosted jobs | 2 | **10** |

`avarok-cmd` is untouched and keeps its own label.

## Why a second label instead of reusing `atlas-cmd`

They make opposite promises.

`atlas-cmd` is safe because it **never checks out PR code at all** — it pins `ref: default_branch`, and `assert-cmd-runner-safe.py` enforces that. This pool exists *to* run PR code, so it cannot make that promise. Its promise is narrower: **only branches in this repository**. A fork's pull request still goes to a hosted runner.

That entire promise lives in one `runs-on` expression, so the assertion now pins it:

- a job naming the cheap pool — by label **or** through `vars.PR_CHEAP_RUNNER` — under a `pull_request` trigger must carry `github.event.pull_request.head.repo.full_name == github.repository`, or CI fails here;
- and it must not then hand that runner an explicit fork ref. The guard chooses the **runner**; it does not choose what gets checked out.

## Controls

Four added to `certification-selftest.sh` — three red, one green:

| control | expect |
|---|---|
| cheap pool on `pull_request`, no guard | rc 1 |
| cheap pool via the variable, still unguarded | rc 1 |
| guarded but checking out a fork ref | rc 1 |
| **the shape actually shipped here** | rc 0 |

Suite: **114 passed, 0 failed**. Also proved by hand against the real `ci.yml` — deleting the guard from this commit's own expression turns the assertion red and names the job; restoring it turns it green.

## One job, deliberately

`typos` only. It is cheap, needs no toolchain, and is a required context.

A label nothing answers leaves a required check **queued forever** — `release-build.yml:236` records `macos-13` doing exactly that. So the rest of the fleet moves only once this job is observed being picked up by a real runner.

`runs-on:` changes only. No job is renamed: `main`'s protection requires 20 contexts **by name**.

## Rollback

Unset the `PR_CHEAP_RUNNER` repository variable and every routed job returns to `ubuntu-latest` with no code change — the same escape hatch `CMD_RUNNER_LABEL` already provides for the command runner, and one `cmd-runner-health.yml` can operate automatically.

### DIFF
--- a/.github/scripts/assert-cmd-runner-safe.py
+++ b/.github/scripts/assert-cmd-runner-safe.py
@@ -70,8 +70,14 @@ def guards_against_forks(job):
     return SAME_REPO_GUARD in blob
 
 
+def routing_blob(job):
+    """The cheap pool's `runs-on`, whitespace-collapsed for comparison."""
+    return " ".join(str(job.get("runs-on")).split())
+
+
 def main():
     problems = []
+    routed = []
     for path in sorted(pathlib.Path(".github/workflows").glob("*.yml")):
         try:
             wf = yaml.safe_load(path.read_text(encoding="utf-8"))
@@ -85,6 +91,7 @@ def main():
             if not isinstance(job, dict):
                 continue
             if uses_pr_pool(job):
+                routed.append((routing_blob(job), f"{path.name}:{name}"))
                 if triggers & FORK_CONTROLLED and not guards_against_forks(job):
                     problems.append(
                         f"{path.name}:{name} runs on {PR_LABEL} under a "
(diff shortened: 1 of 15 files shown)


tbraun96:
/stamp

Releasing certification. Deliberately ONE comment with the command on the first line rather than the usual explain-then-command pair: `cla.yml` fires on `issue_comment` with `cancel-in-progress: true`, so two comments seconds apart cancel the `CLAAssistant` run and leave the PR red on a check that never ran. That happened twice on #907 tonight. This PR is the one that fixes it; until it lands, one comment is the safe shape. The command parser reads only the first word of the first line, so this still stamps.

Status: `cargo test --workspace` and `cargo clippy --tests` both passed on the new self-hosted Rust pool (clippy in 113s). Two failures remain and neither is a code fault:

- **`cargo fmt --check`** — the runner had stable 1.98.1 but `rust-toolchain.toml` pins 1.93.1, and that toolchain was present without its `cargo` component: *"the 'cargo' binary … is not applicable to the '1.93.1' toolchain"*. Fixed at the box (1.93.1 installed with rustfmt, clippy and llvm-tools); re-running.
- **`cargo test --features metal`** — TheTom's fix is working exactly as intended and has exposed the next layer: `xcrun: error: unable to find utility "metal", not a developer tool or in PATH`. The Mac has the Command Line Tools but not the Metal compiler, which ships with Xcode. The build now genuinely tries to compile the 43 `.metal` sources and cannot. That needs the Metal toolchain installed on `apple-48gb-metal`; I have no route to that host from the DGX boxes.

PR #873, opened 2026-09-04
governance: stop the harvest re-triggering itself (≈half the runner queue)

tbraun96:
**The harvest bot was consuming roughly half the repository's runner capacity, in a loop it created itself.**

## The loop

`governance-harvest.yml` triggers on `workflow_run: [CI] completed` and force-pushes `bot/governance-harvest`, opening a pull request. That PR **is** a pull request — so its own CI completing re-triggered the harvest, which pushed the branch again.

Measured today:

| | |
|---|---|
| harvest PRs opened **and closed** | **12** (#854…#871), one every 20–30 min |
| queued runs belonging to that one branch | **11 of 27** |
| other PRs waiting behind it | including two at **zero failures**, queued 2 hours |

## Why it stayed hot

The harvest **closes** its PR whenever it looks unmergeable (the "Superseded automatically" path). An unstamped PR always looks unmergeable: certification is **held** until a human comments `/stamp`, and nobody stamps a bot PR.

So each cycle ran: *open → look broken → close → recreate → run CI → trigger the harvest*.

**The gate was working exactly as designed. The bot read "held" as "broken".**

Worth noting what this cost beyond runners: I stamped and sealed #867 by hand earlier today; the bot replaced it with #871 two hours later, discarding both marks.

## The fix

The job condition now excludes `workflow_run` events whose `head_branch` is the branch this workflow pushes. One clause, and the reasoning is written above it so the next person does not simplify it back.

## The guard

`assert-no-workflow-feedback-loop.py` pins the class for **any** `workflow_run`-triggered workflow. It discovers the pushed branch from that workflow's own `git push --force … refs/heads/<branch>` lines, so it cannot drift from a hand-maintained list.

Against the pre-fix file it refuses and names the branch:

```
REFUSE: governance-harvest.yml is triggered by `workflow_run` and force-pushes
        `bot/governance-harvest`, but no job condition mentions that branch.
```

It lives in the **LoC-cap job** (a required context) rather than in the workflow it protects — a workflow that can re-trigger itself is exactly the thing that will not notice.

## Not fixed here

The bot's "is my PR broken?" test still counts a **held** certification gate as broken. That is a second, independent defect: it will keep closing and recreating its PR whenever CI is merely incomplete, just far less often now that the loop is cut. Worth a follow-up that teaches it the difference between *held* and *failed*.

Touches only `.github/` — **no `PERF_PATHS`, no campaign owed.**

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #941, opened 2026-09-06
ci: stop paying runner overhead for checks that do no work

tbraun96:
Four commits, each independently revertible. Measured before and after; the claims below are numbers, not impressions.

## What the pipeline costs today

A PR run is **25–26 jobs**. Across 11 completed runs:

| | |
|---|---|
| Σ wall across all jobs | **5300–6000 s** |
| makespan (critical path) | 1433–2780 s |
| `release matrix / build windows-x86_64-nvidia-cuda` | **1182 s — 83% of the critical path alone** |
| the nine build legs together | ~4700 s of the ~5900 s total |

And at the cheap end, on the self-hosted pool:

| job | queue | wall | real work |
|---|---|---|---|
| `classify diff (web-only?)` | 28s | 5s | **0s** |
| `kernel shadow structure` | 28s | 6s | **0s** |
| `harvest triage decision table` | 15s | 6s | **0s** |
| `stamp status` | 13s | 6s | **0s** |
| `SPDX license headers` | 15s | 9s | 1s |
| `typos` | 26s | 9s | 2s |

**Eight jobs: 151 s queued, 48 s running, ~7 s of actual work — queue is 76% of their total.** Several spend *zero measurable seconds* on the thing they exist for. The pool is 12 runners wide, so the slots matter more than the seconds.

## The four commits

**1. `classify diff` needs the file list, not the whole commit graph.** Six byte-identical jobs each did `actions/checkout` with `fetch-depth: 0` to compute three booleans — 6×(N+1) full-history fetches per PR. The script already had a stdin mode, so the list now comes from `pulls/{n}/files` (which *is* the three-dot merge-base diff it replaced) and checkout drops to depth 1. Two new failure modes are closed: an unanswered API is no longer indistinguishable from an empty diff, and both endpoints' truncation ceilings (3000 / 300 files) now refuse to classify rather than reasoning about a partial diff. Both answer *build everything*.

**2. One `cheap checks` job for eleven checks.** The inline bodies in `file-size-cap.yml`, `gdn-so-pin.yml` and `tui-threading.yml` were **extracted to scripts first**, so the standalone job and the batch call the same code and cannot disagree during the transition. No fail-fast: every check runs, failures are collected and each `::error` names the context it stands in for.

**3. `cargo fmt` folded into the clippy job** — one toolchain install and one cache restore instead of two, formatting first so a 6 s failure isn't reported after an 88 s clippy run.

**4. The release matrix runs when marked, and tests instead of only building.** Half of the gating already existed and was unused — `release-matrix` is gated on `stamped != 'false'` and `stamp` counts `Expedite`. Missing was `/seal`, now a disjunct (an unsealed PR must still be able to build). And every leg gains `test_mode`, defaulting to `no-run`: `cargo test --release --no-run` compiles and links every test binary for the target, which is strictly more than `cargo build` checked, without gambling nine legs on GPUs the runners don't have.

## Controls, all observed red

| control | result |
|---|---|
| stray file with no SPDX header | RED: `SPDX license headers` |
| `.block_on(` under `tui/` | RED: `No block_on` |
| a 600-line `.rs` under `crates/` | RED: `Enforce <=500 LoC` |
| one corrupted hash in `PINS.sha256` | RED: `Verify GDN binaries` |
| **two broken at once** | RED: **both named** (no fail-fast) |
| API failure treated as an empty diff | 170 passed, **1 failed** |
| truncation guard removed | 169 passed, **2 failed** |
| all restored | **171 passed, 0 failed** |

Self-test 165 → 171 checks.

## Three mistakes worth reading, all caught by assertions rather than review

* The `/seal` gate first landed on **`pr-benchmark-gate`**, not `release-matrix`. The anchor matched *exactly once* — in the wrong job. Unique is not correct. The effect would have been a `/seal` releasing the certification lane, which is not what a seal means.
* The `crate-ci/typos` SHA in the new job was **fabricated from memory**; corrected to the one the existing job pins (`5e38bf262…`).
* The first GDN control corrupted **line 1 of `PINS.sha256`**, which is a comment `sha256sum -c` ignores — so the batch stayed green and the control was measuring nothing. It now corrupts the first real hash line, and `git status` confirms byte-exact restoration.

## Not included: deleting the `metal` feature

Planned for this PR and **split out after measuring**. Deleting `metal` is 4 `cfg` sites. But the goal — `cargo build` with no flags doing the right thing on a Mac — needs the **cuda** side target-aware too, since `default = ["cuda"]` would otherwise activate **123 `cfg(feature = "cuda")` sites across 36 files** on macOS. There are currently **zero** `cfg(target_os = ...)` sites in `crates/`. That is a production-Rust refactor, not a Cargo.toml edit, and it belongs in a perf stack that owes a campaign anyway.

Keeping it out also keeps this PR free of `PERF_PATHS`, so it owes no campaign.

## Not included: the workflow merge

The six `classify diff` jobs are no longer identical — #935 added `is_stack_layer` and `builds_binaries` to `ci.yml`'s copy alone. Folding these workflows so one `changes` job serves all is the follow-up that actually removes the five extra jobs; outputs cannot cross a workflow boundary today.

## Follow-up required before the saving lands

Nothing is deleted here and no required context changes — the nine original jobs still run alongside `cheap checks`. Retiring them needs branch protection repointed first: **add** `cheap checks`, settle, then **remove** the old names, then delete the jobs. `.github/scripts/assert-required-checks-not-comment-cancellable.py` mirrors the required-context list and must move in the same commit.


tbraun96:
Held out of the CI stack (#1113) tonight, with the conflict named rather than guessed at.

`ci: one cheap-checks job for the checks that did nothing but pay for a runner` conflicts in `.github/workflows/file-size-cap.yml` and `tui-threading.yml` with two guard steps that #873 and #868 add to that same job:

- `no workflow_run workflow can re-trigger itself via a branch it pushes` (#873)
- `no required check is cancellable by an event that cannot rewrite it` (#868)

Both are **required-context** steps. Folding them into a consolidated job is the kind of change that loses a required context quietly — the job still runs, still passes, and reports under a name branch protection is not waiting for — so it wants a deliberate pass with `assert-gates-are-wired.py` as the check, not a blind merge resolution.

Once #873 lands, rebasing this PR onto it should be mechanical: the two steps need to move into whichever job the consolidation keeps, and `assert-gates-are-wired.py` will refuse the result if either context disappears. Nothing here is refuted — it is just ordered after.

tbraun96:
Re-cut onto the post-rebrand `main` and made the base of stack #1123, with #622, #876 and #869 merged down into it (each now shows MERGED). One campaign covers the group.

| layer | PR | what it brings |
|---|---|---|
| base | **#941** | classify-diff reads the file list; the cheap-checks job; fmt folded into clippy; the release matrix tests instead of only building; the macOS legs' tests marked `skip` with the reason |
| 2 | #622 | the harvest commit subject is bounded and marked ⚖️; the ledger's FIFO window (#623) |
| 3 | #876 | the AutoMerger skill, the O.R.A.C.L.E agent contract, `pr-containment-scan.py`, `render-stack-map.py` — **without** its stack-model half (see below) |
| 4 | #869 | getrandom 0.4, 133 compatible lockfile bumps, minijinja pinned |

## What the re-cut had to repair, and what found it

- the extracted `check-no-block-on.sh` and `check-file-size-cap.sh` had lost the **"scanned nothing"** assertions their inline bodies carry. A grep over a missing directory exits 2, the `if` reads that as "no hits", and the gate goes green over a tree it never read; `find crates` on a missing directory leaves the violation count at 0 and the cap unenforced repo-wide. Both scripts are main's bodies verbatim now, and the self-test's controls prove each still refuses a tree it cannot scan.
- the cap's **allow-list** travelled with the script and had missed 25 entries added to main since, so it failed on `impl_a3.rs` and `impl_b3.rs`.
- the three consolidated jobs routed to the self-hosted cheap pool **without the fleet gate** that landed after this branch was cut. `assert-cmd-runner-safe.py` refused them: a self-hosted job with no fallback queues forever when the box is down, and a queued required check is ABSENT rather than red.
- the self-test's own controls had to move with the bodies: they run the SCRIPTS now, because extracting `bash .github/scripts/...` into a scratch tree resolves to nothing and passes vacuously — the exact failure this file exists to catch.

## Held out of the stack, each with its cause

- **#876's stack-model half** (its edits to ci.yml, release-build.yml, certification-commands.yml, classify-diff.sh): it implements *top-of-stack certifies, lower layers defer*, read off `pull_request.stack.position`. The 2026-09-12 decision inverted that — layers merge DOWN and the BASE certifies — which is what main implements and what the skill's own rulebook describes. The payload is still the better signal than an API query and its `/stamp`-race fix is real; both want a follow-up that reads the payload and concludes base-certifies. Not a decision to make inside a conflict marker.
- **minijinja 2.24** — see below.
- **#900**: `main`'s own comment refutes its routing until `xcrun -f metal` answers on `apple-48gb-metal`.
- **toml 1.1 (#556)** and **ratatui 0.30 (#557)**: the causes #869 proved still hold.

## The one that was not a patch bump

minijinja 2.21 → 2.24 changes how Python-shaped values render inside a chat template: a bool arrives as `true` where the official Qwen template produces `True`, a null as `none` → `null`. That is tool-call argument text the model reads — prompt content wearing a patch bump's clothes. `tokenizer::tests::qwen_dense_parity` failed on exactly the two cases that pin it, so minijinja is `=2.21.0` with the finding written above the pin.

## Verification

`certification-selftest.sh` 248/248 · all five workflow guards pass · `cargo fmt`, `clippy --workspace --tests`, and 93 test suites green · `governance-evict-test.sh` all cases.

`Cargo.lock` and two `Cargo.toml` files are PERF_PATHS, so the group needs one certification campaign — starting now on the three GB10s.

PR #950, opened 2026-09-06
gate: a new file under src/gate could judge its own PR, and nothing noticed

tbraun96:
**Stack 3 of 3 — the top.** Base: `fix/required-checks-not-displaced-while-pending` (stack 2). Merging this lands the whole stack.

## The hole

`GATE_MACHINERY` excludes the whole `crates/atlas-plugin/src/gate` prefix from every gate's invalidation set. So a file in that directory that decides a verdict and is **not** in `BOUNDARY_FILES` invalidates nothing — and the PR that edits it is then judged **by its own new logic**. That is PR #420's shape, which `BOUNDARY_FILES` exists to close.

The existing guard walks a hardcoded list of **seven verdict functions** and proves each one's defining file re-opens every gate. It is exactly as complete as that list: it catches a function *moving* — it was written after a 500-line split carried `check_record`/`compare` into a new `scoring.rs` — and it is **blind to a verdict function that is new**, because nothing adds the new symbol to the list.

Two files were through it:

| file | why it decides a verdict |
|---|---|
| `amnesty.rs` | decides whether a gate is **excused**. A PR widening the amnesty table excuses itself. Promoted to `BOUNDARY_FILES` here |
| `agreement.rs` | decides which record **sets** are acceptable. On #934, not yet merged — it cannot be listed from this branch because the file does not exist here. **The new test will refuse the rebase that brings it in until it is classified**, which is the mechanism working rather than a note to remember |

## The inversion

A file is no longer asked to prove it is dangerous; it must **declare which it is**. `GATE_MACHINERY_FILES` records the opposite claim — "someone looked and this does not decide a verdict" — with a reason per entry, and `gate_sources_are_all_classified` asserts the two lists are together exactly the directory listing, with no file in both. Adding a source fails the build until its author classifies it, and calling something machinery is a claim in the tree that review can argue with.

## Controls, all observed

| tree | result |
|---|---|
| baseline | 1 passed |
| a new unclassified gate source | **FAILED** — `unclassified gate source(s): [.../newverdict.rs]` |
| a file classified as BOTH | **FAILED** |
| a stale classification (rename/delete) | **FAILED** |
| restored | 1 passed |

My first run of those controls printed `ok` three times at **`0 passed; 0 failed; 901 filtered out`** — the filter said `gate::hardening` where the module is `gate::hardening_tests`, so it selected nothing and reported success. Caught by reading the count.

## Certification

This touches `crates/`, but the only additions are a `#[cfg(test)]` const and a test — **it cannot change inference behaviour**, which is what the campaign protects. Requesting `/expedite` on that basis rather than on convenience.

---

*Reopened as a new PR: #939 was closed at 21:07Z when a bad force-push briefly made this branch identical to its base. GitHub freezes a closed PR's head, so #939 could not be reopened ("state cannot be changed. There are no new commits"). Content is unchanged — same commit, replayed onto the corrected base chain.*

### DIFF
--- a/.benchmarks/agentic-webserver/2026-09-06-823d87f8a7.json
+++ b/.benchmarks/agentic-webserver/2026-09-06-823d87f8a7.json
@@ -1,0 +1,490 @@
+{
+  "schema": 1,
+  "benchmark_id": "agentic-webserver",
+  "benchmark_name": "Agentic Webserver Test",
+  "git_sha": "823d87f8a7",
+  "recorded_at": 1788730397,
+  "target_model": "Qwen/Qwen3.6-35B-A3B-FP8",
+  "params": {
+    "build_timeout_s": "600",
+    "command_timeout_s": "180",
+    "iterations": "10",
+    "max_tokens": "8192",
+    "max_turns": "40",
+    "request_timeout_s": "900",
+    "s_per_turn_budget": "8.5",
+    "serve_timeout_s": "30",
+    "wall_budget_s": "1800"
+  },
+  "command": [
+    "spark",
+    "benchmark",
+    "run",
+    "agentic-webserver",
+    "--param",
+    "build_timeout_s=600",
+    "--param",
+    "command_timeout_s=180",
+    "--param",
+    "iterations=10",
+    "--param",
+    "max_tokens=8192",
+    "--param",
+    "max_turns=40",
+    "--param",
+    "request_timeout_s=900",
+    "--param",
+    "s_per_turn_budget=8.5",
+    "--param",
+    "serve_timeout_s=30",
+    "--param",
+    "wall_budget_s=1800",
+    "--yes",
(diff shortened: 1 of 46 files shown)


tbraun96:
Certified: 11 gates at `823d87f8a7`, one campaign for four PRs. Records and signatures are in `737de1aec1` — every record at the same `git_sha`, ten from one signer (dgx3) with only the second BFCL draw from another box (dgx2).

`#946` sits at the BOTTOM of this stack because it is the only fix here effective **on the branch**: #948 and #949 patch `issue_comment` workflows, which GitHub runs from the default branch, so their position changes nothing until they land. That placement earned itself — the flake #946 fixes broke the required `cargo test --workspace` on four separate PRs today (#871, #880, #938, #891).

Stamping and sealing in that order, at this head, since a mark is a check run bound to a sha and this branch is already current with `main`.

Issue #967, opened 2026-09-08
mtp_carry: the cross-turn drafter slot has no request identity — admitted on a two-token prefix, and reported ON when inert


tbraun96:
The cross-turn MTP drafter carry (`crates/spark-model/src/model/mtp_carry.rs`) is a
**model-level single slot** (`model/types.rs:259`) — one entry for the whole engine,
deposited by whichever sequence finished last (`trait_impl/sequence.rs:251-278`),
adopted by whichever sequence proposes next (`trait_impl/speculative.rs:170-190`).

Three separate comments say it is same-session only. None of them was enforced.
Three distinct defects follow; **D1 and D3 are fixed** on `feat/kat-sharding`
(490768533), **D2 is open**.

---

## D1 — admission was a two-token common prefix, with no identity at all *(fixed)*

`CarriedDrafter::usable_by` admitted any prompt sharing **two** leading tokens.
Every request rendered through one chat template agrees on hundreds, so on serial
traffic every request was eligible to adopt its predecessor's drafter KV.

| where | claims | did |
|---|---|---|
| `mtp_carry.rs:110-113` | "held for the next turn of the same **session**" | no session field existed |
| `mtp_carry.rs:124-128` | "may adopt only if its prompt **starts with exactly these tokens**" | truncated instead; admitted on `common >= 2` |
| `sequence.rs:251-254` | "so the next turn of the **same session** can adopt it" | stored `seq.tokens`, never `seq.session_hash` |

**Fix.** `CarriedDrafter` now stamps `seq.session_hash` at deposit and `usable_by`
refuses anything else, reusing the identity that already exists for SSM snapshot
ownership rather than inventing a second one. `0` refuses rather than allows —
deliberately unlike `SsmSnapshot::session_matches`, and like the sibling
`owns_capture` stamp, which requires a non-zero generation for the same reason.
The refusal is a distinct `CarryOutcome::ForeignSession`; reporting it as
`PrefixMismatch` — the benign, expected outcome of a re-tokenized turn boundary —
is how it stayed invisible.

The lever is preserved where it was measured to pay: `compute_session_hash`
(`session_manager.rs:133-140`) covers the first ≤1024 prompt tokens, so the
long-context session this was built for keeps a stable hash across turns.

**On the safety argument in the module header** (`mtp_carry.rs:47-50`): *"drafter KV
can never corrupt output … costs acceptance, not correctness."* That is sound for
[...]

Issue #936, opened 2026-09-06
BFCL: partitioning the draw four ways changes the score by 4 samples, and the partition is provably exact


tbraun96:
## Summary

Splitting the golden BFCL draw (995 samples) into four stride-partitioned shards and aggregating the results over integer per-subset `(hits, count)` tallies does **not** reproduce the unsharded score. The partition and the aggregation arithmetic are both provably exact, so **the model answered 4 of 995 samples differently** purely as a consequence of being asked in four groups instead of one.

| | overall | normalized | hits |
|---|---|---|---|
| unsharded (995) | **84.22** | **84.12** | 838 / 995 |
| A + B + C + D | **83.82** | **83.36** | **834** / 995 |

**This is not a scoring bug and it is not a partition bug.** Both were checked first (below). The most likely remaining explanation is that a prefix-cache or SSM-state hit is not numerically transparent — i.e. the answer to a sample depends on which requests preceded it.

## Why this matters beyond the sharding work

`normalized_single_turn_score` landed at **83.36 against a floor of 83.32**. The sharded run would have **passed the gate by 0.04** while being **0.76 low** — outside the ±0.4 noise band the entry declares. A silently-wrong score that reports green is the failure mode the gate exists to prevent.

If request order can move an accuracy score, it can move it for the unsharded gate too, whenever anything changes the order or the cache state.

## What was ruled out

**The partition is exact.** Shards are A=251, B=249, C=248, D=247 = 995, and every per-subset count matches the pinned draw plan:

| subset | summed n | subset | summed n |
|---|---|---|---|
| `simple_python` | 248 | `live_multiple` | 105 |
| `multiple` | 124 | `live_irrelevance` | 88 |
| `parallel` | 124 | `simple_java` | 62 |
| `parallel_multiple` | 124 | `simple_javascript` | 31 |
| `live_simple` | 25 | `irrelevance` | 24 |
| `live_parallel_multiple` | 24 | `live_parallel` | 16 |

The selection code is also correct by inspection: `load_shard` advances the within-subset position counter **whether or not the shard keeps the row**, so all four shards stride the same drawn rows and their union is the whole draw.

**The aggregation is exact.** Per-subset `(hits, count)` integers are summed and `score.py`'s category hierarchy (`live` sample-weighted, `non_live` hierarchical, `hallucination` unweighted) is applied once, to the totals. No rounding, no averaging of scores, no reweighting.

[...]

PR #968, opened 2026-09-08
mtp_carry: give the cross-turn drafter carry a request identity, and stop reporting it ON when it is inert

tbraun96:
Three correctness fixes to the cross-turn MTP drafter carry, plus two doc corrections. Split off `feat/kat-sharding` so they are not held behind the BFCL sharding work, which is embargoed by #936. Nothing here touches sharding; every commit is confined to `crates/spark-model/`.

Full write-up, including the audit that found them: **#967**.

## D1 — the carry slot had no request identity

`mtp_carry` is a MODEL-level single slot (`types.rs`): one entry for the whole engine, deposited by whichever sequence finished last, adopted by whichever proposes next. Three comments said it was same-session only. Nothing enforced it — admission was a **two-token common prefix**, and every request rendered through one chat template agrees on hundreds of leading tokens.

The entry now stamps `seq.session_hash` at deposit and `usable_by` refuses anything else, reusing the identity that already exists for SSM snapshot ownership rather than inventing a second one. A zero hash refuses rather than allows, following the sibling `owns_capture` stamp ("blind beats poisoned") rather than `SsmSnapshot::session_matches`, which treats 0 as legacy-allow. The refusal is its own `CarryOutcome`, because reporting it as a prefix mismatch — the benign, expected outcome of a re-tokenized turn boundary — is how it stayed invisible.

**Scope, stated precisely:** a session here is the first ≤1024 prompt tokens, not a client identity. Measured on a real 249-sample BFCL serve log, 249 requests produced 222 distinct hashes, one covering 19 samples. Requests sharing a ≥1024-token preamble still adopt each other's rows — which is sound, because `usable_by` bounds retained rows by the common prefix and `hidden_i` is a pure function of `tokens[0..=i]`.

## D2 — the shared hidden-row interval had no owner

`try_carry_drafter` could pair this request's tokens with another request's hidden states. `mtp_prefill_hidden` is one model-level buffer indexed by absolute position; `mtp_store_range` said which rows were written, never by whom. The cold path gates on an ownership stamp; the warm path had no equivalent, and `CarryOutcome::NoHiddens` is two integer comparisons proving **coverage**, not provenance.

The interval now carries the generation that wrote it. A foreign write takes it over instead of extending it; a foreign read sees an empty interval, which `plan_append` already refuses.

The stamp is deliberately **not** `mtp_capture_gen`: that field is assigned only under `chunk_start == 0`, and a warm turn never starts at 0, so it is `0` for the whole life of exactly the sequences the carry serves. The ticket is drawn unconditionally at `alloc_sequence` from the same atomic that issues capture generations.

## D3 — the startup line reported `carry=ON` when the carry could not run

`resolve_from_env` printed the two env vars. The runtime gates on `levers.drafter.carry && !mtp_multi_seq_mode()`, and `ATLAS_MTP_MAX_SEQS` **defaults to 32** — so the carry is force-disabled unless an operator sets it to 1. Verified on a live rig: the var is unset, the serve runs in-process, and the log nevertheless printed `carry=ON (default)`.

Both readers now share one predicate, `carry_armed`; the report prints the armed state; a configured-but-inert carry warns and names the cap. This one has a cost on the record: it is why a cross-request carry defect was hunted for a night, and why an A/B that turned the carry off was read as evidence when it could not have changed anything.

## Doc corrections

Three comments justified the single carry slot with "MTP is concurrency-1: every spec path is gated `active.len() == 1`". False since the ladder campaign — the cap defaults to 32, and `scheduler/phase_continue_prefills/spec_mixing.rs` already documents the staleness. Separately, `prefill_b.rs` had two comment blocks disagreeing about whether the tail split is conditional; the one describing the removed `peek_matched_tokens` guard is corrected.

A false safety comment is worse than none: the concurrency-1 claim is what made the missing ownership check look deliberate.

## Verification

Every new test was observed **red** against the reintroduced defect and green with the fix, with no other test moving:

| test | control | result |
|---|---|---|
| `a_foreign_session_cannot_adopt_the_slot` | session gate removed | FAILED → ok |
| `an_unstamped_request_cannot_adopt_the_slot` | session gate removed | FAILED → ok |
| `the_shared_template_prefix_is_not_an_identity` | session gate removed | FAILED → ok |
| `configured_carry_is_not_armed_carry_under_a_multi_seq_cap` | dispatch cap ignored | FAILED → ok |
| `a_late_chunk_takes_the_interval_over_instead_of_claiming_foreign_rows` | ownership ignored | FAILED → ok |
| `a_foreign_interval_cannot_satisfy_an_append_plan` | ownership ignored | FAILED → ok |
| `generation_zero_never_matches_anything` | ownership ignored | FAILED → ok |

The D2 tests compose the write and the read in scheduler order, because either half alone is trivially green — it was the composition that was broken. Each carries the unstamped value as an inline control: the late-chunk merge produced `(0, 1000)`, claiming rows 0..600 of another sequence, and the deferred-consume read satisfied `plan_append` with `Some(first_key: 12000, rows: 99)` of this sequence's tokens against another's hiddens. The two honest-case tests (`a_sequence_still_appends_from_its_own_rows`, `consecutive_chunks_of_one_sequence_still_merge`) stay green, so the guard refuses the bad orderings without refusing the feature.

`cargo test -p spark-model --lib` 671 passed · `cargo clippy -p spark-model --lib --tests` clean · `cargo check --workspace --tests` clean, all on this branch off current `main`.

## What this does not fix

- The stamp protects the interval, not the bytes: `mtp_prefill_hidden` is still one unversioned buffer.
- A host-side stamp does not order device work; the cross-stream visibility contract between the capture copy and the catch-up read was already unproven and is untouched.
- The carry remains unreachable on the shipped default (cap 32). This is observable only under `ATLAS_MTP_MAX_SEQS=1`, and it does **not** license raising the cap — the carry slot is still a single slot.
- **This is not a fix for #936.** The carry was never armed in any BFCL run, so it cannot be that mechanism. #936's channel is the SSM snapshot the tail split saves into a globally-evicted pool; see #936 for the measurements.

---

## A regression this PR introduced, found by its own campaign and fixed here

The ownership stamp above first drew its per-sequence ticket from
`mtp_prefill_capture_gen` — the counter `owns_capture` compares against. Every
`alloc_sequence` advanced it, so **any sequence admitted between another's
capture and its first propose silently turned that sequence's drafter prefill
off**. Fewer drafter rows, lower acceptance, slower decode.

| C | parent `main` | with the bug | again | fixed |
|---|---|---|---|---|
| 1 | 18.5 | 18.5 | 18.5 | 18.4 |
| **2** | **30.8** | **23.7** | **23.4** | **27.8** (floor 24.1) |
| 4 | 45.4 | 51.1 | — | 51.8 |
| 8–128 | 63.0 / 90.1 / 105.3 / 116.0 / 115.9 | all within ~1% | — | 65.5 … |

`concurrency-sweep` failed its C=2 floor twice, on two boxes, against a
same-morning control on the parent commit that scored 30.8 on the same box.
C=2 TPOT went 62 → 79 → 66 ms.

Fixed by giving the store ticket its own dispenser (`mtp_store_gen_seq`), and
guarded by a source-level test asserting the draw never names the capture
counter — observed red against the reintroduced defect. Source-level because
the coupling lives at a call site in `alloc_sequence_dispatch` that needs a
whole model to exercise.

**Reviewer note on what nearly happened.** Seven of eight cells matched and C=1
was *identical*, which I first read as proof of no regression. It is the
opposite: nothing is admitted between a lone sequence's capture and its propose,
so C=1 could not move under this bug — the invariant cell was the fingerprint
locating it. Without the control run on the parent commit this would have been
attributed to box noise on a `Sensitivity::Speed` gate whose C=2 cell already
sits near its floor.

**RESOLVED, and it resolved against my earlier reading.** This section
previously said 27.8 "sits below the parent's 30.8" and that one run each side
could not separate a residual gap from spread. The follow-up measurements
settle it: `c2_aggregate_tok_s` on this recipe is **TRIMODAL**, not noisy
around a mean — TPOT clusters at 62-63 / 66-67 / 79 ms, which is ~30.6 / ~27.5
/ ~23.5 tok/s, with nothing in between. 27.8 and 30.8 are two different MODES,
not two draws from one distribution, so their difference is not a gap to
explain. Verified from the run store: clean `main` on 09-07 scored
30.21 -> 23.70 -> 30.54 -> 23.56 in four consecutive runs at one commit.

I had also, on the same evidence, described a "24% C=2 regression" — the phrase
still in this branch's commit subjects. **That number is withdrawn.** It came
from comparing 3 parent samples that all happened to draw the fast mode against
2 candidate samples that both drew the slow one. The dispenser fix stands on
MECHANISM alone — the store ticket sharing the capture counter is a defect
whether or not it costs measurable throughput — and on the source-level test
that goes red against the reintroduced coupling. It does not stand on that 24%,
and nothing here should be read as claiming it.

The commit subjects cannot be corrected without rewriting `f9fd8ca43`, which is
the commit all eleven gate records are measured at; rewriting it would void the
certification and cost a full re-run. Flagging it here rather than silently
paying that.

**Certification status.** The fix touches `crates/`, so the ten gates that had
already passed at the previous pin are void. The campaign is running again from
scratch at `f9fd8ca43`; `concurrency-sweep` (C2 27.8/24.1) and
`concurrency-sweep-dflash2` (C2 48.0/35.4) are green, the rest are in flight.


tbraun96:
/stamp

11 of 11 required gates measured at `f9fd8ca43` and committed (`ac736e938`), run unsharded across dgx2 and dgx3. Local `--pull-request-gate-check` reports `all 11 required gates pass`.

| gate | result |
|---|---|
| bfcl-subset | 84.22 / 84.12, n=995 (floors 83.42 / 83.32) |
| bfcl-subset-echolp | 86.55 / 86.95, n=1004 (floors 86.10 / 86.50) |
| decode-floor | 22.6 tok/s (floor 22.2) |
| ttft-warm-gate | median −0.1%, p90 −0.9% |
| ttft-cold-gate | median +0.4%, p90 +0.1% |
| agentic-webserver | 10/10 ok, 10/10 followed, 5.3 s/turn |
| ssm-state-poisoning-gate | 12 of 12 replays byte-identical |
| vision-fidelity | 14 geometry cells, 3/3 probes, control held |
| video-fidelity | 14/14 legs, control held |
| concurrency-sweep | C2 27.8 (floor 24.1), 8 cells, zero vacuous |
| concurrency-sweep-dflash2 | C2 48.0 (floor 35.4), 5 cells, zero vacuous |

Both accuracy legs are unchanged from the pre-fix campaign and from BENCH.toml's 2026-08-14 measurement — 84.22/84.12 three times now, across two commits.

Two things stated rather than buried. **C2 27.8 sits below the parent's 30.8**; a repetition on the other box gave 27.7, so the fixed pin clusters, and parent-side repetitions are running to say whether the gap is real or C=2 spread. It clears the floor either way. **The `concurrency-sweep` record committed here is the campaign's**, not the one my characterisation repetitions produced — I had believed `--no-save` suppressed gate records and it does not; it only skips `~/.atlas/runs`.

tbraun96:
## Correction: the "24% C=2 regression" in this PR's description is NOT established. I withdraw it.

`concurrency-sweep`'s `c2_aggregate_tok_s` is **bimodal**, and I had that characterised in my own notes before I made the claim. Eleven reps on clean `main` across three boxes:

```
22.8 23.1 23.6 23.6 23.7   <- gap of 3.2, empty ->   26.9 27.5 28.9 29.7 30.2 30.5
slow mode n=5, mean 23.36 (~45% of runs)              fast mode n=6, mean 28.95
```

The effective floor (24.05) sits inside that gap, which makes this metric **a mode detector, not a regression detector** — it fails every slow run and passes every fast one.

My "evidence" for the regression was three parent samples (30.8, 30.8, 30.8 — all fast mode) against two candidate samples (23.7, 23.4 — both slow mode). Three consecutive fast draws is p ≈ 0.55³ ≈ 0.17. That is an ordinary outcome, not a measurement.

**The decisive tell:** the identical slow-mode fingerprint has since appeared at `76cbb8f43` (23.4), which does **not** contain the store-ticket code I blamed — that first appears two commits later. And committed records on `main` from unrelated commits span 27.04-30.41.

### What this does and does not change

**The code fix stands.** Drawing a per-sequence ticket from `mtp_prefill_capture_gen` genuinely advances the counter `owns_capture` compares against, so it really can disable another sequence's drafter prefill. That is a defect by inspection and `mtp_store_gen_seq` is the right fix, guarded by a test observed red. **It is justified by the mechanism, not by a throughput number.**

**The numbers in the PR body are withdrawn.** Specifically: "24% of C=2 decode", the four-column cell table read as before/after, and the TPOT 62 → 79 → 66 ms narrative. TPOT is itself a mode label (fast 62-66, slow 79-81 at identical TTFT and delivered tokens), so it tracked the mode rather than confirming a mechanism.

**The certification is unaffected.** All 11 gates were measured at `f9fd8ca43` and pass on their own thresholds; `concurrency-sweep` cleared 24.1 at 27.8. Nothing here depends on the withdrawn attribution.

### What would actually settle it

Interleaved A/B on ONE box, ≥6 reps per arm, comparing the **mode split** rather than the mean, with TPOT as the mode label. Three pairs are running now and I will extend them; the earlier comparisons were confounded by box (all parent samples on dgx3, most candidate samples on dgx2) and then by time (all parent samples before 14:02Z, all candidate samples after 14:29Z).

I should have checked the metric's distribution before attributing a delta to a diff. I cited the bimodality note twice the same day for a different reason and never applied it to my own diagnosis.

tbraun96:
/seal

Eleven gates, eleven PASSes, all measured at `3cfe4d8c45` on a tree current with `main`.

| gate | box | verdict |
|---|---|---|
| agentic-webserver | spark-43fa | Pass (787s) |
| concurrency-sweep | spark-43fa | Pass — 8 cells, zero errors, **zero vacuous** |
| concurrency-sweep-dflash2 | spark-43fa | Pass — 5 cells, zero vacuous |
| decode-floor | spark-43fa | Pass — 26.2 tok/s vs the 22.2 floor |
| ttft-cold-gate | spark-43fa | Pass — median +0.3% (limit +3.0%) |
| ttft-warm-gate | spark-43fa | Pass — median −0.2% (limit +3.0%) |
| bfcl-subset | spark-43fa | Pass — 84.22 / 84.12, n=995 |
| bfcl-subset-echolp | spark-28c2 | Pass — 86.55 / 86.95, n=1004 |
| ssm-state-poisoning-gate | spark-28c2 | Pass — 12/12 replays byte-identical |
| vision-fidelity | spark-28c2 | Pass — 14 geometry cells, 3/3 probes |
| video-fidelity | spark-28c2 | Pass — 14/14 legs, control held |

Both agreement rules hold structurally: one commit across all eleven, and all six **speed-class** gates on one box under one signing key (`efe157d6f6360027`). The five correctness gates span both boxes, which the rule permits — and which is what let the two 3.5 h BFCL legs run in parallel.

**Why this campaign replaced the previous one rather than repairing it.** The earlier set carried a real defect — six speed records under two keys — and fixing that was necessary but not sufficient: `main` had moved **239 files under PERF_PATHS** since the merge-base, and `record_covers` diffs trees over exactly those paths, so the old campaign described a tree that would never merge. Its records were deleted in the merge commit rather than left beside these, because `Disagreement::Commits` is fatal for every class.

**On reading the throughput numbers:** `decode-floor` scored 26.2 here against 22.6 at the previous pin, and `concurrency-sweep` put C=2 at 31.3 against 27.2 ninety minutes earlier on the same box. Neither is evidence of a speedup — that cell is trimodal (~30.6 / ~27.5 / ~23.5 tok/s). The campaign clears floors; it does not measure a delta.

**Known and not fixed:** the commit subjects on this branch still assert a "24% of C=2 decode" regression that has been withdrawn. Correcting them would rewrite the commit these eleven records are measured at and void the campaign. The withdrawal is recorded in this PR body and in `docs/ROBUSTNESS.md` wave 34.

Issue #835, opened 2026-09-01
concurrency-sweep: the C=2 rung is bimodal and its floor sits inside the gap


tbraun96:
## What

`concurrency-sweep`'s C=2 cell is bimodal on gb10, and the gate's floor sits between the two modes — so the gate flakes for any PR, not just the one that surfaced it.

Measured 2026-09-01, all on the same box, same night:

| build | C=2 aggregate tok/s | verdict |
|---|---|---|
| #831 + fix, run 1 | **22.8** | FAIL |
| #831 + fix, run 2 (same commit) | **28.0** | PASS |
| **`origin/main` `e2dc04aac3`** | **22.9** | **FAIL** |

Floor is `min = 25.0`, `noise = 0.95` (`kernels/gb10/qwen3.8-27b/BENCH.toml`), and the run reports it as 24.1.

## Why it matters

**`main` fails this cell too.** The control leg above was run specifically to separate "PR regression" from "box tonight", and it landed at 22.9 — squarely in the low mode. So this is not attributable to any particular change; any PR that draws the low mode fails a required gate for no reason, and a re-run "fixes" it, which is the worst shape a gate can have.

Every other cell was stable across the same runs — C1 18.2/18.3, C4 44.8-50.8, C8 60.2-60.8, C16 88.7-89.2, C32 104.1, C64 115.4, C128 114.9-115.0. Only C=2 moves, and it moves ~5 tok/s.

## Context that may explain the mode

C=2 is where speculation engages: the sweep logs `speculation ENGAGED at width n=2 (dispatch cap 32)` at that rung, and disengages at the wide ones. So C=2 is the narrowest rung where the MTP verify path is active, and it is plausibly bistable between engaged/disengaged behaviour within a rep.

## What would fix it

Not a floor lowered to 22 — that would make the cell meaningless. Either:
1. find why the rung is bistable and remove the bimodality, or
2. increase reps at C=2 specifically and gate on the median, or
3. if the low mode is legitimate behaviour, split the cell so the gate asserts something that is actually stable.

## Provenance

Surfaced during the 11-gate certification campaign for #831 (merged as `ef8ca11d98`). The `origin/main` control record is deliberately **not** committed to `.benchmarks/` — a FAIL record for main would mislead later gate checks — so the numbers are recorded here instead.

PR #971, opened 2026-09-08
levers: read every ATLAS_* variable once, not per token — and the four flags that had two implementations

tbraun96:
## What this is

Every `ATLAS_*` environment variable read from a per-token, per-layer or
per-forward path, resolved **once** and carried instead.

`std::env::var` allocates a `String` and takes the **process-wide environment
lock**, so concurrent decode threads serialise against each other on it.
Measured on GB10: one resolve of ~30 variables costs **0.57 µs
single-threaded, 4.00 µs at 8 threads, 5.76 µs at 16**. The cost grows with
concurrency, which is exactly why no single-stream benchmark ever showed it —
and why `ModelLevers::from_env` was being called **32,513 times in one
`concurrency-sweep`** while its own doc claimed it was "called once, when the
model is built".

## What this is NOT

**A throughput win, and I am not claiming one.** The arithmetic: ~975K reads
per sweep at the 8-thread cost is ~130 ms against a ~1500 s sweep — about
**0.01%**. An interleaved A/B on one box (merge-base vs this branch, 2 pairs,
all 8 concurrency rungs) is consistent with that: every rung inside ±2.9%
except C=2, where the *baseline arm alone* drew 28.7 then 25.2 tok/s at the
identical commit — a 12% swing against itself. That cell is trimodal and
cannot carry a claim from a handful of draws.

The A/B was run as a **polarity check**, not a speed measurement, with the
decision rule fixed before the first rep: a flipped lever is a
tens-of-percent effect (`ATLAS_DFLASH_OPTION_B`'s flip in August took propose
19.8 → 618.7 ms and 49.9 → 5.5 tok/s with nothing logged), and no rung is
≥15% down in any pair. Rungs where either arm delivered <80% of `osl` are
excluded — the merge-base run hit the known C=8/32/128 truncation, and
comparing a truncated cell would read vacuity as a regression.

## The reason it was worth doing

Four variables turned out to have **more than one independent implementation**:

| Variable | Where |
|---|---|
| `ATLAS_DSPARK_ANCHOR_BIAS` | `forward_block` gated confidence truncation on one read; `markov_argmax_block` gated row-0 bigram bias on its own |
| `ATLAS_DFLASH_DEBUG_DUMP_FULL` | the model half and the drafter half of one dump |
| `ATLAS_DFLASH_EAGLE_FIX` | `verify_dflash_step.rs` and `verify_k2_step.rs`, each per verify step |
| `ATLAS_DUMP_LOGITS_PATH` | `sampler.rs` and `decode_logits_seq.rs` |

Three of the four ship **ON** and gate correctness-relevant behaviour, so a
divergence between two spellings would disable a fix on one lane and leave it
live on the other. Where a second copy is unavoidable — `TransformerModel`
cannot reach the drafter head's levers through a `dyn DraftProposer` — the
duplication is now a **checked invariant** rather than an unchecked one.

## Guards

`hot_path_env_guards.rs` holds one table of 21 guarded files with the
functions in each still permitted to read, attributing every read to its
enclosing `fn`. A count-based budget would not do: deleting a cold read and
adding a hot one keeps the count. `spark-server` has the sibling table for
its scheduler.

Every guard was **observed red before being trusted**, and two of them failed
that check first time:

- the source-level agreement guard used `contains`, and the typo injected to
  break it (`..._FULL_TYPO`) contains the correct name as a prefix — the
  control passed and the guard was measuring nothing;
- a control reported `test result: ok` without running at all, because
  reinjecting the read left a parameter unused and the build failed, and a
  build failure prints no verdict.

## Deliberately not cached

- `refusal::detect` — per request, and `kill_switch_returns_none` mutates its
  variable under a documented `Mutex` because cargo runs tests in parallel
  threads. A `OnceLock` would latch whichever state won the race.
- `cutlass::ctx` — allocates the shared 512 MB workspace once, at lazy init.
- `MtpHead::last_confidence` — reached only when the confidence clamp is
  already armed, which no shipped config does.

Each is named in a guard's allow list with its reason, so the exemption is a
recorded decision rather than an omission.

## Verification

`cargo test --workspace --locked` (77 suites) and `cargo clippy --workspace
--lib --bins --tests` clean at every commit. Note `--lib` alone is not
sufficient here: `spark-server`'s scheduler hangs off `main.rs`, so it
excludes all 2,405 of that crate's bin-target tests.

Draft because it has not yet had CI or a benchmark campaign.


tbraun96:
Closing: already on `main` through #981 (`8a68f2602f`, "Fix #936 + #835 + #971 on one composed branch: --hermetic reaches 0/995 exact equality"), whose message records that it carries this PR's 68-file hot-path diff.

Re-checked today against `main` `9e597531`: all 56 functions this PR adds and all 7 files exist there — e.g. `the_dense_ffn_levers_are_presence_gated_and_their_polarities_hold` at `crates/spark-model/src/layers/ops/model_levers_hot_path_tests.rs:27`. No branch deleted.

PR #972, opened 2026-09-09
Qwen3.8-Flash-Next NVFP4: K-row MTP verify body and HC decode-rows kernels (one Spark, C=1: LRU 46.3 to 51.5 tok/s on this PR, 62.0 with the follow-up)

rrstesiak:
**Base: `wip/exl3-research` (#834) at `dba869c0`, the head this was built and measured on. Merges clean onto the current head `fb8aaca3`. Flash-Next speculation is not on `main`; this lands when the MTP verification layer from #834 does.**

## Where it stands

Qwen3.8-Flash-Next, `nvidia/Qwen3.8-Flash-Next-NVFP4` @ `fc694b54`, one DGX Spark GB10 (hostname `blackbird`), C=1, MTP-2, temp 0, `reasoning_effort` low. Medians of three; streams byte-identical run to run on every cell.

| cell | per-row verify + decode-rows kernels | + stage 1 (K-row HC/FFN body) | + stage 2 (projections at T=K) = **this PR** | + wide-verify fast greedy (next PR) | vLLM, published |
|---|---|---|---|---|---|
| LRU 512 (Richard's prompt from `bench/qwen4_exp`) | 46.3 | 48.6 | **51.5** (48.45 / 51.54 / 51.39) | 62.0 | 46 |
| MinHeap 300 (code) | 42.4 | 50.0 | **49.9** (47.03 / 49.91 / 49.89) | 58.6 | |
| Volvo 300 (prose) | 39.6 | 47.4 | **47.6** (47.68 / 47.58 / 47.54) | 54.0 | |

Base code serial number on this pack and prompt: 20.8. The vLLM column is the one published single-Spark number on this pack; its prompt is not matched to these cells.

nsys, Volvo, one decode step (medians over ~390 steps): per-row bodies 58.7 ms, 1701 launches; stage 1 52.3 ms, 1383 launches. The kernel reworks in the last three commits were measured on the full stack; their LRU deltas were +2.7, +1.2 and +0.8 tok/s with streams identical to the prior reference except where the reduction order changed.

## What it does

All three paths ship ON. Each keeps its variable as the kill switch, explicit `0`, the same convention as the loader switch: `ATLAS_HC_DECODE_ROWS=0`, `ATLAS_QWEN4EXP_MTP_HC_ATTN_ROWS=0`, `ATLAS_QWEN4EXP_MTP_HC_ATTN_ROWS_QKV=0`. With any of them at `0` that path is byte-identical to the base, which is what the A/B and the rollback need.

[...]

PR #973, opened 2026-09-09
Qwen3.8-Flash-Next NVFP4: wide MTP verify takes the GPU argmax when the host pipeline provably cannot change the pick (one Spark, C=1, on this PR's tree: LRU 57.2 to 62.0 tok/s)

rrstesiak:
**Stacked on #972: base `wip/exl3-research` (#834) at its current head `fb8aaca3`, then #972's ten commits, then this change. Until #972 merges, GitHub shows #972's commits here as well; the diff that belongs to this PR is the last two.**

## Where it stands

Qwen3.8-Flash-Next, `nvidia/Qwen3.8-Flash-Next-NVFP4` @ `fc694b54`, one DGX Spark GB10, C=1, MTP-2, temp 0, `reasoning_effort` low. Medians of three; streams byte-identical run to run on every cell. Both columns are the same build of this PR's head `4b2ed804`, quiet box, the only difference being the switch.

| cell | this tree, `ATLAS_MTP_THINK_FAST_GREEDY=0` (the #972 path) | **this tree, default (this PR on)** | vLLM, published |
|---|---|---|---|
| LRU 512 (Richard's prompt from `bench/qwen4_exp`) | 57.2 (54.18 / 57.39 / 57.20) | **62.0** (58.69 / 61.96 / 61.92) | 46 |
| MinHeap 300 (code) | 51.9 (49.06 / 51.85 / 51.85) | **58.5** (54.73 / 58.45 / 58.78) | |
| Volvo 300 (prose) | 49.9 (50.01 / 49.91 / 49.87) | **54.0** (54.03 / 54.03 / 53.92) | |

On LRU the two streams are identical, so this stream has no exact tie at the maximum and the gain is pure step time. On MinHeap and Volvo the streams differ past the first tie, as described below. The same change measured 51.5 to 57.0 on LRU on the pre-rebase tree before #972's kernel reworks; nsys on that tree showed the step's biggest idle gap, about 3.4 ms of 52, with no CUDA calls: the host was copying the `[K, vocab]` logits and running the full pre-sample pipeline for every row of every verify step.

## What it does

`verify_mtp_wide::finish` copied the K-row BF16 logits to the host and ran `process_seq_logits` (248k dequant, the eight-stage pipeline, host argmax) on every row of every step, in-think or not, discarding the graphed verify's own GPU argmax IDs. It never reached the existing fast arms: the chat fast path is gated on not being inside thinking, and the wide verify did not call it at all.

`row_fast_greedy()` proves, on the live per-row state, that `process_seq_logits` would return the raw argmax: temp 0, no grammar, no logprobs, no diagnostics, penalties classified argmax-preserving by the SSOT `fast_greedy` gate (reduce-only needs the argmax immune, one 2-byte D2H), no confidence stage armed, no forced-think-end injector armed or deferring, no pin-to-tool-call, and the pick is not a structural id. Eligible rows emit the GPU argmax. The first ineligible row triggers one lazy D2H and the exact host path for the rest. Serve log: `MTP verify fast greedy ACTIVE`, plus one line per fallback reason seen.

Ships ON. `ATLAS_MTP_THINK_FAST_GREEDY=0` restores the unconditional D2H and host path byte-identically, the same convention as `ATLAS_NO_FAST_GREEDY_CHAT`. The slow arm keeps the scheduler's borrowed staging buffer from the aux-collect change.

## Tie-break disclosure

Emitted tokens can differ from the slow path where two logits tie exactly at the maximum: the GPU argmax kernel and the host scan resolve such ties by different index rules. This is the trade the shipped chat fast path already makes (its own comment cites #237); this PR extends it to MTP verify rows inside the think block, which until now always took the host path. Run to run the streams are byte-identical.

## Tests and gates

- Three runs per cell, byte-identical text within each tag; on MinHeap and Volvo the text differs from the switch-off run only past the first exact tie; on LRU it is identical.
- Fallback reasons observed on the record cells: structural argmax, think-end injector. Both took the exact host path.
- MTP-2 on this stack is deterministic but not serial-exact, as on the base.

## Reproduction

```
ATLAS_TARGET_HW=gb10 ATLAS_TARGET_MODEL=qwen3.8-flash-next ATLAS_TARGET_QUANT=nvfp4 cargo build --release -p spark-server
```
Serve per `bench/qwen4_exp/serve_nvidia_mtp.sh` with `MTP_DRAFTS=2`, `--gpu-memory-utilization 0.76`, `ATLAS_QWEN4EXP_BF16_GDN=0`. Control on the same build: `ATLAS_MTP_THINK_FAST_GREEDY=0`. Prompts and protocol as in #972.

## Credits

Richard's batched verify argmax on this branch (one launch, one D2H in the K-row verify tail) is what this PR's rows now read; an equivalent commit from this series was dropped in its favor. The borrowed staging buffer in the slow arm is his aux-collect change.


tbraun96:

tbraun96:

PR #981, opened 2026-09-10
Fix #936 + #835 + #971 on one composed branch: --hermetic reaches 0/995 exact equality

### LINKED ISSUE #936: BFCL: partitioning the draw four ways changes the score by 4 samples, and the partition is provably exact

tbraun96:
Composes the three blocked issues onto one branch so they certify once instead
of four times. The gate judges every layer against `main`, not its own base,
and `PERF_PATHS[0]` is the bare string `crates` — so a native 4-layer stack is
four distinct trees (~17 GPU-h) where this is one (~4.5 GPU-h).

## What lands

**#936 — the KAT split does not reproduce the unsharded score.**
The cause is cross-request state, and this branch closes it under one name.

- `spark serve --hermetic` — one recorded regime that expands *in code* to the
  channel closures, so they cannot drift apart and the record carries a single
  unambiguous name. It closes the radix KV prefix cache (keyed on token content
  with no session component) and pins the MTP gate (which does not only switch
  arms, it *probes*: `tokens_since_event` is cumulative and never reset at a
  request boundary, so which request is serving when the counter rolls over is
  a function of everything served before it).
- The snapshot session gate is now **one predicate**, read by both lookups.
  It existed as two copies — and only `lookup_tiered` is on the serving path,
  so editing the other would have compiled, read correctly, and changed nothing.
  Under `--hermetic` it gates every entry, not only `is_tail`.
- `kat-equality-gate` — replays a draw against one server in ≥2 request orders
  and requires every `sample_id` byte-identical.

**#835 — `c2_aggregate_tok_s` is trimodal and the floor sits in the empty gap.**
No floor is moved. The run now **publishes** `c{c}_accept_len` (1.00 on the
serial arm, ~2.3 on MTP — near-zero within-mode variance) plus
`non_mtp_arm_cells`, so a human reading a low C=2 can see whether the serial arm
explains it. That is what #835 needs: the false-regression report comes from a
reader who cannot tell a mode change from a regression, and the record now tells
them.

★ It PUBLISHES and does not GATE, and that distinction was bought with a
campaign. The first implementation treated a serial-arm cell as
non-comparable — excluded from scoring, run failed INCONCLUSIVE. Its tests
passed and its negative control went red, and it was still wrong: at wide batch
the MTP gate drops speculation ON PURPOSE, so C=8 upward legitimately run
serial and those rungs' floors were calibrated on runs that did. The pin dropped
five of eight cells, made `peak_aggregate_tok_s` read 49.5 (from C=4) instead of
~115 (from C=64), and failed a gate carrying nine consecutive passing records —
while C=2, the cell the issue is about, ran the MTP arm and passed at 30.02
against a 25.0 floor. The benchmark cannot know which arm SHOULD have run;
publishing the fact is informative, refusing on it asserts knowledge it does not
have. See `docs/ROBUSTNESS.md`.

**#971 — rustdoc blocker.** A `pub` field linked a `pub(super)` item under
`deny(warnings)`, so CI aborted on `spark-model` and never reached
`spark-server` at all. Fixed, and `spark-server` now has its first clean
rustdoc pass.

## ★ What this PR does NOT turn on

**The BFCL gates still certify UNCLOSED in this PR, and sharded certification
is not enabled.** Both are deliberate and neither is a loose end.

`bfcl-subset`'s baseline pins no serve overrides. A hermetic run's record
carries three (`hermetic` plus the two it expands into), and `check_record`
compares record against baseline in both directions — so a hermetic record
cannot discharge that gate until its entry pins them, which is PR 2's job. The
campaign for THIS branch therefore runs the gates in the shipped regime, which
matches their baselines exactly.

The shard descriptors and the `bfcl-subset` group are registered here, but a
whole-draw record still satisfies the group gate, so certifying unsharded
remains the path this PR takes. Sharding becomes SAFE to rely on when hermetic
becomes the gate regime — measured here, declared there.

In short: this PR ships the mechanism and proves equality is reachable. It does
not yet change how the KAT gates are measured.

## What this PR deliberately does NOT claim

- **No floor is declared.** `--hermetic` changes the scores — measured, not
  assumed: the whole BFCL leg scores 84.22 / 84.12 on `main` and 83.92 / 84.22
  under the closures. Floors are re-cut from measured runs in a follow-up. No
  bar certifies itself.
- **`kat-equality-gate` registers as a PROMOTION CANDIDATE, not a required
  gate**, for the same reason: its bar is that the shipped regime is
  order-independent, and a gate may not certify itself in the change that first
  records it. It accrues visible debt meanwhile, so "ungated" never reads as
  "unaffected".
- **The MTP dwell-counter defect is recorded, not fixed.**
  `SWITCH_DWELL_WINDOWS = 2` does not require two *fresh* comparisons — `stale`
  means "the economics moved", never "this number is old" — so one probe can
  carry both windows. The fix changes when arms switch, which is
  throughput-affecting, and this branch already carries #971's 68-file hot-path
  diff. Own PR, own A/B. See `docs/ROBUSTNESS.md`.

## The measurement

The acceptance criterion was EXACT equality — the whole BFCL draw and its own
four shards answering identically for every `sample_id`, not "within noise".

| arm | serve config | n | disagreeing |
|---|---|---|---|
| historical (pin `e897463b54`) | shipped | 995 | 12 |
| base (current `main`) | shipped | 995 | **12** |
| C | `mtp_gate=force` + `enable_prefix_caching=false` | 995 | **0** |
| **D** | **`--hermetic`, the shipped flag** | **995** | **0** |

Arm D's whole leg is byte-identical to arm C's ACROSS TWO BOXES
(`a9aadaa1…`) while base differs — so the flag is not approximately the
hand-configured pair, it is exactly it, and the engine is bit-reproducible
under hermetic across machines.

Two things follow. **#968 did not close this** — the count is identical to the
one measured months ago at a pin that is not an ancestor of `main`, so the
drafter-carry session gate that landed in between moved it by zero, and the
channel is the prefix cache and the MTP probe. And **equality is reachable**,
which the plan listed as its first risk.

Both levers were proven armed and proven to be the only difference: the base
arm logs `Prefix caching: ENABLED (radix tree)` / `mtp_gate=auto`, arm C logs
`Prefix caching: disabled` / `mtp_gate=force`, and every other entry in the
`kernel flags:` line is identical. That check is not ceremony —
`enable_prefix_caching` DEFAULTS to false, so the override could have been a
no-op, and only the base arm's log proves the recipe had turned it on.

★ **The instrument is the per-sample diff, never the score.** The whole leg
scores 84.22 / 84.12 and the four shards recombine to close to the same totals,
so a score comparison — the obvious thing to run — reports agreement while
twelve samples answer differently. Two different sets of answers can total
identically. An earlier reading of this same bug went wrong exactly here, by
comparing aggregate rates when the mechanism predicts an invariant population
rate with varying membership.

## Verification

- `cargo test --workspace --locked` — 83 suites green
- `cargo clippy --workspace --lib --bins --tests` — 0 lints
- `ATLAS_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo doc --workspace --no-deps` — exit 0
- `bash .github/scripts/certification-selftest.sh` — 226 passed, 0 failed
- file-size cap — 1993 files scanned, 0 over cap and not allow-listed
- Every new guard has a **negative control that was run and observed red**;
  `docs/ROBUSTNESS.md` wave 35 names each one, including the control that came
  back GREEN and thereby found a decorative test.

## What PR 2 must do, and why it cannot be this PR

Nothing here declares a bar it also measures. PR 2 does, and its contents are
already determined by what this branch measured — it is bookkeeping on evidence,
not a fresh design:

1. **Re-cut the BFCL floors from hermetic runs.** `--hermetic` is not
   score-neutral: 84.22 / 84.12 unclosed vs 83.92 / 84.22 closed on the whole
   leg. Both clear the committed bars today, so nothing is blocked — but a floor
   cut on the unclosed regime does not describe the closed one, and the gap is
   the wrong side of the noise term to leave to chance.

2. **Add the `kat-equality-gate` BENCH.toml entry**, pinning `hermetic = "true"`
   AND `enable_prefix_caching = "false"` AND `mtp_gate = "force"`. All three:
   `gate::bench` now REFUSES an entry that pins the first without the others,
   because `--hermetic` expands into them and `check_record` compares the record's
   overrides against the baseline's pins in both directions. The refusal exists
   so this is a parse error rather than a discovery made after a campaign.

3. **Promote the gate from `PROMOTION_CANDIDATES` to `REQUIRED`** — array lengths
   are in the types, so `REQUIRED` goes 11 → 12, `REQUIRED_GATES` likewise, and
   the `NOT_REQUIRED` entry (which currently records "not required YET, and
   deliberately not in the PR that introduces it") comes out, 6 → 5.

4. **Pin the `sample_cap` that was actually measured.** The default is 0 (the
   whole draw), which is honest but costs ~2× a BFCL leg. A smaller cap is
   defensible for a routine gate and must be stated with its power: the effect
   this gate exists to catch was 12 samples in 995, so a cap of N has
   proportionally less chance of seeing a reintroduction. Say the number, do not
   inherit it.

Each of those is a bar, and a bar may not be cut from the run that judges it.

## Campaign: 11 of 11, one commit

Measured at `5a8293500`, every record signed, verified by the product's own
reader (`spark benchmark --pull-request-gate-check` → *all 11 required gates
pass*).

| gate | class | result |
|---|---|---|
| `agentic-webserver` | speed | 10/10 · 4.711 s/turn ≤ 8.500 · Σwall 758s |
| `ttft-cold-gate` | speed | median +1.4% (limit +3.0%) |
| `ttft-warm-gate` | speed | median +0.8% · p90 −0.1% |
| `decode-floor` | speed | 26.1 tok/s vs 22.2 floor · `accept_len` 2.68 |
| `concurrency-sweep` | speed | 8 cells, every populated floor met, peak 116.1 @ C=64 |
| `concurrency-sweep-dflash2` | speed | 5 cells, every populated floor met |
| `bfcl-subset` | correctness | 84.22 / 84.12 (floors 83.42 / 83.32) n=995 |
| `bfcl-subset-echolp` | correctness | 86.55 / 86.95 (floors 86.10 / 86.50) n=1004 |
| `ssm-state-poisoning-gate` | correctness | pass |
| `vision-fidelity` | correctness | pass |
| `video-fidelity` | correctness | pass |

All six SPEED gates ran on one box under one signing key, as `gate::agreement`
requires. Correctness gates may span boxes and did.

**No speed regression anywhere** — the plan's headline risk for #971's 68-file
hot-path diff. All eight concurrency rungs land within ~1% of `main`'s last
passing record; cold TTFT moved +1.4% against a 3% limit and warm's p90 is
negative.

## ⚠ This PR should not be self-sealed

Sealing a certified campaign is delegated by default here, with an enumerated
list of exceptions. One applies: *the merge lands a `BOUNDARY_FILE` change whose
only evidence is this campaign.* This branch touches **7 of the 15 declared
BOUNDARY_FILES**, four of them new — including `gate/coverage.rs`, which
declares that list, and `gate/bench.rs`, which now refuses an under-pinned
hermetic baseline.

Those files decide whether any gate passes, so offering the campaign they govern
as proof that they are correct is circular. Unit tests and the 226-check
certification self-test cover the logic; the campaign is the only *behavioural*
evidence, which is precisely the condition the rule names.


tbraun96:
## Certification evidence

**Campaign: 11 of 11, one commit (`5a8293500`), every record signed.**
Verified two ways — `record_agreement` over the set `git diff --diff-filter=AM
base...HEAD` reports ("all 11 added record(s) agree: one commit, and signer
agreement holds for every speed-class gate"), and `spark benchmark
--pull-request-gate-check` reports "all 11 required gates pass".

| gate | class | result |
|---|---|---|
| `agentic-webserver` | speed | 10/10 · 4.711 s/turn ≤ 8.500 · Σwall 758s |
| `ttft-cold-gate` | speed | median +1.4% (limit +3.0%) |
| `ttft-warm-gate` | speed | median +0.8% · p90 −0.1% |
| `decode-floor` | speed | 26.1 tok/s vs 22.2 floor · accept_len 2.68 |
| `concurrency-sweep` | speed | 8 cells, every populated floor met, peak 116.1 @ C=64 |
| `concurrency-sweep-dflash2` | speed | 5 cells, every populated floor met |
| `bfcl-subset` | correctness | 84.22 / 84.12 (floors 83.42 / 83.32) n=995 |
| `bfcl-subset-echolp` | correctness | 86.55 / 86.95 (floors 86.10 / 86.50) n=1004 |
| `ssm-state-poisoning-gate` | correctness | pass |
| `vision-fidelity` | correctness | pass |
| `video-fidelity` | correctness | pass |

All six SPEED gates carry ONE signing key (`efe157d6f636…`), as `gate::agreement`
requires. The two correctness gates on the second box carry another
(`02156264cbf7…`), which that rule explicitly permits.

**The measurement this branch exists for.** Whole BFCL draw vs its own four
shards (251+249+248+247 = 995), one box, one commit, temperature 0:

| arm | serve config | n | disagreeing |
|---|---|---|---|
| historical (pin `e897463b54`) | shipped | 995 | 12 |
| base (current `main`) | shipped | 995 | **12** |
| C | `mtp_gate=force` + `enable_prefix_caching=false` | 995 | **0** |
| **D** | **`--hermetic`, as shipped** | **995** | **0** |

#968 moved the count by zero, so the channel is the prefix cache and the MTP
refresh probe. Arm D's whole leg is byte-identical to arm C's across two
different boxes (`a9aadaa1…`) while base differs — so the flag is exactly the
hand-configured pair, not approximately it. The plan's demanded negative control
is the base arm going red on twelve.

**No speed regression anywhere** — all eight concurrency rungs within ~1% of
main's last passing record; this was the stated risk for #971's 68-file hot-path
diff.

**Required checks: 20/20 green.** Two non-required checks are red and both are
expected: `seal status` (red until sealed) and `PR categorize (advisory)`, which
fails on shallow-clone depth and correctly refuses to classify from the title
and body alone.

**Boundary files.** This lands changes to 7 of the 15 declared `BOUNDARY_FILES`,
four of them new. Sealing it was held for a human on exactly that ground and has
now been authorised by the repository owner in-session.

PR #982, opened 2026-09-10
gate: promote kat-equality-gate to REQUIRED — 36 divergences open, 0 closed

tbraun96:
Promotes `kat-equality-gate` from `PROMOTION_CANDIDATES` to `REQUIRED`, with its instrument pinned in `BENCH.toml` — the follow-up #981 said it owed. Along the way it fixes two defects #981 shipped.

## The measurement

Same box, same cap, same instrument — only the serve regime differs:

| | open | pinned (`hermetic`) |
|---|---|---|
| **diverged** | **36 of 257** | **0 of 257** |
| identical | 221 | 257 |
| unmeasured | 0 | 0 |
| empty_replies | 0 | 0 |
| wall | 3878 s | 4061 s |

The open arm is the negative control. The record (`.benchmarks/kat-equality-gate/2026-09-10-4db613c1d8.json`) carries `serve_overrides {enable_prefix_caching: false, hermetic: true, mtp_gate: force}` and `params {max_new_tokens: 1024, orders: 2, sample_cap: 257}`, so the lever is proven armed rather than assumed.

**36 is not a regression against #936's 12.** Different instruments: #936 compared the whole draw against its own four shards; this gate reverses the request order against ONE server, which perturbs more. The driver's cap warning and `sample_cap` help both cited the 12 as though it described this gate — corrected here.

## The cap is a prefix, not a sample

`sample_cap` calls `Vec::truncate`, and the draw concatenates subsets in sorted name order, so the cap decides **which subsets are compared**, deterministically. 257 is the end of `live_parallel_multiple` — the smallest prefix covering every subset order-dependence has been observed in (10 `live_irrelevance`, 1 `live_multiple`, 1 `live_parallel_multiple`). It gives up the 738-sample `non_live` tail, which has never produced one.

The control was re-run **at the cap** rather than inherited, because truncating changes what ran before every surviving sample: a control taken on the whole draw describes a different instrument.

Cost: **~65 min**. The whole draw in two orders would be ~4.2 h on every PR, against a ~4.5 h campaign.

## Two defects in #981

- **The gate generated at the wrong budget.** `kat_equality` defaulted `max_new_tokens: 512` against BFCL's 1024, while its own doc says the body "matches BFCL's exactly … a body that differed would answer about a regime nobody measures". Every committed `bfcl-subset` record, and both the 12/995 and 0/995 measurements, were at 1024. There were four literals for one value; now one exported `MAX_NEW_TOKENS`.
- **The equality driver re-opened every other gate.** It shipped in no sibling exclusion list, so editing the detector cost a full 11-gate recampaign (`ssm_poison` is excluded by five arrays). Added to all eight siblings — deliberately not to its own, so a change to the detector still re-opens the detector.

Also corrects `group.rs`'s closure prose, which named `ssm_cache_slots = "0"` plus a floor re-cut; what shipped is `--hermetic`, and no floor re-cut was needed.

## What this does NOT do

- **No BFCL floor changes.** The closed regime scores 83.92 / 84.22 against bars of 83.82 / 83.72 with noise 0.4.
- **No `hermetic` pin on `bfcl-subset` / `bfcl-subset-echolp`.** That would invalidate all 30+ historical whole-draw records and turn every open PR red.

## Verification

`PREFLIGHT CLEAN` — file-size cap (the workflow's own step), `cargo fmt --check`, `cargo doc`, clippy (0 lints), `cargo test --workspace --locked` (83 suites), certification self-test.

Negative controls run for each new guard: restoring the 512 fails the budget test (`left: 512, right: 1024`); removing one sibling exclusion fails the coverage test (`vision-fidelity`, `video-fidelity` re-open).


tbraun96:
/stamp

Campaign complete — `--pull-request-gate-check` at `038a50b1d` reports **12 of 12 PASS**.

This PR edits `coverage.rs` and `group.rs` (both `BOUNDARY_FILES`), so it re-opened every gate; the 11 owed records are committed in `038a50b1d`, split by the single-signer rule (six speed gates on one box, five correctness on the other). `kat-equality-gate` carries its own record from `4db613c1d`.

No speed movement (peak 116.0 vs 116.1, decode 26.3 vs 26.1) and BFCL at 84.22 / 84.12 — the same numbers main carries — which is the expected result for a change that is gate bookkeeping rather than a hot path.

tbraun96:
## Evidence for the seal

`--pull-request-gate-check` at `cd1f02c29`: **12 of 12 PASS**.

### The record set hangs together
All twelve added records carry **one** `git_sha` (`12c2875ca4`), verified with CI's own filter (`git diff --diff-filter=AM base...HEAD -- .benchmarks`), and signers split correctly:

| class | signer | gates |
|---|---|---|
| Speed | `efe157d6f6360027` (one key, as required) | decode-floor, concurrency-sweep, -dflash2, agentic-webserver, ttft-cold, ttft-warm |
| Correctness | `02156264cbf75bd7` | bfcl-subset, -echolp, ssm-poison, vision, video, kat-equality |

Both keys are in `.github/record-signers/`. Correctness may span signers; Speed may not, and does not.

⚠️ This needed a fix: the set originally spanned **two** commits, because a promotion must earn its own gate's record *before* running the campaign. `Disagreement::Commits` is fatal for every class, and `--pull-request-gate-check` does not catch it — it reported 12/12 regardless, since coverage and agreement are different questions. kat was re-run at the campaign sha and the odd record deleted.

### Checked against the enumerated hold-conditions
- **No gate failed**, so no re-run-until-green. The kat re-run was for the agreement rule, returned the same 0 divergences, and did so on a **different box** — an independent reproduction of the Correctness-class premise.
- **No bar was cut from the run that certifies it.** `diverged`/`unmeasured` are absolute invariants; `orders`/`samples` restate param pins; `empty_replies = 0` came from the *open-arm control*, a different run.
- **Boundary files** (`coverage.rs`, `group.rs`) are list edits and a doc comment, evidenced by unit tests whose negative controls were run and observed red — plus the dedicated 36→0 open/closed measurement, which is separate from this campaign. Flagging it explicitly since it is the closest condition to firing.

### Known limits, stated rather than implied
- The cap is a **prefix**, not a subsample. It gives up 738 samples: `live_simple` (25 — a **live** subset) plus the non_live half (713). No `live_simple` divergence has been observed by either instrument, but that is a measured absence, not a structural one.
- A green here certifies the **closed** regime. `bfcl-subset` pins no serve overrides and is scored **open** — the regime this gate measures at 36/257 — so this says nothing about the conditions `84.22 / 84.12` were taken under.

Sealing on the repo owner's standing delegation.

PR #983, opened 2026-09-11
gate(thresholds): raise three bars that had stopped being able to fail

tbraun96:
Raises three bars that had stopped being able to fail. **`BENCH.toml` only**, so this invalidates no record and owes no campaign — `NON_COMPILED_KERNEL_FILES` exists precisely so that ratcheting a bar does not destroy the record that justified the ratchet.

| bound | from | to | effective | basis |
|---|---|---|---|---|
| decode-floor `server_decode_tok_s` | 22.7 | **25.0** | 24.5 | n=12, mean 26.125, σ 0.062 |
| ssm-poison `min_cached_prompt_tokens` | 1.0 | **900.0** | — | 1026.0 in 54/54 passing records |
| ssm-poison `sum_wall_s` | 1200.0 | **200.0** | — | observed max 114.4 on the pinned instrument |

## decode-floor — and why the level moved

A ratchet that cannot say why is a ratchet from a hot box. This one is a **step**, not drift:

| window | `server_decode_tok_s` |
|---|---|
| 2026-08-23 → 09-07 (16 records) | 22.4 – 23.5 |
| 2026-09-08 → now | 26.09 / 26.13 / 26.19 + 12 fresh runs at 26.125 |

The window contains `f9fd8ca43` — *"mtp-carry: give the store ticket its own dispenser — sharing the capture counter cost 24% of C=2 decode"* (#968). `accept_len_mean` is **2.6826 across the entire span**, so this is not better speculation; it is that fix.

**At the old 22.2, a regression undoing #968 would have landed near 23 and passed silently.** That is the hole this closes.

The 12-run set ran on dgx2 (spark-43fa) — the *slowest* box, which is where the superseded bar was drawn from, so the bases are comparable. 24.5 sits ~26σ and 6.2% below the mean, wider than the mean−5% the sweeps use: deliberately conservative rather than the tightest the data allows (25.3).

## ssm-poison — two dead bounds

`min_cached_prompt_tokens` had a floor of **1**, and its own comment said to ratchet "after the first recorded run that carries this metric". That was 54 runs ago; every passing record reads exactly **1026.0**, zero variance. A floor of 1 separated "the restore path ran" from a cold engine and nothing else.

`sum_wall_s` had a ceiling of **1200** against an observed max of **114.4** — a blowup detector 10× too loose to detect a blowup. 200 is 1.75× the worst observed and is sized to catch what matters: one round hanging to the 300 s request timeout exceeds it alone, and a uniform 50% slowdown lands near 170.

★ Three older records read 306.8 / 320.8 / 321.4 and are deliberately **not** the basis — they predate the `disable_thinking` pin, so they measured thinking-on generation (~3× the work), and `check_record` would refuse them against this baseline anyway.

## Verification

- `--pull-request-gate-check` at this head: **12 of 12 PASS**.
- **Negative control** — the bars are live, not decorative. Pushing them past the observed values (decode 25.0→30.0, min_cached 900→2000, sum_wall 200→50) turns both gates **FAIL**; restoring turns them **PASS**.
- `check_bench_threshold_prose.py`: 88 bounds, 0 contradictions.
- No bar here was cut from a run this PR certifies; every basis predates it.

### DIFF
--- a/kernels/gb10/qwen3.6-35b-a3b/BENCH.toml
+++ b/kernels/gb10/qwen3.6-35b-a3b/BENCH.toml
@@ -449,15 +449,46 @@ max = 0.0
 # restored from anchors 992-1152). The floor of 1 is a DEFINITION, not a
 # measurement: it distinguishes "the restore path ran" from "prefix caching
 # was off and the gate replayed against a cold engine" — the configuration
-# under which this gate once returned a green PASS proving nothing. Ratchet
-# toward the observed anchor scale (~992 on this checkpoint) after the first
-# recorded run that carries this metric; do not guess a higher floor before
-# then.
+# under which this gate once returned a green PASS proving nothing.
+#
+# ★ RATCHETED 2026-09-11, 1.0 -> 900.0, discharging this comment's own
+# instruction to "ratchet toward the observed anchor scale after the first
+# recorded run that carries this metric". That condition was met 54 runs ago:
+# EVERY passing record carrying the metric reads exactly 1026.0 — 54 of 54,
+# zero variance, across four weeks and three checkpoint-config generations.
+# The floor of 900.0 sits ~12% under that invariant, which is margin for a
+# different anchor scale rather than for noise; there is no noise to cover.
+#
+# The one record that does not read 1026 is 2026-08-15-d4c00509b7, which reads
+# 0 — a FAIL whose reference round never ran (rounds 0). That is the vacuity
+# case this bound exists for, and it confirms the bound catches it. What 1.0
+# could not catch is the case in between: a server caching a handful of tokens
+# rather than a real anchor would have passed a floor of 1 while proving as
+# little as the cold engine did.
 [benchmarks.metrics.min_cached_prompt_tokens]
-min = 1.0
+min = 900.0
 
 [benchmarks.metrics.sum_wall_s]
-max = 1200.0
+# The blowup detector, not a perf gate. The ceiling of 200.0 replaces 1200.0
+# as of 2026-09-11, discharging the entry note's "generous pending the first
+# recorded run; ratchet after measurement, never to the best".
+#
+# Measured on THIS entry's pinned instrument (disable_thinking = "true",
+# ssm_cache_slots = "256"): 54 passing records spanning 75.8 to 114.4 s, in
+# three levels — ~110-114 (2026-08-15 to 08-23), ~82-85 (08-23 onward) and
+# 75.8-76.6 on 09-08, the same f9fd8ca43 step the decode floor records. 200.0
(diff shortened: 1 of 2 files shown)


tbraun96:
## Evidence for the seal

`--pull-request-gate-check` at this head: **12 of 12 PASS**, using the records already on main — this PR adds none.

### The bars are live, not decorative
The check that matters for a ratchet is whether the new number is actually enforced. Pushing all three past their observed values:

| mutation | result |
|---|---|
| decode 25.0 → 30.0 | `FAIL decode-floor` |
| min_cached 900 → 2000, sum_wall 200 → 50 | `FAIL ssm-state-poisoning-gate` |
| restored | both `PASS` |

### Hold-conditions checked
- **No records added** (`git diff --diff-filter=AM` over `.benchmarks` is empty), so `Disagreement::Commits` has nothing to judge and no signer question arises.
- **No `BOUNDARY_FILE` touched** — the diff is two `BENCH.toml` files, and no `BENCH.toml` is in `BOUNDARY_FILES`. They are in `NON_COMPILED_KERNEL_FILES` instead, which is why this owes no campaign.
- **No bar certifies itself.** This is the one condition worth stating rather than ticking. A threshold does move here — that is the whole PR — but nothing is measured *by* it: the decode basis is a 12-run set taken at `12c2875ca`, and the ssm-poison basis is 54 records spanning four weeks. Both predate this commit. That is measure-then-declare, the same shape the 2026-09-06 ratchet used (its ten-run set likewise lives in the note, not as ten committed records).

### Known limit
The 12-run calibration set is documented in the BENCH.toml note but only one of its runs is committed as a record. That matches the precedent for this entry, and the committed record (26.3) independently clears the new bar — but it does mean the set itself is not independently verifiable from the tree.

Sealing on the repo owner's standing delegation.

tbraun96:
Re-sealing after `76e279d8b`, which corrects a false causal claim in the previous commit's prose.

The **bar is unchanged at 25.0**. What changed is the explanation beside it: I had credited the decode step to #968, and the repo's own records refute that — `f9fd8ca43`'s message says "C=1 IS IDENTICAL, AND THAT IS THE DIAGNOSTIC", and the campaign measured *at* that commit reads decode 22.58 / C1 18.42, both pre-step. That record was dropped in `3cfe4d8c45`, which is the only reason the surviving records looked like a clean break. The leading candidate is now #965, recorded explicitly as a candidate rather than a finding.

Evidence for the seal is otherwise as before: no records added, no `BOUNDARY_FILE` touched, 12/12 gates PASS, and the negative control still holds (pushing the three bars past their observed values fails both gates; restoring passes them).

Sealing on the repo owner's standing delegation.

Issue #927, opened 2026-09-06
Goal: H100 Qwen3.8-27B-FP8 batched throughput at C=16 should match or beat the control engine (today 65 vs 790 tok/s)


TheTom:
## Goal

Match or beat the control engine's aggregate throughput at C=16 on a single H100 for Qwen/Qwen3.8-27B-FP8, on both frozen shapes, with the same pinned checkpoint.

## Where we are (2026-09-06, single H100, rental evidence)

| Shape | Atlas native FP8 (batch-4 profile, `5cde118`) | Control (FP8, cap 512) | Ratio |
|---|---:|---:|---:|
| 1,193 in / 256 out, C=16 | 65.0 tok/s aggregate, TTFT p50 26.3 s | 790.5 tok/s, TTFT p50 1.32 s | 12.2× |
| 4,593 in / 512 out, C=16 | 55.1 tok/s aggregate, TTFT p50 63.7 s | 638.8 tok/s, TTFT p50 3.45 s | 11.6× |

Same checkpoint weights; arithmetic differs (Atlas W8A16/W8A8 + FP8 KV, control dynamic W8A8 + BF16 KV). Not certified campaign numbers; raw evidence under `docs/campaigns/hopper-atlas-vs-vllm-2026-09/evidence/rental-h100-20260905/` on TheTom/atlas `campaign/docs-hopper-2026-09`.

## Why the gap is 12× and not 2×

It is a stack of independent limits, each with its own issue:

1. **Batch capacity.** The H100 recipe runs `--max-batch-size 4` because the default batch-32 reserve is 45.8 GiB (#915). C=16 is therefore 4 active rows plus 12 queued; the TTFT numbers above are queueing time. This alone caps aggregate at roughly 4× the per-row rate.
2. **Per-row decode on the native FP8 path.** ~17–19 tok/s per active row after the kernel-selection fix (#925), from ~3 before; the W8A16 batched GEMV is unfused and per-projection (#917).
3. **LM head.** BF16 head took the scalar GEMM for batched rows; batched-GEMV branch was present but not dispatched (#926, fix `7c786cc`).
4. **No Hopper W4A4 / FP8 tensor-core dense GEMM.** Prefill on sm_90 runs the W4A16 `t_m128` fallback (NVFP4 profile) or the W8A16 pipelined GEMM (native), not wgmma FP8 (#913 context, #917).

## Acceptance

[...]

PR #984, opened 2026-09-11
spark-model: preserve scalar accumulation order in FP8 batched GEMV and batch native o_proj (#932)

### LINKED ISSUE #932: w8a16_gemv_batch4 is not bit-identical to the scalar w8a16_gemv: pairwise accumulation gives 1-ulp BF16 drift in multi-row decode
## Summary

`kernels/gb10/common/w8a16_gemv_batch4.cu` is not bit-identical to the scalar `w8a16_gemv.cu` it replaces for multi-row decode. On the H100 with the Qwen/Qwen3.8-27B-FP8 attention output projection at 4 rows, 3 of the BF16 outputs differed from the scalar reference by one ulp (max |diff| 0.00048828125). Cause: the batched kernel sums products in pairs before accumulating, the scalar kernel accumulates one product at a time, so the FP32 rounding sequence differs.

Found by the bit-parity oracle added for the attention o_proj batching change (`6b76f3d` on `hopper/sm90-target-tdd-2026-09`); the known-bad harness cases (output-bit, row-offset guard, non-finite) all refused correctly, so this is the kernel, not the test.

## Why it matters

- The batched W8A16 GEMV is the kernel behind the native-FP8 multi-row paths that shipped today: the dense FFN batched decode (#925, `5cde118`) and the BF16 head (#930) sit on the same family. The C=16 numbers measured on the rental therefore ran with the pair-summing variant, and batched vs single-row decode of the same sequence can diverge by an ulp per projection per layer. That is the kind of drift that turns into different greedy tokens after a few hundred steps (compare #435 for speculative decode).
- Atlas's stated contract is that batching does not change results. The existing `w4a16_batch_bitparity_microtest.rs` enforces it for the NVFP4 batch kernels; the W8A16 batch kernels had no such oracle until today.

## Fix

Accumulation order in `w8a16_gemv_batch4.cu` corrected to match the scalar kernel one product at a time; the strict bit-parity test is kept. The 16-row variant (`w8a16_gemv_batch16` or equivalent) is being checked for the same issue. Commit on the campaign branch to follow; will be split into its own PR like #921–#925 and #930.

## Acceptance

- `w8a16_batch_bitparity_microtest` (RAW BF16 bytes) passes for batch4 and batch16 against the scalar kernel at Qwen3.8's real shapes (K=5120; N=12288 / 1024 / 5120) with unequal block scales and non-zero view offsets.
- Re-measure the native-FP8 C=16 ladder with the corrected kernel and report any throughput delta (the fix should be free; if it is not, that is a second finding).

TheTom:
## Summary

Two things, one file apart. `w8a16_gemv_batch4.cu` summed each pair of products *before* touching the accumulator, so the batched FP8 GEMV was not bit-identical to the scalar `w8a16_gemv` it stands in for — 1-ulp BF16 drift in multi-row decode (#932). And the native-FP8 attention output projection never used the batched kernel at all: it looped `w8a16_gemv` once per row, re-reading the whole block-scaled `o_proj` weight for every concurrent sequence (#927's bucket).

Fixes #932. Refs #927.

## What was wrong

**The kernel.** `w8a16_gemv_batchm_impl` dequantizes each FP8 weight byte once and MACs it into M independent FP32 accumulators — the whole point being that batching cannot change the answer. The inner loop did:

```c
acc[t] += __bfloat162float(lo) * wf[j * 2]
        + __bfloat162float(hi) * wf[j * 2 + 1];
```

The scalar kernel accumulates one product at a time. Summing the pair first rounds once more, in a different place, so the FP32 sequence differs and the BF16 output can land one ulp away. On the H100 at the real Qwen3.8-27B o_proj shape, 3 of the M=4 outputs differed from the scalar reference (max |diff| 0.00048828125). Atlas's contract is that batching does not change results — the NVFP4 batch kernels have had `w4a16_batch_bitparity_microtest` enforcing exactly this for a while; the W8A16 family had no such oracle.

**The dispatch.** `multi_seq/attn/o_proj.rs` had a plain `for i in 0..n` over `ops::w8a16_gemv` in the native-FP8 arm. At C=4 decode the four sequences share one weight matrix, so that is four full DRAM passes over `o_proj` where one would do.

## What the fix does

- `kernels/gb10/common/w8a16_gemv_batch4.cu`: the pair is split into two sequential `+=`, which is literally the scalar kernel's order. Five lines. The header comment's parity claim is updated from "verify with a cos>=0.9999 microtest" to exact BF16 output comparison, because that is now what is checked.
- `multi_seq/attn/o_proj.rs`: the native-FP8 arm steps by 4 and calls `ops::w8a16_gemv_batch4` when `n > 1`, the `w8a16_gemv_batch4` handle resolved, the weight is `Fp8BlockScaled`, and both `hidden_size` and `q_dim` are multiples of 128. Anything else — single-row decode, a missing kernel (older bundles), per-row scales, an unaligned dim — keeps the exact scalar loop it had. The tail group passes `(n - i).min(4)` so a row count that is not a multiple of 4 is handled by the kernel's own M argument rather than by a separate scalar pass.
- `qwen3_attention/{types,init}.rs`: the optional `w8a16_gemv_batch4_k` handle, resolved through `try_kernel`, so `KernelHandle(0)` simply retains scalar dispatch.

Nothing is copied, staged or requantized: both arms read the same FP8 weight and the same block-scale buffer, and the batched call takes the same per-row view offsets the scalar loop computed.

## Oracle and evidence

**Bit-parity, H100, 2026-09-06 02:44 UTC**, source `8882f08` (these four commits), `native_fp8_oproj_batch_microtest` at the real o_proj shape N=5120, K=6144:

- batch4 at M=2/4/5 and batch16 at M=8/16 all report `unequal_bf16=0 max_abs=0 cosine=1.0` against the scalar `w8a16_gemv`.
- The known-bad controls (output-bit mismatch, row-offset guard, non-finite) are all still refused, so the harness can fail.
- Before the fix the same test reported 3 unequal outputs at M=4.

Recorded on #932. `97375dfc` widened that oracle from batch4 to the shared batch16 template, since both instantiate `w8a16_gemv_batchm_impl` and the pair-sum was in the template, not in the M=4 wrapper.

**PTX, both targets** (see Test plan for the exact commands): `w8a16_gemv_batch4` and `w8a16_gemv_batch16` both still compile strict at `sm_121a` and `sm_90a`, 0 spill bytes, smem unchanged at 1152/1536 bytes. Register use moves slightly down, not up — sm_121a 48 -> 46 (batch4) and 72 -> 71 (batch16); sm_90a 55 -> 48 (batch4), 64 unchanged (batch16). The PTX itself of course differs: the change is a rounding-order change, and a kernel whose PTX did *not* move here would mean the fix had not landed.

**Host dispatch**, three new tests over the mock backend (`o_proj_tests.rs`): that 4 rows become one launch of the batched handle carrying the right `M`, that 2/5/16 rows chunk into `ceil(n/4)` launches with the per-group input and output offsets intact, that the projection allocates nothing, and that all four fallback conditions still reach the scalar kernel — `n == 1`, a null batched handle, a `hidden_size` that is not 128-aligned, and `Fp8PerRow` scales.

### Throughput: context, not a claim

The H100 native-FP8 ladder on the integration branch that carries these commits (`1166fc154`, 2026-09-11) measures **76.06 tok/s** aggregate at C=16 / 1024x256, against **73.8** on the 2026-09-06 reference `7c786cc` — which predates all four commits here. 4096x512 moves 61.4 -> 63.46 on the same pair. Rep-to-rep spread was 0.02-0.05%, so the +3% is outside noise.

That is 37 commits of difference including a full merge of `main`, and **it was not bisected**. Treat it as "these commits are in a tree that did not regress and measured a little better", not as a measured gain for this PR. The honest statement of what this PR is worth is the bit-parity result above; the C=4/C=16 A/B that would isolate the o_proj batching is still owed and needs GPU time (see below).

## Test plan

- [x] `cargo test -p spark-model --features cuda --lib` — **891 passed, 0 failed, 11 ignored** (902 total; 3 of them the new o_proj dispatch tests)
- [x] `cargo test -p spark-model --features cuda --lib o_proj` — **7 passed, 0 failed**
- [x] PTX, strict, both targets — `nvcc --ptx -arch=sm_121a` and `-arch=sm_90a` with `-DATLAS_NO_WARP_BLOCKSCALE_MMA --Werror all-warnings` on `kernels/gb10/common/w8a16_gemv_batch4.cu`, plus `--cubin -Xptxas -v` for the resource numbers quoted above. Both entry points emit at both arches; 0 spill stores / 0 spill loads.
- [x] Tested against real hardware — H100 bit-parity receipt above (2026-09-06)
- [ ] `scripts/hopper_ptx_gate.sh --hw gb10 --model qwen3.8-27b --jobs 8 --strict` — **the script does not exist on `main`**; it lives on the #895 campaign branch. The direct `nvcc` invocation above is the substitute and is reported rather than claimed as the gate.

## GB10 perf-gate records

This diff touches `crates/` and `kernels/`, both PERF_PATHs, so `record_covers` invalidates every committed `.benchmarks/` record the moment it lands. A GB10 re-measure is owed before merge — that is the required check talking, not a nicety, and it is also where the C=16 A/B that this PR's throughput section deliberately does not claim would come from.

## Notes for reviewers

The 128-alignment and `Fp8BlockScaled` guards are the kernel's real preconditions, not padding: `w8a16_gemv_batch4` indexes `block_scale[n/128][k/128]` and loads rows as `uint4`, so a per-row-scaled weight or an unaligned dim must keep the scalar path. The tests pin all four fallbacks rather than only the happy case.

`n > 1` rather than `n >= 4` is deliberate: the kernel handles M in 1..=4 and the tail group already relies on that, so there is no reason to hand 2 or 3 rows back to a loop that reads the weight matrix twice or three times.

The four commits are cherry-picked from `hopper/sm90-target-tdd-2026-09` (#895) with `-x` provenance lines. One extra commit on top adapts the new test to `main`: `ForwardContext` gained a `decode_step` field after the source branch was cut, and the test's struct literal now names it `false` — prefill shape, `attn_metadata` is `None`, and the output projection reads none of the decode scalars that flag guards. Same value and same rationale as the sibling `dense_ffn_native_batch_tests` fixture. No other conflict: the four picks applied clean.

The example is named `native_fp8_oproj_batch_microtest`, not the `w8a16_batch_bitparity_microtest` #932's acceptance criteria names. Same oracle and same shapes; the name is the one the source branch used and renaming it would orphan the H100 log above.

## CLA

### DIFF
--- a/crates/spark-model/Cargo.toml
+++ b/crates/spark-model/Cargo.toml
@@ -279,3 +279,7 @@ required-features = ["cuda", "gpu-examples"]
 [[example]]
 name = "w4a16_qg_batch_bitparity_microtest"
 required-features = ["cuda", "gpu-examples"]
+
+[[example]]
+name = "native_fp8_oproj_batch_microtest"
+required-features = ["cuda", "gpu-examples"]
--- a/crates/spark-model/examples/native_fp8_oproj_batch_microtest.rs
+++ b/crates/spark-model/examples/native_fp8_oproj_batch_microtest.rs
@@ -5,11 +5,12 @@ use anyhow::{Result, ensure};
 use half::bf16;
 use spark_model::layers::ops;
 use spark_runtime::cuda_backend::AtlasCudaBackend;
-use spark_runtime::gpu::{DevicePtr, GpuBackend};
+use spark_runtime::gpu::{DevicePtr, GpuBackend, KernelLaunch};
 
 const N: usize = 5120;
 const K: usize = 6144;
-const MAX_M: usize = 5;
+const MAX_M: usize = 16;
+const ORIGINAL_M: usize = 5;
 const GUARD: usize = 64;
 
 fn upload(gpu: &dyn GpuBackend, bytes: &[u8]) -> Result<DevicePtr> {
@@ -58,6 +59,7 @@ fn main() -> Result<()> {
     let gpu = AtlasCudaBackend::new(0, &atlas_kernels::ptx_modules())?;
     let scalar = gpu.kernel("w8a16_gemv", "w8a16_gemv")?;
     let batch4 = gpu.kernel("w8a16_gemv_batch4", "w8a16_gemv_batch4")?;
+    let batch16 = gpu.kernel("w8a16_gemv_batch4", "w8a16_gemv_batch16")?;
     let mut state = 0x51a7_8a16_2026_u64;
     let mut random = || {
         state = state.wrapping_mul(6364136223846793005).wrapping_add(1);
@@ -69,7 +71,7 @@ fn main() -> Result<()> {
             ((x % 127) as u8) | (((x >> 7) & 1) as u8 * 128)
         })
         .collect();
-    let acts: Vec<u8> = (0..MAX_M * K)
+    let mut acts: Vec<u8> = (0..ORIGINAL_M * K)
         .flat_map(|_| {
             bf16::from_f32(((random() % 2049) as f32 - 1024.0) / 1024.0)
                 .to_bits()
@@ -79,6 +81,13 @@ fn main() -> Result<()> {
     let scales: Vec<u8> = (0..N / 128 * (K / 128))
         .flat_map(|_| (((random() % 16 + 1) as f32) / 1024.0).to_le_bytes())
         .collect();
+    // Keep the observed M4 red fixture's first five rows and scales unchanged.
+    // Additional rows are generated only after the original scale draws.
+    acts.extend((0..(MAX_M - ORIGINAL_M) * K).flat_map(|_| {
+        bf16::from_f32(((random() % 2049) as f32 - 1024.0) / 1024.0)
+            .to_bits()
+            .to_le_bytes()
+    }));
     let weight = upload(&gpu, &weights)?;
     let input_base = upload(&gpu, &[vec![0x5a; GUARD], acts, vec![0x5a; GUARD]].concat())?;
     let input = input_base.offset(GUARD);
@@ -90,7 +99,7 @@ fn main() -> Result<()> {
     let scalar_out = scalar_base.offset(GUARD);
     let batch_out = batch_base.offset(GUARD);
     let mut first_oracle = true;
-    for m in [2_usize, 4, 5] {
+    for m in [2_usize, 4, 5, 8, 16] {
         gpu.copy_h2d(&sentinel, scalar_base)?;
         gpu.copy_h2d(&sentinel, batch_base)?;
         for row in 0..m {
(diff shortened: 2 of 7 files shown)


tbraun96:

* Of the **6 identifiers this PR introduces that `main` does not have**, **6** are present on #895.
* No gaps at all.

TheTom:

tbraun96:

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1027 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1027; this fork PR is closed.

Issue #915, opened 2026-09-06
Dense 27B on an 80 GB GPU: inference reserve is 45.8 GiB (8 rollback snapshots x max-batch-size 32) and boot is refused; reserve is not sized from free memory


TheTom:
## Summary

Serving Qwen/Qwen3.8-27B-FP8 on a single 80 GB H100 with the hopper 27B recipe (`--max-batch-size 32`, `--max-num-seqs 128`, `--max-seq-len 24576`, FP8 KV) refuses to boot: the inference reserve alone is 45.8 GiB.

```
Preflight reserve: inference=45823 MB ...
Weights: 57.2 GB (pre-KV) ... exceeded the 71.3 GB budget
```

The reserve is dominated by the SSM rollback ring: 8 rollback snapshots x max-batch-size 32 of per-sequence GDN state for a 27B dense model. It is sized from the batch flag, not from what the GPU can hold after weights.

## Workaround in place

Fork branch `hopper/sm90-target-tdd-2026-09`, commit `9d68c96b6a37661179ef4c9ecc8c4d239c810c78` "campaign: bound Qwen3.8 H100 active batch capacity": the H100 recipe sets `--max-batch-size 4` (everything else unchanged: 8-slot rollback ring, 16 prefix snapshots, ctx 24576, spec off). Reserve drops to 8.54 GiB; boot passes in 22 s with 49.4 GB actually free after weights. Note the preceding commit `ff751807` caps the control recipe's sequence count for the same reason on the other engine and is unrelated to Atlas.

That is a recipe change, not a fix. C=16 on this profile means 4 active decode rows and 12 queued.

## What a fix looks like

- Size the rollback ring from the post-weights free memory and the requested batch, and either shrink the ring depth or clamp the effective batch with a logged warning, instead of refusing to boot.
- Or make the ring depth a first-class flag with a documented memory formula so operators can trade rollback depth for batch on smaller GPUs.
- Preflight should print the formula (`snapshots x batch x per-seq state bytes`) so the operator sees why 45.8 GiB was asked for.

Related: #740 (KV budget on unified memory) and the discrete-GPU free-memory fix in #904; this is the remaining piece of "boot on an 80 GB card with the default recipe".

## Evidence

[...]

PR #985, opened 2026-09-11
spark-server: size the SSM rollback ring from free memory instead of refusing (#915)

TheTom:
## Summary

The Phase-C SSM decode-rollback ring was reserved as a **constant** depth (8) times `--max-batch-size`, a flag that says nothing about what the card can hold. On one 80 GB H100 with the Qwen3.8-27B hopper recipe that is 8 x 32 x 151.5 MiB = 37.88 GiB of ring inside a 45,823 MiB inference reserve, against a 71.3 GiB budget already carrying the weights — and the serve refused to boot. The shipped workaround was `--max-batch-size 4`: paying for rollback depth with four fifths of the concurrency.

This makes the depth a real flag with a printed formula, and lets preflight shrink it instead of refusing.

Refs #915. **Deliberately not `Fixes`** — see "What this does not fix", which is the most important section here.

## What changed

- **`--ssm-decode-ring-slots <auto|N>`** (N in 0..=8, default `auto`) pins the depth on *both* sides — the preflight reserve and `TransformerModel::new`'s allocation — and disables the fit. A serve that does not fit at the pinned N is **refused rather than booted at a depth its recipe does not record**. `ATLAS_SSM_DECODE_RING=1|0` keeps meaning 8 and 0, and is consulted only when the flag is `auto`.
- **Preflight auto-fit**: on `auto`, walks the depth down 8 -> 4 -> 2 -> 1 -> 0 until the reserve fits, publishes the chosen depth so the allocation matches the reservation exactly, and logs the formula at WARN.
- **The refusal carries the formula too**, so an operator reading only the failure can see where the reserve went instead of inferring it.
- **Depth-0 correctness fix**, which the auto-fit is what makes reachable: an SSM model with the ring disabled fell into `rollback_to_boundary`'s pure-attention any-boundary arm, rewinding tokens while the recurrent state stayed conditioned on the discarded tail. It now declines with `NoSsmSnapshot` — the fail-open the ring's SSOT already documented. Without this, turning the ring off would have been a silent correctness bug rather than a capacity trade.
- **Docs**: `SsmDecodeRing`'s module doc and `Model::decode_rollback_ring_slots` both still quoted `ROLLBACK_RESTEER_CAP + 1` as the depth. That was already wrong when the depth became 8; it is doubly wrong now that the depth is whatever the flag or the fit published. Doc-only commit.
- **File splits**, mechanical: `preflight.rs` (515 LoC, over the cap) into `preflight/{decode_ring,refusal}.rs`, and `ssm_reserve.rs` (608) into `ssm_reserve/decode_ring.rs`. Every public path is unchanged by re-export.

## What this does not fix

A batch-16 and batch-32 boot were attempted on an 80 GB H100 on **2026-09-11** against a tree carrying these commits. **Both still refused**, and the probe explains why. This is a real negative result and it belongs in the body, not in a follow-up issue nobody reads.

The fitter runs at preflight, ~1 s into boot, and sizes the reserve against **pre-load free memory**. Its own WARN says so:

```
SSM decode-rollback ring auto-fit — ring: 4 slots x 64 seqs x 151.5 MB/seq = 37.88 GB
(was 75.75 GB); reserve 54.13 of 78.57 GB. Sized from free memory, not from
--max-batch-size 64 (#915): rollback depth yields, concurrency does not.
Pass --ssm-decode-ring-slots N to pin a depth (and be refused rather than shrunk).
```

`reserve 54.13 of 78.57 GB` — 78.57 GB is free memory *before* the weights load. The quantity that actually has to accommodate the reserve is the **post-load KV headroom**, which the same boot measures four minutes later as `71.3 GB budget − 58.1 GB already live = 13.2 GB`. The fitter is sizing against a number ~6x larger than the real constraint. Direct sweep of preflight (each point ~20 s, since preflight logs long before the load finishes):

| `--max-batch-size` | depth chosen | ring GB | inference reserve | fired? |
|---|---|---|---|---|
| 4 | 8 | 4.73 | 8,745 MB | no |
| 16 | 8 | 18.94 | 25,107 MB | no |
| 32 | 8 | 37.88 | 46,923 MB | no |
| 40 | 8 | 47.34 | 57,831 MB | no |
| **64** | **8 -> 4** | 37.88 | 51,771 MB | **yes** |
| **128** | **8 -> 2** | 37.88 | 61,467 MB | **yes** |

So the mechanism works and the plumbing is correct — the depth it picks is honoured end to end — but it stays silent for every batch size from 4 to 40, and even when it fires at 64/128 it targets ~37.88 GB of ring against the 13.2 GB that exists, so the shrunk configuration is still unbootable. **At this tree the auto-fit cannot produce a bootable configuration at any `--max-batch-size` above ~4 on an 80 GB H100.**

Against #915's three asks, honestly scored:

| #915 asks | state |
|---|---|
| Make ring depth a first-class flag with a documented memory formula | **done**, verified on hardware |
| Preflight prints the formula so the operator sees why N GiB was asked for | **done**, verified on hardware |
| Size the ring from **post-weights** free memory, shrink instead of refusing | **partial** — shrinks, but against the pre-load number, which is the wrong yardstick |

Fixing the yardstick means either moving the fit to after the weight load (where the 58.1 GB live figure exists) or teaching preflight to predict it — the gap is the load's transient/duplicate allocations, ~29 GB on this boot, not the resident weights. That is a separate change with its own hardware bill, and stacking it under a review of this one would hide both.

## What it did buy, measured

The flag half is not theoretical: it is what made any batch-16 or batch-32 H100 number exist at all on 2026-09-11. With `--ssm-decode-ring-slots 1` at `--max-batch-size 16`, the same recipe that refused now boots in 253 s and reports the pin taking, with **no** ring WARN — the documented behaviour for an explicit depth:

```
kernel flags: ... ssm_rollback_mode=Snapshot ssm_decode_ring_slots=1 ...
Preflight reserve: inference=8139 MB, buffer_arena=3658 MB (pre-load free: 78.6 GB);
  ring: 1 slots x 16 seqs x 151.5 MB/seq = 2.37 GB
KV cache: 79.2 GB total x 90% util = 71.3 GB budget; 58.1 GB pre-KV + 7.9 GB
  reserve -> 5.2 GB for KV -> 8585 blocks x 16 tok/block = 137360 max KV tokens
```

That serve then ran a clean ladder (`LADDER_EXIT=0`, zero errors on 12 recorded reps) and a 16-way concurrent arithmetic probe at **16/16 correct**. Before this PR the only way to reach that state was to cut the batch by 4x.

## Test plan

- [x] `cargo test -p spark-model --features cuda --lib ssm_reserve` — **25 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda --lib` — **893 passed, 0 failed, 11 ignored**
- [x] `cargo test -p spark-server --bins --features cuda,nccl preflight` — **26 passed, 0 failed**
- [x] `cargo test -p spark-server --bins --features cuda,nccl rollback` — **26 passed, 0 failed**
- [x] `cargo test -p spark-server --bins --features cuda,nccl ssm_decode_ring` — **14 passed, 0 failed**
- [x] `cargo test -p spark-server --bins --features cuda,nccl flag` — **48 passed, 0 failed**
- [x] Tested against real hardware — 80 GB H100, 2026-09-11: the negative result and the pinned-depth boot above
- [ ] PTX gate — not applicable, no `kernels/` file is touched

## GB10 perf-gate records

## Notes for reviewers

The depth-0 rollback fix is the part to read closest. It is a correctness change riding along with a capacity change, and it is only in scope because the auto-fit is what makes depth 0 reachable in the first place — before this PR nothing could put an SSM model in that state. If you would rather see it split, say so and it can lead its own PR; the rest of this stack does not depend on it.

`--ssm-decode-ring-slots N` refusing rather than shrinking is deliberate and is the asymmetry worth arguing about: `auto` may silently give you a shallower ring, an explicit N may not. The reasoning is that a recipe that records a depth should either get that depth or fail loudly, since a serve quietly running at a different rollback depth than its recipe claims is unreproducible.

Both commits are cherry-picked from `fix/915-ring-autofit` on the fork with `-x` provenance lines. Both applied to current `main` with **no conflicts**; nothing was adapted.

## CLA

### DIFF
--- a/crates/spark-model/src/model/impl_a1.rs
+++ b/crates/spark-model/src/model/impl_a1.rs
@@ -275,8 +275,9 @@ impl TransformerModel {
         // therefore unreachable, and on this model it is NOT cheap: 8 slots x
         // max_batch x the full SSM blob (27B: 158.9 MB) = ~19.9 GB at batch 16,
         // allocated up front. Skip it when speculative decode is on.
-        // The ring-depth decision (env overrides + speculative/watchdog
-        // skip) is SSOT'd in `crate::ssm_reserve::decode_rollback_ring_slots`
+        // The ring-depth decision (the published `--ssm-decode-ring-slots` /
+        // #915 auto-fit depth, env overrides, speculative/watchdog skip) is
+        // SSOT'd in `crate::ssm_reserve::decode_rollback_ring_slots`
         // — spark-server's `preflight_reserve` calls the SAME helper, so the
         // GPU reservation and this allocation cannot drift. The scheduler
         // keys off `decode_rollback_ring_slots()`, so a 0 here disables save
--- a/crates/spark-model/src/ssm_reserve.rs
+++ b/crates/spark-model/src/ssm_reserve.rs
@@ -1,48 +1,32 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 
-//! SSOT for the Phase-C decode-rollback ring depth.
+//! SSOT for the SSM/linear-attention GPU reserve terms.
 //!
-//! Two call sites MUST agree on this number or a serve either
-//! under-reserves (runtime CUDA alloc failure after weights load) or
-//! over-reserves (preflight refuses batch sizes the runtime could fund):
+//! Every term here is computed TWICE — once by `spark-server`'s
+//! `preflight_reserve` before the weights load, once by the allocating call
+//! site (`SsmStatePool::new`, `TransformerModel::new`) after — and the two
+//! MUST agree or a serve either under-reserves (runtime CUDA alloc failure)
+//! or over-reserves (preflight refuses a configuration the runtime could
+//! fund). Each function below is the one place that decision is made.
 //!
-//! * `spark-server` `preflight_reserve` — sizes the SSM-snapshot GPU
-//!   reservation before weights load;
-//! * `TransformerModel::new` (`impl_a1.rs`) — allocates the actual ring.
-//!
-//! The ring's ONLY writer (scheduler `snapshot_boundary_if_ssm`) and reader
-//! (content-loop `rollback_to_boundary`) live on the PLAIN decode path — the
-//! speculative path does its rejection rollback through the verify snapshot,
-//! never this ring. Under `--speculative` the ring is unreachable, and it is
-//! NOT cheap: 8 slots × max_batch × the full SSM blob (27B: 158.9 MB) is
-//! ~19 GB at batch 16 and ~38 GB at batch 32. Reserving it unconditionally
-//! while the runtime skipped it capped the native batch at ~20 on GB10
-//! (SSM reserve 75.2 GB vs an 85.2 GB budget at util 0.70).
-//!
-//! Env contract (read HERE and nowhere else):
-//!
-//! * `ATLAS_SSM_DECODE_RING=1` force-allocates the ring even under spec
-//!   (mixed workloads whose grammar-bound sequences fall to plain decode and
-//!   should keep loop re-steer); `=0` force-disables it even without spec.
-//! * `ATLAS_DISABLE_WATCHDOGS=1|true` (trimmed, case-insensitive — mirrors
-//!   spark-server's `parse_disable_watchdogs`): the ring's only reader can
-//!   never fire, so the ring is skipped.
+//! The Phase-C decode-rollback ring DEPTH — its publication cell, the
+//! `ATLAS_SSM_DECODE_RING` / `ATLAS_DISABLE_WATCHDOGS` contract and the #915
+//! auto-fit — lives in the `decode_ring` sibling module and is re-exported
+//! here, so every existing `ssm_reserve::decode_rollback_ring_slots` path is
+//! unchanged.
 
-/// Outcome of the ring-depth decision.
-///
-/// `skip_reason` is `Some` only for the IMPLICIT skip (speculative decode /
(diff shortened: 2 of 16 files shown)


TheTom:
H100 follow-up (2026-09-11, round 6, tip 456109bf3 on the campaign branch): once the native-FP8 residency fix landed (pre-KV 58.1 → 33.2 GB), `--ssm-decode-ring-slots auto` boots batch 16 with the full 8-slot ring (`ring: 8 slots x 16 seqs x 151.5 MB/seq = 18.94 GB`, KV 13.6 GB / 355k tokens) and the C=16 decode step halved (TPOT 100.7 → 55.6 ms on 1193/256; aggregate 135.8 → 228.0 tok/s). Batch 32 under `auto` still refuses (`33.2 + 45.8 = 79.0 GB` vs the 71.3 GB budget) because this fitter measures against pre-load free memory; a second pass that sizes the ring against the predicted post-load KV headroom (weights + derived + arena + a KV floor) is in progress on the campaign branch and will be a follow-up PR. With an explicit `--ssm-decode-ring-slots 4`, batch 32 boots and matches batch 16 at C=16 to within 0.1%.

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1035 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1035; this fork PR is closed.

PR #986, opened 2026-09-11
spark-model: batch native FP8 Q/K/V projections across decode rows (#927)

TheTom:
> **Stacked on #984.** Its four commits are the first four here; this PR's own content is the single commit `spark-model: batch native FP8 Q/K/V projections across decode rows (O13)`. Review that one — or `git diff pr/fp8-batch-parity-oproj..` — and merge #984 first. Base is `main` rather than #984's branch so the PR does not become unreviewable if #984 is rebased.

## Summary

Multi-row decode dispatched the native-FP8 (W8A16 block-scaled) Q/K/V projections as **three scalar `w8a16_gemv` launches per row**. The batched FP8 GEMV kernels write a contiguous `[M, N]` output, while the multi-seq QKV buffer is strided by `per_seq_qkv` — 14,336 BF16 elements on Qwen3.8-27B, with Q at offset 0, K after Q and V after K inside every row. The batched kernel could not address that layout, so every concurrent sequence re-read the whole q/k/v weight. In the C=4 decode profile that bucket (qkvz / out_proj) is ~6.3 ms of a 44.7 ms step, the third largest.

Refs #927.

## What was wrong

Not a missing kernel — a missing stride. `w8a16_gemv_batchm_impl` assumed A and C were packed `[M, K]` and `[M, N]`. The decode QKV buffer is neither: it is one row per sequence, `per_seq_qkv` elements apart, with the three projections living at fixed offsets *inside* each row. Batching across rows therefore had nowhere to write, and the dispatch fell back to `n x 3` scalar launches over weights that all `n` rows share.

## What the fix does

**Kernel** (`kernels/gb10/common/w8a16_gemv_batch4.cu`). `w8a16_gemv_batchm_impl` gains `a_row_stride` and `c_row_stride`, in elements, and two new entry points expose them: `w8a16_gemv_batch4_strided` and `w8a16_gemv_batch16_strided`. The existing contiguous entry points pass K and N for those arguments, which is literally what the body computed before — so they are unchanged by construction, and that is verified against PTX rather than asserted (below).

**Dispatch.** A new `ms_qkv_batchm_fp8` tier in its own module (`multi_seq/qkv_fp8_batch.rs`, kept out of the already-oversized `qkv.rs`) takes `n` in 2..=8 when all of q/k/v are block-scaled FP8, the dims are multiples of 128, both strided handles resolved, and `ATLAS_NO_FP8_QKV_BATCH` is unset. One launch per projection, straight into the per-seq offsets the scalar loop already used, then the same gated-Q deinterleave the per-sequence FP8 branch applies. LoRA and the q/k norms stay where `ms_phase_qkv` runs them for every branch — this tier does not move them.

The tier sits after the NVFP4 tiers and before the dense-BF16 one; they are mutually exclusive by weight type, so that ordering is readability, not precedence. It branches on the **padded** ctx `n` — the value the CUDA-graph cache is keyed by — never on the unpadded sequence count, so a captured graph cannot be replayed under a different dispatch than it was captured with.

## Oracle and evidence

**The entry points that already existed are byte-for-byte unaffected.** `w8a16_gemv_batch4` and `w8a16_gemv_batch16` were compiled from #984's tree and from this one, at `sm_90a` and `sm_121a`, and their entry bodies compared:

```
sm_90a   w8a16_gemv_batch4 : instruction-identical after symbol-name normalisation (606 lines)
sm_90a   w8a16_gemv_batch16: instruction-identical after symbol-name normalisation (1906 lines)
sm_121a  w8a16_gemv_batch4 : instruction-identical after symbol-name normalisation (509 lines)
sm_121a  w8a16_gemv_batch16: instruction-identical after symbol-name normalisation (1535 lines)
```

The only raw difference is the template's mangled shared-symbol name (`...PS0_jjjE` -> `...PS0_jjjjjE`, the two new parameters) and the `// demoted variable` shared declarations moving to module scope now that four entries instantiate the template instead of two. Not one instruction moves. Resource usage is identical across all four entry points on both targets, with no spills:

| target | batch4 / batch4_strided | batch16 / batch16_strided |
|---|---|---|
| sm_90a | 48 regs, 1152 B smem | 64 regs, 1536 B smem |
| sm_121a | 46 regs, 1152 B smem | 71 regs, 1536 B smem |

**Bit-exactness against the scalar path, H100, 2026-09-11 05:19 UTC.** `native_fp8_qkv_batch_microtest` at the real Qwen3.8-27B shapes (N=12288 / K=5120 / stride=14336 and the k/v shapes), each row compared against production scalar `w8a16_gemv`:

```
KNOWN_BAD output-bit: refused: batch output differs from production scalar BF16 bits
KNOWN_BAD gap:        refused: batched projection wrote outside its extent at byte 24642
KNOWN_BAD guard:      refused: batched projection wrote outside its extent at byte 0
KNOWN_BAD nonfinite:  refused: nonfinite projection output

q_proj   M=8 N=12288 K=5120 stride=14336 kernel=batch16 unequal_bf16=0 max_abs=0.000000000
full-qkv M=8 row_elems=14336 K=5120                     unequal_bf16=0 max_abs=0.000000000
ALL PASS: Qwen3.8-27B q/k/v shapes, strided batch4 M2/M3/M4 and batch16 M5/M8,
          exact scalar equivalence, gaps and guards intact
O13_EXIT=0
```

That is **20/20 positive cases** — M in {2,3,4} on `batch4`, M in {5,8} on `batch16`, each crossed with {q_proj, k_proj, v_proj, full-qkv} — every one `unequal_bf16=0`, with four negative controls refused first, so the harness demonstrably can fail. Each single-projection case leaves the other two slots as gaps and requires them, the rows past M, and the guard bands to stay sentinel: the whole risk of a strided write is writing outside the extent, so that is what is asserted.

The serving-level counterpart from the same session: a 16-way concurrent arithmetic probe, all 16 sharing one decode step, **16/16 correct** in 4.05 s wall with exact string matching.

Note the microtest only means anything with `ATLAS_TARGET_HW=hopper ATLAS_TARGET_MODEL=qwen3.8-27b ATLAS_TARGET_QUANT=nvfp4` exported, and nothing in `env.sh` supplies them. Worth fixing separately.

**Host dispatch** tests (`qkv_fp8_batch_tests.rs`) cover n=1, n in 2..=8, missing handles, per-row scales, unaligned dims and the `ATLAS_NO_FP8_QKV_BATCH` kill switch.

**No throughput number is claimed here.** The ~6.3 ms of 44.7 ms figure is the *pre-change* profile that motivated the work; the A/B that would say what this bought is owed on GB10 and is not something this PR asserts.

## Test plan

- [x] `cargo test -p spark-model --features cuda --lib qkv` — **18 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda --lib o_proj` — **7 passed, 0 failed** (#984's tests, unmoved)
- [x] `cargo test -p spark-model --features cuda --lib` — **898 passed, 0 failed, 11 ignored** (891 on #984, +7 new dispatch tests)
- [x] PTX, strict, both targets — `nvcc --ptx -arch=sm_121a` and `-arch=sm_90a` with `-DATLAS_NO_WARP_BLOCKSCALE_MMA --Werror all-warnings`; all four entry points emit, plus the entry-body identity comparison and `--cubin -Xptxas -v` resource numbers above
- [x] Tested against real hardware — H100 microtest receipt above (2026-09-11)
- [ ] `scripts/hopper_ptx_gate.sh --hw gb10 --model qwen3.8-27b --jobs 8 --strict` — **the script does not exist on `main`**; it lives on the #895 campaign branch. The direct `nvcc` work above is the substitute and is reported as such, not claimed as that gate.

## GB10 perf-gate records

This diff touches `crates/` and `kernels/`, both PERF_PATHs, so `record_covers` invalidates every committed `.benchmarks/` record on landing. A GB10 re-measure is owed before merge — and it is also where the decode A/B this PR declines to claim would come from.

## Notes for reviewers

The reason the pre-existing entry points get a PTX identity check rather than a "it is obviously the same" argument: the refactor threads two new parameters through a template that four entry points now instantiate, and the compiler is free to re-register-allocate the lot. "Obviously the same" is exactly the claim that is cheap to make and cheap to check, so it is checked.

`n` in 2..=8 rather than 2..=16 is the conservative bound: `batch16_strided` is wired and covered by the oracle at M=5 and M=8, but no decode profile on the target hardware reaches the padded ctx above 8, so widening it would ship an untested band. Raising the ceiling is a one-line change once a profile needs it.

The kernel's accumulation order is the one #932 fixed and is not touched here — which is why this PR is stacked on #984 rather than racing it: both edit `w8a16_gemv_batch4.cu`, and landing them in the other order would put the strided template on top of the pair-summing accumulator.

Cherry-picked from `fix/o13-fp8-qkv-batch` on the fork with an `-x` provenance line. Applied to #984's branch with **no conflicts**; nothing was adapted.

## CLA

### DIFF
--- a/crates/spark-model/Cargo.toml
+++ b/crates/spark-model/Cargo.toml
@@ -283,3 +283,7 @@ required-features = ["cuda", "gpu-examples"]
 [[example]]
 name = "native_fp8_oproj_batch_microtest"
 required-features = ["cuda", "gpu-examples"]
+
+[[example]]
+name = "native_fp8_qkv_batch_microtest"
+required-features = ["cuda", "gpu-examples"]
--- a/crates/spark-model/examples/native_fp8_qkv_batch_microtest.rs
+++ b/crates/spark-model/examples/native_fp8_qkv_batch_microtest.rs
@@ -1,0 +1,351 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+//! Actual Qwen3.8-27B Q/K/V projection shapes: the production per-sequence
+//! scalar `w8a16_gemv` loop versus ONE strided batched launch per projection
+//! (`w8a16_gemv_batch{4,16}_strided`, issue #927 / O13).
+//!
+//! No new kernel math is claimed, so the bar is exact: identical BF16 output
+//! bytes, and every byte the batched path must NOT touch still holding its
+//! sentinel (the rows past M, the unused K/V region of a Q-only run, and the
+//! guard bands either side of every buffer).
+//!
+//! Shapes come from `kernels/gb10/qwen3.8-27b/MODEL.toml`: hidden_dim 5120,
+//! head_dim 256, q_heads 24, kv_heads 4, attn_output_gate = true. So
+//! K = 5120, q_dim = 6144, q_proj_dim = 2*q_dim = 12288 (interleaved [Q|gate]),
+//! kv_dim = 1024, and one sequence's [Q|K|V] block is 14336 BF16 elements —
+//! the `per_seq_qkv` row stride the batched kernels have to honour.
+//!
+//! Run (H100):
+//!   cargo run --release -p spark-model --features cuda,gpu-examples \
+//!     --example native_fp8_qkv_batch_microtest
+use anyhow::{Result, ensure};
+use half::bf16;
+use spark_model::layers::ops;
+use spark_runtime::cuda_backend::AtlasCudaBackend;
+use spark_runtime::gpu::{DevicePtr, GpuBackend, KernelHandle};
+
+const K: usize = 5120; // hidden_dim
+const Q_PROJ_DIM: usize = 12288; // 2 * q_heads * head_dim (gated)
+const KV_DIM: usize = 1024; // kv_heads * head_dim
+const PER_SEQ_QKV: usize = Q_PROJ_DIM + 2 * KV_DIM; // 14336 BF16 elements
+const MAX_M: usize = 8;
+const GUARD: usize = 64; // bytes of sentinel either side of every buffer
+const SENTINEL: u8 = 0x5a;
+
+/// One projection's slot inside a sequence's [Q|K|V] block.
+struct Proj {
+    name: &'static str,
+    weight: DevicePtr,
+    scale: DevicePtr,
+    /// Element offset of this projection inside the row.
+    offset: usize,
+    /// Output width (N).
+    n: usize,
+}
+
+fn upload(gpu: &dyn GpuBackend, bytes: &[u8]) -> Result<DevicePtr> {
+    let ptr = gpu.alloc(bytes.len())?;
+    gpu.copy_h2d(bytes, ptr)?;
+    Ok(ptr)
+}
+
+fn values(bytes: &[u8]) -> Vec<f64> {
+    bytes
+        .chunks_exact(2)
+        .map(|x| bf16::from_bits(u16::from_le_bytes([x[0], x[1]])).to_f32() as f64)
+        .collect()
+}
+
+/// The real comparison oracle, exercised against deliberately-corrupted input
+/// below so a green run cannot be a vacuous one.
+fn check(observed: &[u8], baseline: &[u8], sentinel: &[u8], live: &[(usize, usize)]) -> Result<()> {
+    ensure!(
+        observed.len() == sentinel.len() && baseline.len() == sentinel.len(),
+        "output extent mismatch"
+    );
+    // Every byte outside the live extents must still be sentinel, on BOTH runs.
+    let mut mask = vec![false; sentinel.len()];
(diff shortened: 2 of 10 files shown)


tbraun96:

* Of the **29 identifiers this PR introduces that `main` does not have**, **28** are present on #895.
* The single name not found is a TEST name. Coverage is not lost: `o_proj_tests.rs` carries **3** tests on this PR and **14** on #895 (0 on `main`).

TheTom:

tbraun96:

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1028 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1028; this fork PR is closed.

Issue #929, opened 2026-09-06
Coherency gate: the word-reversal known-answer probe is failed by every engine and model tried; it gates on the model, not the engine


TheTom:
## Summary

The coherency gate's known-answer probe "spell refrigerator backwards" (`bench/hopper_ab/coherency_gate.py`) was failed by every engine and every model tried on the H100: Atlas and the control on Qwen3.6-35B-A3B-FP8 (both `rotarefiger`), the control on Qwen3.8-27B-FP8 (runs out of budget mid-explanation), Butter on Qwen3.8-27B Q4 (`rotrateregirifer`), Atlas native FP8 on Qwen3.8-27B (same as control). Only Atlas's NVFP4 profile on Qwen3.8-27B produced `rotaregirfer` — and that flipped to a miss under native-FP8 arithmetic on the same weights. Every other probe (determinism, tool call, think-leak, 17×23, Tokyo) passed on every leg.

Because a failed gate blocks the scored ladder by design, no scored cell could be produced all day; the matched measurements only exist because the operator granted an explicit "reversal-only" exception, recorded per cell in `quality-policy-*.json` with `certification_claimed: false`.

## Why this is a harness bug

Character-level string reversal is a known weak spot for BPE models at temperature 0 and is sensitive to arithmetic (the NVFP4 vs FP8 flip above). A probe that the reference implementation fails on the reference model measures the model, not the engine. The gate's purpose is to catch engine defects (garbage, looping, wrong tool parsing, leaked thinking), so the known-answer set should be things the model reliably gets right and an engine bug would break.

## Proposal

- Replace the reversal probe with 2–3 probes calibrated on the control engine at temperature 0 for each pinned checkpoint (e.g. a second arithmetic item, a short factual item with a single-token answer, a JSON-shape item), and store the expected answers per checkpoint in the recipe, not hard-coded.
- Keep reversal, if wanted, as a *recorded* quality observation that does not gate.
- Add a calibration step to the campaign runbook: run the known-answer set on the control first; any probe the control misses is removed from the gate for that checkpoint and logged.
- Make the exception path first-class: a documented, per-probe allow-list in the recipe with the authorization recorded in the artifact, instead of an ad-hoc wrapper.

## Evidence

[...]

Issue #928, opened 2026-09-06
Goal: H100 Qwen3.8-27B-FP8 single-request latency should match or beat the control engine (today TTFT 0.93-3.2 s vs 0.29-0.39 s, TPOT 16-24 ms vs 12 ms)


TheTom:
## Goal

Match or beat the control engine's single-request latency on a single H100 for Qwen/Qwen3.8-27B-FP8: TTFT and per-token decode, both frozen shapes, same pinned checkpoint.

## Where we are (2026-09-06, single H100, C=1, temperature 0)

| Shape | Metric | Atlas NVFP4 profile (default) | Atlas native FP8 (`5cde118`) | Control (FP8) |
|---|---|---:|---:|---:|
| 1,193 / 256 | TTFT p50 | 931 ms | 1,074 ms | 287 ms |
| 1,193 / 256 | TPOT | 15.9 ms | 21.3 ms | 11.9 ms |
| 1,193 / 256 | decode tok/s | 51.2 | 39.3 | 76.9 |
| 4,593 / 512 | TTFT p50 | not run | 3,195 ms | 385 ms |
| 4,593 / 512 | TPOT | not run | 24.0 ms | 12.0 ms |
| 4,593 / 512 | decode tok/s | not run | 33.1 | 78.5 |

Boot is the one place Atlas leads: 21–24 s vs 89–105 s warm for the control.

Raw evidence: `docs/campaigns/hopper-atlas-vs-vllm-2026-09/evidence/rental-h100-20260905/` on TheTom/atlas `campaign/docs-hopper-2026-09` (cells `qwen38.atlas.c.lat.c1`, `benchmark.qwen38.atlas.native-lat02`, `benchmark.qwen38.atlas.native-agent01`, `benchmark.qwen38.vllm.lat01`, `benchmark.qwen38.vllm.agent01`).

## Causes on record

- **Prefill (3–8× behind).** Hopper has no NVFP4 MMQ path (#913 removed the trap; there is no sm_90 kernel behind it) and the native-FP8 dense GEMM is W8A16 BF16-MMA, not FP8 tensor-core (#917). The control runs a block-scaled FP8 dynamic-quant GEMM. The pipelined GEMM swap (#925) already took native TTFT from 3.41 s to 1.07 s on the short shape; the rest needs a Hopper FP8 W8A8 dense GEMM with dynamic activation quant, the same approach Atlas's attention/GDN prefill already uses.
[...]

Issue #917, opened 2026-09-06
Native FP8 (W8A16) dense path on H100 is 1.8x slower decode and 3.7x slower TTFT than the NVFP4 requant path; batched decode collapses to ~3 tok/s per row


TheTom:
## Summary

With the native FP8 profile actually native (`ATLAS_DENSE_FP8=1`, `--lm-head-dtype bf16`, multi-row dispatch fix `b3025ab1`), Qwen/Qwen3.8-27B-FP8 on one H100 at C=1, 1,193 prompt tokens / 256 output, temperature 0:

| Profile | Decode tok/s | TTFT p50 | TPOT | E2E |
|---|---|---|---|---|
| Default (FP8 checkpoint requantized to NVFP4, W4A16 fallback prefill) | 51.2 | 931 ms | 15.9 ms | 5.00 s |
| Native FP8 (W8A16 + BF16 head) | 28.9 | 3,409 ms | 21.3 ms | 8.85 s |

Three repetitions each, spread under 0.4 %. Boot and memory are the same (28.75 GB weights, ~21 s). The loader comment in `weight_loader/qwen35_dense.rs` already predicted this ("NOT a perf win ... FP8 W8A16 kernels are unfused per-projection GEMV, non-transposed `w8a16_gemm` prefill").

At C=16 (4 active decode rows, 12 queued on this profile) the native FP8 batch path collapsed to roughly **3 decode tokens/s per active request** (`Done: 256 tokens (length) 3.0 tok/s, TTFT=3395.6ms` in the serve log). The rung was stopped by the operator to save rental budget; partial C=16 logs are retained. The same recipe on the NVFP4 profile is expected to be tens of tok/s per row, so the multi-row W8A16 GEMV path (`w8a16_gemv_batch4`) is the primary suspect, ahead of the prefill GEMM.

## Why it matters

On Hopper and sm_100 there is no NVFP4 W4A4 MMQ path (see #913), so the NVFP4 profile is itself running fallback prefill. The FP8 path is the one that should be fast on these GPUs, and it is the one that preserves the checkpoint's weights bit-for-bit. Right now the choice is "fast but requantized" or "native but 2x slower".

## Leads

- Dense FFN FP8 prefill uses the basic `w8a16_gemm`; another layer already uses `w8a16_gemm_pipelined` with the same arithmetic. A narrow swap is being evaluated.
[...]

PR #987, opened 2026-09-11
spark-model: W8A8 block-scaled dense FFN prefill (#917)

TheTom:
## Summary

On a native-FP8 checkpoint the dense FFN's prefill GEMMs ran **W8A16**: BF16 activations against E4M3 weights, so the MMA is the BF16 tensor-core path and the FP8 bytes buy memory, not math. The attention Q/K/V/O projections and the MoE shared expert had already moved to the W8A8 block-scaled path; on a dense model the FFN is most of the prefill FLOPs and it was the one large consumer left behind.

This gives it the same arithmetic vLLM uses — per-token 1×128 FP32 activation scales folded against the checkpoint's 128×128 FP32 weight scales in an FP32 epilogue, accumulated by `mma.sync.m16n8k32.e4m3` (native on sm_90a and sm_121).

Refs #917 and #928.

**Scope note, up front.** The cuBLASLt arm (`ATLAS_CUBLAS_GEMM=1`) is **not validated on hardware and should not be enabled**. It is wired and it is very fast — 1,140 TFLOP/s at gate/up M=1193 on an H100 — but the 2026-09-11 microtest found it numerically **wrong**: rel_rms 7.7e-2–8.8e-2 against the W8A16 reference at M=1193 where the in-tree kernel manages 2.5e-2, from a VEC128 activation-scale layout mismatch. The fix for that lives on a follow-up branch and gets its own PR behind its own H100 receipt. **Every number below is the default kernel path**, which is what the serve selects unless `ATLAS_CUBLAS_GEMM` is set.

## What was wrong

`w8a16_gemm_pipelined` on the dense FFN shapes. Measured on H100 2026-09-11 at the real Qwen3.8-27B dims (H=5120, INTER=17408; gate/up N=17408 K=5120, down N=5120 K=17408), it turns **15–73 TFLOP/s** — the low end at the short prefill chunk M=64, the high end at M=1193. The FP8 weight bytes were being dequantized into a BF16 MMA rather than feeding the FP8 tensor cores the checkpoint's format already permits.

## What the fix does

- **`dense_ffn_w8a8_prefill.rs`, a new child module** of `dense_ffn.rs` (which is already at the CI size cap), holding the whole selection rule, the activation quantizer call, the GEMM call and the log-once route latch. `w8a8_prefill_selected` is a pure function of shape, scale format and kernel handles so the CPU tests can pin every clause: `m > 4` (M≤4 stays on the batch4 GEMV, which streams each weight once and beats any MMA tile there), `ATLAS_FP8_SINGLE_SCALE` not set, `Fp8BlockScaled` scales (a per-row layout would be read as `[N/128, K/128]`), `n` and `k` multiples of 128, and both kernel handles resolved — a model shadow without them keeps the W8A16 branches.
- **`dense_ffn.rs`**: the `w8_gemm!` dispatch macro gains one arm, placed **after** the `w8a16_gemv_batch4` decode arm and **ahead** of the W8A16 tile arms. Gate and up share a single activation quantization (they read the same `[M, K]` input; quantizing per projection would pay the per-token E4M3 cast twice per layer for identical bytes); `down` quantizes the post-SiLU product into the same scratch, which is safe because every launch is on one stream.
- **`ATLAS_FFN_W8A16_ONLY`** (presence, not `=1` — `…=0` meaning "on" is a trap for an operator reaching for an escape hatch mid-incident) restores the old path. The selected arithmetic is logged **once at INFO**, and the unchanged W8A16 path carries a counterpart line, so the absence of the W8A8 line is never ambiguous between "kill switch set" and "log lost".
- **`dispatch_proj.rs`**: `cublas_fp8_proj` is split so the already-quantized caller has an entry point (`cublas_fp8_proj_prequant`), and `cublas_fp8_m_pad` becomes the SSOT for the padded M extent.
- **Buffer sizing** (`spark-runtime`). cuBLASLt rejects a scale-tensor M extent that is not a multiple of 4, so the helper pads M to 16 and the matmul **writes** the phantom output rows and **reads** the phantom activation rows. `expert_gate_out` / `expert_up_out` / `moe_output` / `fp8_act` / `ffn_act_a` / `ffn_act_scale` now carry that 16-row headroom: without it a prefill chunk exactly as wide as `max_batch_tokens` writes up to 15 rows past its buffer, straight into the neighbouring arena buffer. Cost is ≤15 rows per buffer, ~0.5 MB on a 27B. The padded activation **bytes** are zeroed alongside their scales, because a zero scale kills a phantom row's contribution but `NaN * 0.0` is `NaN`.

Nothing is copied, staged or requantized on the weight side: both arms read the same FP8 weight and the same `[N/128, K/128]` block-scale buffer.

## Oracle and evidence

All of this is **diagnostic single-H100 evidence, not certified campaign numbers.** 1×H100 80 GB, `Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`, native FP8 profile, `--max-batch-size 16 --ssm-decode-ring-slots 1 --ssm-batched-recurrent true`, temperature 0, 1 warm-up + 3 reps, frozen ladder harness `bench/ladder38/harness_w55_conc_ladder.py`, engine tip `5f78270dc` (these three commits). 2026-09-11.

### Microtest — `native_fp8_ffn_w8a8_microtest`, kernel path

Real Qwen3.8-27B dense-FFN shapes, CUDA-event timing, 10 iterations after warm-up. M=64 is a short prefill chunk; M=1193 is the trace prompt.

| shape | M | W8A16 ms / TFLOP/s | W8A8 kernel ms / TFLOP/s | speedup | max_abs | cosine | rel_rms |
| gate/up | 64 | 0.527 / **21.64** | 0.172 / **66.41** | 3.07× | 5.375 | 0.999680 | **0.0253** |
| gate/up | 1193 | 2.900 / **73.34** | 1.441 / **147.53** | 2.01× | 5.500 | 0.999680 | **0.0253** |
| down | 64 | 0.741 / **15.40** | 0.450 / **25.36** | 1.65× | 9.000 | 0.999679 | **0.0254** |
| down | 1193 | 3.169 / **67.11** | 1.455 / **146.21** | 2.18× | 9.500 | 0.999679 | **0.0253** |

**2.0–3.1× over W8A16**, cosine ≥ 0.9996 everywhere.

At the example's default `rel_rms` bound the run exits 1 with exactly four failures, one per cell:

```
FAIL: gate/up M=64: rel_rms 0.0253 > 0.02        (x4, one per shape/M)
Error: 4 gate(s) failed
W8A8_KERNEL_EXIT=1
```

Re-run with `ATLAS_W8A8_REL_RMS_GATE=0.03` — **identical numbers**, and:

```
RESULT: PASS (all shapes within cosine/rel_rms/ULP gates)
W8A8_KERNEL_G03_EXIT=0
```

Both readings are reported rather than only the passing one. The example's own header argues 2.53% is the **E4M3 per-element floor, not a defect**: 3 stored mantissa bits give ~2.5% RMS per element and a dot product of independent terms does not average it down, so the predicted band is ~2–2.6%, sitting right on a 2% gate. The measurement lands inside that band on all four cells with cosine 0.99968 on all four — isotropic noise, not a systematic shear. **The 2% constant is what looks wrong here**, set below the arithmetic's own floor so it can only ever fail; it is under review and deliberately **not** changed in this PR, because moving a gate in the same commit that first trips it is how gates stop meaning anything.

### cuBLASLt arm — measured, failing, and not enabled

Reported because it is in the diff, not because it is claimed:

| shape | M | cuBLASLt ms / TFLOP/s | vs kernel max_ulp | vs kernel rel_rms | vs W8A16 cosine |
|---|---|---|---|---|---|
| gate/up | 64 | 0.048 / 237.34 | 32 591 | 1.12e-2 | 0.999620 |
| gate/up | 1193 | 0.187 / **1140.16** | 33 695 | **7.67e-2** | **0.996738** ✗ |
| down | 64 | 0.045 / 253.98 | 32 747 | 1.12e-2 | 0.999618 |
| down | 1193 | 0.211 / 1008.89 | 33 675 | **8.82e-2** | **0.995785** ✗ |

The microtest's contract is that the two arms take the same quantized inputs through the same FP32 epilogue and should agree to ~4 ULP; measured is ~33,000 ULP with 84–94% of output elements differing, and at M=1193 cuBLASLt is *further* from the W8A16 reference than the W8A8 kernel is. That is not a second valid rounding of the same math. Relaxing the rel-RMS bound does not rescue it — 6 gates still fail at `ATLAS_W8A8_REL_RMS_GATE=0.03` — so the failure is orthogonal to the gate-bound question above. Root cause is the VEC128 activation-scale layout; it is fixed on a follow-up branch that will come with its own H100 receipt. Until then the flag is off by default and should stay off.

### Ladder — before vs after, same recipe

Round 2 (tip `fbbe70767`, W8A16 prefill) → round 3 (these commits, W8A8 kernel). Shapes are the harness's `1024×256` and `4096×512` labels — **1193** and **4593** prompt tokens against 256 / 512 output.

| shape | C | tok/s before → after | TTFT p50 before → after | TPOT p50 before → after |
|---|---|---|---|---|
| 1193 / 256 | 1 | 39.41 → **41.78** | 1,071 ms → **707 ms** | 21.27 → 21.26 |
| 1193 / 256 | 16 | 66.05 → **113.03** (agg) | 9,070 → 5,957 ms | 207.57 → **117.58** |
| 4593 / 512 | 1 | 33.29 → **35.16** | 3,125 ms → **2,310 ms** | 23.98 → 23.97 |
| 4593 / 512 | 16 | 57.79 → **91.76** (agg) | 26,453 → 19,611 ms | 225.63 → 137.36 |

`LADDER_EXIT=0`, **zero errors** in all 12 recorded reps, rep-to-rep spread ≤2.26%. TTFT falls 34% and 26% at C=1, which is the prefill change showing up where prefill is measured; the C=16 aggregate nearly doubles on the short shape. The 28-token smoke request's TTFT moved 174.8 ms → 100.9 ms on the same box.

The arithmetic change is visible in the serve log rather than inferred:

```
[atlas] dense FFN prefill: W8A8 block-scaled via kernel (per-token 1x128 act
  scales x 128x128 weight scales, FP32 epilogue; vLLM-equivalent FP8 numerics).
  ATLAS_FFN_W8A16_ONLY=1 restores W8A16.
```

No W8A16 line was emitted alongside it, so no fallback diagnosis is needed, and `ATLAS_CUBLAS_GEMM` was unset on that leg.

### Coherency

`bench/hopper_ab/coherency_gate.py`, same session, after `--selftest` was shown to refuse its own known-bad fixtures first:

| probe | verdict |
|---|---|
| determinism | **PASS** — two identical requests byte-identical |
| tool call | **PASS** — `get_weather` with `{"city":"Reykjavik","days":3}`, `finish_reason: tool_calls` |
| think-leak | **PASS** |
| known-answer `391` | **PASS** |
| known-answer `Tokyo` | **PASS** |
| known-answer word-reversal | **FAIL** |

The word-reversal probe fails on vLLM too (#929) and its failure mode here is a tokenizer artifact rather than an arithmetic one: the model enumerates all twelve letters of "refrigerator" in the correct order and then drops one on concatenation. A 16-way concurrent arithmetic probe was **16/16 correct** in 2.51 s wall on the same config.

**What this evidence does not establish.** One box, one model, one session, not bisected against anything but round 2's tip — which differed by exactly these three commits, which is the strongest part of the claim. It is not a certification run, it says nothing about GB10, and the precision trade is a real trade: W8A8 is lossier than W8A16 by construction, which is why the route is logged and `ATLAS_FFN_W8A16_ONLY` exists.

## Test plan

- [x] `cargo test -p spark-model --features cuda w8a8` — **12 passed, 0 failed** (the new dispatch/sizing tests)
- [x] `cargo test -p spark-model --features cuda ffn` — **21 passed, 0 failed**
- [x] `cargo test -p spark-runtime --lib --features cuda buffers` — **10 passed, 0 failed**, including `mixed_dense_moe_sizes_for_widest_ffn` with the pad pinned
- [x] `cargo build -p spark-model --example native_fp8_ffn_w8a8_microtest --features cuda,gpu-examples` — builds clean
- [x] `cargo clippy -p spark-model -p spark-runtime -p spark-server --tests --features cuda,nccl` — clean, no warnings emitted. `spark-server` is in the invocation only because `nccl` is its feature, not `spark-model`'s or `spark-runtime`'s; the two crates this PR edits are both covered.

- [ ] `scripts/hopper_ptx_gate.sh` — **not applicable**: these three commits touch no `.cu` and no `kernels/` file at all. `per_token_group_quant_fp8` and `fp8_gemm_t_blockscaled` are the entry points the attention W8A8 prefill already resolves; nothing new is compiled.
- [x] Tested against real hardware — H100 receipts above (2026-09-11), kernel path.
- [ ] cuBLASLt arm on hardware — **fails, see above**; the fix and its receipt are a separate PR.

## GB10 perf-gate records

This diff touches `crates/`, a PERF_PATH, so `record_covers` invalidates every committed `.benchmarks/` record on landing. A GB10 re-measure is owed before merge. The H100 ladder above is diagnostic evidence from a different card and does not substitute for it.

## Notes for reviewers

**Where the arm sits in `w8_gemm!` matters.** It is after the `w8a16_gemv_batch4` arm (which `main` has via #925) and before the W8A16 tile arms. The `m > 4` clause in the selector makes that ordering redundant rather than load-bearing — both guards say the same thing — but they are written independently on purpose: the macro arm is a dispatch detail and the selector is the documented rule, and a future edit to either should not be able to silently route a 2-row decode into an MMA tile.

**The three buffer sizes are the risky part of this diff, not the GEMM.** `expert_gate_out`, `moe_output` and the `ffn_act_*` scratch all gained 16-row rounding, and the failure mode if any of them had been missed is a cross-buffer write during a full-width prefill chunk rather than a wrong number — silent, shape-dependent, and invisible to a microtest. That is why `w8a8_gemm` bounds-checks the padded output against the arena's reported capacity before choosing cuBLASLt and falls back to the kernel rather than trusting the sizing, why the accessors expose `*_bytes()` at all, and why the third commit pins the pad in `mixed_dense_moe_sizes_for_widest_ffn` rather than letting the constant drift.

**On the failing 2% gate.** The honest reading is that the example shipped a bound below its own arithmetic's floor, and the first thing to run against it proved it. Changing the constant belongs in a change that argues for the new number on its own, not in the one that has an interest in passing.

The three commits are cherry-picked from `fix/ffn-w8a8-prefill` on the fork with `-x` provenance lines. **One conflict**, in `crates/spark-model/Cargo.toml`: the source branch's `[[example]]` block also registers `native_fp8_oproj_batch_microtest` (#984) and `native_fp8_qkv_batch_microtest` (#986), neither of which exists on `main`. Resolved by keeping only `native_fp8_ffn_w8a8_microtest`, this PR's own example. Everything else — including `dense_ffn.rs`, where the `w8_gemm!` macro was the thing most likely to have moved — applied to identical pre-images and was not adapted.

The later cuBLASLt VEC128 scale-layout fix is deliberately **not** in this branch. It is the thing that would make the fast path correct, and it should be reviewed against its own hardware receipt rather than smuggled in behind a set of numbers it did not produce.

## CLA

### DIFF
--- a/crates/spark-model/Cargo.toml
+++ b/crates/spark-model/Cargo.toml
@@ -279,3 +279,7 @@ required-features = ["cuda", "gpu-examples"]
 [[example]]
 name = "w4a16_qg_batch_bitparity_microtest"
 required-features = ["cuda", "gpu-examples"]
+
+[[example]]
+name = "native_fp8_ffn_w8a8_microtest"
+required-features = ["cuda", "gpu-examples"]
--- a/crates/spark-model/examples/native_fp8_ffn_w8a8_microtest.rs
+++ b/crates/spark-model/examples/native_fp8_ffn_w8a8_microtest.rs
@@ -1,0 +1,394 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+//! Dense-FFN W8A8 block-scaled prefill at the real Qwen3.8-27B shapes (#917/#928).
+//!
+//! WHY. The native-FP8 dense FFN ran its prefill GEMMs as W8A16 — BF16
+//! activations against E4M3 weights, so the MMA is the BF16 tensor-core path.
+//! On H100 (2026-09-11, 1193-token prompt) that put TTFT at 1075 ms against
+//! vLLM's 287 ms, with `w8a16_gemm_pipelined` turning ~12 TFLOP/s on these
+//! shapes. This microtest measures the replacement: the same W8A8 block-scaled
+//! arithmetic the attention projections already use, with cuBLASLt as the
+//! Hopper fast path.
+//!
+//! What it reports, per shape and per M:
+//!   * W8A8 (kernel) vs W8A16 — max_abs, cosine, relative RMS. W8A8 quantizes
+//!     the ACTIVATION to E4M3 per 128-wide K group, which W8A16 does not; the
+//!     difference is vLLM's dynamic W8A8 arithmetic and a deliberate precision
+//!     trade, not a defect.
+//!   * cuBLASLt vs the W8A8 kernel — same quantized inputs, same FP32
+//!     epilogue, so these should be near bit-identical; the max BF16 ULP
+//!     distance is printed so "near" is a number and not an adjective.
+//!   * TFLOP/s for each path (CUDA events, 10 iterations after warm-up).
+//!
+//! NUMERICS FLOOR — read before judging a marginal `rel_rms`. E4M3 carries 3
+//! stored mantissa bits, so round-to-nearest costs ~2.5% RMS relative error per
+//! element, and for a dot product of independent terms that error does NOT
+//! average down relative to the signal: the expected `rel_rms` of this
+//! comparison on random inputs is ~2-2.6%, sitting right on the 2% gate.
+//! Cosine is the robust metric (~0.9997 at that error). `ATLAS_W8A8_REL_RMS_GATE`
+//! overrides the bound for a measurement run; the value used is always printed.
+//!
+//! Run (H100):
+//!   cargo run --release -p spark-model --features cuda,gpu-examples \
+//!     --example native_fp8_ffn_w8a8_microtest
+//!   ATLAS_CUBLAS_GEMM=1 cargo run --release -p spark-model \
+//!     --features cuda,gpu-examples --example native_fp8_ffn_w8a8_microtest
+
+use anyhow::{Result, bail};
+use half::bf16;
+use spark_model::layers::ops;
+use spark_model::weight_map::{Fp8Weight, WeightQuantFormat};
+use spark_runtime::cuda_backend::AtlasCudaBackend;
+use spark_runtime::gpu::{DevicePtr, GpuBackend};
+
+// CUDA driver event API — kernel-only timing. Wall-clock `Instant` carries a
+// ~0.3 ms per-launch host floor that swamps the signal on these shapes.
+// Signatures mirror `examples/w8a16_microtest.rs` (SSOT for these decls).
+unsafe extern "C" {
+    fn cuEventCreate(event: *mut u64, flags: u32) -> i32;
+    fn cuEventRecord(event: u64, stream: u64) -> i32;
+    fn cuEventSynchronize(event: u64) -> i32;
+    fn cuEventElapsedTime(ms: *mut f32, start: u64, end: u64) -> i32;
+    fn cuEventDestroy_v2(event: u64) -> i32;
+}
+
+/// Qwen3.8-27B dense FFN: hidden 5120, intermediate 17408.
+const H: usize = 5120;
+const INTER: usize = 17408;
+/// 64 = a short prefill chunk; 1193 = the prompt length in the #917 H100 trace.
+const BATCHES: [usize; 2] = [64, 1193];
+const BLOCK: usize = 128;
+const ITERS: u32 = 10;
+const WARMUP: u32 = 3;
+const COSINE_GATE: f64 = 0.999;
+const REL_RMS_GATE: f64 = 0.02;
+
+struct Rng(u64);
+impl Rng {
+    fn next_u32(&mut self) -> u32 {
+        self.0 = self.0.wrapping_mul(6364136223846793005).wrapping_add(1);
+        (self.0 >> 32) as u32
+    }
+    /// Uniform in [-1, 1).
+    fn unit(&mut self) -> f32 {
+        (self.next_u32() % 2049) as f32 / 1024.0 - 1.0
+    }
+}
+
+fn upload(gpu: &dyn GpuBackend, bytes: &[u8]) -> Result<DevicePtr> {
+    let ptr = gpu.alloc(bytes.len().max(1))?;
+    gpu.copy_h2d(bytes, ptr)?;
+    Ok(ptr)
+}
+
+fn download_bf16(gpu: &dyn GpuBackend, ptr: DevicePtr, elems: usize) -> Result<Vec<u16>> {
+    let mut raw = vec![0u8; elems * 2];
+    gpu.copy_d2h(ptr, &mut raw)?;
+    Ok(raw
+        .chunks_exact(2)
+        .map(|c| u16::from_le_bytes([c[0], c[1]]))
+        .collect())
+}
+
+fn to_f64(bits: &[u16]) -> Vec<f64> {
+    bits.iter()
+        .map(|b| bf16::from_bits(*b).to_f32() as f64)
+        .collect()
+}
+
+/// BF16 bits → a monotonically ordered integer, so `|ord(a) - ord(b)|` is the
+/// ULP distance (the standard sign-magnitude → two's-complement remap).
+fn ord(bits: u16) -> i32 {
+    if bits & 0x8000 != 0 {
+        -((bits & 0x7fff) as i32)
+    } else {
+        bits as i32
+    }
+}
+
+struct Compare {
+    max_abs: f64,
+    cosine: f64,
+    rel_rms: f64,
+    max_ulp: i32,
+    unequal: usize,
+}
+
+fn compare(a_bits: &[u16], b_bits: &[u16]) -> Compare {
+    let (a, b) = (to_f64(a_bits), to_f64(b_bits));
+    let (mut dot, mut na, mut nb, mut max_abs, mut sq_diff, mut sq_ref) =
+        (0.0, 0.0, 0.0, 0.0_f64, 0.0, 0.0);
+    for (x, y) in a.iter().zip(&b) {
+        dot += x * y;
+        na += x * x;
+        nb += y * y;
+        max_abs = max_abs.max((x - y).abs());
+        sq_diff += (x - y) * (x - y);
+        sq_ref += y * y;
+    }
+    Compare {
+        max_abs,
+        cosine: dot / (na.sqrt() * nb.sqrt()),
+        rel_rms: (sq_diff / sq_ref.max(f64::MIN_POSITIVE)).sqrt(),
+        max_ulp: a_bits
+            .iter()
+            .zip(b_bits)
+            .map(|(x, y)| (ord(*x) - ord(*y)).abs())
+            .max()
+            .unwrap_or(0),
+        unequal: a_bits.iter().zip(b_bits).filter(|(x, y)| x != y).count(),
+    }
+}
+
+/// GPU time per iteration for `launch`, in seconds (CUDA events, no host sync
+/// between iterations).
+fn time_gpu(
+    gpu: &dyn GpuBackend,
+    stream: u64,
+    mut launch: impl FnMut() -> Result<()>,
+) -> Result<f64> {
+    for _ in 0..WARMUP {
+        launch()?;
+    }
+    gpu.synchronize(stream)?;
+    let (mut ev_start, mut ev_end) = (0u64, 0u64);
+    for (ev, what) in [(&mut ev_start, "start"), (&mut ev_end, "end")] {
+        let rc = unsafe { cuEventCreate(ev, 0) };
+        if rc != 0 {
+            bail!("cuEventCreate({what}) failed: status {rc}");
+        }
+    }
+    if unsafe { cuEventRecord(ev_start, stream) } != 0 {
+        bail!("cuEventRecord(start) failed");
+    }
+    for _ in 0..ITERS {
+        launch()?;
+    }
+    if unsafe { cuEventRecord(ev_end, stream) } != 0 {
+        bail!("cuEventRecord(end) failed");
+    }
+    if unsafe { cuEventSynchronize(ev_end) } != 0 {
+        bail!("cuEventSynchronize failed");
+    }
+    let mut ms: f32 = 0.0;
+    if unsafe { cuEventElapsedTime(&mut ms, ev_start, ev_end) } != 0 {
+        bail!("cuEventElapsedTime failed");
+    }
+    unsafe {
+        cuEventDestroy_v2(ev_start);
+        cuEventDestroy_v2(ev_end);
+    }
+    Ok((ms as f64 / 1e3) / ITERS as f64)
+}
+
+fn tflops(m: usize, n: usize, k: usize, secs: f64) -> f64 {
+    2.0 * m as f64 * n as f64 * k as f64 / secs / 1e12
+}
+
+struct Shape {
+    label: &'static str,
+    n: usize,
+    k: usize,
+}
+
+fn main() -> Result<()> {
+    let gpu = AtlasCudaBackend::new(0, &atlas_kernels::ptx_modules())?;
+    let stream = 0u64;
+    let w8a16_k = gpu.kernel("w8a16_gemm_pipelined", "w8a16_gemm_pipelined")?;
+    let quant_k = gpu.kernel("per_token_group_quant_fp8", "per_token_group_quant_fp8")?;
+    let w8a8_k = gpu.kernel("fp8_gemm_t_blockscaled", "fp8_gemm_t_blockscaled")?;
+    let want_cublas = std::env::var("ATLAS_CUBLAS_GEMM").as_deref() == Ok("1");
+    let rel_rms_gate = std::env::var("ATLAS_W8A8_REL_RMS_GATE")
+        .ok()
+        .and_then(|v| v.parse::<f64>().ok())
+        .unwrap_or(REL_RMS_GATE);
+
+    println!(
+        "dense-FFN W8A8 microtest — H={H} INTER={INTER}  cuBLASLt={}  gates: cosine>={COSINE_GATE} rel_rms<={rel_rms_gate}",
+        if want_cublas {
+            "on (ATLAS_CUBLAS_GEMM=1)"
+        } else {
+            "off"
+        }
+    );
+
+    let mut rng = Rng(0x9_17_09_28_2026);
+    let max_m_pad = BATCHES
+        .iter()
+        .map(|m| m.div_ceil(16) * 16)
+        .max()
+        .expect("BATCHES is non-empty");
+    let max_k = H.max(INTER);
+    let max_n = H.max(INTER);
+
+    // Activations [max_m_pad, max_k] BF16 — one buffer, sliced per shape. The
+    // padded rows exist because the cuBLASLt arm reads ceil16(M) rows.
+    let act_host: Vec<u8> = (0..max_m_pad * max_k)
+        .flat_map(|_| bf16::from_f32(rng.unit()).to_bits().to_le_bytes())
+        .collect();
+    let act = upload(&gpu, &act_host)?;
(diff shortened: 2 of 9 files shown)


tbraun96:

**Evidence of containment** (measured 2026-09-11 against `pr895` = `762fce294`, `avarok/main`, `pr987`):

* Of the 11 files this PR changes relative to `main`, **0** are still byte-identical to `main` on #895 — #895 has independently modified every one.
* `crates/spark-model/src/layers/dense_ffn_w8a8_prefill.rs` exists on #895 and is **this PR's module with #989's VEC128 `kmajor` work already applied on top** (`git diff pr895 pr987` on that file shows only the removal of the later additions).
* Symbol counts, `git grep -c ... -- crates/`:

  | symbol | #987 | #895 |
  |---|---|---|
  | `w8a8_prefill_selected` | 2 | 4 |
  | `log_w8a8_prefill_route` | 1 | 1 |
  | `cublas_fp8_proj_prequant` | 4 | 13 |
  | `cublas_fp8_m_pad` | 3 | 13 |

A trial merge of this PR into #895 conflicts in 8 files, all of them files where #895 holds a **newer** version of the same code. Resolving that would mean hand-merging FP8 GEMM dispatch against a branch that already contains a further-evolved copy — the failure mode being a silent numerical defect, not a build error.

**One finding from this PR carries forward and is not yet in #895.** The selection rule here — and on #895 — has no upper `M` bound, and on GB10 that is a measured regression:

| arm | TTFT median, M=949, `Qwen3.6-27B-FP8`, GB10 |
|---|---|
| W8A8 (this rule) | **3343.3 ms** |
| W8A16 (with an M ceiling) | **2560.4 ms** |

−23.4%, n=5/leg, complete separation (max capped 2588.0 < min uncapped 3248.0), both arms proven distinct from the serve log. W8A8 throughput is flat at ~14 TFLOP/s from M=128–2048 on sm_121 while W8A16 climbs to ~26; on H100 W8A8 is 2.0–3.1× at every M measured, which is why the fix is a **per-arch ceiling** rather than a kill switch. That lands separately on #895's branch via its `[defaults]` / `TARGET_DEFAULTS` table.

Reachability note for whoever picks this up: this arm is reached only under `ATLAS_DENSE_FP8=1`. `git grep ATLAS_DENSE_FP8` returns one hit — its own reader. No gate, no CI job and no default serve exercises it today.

TheTom:

tbraun96:

tbraun96:
## Measured: this arm is a 23.4% TTFT regression on GB10 above a shape-dependent M

Not asking you to change this PR — the kernel work is right and the numbers below were taken *with* it. Recording the measurement here so it is attached to the code it describes rather than living on a scratch branch.

**Served A/B, spark-256a (GB10/sm_121), `Qwen/Qwen3.6-27B-FP8`, M=949, byte-identical 949-token prompt, 1 warm-up + 5 timed, TTFT median. Two binaries differing only in an upper-M bound on `w8a8_prefill_selected`:**

| arm | route (read back from the serve log) | TTFT median |
|---|---|---|
| no cap (this PR's rule) | `dense FFN prefill: W8A8` | **3343.3 ms** |
| capped → falls back | `dense FFN prefill: W8A16` | **2560.4 ms** |

**−782.9 ms, −23.4%.** Complete separation: max(capped) 2588.0 < min(uncapped) 3248.0. Reproduced in a second independent run at 2555.8 ms (0.2%).

**Why, and why it is not an argument against the PR.** W8A8 feeds the FP8 tensor cores instead of dequantizing into a BF16 MMA, and it pays a per-token activation quantization plus an FP32 scale epilogue that W8A16 never pays. That cost is fixed per token, so W8A8 wins only while the GEMM is small enough that the quantization is not the bill. On H100 it is 2.0–3.1× at every M — your microtest numbers. On sm_121 it is **flat at ~14 TFLOP/s from M=128 to M=2048** while W8A16 climbs to ~26 and stays there. A kernel whose throughput does not move with M is not compute-bound.

So the right shape of fix is a **per-arch upper bound on M**, not a kill switch. Measured crossovers at the real Qwen3.8-27B dims: gate/up (N=17408, K=5120, widening) ~M=64–128; down (N=5120, K=17408, narrowing) ~M=384–512 — about 6× apart, so it wants two values, not one.

**Reachability, which matters for how urgent this is.** This whole path runs only under `ATLAS_DENSE_FP8=1`, and `git grep ATLAS_DENSE_FP8` returns one hit — its own reader. No CI job, no `BENCH.toml` entry and no gate sets it, so no certification record has ever exercised this code. It is off by default because on GB10 native dense FP8 loses to the NVFP4 fallback outright — same box, same binary, same prompt: default **1437.2 ms** vs 2555.8 ms capped vs 3343.3 ms uncapped. So the ceiling is worth 23.4% *within* the dense-FP8 path, and that path is still the slower choice on this arch.

**Why I am not just pushing the fix here.** I implemented it as two `u32` rows in the `[defaults]` table (`kernels/<hw>/HARDWARE.toml` → `atlas_kernels::TARGET_DEFAULTS`), which is the SSOT mechanism that matches how this repo already expresses per-arch divergence — but that mechanism is **not in this PR and not in #996**; it currently exists only on the umbrella branch. Adding a second, self-contained ceiling here would duplicate a mechanism that is about to land properly.

Suggestion: land this PR as-is, land the `[defaults]` mechanism in its own clean PR, then the ceiling is two rows of data plus one `&& m <= max_m` clause. I have that change written and tested (controls both ways: defeating the bound turns exactly the two new ceiling tests red, 15 pass / 2 fail; restoring gives 17/17) and will open it as a follow-up the moment the mechanism has a home. Say the word if you would rather it went in here instead.

TheTom:
## Agreed — land as-is, ceiling goes on the defaults PR. All three of your code claims reproduce.

I verified each one against the branch rather than accepting it:

* **No upper `M` bound.** `crates/spark-model/src/layers/dense_ffn_w8a8_prefill.rs:88-102` — the rule is an eight-clause conjunction (`!w8a16_only && m > 4 && fp8_blockscaled_prefill && scale_format == Fp8BlockScaled && k % 128 == 0 && n % 128 == 0 && quant_k.0 != 0 && gemm_k.0 != 0`). There is a lower bound at `m > 4` and nothing above it. You are right.
* **One reader for the lever.** `git grep ATLAS_DENSE_FP8` on this branch returns two lines, both in `crates/spark-model/src/weight_loader/qwen35_dense.rs` — `:167` is the read, `:165` is that function's own doc comment. One actual reader, as you said. No gate, no CI job, no `BENCH.toml` entry reaches it.
* **`TARGET_DEFAULTS` has no home yet.** `git grep -c TARGET_DEFAULTS` is **0** on `avarok/main` and **0** on #996's branch; it exists only on #895's (`crates/atlas-kernels/src/target_defaults.rs`, `build_defaults.rs`, `crates/spark-model/src/layers/ops/target_defaults.rs`). Correct, and it is the reason not to bolt a second self-contained ceiling in here.

**So: your suggested sequencing, not the alternative.** Do not put the ceiling in #987. The `[defaults]` / `TARGET_DEFAULTS` mechanism is being cut right now as its own clean PR off the umbrella; once it is up, the ceiling is the two `u32` rows plus the `&& m <= max_m` clause you already have written and control-tested. Duplicating the mechanism here to land it a few hours earlier would leave a second SSOT to unpick.

**Corroboration from our side, which I think strengthens the two-value shape.** Your ceiling is already on the umbrella as `d4112c3eb`, and our round-14 H100 campaign ran with it: the boot line reads `w8a8_prefill_max_m=max/max` with no `(env)` tag — `w8a8_prefill_max_m_widening = 4294967295` and `w8a8_prefill_max_m_narrowing = 4294967295` on hopper. A deliberate no-op on the arch where W8A8 is 2.0–3.1x, so the per-arch mechanism is already shown not to disturb H100; the GB10 rows are the half that carries the win. And the widening/narrowing split is already the shape the lever has, which matches your measured ~6x crossover gap (gate/up ~M=64–128, down ~M=384–512) — it wants two values and it has two.

**One scoping note for whoever prioritises the follow-up.** By your own numbers the default GB10 serve is the NVFP4 fallback at 1437.2 ms, against 2555.8 ms capped and 3343.3 ms uncapped on the dense-FP8 path. So the ceiling is a 23.4% win *inside* a path that is off by default and still the slower choice on that arch. Worth landing, correctly, on the mechanism — not worth holding #987 for.

Thanks for attaching the measurement to the code it describes instead of leaving it on a scratch branch. That is the right place for it.

TheTom:
Merged base #986 forward (merge commit `d9d9bd802`).

Conflict: `crates/spark-model/Cargo.toml` — add/add on adjacent `[[example]]` stanzas (both layers were cherry-picked from `b2a7ae16e`). Resolved to the **union of both sides**, ordered as `hopper/h100-integration-2026-09-10` orders them.

PR's own diff unchanged (verified): all 8 non-`Cargo.toml` files byte-identical to `d6d1073b9`; the `Cargo.toml` delta is a pure addition of #986's two stanzas.

Distance from the measured tree (`4f00cf199`) on that file: **+0 / −57** — a strict subset, the deletions being only the stanzas from layers this PR doesn't carry.

Gates on Spark 1 (GB10, CUDA 13): `cargo fmt --all -- --check` ok · clippy `-p spark-model --features cuda` / `-p spark-runtime --features cuda` `-D warnings` ok · `cargo test -p spark-model --lib` 910 passed 0 failed · `-p spark-runtime --lib` 297 passed 0 failed · `cargo test -p atlas-kernels --no-default-features` ok · CI file-size cap (≤500 LoC, `.github/workflows/file-size-cap.yml` step run verbatim) ok.

TheTom:
Agreed on all three points, and verified on my side: the arm is only reachable under `ATLAS_DENSE_FP8=1` (one reader, `weight_loader/qwen35_dense.rs`), so merging this PR as-is leaves the GB10 default route byte-for-byte where it was; the Spark 2 regression run against `main` will carry that receipt on this PR when it lands.

The mechanism now has its home: #1006 (`[defaults]` per target, baked into `TARGET_DEFAULTS`, resolved with the `ATLAS_*` grammar and printed on the boot line), stacked on #996. It deliberately omits the `w8a8_prefill_max_m_*` rows because their consumer is this PR — the contract in `target_defaults.rs` is one commit per row landing with the arm that reads it. So please open your follow-up as you described (two rows + the `&& m <= max_m` clause + the two ceiling tests), stacked on #1006 → #987; H100 keeps `u32::MAX` on both (2.0–3.1× at every M measured, rounds 11–15).

TheTom:
Follow-up you proposed is up: **#1019** (`pr/w8a8-prefill-max-m`) — per-arch upper M bound on W8A8 FFN prefill.

Stacked on this PR after merging #1006 (TARGET_DEFAULTS). Cherry-picked `-x` from your `d4112c3eb` (author kept). GB10 64/384, Hopper `u32::MAX`, B200 `u32::MAX`. Spark 2 GB10 receipt still pending.

Not landed in this PR, as agreed.

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1029 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1029; this fork PR is closed.

PR #988, opened 2026-09-11
spark-model: batch native FP8 FFN decode across 5..32 rows with w8a16_gemv_batch16 (#927)

TheTom:
> **Stacked on #986 (and #984), and on #987.** Their six commits are the first six here; this PR's own content is the last three — `spark-model: batch native FP8 FFN decode across 5..32 rows with w8a16_gemv_batch16 (#927)`, a one-line `mut` cleanup on the microtest launcher, and `spark-model: make the batch16 dense-FFN decode tier opt-in (#927)` (`8db757c77`, the review fix below). Review those — or `git diff pr/o13-fp8-qkv-batch..` for this PR plus #987, `git diff 793384e03..` for this PR alone. Merge #984, then #986, then #987. Base is `main` rather than a stack branch so this PR does not become unreviewable if any of them is rebased.
>
> #987 is in the base because it is not optional here: this PR rewrites the same `w8_gemm!` macro #987 adds its arm to, and threads the new rungs *ahead* of that arm. See **Conflicts** below.

## Summary

> **⚠️ Revised 2026-09-11 after a single-variable A/B (round 5). The FFN batch16 tier is now DEFAULT OFF and opt-in via `ATLAS_FFN_BATCH16=1`.** It measured a **net loss** in serving on H100 and has **never been measured on GB10**. An earlier revision of this body quoted a **−15.5% step** from a round-4 profile; that number was confounded and **has been withdrawn** — see [What changed, and why the headline is gone](#what-changed-and-why-the-headline-is-gone). What this PR still lands on its own merits is the **bit-exactness** result and the **wrapper/handle-pairing correctness fix**; the perf tier ships disarmed until a target earns it.

Raising `--max-batch-size` from 4 to 16 on 1×H100 made native-FP8 decode **slower in aggregate**: the step cost 44 ms at 4 active rows and **224 ms at 16**, so C=16 throughput FELL from 76 to 62 tok/s. Five extra rows cost five times the step.

That is a dispatch gap, not a missing kernel. Every native-FP8 decode site stopped at the M≤4 batched GEMV and handed 5..16 rows to tile GEMMs that pad M to a 128-row MMA tile — at M=16, seven eighths of every tile is padding, on projections that are purely weight-bandwidth bound. The MAX_M=16 instantiation of the same template, `w8a16_gemv_batch16`, already exists in `kernels/gb10/common/w8a16_gemv_batch4.cu` (#984/#986); nothing dispatched to it.

This PR wires it in at six sites and adds a second rung for 17..=32 rows. **No kernel source changes** — this PR's diff touches zero files under `kernels/`.

The dense-FFN rungs are **off unless `ATLAS_FFN_BATCH16=1`**. The other five sites stay default-on: they are separate levers on their own kernel handles, they were on in **both** halves of the A/B that condemned the FFN tier, and each is bit-identical per row to the M=1 GEMV it replaces.

Refs #927.

## What was wrong

`dense_ffn.rs`'s `w8_gemm!` claimed only `(1..=4)` for `w8a16_gemv_batch4`. At m = 5..16 it fell through to the transposed `w8a16_gemm_n128_m128` / `w8a16_gemm_pipelined` arms — the 5–12 TFLOP/s class at these shapes. The same flat-4 ceiling was in five more places:

- **attention `o_proj`** — `step` was a flat 4, so 16 concurrent rows read the o_proj weight `ceil(n/4)` = **4 times**.
- **multi-seq FP8 QKV** (#986's module) — the band read `2..=8` against a comment claiming the padded-n ladder was `[2,4,8]`. The ladder has had rungs **12 and 16** since the C=[1,2,4,8,16] concurrency work, so padded_n 12 and 16 were falling back to the per-sequence scalar loop: `3n` launches and `n` full weight passes per attention layer per step.
- **GDN/SSM verify QKVZ and out_proj** — `R = Σ ks` above 4 went straight to the tile GEMMs.
- **BF16 LM head** — the batched-GEMV band was the frozen constant 8, with no per-site lever.

And one latent trap: `ssm_batched.rs` already *selected* `w8a16_gemv_batch16_k` for n > 4, then launched it through the `ops::w8a16_gemv_batch4` wrapper. That worked only because the wrapper did not check `m`.

## What the fix does

**Dispatch, `dense_ffn.rs`.** `w8_gemm!` gains an explicit, commented arm ladder, decode rungs first:

```
1. m <= 4      w8a16_gemv_batch4        one weight pass, 4-row tier
2. m 5..=16    w8a16_gemv_batch16       OPT-IN, off by default
3. m 17..=32   w8a16_gemv_batch16 x2    OPT-IN, off by default
4. W8A8 block-scaled prefill            (#917/#928)
5. transposed / pipelined / base W8A16 tile GEMMs
```

**Rungs 2–3 are disarmed unless `ATLAS_FFN_BATCH16=1`, so a stock serve runs rungs 1, 4 and 5 exactly as it did before this PR.** They live in a new child module `dense_ffn_batch16_decode.rs` (`dense_ffn.rs` is already at the CI size cap) together with the selection rule, the opt-in, the receipt that made it opt-in, and the consequence below. The gate is latched once per layer into `DenseFfnLayer::batch16_enabled` rather than read per dispatch: that fixes the route before any CUDA-graph capture, and it lets the dispatch tests pin **both** polarities, which a process-global `OnceLock` could not. 17..=32 runs the same kernel twice on contiguous row halves — the FFN activations and outputs are contiguous `[m,k]`/`[m,n]`, so a half is a byte offset, no staging and no strided variant. `ceil(m/2)` splits so both halves are ≤16 and the first is the wider one.

**🪤 A real boundary move, and it is the one that sank the tier.** #987's W8A8 prefill rule starts at `m > 4`, so with rungs 2–3 armed ahead of it the W8A8 path begins at **m > 32** in practice — and a *prefill* of 5..=32 tokens (a very short prompt, or the **tail chunk of a chunked prefill**) takes the GEMV too. That is not a footnote: it is where the +50 ms TTFT below comes from, because the tier dispatches by **row count** and not by phase. A 1193-token prompt arrives as `1168 + 25`, and the 25-row tail lands in the band. Disarmed by default now; `ATLAS_FFN_BATCH16=1` is what opts back in. The W8A8 activation quantizer is also gated on `batch16.is_none()` rather than only the match arm, so at m=5..=32 it is not launched as dead work whose result no arm reads.

**The five other sites — separate levers, still default-on.** None of them reads `ATLAS_FFN_BATCH16`; each keys off its own kernel handle. They were armed in **both** halves of the A/B below, so none of that table's movement is theirs to claim or to blame. o_proj walks in 16-row groups instead of 4; the multi-seq FP8 QKV band widens 2..=8 → 2..=16 (the upper edge is the kernel's MAX_M, not a tuning choice — it clamps rather than erroring); the GDN/SSM verify QKVZ and out_proj gain `(5..=16)` arms placed *after* the existing NVFP4 arms, so a checkpoint carrying both formats keeps the format it picks today; `ssm_batched.rs` now carries the wrapper and the handle as one pair.

**`ops::w8a16_gemv_batch4` now refuses `m > 4`.** The kernel is `w8a16_gemv_batchm_impl<4>`: at M=5 it computes rows 0..3 and never writes rows 4.. — stale memory, not a launch failure. `ops::w8a16_gemv_batch16` is its contiguous sibling with the same signature and launch geometry. This is what forces the `ssm_batched.rs` pairing above; without it, that path would now `Err` at every C=8/16 step instead of silently reading the right kernel through the wrong bound.

**LM head.** The BF16 decode band becomes a resolved value behind `ATLAS_LM_HEAD_BATCHM_MAX`, clamped to the kernel's compile-time row bound. **The default is the frozen 8, so GB10 bits are unchanged by construction** — the A/B behind that constant measured the GEMV *negative* above 8 on GB10 (−14.4% at C=16, commits 84d5b763c / 78d276832), and `DENSE_GEMV_BATCHM_DECODE_MAX_M` is deliberately not edited. H100 sets 16 because on that machine it is the tile GEMM that loses at decode widths. The lever moves this head and nothing else.

Both new levers are `OnceLock`-cached, which is also what keeps the route constant across CUDA-graph replays — a per-call `env::var` could change the captured launch set between capture and replay. The QKV band branches on the **padded** ctx `n`, the value the graph cache is keyed by, never on the unpadded sequence count.

## Oracle and evidence

**Every row is bit-identical to the scalar `w8a16_gemv` that M=1 decode runs** — same template, same K-iteration order, same per-row reduction tree, accumulators independent, `M` in no row's operand sequence. So these widths move **off** a reassociating tile GEMM and **onto** the bits decode already produces — the direction that removes a numerics seam rather than adding one.

`examples/native_fp8_ffn_batch16_microtest.rs`, 1×H100 80 GB, 2026-09-11, at the real Qwen3.8-27B FFN shapes (gate/up N=17408 K=5120; down N=5120 K=17408), each row compared against production scalar `w8a16_gemv`. `MT_EXIT=0`, and both negative controls (`KNOWN_BAD gate/up output-bit`, `KNOWN_BAD down output-bit`) are **refused** first, so the harness is shown able to fail before it is believed.

| shape | M | passes | parity | batch16 ms / GB/s | pipelined ms / GB/s | speedup |
|---|---|---|---|---|---|---|
| gate/up | 5 | 1 | `unequal_bf16=0 max_abs=0.000000000 guards=ok` | 0.119 / **751.7** | 0.630 / 141.5 | **5.31×** |
| gate/up | 8 | 1 | `unequal_bf16=0 max_abs=0.000000000 guards=ok` | 0.153 / **581.7** | 0.612 / 145.6 | **3.99×** |
| gate/up | 16 | 1 | `unequal_bf16=0 max_abs=0.000000000 guards=ok` | 0.260 / 342.4 | 0.586 / 152.1 | **2.25×** |
| gate/up | 32 | 2 | `unequal_bf16=0 max_abs=0.000000000 guards=ok` | 0.547 / 326.2 | 0.564 / 158.0 | 1.03× |
| down | 5 | 1 | `unequal_bf16=0 max_abs=0.000000000 guards=ok` | 0.132 / **674.1** | 0.878 / 101.5 | **6.64×** |
| down | 8 | 1 | `unequal_bf16=0 max_abs=0.000000000 guards=ok` | 0.190 / 469.4 | 0.841 / 106.0 | **4.43×** |
| down | 16 | 1 | `unequal_bf16=0 max_abs=0.000000000 guards=ok` | 0.330 / 270.3 | 0.798 / 111.7 | **2.42×** |
| down | 32 | 2 | `unequal_bf16=0 max_abs=0.000000000 guards=ok` | 0.721 / 247.3 | 0.780 / 114.2 | 1.08× |

```
ALL PASS: real Qwen3.8-27B FFN shapes, batch16 M5/M8/M16 and 2x-halves M32
exactly equal to scalar w8a16_gemv, guards intact      MT_EXIT=0
```

The win is a *bandwidth* win and its size is set by how many times the weights are streamed, so it decays monotonically as M rises (5.31 → 3.99 → 2.25). **At M=32 the tier makes two passes by design and the win collapses to 1.03×/1.08×** — the 17..32 rung is essentially free but is not where the value is. The value is the 5..16 band, which is exactly the band a `--max-batch-size 16` serve decodes in.

### What changed, and why the headline is gone

An earlier revision of this body led with **−15.5% total step** and **−41.3% `attn`**, from a round-4 per-phase profile. Round 5 ran the experiment that profile could not: a **single-variable A/B** — same binary, same 16-way burst, the tier the only thing that differs — plus a control boot proving the two binaries agree on the non-cuBLASLt path to within 0.6%. Both round-4 numbers are withdrawn. The −15.5% was a two-lever, cross-binary comparison; the −41.3% `attn` attribution is **refuted outright**, because turning the tier off moves `attn` *further down* (19.90 → 18.82 ms) instead of back up.

What replaces them is smaller, negative, and single-variable.

### The A/B: the tier is a net loss on H100

1×H100 80 GB, `Qwen/Qwen3.8-27B-FP8`, native FP8, `--max-batch-size 16`, temp 0 / seed 42, frozen ladder, env verified live in the running process. Serve **G** = tier on, serve **E** = tier off, **same round-5 binary**.

| shape | metric | tier ON (G) | **tier OFF (E)** | Δ from turning it off |
|---|---|---|---|---|
| 1024×256 | **C=16 agg** | 121.37 tok/s | **128.00 tok/s** | **+5.5%** |
| 1024×256 | **C=16 TPOT p50** | 107.31 ms | **101.98 ms** | **−5.0%** |
| 1024×256 | C=16 TTFT p50 | 6 327.6 ms | 5 966.7 ms | −5.7% |
| 1024×256 | C=1 TPOT | 21.29 ms | 21.26 ms | −0.1% (control) |
| 1024×256 | **C=1 TTFT p50** | 743.8 ms | **706.6 ms** | **−5.0%** |
| 4096×512 | **C=16 agg** | 98.98 tok/s | **102.13 tok/s** | **+3.2%** |
| 4096×512 | **C=16 TPOT p50** | 123.07 ms | **118.50 ms** | **−3.7%** |
| 4096×512 | C=1 TPOT | 23.99 ms | 23.98 ms | −0.04% (control) |

**Eight cells move in the tier-off direction, three are flat, none favours the tier.** Tier-off C=1 also reproduces the pre-#927 binary to the last digit (41.77/41.78 tok/s, 706.6/706.8 ms TTFT, 21.26/21.26 ms TPOT), which is the control that licenses reading the C=16 rows.

**The smoke TTFT is a step function, not a rate change:**

| serve | tier | 28-token smoke TTFT |
|---|---|---|
| pre-#927 binary | absent | **100.86 ms** |
| tier OFF | off | **100.90 ms** |
| tier ON | on | **149.81 / 150.00 ms** |

A fixed **~+49 ms** on a 28-row prompt, across three binaries, with the tier as the only difference — the chunked-prefill tail-chunk consequence flagged above, paid per request. On the ladder it shows up as +37 ms (short shape, 25-row tail) and +28 ms (long shape, 17-row tail).

### Per-phase, single-variable

`ATLAS_MS_PROFILE=1` on both boots, n=16, no `ATLAS_SSM_MS_PROFILE`. ⚠ The profile forces eager execution (CUDA graphs off), so absolutes are an upper bound — **read the ratios**.

| phase (n=16) | tier ON | **tier OFF** | Δ |
|---|---|---|---|
| total step | 86.80 ms | **82.32 ms** | **−5.2%** |
| `ssm` (48 layers) | 63.31 ms | **59.91 ms** | **−5.4%** |
| `attn` (16 layers) | 19.90 ms | **18.82 ms** | **−5.4%** |
| `head` | 3.58 ms | 3.58 ms | **0.00%** |
| `ssm` per layer | 1319 µs | **1248 µs** | −5.4% |

Two independent instruments agree to 0.2 percentage points (profile −5.2% total step, ladder −5.0% C=16 TPOT), one eager and one with CUDA graphs on. And `ssm` per layer lands on **1248 µs** against a pre-#927 **1252 µs**: the tier's cost is the *whole* of the `ssm` regression round 4 reported, not part of it.

Two honest notes on that table:

- **`attn` improves too, by the same 5.4%.** The o_proj and QKV arms were armed in **both** boots, so this movement is not evidence for or against them; it moved while they were untouched. The mechanism is not isolated here and is not claimed.
- **Round 4's `attn` −41.3% is refuted, not re-explained.** If the tier had caused it, disarming the tier would send `attn` back toward round 3's 33.90 ms. It does not. The remaining candidates are outside this PR's scope and are named as unresolved rather than reassigned.

### GB10: never measured

`w8a16_gemv_batch16` is an instantiation inside `kernels/gb10/common/w8a16_gemv_batch4.cu`, so its handle resolves on **every** target carrying that module — GB10 included. The original presence-keyed kill switch therefore made this tier **default-on for GB10 on the strength of an H100 number that has since come out negative**, with no GB10 reading of its own. That is the second reason for the inversion, and it is the one that would have shipped. `ATLAS_FFN_BATCH16=1` is available for whoever runs that A/B; if GB10 wins, the lever to flip is the module's default, with the receipt beside it.

- **Diagnostic single-H100 evidence, not a perf gate.** One box, one model, one quantization. It is enough to disarm a tier on that box; it is not enough to arm one anywhere.
- **The cross-binary confound is closed for the A/B above, and only for it.** Round 4's comparison spanned two binaries; round 5 added a control boot (round-4 config on the round-5 binary) that reproduces every round-4 cell to within 0.6%, most to within 0.1%. That is what makes the ON/OFF table single-variable. The withdrawn −15.5% did not have it.
- **The head band default stays 8 and no GB10 bits change by default.** The `ATLAS_LM_HEAD_BATCHM_MAX=16` in the H100 recipe is an H100 setting; `DENSE_GEMV_BATCHM_DECODE_MAX_M` is untouched.
- **The FFN tier is off unless `ATLAS_FFN_BATCH16=1`.** With it unset — every target, including GB10 — 5..=32 route exactly where they did before #927.

**Host dispatch tests** pin every rung as a pure function of the row count, the handle's presence and the opt-in, without building a `ForwardContext`: `dense_ffn_batch16_decode_tests.rs` (the rungs, both halves splits, the handle-absent fallback, the unarmed default), plus the o_proj, QKV-band, SSM-verify and LM-head-lever cases (including the lever clamping and defaulting on junk).

Two of them pin the **default** specifically, through `forward_prefill` and not only through the pure rule. `stock_serve_is_the_pre_927_routing` drives a layer with the batch16 handle **loaded** and the tier **unarmed**, and asserts the M-padded tile GEMM at every width in 5..=32; `the_batch4_rung_is_untouched_by_the_opt_in` pins that 1..=4 keeps its pre-#927 arm either way. A later flip back to default-on has to change a test that says so out loud.

## Test plan

Re-run at `8db757c77` (the opt-in commit), Spark 1 (GB10), isolated worktree, `--jobs 8`:

- [x] `cargo test -p spark-model --features cuda ffn` — **35 passed, 0 failed** (was 33; +2 default-pinning tests)
- [x] `cargo test -p spark-model --features cuda o_proj` — **10 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda qkv` — **19 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda lm_head` — **9 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda batch16` — **18 passed, 0 failed** (was 16)
- [x] `cargo test -p spark-model --features cuda --lib` — **930 passed, 0 failed, 11 ignored** at the earlier tip (910 on the stack base `793384e03`); the filtered runs above are the re-gate for the opt-in commit
- [x] `cargo build -p spark-model --examples --features cuda,gpu-examples` — clean, `Finished dev profile`
- [x] `cargo clippy -p spark-model -p spark-server --tests --features cuda,nccl` — clean, **0 warnings emitted**
- [x] Tested against real hardware — H100 microtest, ladder, coherency gate and 16-way probe above (2026-09-11)
- [n/a] PTX gate on `w8a16_gemv_batch4.cu` — **this PR changes no kernel source.** `git diff 793384e03.. -- kernels/` is empty, and all four entry points (`w8a16_gemv_batch4`, `w8a16_gemv_batch16`, and the two `_strided` siblings) already exist on the base at `w8a16_gemv_batch4.cu:213/228/255/270`. #986 carries the PTX identity receipt for that file at `sm_90a` and `sm_121a`; there is nothing here to re-gate. The one new thing is a *registration* (`try_kernel("w8a16_gemv_batch4", "w8a16_gemv_batch16")`), which is why the H100 build's kernel count moved 188 → 189.

## GB10 perf-gate records

The default-off inversion **shrinks** what that re-measure has to cover, and this is the reviewer's catch: with `ATLAS_FFN_BATCH16` unset the GB10 dense-FFN dispatch is byte-for-byte pre-#927, so the FFN tier can no longer change GB10 numbers without someone asking for it. The o_proj, QKV and SSM rungs **do** still engage on GB10 whenever a serve decodes 5..32 native-FP8 rows — they are bit-exact per row, but they are still a routing change and the re-measure is their first GB10 reading. The LM-head band stays 8 there.

## Conflicts

Cherry-picked from `fix/decode-batch16-gemv` on the fork with `-x` provenance lines. Every resolution:

1. **Cherry-picking onto #986's branch alone did not apply.** `dense_ffn.rs` conflicted in `w8_gemm!`: this commit's macro rewrite threads #987's `$a8` / `$cap` parameters and adds a `w8a8_reachable` guard against #987's `prefill_w8a8_selected`, none of which exist without #987. Resolving that by hand would have meant *rewriting* the change rather than applying it. **#987's three commits were cherry-picked into the base first**, after which `dense_ffn.rs` applied with **no conflict at all** and nothing in it was adapted.
2. **`crates/spark-model/Cargo.toml`, twice** — both times the `[[example]]` list, and both times a pure union. #987 adds `native_fp8_ffn_w8a8_microtest`, #984/#986 add `native_fp8_oproj_batch_microtest` and `native_fp8_qkv_batch_microtest`, this commit adds `native_fp8_ffn_batch16_microtest`; git could not line up four sibling stanzas. All four are present exactly once, with identical `required-features = ["cuda", "gpu-examples"]`. No stanza was dropped, edited or renamed — only ordered.

Two files differ from the source branch's tree, both expected and neither this PR's content: `Cargo.toml` (the ordering above) and `o_proj_tests.rs` (which carries #984's own `Adapt the o_proj dispatch test to main's decode_step context field` commit, already in the base). Everything else in the PR's own diff is byte-identical to the source.

## CLA

### DIFF
--- a/crates/spark-model/Cargo.toml
+++ b/crates/spark-model/Cargo.toml
@@ -281,6 +281,10 @@ name = "w4a16_qg_batch_bitparity_microtest"
 required-features = ["cuda", "gpu-examples"]
 
 [[example]]
+name = "native_fp8_ffn_batch16_microtest"
+required-features = ["cuda", "gpu-examples"]
+
+[[example]]
 name = "native_fp8_ffn_w8a8_microtest"
 required-features = ["cuda", "gpu-examples"]
 
--- a/crates/spark-model/Cargo.toml
+++ b/crates/spark-model/Cargo.toml
@@ -281,6 +281,10 @@ name = "w4a16_qg_batch_bitparity_microtest"
 required-features = ["cuda", "gpu-examples"]
 
 [[example]]
+name = "native_fp8_ffn_w8a8_microtest"
+required-features = ["cuda", "gpu-examples"]
+
+[[example]]
 name = "native_fp8_oproj_batch_microtest"
 required-features = ["cuda", "gpu-examples"]
 
--- a/crates/spark-model/Cargo.toml
+++ b/crates/spark-model/Cargo.toml
@@ -285,12 +285,16 @@ name = "native_fp8_ffn_batch16_microtest"
 required-features = ["cuda", "gpu-examples"]
 
 [[example]]
-name = "native_fp8_ffn_w8a8_microtest"
+name = "native_fp8_oproj_batch_microtest"
 required-features = ["cuda", "gpu-examples"]
 
 [[example]]
-name = "native_fp8_oproj_batch_microtest"
+name = "native_fp8_ffn_w8a8_microtest"
 required-features = ["cuda", "gpu-examples"]
+# The cuBLASLt-vs-kernel acceptance rule is CPU-testable; see the `tests`
+# module in `examples/common/native_fp8_bf16_compare.rs`, which the example
+# pulls in with `#[path]`.
+test = true
 
 [[example]]
 name = "native_fp8_qkv_batch_microtest"
--- a/crates/spark-model/examples/common/native_fp8_bf16_compare.rs
+++ b/crates/spark-model/examples/common/native_fp8_bf16_compare.rs
@@ -1,0 +1,311 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! Split out of `native_fp8_ffn_w8a8_microtest.rs` to keep it under the 500-LoC
+//! cap. The rule and its comparator are the half of that example that needs no
+//! GPU, which is also the half worth testing on its own — moving them here puts
+//! the CPU tests beside the code they pin.
+//!
+//! Test-only harness code; no serving path runs any of it.
+
+use half::bf16;
+
+/// cuBLASLt-vs-kernel acceptance. The two implementations consume the SAME FP8
+/// bytes and the SAME FP32 scales, so the only licensed difference between them
+/// is the ORDER of the FP32 accumulation (tile shape, split-K, epilogue
+/// association). That is a real difference and it has a bounded consequence:
+/// each FP32 partial sum moves by a few parts in 1e7, which is far under one
+/// BF16 step (2^-8 relative), so an output lands on a different BF16 value only
+/// when the exact sum sits within that sliver of a rounding boundary — and then
+/// it moves by exactly ONE step, never two. A layout bug does not look like
+/// that; it permutes scales, which moves elements by whole factors.
+///
+/// Hence the acceptance below is a **tolerance on every element**, not a
+/// bit-identity check:
+///
+///   * `|a - b| <= 1 BF16 ULP at max(|a|, |b|)` — one rounding step, measured
+///     at the larger operand's magnitude so a pair straddling a binade boundary
+///     is judged by the coarser grid it actually shares.
+///   * OR both values are under [`CUBLAS_SMALL_MAGNITUDE`]. Near zero the
+///     output is what is left after cancellation between O(K) terms of much
+///     larger magnitude, so its remaining bits — including its SIGN — are set
+///     by accumulation order and carry no information. 0.05 is ~1.7x the
+///     largest sign-flipped magnitude measured on H100 (0.0148, 0.0167, 0.0237,
+///     0.0297 on 2026-09-11) and ~2e-4 of the shapes' output range, i.e. below
+///     anything the downstream SiLU/BF16 store can distinguish.
+///
+/// The count of elements outside that bound is the gate and must be **0**.
+/// `sign_flips`, `unequal_bf16` and `max_ulp` are REPORTED, not gated.
+///
+/// WHY `max_ulp` IS NO LONGER A GATE. The previous bound was `max_ulp <= 2` on
+/// an ORDINAL ULP distance — `|ord(a) - ord(b)|` over sign-magnitude BF16 — and
+/// that metric scores a sign flip on a value of magnitude `v` as `2 * ord(v)`,
+/// a number in the tens of thousands however tiny `v` is. The 2026-09-11 H100
+/// run measured exactly that failure mode: cosine 0.9999996, rel_rms
+/// 9.1e-4-9.3e-4, `max_abs` equal to one BF16 ULP at the largest outputs
+/// (1.000 and 2.000), and `max_ulp ~= 31 000` decoding to sign flips at
+/// |v| <= 0.03. A bound no correctly-ordered FP32 accumulation can satisfy is
+/// a bit-identity gate wearing a tolerance's clothes; this is the tolerance it
+/// was pretending to be.
+///
+/// `rel_rms` moves 1e-3 -> 2e-3 for headroom over the same run's 9.29e-4 while
+/// staying an order of magnitude under the ~2.5e-2 E4M3 floor the W8A16
+/// comparison sits on — which is what keeps these a LAYOUT test and not a
+/// precision test. The row-major control fails all three: rel_rms 1.1e-2-8.8e-2,
+/// cosine down to 0.9961, and `max_abs` 55.6 against a one-ULP bound of 2.
+pub(crate) const CUBLAS_SMALL_MAGNITUDE: f64 = 0.05;
+pub(crate) const CUBLAS_COSINE_GATE: f64 = 0.99999;
+pub(crate) const CUBLAS_REL_RMS_GATE: f64 = 2e-3;
+
+fn to_f64(bits: &[u16]) -> Vec<f64> {
+    bits.iter()
+        .map(|b| bf16::from_bits(*b).to_f32() as f64)
+        .collect()
+}
+
+/// BF16 bits → a monotonically ordered integer, so `|ord(a) - ord(b)|` is the
+/// ULP distance (the standard sign-magnitude → two's-complement remap).
+fn ord(bits: u16) -> i32 {
+    if bits & 0x8000 != 0 {
+        -((bits & 0x7fff) as i32)
+    } else {
+        bits as i32
+    }
+}
+
+/// One BF16 ULP at `bits`' magnitude — the gap between it and its neighbour.
+/// BF16 stores 7 mantissa bits, so a normal value with unbiased exponent `e`
+/// steps by `2^(e - 7)`; every subnormal steps by the smallest normal's step,
+/// `2^-133`. Inf/NaN report an infinite step so nothing is judged "close" to
+/// them by arithmetic accident (the `<=` below still rejects a NaN difference).
+fn bf16_ulp(bits: u16) -> f64 {
+    let biased_exp = ((bits >> 7) & 0xff) as i32;
+    match biased_exp {
+        0xff => f64::INFINITY,
+        0 => (-133.0_f64).exp2(),
+        e => ((e - 127 - 7) as f64).exp2(),
+    }
+}
+
+/// The per-element acceptance described on [`CUBLAS_SMALL_MAGNITUDE`]: one BF16
+/// rounding step at the larger magnitude, or both values under the small-value
+/// escape.
+fn within_one_bf16_ulp(a_bits: u16, b_bits: u16) -> bool {
+    let a = bf16::from_bits(a_bits).to_f32() as f64;
+    let b = bf16::from_bits(b_bits).to_f32() as f64;
+    if a.abs() < CUBLAS_SMALL_MAGNITUDE && b.abs() < CUBLAS_SMALL_MAGNITUDE {
+        return true;
+    }
+    let ulp = if a.abs() >= b.abs() {
+        bf16_ulp(a_bits)
+    } else {
+        bf16_ulp(b_bits)
+    };
+    (a - b).abs() <= ulp
+}
+
+/// Opposite signs with neither value zero. Reported for diagnosis: on these
+/// shapes every flip seen so far has been a cancellation residue under the
+/// small-magnitude escape, and a flip on a LARGE output would fail the
+/// one-ULP bound anyway.
+fn is_sign_flip(a_bits: u16, b_bits: u16) -> bool {
+    let a = bf16::from_bits(a_bits).to_f32();
+    let b = bf16::from_bits(b_bits).to_f32();
+    a != 0.0 && b != 0.0 && (a < 0.0) != (b < 0.0)
+}
+
+pub(crate) struct Compare {
+    pub(crate) max_abs: f64,
+    pub(crate) cosine: f64,
+    pub(crate) rel_rms: f64,
+    pub(crate) max_ulp: i32,
+    pub(crate) unequal: usize,
+    /// Elements failing [`within_one_bf16_ulp`] — the cuBLASLt-vs-kernel gate.
+    pub(crate) over_bound: usize,
+    pub(crate) sign_flips: usize,
+}
+
+pub(crate) fn compare(a_bits: &[u16], b_bits: &[u16]) -> Compare {
+    let (a, b) = (to_f64(a_bits), to_f64(b_bits));
+    let (mut dot, mut na, mut nb, mut max_abs, mut sq_diff, mut sq_ref) =
+        (0.0, 0.0, 0.0, 0.0_f64, 0.0, 0.0);
+    for (x, y) in a.iter().zip(&b) {
+        dot += x * y;
+        na += x * x;
+        nb += y * y;
+        max_abs = max_abs.max((x - y).abs());
+        sq_diff += (x - y) * (x - y);
+        sq_ref += y * y;
+    }
+    Compare {
+        max_abs,
+        cosine: dot / (na.sqrt() * nb.sqrt()),
+        rel_rms: (sq_diff / sq_ref.max(f64::MIN_POSITIVE)).sqrt(),
+        max_ulp: a_bits
+            .iter()
+            .zip(b_bits)
+            .map(|(x, y)| (ord(*x) - ord(*y)).abs())
+            .max()
+            .unwrap_or(0),
+        unequal: a_bits.iter().zip(b_bits).filter(|(x, y)| x != y).count(),
+        over_bound: a_bits
+            .iter()
+            .zip(b_bits)
+            .filter(|(x, y)| !within_one_bf16_ulp(**x, **y))
+            .count(),
+        sign_flips: a_bits
+            .iter()
+            .zip(b_bits)
+            .filter(|(x, y)| is_sign_flip(**x, **y))
+            .count(),
+    }
+}
+
+/// CPU tests for the cuBLASLt-vs-kernel acceptance. They need no GPU, no
+/// kernels and no model: the point of pulling the bound out into
+/// [`within_one_bf16_ulp`] is that the rule can be pinned on synthetic vectors
+/// instead of only being exercised by the H100 run it gates.
+///
+/// Enabled by `test = true` on this example's `[[example]]` stanza, so
+/// `cargo test -p spark-model --example native_fp8_ffn_w8a8_microtest
+/// --features cuda,gpu-examples` runs them.
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// The BF16 value one representable step above `x`, by construction rather
+    /// than by arithmetic: increment the significand of the stored bits.
+    fn next_bf16_up(x: f32) -> u16 {
+        let b = bf16::from_f32(x).to_bits();
+        assert!(b & 0x8000 == 0, "helper is for positive values");
+        b + 1
+    }
+
+    #[test]
+    fn ulp_is_the_gap_to_the_next_bf16() {
+        // 1.0 sits at the bottom of its binade: 7 stored mantissa bits -> 2^-7.
+        let one = bf16::from_f32(1.0).to_bits();
+        assert_eq!(bf16_ulp(one), 2.0_f64.powi(-7));
+        // And the step really is the distance to the neighbour.
+        let up = bf16::from_bits(next_bf16_up(1.0)).to_f32() as f64;
+        assert!((up - 1.0 - bf16_ulp(one)).abs() < 1e-12);
+        // The H100 run's max_abs values are exactly one ULP at their magnitude:
+        // 1.000 in [128, 256) and 2.000 in [256, 512).
+        assert_eq!(bf16_ulp(bf16::from_f32(200.0).to_bits()), 1.0);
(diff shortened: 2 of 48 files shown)


TheTom:
H100 A/B receipt for this PR's FFN tier (2026-09-11, round 5, 1×H100 80 GB, Qwen/Qwen3.8-27B-FP8 rev 017b9c7a, native FP8, batch 16, `--ssm-decode-ring-slots 1 --ssm-batched-recurrent true`, `ATLAS_LM_HEAD_BATCHM_MAX=16`, temp 0, 1 warmup + 3 reps, same binary `e739f6ba…` for both legs):

| shape | metric | tier ON (default) | tier OFF (`ATLAS_FFN_NO_BATCH16=1`) |
|---|---|---:|---:|
| 1193/256 | C=16 agg tok/s | 121.4 | **128.0** (+5.4%) |
| 1193/256 | C=16 TPOT p50 | 107.4 ms | **102.0 ms** |
| 4593/512 | C=16 agg tok/s | 99.0 | **102.1** (+3.1%) |
| 4593/512 | C=16 TPOT p50 | 123.1 ms | **118.5 ms** |
| 28-token smoke | TTFT | 150 ms | **101 ms** |

Per-phase (`ATLAS_MS_PROFILE`, n=16): ssm 63.3 → 59.9 ms, attn 19.9 → 18.8 ms with the tier off; head unchanged. So on Hopper the batch16 GEMV is slower than the tile GEMM it replaces at M=16 (the microtest's 342/270 GB/s is FP32-FMA-bound, ~10% of HBM), and the tier also intercepts the 17–25-row tail prefill chunk. The bit-exactness and the wrapper/handle pairing fix stand; the throughput claim in the body does not on this GPU (round 4's +7.4% was confounded with the head-band lever). Suggest keeping the tier but defaulting it off for `hopper`/`b200` targets (or via HARDWARE.toml), and re-measuring on GB10 where the bandwidth ratio is different. A tensor-core 16-row-tile W8A16 GEMM (`w8a16_gemm_m16`, on the campaign branch) is the follow-up for this width.

TheTom:
Good catch, and it was the right one to make — thank you. `w8a16_gemv_batch16` resolves on GB10 as well as H100, so the presence-keyed kill switch made this tier default-on for a target where it has never been measured, on the strength of an H100 number that came out **negative** once it was measured properly. `8db757c77` inverts the lever: `ATLAS_FFN_NO_BATCH16` (presence, default on) becomes `ATLAS_FFN_BATCH16=1` (value, **default OFF on every target**), latched once per layer into `DenseFfnLayer::batch16_enabled` so the route is fixed before any CUDA-graph capture and the dispatch tests can pin both polarities — `stock_serve_is_the_pre_927_routing` now drives `forward_prefill` with the handle loaded and the tier unarmed and asserts the M-padded tile GEMM across 5..=32. I have also rewritten the PR body: the **−15.5% step** headline is withdrawn (it was a two-lever, cross-binary round-4 comparison) and replaced with the round-5 single-variable A/B — same binary, same burst, the tier the only difference: C=16 aggregate **121.4 → 128.0 tok/s** and TPOT p50 **107.4 → 102.0 ms** with the tier OFF, plus a **150 → 101 ms** smoke TTFT step from the chunked-prefill tail chunk landing in the 5..=32 band. Round 4's `attn` −41.3% attribution is reported as **refuted**, not reassigned: disarming the tier moves `attn` further down (19.90 → 18.82 ms), not back up. On your second question — the o_proj groups-of-16, the QKV band widening and the SSM verify arms are **independent** of this lever (they key off their own kernel handles and never read the switch, and they were armed in both halves of that A/B), so they stay default-on and none of the table's movement, the `attn` −5.4% included, is claimed for them; the file says so explicitly. Gates on Spark 1 at `8db757c77`, isolated worktree: `cargo test -p spark-model --features cuda` for `ffn` **35 passed** (was 33, +2 default-pinning tests), `batch16` **18 passed** (was 16), `o_proj` **10**, `qkv` **19**, `lm_head` **9**, all 0 failed; `cargo clippy -p spark-model --tests --features cuda -- -D warnings` clean; `cargo fmt --check` clean.

tbraun96:

* Of the **97 identifiers this PR introduces that `main` does not have**, **92** are present on #895.
* The five names not found are one function and four test names. `ffn_batch16_enabled()` was **renamed** `ffn_batch16_tier()` on #895 when the lever moved from an env read to `[defaults] ffn_batch16_tier`; the module `dense_ffn_batch16_decode.rs` (absent on `main`) is present on both. Coverage is not lost: `dense_ffn_batch16_decode_tests.rs` has **13** `#[test]` on this PR and **13** on #895.

TheTom:

tbraun96:

TheTom:
Merged base #989 forward (merge commit `bc55e4255`).

Conflicts (7 files) and the rule applied to each:

* `dense_ffn_w8a8_prefill.rs`, `dense_ffn_w8a8_prefill_tests.rs`, `examples/native_fp8_ffn_w8a8_microtest.rs`, `ops/dispatch_proj.rs`, `spark-runtime/src/buffers/accessors.rs`, `spark-runtime/src/buffers/sizes.rs` — **took the base's (#989's) side**. #988's side of each is byte-identical to #987's original, so #988 carries no hunk of its own here; #989's VEC128 scale-layout work is the only real change.
* `dense_ffn.rs` — both sides changed it. **Took #989's side and re-applied #988's own patch verbatim** (the #927 batch16 rungs, the `w8a8_reachable` gate, the `batch16_decode` child module). Verified mechanically that the result equals `#989-side + #988's-own-patch`.
* `crates/spark-model/Cargo.toml` — git auto-merged this one **without reporting a conflict and produced a duplicate `[[example]]` stanza** for `native_fp8_ffn_w8a8_microtest` (#988 places it before the `oproj_batch` stanza, #989 after). Kept the single copy the measured tree carries, the one with `test = true`. Worth flagging: a silent duplicate is the failure mode to watch for on the rest of this stack.

PR's own diff unchanged (verified): of the 32 files that differ from #988's old head, 31 are exactly #989's side and the 32nd is the `dense_ffn.rs` union; `dense_ffn_batch16_decode.rs` and its tests are untouched.

Distance from the measured tree (`4f00cf199`) across the 8 conflicted files: **+187 / −839** (the insertions are #927 content the integration tree carries in a later form).

Gates on Spark 1 (GB10, CUDA 13): fmt ok · clippy `-p spark-model` / `-p spark-runtime` (`--features cuda`, `-D warnings`) ok · `cargo test -p spark-model --lib` 945 passed 0 failed · `-p spark-runtime --lib` 301 passed 0 failed · `-p atlas-kernels --no-default-features` ok · CI file-size cap ok.

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1031 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1031; this fork PR is closed.

PR #989, opened 2026-09-11
spark-runtime: feed cuBLASLt the documented VEC128 scale layout and scope ATLAS_CUBLAS_GEMM per projection family (#917)

TheTom:
## Summary

#987 landed the dense FFN's W8A8 block-scaled prefill and shipped its cuBLASLt arm **disabled and documented as wrong**: 1,140 TFLOP/s at gate/up M=1193 on an H100, and rel_rms 7.7e-2–8.8e-2 against the same quantized inputs the in-tree kernel consumed. This is the PR that makes that arm correct, and then makes it reachable.

Two defects, one per source of the failure:

1. **The activation scales were handed to cuBLASLt in the wrong order.** cuBLASLt reads VEC128 factors with the **token** index contiguous; `per_token_group_quant_fp8` writes them with the **K-group** index contiguous. The buffer was passed untransposed, so cuBLASLt read a permutation of plausible per-token scales — which is why it looked like precision loss rather than garbage, and why it got worse as M grew.
2. **`ATLAS_CUBLAS_GEMM` was one global boolean.** The variable that armed the dense-FFN arm also armed the SSM `in_proj_qkvz` arm, which routed through a cached FP8→BF16 weight dequant allocating 160 MiB per layer outside the buffer ledger — ~10.3 GiB across 48 SSM layers. On an 80 GB H100 that killed the serve at layer 36 before the FFN arm under test could be exercised end to end, and no knob separated the two.

With both fixed, the arm serves: memory flat to the MiB across requests, coherency 4/4, and **C=1 TTFT 743.6 → 521.5 ms** on the 1193/256 shape and **2,322.7 → 1,474.9 ms** on 4593/512, with C=1 TPOT unchanged to two decimal places as the control.

A fifth commit rewrites the microtest's cuBLASLt-vs-kernel acceptance from a bit-identity check into a tolerance. That is a gate change and it gets its own argument below, in its own section, rather than being folded into the numbers it lets through.

Refs #917 and #928.

**Stacked on #987** (`pr/ffn-w8a8-prefill`), which this needs for `cublas_fp8_proj_prequant` and the FFN arm itself. The base is `main`, so the diff here contains #987's three commits as well. **Review this PR's own five with:**

```
git diff pr/ffn-w8a8-prefill..
```

### 1. The VEC128 activation-scale layout

cuBLAS 13.4, "128-element 1D and 128x128 2D Block Scaling For FP8 Data Types" → "Scaling factors layouts", specifies `CUBLASLT_MATMUL_MATRIX_SCALE_VEC128_32F` factors as **M-major for A and N-major for B**, shape `MN x ceil(K/128)` — the **MN** index is the contiguous one. `per_token_group_quant_fp8` writes `a_scale[m * (K/128) + kg]`: K-group contiguous, `[M, K/128]`. That buffer went to cuBLASLt untransposed, i.e. transposed relative to what the API documents.

Measured on 1×H100 (2026-09-11, `native_fp8_ffn_w8a8_microtest`, tip `5f78270dc`, cuBLASLt 13.1), against the in-tree `fp8_gemm_t_blockscaled` on the **same** quantized bytes:

| shape | M | rel_rms vs kernel | ordinal max_ulp | elements differing | cosine vs W8A16 |
|---|---|---|---|---|---|
| gate/up | 64 | 1.12e-2 | 32 591 | 84% | 0.999620 |
| gate/up | 1193 | **7.67e-2** | 33 695 | 87% | **0.996738** ✗ |
| down | 64 | 1.12e-2 | 32 747 | 84% | 0.999618 |
| down | 1193 | **8.82e-2** | 33 675 | 94% | **0.995785** ✗ |

Two correct W8A8 implementations over identical FP8 bytes with FP32 accumulation agree to ~1e-3 and do not degrade with M. These degrade with M by 8×, and at M=1193 cuBLASLt was *further* from the W8A16 reference than the W8A8 kernel was. That is a scale-tensor layout, not arithmetic — a bigger M mixes more mis-indexed scales.

Two things the same doc section **clears** of suspicion, and both are now dispatch clauses rather than assumptions: the BLK128x128 (weight) / VEC128 (activation) mode pairing is listed as supported, and the BLK128x128 factors are K-major with shape `L4 x ceil(N/128)` where the column stride must be a multiple of 4 — which **is** the checkpoint's row-major `[N/128, K/128]` grid so long as `K/128 % 4 == 0`, true for every shape served today.

### 2. The global lever and the off-ledger SSM dequant

`ops::cublas_bf16_proj` caches a BF16 dequant of the FP8 weight, keyed by weight pointer. For the SSM's fused `in_proj_qkvz` (`[10240, 5120]` + `[6144, 5120]`, 2 B/elem) that is **167 772 160 bytes per layer**, allocated outside the buffer arena's ledger. On 1×H100 with `Qwen/Qwen3.8-27B-FP8` at tip `5f78270dc`:

```
one 28-token prefill consumed 6120 MiB and died at layer 36:
  cuMemAlloc_v2 failed: status 2, requested 167772160 bytes
```

Dropping `--gpu-memory-utilization` to 0.84 bought enough headroom to finish the allocation and the serve then answered **garbage** — all four coherency probes failed, with output like `"(with the exception of the following)\n\n(1)\n\n(2)"`, because the FFN arm was still on the row-major scales. So the one variable that could enable #917's fast path also guaranteed the box either died or lied.

## What the fix does

- **`fp8_scale_transpose.cu` / `fp8_act_scale_to_kmajor`** (new, gb10 tree) rewrites the quantizer's `[M, K/128]` into the `[K/128, ceil16(M)]` cuBLASLt reads, **zero-filling the pad rows** the M-pad adds — the phantom tokens' FP8 bytes are already zeroed by #987's caller, and a zero scale keeps their contribution defined rather than reading whatever the scratch held. The quantizer's own output is untouched, because the in-tree kernel indexes it directly. Cost is `M_pad × L` floats — 0.65 MB at the widest Qwen3.8-27B FFN shape against ~89 MB of FP8 weight the same GEMM streams, which is why it is a plain strided-read/coalesced-write pass and not a tiled transpose.
- **`spark_runtime::cublaslt::scale_layout`** is the SSOT for that index math, carries the doc quotes, and is pinned by CPU unit tests — including one that asserts the two readings *disagree* at the measured shape, so the module cannot be quietly turned back into a no-op.
- **`ATLAS_CUBLAS_SCALE_LAYOUT=kmajor|rowmajor`** (default `kmajor`) feeds the old reading on demand, so one box can show both. The microtest labels it `pre-fix control, expected to FAIL`.
- **`ATLAS_CUBLAS_GEMM` becomes a comma-separable family set**: `all|ffn|attn|ssm|head|off`, with `1`/`true` still meaning `all` so existing launch scripts are unchanged. It resolves once into `CublasScope` on `GemmDispatch`, every dispatch site reads **its own field**, unknown tokens are dropped with a warning and **never widen the set** (a typo must not silently arm an arm), and the resolved scope is logged at model build — not lazily on first use, so an operator sees it before any traffic:

```
[atlas] ATLAS_CUBLAS_GEMM="ffn" -> cuBLASLt arms ffn=true attn=false
  ssm=false head=false (head has no consumer yet)
```

  `head` is parsed and carried but has **no consumer yet**, and the log line says so rather than implying one.
- **The SSM cuBLASLt arm no longer dequantizes.** It moves inside the existing W8A8 branch and consumes the FP8 weight directly through #987's `cublas_fp8_proj_prequant` — the same helper and the same k-major VEC128 scale layout as the FFN arm, so one fix covers both. Activation, scales and the k-major transpose are all arena buffers (`fp8_act_scale_kmajor` is new and ledger-owned), and `ssm_qkvz` / `ssm_deinterleaved` now carry the `ceil16(M)` row pad the matmul writes. The arm declines to the in-tree kernel when any precondition is missing, including the output-buffer bounds check. `trait_prefill_proj.rs` now reads `NO cuBLASLt-BF16 ARM HERE, deliberately` at the site the dequant used to live at.
- **Two buffer-sizing tests pin the pad** rather than letting the constant drift — the failure mode if a size were missed is a cross-buffer write during a full-width prefill chunk: silent, shape-dependent, and invisible to a microtest.

### The gate change, argued separately

The microtest accepted the cuBLASLt arm against the kernel at `max_ulp <= 2`, on an **ordinal** BF16 distance (`|ord(a) - ord(b)|` over the sign-magnitude remap). That metric scores a **sign flip** on a value of magnitude *v* as `2·ord(v)` — tens of thousands however small *v* is — so the bound is only reachable by two implementations that agree bit for bit. **The two arms are not required to agree bit for bit.** They consume the same FP8 bytes and the same FP32 scales; what differs is the *order* of the FP32 accumulation, and no correct FP8 GEMM has to reproduce another's tile shape.

The 2026-09-11 k-major run measured exactly that residue: cosine 0.9999996, rel_rms 9.09e-4–9.29e-4, 11% of elements differing by one BF16 rounding step, `max_abs` **1.000** at gate/up and **2.000** at down — which are *precisely* one BF16 ULP at those shapes' largest outputs, not an error — and `max_ulp ≈ 31 000`, which decodes to sign flips at |v| = **0.0148, 0.0167, 0.0237, 0.0297**. The alternative reading (two same-signed values ~242 binades apart) is excluded by `max_abs ≤ 2.0`. Everything the run reports says "one rounding step apart"; the one gate that failed is the one measuring something else.

The acceptance is now **per element**: `|a − b| ≤ one BF16 ULP at max(|a|, |b|)`, **or** both values under **0.05**. The first is the licensed difference — an FP32 partial sum moved by a few parts in 1e7 changes the stored BF16 only when the exact sum sits inside that sliver of a rounding boundary, and then by exactly one step. The second is cancellation: near zero the output is the residue of O(K) much larger terms, so its remaining bits *and its sign* are set by accumulation order and carry no information; 0.05 is ~1.7× the largest flipped magnitude measured and ~2e-4 of these shapes' output range. **The count outside that bound is the gate and must be 0**; `sign_flips`, `unequal_bf16` and the ordinal `max_ulp` are still printed, because the diagnosis of a future failure lives in those numbers. `rel_rms` goes 1e-3 → **2e-3** for headroom over the measured 9.29e-4, still an order of magnitude under the ~2.5e-2 E4M3 floor the W8A16 comparison sits on — which is what keeps this a **layout** test and not a precision test. `cosine >= 0.99999` is unchanged.

**This does not soften the control.** `ATLAS_CUBLAS_SCALE_LAYOUT=rowmajor` is still expected to FAIL and still fails every clause: rel_rms 1.12e-2–8.82e-2 against 2e-3, cosine down to 0.9961 against 0.99999, and `max_abs` 55.625 against a one-ULP bound of 2.0 at that magnitude.

**What is not yet measured, stated plainly:** the round-4 binary carried the *old* gate, so the 0-elements-over-bound figure is an inference from that run's aggregate statistics (max_abs equal to one ULP, rel_rms 9.2e-4, 11% differing), not a direct reading. The new gate owes its own H100 receipt. What *is* directly measured is that every other number it checks already passes and the control still fails.

## Oracle and evidence

All of this is **diagnostic single-H100 evidence, not certified campaign numbers.** 1×H100 80 GB, cuBLASLt 13.1, `Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`, native FP8 profile, `--max-batch-size 16 --ssm-decode-ring-slots 1 --ssm-batched-recurrent true`, temperature 0, 1 warm-up + 3 recorded reps, frozen ladder harness `bench/ladder38/harness_w55_conc_ladder.py`. 2026-09-11.

### Microtest, k-major vs row-major on the same binary and the same box

`native_fp8_ffn_w8a8_microtest`, tip `2962cfed7`, both legs at `ATLAS_W8A8_REL_RMS_GATE=0.03` (#987's reading of the E4M3 floor). Banner: `scale_layout=kmajor [K/128,M_pad] (documented)`.

| shape | M | kernel ms / TFLOP/s | **cuBLASLt ms / TFLOP/s** | vs W8A16 | vs kernel |
|---|---|---|---|---|---|
| gate/up | 64 | 0.170 / 67.15 | 0.050 / **226.82** | 10.40× | 3.38× |
| gate/up | 1193 | 1.441 / 147.56 | 0.187 / **1139.46** | 15.55× | 7.72× |
| down | 64 | 0.447 / 25.53 | 0.047 / **241.97** | 15.62× | 9.48× |
| down | 1193 | 1.453 / 146.33 | 0.213 / **999.29** | 14.90× | 6.83× |

Agreement with the in-tree kernel, with the row-major control's numbers beside it:

| shape | M | elements differing | max_abs | **cosine** | **rel_rms** | rowmajor rel_rms | rowmajor cosine |
| gate/up | 64 | 11% | 1.000 | **0.999999569** | **9.29e-4** | 1.12e-2 | 0.999937 |
| gate/up | 1193 | 11% | 1.000 | **0.999999575** | **9.22e-4** | 7.67e-2 | 0.997055 |
| down | 64 | 11% | 2.000 | **0.999999587** | **9.09e-4** | 1.12e-2 | 0.999937 |
| down | 1193 | 11% | 2.000 | **0.999999577** | **9.21e-4** | 8.82e-2 | 0.996102 |

`vs W8A16` is now **identical to the W8A8 kernel's own** on all four cells (max_abs 5.375/5.500/9.000/9.500, cosine 0.999680/0.999680/0.999679/0.999679, rel_rms 0.0253–0.0254): cuBLASLt and the kernel are the *same answer* against the reference, which is the thing that was untrue before.

| metric | rowmajor | **kmajor** | improvement |
|---|---|---|---|
| rel_rms @ M=1193 gate/up | 7.67e-2 | **9.22e-4** | **83×** |
| rel_rms @ M=1193 down | 8.82e-2 | **9.21e-4** | **96×** |
| cosine @ M=1193 gate/up | 0.997055 | **0.999999575** | residual −99.986% |
| max_abs @ M=1193 gate/up | 55.625 | **1.000** | **56×** |
| elements differing | 84–94% | **11%** | |
| `vs W8A16` cosine gates | 2 failures | **0 failures** | |
| M-dependence | 8× worse at M=1193 | none | |

Throughput is unchanged by the layout fix (1139 vs 1143 TFLOP/s at gate/up M=1193) — it costs one 0.65 MB transpose per GEMM. **The control is the load-bearing part of this table**: it reproduces the pre-fix numbers *to the digit* (32 591 / 33 695 / 32 747 / 33 675 ULP, 1.12e-2 / 7.67e-2 / 1.12e-2 / 8.82e-2 rel_rms), on the new binary, reachable by one environment variable. The improvement is the fix, not drift.

### Serve with `ATLAS_CUBLAS_GEMM=ffn` — tip `c1b67f022`

Same recipe as #987's leg plus `ATLAS_LM_HEAD_BATCHM_MAX=16`; the only script delta is four lines exporting `ATLAS_CUBLAS_GEMM=ffn`. Reserve and KV budget are within 0.02% of the control boot, so nothing below is confounded by a different cache size. **Zero `ERROR` lines in the whole serve log**, and the scoped lever adds no WARN.

**Memory — the leak is gone.**

| point | `memory.used` | Δ |
|---|---|---|
| ready, no request yet | 74 813 MiB | — |
| after request 1 (28-tok smoke) | 74 813 MiB | **0** |
| after request 2 (~20 tok) | 74 813 MiB | **0** |
| after request 3 (2103-tok prompt) | 74 813 MiB | **0** |
| after request 4 (2263-tok prompt) | 74 813 MiB | **0** |
| after the entire 4-shape ladder (16 reps, C=16, 4593-tok prompts) | 74 841 MiB | +28 MiB total |

Flat **to the MiB** across four requests, against 6,120 MiB consumed by a single 28-token prefill before. The +28 MiB across a nine-minute ladder running 16 concurrent 4593-token sequences is paged-KV bookkeeping, not a per-request allocation.

**The prefill route line says cuBLASLt**, which is the first time the arm has been reached by a serving request:

```
spark_model::layers::dense_ffn::w8a8_prefill:
  [atlas] dense FFN prefill: W8A8 block-scaled via cuBLASLt (per-token 1x128
  act scales x 128x128 weight scales, FP32 epilogue; vLLM-equivalent FP8
  numerics). ATLAS_FFN_W8A16_ONLY=1 restores W8A16.
```

No `log_w8a16_prefill_route` line accompanies it, so it is not a fallback.

**Coherency — `COH_EXIT=0`, `passed: true`, 4/4.** `determinism_ok`, `toolcall_ok` (`get_weather` with `{"city":"Reykjavik","days":3}`, `finish_reason: tool_calls`), `think_leak_ok`, and `known_answer_ok` with all three sub-probes OK including the word-reversal one that fails on vLLM too (#929). The pre-fix serve failed **all four**. A 16-way concurrent arithmetic probe was **16/16** correct in 3.22 s.

**Ladder** — `LADDER_EXIT=0`, zero errors in all 12 recorded reps, every `finish_reason: length`, rep spread 0.02–0.14%. Against the same binary and recipe with the arm off:

| shape | metric | arm off | **`ATLAS_CUBLAS_GEMM=ffn`** | Δ |
|---|---|---|---|---|
| 1193 / 256 | **C=1 TTFT p50** | 743.6 ms | **521.5 ms** | **−29.9%** |
| 1193 / 256 | C=1 tok/s | 41.49 | **43.07** | +3.8% |
| 1193 / 256 | **C=1 TPOT** | 21.27 ms | **21.25 ms** | **−0.09% (control)** |
| 1193 / 256 | C=16 agg | 121.42 | **135.80** | **+11.8%** |
| 1193 / 256 | C=16 TTFT p50 | 6 329.2 ms | **4 444.5 ms** | −29.8% |
| 4593 / 512 | **C=1 TTFT p50** | 2 322.7 ms | **1 474.9 ms** | **−36.5%** |
| 4593 / 512 | C=1 tok/s | 35.12 | **37.28** | +6.2% |
| 4593 / 512 | **C=1 TPOT** | 23.99 ms | **23.99 ms** | **0.00% (control)** |
| 4593 / 512 | C=16 agg | 99.03 | **118.62** | **+19.8%** |
| 4593 / 512 | C=16 TTFT p50 | 19 779.9 ms | **12 570.8 ms** | −36.4% |

**The shape is exactly right for a prefill-only lever, which is why it is worth trusting.** The dense-FFN cuBLASLt arm is a prefill route, so single-stream decode is the control this experiment needed: C=1 TPOT moves 0.02 ms on the short shape and **0.00 ms** on the long. Everything that moves, moves in TTFT, by ~30% on the short prompt and ~36% on the long — the right sign and the right length-dependence for replacing a 146–148 TFLOP/s kernel with a 999–1139 TFLOP/s one. The C=16 aggregate gains exceed the C=16 TPOT gains and that is not double-counting: cutting prefill by a third both shortens the concurrent ramp and removes prefill/decode interference during it.

The single-stream TTFT gap to vLLM closes from 2.59× to **1.82×** on the short shape and from 6.03× to **3.83×** on the long one.

**What this evidence does not establish.** One box, one card, one model, one session. It is not a certification run, it says nothing about GB10, and the new microtest tolerance has not itself been run on hardware (see the gate section). The cuBLASLt arm remains **off by default** — `ATLAS_CUBLAS_GEMM` unset is the empty set — so nothing here changes what a serve does unless an operator asks for it.

## Test plan

All on Spark 1 (GB10), CPU-only, `ATLAS_SKIP_BUILD=1`, `nice`, `--jobs 8`, against this branch fetched from the fork at `7cdb0c172`.

- [x] `cargo test -p spark-runtime --lib --features cuda cublas` — **4 passed, 0 failed** (`kmajor_adapter_transposes_and_zero_fills_the_pad`, `the_two_readings_disagree_at_the_measured_shape`, `documented_extents_match_the_shapes_we_serve`, `adapter_is_identity_free_only_when_one_dimension_is_trivial`)
- [x] `cargo test -p spark-model --features cuda w8a8` — **19 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda cublas` — **6 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda dispatch` — **44 passed, 0 failed**, including the lever grammar table, `unknown_families_are_dropped_and_reported_never_widening_the_set`, and the two source-scan tests `the_global_cublas_gemm_boolean_has_no_readers_left` / `each_cublas_family_is_read_only_from_its_own_projection_area`
- [x] `cargo test -p spark-model --features cuda ssm` — **141 passed, 0 failed**
- [x] `cargo build -p spark-model --example native_fp8_ffn_w8a8_microtest --features cuda,gpu-examples` — clean, `Finished dev profile`
- [x] `cargo test -p spark-model --example native_fp8_ffn_w8a8_microtest --features cuda,gpu-examples` — **8 passed, 0 failed** (the new CPU tests for the tolerance)
- [x] `cargo clippy -p spark-model -p spark-runtime -p spark-server --tests --features cuda,nccl` — clean, **0 warnings emitted**
- [x] PTX gate on the new `kernels/gb10/common/fp8_scale_transpose.cu`, **both arches**, `nvcc --ptx` and `-cubin -Xptxas -v`:

```
sm_121a  fp8_act_scale_to_kmajor: 0 bytes stack frame, 0 bytes spill stores,
         0 bytes spill loads, 14 registers, 0 barriers
sm_90a   fp8_act_scale_to_kmajor: 0 bytes stack frame, 0 bytes spill stores,
         0 bytes spill loads, 14 registers, 0 barriers
```

  One `.visible .entry fp8_act_scale_to_kmajor` on both, identical resource usage — it is an index rewrite with one predicated load, so the arch delta being zero is the expected result rather than a lucky one.
- [x] Tested against real hardware — the H100 receipts above (2026-09-11). The **new microtest tolerance** is the exception and is called out in its own section.

**The CPU tests for the tolerance.** `within_one_bf16_ulp`, `bf16_ulp` and `is_sign_flip` are pure functions so the rule can be pinned on synthetic vectors instead of only being exercised by the H100 run it gates: one-ULP pairs accepted at six magnitudes and two-ULP pairs rejected at the same six; the four *measured* sign-flip magnitudes accepted and a flip at 0.06 and at 200.0 rejected; `bf16_ulp` checked against the actual distance to the next representable BF16, and against the run's `max_abs` values (1.000 at 200.0, 2.000 at 400.0); NaN never accepted; a binade-straddling pair judged at the coarser grid, with the deliberate cost of that choice asserted rather than hidden; and `compare()` counting exactly one over-bound element and one sign flip in a hand-built four-element vector. `test = true` on the `[[example]]` stanza is what runs them.

## GB10 perf-gate records

This diff touches `crates/`, a PERF_PATH, so `record_covers` invalidates every committed `.benchmarks/` record on landing. A GB10 re-measure is owed before merge. **By default the GB10 dispatch is unchanged**: `ATLAS_CUBLAS_GEMM` unset resolves to the empty set and no cuBLASLt arm is taken, so the re-measure should reproduce #987's numbers — but it is not a formality, because the new `fp8_act_scale_kmajor` arena buffer and the `ceil16(M)` pad on `ssm_qkvz` / `ssm_deinterleaved` change the arena's size on every target, armed or not. The H100 ladder above is diagnostic evidence from a different card and does not substitute for it.

## Conflicts

Cherry-picked from `fix/cublaslt-vec128-scale-layout` and `fix/cublas-lever-scoping-and-ssm-leak` on the fork with `-x` provenance lines, onto #987's branch. **No conflicts** — all four applied to identical pre-images and nothing was adapted. Verified by diffing this PR's own range against the source branches' equivalent range: the only differences are the hunk line offsets in `dense_ffn.rs` and `qwen3_ssm/tests.rs` (the source branch also carries #988's batch-16 commits, which are not in this stack) and the deletion below. Every hunk body is byte-identical.

**One deliberate deletion.** `2962cfed7` adds `fp8_scale_transpose.cu` to three kernel trees: the real file under `kernels/gb10/common/` and two symlinks to it under `kernels/hopper/common/` and `kernels/b200/common/`. **`main` has neither of those trees** — it carries `gb10`, `metal`, `strix`, `strix-hip`, and the two symlinks were the only entries under the new directories, so applying them would have created two hardware trees as a side effect of a kernel addition. Both symlinks were dropped; the gb10 file is unchanged. `python3 scripts/check_kernel_shadows.py` passes and still reports exactly the four trees `main` has. The symlinks belong with whatever introduces those trees, not here.

## CLA


tbraun96:

**Evidence of containment** (measured 2026-09-11 against `pr895` = `762fce294`, `avarok/main`, `pr989`):

* Of the 34 files this PR changes relative to `main`, **0** are still byte-identical to `main` on #895 — #895 has independently modified every one.
* The symbols that are *unique to this PR* (absent from #987, which this PR builds on) are all present on #895, `git grep -c ... -- crates/`:

  | symbol | #987 | #989 | #895 |
  |---|---|---|---|
  | `fp8_act_scale_to_kmajor` | 0 | 10 | **17** |
  | `blk128x128_stride_ok` | 0 | 5 | **9** |
  | `ffn_act_scale_kmajor` | 0 | 7 | **7** |
  | `cublas_fp8_proj_prequant` | 4 | 6 | **13** |

  Zero on #987 and non-zero on #895 is the decisive column: #895 cannot have acquired these from #987, so it carries **this** PR's contribution.

* `crates/spark-model/src/layers/dense_ffn_w8a8_prefill.rs` on #895 is #987's module with this PR's VEC128 `kmajor` adapter applied — including the `scale_layout_ready` guard, the `blk128x128_stride_ok(k)` clause and the widened `cublas_fp8_proj_prequant` signature.

A trial merge into #895 conflicts in 11 files, every one of them a file where #895 holds a newer version of the same code.

The correctness argument this PR makes stands and is worth restating for the record, since it is the reason the cuBLASLt arm must not ship without the transpose: cuBLASLt reads the VEC128 activation scales token-contiguous, so without `fp8_act_scale_to_kmajor` the GEMM is fast and **wrong** — 1140 TFLOP/s at rel_rms 7.7e-2 against the in-tree kernel's 2.5e-2 (H100, 2026-09-11).

TheTom:

tbraun96:

TheTom:
Merged base #987 forward (merge commit `128c6d428`). #989 already contained #987's old head, so this brings #986's layer in through it.

Conflict: `crates/spark-model/Cargo.toml` — #989 adds `test = true` to the `native_fp8_ffn_w8a8_microtest` stanza while #986 appends the `qkv_batch` stanza right after it. Resolved to the **union**, which is byte-identical to `hopper/h100-integration-2026-09-10` for that region.

PR's own diff unchanged (verified): every one of #989's 31 non-`Cargo.toml` files byte-identical to `e5f92d1a9`.

Distance from the measured tree (`4f00cf199`) on that file: **+0 / −53** — a strict subset.

Gates on Spark 1 (GB10, CUDA 13): fmt ok · clippy `-p spark-model` / `-p spark-runtime` (`--features cuda`, `-D warnings`) ok · `cargo test -p spark-model --lib` 923 passed 0 failed · `-p spark-runtime --lib` 301 passed 0 failed · `-p atlas-kernels --no-default-features` ok · CI file-size cap ok.

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1030 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1030; this fork PR is closed.

Issue #916, opened 2026-09-06
ATLAS_DENSE_FP8=1 still computes in NVFP4 on two paths: multi-row decode (M4-M8) ignores the FP8 overlay, and the LM head is requantized unless --lm-head-dtype bf16


TheTom:
## Summary

`ATLAS_DENSE_FP8=1` is documented as "load the checkpoint's block-scaled FP8 weights and dispatch W8A16 instead of requantizing to NVFP4". Two paths still computed in NVFP4 with the flag set, so a run labeled "native FP8" was not:

1. **Multi-row decode (M4 to M8).** `can_forward_km` accepted the resident NVFP4 fallback weights and `forward_km` read them without consulting the FP8 overlay. M2/M3 already had the format-aware guard. Any concurrent decode with 4+ active sequences silently switched the dense FFN to 4-bit weights.
2. **LM head.** The flag has no effect on `factory/lm_head_setup.rs`; the head is quantized to NVFP4 at load (`LM head quantized to NVFP4 (vocab=248077)`) unless `--lm-head-dtype bf16` is passed explicitly. `--lm-head-dtype fp8` requantizes the checkpoint's BF16 head to per-row FP8, so it does not preserve the original either.

Found by a source audit during the H100 rental while preparing a same-weight-precision comparison for Qwen/Qwen3.8-27B-FP8.

## Fix for (1), on the fork

TheTom/atlas `hopper/sm90-target-tdd-2026-09` commit `b3025ab19d30416a3bbef201f34989bf338e1acd` "spark-model: preserve native FFN weights in multi-row decode": when a native BF16/FP8 overlay is installed, the M4 to M8 entry delegates to the existing format-aware `forward_prefill`. Default NVFP4 behavior is unchanged. Regression test `dense_ffn_native_batch_tests::native_fp8_multi_row_dispatch_preserves_checkpoint_weights` covers M4/M5/M8 with valid fallback pointers and asserts all three projections pick the native kernel (red on the old source: "native M4 dispatched 0 of 3 required FP8 projections"). spark-model suite 647 passed.

The committed fix is a CPU dispatch proof. A GPU kernel trace showing `w8a16_gemm` rather than `w4a16_gemv_batch4` on the M4 path is still owed.

## Proposed for (2)

Either make `ATLAS_DENSE_FP8=1` imply "keep the checkpoint head dtype" (BF16 here), or have the loader log loudly that the head is being requantized while the flag is set. Today the only signal is one INFO line.

[...]

Issue #736, opened 2026-08-23
Teardown sweep reclaims 43.57 GB across 1688 unowned allocations — 5 loader sites need ledger owners


rsafier:
## What

On serve shutdown (qwen3.8-27B-NVFP4 + DFlash2 drafter, GB10, branch `wip/dflash2-full` @ 8eb7fd71), the teardown backstop sweep reports:

```
WARN spark_runtime::cuda_backend: sweep: 1688 allocation(s) totalling 43.57 GB had no owner; largest sites:
  crates/spark-model/src/layers/dense_ffn.rs:653            (9180.0 MB x192)
  crates/spark-model/src/weight_map/quant_helpers.rs:98     (5625.0 MB x65)
  crates/spark-model/src/weight_map/loaders_fp8.rs:210      (5065.7 MB x185)
  crates/spark-model/src/weight_loader/qwen35_dense.rs:1363 (3840.0 MB x48)
  crates/spark-runtime/src/weights/loader/load_fns.rs:200   (3670.5 MB x81)
```

## Why it matters

`sweep_unreleased` (cuda_backend.rs) is documented as the backstop for allocations no `ModelResource` covers, and its own doc says "a non-zero count after a clean teardown is a leak" — the log line was added (2026-08-19 ledger work) precisely so the sites could be tracked down rather than counted. 43.57 GB is roughly the whole weight footprint of this serve: essentially none of the loader-side weight memory has a registered owner.

Consequences beyond hygiene:

1. **The alloc ledger's attribution story has a 43 GB blind spot.** The ledger (size + call site, logged at end of load) is what the util-pledge honesty work leans on; the largest single consumer class on the box is exactly the part that self-attributes as "no owner".
2. **Process teardown is the only thing freeing these.** Anything that tears a model down without exiting the process (model swap, future multi-model serving, error-path reload) would leak the full weight set — the sweep only runs in the full teardown path.
3. **Double-quant staging may be hiding in here.** The `quant_helpers.rs:98` and `load_fns.rs:200` rows may include load-time staging buffers that should have been freed at load end, not carried to teardown — the earlier ledger pass caught a 9.2 GB duplicate FFN repack the same way. Can't tell staging from steady-state today, which is the point of giving them owners.

## Fix direction

Register the five sites' allocations under real owners (layer structs / loader stages) so `ModelResource::release` covers them and the sweep goes back to meaning "leak":

[...]

PR #991, opened 2026-09-11
spark-model: stop building the unreachable NVFP4 sidecars and W8A16 twins on the native-FP8 dense route (#915)

TheTom:
## Summary

On a native-FP8 checkpoint the dense loader builds a full set of **NVFP4 fallback weights and transposed twins that no kernel on that route can reach**. Measured on 1×H100, 2026-09-11, `Qwen/Qwen3.8-27B-FP8`: the checkpoint is **28.75 GB**, but **58.1 GB** was live before the KV cache was sized. The extra 28 GB is derived copies — 18.4 GiB of it provably unreadable — and none of it had an owner, so the teardown sweep reclaimed **28.01 GB across 1,980 allocations** and warned that each was "memory whose owner is unaccounted for".

This stops building the unreachable copies, gives the ones that stay a real owner on the `WeightStore`, and frees the four SSM store tensors the fused `[QKV|Z]` FP8 concat had already consumed. Pre-KV residency falls **58.1 → 33.2 GB**; the KV cache goes 5.2 → 13.6 GB (137,216 → 355,280 tokens) on the same card.

Refs #915, #916, #917 and #736.

**On #736 specifically — this is `Refs`, not `Fixes`.** That issue asks for ledger owners at five named loader sites. This PR removes four of them from existence (`loaders_fp8.rs:229/230`, `quantized.rs:261/262`) and adopts what stays, and the sweep's unowned total falls 28.01 GB → 1.12 GB with **zero** contribution from `load_layers`. But `quant_helpers.rs:98` — one of the five — is still unowned in the after-log (710 MB ×7, the quantizer scratch buffers), and `dense_ffn.rs:653` / `load_fns.rs:200` are not addressed by name. The issue's exact ask is not fully met by this PR alone.

**No arithmetic changes in this PR.** Every GEMM, every scale, every epilogue is untouched. This is a residency change: the same math over the same bytes, with the copies nothing reads no longer allocated.

## What was wrong

`ATLAS_MEM_PROFILE` recorded GPU-free falling a perfectly linear **437 MB per layer across all 64 layers (28.0 GB)** *after* the checkpoint was already resident, so all of it was derived, not loaded. The ledger named six sites:

| site | bytes × count | tensor |
|---|---|---|
| `weight_map/loaders_fp8.rs:229` | 8,960 MB ×256 | `quantize_to_nvfp4` packed `[N,K/2]` |
| `weight_map/quantized.rs:261` | 8,960 MB ×256 | the transposed twin of the same |
| `weight_loader/qwen35_dense.rs:135` | 3,840 MB ×48 | SSM fused `[QKV\|Z]` FP8 concat |
| `weight_map/quantized.rs:643` | 1,600 MB ×64 | attention `Fp8WeightTransposed::weight_t` |
| `weight_map/loaders_fp8.rs:230` | 1,120 MB ×256 | NVFP4 per-16 group scales |
| `weight_map/quantized.rs:262` | 1,120 MB ×256 | the transposed twin of those scales |

The ×256 counts are `64 layers × 3` dense-FFN projections plus `16 attention layers × 4` q/k/v/o — 8,160 + 800 MB, which is exactly the 8,960 reported.

Three families, **two of them unreachable**:

- **NVFP4 gate/up/down + twins, 64 layers — 18.4 GiB.** `DenseFfnLayer::forward` and `forward_prefill_inner` both return from inside their `if let Some(ref fp8w) = self.fp8_weights` arm; `forward_k2`/`k3`/`km` redirect through `native_small_batch_uses_prefill`; and the `w8_gemm!` macro binds `gate_t`/`up_t`/`down_t` to a **literal `None`** on every rung, so even the W8A16 fallback reads the original `[N,K]` E4M3 bytes. Nothing on the route can read an NVFP4 buffer.
- **NVFP4 q/k/v/o + twins + the fused `[q|k|v]` twin, 16 layers — ~2.4 GiB.** `set_fp8_weights` **overwrites** `q/k/v/o_weight`, so these were orphaned the instant they were installed.
- **FP8 prefill twins, 16 layers — 1,600 MB.** Q and O are reached only after the W8A8 arm declines. **K and V are not** — `prefill/cache_skip_qkv.rs` has no W8A8 arm at all and is the first-chunk path of every request. K/V twins stay.

Separately, the native-FP8 GDN arm copies `in_proj_qkv` and `in_proj_z` into one fused device buffer and rebuilds `in_proj_ba` through a host interleave, then leaves all four store originals resident for the life of the model — 80 MiB + ~1 MB per SSM layer, **~3.9 GB across the 48 GDN layers**. That is exactly what the fused copy costs, so the pair was a straight duplicate charged to the KV budget.

## What the fix does

**`fp8_residency.rs`, a new child module** of `qwen35_dense.rs`, is the decision table. The loader runs *before* dispatch exists, so it cannot read a `ForwardContext`; instead it derives the route from **the same resolvers the forward pass uses** — `GemmDispatch::from_env`, the exact constructor `model/impl_a1.rs:836` calls, plus the four un-memoised `ATLAS_ATTN_*` reads the hot arms make. Both are pure functions of the environment and a serve never rewrites its own environment, so the loader's answer and the context's answer cannot disagree. `DenseFp8Plan::resolve` is a pure function of `DenseFp8Inputs`, which is why the CPU tests can pin every clause without touching the process environment.

`ATLAS_FFN_W8A16_ONLY` is deliberately **not** an input: it steers between rungs of the same `w8_gemm!` match, whose transposed operand is that literal `None`, so it selects between two kernels that both read the original FP8 bytes and cannot resurrect an NVFP4 reader.

**`ATLAS_DENSE_FP8_KEEP_NVFP4`** (presence, not `=1` — `…=0` meaning "on" is a trap for an operator reaching for an escape hatch mid-incident) restores the pre-#915 footprint wholesale, for bisecting a suspected gap in that table.

**`dense_ffn.rs`** — `QuantizedWeight::null()` becomes a legal shape for a layer carrying an FP8 overlay. `finalize_q4k_load` and `finalize_nvfp4_mmq_load` return early when `fp8_weights` is installed, for the same reason they already do for packed-Q2: the repack is dead work, and over NULL NVFP4 sources it is a CUDA-700. `finalize_nvfp4_mmq_load` is the sharper case — it is active **by default** wherever its four kernels exist, and it also **frees** the transposed copies on the assumption that the MMQ arm will serve prefill.

`can_forward_km` keeps answering `true` for an FP8 layer. It gates the 4..=8-row verify branch in `multi_seq/ffn.rs:145`; letting a NULL `gate_proj` flip it to `false` would quietly reroute that branch, and **this change must shrink the footprint without moving the dispatch**. `forward_km` redirects to `forward_prefill` for an FP8 layer anyway.

**`DerivedStore` (`spark-runtime/src/weights/derived.rs`)** is the owner for what stays. It hangs off `WeightStore` because these buffers are derived from store tensors, share the store's lifetime, and the store is already the last `ModelResource` released before the sweep — so adopting a buffer moves it from "swept" to "released", which is the difference the ledger reports. Interior mutability (`parking_lot::Mutex`, no poisoning, so a panic elsewhere cannot block teardown) because loaders take `&WeightStore`; threading a `&mut` through `ModelWeightLoader::load_layers` would change every architecture's loader to fix a bookkeeping gap in one of them. `disown()` exists for the transient half of a fused concat — adopted by the generic loader that produced it, freed by the loader that consumed it — because without it that pointer would be freed twice. `release()` drains first, so a failure part-way through cannot leave a freed pointer in the list to be freed again.

Adopted: the SSM fused `[QKV|Z]` FP8 concat (3,840 MB ×48, the hottest tensor on that route, with no un-fused dispatch to fall back to), its block-scale grid, the interleaved BA weight, and every widened FP8 block scale.

**`prune_after_load`** frees the four SSM store names the fused concat consumed, and **only for layers where the arm actually ran**. The arm's condition moves into `gdn_fp8_arm_selected` so the loader and the pruner cannot drift — a drift here is a use-after-free with no diagnostic. `out_proj.weight` is **not** pruned (it *is* `out_proj_fp8w.weight`, bound zero-copy), and neither are conv1d / A_log / dt_bias / norm, which are aliased or not depending on their on-disk dtype.

Load ends with one INFO line — weights / derived / twins / freed / not built — so the next serve log proves the residency without a profiling rerun.

## Oracle and evidence

**Diagnostic single-H100 evidence, not certified campaign numbers.** 1×H100 80 GB, `Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`, native FP8 profile, `--lm-head-dtype bf16`, 2026-09-11. Before = round 3/5 at `c1b67f022`; after = round 6 at integration tip `456109bf3`.

### The residency lines, verbatim

```
native FP8 dense residency: weights 30.87 GB, derived 4.24 GB
  (twins: attn-fp8-t, ssm-qkvz-fp8), freed 0.00 GB, not built 23.31 GB

native FP8 GDN: released 288 store tensors (4.07 GB) consumed by the fused
  [QKV|Z] concat and the BA interleave; out_proj/conv1d/A_log/dt_bias/norm
  kept (still aliased)
```

### Memory — this PR's own receipt

| line | before (`c1b67f022`) | after (`456109bf3`) | Δ |
|---|---|---|---|
| `build residency: load_layers` | **−26.045 GiB** | **−3.996 GiB** | −22.05 GiB |
| `build residency: prune_after_load` | +0.000 GiB | **+3.750 GiB** | the prune now returns memory |
| cumulative after build | 26.738 GiB | **0.939 GiB** | −25.8 GiB |
| **pre-KV** | **58.1 GB** | **33.2 GB** | **−24.9 GB** |
| **KV** | **5.2 GB** | **13.6 GB** | **+161%** |
| **max KV tokens** | **137,216** | **355,280** | **2.59×** |
| teardown unowned | **28.01 GB / 1,980 allocs** | **1.12 GB / 124 allocs** | **−96%** |

A `grep -c` for the four sidecar/twin ledger sites (`loaders_fp8.rs:229`, `:230`, `quantized.rs:261`, `:262`) over the whole after-serve log returns **0**. By file, `loaders_fp8.rs` collapses 10,080 MB → **4.7 MB**, `quantized.rs` 11,680 MB → 160 MB, and `quantized.rs:643` (1,600 MB) is gone. The 76.14 GB ledger *total* is unchanged only because the freed memory was immediately re-spent on the ring and the KV cache.

The remaining 1.12 GB of unowned bytes is `quant_helpers.rs:98` (710 MB ×7) and four `impl_a1_init.rs` sites (352 MB) — **nothing from `load_layers`**. Both before and after also emit `backend drop reclaimed …` after the sweep, so nothing leaked in either case; what changed is the accounting.

Coherency gate `4/4` (`COH_EXIT=0`) on the after-build, 16/16 on a concurrent arithmetic probe, and `nvidia-smi` moved **+16 MiB across a full ladder** — memory flat, no leak.

### Throughput — read this carefully

At that config, with the other levers in the round-6 recipe (batched GDN, head band 16, cuBLASLt FFN prefill, `ATLAS_FFN_NO_BATCH16=1`) and the **full 8-slot SSM decode ring at batch 16** (`ring: 8 slots x 16 seqs x 151.5 MB/seq = 18.94 GB`), which the freed memory is what made bootable:

| shape | C | tok/s before → after | TPOT before → after |
|---|---|---|---|
| 1193 / 256 | 1 | 43.63 (TTFT 442 ms) | — |
| 1193 / 256 | 16 | 135.80 → **228.02** agg (+67.9%) | 100.74 → **55.62 ms** |
| 4593 / 512 | 1 | 37.41 (TTFT 1,420 ms) | — |
| 4593 / 512 | 16 | 118.62 → **177.60** agg (+49.7%) | — |

**Say plainly what that delta is.** It is **the ring depth the freed memory funded, not this PR's arithmetic** — which is unchanged. Before the fix the 8-slot ring at batch 16 could not fit and the serve ran on 1 slot; after it, `--ssm-decode-ring-slots auto` was accepted on the first boot. The throughput numbers belong to the recipe. **The memory numbers above are this PR's own receipt.**

**What this evidence does not establish.** One box, one model, one session. It says nothing about GB10. The C=16 throughput leg is confounded by the rest of the round-6 recipe by construction and is reported as context, not as a claim.

## Test plan

- [x] `cargo test -p spark-model --features cuda loader` — **61 passed, 0 failed** (1 ignored)
- [x] `cargo test -p spark-model --features cuda dense` — **98 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda residency` — **29 passed, 0 failed** (the new decision-table suite)
- [x] `cargo test -p spark-model --features cuda ledger` — **3 passed, 0 failed**
- [x] `cargo test -p spark-runtime --lib --features cuda derived` — **8 passed, 0 failed**
- [x] `cargo test -p spark-runtime --lib --features cuda teardown` — **8 passed, 0 failed**
- [x] `cargo test -p spark-runtime --lib --features cuda weights` — **93 passed, 0 failed** (3 ignored)
- [x] `cargo clippy -p spark-model -p spark-runtime -p spark-server --tests --features cuda,nccl` — clean, no warnings emitted. `spark-server` is in the invocation only because `nccl` is its feature; the two crates this PR edits are both covered.
- [x] Tested against real hardware — the H100 receipts above (2026-09-11).

- [ ] `scripts/hopper_ptx_gate.sh` — **not applicable**: these five commits touch **no `.cu`, no `.cuh` and no `kernels/` file at all**. Nothing new is compiled; this PR only decides which buffers get allocated.

The new CPU tests are the load-bearing part, because the claim they pin is a negative — "no kernel on this route reads an NVFP4 buffer" — and a negative cannot be proved by a benchmark:

- `fp8_residency_tests.rs` (17 cases) is the decision table under test: the default route builds no NVFP4 and only the KV FP8 twins; the KV twins survive **every** default route variation; the FFN and attention overlays are decided independently; `ATLAS_FP8_SINGLE_SCALE` brings the Q and O twins back; a target without the W8A8 kernels keeps every FP8 twin; each `ATLAS_ATTN_*` lever keeps exactly the NVFP4 copies it can reach; the escape hatch restores the pre-#915 loader and **cannot invent FP8 twins on a non-FP8 layer**; and the H100 ledger rows reproduce from the model shapes.
- `dense_ffn_fp8_residency_tests.rs` pins the claim the 18.4 GiB rested on — `every_small_batch_entry_point_runs_without_nvfp4_weights`, `the_mmq_finalizers_are_no_ops_under_a_native_fp8_overlay`, `the_verify_arm_still_selects_this_layer_without_nvfp4_weights`. **Each is red before this change.**
- `teardown_tests.rs` adds `releasing_frees_adopted_derived_buffers_too`, `releasing_twice_is_harmless_for_derived_buffers`, and `an_unadopted_derived_buffer_is_what_the_sweep_would_report` — the last one is the regression the whole PR is about, written so it fails if adoption is ever dropped.

## GB10 perf-gate records

This diff touches `crates/`, a `PERF_PATH`, so `record_covers` invalidates every committed `.benchmarks/` record on landing. **A GB10 re-measure is owed before merge.** The H100 evidence above is from a different card and does not substitute for it.

## Notes for reviewers

**The risk in this diff is a false negative in the decision table, not a wrong number.** If `DenseFp8Plan::resolve` says "not reachable" about a buffer some lever *can* route to, the failure is a NULL read at a dispatch site — a CUDA-700, loud and immediate, not a silent precision loss. That is the intended failure mode and it is why the table is a pure function with 17 CPU cases over it rather than a set of `if`s at the allocation sites. The contract is stated in the module header: **any new lever that can route a native-FP8 layer back onto an NVFP4 kernel must be added to `DenseFp8Plan::resolve` as well as to the dispatch site.**

**Why K and V twins stay and Q and O go.** `prefill/cache_skip_qkv.rs` has no W8A8 arm, and it is the first-chunk path of *every* request. Dropping the K/V twins would have been the same 1,600 MB and would have broken the common case. This asymmetry is the one place the residency rule is not "FP8 means no twin", and it is pinned by `the_kv_fp8_twins_survive_every_default_route_variation`.

**`prune_after_load` is the use-after-free surface.** It frees store tensors by name, and the only thing keeping it honest is that the loader and the pruner ask `gdn_fp8_arm_selected` the same question. Reviewers should check that single shared predicate rather than the two call sites — if those ever drift, the symptom is a corrupted weight with no diagnostic, not a crash.

**`can_forward_km` deliberately still returns `true` for a layer with NULL NVFP4 weights.** That reads like a bug and is not: it is a dispatch gate, and this PR's whole discipline is to change the footprint without moving the dispatch. `forward_km` redirects FP8 layers to `forward_prefill` before any NVFP4 pointer is touched.

All five commits **cherry-picked cleanly onto `main`** with `-x` provenance lines — **no conflicts**, no adaptation, every hunk applied to an identical pre-image. `dense_ffn.rs` (the file most likely to have moved under #987/#988) auto-merged. This PR is **not stacked** on anything.

## CLA

### DIFF
--- a/crates/spark-model/src/layers/dense_ffn.rs
+++ b/crates/spark-model/src/layers/dense_ffn.rs
@@ -467,6 +467,14 @@ impl DenseFfnLayer {
         if self.q2_weights.is_some() {
             return Ok(());
         }
+        // Native FP8 (#915): `forward_prefill_inner` returns from inside its
+        // `self.fp8_weights` arm long before the Q4_K/MMQ arm, so this repack
+        // is dead work — and since #915 the loader installs NULL NVFP4 source
+        // weights on that route, over which `ensure_q4k_weight`'s dequant is a
+        // CUDA-700 illegal access. Same reason as the packed-Q2 guard above.
+        if self.fp8_weights.is_some() {
+            return Ok(());
+        }
         let q4k_active = self.q4k_mmq_nc_k.0 != 0
             && self.q4k_quant_act_k.0 != 0
             && self.q4k_quant_w_k.0 != 0
@@ -563,6 +571,13 @@ impl DenseFfnLayer {
         // present) and repacks the NVFP4 gate/up — over NULL pointers that's a
         // CUDA-700 illegal access. Packed-Q2 uses its own decode/prefill path.
         if self.q2_weights.is_some() {
+            return Ok(());
+        }
+        // Native FP8 (#915): same reasoning as `finalize_q4k_load`, and this
+        // one is active by DEFAULT wherever the W4A4-MMQ kernels exist — so on
+        // a target that ships them it would repack NULL NVFP4 gate/up AND free
+        // `_t` copies for a prefill arm the FP8 early-return never reaches.
+        if self.fp8_weights.is_some() {
             return Ok(());
         }
         let active = self.nvfp4_mmq_nc_k.0 != 0
@@ -1552,7 +1567,15 @@ impl DenseFfnLayer {
     /// (batchm kernel present AND NVFP4 weights loaded — the batchm GEMV
     /// reads the non-transposed NVFP4 layout).
     pub fn can_forward_km(&self, m: u32) -> bool {
-        self.batchm_kernel(m).0 != 0 && !self.weights.gate_proj.weight.is_null()
+        self.batchm_kernel(m).0 != 0
+            && (!self.weights.gate_proj.weight.is_null()
+                // Native FP8 (#915): the loader no longer builds the NVFP4
+                // fallback, but `forward_km` redirects to `forward_prefill` for
+                // an FP8 layer anyway (`native_small_batch_uses_prefill`), so
+                // the answer must stay `true` or the n=4..8 verify arm in
+                // `multi_seq/ffn.rs:145` would fall through to a DIFFERENT
+                // branch and quietly change the routing this fix must not touch.
+                || self.fp8_weights.is_some())
     }
 
     /// K=m (m<=8) speculative verify: batched GEMV for m tokens.
@@ -2704,6 +2727,12 @@ mod native_batch_tests;
 #[cfg(test)]
 #[path = "dense_ffn_kernel_tests.rs"]
 mod kernel_tests;
+
+/// #915: the native-FP8 route must work with NO NVFP4 fallback weights, so the
+/// loader can stop allocating 18.4 GiB of them.
+#[cfg(test)]
+#[path = "dense_ffn_fp8_residency_tests.rs"]
+mod fp8_residency_tests;
 
 #[cfg(test)]
 mod tests {
--- a/crates/spark-model/src/layers/dense_ffn_fp8_residency_tests.rs
+++ b/crates/spark-model/src/layers/dense_ffn_fp8_residency_tests.rs
@@ -1,0 +1,203 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! The native-FP8 dense FFN must work with NO NVFP4 fallback weights (#915).
+//!
+//! WHY. On 1xH100, 2026-09-11, `Qwen/Qwen3.8-27B-FP8`, the loader built an
+//! NVFP4 gate/up/down plus a transposed twin for all 64 layers — 18.4 GiB — on
+//! the strength of a comment that said the batched spec-decode paths had no FP8
+//! branch. They do: `forward_k2`/`forward_k3`/`forward_km` all redirect through
+//! `native_small_batch_uses_prefill`, and `w8_gemm!` binds its transposed
+//! operand to a literal `None`. These tests pin that, so the loader can install
+//! `QuantizedWeight::null()` and the 18.4 GiB stays unallocated.
+//!
+//! Every case here is RED before #915: with a NULL `gate_proj`,
+//! `can_forward_km` returned false (silently rerouting the 4..8-row verify arm)
+//! and `finalize_nvfp4_mmq_load` — active by DEFAULT wherever its kernels
+//! exist — repacked over the null pointers.
+
+use super::{DenseFfnLayer, DenseFfnWeights};
+use crate::layer::{ForwardContext, MoeLoraRoute};
+use crate::layers::ops::{DerivedWeights, GemmDispatch, ModelLevers, ModelStats};
+use crate::weight_map::{Fp8Weight, QuantizedWeight, WeightQuantFormat};
+use atlas_core::config::ModelConfig;
+use spark_runtime::buffers::BufferArena;
+use spark_runtime::gpu::mock::MockGpuBackend;
+use spark_runtime::gpu::{GpuBackend, KernelHandle};
+
+fn config() -> ModelConfig {
+    let mut config = ModelConfig::qwen3_next_80b_nvfp4();
+    config.hidden_size = 128;
+    config.intermediate_size = 128;
+    config.num_experts = 1;
+    config.num_experts_per_tok = 1;
+    config.moe_intermediate_size = 128;
+    config.vocab_size = 128;
+    config
+}
+
+/// Exactly what `qwen35_dense.rs` now builds on the native-FP8 route: NULL
+/// NVFP4 fallbacks, no transposed twins, an FP8 overlay on top.
+fn native_fp8_layer(gpu: &MockGpuBackend) -> DenseFfnLayer {
+    let mut layer = DenseFfnLayer::new(
+        DenseFfnWeights {
+            gate_proj: QuantizedWeight::null(),
+            up_proj: QuantizedWeight::null(),
+            down_proj: QuantizedWeight::null(),
+            gate_proj_t: None,
+            up_proj_t: None,
+            down_proj_t: None,
+        },
+        gpu,
+    )
+    .unwrap();
+    layer.w8a16_gemm_k = KernelHandle(0xF08);
+    layer.act_mul = KernelHandle(0xAC7);
+    let fp8 = Fp8Weight {
+        weight: gpu.alloc(128 * 128).unwrap(),
+        row_scale: gpu.alloc(4).unwrap(),
+        n: 128,
+        k: 128,
+        scale_format: WeightQuantFormat::Fp8BlockScaled,
+    };
+    layer.set_fp8_weights(fp8, fp8, fp8);
+    layer
+}
+
+/// Turn on every handle the two load-time MMQ finalizers check, so the test
+/// exercises the arm a real Blackwell/GB10 kernel set would enable rather than
+/// passing because the handles happen to be zero.
+fn arm_the_mmq_finalizers(layer: &mut DenseFfnLayer) {
+    for h in [
+        &mut layer.q4k_mmq_nc_k,
+        &mut layer.q4k_quant_act_k,
+        &mut layer.q4k_quant_w_k,
+        &mut layer.dequant_nvfp4_bf16_k,
+        &mut layer.nvfp4_mmq_nc_k,
+        &mut layer.nvfp4_quant_act_k,
+        &mut layer.nvfp4_repack_k,
+        &mut layer.nvfp4_silu_scaled_k,
+    ] {
+        *h = KernelHandle(0xBEEF);
+    }
+}
+
+fn with_ctx(gpu: &MockGpuBackend, f: impl FnOnce(&ForwardContext, &BufferArena)) {
+    let config = config();
+    let buffers = BufferArena::new(&config, 8, 256, 256, 8, gpu).unwrap();
+    let dispatch = GemmDispatch::defaults();
+    let derived = DerivedWeights::new();
+    let levers = ModelLevers::defaults();
+    let stats = ModelStats::new();
+    let ctx = ForwardContext {
+        buffers: &buffers,
+        hc_row_offset: 0,
+        gpu,
+        config: &config,
+        dispatch: &dispatch,
+        derived: &derived,
+        levers: &levers,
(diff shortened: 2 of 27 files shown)


tbraun96:
Closing as **superseded by #895**, at the repo owner's direction — the same disposition #985 received into #992.

Every commit on this branch is already present in `hopper/sm90-target-tdd-2026-09`: `git cherry refs/remotes/pr/895 refs/remotes/pr/991 main` reports **0 new**, and the files this PR *creates* (`fp8_residency.rs`, `predicted_residency.rs`, `preflight/headroom.rs`) already exist there. Diffing #895 against this branch over only this PR's own files shows #895 carrying strictly more, so nothing here is lost by closing.

#895 is going in at the bottom of the stack, which would leave this PR with nothing to merge.

No work is discarded — reopen if the containment check above turns out to be wrong.

TheTom:

TheTom:
Merged base #985 forward (merge commit `95a83c54f`).

#985's head was not an ancestor of #991, but the two layers touch disjoint files: **no conflicts**.

PR's own diff unchanged (verified): all 13 of #991's files byte-identical to `d6f5b768b`, and every file that differs from #985's head is one of those 13.

Distance from the measured tree: n/a — nothing was resolved.

Gates on Spark 1 (GB10, CUDA 13): fmt ok · clippy `-p spark-model --features cuda` / `-p spark-runtime --features cuda` / `-p spark-server --features cuda,nccl`, `-D warnings`, all ok · `cargo test -p spark-model --lib` 914 passed 0 failed · `-p spark-runtime --lib` 305 passed 0 failed · `-p spark-server --lib` 114 passed 0 failed · `-p atlas-kernels --no-default-features` ok · CI file-size cap ok.

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1036 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1036; this fork PR is closed.

PR #992, opened 2026-09-11
spark-server: size the SSM rollback ring against the predicted post-load KV headroom (#915)

### LINKED ISSUE #915: Dense 27B on an 80 GB GPU: inference reserve is 45.8 GiB (8 rollback snapshots x max-batch-size 32) and boot is refused; reserve is not sized from free memory

TheTom:
## Summary

The #915 auto-fit's first pass (#985) shrinks the SSM decode-rollback ring, but it sizes it against **pre-load free memory** — 78.6 GB on an 80 GB H100, before a single weight byte is resident. The quantity that actually has to accommodate the reserve is the **post-load KV headroom**, and a different stage enforces it four minutes later with no shrink path, only a refusal. So the fitter stayed silent at every batch size that mattered and the serve died anyway.

This fits against the same formula that stage enforces:

```text
budget          = total_mem x --gpu-memory-utilization
pre_kv_estimate = weights + derived + buffer_arena + fixed_reserve
headroom        = budget - pre_kv_estimate
```

and takes the largest ladder depth for which `ring_bytes + kv_floor <= headroom`.

On 1xH100 with `--max-batch-size 32` and **no `--ssm-decode-ring-slots` token in the argv at all**, the serve now boots unaided in 256 s at the depth an operator previously had to pass by hand. The same recipe one binary earlier was refused.

**Fixes #915** — see "Against #915's three asks" below; #985 closed two of them, this closes the third.

## What changed

Two commits, both `spark-server`-led:

- **`preflight/headroom.rs` (new, 378 lines)** is the yardstick. It computes the budget/pre-KV/headroom subtraction above from what preflight can already see, and `describe()` prints every term in the order the arithmetic runs, so an operator reading one INFO line can redo it.
- **Two terms were not knowable pre-load and now are.** `weights` is the checkpoint's on-disk byte count, read from the safetensors index — an upper bound on residency, because the loader prunes and never grows (round 6's 28.75 GB checkpoint settles at ~24.7 GB), so it can only ever choose a *smaller* ring. `derived` is `weight_loader::predicted_residency::predicted_derived_bytes`, which evaluates the native-FP8 dense loader's own residency plan (`fp8_residency::DenseFp8Plan`, #991) from `config.json` alone.
- **Neither estimate is allowed to be a guess.** The post-load yardstick is used **only** where the derived prediction is available, which is the same route on which the on-disk count is sound. `predicted_derived_bytes` declines explicitly — not the Qwen3.5-dense loader, `ATLAS_DENSE_FP8` unset, `--tp-size > 1`, `config.json` not declaring block-scaled FP8, `ATLAS_DENSE_FP8_KEEP_NVFP4` set, or any NVFP4 fallback lever armed — and every declining route keeps the pre-load behaviour.
- **The fit decision is logged at INFO on every boot**, with all its terms, in both branches — so "the fitter approved this" can be told from "the fitter never looked". The pre-load branch names *why* it fell back.
- **`--ssm-decode-ring-slots N` is unchanged**: it still pins the depth and is refused rather than shrunk.
- **The refusal's text is now tested, and stops reading a global.** `ring_note` read `published_decode_ring_slots()` directly — a process `OnceLock`, so one caller sealing it decided the wording for every later refusal in the same process, and made the text untestable in a shared test binary. The flag is an input on `Refusal` now, passed by the one call site that knows it. Refusal is the last resort after #985 and is the only thing an operator has to act on, so it earns a test.
- **Supporting factorings**: `weight_map::config_declared_variant` (step 1 of `detect_nvfp4_variant`, so a caller holding only `config.json` asks the same question); `qwen3_attention::W8A8_PREFILL_KERNELS` names the two kernels once for `init.rs`, `has_w8a8_prefill_kernels` and the prediction; the post-load memory audit moves to `preflight/post_load_audit.rs`, which keeps `preflight.rs` under the 500-line cap.

## Why the first pass could not work — the sweep it was measured against

From #985's own probe, on the same card:

| `--max-batch-size` | ring chosen | fired? | serve |
|---|---|---|---|
| 4 | 8 | no | boots |
| 16 | 8 (18.94 GB) | no | **refused** by the KV stage |
| 32 | 8 (37.88 GB) | no | **refused** by the KV stage |
| 64 | 8 -> 4 | yes | still refused |

The fitter's own WARN said it: `reserve 54.13 of 78.57 GB`. The KV stage measures something else entirely, and that is the number the ring has to live inside.

## Evidence

**Diagnostic single-H100 evidence, not certified campaign numbers.** 1xH100 80 GB (`atlas-testing-jawn`), `Qwen/Qwen3.8-27B-FP8`, native FP8, `--lm-head-dtype bf16`, `--gpu-memory-utilization 0.90`, `--max-seq-len 24576`, FP8 KV, 2026-09-11 (round 8, stage 2; round-6 comparands from the same box and recipe at `--max-batch-size 16`). Build at the integration tip carrying these two commits, `cargo build --locked --release`, exit 0.

### The boot, with no ring flag in the argv

`--max-batch-size 32`, and the `--ssm-decode-ring-slots auto` token was **deleted from the serve script, not re-spelled** — so what boots below is the compiled default. Confirmed in the recorded argv: there is no `--ssm-decode-ring-slots` token in it.

```text
INFO preflight: SSM decode-ring fit against the predicted post-load KV headroom:
  budget 71.26 GB (90% util) = weights 28.75 + derived 3.95 + arena 3.58 +
  fixed reserve 7.95 -> headroom 27.04 GB; ring(4) 18.94 + KV floor 4.00 = 22.94 GB

WARN preflight: SSM decode-rollback ring auto-fit — ring: 4 slots x 32 seqs x
  151.5 MB/seq = 18.94 GB (was 37.88 GB); ring + KV floor 22.94 of 27.04 GB.
  Sized from the predicted post-load KV headroom, not from --max-batch-size 32
  (#915): rollback depth yields, concurrency does not. Pass
  --ssm-decode-ring-slots N to pin a depth (and be refused rather than shrunk).

INFO preflight: Preflight reserve: inference=27531 MB, buffer_arena=3664 MB
  (pre-load free: 78.6 GB); ring: 4 slots x 32 seqs x 151.5 MB/seq = 18.94 GB

INFO factory::build: KV cache: 79.2 GB total x 90% util = 71.3 GB budget;
  33.2 GB pre-KV + 26.9 GB reserve -> 11.2 GB for KV -> 18317 blocks x
  16 tok/block = 293072 max KV tokens
```

**READY at +256 s, first try, no manual pin, no retry.** A full-log grep for `fallback|refus|No memory left|pre-load reserve|shrunk|backed off` returns only the auto-fit WARN itself and an unrelated `Remote image fetching disabled` notice — the post-load fit is the only path taken.

### It derives exactly what round 6 had to pin by hand

| line | r6 (bs32, ring `auto`, first pass) | r6 (bs32, **hand-pinned 4**) | **r8 (bs32, no flag)** |
|---|---|---|---|
| ring | 8 x 32 = 37.88 GB | 4 x 32 = 18.94 GB | **4 x 32 = 18.94 GB** |
| `Preflight inference` | 46,923 MB | 27,531 MB | **27,531 MB** |
| KV | — | 11.2 GB | **11.2 GB** |
| blocks / max KV tokens | — | 18,317 / 293,072 | **18,317 / 293,072** |
| boot | **REFUSED** (`33.2 + 45.8 = 79.0 GB` committed against a 71.3 GB budget) | OK, after a manual retry | **OK, first try, no flag** |

Every figure in the last column matches the hand-pinned column to the byte. The change turns a boot refusal plus an operator retry into an automatic, logged, one-line decision — and the WARN says why it is allowed to shrink and how to opt out.

### Prediction accuracy — the load-bearing estimate, checked against the loader's own log

The fit line's `derived 3.95` and the loader's `native FP8 dense residency: ... derived 4.24 GB` (round 6, `--max-batch-size 16`, same box and recipe) are **the same 4,242,882,560 bytes**: `predicted_residency` is exact on this route, and the two lines differ only in units — `headroom.rs` divides by 2^30 while `fp8_residency.rs` divides by 1e9, and both print the literal `GB`. 4,242,882,560 / 2^30 = 3.951; / 1e9 = 4.243.

`the_27b_prediction_matches_the_round6_residency_summary` pins the total at that exact integer and asserts the relative error against the logged 4.24 GB is **under the 5% bound** — it lands at 0.07%. The two component assertions (16 attention layers x (K twin + V twin) = 167,813,120 B; 48 GDN layers x 84,897,280 = 4,075,069,440 B) are recomputed independently in the test, so a change to `ssm_concat_bytes` cannot silently redefine what is being checked.

### Throughput — a scheduler control, and it behaves like one

Ladder at `1024x256`, C=16 only:

| cell | r8 bs32 auto | r8 bs16 control (same binary) |
|---|---|---|
| C=16 aggregate | **227.72 tok/s** | 227.68 tok/s |
| C=16 TPOT p50 | **55.66 ms** | 55.65 ms |

A 0.02% difference against a 0.11-0.24% rep spread — i.e. **no throughput claim is being made here**. Raising the cap to 32 does nothing at C=16 because C=16 cannot fill more than 16 rows. What it still costs is KV: 13.6 -> 11.2 GB, 355,280 -> 293,072 tokens, because the reserve scales with the cap whether or not the rows are used. **bs32 remains strictly worse than bs16 on this box unless offered concurrency actually exceeds 16.** What this PR changes is that the answer is now "and it costs you 2.4 GB of KV" instead of "and it will not boot".

Gates on that boot: smoke PASS (`391`, TTFT 72.54 ms), coherency `COH_EXIT=0` **4/4**, 16-way concurrent arithmetic probe **16/16**, and GPU memory moved 73,851 -> 73,879 MiB across ready + gates + ladder — flat, no leak.

### One reporting gap, named rather than left to be found

`kernel flags: ... ssm_decode_ring_slots=auto` prints the flag **as parsed, not as fitted**. It is emitted 449 ms *before* the preflight fit, so it structurally cannot carry the chosen depth; there is exactly one `ssm_decode_ring_slots` occurrence in the whole log and it says `auto`. The ring really is 4 slots — the fit line, the WARN and the `Preflight reserve` line all agree, and the resulting KV is bit-identical to the hand-pinned boot. This is a reporting gap, not a behaviour gap, but anyone grepping `kernel flags` to learn the effective ring depth on an `auto` boot will read the wrong number. Fixing it means re-emitting or deferring that line, which is a `serve_flags` change with nothing to do with the fit, so it is not in this PR.

## Against #915's three asks

| #915 asks | state |
|---|---|
| Size the ring from the post-weights free memory, and shrink instead of refusing to boot | **done here**, verified on hardware: bs32 with no flag boots at ring 4 and 11.2 GB of KV |
| Make ring depth a first-class flag with a documented memory formula | **done in #985**, verified on hardware |
| Preflight prints the formula so the operator sees why N GiB was asked for | **done in #985**; this PR adds the fit's own INFO line with every term of the subtraction |

**What remains outside the fix, stated plainly.** The post-load yardstick answers only for the native-FP8 dense route at `--tp-size 1`. Every other model keeps the pre-load free-memory yardstick — and logs, at INFO, that it did and why. That is deliberate: the alternative is predicting residency for loaders that requantise at load, where the on-disk byte count is not an upper bound on anything, and a wrong prediction there would refuse a serve that would have booted. Extending the prediction to another loader is a per-loader change with its own hardware bill. The route in #915's title and evidence is covered, which is why this is `Fixes`.

## Test plan

- [x] `cargo test -p spark-model --features cuda --lib ssm_reserve` — **25 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda --lib residency` — **29 passed, 0 failed**
- [x] `cargo test -p spark-server --bins --features cuda,nccl preflight` — **42 passed, 0 failed**
- [x] `cargo test -p spark-server --bins --features cuda,nccl ssm_decode_ring` — **14 passed, 0 failed**
- [x] `cargo clippy -p spark-model -p spark-server -p spark-runtime --tests --features cuda,nccl` — clean, no warnings emitted
- [x] Tested against real hardware — 1xH100 80 GB, 2026-09-11: the no-flag bs32 boot and its gates above
- [ ] PTX gate — not applicable, **no `.cu` or `.cuh` file is touched** (confirmed against the full 38-file diff)

## GB10 perf-gate records

## Notes for reviewers

**Stacked on #985 and #991.** #985's two commits are the first two here and #991's six the next six; this PR's own content is the **last two**. Review those:

```
git diff pr/915-ssm-ring-autofit..
```

The base is `main` rather than either stack branch, so this PR does not become unreviewable if #985 or #991 is rebased — and so the diff shown by GitHub carries all eight commits. Merge #985, then #991, then this.

The dependency on #991 is real and not bookkeeping: `predicted_derived_bytes` evaluates `fp8_residency::DenseFp8Plan`, which #991 introduces, and it is the whole reason `derived` can be a computed term rather than a fudge factor. Cherry-picking #985's and #991's commits onto this branch applied with **no conflicts**; nothing was adapted. This PR's own two were cherry-picked with `-x` provenance lines and also applied clean.

The part to read closest is `headroom.rs`'s "What it refuses to estimate" section and the gate list in `predicted_derived_bytes`. The correctness of this change is entirely a question of whether those gates are tight enough — a prediction that is available when it should not be will under-size the ring on a route whose residency it cannot see, and nothing downstream would catch that. `DenseFp8Plan::resolve` being a pure function of `DenseFp8Inputs` is what lets the CPU tests pin every clause without touching the process environment.

## CLA

### DIFF
--- a/crates/spark-model/src/layers/dense_ffn.rs
+++ b/crates/spark-model/src/layers/dense_ffn.rs
@@ -467,6 +467,14 @@ impl DenseFfnLayer {
         if self.q2_weights.is_some() {
             return Ok(());
         }
+        // Native FP8 (#915): `forward_prefill_inner` returns from inside its
+        // `self.fp8_weights` arm long before the Q4_K/MMQ arm, so this repack
+        // is dead work — and since #915 the loader installs NULL NVFP4 source
+        // weights on that route, over which `ensure_q4k_weight`'s dequant is a
+        // CUDA-700 illegal access. Same reason as the packed-Q2 guard above.
+        if self.fp8_weights.is_some() {
+            return Ok(());
+        }
         let q4k_active = self.q4k_mmq_nc_k.0 != 0
             && self.q4k_quant_act_k.0 != 0
             && self.q4k_quant_w_k.0 != 0
@@ -563,6 +571,13 @@ impl DenseFfnLayer {
         // present) and repacks the NVFP4 gate/up — over NULL pointers that's a
         // CUDA-700 illegal access. Packed-Q2 uses its own decode/prefill path.
         if self.q2_weights.is_some() {
+            return Ok(());
+        }
+        // Native FP8 (#915): same reasoning as `finalize_q4k_load`, and this
+        // one is active by DEFAULT wherever the W4A4-MMQ kernels exist — so on
+        // a target that ships them it would repack NULL NVFP4 gate/up AND free
+        // `_t` copies for a prefill arm the FP8 early-return never reaches.
+        if self.fp8_weights.is_some() {
             return Ok(());
         }
         let active = self.nvfp4_mmq_nc_k.0 != 0
@@ -1552,7 +1567,15 @@ impl DenseFfnLayer {
     /// (batchm kernel present AND NVFP4 weights loaded — the batchm GEMV
     /// reads the non-transposed NVFP4 layout).
     pub fn can_forward_km(&self, m: u32) -> bool {
-        self.batchm_kernel(m).0 != 0 && !self.weights.gate_proj.weight.is_null()
+        self.batchm_kernel(m).0 != 0
+            && (!self.weights.gate_proj.weight.is_null()
+                // Native FP8 (#915): the loader no longer builds the NVFP4
+                // fallback, but `forward_km` redirects to `forward_prefill` for
+                // an FP8 layer anyway (`native_small_batch_uses_prefill`), so
+                // the answer must stay `true` or the n=4..8 verify arm in
+                // `multi_seq/ffn.rs:145` would fall through to a DIFFERENT
+                // branch and quietly change the routing this fix must not touch.
+                || self.fp8_weights.is_some())
     }
 
     /// K=m (m<=8) speculative verify: batched GEMV for m tokens.
@@ -2704,6 +2727,12 @@ mod native_batch_tests;
 #[cfg(test)]
 #[path = "dense_ffn_kernel_tests.rs"]
 mod kernel_tests;
+
+/// #915: the native-FP8 route must work with NO NVFP4 fallback weights, so the
+/// loader can stop allocating 18.4 GiB of them.
+#[cfg(test)]
+#[path = "dense_ffn_fp8_residency_tests.rs"]
+mod fp8_residency_tests;
 
 #[cfg(test)]
 mod tests {
--- a/crates/spark-model/src/layers/dense_ffn_fp8_residency_tests.rs
+++ b/crates/spark-model/src/layers/dense_ffn_fp8_residency_tests.rs
@@ -1,0 +1,203 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! The native-FP8 dense FFN must work with NO NVFP4 fallback weights (#915).
+//!
+//! WHY. On 1xH100, 2026-09-11, `Qwen/Qwen3.8-27B-FP8`, the loader built an
+//! NVFP4 gate/up/down plus a transposed twin for all 64 layers — 18.4 GiB — on
+//! the strength of a comment that said the batched spec-decode paths had no FP8
+//! branch. They do: `forward_k2`/`forward_k3`/`forward_km` all redirect through
+//! `native_small_batch_uses_prefill`, and `w8_gemm!` binds its transposed
+//! operand to a literal `None`. These tests pin that, so the loader can install
+//! `QuantizedWeight::null()` and the 18.4 GiB stays unallocated.
+//!
+//! Every case here is RED before #915: with a NULL `gate_proj`,
+//! `can_forward_km` returned false (silently rerouting the 4..8-row verify arm)
+//! and `finalize_nvfp4_mmq_load` — active by DEFAULT wherever its kernels
+//! exist — repacked over the null pointers.
+
+use super::{DenseFfnLayer, DenseFfnWeights};
+use crate::layer::{ForwardContext, MoeLoraRoute};
+use crate::layers::ops::{DerivedWeights, GemmDispatch, ModelLevers, ModelStats};
+use crate::weight_map::{Fp8Weight, QuantizedWeight, WeightQuantFormat};
+use atlas_core::config::ModelConfig;
+use spark_runtime::buffers::BufferArena;
+use spark_runtime::gpu::mock::MockGpuBackend;
+use spark_runtime::gpu::{GpuBackend, KernelHandle};
+
+fn config() -> ModelConfig {
+    let mut config = ModelConfig::qwen3_next_80b_nvfp4();
+    config.hidden_size = 128;
+    config.intermediate_size = 128;
+    config.num_experts = 1;
+    config.num_experts_per_tok = 1;
+    config.moe_intermediate_size = 128;
+    config.vocab_size = 128;
+    config
+}
+
+/// Exactly what `qwen35_dense.rs` now builds on the native-FP8 route: NULL
+/// NVFP4 fallbacks, no transposed twins, an FP8 overlay on top.
+fn native_fp8_layer(gpu: &MockGpuBackend) -> DenseFfnLayer {
+    let mut layer = DenseFfnLayer::new(
+        DenseFfnWeights {
+            gate_proj: QuantizedWeight::null(),
+            up_proj: QuantizedWeight::null(),
+            down_proj: QuantizedWeight::null(),
+            gate_proj_t: None,
+            up_proj_t: None,
+            down_proj_t: None,
+        },
+        gpu,
+    )
+    .unwrap();
+    layer.w8a16_gemm_k = KernelHandle(0xF08);
+    layer.act_mul = KernelHandle(0xAC7);
+    let fp8 = Fp8Weight {
+        weight: gpu.alloc(128 * 128).unwrap(),
+        row_scale: gpu.alloc(4).unwrap(),
+        n: 128,
+        k: 128,
+        scale_format: WeightQuantFormat::Fp8BlockScaled,
+    };
+    layer.set_fp8_weights(fp8, fp8, fp8);
+    layer
+}
+
+/// Turn on every handle the two load-time MMQ finalizers check, so the test
+/// exercises the arm a real Blackwell/GB10 kernel set would enable rather than
+/// passing because the handles happen to be zero.
+fn arm_the_mmq_finalizers(layer: &mut DenseFfnLayer) {
+    for h in [
+        &mut layer.q4k_mmq_nc_k,
+        &mut layer.q4k_quant_act_k,
+        &mut layer.q4k_quant_w_k,
+        &mut layer.dequant_nvfp4_bf16_k,
+        &mut layer.nvfp4_mmq_nc_k,
+        &mut layer.nvfp4_quant_act_k,
+        &mut layer.nvfp4_repack_k,
+        &mut layer.nvfp4_silu_scaled_k,
+    ] {
+        *h = KernelHandle(0xBEEF);
+    }
+}
+
+fn with_ctx(gpu: &MockGpuBackend, f: impl FnOnce(&ForwardContext, &BufferArena)) {
+    let config = config();
+    let buffers = BufferArena::new(&config, 8, 256, 256, 8, gpu).unwrap();
+    let dispatch = GemmDispatch::defaults();
+    let derived = DerivedWeights::new();
+    let levers = ModelLevers::defaults();
+    let stats = ModelStats::new();
+    let ctx = ForwardContext {
+        buffers: &buffers,
+        hc_row_offset: 0,
+        gpu,
+        config: &config,
+        dispatch: &dispatch,
+        derived: &derived,
+        levers: &levers,
+        stats: &stats,
+        attn_metadata: None,
+        profile: false,
+        comm: None,
+        graph_capture: false,
+        decode_step: false,
+        gdn_exact_replay: false,
+        token_ids: None,
+        host_token_ids: None,
+        routed_lora_layers: None,
+        midchunk_capture: None,
+        moe_lora_route: MoeLoraRoute::Fold,
+    };
+    f(&ctx, &buffers);
+}
+
+#[test]
+fn the_verify_arm_still_selects_this_layer_without_nvfp4_weights() {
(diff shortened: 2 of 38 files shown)


TheTom:
Thank you for catching this — the citation was genuinely dangling. `kernels/hopper/qwen3.8-27b/MODEL.toml` exists only on `hopper/h100-integration-2026-09-10`; on `main` the one MODEL.toml for this checkpoint is `kernels/gb10/qwen3.8-27b/MODEL.toml`, so all three references now point there (`5da6b91b6`). The gb10 file is the right source rather than a stand-in: its architecture block is target-independent and the Hopper copy repeats those fields verbatim, adding nothing but `expected_absent` entries — so the fixture is still the checkpoint the H100 rounds served, just cited at a path that exists in this tree. No test reads the file, the shapes were already inlined in `qwen38_27b()`, and nothing about the fixture or the pinned integers changes. While verifying the source I found a second, smaller inaccuracy and fixed it in the same commit: the GDN head geometry (16×128 key heads, 48×128 value heads) was attributed to the MODEL.toml and is not in it — **no** MODEL.toml carries the linear-attention head fields, `ModelConfig` reads them from the checkpoint's `config.json` — which matters because the 83,886,080-byte fused `[QKV|Z]` term is arithmetic over exactly those numbers, so both doc comments now say where they come from. The same commit cherry-picked cleanly onto `hopper/h100-integration-2026-09-10` (`5404a2b14`) to keep the two branches from diverging on this text. Gates on Spark 1 at `5da6b91b6`, isolated worktree, all green: `cargo test -p spark-server --bins --features cuda,nccl preflight` **42 passed**, `headroom` **13 passed**, `cargo test -p spark-model --features cuda residency` **29 + 1 passed**, all 0 failed; `cargo doc --workspace --no-deps` clean; `cargo fmt --all --check` clean. One note on the commit subject: it reads `spark-model:` rather than `spark-server:`, because both touched files live in `crates/spark-model`.

tbraun96:
Closing as **superseded by #895**, at the repo owner's direction — the same disposition #985 received into #992.

Every commit on this branch is already present in `hopper/sm90-target-tdd-2026-09`: `git cherry refs/remotes/pr/895 refs/remotes/pr/992 main` reports **0 new**, and the files this PR *creates* (`fp8_residency.rs`, `predicted_residency.rs`, `preflight/headroom.rs`) already exist there. Diffing #895 against this branch over only this PR's own files shows #895 carrying strictly more, so nothing here is lost by closing.

#895 is going in at the bottom of the stack, which would leave this PR with nothing to merge.

No work is discarded — reopen if the containment check above turns out to be wrong.

TheTom:

TheTom:
Merged base #991 forward (merge commit `d81b6f4b9`).

#992 cherry-picked #991's layer from `b2a7ae16e` under its own SHAs, so the three files it then extended collide content-wise with #991's originals. In every case **#992's side is #991's content plus #992's own addition, and the base's side is simply that addition's absence** — so the PR's own side wins all three:

* `qwen3_attention/types_weights.rs` — `W8A8_PREFILL_KERNELS` and `w8a8_prefill_kernels_loaded`, which the pre-load residency prediction reads (#915);
* `qwen3_attention/mod.rs` — the re-export of those two;
* `weight_loader/qwen35_dense.rs` — `pub mod fp8_residency`, published so spark-server's preflight can price its terms with the loader's own shape helpers.

Cross-check: `hopper/h100-integration-2026-09-10` carries all three symbols, so this is the side the measured tree is built on.

PR's own diff unchanged (verified): all 26 of #992's files byte-identical to `5da6b91b6`; every file differing from the base is inside that set.

Distance from the measured tree (`4f00cf199`) across the 3 conflicted files: **+14 / −165**.

Also note this advances the branch's main point from `b2a7ae16e` to `ea57dc3da`, which #991 already carried.

Gates on Spark 1 (GB10, CUDA 13): fmt ok · clippy `-p spark-model --features cuda` / `-p spark-runtime --features cuda` / `-p spark-server --features cuda,nccl`, `-D warnings`, all ok · `cargo test -p spark-model --lib` 922 passed 0 failed · `-p spark-runtime --lib` 305 passed 0 failed · `-p spark-server --lib` 114 passed 0 failed · `-p atlas-kernels --no-default-features` ok · CI file-size cap ok.

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1037 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1037; this fork PR is closed.

PR #993, opened 2026-09-11
spark-model: stage the SwiGLU once for the native-FP8 decode FFN down projection (#928)

TheTom:
## Summary

The native-FP8 M=1 decode FFN reads the same 89.1 MB of weights per layer twice: once for `gate`+`up`, once for `down`. nsys says the first read runs at **1,979 GB/s** and the second at **858 GB/s** — the `down` projection is **30.4% of a 21.891 ms C=1 step** and the single most concentrated latency target on the board.

It is not a bandwidth shape problem. `w8a16_gemv_silu_input` fuses the SwiGLU into the down GEMV and evaluates it **per output column, not once**: every one of the `ceil(N/4)` CTAs gives each of its 4 outputs a 64-lane team, and every team walks the whole K computing `silu(gate[k])*up[k]` for itself. That is `N*K` = 5,120 × 17,408 = **89.1 M** SwiGLU evaluations per launch where the token needs 17,408 — a 5,120× redundancy — and each one is an `__expf` plus a **true IEEE FP32 divide** (`kernels/gb10/common/KERNEL.toml` builds with `--fmad=false` and no `-use_fast_math`, so `g / (1 + e^-g)` is the division sequence, not a reciprocal). ptxas puts the fused kernel at **53 registers → 4 CTAs/SM**, against 32 registers → 8 CTAs/SM for the plain GEMV. It is issue-bound and half-occupancy, not bandwidth-bound.

This PR stages `silu(gate)*up` once with the elementwise kernel prefill already uses, then runs `down` as a plain weight-streaming GEMV. On the real Qwen3.8-27B `down` shape that is **105.7 µs → 58.9 µs, 843 → 1,513 GB/s, 1.79×**, and at serving scale **C=1 TPOT 21.28 → 17.89 ms (−15.9%)**.

**That is the whole PR now.** It is Rust-only: `git diff avarok/main -- kernels/` is **empty** — no `.cu` file is added, modified or deleted.

Refs #928.

**Base is `main`.** The FP8 decode arm this PR rewrites is the one already on `main`; it does not need #988's 5..32-row ladder, and the two do not overlap (#988 adds rungs *above* M=4, this changes what M=1 runs).

## 🪤 What was removed from this PR, and why

Earlier revisions of this branch also added `kernels/gb10/common/w8a16_gemv_splitk.cu` — a split-K W8A16 decode GEMV behind `ATLAS_FFN_DOWN_SPLITK`, default OFF — aimed at the *second* cause below (wave quantisation). **It has been deleted**, along with its ops wrappers, its `SplitKGemv`/`SplitKPlan` plan code, its FP32 partial scratch, the lever, and the split-K arms of the microtest. Two independent reasons:

1. **The H100 measured it a null-to-loss, on both shapes it was written for.** `down`: **61.8 µs with the lever on against 58.9 µs without it (4.9% slower)**. Attention k/v (N=1024, K=5120): **0.67×, a 33% regression**. At serving scale the lever cost C=1 TPOT 17.89 → 17.98 ms (+0.50%) and was a null at C=16. A kernel that loses on its own target shape does not earn a lever, a second launch, a reassociated combine, ~160 KB/layer of partials, or review surface.
2. **It was in the wrong tree.** Per the maintainer's kernel-placement rule, Hopper-tuned kernels belong under `kernels/hopper/` as real files, and `kernels/gb10/common` changes only for shared correctness fixes. The split-K kernel was a Hopper occupancy experiment living in the shared GB10 tree. Removing it settles the placement question by not raising it.

The nsys attribution for cause 2 is still in the module doc, now recorded as **LEFT OPEN** with the split-K result as its receipt, so the next attempt starts from a fresh profile rather than from a plan that has already been measured and lost.

## What was wrong

nsys, 1×H100 80 GB, `Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`, native FP8, 2026-09-11, C=1 steady-state decode step **21.891 ms**, GPU busy 96%:

| kernel | shape | grid | launches | µs each | ms/step | GB/s |
|---|---|---|---|---|---|---|
| `w8a16_gemv_silu_input` (down) | N=5120 K=17408 | 1280 | 64 | 103.9 | **6.65 (30.4%)** | **858** |
| `w8a16_gemv_dual` (gate+up) | N=17408×2 K=5120 | 4352 | 64 | 90.1 | 5.77 | **1,979** |
| `w8a16_gemv` | N=16384 K=5120 | 4096 | — | — | — | 1,852 |

Same bytes on the first two rows, 43% of the rate. Two independent causes, and the module doc names both:

1. **The fused SwiGLU is recomputed per output.** The redundancy, the transcendental cost and the `--fmad=false` divide are above. Counting ~20 ops per K element against the dual GEMV's ~6, the fused kernel issues ~56 M warp instructions per launch — ~60 µs of pure issue on 132 SMs — on top of a 26.6 µs weight stream. The NVFP4 arm reached this verdict with ncu (SM 57% vs memory 23%) and has staged the activation once ever since; **the FP8 arm simply never got the same treatment.** This is the cause the PR fixes.
2. **The grid is a pure function of N — left open.** `w8a16_gemv.cu` puts `N_PER_BLOCK=4` outputs in a 256-thread CTA, so K never enters the CTA count. ~8 CTAs co-reside per SM: grid 4,352 is ~4.1 waves (a ~2% tail), grid 1,280 is **~1.2 waves**. Split-K was the obvious lever for this and lost on the measurement (above), so nothing here addresses it.

## What the fix does

**Split-SiLU is the new DEFAULT for the native-FP8 decode `down` projection.** `dense_ffn.rs`'s FP8 decode branch becomes a three-way arm resolved by `dense_ffn_fp8_down::fp8_down_arm` — a pure function of the resolved handles and the lever, so every combination is pinned by CPU tests:

| arm | what runs | when |
|---|---|---|
| `SplitSilu` (default) | `w8a16_gemv_dual` → `moe_silu_mul` stages `silu(gate)*up` once → plain `w8a16_gemv` for down | SiLU + dual + `moe_silu_mul` + `w8a16_gemv` all resolve, and `ATLAS_NO_DECODE_SPLIT_SILU` is unset |
| `FusedSilu` | what shipped before: `w8a16_gemv_dual` → `w8a16_gemv_silu_input` | the kill switch, or a target without `moe_silu_mul` |
| `PerProjection` | the 4-launch `w8a16_gemv` ×2 + `moe_silu_mul` + down | no fused kernels, or a non-SiLU activation |

The staging launch is one elementwise pass over K — microseconds — and CUDA graphs amortise the launch. **It uses only kernels that already ship**: `moe_silu_mul` and `w8a16_gemv`, both already on `main` and both already on this exact data in prefill. The arm rule, the nsys table and the ptxas receipt live in the child module `dense_ffn_fp8_down.rs` (`dense_ffn.rs` is already at the CI size cap).

**🪤 NUMERICS, and it is a default-on change.** Split-SiLU is **not** bit-identical to the fused kernel it replaces. `moe_silu_mul` rounds `g*(1/(1+e^-g))*u` to BF16 before the GEMV consumes it, where the fused kernel keeps `(g/(1+e^-g))*u` in FP32 all the way into the dot product — one BF16 activation round, plus a reciprocal-vs-divide difference. Measured on the real shape: `unequal=2130 max_abs=0.500000 max_ulp=31195` (0.5 is the BF16 quantisation step at that exponent, not a large absolute error). **It IS the arithmetic prefill already runs, and the same trade the NVFP4 arm has shipped by default since its own ncu receipt.** `ATLAS_NO_DECODE_SPLIT_SILU` restores the fused kernel bit-for-bit. There is no kernel-level tolerance gate on this path; what covers it is the serving evidence below — four coherency probes, 4/4, in every cell.

## Oracle and evidence

`examples/native_fp8_ffn_down_gemv_microtest`, 1×H100 80 GB, 2026-09-11, at the real Qwen3.8-27B `down` shape. `MT_EXIT=0`.

| shape | route | µs | GB/s | vs OLD |
|---|---|---|---|---|
| **down** N=5120 K=17408 (89.1 MB/layer) | OLD fused `silu_input` | 105.7 | **843** | 1.00× |
| | **NEW split-SiLU + plain GEMV (the new default)** | **58.9** | **1,513** | **1.79×** |

The OLD row reproduces the nsys measurement of the same kernel (105.7 µs / 843 GB/s in the microtest against **103.9 µs / 858 GB/s** in the serving trace, 1.7% apart), so the microtest and the engine agree on the baseline before anything is claimed about the new one.

**The module's own stated target (≥1,700 GB/s / ~52 µs) is still MISSED** — the log prints `staged miss` — so ~12% is left on this kernel and this PR does not claim otherwise.

**The microtest's hard `unequal=0` gate now pins the thing the default arm actually does that nothing else covers.** Production stages the SwiGLU **in place**: `ops::silu_mul(gate_out, up_out, gate_out)` passes one buffer as both the `gate` input and the `output` of a kernel whose parameters are both `__restrict__`. `moe_silu_mul` is one thread per element reading its own index before writing it, so the oracle requires that route to produce the same activation bytes **and** the same projection bytes as staging out of place. (It replaces the `splits=1` bit-identity gate, which went out with the split-K kernel.)

1×H100 80 GB, `Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`, native FP8. Recipe frozen across all cells: `--max-batch-size 16`, ring auto, batched GDN, head band 16, `ATLAS_CUBLAS_GEMM=ffn`, `ATLAS_FFN_NO_BATCH16=1`. Temp 0 / seed 42, thinking off, frozen ladder, 1 warmup + 3 recorded reps. Cell **U** is this PR's default on the new tip; the baseline is the like-for-like predecessor one binary earlier (round 8 serve C on the short shape, round 6 serve I on the long one, which round 8 showed equivalent within one rep spread).

| cell | tok/s agg C=1 | TTFT p50 C=1 | TPOT C=1 | tok/s agg C=16 | TPOT C=16 | spread |
|---|---|---|---|---|---|---|
| baseline | 43.56 | 445.9 ms | 21.28 ms | 227.68 | 55.65 ms | 0.24% |
| **U** (this PR's default) | **51.13** | 443.0 ms | **17.89 ms** | **235.47** | **53.42 ms** | 0.15% |

| cell | tok/s agg C=1 | TTFT p50 C=1 | TPOT C=1 | tok/s agg C=16 | TPOT C=16 |
|---|---|---|---|---|---|
| baseline | 37.41 | 1,420.0 ms | 24.00 ms | 177.60 | 66.64 ms |
| **U** | **42.86** | 1,418.6 ms | **20.60 ms** | 182.08 | 64.43 ms |

**C=1 TPOT 21.28 → 17.89 ms (−15.9%) short and 24.00 → 20.60 ms (−14.2%) long, against a 0.15% rep spread.** Aggregate +17.4% and +14.6%. **The arithmetic closes:** the microtest saves 46.8 µs/layer × 64 layers = **3.00 ms**, and the measured TPOT saving is **3.39 ms** — the extra ~0.4 ms is the `down` projection's share of the long-shape step and rep-to-rep placement, not a discrepancy. C=16 carries **+3.4%** (55.65 → 53.42 ms), smaller in proportion because at 16 rows the FFN has already been moved onto cuBLASLt by the recipe.

Coherency gate **4/4 PASS** (`passed: true`; `'391' OK; 'Tokyo' OK; 'rotaregirfer' OK`), 16-way concurrent arithmetic probe **16/16** correct, device memory flat at **73,829 MiB at ready** and back to 0 MiB after shutdown, zero errors in every recorded rep, every `finish_reason: length`.

The split-K cell (**V**, `ATLAS_FFN_DOWN_SPLITK=1`) from the same round measured C=1 TPOT **17.98 ms** and aggregate 50.90 tok/s against U's 17.89 / 51.13 — a 0.50% TPOT loss against a 0.02% C=1 rep spread, and a null at C=16. That cell is why the lever is gone rather than shipped off-by-default; the numbers are recorded here and nothing in the tree depends on them any more.

- **Diagnostic single-H100 evidence, not a perf gate.** One box, one model (`Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`), one quantization, one recipe. Nothing here is a claim about GB10, about other models, or about a released number.
- **The numerics change is default-on and covered only at serving level.** The split-SiLU default is not bit-equal to what it replaces and there is no kernel-level tolerance gate on that path. The coherency probes are what stands behind it. `ATLAS_NO_DECODE_SPLIT_SILU` is the off switch.
- **🪤 The default emits no route line.** The FFN decode path changes representation and gets 1.79× faster, and the boot log says nothing about it — an operator cannot confirm from the log which `down` path a server is running. That is a reporting gap on a default that changes numerics, and it is worth a follow-up rather than a silence.

## Test plan

- [x] `cargo test -p spark-model --features cuda ffn` — **16 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda gemv` — **28 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda down` — **20 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda lever` — **29 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda --lib` — **895 passed, 0 failed, 11 ignored** (888 on `main`, **+7** new arm/dispatch tests)
- [x] `cargo clippy -p spark-model --tests --features cuda` — **clean, 0 warnings emitted**
- [x] `cargo fmt -p spark-model -- --check` — clean
- [x] `python3 scripts/check_kernel_shadows.py` — **`kernel shadow structure: OK (4 hardware trees scanned: gb10, metal, strix, strix-hip)`**
- [x] **`git diff avarok/main -- kernels/` is EMPTY** — the branch no longer touches a single `.cu` file, so there is no PTX gate to run and no shadow tree to mirror
- [x] Tested against real hardware — the H100 microtest, ladder, coherency gate and 16-way probe above (2026-09-11)

## GB10 perf-gate records

This diff touches `crates/` (a PERF_PATH), so `record_covers` invalidates every committed `.benchmarks/` record on landing. **A GB10 re-measure is owed before merge, and it is not a formality here** — the split-SiLU arm is the DEFAULT, so a GB10 native-FP8 serve takes it at M=1 the moment this lands. Every number above is H100; GB10's own arithmetic on this arm is unmeasured.

## Conflicts

Cherry-picked from `fix/w8a16-gemv-long-k-bandwidth` on the fork with `-x` provenance lines, then reduced to the split-SiLU change by the removal commit at the tip. The original cherry-pick took three resolutions, all mechanical:

1. **`crates/spark-model/src/layers/dense_ffn.rs`, the module-declaration tail, twice.** The source branch sits on an integration line that also carries `dense_ffn_w8a8_prefill` (#987), `dense_ffn_batch16_decode` (#988), `dense_ffn_m16_tc` and `dense_ffn_fp8_residency_tests` (#991) as sibling child modules; `main` has none of them, so the `pub mod` block and the `#[cfg(test)] mod` block both conflicted wholesale. Resolved by keeping **only** `fp8_down` and `fp8_down_tests`, and rewording that module's doc comment, whose original text referred to siblings that do not exist here. **The dispatch itself — the struct fields, the kernel registrations and the arm in `forward` — applied with no conflict at all.**
2. **`crates/spark-model/src/layers/ops.rs`, twice** (the `mod` list and the `pub use` list) — the same cause, the same resolution: the source branch's block also declares `w8a16_gemm_m16` and `w8a16_gemv_ncol`, which belong to other work. Both of this branch's additions there have since been removed with the split-K arm, so `ops.rs` is now unchanged from `main`.
3. **`crates/spark-model/Cargo.toml`** — the `[[example]]` list, a pure union with the same shape as the one #988 hit. Only `native_fp8_ffn_down_gemv_microtest` was added.

The source branch's `kernels/hopper/common/` and `kernels/b200/common/` split-K symlinks were dropped at cherry-pick time (`main` has no such trees), and the `kernels/gb10/common/w8a16_gemv_splitk.cu` they pointed at is gone too, for the reasons at the top. The branch's diff against `main` is now Rust-only.

## CLA


TheTom:
Note for reviewers on where the Hopper (sm_90a) target lives: this PR, like #984–#994, is cherry-picked onto `main`, and `main` has no Hopper target — `kernels/hopper/` and `kernels/b200/` exist only on the campaign branch behind PR #895. The Hopper target is not a separate set of kernel sources: it is a `kernels/hopper/HARDWARE.toml` with `arch = "sm_90a"` plus relative symlinks to the same `.cu` files under `kernels/gb10/`, compiled for sm_90a (the arch-specific pieces are a handful of guards, e.g. the NVFP4 MMQ entry points and the warp block-scale MMA are excluded on sm_90a, and the device-arch preflight + PTX gate). The fix PRs are arch-neutral and apply to gb10 and hopper alike; the H100 receipts in their bodies were measured on the campaign branch, which is `main` + the Hopper target + these fixes. A slim "Hopper + B200 target" PR is being cut from #895's original target commits so the sm_90a target can be reviewed without the whole umbrella.

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1038 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1038; this fork PR is closed.

PR #994, opened 2026-09-11
spark-model: W8A8 cuBLASLt decode projections for 5..16 rows on the SSM and attention paths (#927)

TheTom:
> **Stacked on #989 (and #987), and on #984 → #986 → #988.** This PR's own content is the **last four** of the twenty-one commits here; the seventeen below them belong to those five PRs (#987's three and #989's six, then #984's five, #986's one and #988's two). Review this PR alone with `git diff f09940890..`, or just the commit titled `spark-model: W8A8 cuBLASLt decode projections for 5..16 rows on the SSM and attention paths (#927)` plus its three follow-ups. Merge order: **#984 → #986 → #987 → #988 → #989 → this**. Base is `main` rather than a stack branch so this PR does not become unreviewable if any of them is rebased.
>
> Why all five are in the base, in one line each: **#989** supplies `cublas_fp8_proj_prequant`, the k-major VEC128 scale transpose and the per-family `ATLAS_CUBLAS_GEMM` scoping this PR's `ssm`/`attn` bits are read from — without it there is no correct cuBLASLt W8A8 call and no way to arm one family without arming all; **#987** is #989's own base and owns the W8A8 prefill machinery; **#986** owns `qkv_fp8_batch.rs`, the module this PR's Q/K/V arm is inserted into; **#988** owns `ops::ContiguousBatchGemv` and `ops::w8a16_gemv_batch16`, the wrapper/handle pair the extracted SSM tier resolution returns, and the `w8a16_gemv_batch16` tier this PR falls back to. See **Conflicts** below.

## Summary

At 16 concurrent decode rows on 1×H100, **half the step is one family of kernels running at a ninth of the memory roofline.** nsys, `Qwen/Qwen3.8-27B-FP8`, steady-state n=16 step **43.595 ms**: `w8a16_gemv_batch16` and its `_strided` sibling account for **21.85 ms = 50.1%** of it, and the SSM `in_proj_qkvz` alone is 48 launches × 235 µs = **11.29 ms at 357 GB/s** against an HBM3 roofline of 3,350 GB/s.

Meanwhile, in the same step and on the same box, the dense FFN runs **those same 16 rows** through cuBLASLt W8A8 at ~128 µs/layer for 267 MB of weights — roughly 2,100 GB/s-equivalent. The projections were simply never given that path.

This PR gives it to them: the SSM `in_proj_qkvz`/`out_proj` and the attention `q`/`k`/`v`/`o` projections take cuBLASLt W8A8 for 5..16 rows, with a strided `ldc` so the interleaved `[Q|K|V]` decode buffer can be written without staging. It also batches the multi-seq gated RMS norm (**768 → 48 launches**) and removes the attention prefill's BF16-dequant arm, which the recipe's `ATLAS_CUBLAS_GEMM=ffn,ssm,attn` would otherwise re-arm into an off-ledger 2 B/element weight twin.

Measured: **C=16 TPOT 53.42 → 33.97 ms (−36.4%)** and aggregate **235.47 → 346.41 tok/s (+47.1%)** on the short shape, **64.43 → 43.89 ms / 182.08 → 252.93 tok/s** on the long one, with **C=1 TPOT unchanged at 17.88 vs 17.89 ms** as the built-in null check. TTFT moves too, because the SSM and attention *prefill* also land on cuBLASLt: **443.0 → 370.0 ms** and **1,418.6 → 1,174.1 ms**.

Refs #927 and #928.

## What was wrong

nsys, 1×H100 80 GB, `Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`, native FP8, `--max-batch-size 16`, 2026-09-11, steady-state n=16 decode step **43.595 ms** (idle 4.2%), `--cuda-graph-trace=node`. Resolved by grid shape:

| kernel | launches | µs/step | what |
|---|---|---|---|
| `w8a16_gemv_batch16` GrdX 4096 | 48 | **11,294** | SSM `in_proj_qkvz` N=16384 K=5120, 235 µs each @ **357 GB/s** |
| `w8a16_gemv_batch16` GrdX 1280 | 64 | 6,814 | SSM `out_proj` + attn `o_proj`, N=5120, 106 µs each |
| `w8a16_gemv_batch16_strided` GrdX 3072 | 16 | 2,893 | attn `q_proj` N=12288, 181 µs each |
| `w8a16_gemv_batch16_strided` GrdX 256 | 32 | 846 | attn `k_proj` + `v_proj` N=1024, 26 µs each |

**21.85 ms = 50.1% of the step.** The mechanism is not a mystery: `w8a16_gemv_batch16` reads the weight once and then pays **~16 scalar FFMA per weight byte**. The same `in_proj_qkvz` weight read costs **45.3 µs at M=1** (1,852 GB/s) and **235.3 µs at M=16** — 5.2× the time for 1× the bytes. The tier is a bandwidth tier being asked to do arithmetic, and it degrades monotonically as rows are added.

Two smaller things in the same step, both fixed here because the first change forces them:

- **The multi-seq gated RMS norm fires 768 times** — 48 SSM layers × 16 sequences — for 1.612 ms, i.e. **2.1 µs each**. At that size it is launch and tail overhead, not work, and it was the only per-layer kernel left scaling with the row count; its neighbours `gated_delta_rule_decode_f32_strided` and `causal_conv1d_update_l2norm_f32_strided` are already at 48. It is 3.70% of the step.
- **`ctx.dispatch.cublas.attn` routed the attention *prefill* projections to `ops::cublas_bf16_proj`**, which dequantizes the FP8 weight to BF16 and caches the copy — 2 bytes per weight element, allocated outside the buffer ledger. That is the same class of leak that cost the SSM QKVZ arm ~10.3 GiB and killed a 28-token H100 prefill at layer 36 in #917 round 3. It did not matter while nothing armed `attn`. **This PR's serving recipe is `ATLAS_CUBLAS_GEMM=ffn,ssm,attn`**, so arming `attn` for decode would have re-armed that prefill arm.

## What the fix does

**`ops::dispatch_proj_decode` is the SSOT** for the routing rule, the strided-output arithmetic and the kill switch; `qwen3_ssm::decode_w8a8_proj` and `qwen3_attention::…::w8a8_decode` are the two call sites. Per projection: one `per_token_group_quant_fp8` over the normed input, one `fp8_act_scale_to_kmajor` into the VEC128 layout cuBLASLt documents (#989), then `nvjet` against the checkpoint's own FP8 bytes and its 128×128 FP32 block scales, FP32 epilogue. **Zero device memory** — every buffer is arena scratch that already exists, and the weight is consumed in FP8, never dequantized.

**🪤 The strided output is the whole difficulty, and the phantom rows are the trap.** The multi-seq QKV buffer is `[n, per_seq_qkv]` — Q at offset 0 of each sequence's slot, K at `q_proj_bytes`, V after it — so a contiguous `[M, N]` GEMM cannot write it. `fp8_gemm_act_weight_t_blkscaled_ldc` takes the output row pitch, which is exactly what cuBLASLt's D layout already has (`ldc` on a column-major `[N, M]` *is* a row-major `[M, N]`'s row pitch), so one extra parameter buys the strided write with no staging buffer and no scatter kernel. But `cublas_fp8_proj_prequant` hands cuBLASLt `ceil16(M) = 16` for every row count in this band, **and those phantom rows are written**. With a contiguous output they land past the live rows; with this one they land in decode slots `n..16` — slots belonging to sequences not in this step, whose contents are re-projected before anything reads them. That is correct *only while the buffer really has 16 slots*, so the selector bounds the **full padded write extent** (`(m_pad - 1) * ldc + n` elements, from each projection's own base — V's is the tightest) against the arena, and a buffer that cannot hold it **declines** to the `w8a16_gemv_batch16_strided` tier rather than writing past its end.

**The band is 5..16 rows. M=1 decode is untouched** and still runs the bit-exact scalar `w8a16_gemv`; the batch oracles still compare the GEMV tiers to the scalar loop bit-for-bit.

**Every selector is a pure function of the PADDED ctx `n`** (the `padded_batch_n` ladder the CUDA-graph cache is keyed by), the model's resolved `GemmDispatch`, and handles and capacities fixed at model build. Nothing reads the environment per step. `ATLAS_NO_W8A8_DECODE_PROJ` (presence, `OnceLock`-cached) keeps every one of these projections on today's tiers.

**`gated_rms_norm_f32_input_strided`** is the new kernel entry point, `ops::gated_rms_norm_strided` its launcher: `grid = (heads_per_seq, num_seqs, 1)` instead of one launch per sequence. **Bit-identical** to the packed kernel — each block still owns exactly one `(sequence, head)` row, computes the same reduction over the same elements in the same order and writes the same addresses; only the base address differs. That is the same argument `rms_norm_strided` already makes in tree.

**The attention prefill's BF16-dequant arm is removed**, and `prefill_w8a8.rs` replaces it where a replacement is possible. `paged_oproj.rs` gets the cuBLASLt W8A8 arm — same arithmetic the prefill already computes, handed to cuBLASLt instead of the in-tree kernel, no dequant and no allocation. **`cache_skip_qkv.rs` gets no replacement arm and says so in the source**: that path writes q/k/v into three back-to-back regions of one buffer, and cuBLASLt writes `ceil16(M)` rows, so with a prefill token count that is not a multiple of 16 the pad would cross into the next projection's region. It keeps the transposed W8A16 kernels, which are byte-identical to what an un-armed serve already runs.

**🪤 NUMERICS, stated plainly.** These rows move from W8A16 (BF16 activation × E4M3 weight) to W8A8 (dynamic per-token 1×128 E4M3 activation quant × the checkpoint's 128×128 weight scales, FP32 epilogue) — **exactly the arithmetic the dense FFN already uses at these same widths, and vLLM's.** It is a deliberate precision trade, not a defect: E4M3 keeps 3 stored mantissa bits, so the floor of a W8A8-vs-W8A16 comparison on these shapes is ~2–2.6% relative RMS at cosine ~0.9997, and the measurements below sit exactly on that floor.

## Oracle and evidence

`examples/native_fp8_decode_proj_w8a8_microtest`, 1×H100 80 GB, 2026-09-11, at the real Qwen3.8-27B projection shapes. `MT_EXIT=0`. Gates: **`cosine >= 0.999`, `rel_rms <= 3.0e-2`** — a quantisation-noise tolerance, not a bit gate, because the E4M3 activation-quant floor is ~2–2.6e-2 and a bit gate on a deliberate precision change would be meaningless.

**The blocking structural question is answered first: cuBLASLt DOES return an algorithm at `ldc = 14336`.** `q_proj`/`k_proj`/`v_proj` are the strided family (one sequence's `[Q|K|V]` block is 14,336 BF16 elements), they are reported `strided`, they return finite results with intact gaps at every M, and they are the *fastest* arm at M=16.

| M | projection | cosine | rel_rms | phantom rows | verdict |
|---|---|---|---|---|---|
| 5 | ssm `in_proj_qkvz` | 0.999672 | 2.5633e-2 | 5..16 finite, `max_abs=0.000` | PASS |
| 5 | attn `q_proj` (strided) | 0.999674 | 2.5568e-2 | 5..16 finite | PASS |
| 8 | ssm `in_proj_qkvz` | 0.999675 | 2.5525e-2 | 8..16 finite | PASS |
| 16 | ssm `in_proj_qkvz` | 0.999678 | 2.5408e-2 | — | PASS |
| 16 | attn `o_proj` | 0.999680 | 2.5310e-2 | — | PASS |

All 18 cells are in the log: **cosine spans 0.999670–0.999700, rel_rms 2.4516e-2–2.5715e-2.** The deviation is **flat in M** and sits right at the E4M3 floor — which is the thing that says this is activation quantisation and not an indexing bug. And the phantom rows are `finite` with `max_abs=0.000` in the gap regions, i.e. the write extent is where the argument above says it is.

**Four negative controls, all refused, so the harness is shown able to fail before it is believed:**

```
KNOWN_BAD gap:       refused: route wrote outside its extent at byte 458814
KNOWN_BAD guard:     refused: route wrote outside its extent at byte 0
KNOWN_BAD nonfinite: refused: nonfinite projection output (phantom rows included)
KNOWN_BAD value:     refused: cosine 0.068274 < 0.999
```

Timing — µs per projection, 20 reps, `batch16` → W8A8:

| projection | M=5 | M=8 | **M=16** | **M=16 speedup** |
|---|---|---|---|---|
| ssm `in_proj_qkvz` (N=16384 K=5120) | 112.8 → 42.6 | 147.3 → 42.7 | **241.4 → 40.9** | **5.90×** |
| ssm `out_proj` (N=5120 K=6144) | 44.6 → 22.1 | 58.0 → 22.1 | **104.0 → 20.3** | **5.12×** |
| attn `q_proj` (N=12288, strided) | 87.6 → 37.2 | 113.8 → 37.8 | **185.3 → 35.3** | **5.25×** |
| attn `o_proj` (N=5120 K=6144) | 45.1 → 22.3 | 58.7 → 23.0 | **104.3 → 20.4** | **5.11×** |
| attn `k_proj` (N=1024, strided) | 13.3 → **23.7 (LOSS)** | 17.0 → **23.8 (LOSS)** | 27.0 → 21.9 | 1.23× |
| attn `v_proj` (N=1024, strided) | 13.3 → **23.9 (LOSS)** | 17.0 → **23.8 (LOSS)** | 26.9 → 21.9 | 1.23× |

In GB/s-equivalent the whole story is one line: **the batch16 GEMV tier decays with M — `in_proj` 743 → 569 → 348 GB/s as M goes 5 → 8 → 16, because it re-reads the weights per row-group — while cuBLASLt W8A8 is flat at ~2,000 GB/s (1,968 / 1,966 / 2,053).** The GEMV tier is losing a race it structurally cannot win.

**🪤 And the two small-N projections are a real regression at M=5 and M=8.** `k_proj` and `v_proj` (N=1024) go 13.3 → 23.7 µs, **1.8× slower**, because a 1,024-wide output cannot fill a cuBLASLt tile and the fixed launch/heuristic cost dominates. They are only 2 × 26.9 µs of the M=16 step so the bundle still wins by a wide margin, but this is not rounded away: **a per-projection output-width threshold is free money and is not in this PR.**

1×H100 80 GB, `Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`, native FP8. **Same binary in both cells** — X differs from U only by the environment variable `ATLAS_CUBLAS_GEMM=ffn,ssm,attn`, verified live in `/proc/<pid>/environ` rather than merely passed. Recipe otherwise frozen: `--max-batch-size 16`, ring auto, batched GDN, head band 16, `ATLAS_FFN_NO_BATCH16=1`. Temp 0 / seed 42, thinking off, frozen ladder, 1 warmup + 3 recorded reps, rep spread 0.15–0.16%.

| cell | tok/s agg C=1 | TTFT p50 C=1 | TPOT C=1 | **tok/s agg C=16** | TTFT p50 C=16 | **TPOT C=16** |
|---|---|---|---|---|---|---|
| **U** (recipe, `ffn` only) | 51.13 | 443.0 ms | 17.89 ms | 235.47 | 3,758.2 ms | 53.42 ms |
| **X** (`ffn,ssm,attn`) | 51.92 | **370.0 ms** | **17.88 ms** | **346.41** | **3,166.2 ms** | **33.97 ms** |

| cell | tok/s agg C=1 | TTFT p50 C=1 | TPOT C=1 | **tok/s agg C=16** | TTFT p50 C=16 | **TPOT C=16** |
|---|---|---|---|---|---|---|
| **U** | 42.86 | 1,418.6 ms | 20.60 ms | 182.08 | 12,019.6 ms | 64.43 ms |
| **X** | 43.76 | **1,174.1 ms** | 20.60 ms | **252.93** | **9,927.5 ms** | **43.89 ms** |

**C=16 TPOT 53.42 → 33.97 ms (−36.4%), aggregate 235.47 → 346.41 (+47.1%)** short; **64.43 → 43.89 ms (−31.9%) / 182.08 → 252.93 (+38.9%)** long.

**The arithmetic over-closes, and that is the interesting part.** The kernels this replaces are 21.85 ms of the 43.6 ms step and the microtest says they get ~5.2× faster, i.e. ≈17.6 ms removed — against **21.7 ms of TPOT actually removed**. The measurement is *larger* than the decode arithmetic predicts because X also moves the SSM and attention **prefill** to cuBLASLt, which is the second, separately valuable result:

**TTFT drops on every cell.** C=1 short **443.0 → 370.0 ms (−16.5%)**, C=1 long **1,418.6 → 1,174.1 ms (−17.2%)**, C=16 short **3,758 → 3,166 ms**, C=16 long **12,020 → 9,928 ms**.

**The null check passes to three digits.** The route lines say "M=1 decode is untouched", and **C=1 TPOT is 17.88 ms in X against 17.89 ms in U** — a built-in control on a change that must not touch single-row decode, at a width where none of these selectors can engage. C=1 long is 20.60 vs 20.60.

The four decode route lines and the two prefill ones, verbatim from X's boot log:

```
[atlas] SSM in_proj_qkvz decode (n=16 rows, N=16384 K=5120): W8A8 block-scaled via
  cuBLASLt (per-token 1x128 act scales x 128x128 weight scales, FP32 epilogue;
  vLLM-equivalent FP8 numerics), replacing w8a16_gemv_batch16.
  ATLAS_NO_W8A8_DECODE_PROJ restores the GEMV tier; M=1 decode is untouched.
[atlas] SSM out_proj decode (n=16 rows, N=5120 K=6144): …
[atlas] attention q/k/v decode (n=16 rows, K=5120): …
[atlas] attention o_proj decode (n=16 rows, K=6144): …
[atlas] SSM QKVZ prefill:  W8A8 block-scaled via cuBLASLt …
[atlas] attention prefill: W8A8 block-scaled via cuBLASLt …
[atlas] ATLAS_CUBLAS_GEMM="ffn,ssm,attn" -> cuBLASLt arms ffn=true attn=true ssm=true
  head=false (head has no consumer yet)
```

**Memory is flat to the megabyte and no coherency gate is paid.** Both cells: **73,829 MiB at ready**, +4…+8 MiB across the gates, +8…+36 MiB across the ladder, and back to **0 MiB** after shutdown. KV is identical (`13.6 GB → 22205 blocks → 355,280 max KV tokens`). **Coherency gate 4/4 PASS in both** (`passed: true`; `'391' OK; 'Tokyo' OK; 'rotaregirfer' OK`), 16-way concurrent arithmetic probe **16/16** correct — and its wall time is a free side-measurement that tracks the result: **1.66 s (U) → 1.48 s (X)**. Zero errors in every recorded rep, every `finish_reason: length`.

- **Diagnostic single-H100 evidence, not a perf gate.** One box, one model (`Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`), one quantization, one recipe. Nothing here is a claim about GB10 or a released number.
- **The arm requires `ATLAS_CUBLAS_GEMM=ffn,ssm,attn`.** It is not on by default; an unset or `ffn`-only serve runs today's tiers byte-for-byte. The per-family scoping that makes `ssm,attn` expressible is #989's, which is why that PR is in the base.
- **`k_proj`/`v_proj` (N=1024) are 1.8× SLOWER at M=5 and M=8.** The bundle wins anyway, and the selector has no width threshold today. That is a known follow-up, not a solved problem.
- **The numerics change is real and default-on once the lever is set.** ~2.5e-2 relative RMS at cosine ~0.9997 against W8A16, which is the E4M3 floor and the same trade the dense FFN and vLLM already take. M=1 is untouched. `ATLAS_NO_W8A8_DECODE_PROJ` is the off switch.
- **The attention prefill loses a route it had.** A serve that armed `ATLAS_CUBLAS_GEMM=attn` before this PR got `cublas_bf16_proj` on the prefill projections; after it, `paged_oproj` gets cuBLASLt W8A8 instead and `cache_skip_qkv` gets the in-tree transposed W8A16 kernels. The second is a behaviour change for that configuration, and it is in the direction of not allocating an unledgered BF16 twin of every attention weight.

**Host dispatch tests** pin the selector as a pure function of row count, handles, capacities and the kill switch, without building a `ForwardContext`: the write-extent bound (including the tightest-base V case), the padded-`n` keying, the scope bits, the decline paths, and the gated-norm launch-count arithmetic.

## Test plan

- [x] `cargo test -p spark-model --features cuda --lib ssm` — **146 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda --lib qkv` — **23 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda --lib o_proj` — **10 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda --lib w8a8` — **35 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda --lib cublas` — **6 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda --lib gated_rms` — **4 passed, 0 failed**
- [x] `cargo test -p spark-runtime --features cuda --lib cublas` — **4 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda --lib` — **971 passed, 0 failed, 11 ignored** (943 on the stack base `f09940890`, **+28** new dispatch/extent/launch-count tests)
- [x] `cargo clippy -p spark-model -p spark-runtime -p spark-server --tests --features cuda,nccl` — clean, **0 warnings emitted**
- [x] **PTX gate on `kernels/gb10/common/rms_norm.cu`**, the one kernel source this PR changes — `nvcc --ptx` succeeds at **both** `-arch=sm_121a` and `-arch=sm_90a`, 19 `.visible .entry` symbols each and the new `gated_rms_norm_f32_input_strided` present in both. `-cubin -Xptxas -v`: the new entry uses **30 registers / 128 bytes smem / 0 bytes spill stores / 0 spill loads** at `sm_90a` and **33 registers / 128 B smem / 0 spills** at `sm_121a`
- [x] Tested against real hardware — the H100 microtest, ladder, coherency gate and 16-way probe above (2026-09-11)

## GB10 perf-gate records

This diff touches `crates/` and `kernels/`, both PERF_PATHS, so `record_covers` invalidates every committed `.benchmarks/` record on landing. A GB10 re-measure is owed before merge. **Two of the three changes here are inert on an un-armed GB10 serve and one is not:** the decode and prefill W8A8 arms need `ATLAS_CUBLAS_GEMM` to name `ssm`/`attn`, which no GB10 recipe does today, but **`gated_rms_norm_strided` engages on every multi-seq SSM decode step regardless of any lever**. It is argued bit-identical and its launch count is unit-tested, but it is the part of this PR a GB10 re-measure actually exercises.

## Conflicts

Cherry-picked from `feat/w8a8-cublas-decode-projections` on the fork with `-x` provenance lines.

1. **Cherry-picking onto #989's branch alone did not apply.** Three files conflicted structurally rather than textually: `qkv_fp8_batch.rs` did not exist (it is #986's module, and this PR's Q/K/V arm is an insertion *into* it); `attn/o_proj.rs`'s FP8 branch was the pre-#986/#988 per-token loop rather than the batch ladder this arm is placed ahead of; and `ssm_batched.rs`'s projection blocks — which this PR *extracts* into `ssm_batched_proj.rs` — were the pre-#988 versions, without the `ContiguousBatchGemv` wrapper/handle pairing the extracted tier returns. Resolving those by hand would have meant **rewriting** the change rather than applying it. **#984's five commits, #986's one and #988's two were cherry-picked into the base first**, after which all three files applied with **no conflict at all** and nothing in them was adapted. This is the same resolution #988 itself used for #987.
2. **`crates/spark-model/Cargo.toml`, twice** — both times the `[[example]]` list, and both times a pure union that git could not line up because four PRs add sibling stanzas. Once while building the base (#984's `native_fp8_oproj_batch_microtest` against #989's `native_fp8_ffn_w8a8_microtest`, which had grown a `test = true` line) and once for this PR's own `native_fp8_decode_proj_w8a8_microtest`. Every stanza is present exactly once with its original `required-features`; none was dropped, edited or renamed. The source branch's neighbouring `native_fp8_attn_decode_batch_microtest` stanza belongs to other work and was left out.

Verified after the fact: this PR's own diff (`git diff f09940890..`) is **byte-identical** to the source branch's own diff apart from hunk line numbers and two context lines in files the integration branch has since refactored (`W8A8_PREFILL_KERNELS` constants that are not on `main`). No behaviour was adapted.

## CLA


TheTom:
`cargo test --workspace` was red on `kernel_shadow_detector::no_model_shadow_drops_a_common_kernel`: this branch adds `gated_rms_norm_f32_input_strided` to `kernels/gb10/common/rms_norm.cu`, and ten model directories shadow that file, so the detector reported ten dropped-kernel rows.

Pushed `7a13851` (append-only, no force) which adds the kernel to `[shadow_exempt] norm` in `kernels/gb10/common/KERNEL.toml`, with the reason the table requires.

Why exempt rather than back-port: those ten paths are five distinct sources that implement STANDARD RMS norm (`x * weight / rms`) where `common/` implements offset-from-1 (`x * (1 + weight) / rms`), with 1,100+ lines of divergence each — re-deriving them from `common/` would silently renormalise five models, and appending only this one kernel would add a tenth copy of a function whose sole caller is optional. The kernel is a launch-count optimisation: its one dispatch site is a `try_kernel`, and the documented fallback on handle 0 is the per-sequence loop, which is the arm all ten of those models run today.

The exemption suppresses the build warning only — exempt pairs stay in `TargetPtxSet::shadowed_dropped`, so `kernel_audit::classify_failures` still hard-errors if a model's dispatch ever actually asks for it.

Same fix is on `feat/target-defaults-reconciled`, where the H100 integration line hit the identical red (fourteen paths there, counting the b200/hopper `rms_norm.cu` symlinks).

TheTom:
Merged base #988 forward (merge commit `552926ed4`).

#994 already contained #989's old head (`e5f92d1a9`); what it lacked was #988's "make the batch16 dense-FFN decode tier opt-in" commit, plus the #984/#986/#987 layers under the base's SHAs rather than its own cherry-picked copies — hence the add/add collisions.

Conflicts (10 files). The two sides own **disjoint** files, which decides every one:

* **Base's side** — `dense_ffn.rs`, `dense_ffn_batch16_decode.rs`, `dense_ffn_batch16_decode_tests.rs`, `dense_ffn_kernel_tests.rs`, `qwen3_ssm/tests.rs`: changed by #988's opt-in commit, untouched by #994's own work.
* **#994's own side** — `multi_seq/attn/o_proj.rs`, `multi_seq/mod.rs`, `multi_seq/qkv_fp8_batch.rs`, `qwen3_ssm/trait_decode_multi_seq/ssm_batched.rs`: the base leaves these exactly as #994 inherited them; only #994 changed them.
* `crates/spark-model/Cargo.toml` — union of the `[[example]]` stanzas, no duplicates (checked by parsing the TOML).

PR's own diff unchanged (verified): of the 32 files in #994's own diff, **31 are byte-identical to `7a1385154`** and the 32nd is the `Cargo.toml` union; no file outside that set differs from the base.

Distance from the measured tree (`4f00cf199`) across the 10 conflicted files: **+235 / −577**.

Gates on Spark 1 (GB10, CUDA 13): fmt ok · clippy `-p spark-model` / `-p spark-runtime` (`--features cuda`, `-D warnings`) ok · `cargo test -p spark-model --lib` 973 passed 0 failed · `-p spark-runtime --lib` 301 passed 0 failed · `-p atlas-kernels --no-default-features` ok · CI file-size cap ok.

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1032 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1032; this fork PR is closed.

crates/spark-model/src/layers/ops/dispatch_proj_decode.rs
```diff
@@ -0,0 +1,332 @@
...
+
+/// `ATLAS_NO_W8A8_DECODE_PROJ` kill switch: PRESENCE (any value, including
+/// empty) keeps every 5..16-row decode projection on today's `w8a16_gemv_batch16`
+/// tiers. Presence rather than `=1` for the same reason `ATLAS_FFN_W8A16_ONLY`
+/// is presence-checked — an operator reaches for it while a serve is
+/// misbehaving, and `...=0` meaning "on" is a trap.
+///
+/// `OnceLock`-cached and therefore constant for the life of the process, which
+/// is what makes it safe to branch on under CUDA-graph capture.
+pub fn w8a8_decode_proj_disabled() -> bool {
```
mocogit, line 62: **[P1] Resolve this kill switch per model instead of caching it process-wide.** `GemmDispatch` is deliberately rebuilt with `GemmDispatch::from_env()` for each `TransformerModel` because this process supports swapping models/recipes with different levers. This `OnceLock` is initialized by the first decode and survives every later model load. A later recipe that sets `ATLAS_NO_W8A8_DECODE_PROJ` can therefore still execute this precision-changing W8A8 route (or remain disabled after the flag is removed), defeating the emergency rollback control. Please add the flag to the model-resolved dispatch/config and pass that boolean into `decode_w8a8_selected`; the selector already accepts `disabled`, so graph branching remains stable without process-global state.

mocogit (commented):
Reviewed the PR's own diff (the W8A8 decode-projection changes, not the stacked base). I found one blocking model-lifecycle issue below. The padded write-extent checks, all-or-nothing Q/K/V fallback, shared activation-quant scratch, LoRA continuation, and existing W8A16 fallbacks otherwise look internally consistent.

#994 is closed and these commits were restacked unchanged as #1032, so the same finding applies to the active successor.

PR #995, opened 2026-09-11
spark-model: accumulate FP8 KV calibration over the requested token window before freezing (#919)

### LINKED ISSUE #919: FP8 KV calibration freezes on the first 13-token request despite --fp8-kv-calibration-tokens 256; cached_tokens reported without a prefix-cache hit
## Summary

Two accounting observations from Atlas serve logs on the H100 rental (Qwen3.6-35B-A3B-FP8, hopper target), recorded so they do not get lost. Neither has a fix on the fork; both need confirmation on GB10.

### 1. FP8 KV calibration freezes on the first request

With `--kv-cache-dtype fp8 --fp8-kv-calibration-tokens 256`, the calibration statistics were frozen after the readiness probe's 13-token request, not after 256 observed tokens. If the flag is meant as "accumulate until N tokens", the first tiny request should not end calibration. If it is meant as "calibrate on the first request, cap at N", the name is misleading and the scales are set by whatever the health check sends.

Expected: accumulate until 256 tokens (or expose an explicit warm-up prompt for calibration), and log when the freeze happens and on how many tokens.

### 2. `usage.prompt_tokens_details.cached_tokens` non-zero with no cache hit

A chat completion reported `cached_tokens: 48` while the server log for the same request showed a full prefill and no prefix-cache hit. Either the counter counts something other than prefix-cache reuse (in which case the field name is wrong for OpenAI-compat clients) or it is reading stale state from a previous session slot.

Expected: `cached_tokens` equals the number of prompt tokens served from the prefix cache for that request, zero otherwise.

## Why file these now

Both affect how numbers get read in a benchmark: (1) changes numerical behavior based on the first request a harness happens to send, and (2) invalidates any prefix-cache claim built on the usage field. Neither blocked serving.

TheTom:
## Summary

Two accounting bugs, one per half of #919, both found by reading H100 serve logs next to the usage JSON the same server emitted.

**`--fp8-kv-calibration-tokens 256` calibrated on 13 tokens.** The online FP8 KV calibrator froze the per-tensor K/V scales on the FIRST `observe`, so the flag's token count was decoration: every serve logged `freezing per-tensor scales on the first observed tokens` and meant it — the first observe is the readiness probe, and a 24k-context serve got the numerical basis for its entire KV cache from a handful of tokens of the server's own warmup text. The amax now accumulates **across requests** until the requested count is reached, and the entries the window already wrote are made to agree with the final scale rather than left behind it.

**`usage.prompt_tokens_details.cached_tokens` reported prefix-cache hits that never happened.** The field was stamped from the lookup result, which three paths find and then discard — hence `cached_tokens: 48` next to a full-prefill log line. It now reports the tokens whose KV was actually read out of the cache.

Fixes #919.

### 1. The freeze fired before the window could accumulate

`--fp8-kv-calibration-tokens N` reads like "accumulate until N tokens". It was implemented as "freeze on the first observe, whatever that is".

The freeze-on-first-observe rule was not an accident, and this PR does not remove it. Paged attention reads a sequence's whole history in one pass with ONE `k_scale`/`v_scale`, so a scale that moves under live cache entries dequantizes them through the wrong basis — the 2026-07-25 hardening froze on the first observe precisely to guarantee that the scale which quantized an entry is the scale which dequantizes it. The problem is that "the first observe" is whatever request arrives first, and on every real deployment that is the readiness probe.

Measured on 1×H100, `Qwen/Qwen3.8-27B-FP8`, native FP8, `--kv-cache-dtype fp8 --fp8-kv-calibration-tokens 256 --kv-high-precision-layers auto`: the startup line said `freezing per-tensor scales on the first observed tokens`, and the scales froze on the ~13–31-token readiness/smoke request. 256 was never reached because nothing ever waited for it.

### 2. `cached_tokens` was the lookup length, not the reuse length

`SequenceState::cached_prefix_tokens` is `PrefixMatch::matched_tokens` — the lookup result — and the three report sites (`phase_promote_prefills.rs`, `prefill_a_step.rs`, `prefill_b_step.rs`) stamped `usage.prompt_tokens_details.cached_tokens` straight from it, *before* the code decided whether to use the match. Three paths find a match, `inc_ref` its blocks onto the block table, and then recompute all of its KV anyway:

* a hybrid-SSM model with no usable SSM snapshot (`"…but no SSM snapshot — recomputing all KV"`),
* the exact-leaf snapshot shortcut bypass (default ON),
* a Marconi restore declined below `marconi_min_tokens()` or by the session gate.

The reported `48` is 3 × the default `--block-size 16`, exactly the shape a block-aligned lookup produces. The log was right; the usage field was wrong. This is worse than a cosmetic mismatch for OpenAI-compat clients and for anything that reads the field as a cache-hit signal — `atlas-plugin`'s own concurrency benchmark does exactly that.

### The window accumulates; the window's own writes are requantized

`layers/fp8_calibration/state.rs` is the new SSOT for "when does the scale freeze, and on what data", and it is deliberately GPU-free so the decision is unit-testable without a device. `record()` folds each observation into a running amax, and returns one of three things: `Stage` (still inside the window), `Freeze { k_scale, v_scale, tokens_seen }` (this batch reached it), or `Frozen`. The freeze uses the amax over **every token in the window, the crossing batch included** — so a 13-token probe followed by a 243-token request freezes on the max of both.

Keeping the write-scale-equals-read-scale invariant is the whole difficulty, because entries written *during* the window were quantized at a provisional scale. Three options were considered:

* **(a) hold the window's tokens in BF16** — impossible without a second pool: the FP8 pools are sized at 1 byte/element and attention must be able to read those tokens back *during* the window.
* **(b) rescale the stored FP8 codes in place** — needs a new dequant/requant kernel in all five arch trees (`kernels/{hopper,b200,gb10,strix,strix-hip}`) and loses precision twice.
* **(c) chosen** — `layers/fp8_calibration/staging.rs` keeps the original BF16 K/V of the window's batches plus their `slot_mapping`s, and at the freeze replays them through the **existing** `reshape_and_cache_fp8` kernel at the frozen scale, in write order, before the crossing batch is written. No new kernel, and the rewritten entries are quantized **once, from the original BF16**, rather than rescaled in the FP8 code domain.

Replay is chronological, so "last writer wins" per slot is identical to the original write order even if a block was freed and re-allocated inside the window. Cost is `window_tokens × num_kv_heads × head_dim × 2 B × 2` device bytes per attention layer, freed at the freeze — 512 KiB/layer at 256 tokens × 4 KV heads × 128 dims — and `MAX_STAGED_TOKENS = 4096` clamps a pathological flag value (with a warning, once, from layer 0).

`write_kv_cache.rs` now hands `observe()` an `Fp8KvWriteTarget` describing where the batch is about to go, and still calls it strictly *before* the write, so `effective_fp8_scales()` returns exactly the scale this batch is written with.

Each attention layer logs its own freeze, which is the second half of what #919 asked for:

```
FP8 KV scales frozen after 1198 tokens (requested 256) on attn layer 2:
k_scale=0.064453 (amax=14.438), v_scale=0.038225 (amax=8.562), headroom=2.00;
requantized 30 staged tokens in 4 batches, 0 unstaged
```

The startup line stops lying too: `accumulating per-tensor K/V amax over the first {N} observed tokens (across requests, readiness probe included) before freezing the scales`.

**🪤 The window is a floor, not a target, and this is the honest caveat.** The freeze is evaluated per observation batch, so the first observation that *crosses* the requested count freezes on everything it saw. On the H100 run below, a 1,193-token prefill arrives as one observation and the window closes at **1,198 tokens against a requested 256** — a 4.7× overshoot. That is benign (more calibration data, and the 30 staged tokens are correctly replayed) but a deployment that wants a tight 256-token basis does not get one. Reads *inside* the window use a conservative provisional scale (`PROVISIONAL_SCALE = 2.0`, i.e. headroom up to |x| = 896) rather than a data-derived one; that is coarse, but it is consistent, because every read during the window uses the same value. And `--fp8-kv-calibration-tokens 1` reproduces the old freeze-on-first-observe behaviour exactly — nothing is staged, so nothing is rewritten.

### `cached_tokens` is stamped after the reuse decision

`model/trait_impl/prefix_reuse.rs` is the new SSOT: `reused_prefix_tokens(matched, kv_write_start, skip)` returns `0` when the match was found and thrown away, and `kv_write_start.min(matched)` otherwise — never more than the matched prefix, because an SSM snapshot deeper than the match still only reuses the match. The three lookup sites (`prefill_a.rs`, `prefill_b/prefix_lookup.rs`, `prefill_c.rs`) stamp `SequenceState::reused_prefix_tokens` **after** the skip decision, and the three report sites read that instead.

`cached_prefix_tokens` keeps its meaning and its value untouched: `free_sequence` release, `cache_sequence` double-bump avoidance and the chunked-prefill KV write floor all key off the lookup length, and breaking that would be a refcount bug. The reported count is a second, narrower number.

### FP8 KV calibration — 1×H100, diagnostic, not a GB10 record

1×H100 80 GB, `Qwen/Qwen3.8-27B-FP8`, native FP8, serve flags `--kv-cache-dtype fp8 --fp8-kv-calibration-tokens 256 --kv-high-precision-layers auto --max-seq-len 24576 --max-batch-size 16`, 2026-09-11. **This is a single-H100 diagnostic run on rented hardware, not a GB10 perf-gate record** — see the perf-gate section.

**Before:** every serve logged `freezing per-tensor scales on the first observed tokens` and froze on the ~13–31-token readiness/smoke request.

**After**, and the freeze is checked at three checkpoints rather than one:

| checkpoint | tokens the server has seen | `FP8 KV scales frozen` lines |
|---|---|---|
| at READY (`/v1/models` answering, readiness probe done) | readiness probe only | **0** |
| after a 27-in/4-out smoke request | ~31 | **0** |
| after the first 1,193-token request | 1,198 | **12** |

**Exactly 12 lines fire, one per FP8 attention layer** — `attn layer 2` through `attn layer 13`. That is the boot's own `Selective boundary KV cache: 4/16 attention layers at bf16, rest at fp8`: the four boundary layers have no FP8 scales to calibrate, and they log nothing. Every line reports `0 unstaged`, i.e. the replay is complete and no token was written under the provisional basis and left there.

The per-layer spread is why per-tensor-per-layer calibration is worth doing at all:

| | min | max | ratio |
|---|---|---|---|
| `k_scale` | 0.064453 (layer 2, amax 14.438) | 0.102679 (layer 7, amax 23.000) | 1.59× |
| `v_scale` | 0.028460 (layer 3, amax 6.375) | 0.273438 (layer 12, amax 61.250) | **9.61×** |

V amax spans 6.4 → 61.3 across twelve layers. A single global KV scale has to cover both ends.

**Coherency gate 4/4** on the same serve (`determinism_ok`, `toolcall_ok`, `think_leak_ok`, `known_answer_ok`), 16/16 on the concurrent known-answer probe.

**The anticipated cost measures null.** Calibration keeps decode eager until every layer has frozen, so the first request could in principle pay a graphs-eager penalty. It does not: **C=1 TPOT 17.87 ms on the very first request** (the one the window closes inside) **vs 17.88 ms warm** — because on this shape the freeze lands inside the prefill, before decode starts.

**And the ladder does not move.** Twelve cells against the immediately preceding tip, all inside rep-to-rep spread:

| | previous tip | this change | delta |
|---|---|---|---|
| tok/s agg C=1 | 51.96 | **51.91** | −0.10% |
| TTFT p50 C=1 | 369.3 ms | **371.2 ms** | +0.5% |
| TPOT C=1 | 17.87 ms | **17.88 ms** | +0.06% |
| tok/s agg C=16 | 366.19 | **366.37** | +0.05% |

That is the null result this PR wants: the calibration window changes *when* the scales freeze and *what* they are computed from, and costs nothing in steady state.

### `cached_tokens` — CPU tests, no hardware needed

`model/trait_impl/prefix_reuse.rs` carries **6** unit tests over the decision function, each named for the case it pins:

* `a_discarded_match_reports_zero_not_the_lookup_length` — the literal #919 case, `reused_prefix_tokens(48, 0, false) == 0`.
* `the_exact_leaf_bypass_reports_zero` — the worst version, a full-prompt 4,593-token match recomputed end to end must report 0, not 4,593.
* `a_real_skip_reports_the_reused_prefix`, `an_intermediate_snapshot_reports_only_what_it_skipped`, `the_report_never_exceeds_the_matched_prefix`, `no_match_reports_zero` — the genuine-reuse and clamping directions, so the fix cannot be "return 0 always".

`scheduler/cached_tokens_tests.rs` pins the three report sites themselves: it reads `phase_promote_prefills.rs`, `prefill_a_step.rs` and `prefill_b_step.rs` as source and asserts none of them still binds `cached_prompt_tok` from `cached_prefix_tokens` and that each binds it from `reused_prefix_tokens`. It is a source-level assertion rather than a behavioural one because the report sites have no seam to inject a sequence through; it is **red before the fix**, which is the property that matters.

`atlas-plugin`'s `a_requested_warm_path_requires_observed_cached_tokens` is unchanged by this PR and needs no change — but it is now **testing something real**. It asserts that a requested warm path with a zero `cached_prompt_tokens` on any request is `cache_uncontrolled` and not `comparable`. Against the old server that check could be satisfied by a discarded lookup, so a benchmark could report a controlled warm cache it did not have.

### Calibration state and staging

`layers/fp8_calibration/state_tests.rs` — 8 tests on the pure state machine, including `readiness_probe_counts_toward_the_window_but_does_not_freeze_it`, `a_single_oversized_first_request_freezes_on_all_of_it` (the overshoot above, pinned as intended behaviour rather than left implicit), `many_small_requests_accumulate_across_the_window`, `a_one_token_window_freezes_immediately_like_before` (the pre-#919 escape hatch), `the_frozen_scale_never_shrinks_below_the_pre_freeze_amax` and `later_observes_never_move_a_frozen_scale`.

`layers/fp8_calibration/tests.rs` — 11 tests on the glue, including `the_freeze_replays_the_staged_window_and_releases_its_buffers`, `a_one_token_window_stages_nothing`, `only_plain_fp8_kv_runs_online_calibration`, and the four `graphs_ready_*` cases that pin CUDA-graph readiness against a mix of frozen and warm layers.

## Test plan

All on Spark 1 (GB10), CPU-only, isolated worktree, shared `CARGO_TARGET_DIR`, `ATLAS_SKIP_BUILD=1`, `CUDARC_CUDA_VERSION=13000`, against this branch fetched from the fork at `90e248cca`.

- [x] `cargo test -p spark-model --features cuda fp8_calibration` — **19 passed, 0 failed** (896 filtered out)
- [x] `cargo test -p spark-model --features cuda calib` — **19 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda kv` — **28 + 1 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda prefix_reuse` — **6 passed, 0 failed** (the six `prefix_reuse.rs` cases above)
- [x] `cargo test -p spark-runtime --lib --features cuda kv` — **56 passed, 0 failed**
- [x] `cargo test -p spark-server --bins --features cuda,nccl kv` — **25 passed, 0 failed**
- [x] `cargo test -p spark-server --bins --features cuda,nccl calib` — **2 passed, 0 failed**
- [x] `cargo test -p spark-server --bins --features cuda,nccl cached_tokens` — **1 passed, 0 failed** (the three-report-site source assertion)
- [x] `cargo test -p atlas-plugin concurrency` — **37 + 1 passed, 0 failed**
- [x] `cargo clippy -p spark-model -p spark-server -p spark-runtime -p atlas-plugin --tests --features cuda,nccl` — clean, no warnings emitted
- [x] `cargo doc --workspace --no-deps` — clean. `spark-model` carries `#![deny(warnings)]`, so `rustdoc::private_intra_doc_links` is a hard error there; see the third commit

- [x] Tested against real hardware — the 1×H100 receipt above (2026-09-11). No GB10 serve was run for this PR.

**No `.cu` and no `kernels/` file is touched** — confirmed by `git diff --name-only`. The diff is 21 files, all under `crates/`.

## GB10 perf-gate records

This diff touches `crates/`, a PERF_PATH, so `record_covers` invalidates the committed `.benchmarks/` records the moment it lands, and a GB10 re-measure is owed before merge. That is the required check talking.

The evidence above is a **1×H100 diagnostic**, deliberately labelled as such: it is where the bug was observed and where the fix was verified, and it is the right box for that, but it is not a GB10 record and is not offered as one. The ladder table is included to show the change is performance-neutral on the box it was measured on, not to claim a GB10 number.

## Notes for reviewers

The two halves share an issue and nothing else. They are in one PR because they are one issue and each is small; they can be read independently — part 1 is `layers/fp8_calibration*` plus `write_kv_cache.rs` and the two CLI doc/hint sites, part 2 is `prefix_reuse.rs` plus the three stamp sites and the three report sites.

`fp8_calibration.rs` was a single file and is now a module with `state.rs` (pure decision), `staging.rs` (BF16 staging and replay) and the original glue, which is why the diff shows 526 lines moving in the parent file. The split is what makes the freeze decision testable without a GPU; the 19 tests under the `fp8_calibration` filter all run on CPU.

**Post-freeze EMA recalibration stays OFF and stays unsafe.** `ATLAS_FP8_KV_EMA_RECAL=1` still exists and is still default-off. The staging introduced here covers the calibration window only, so moving a scale *after* the freeze still re-bases every already-written entry; the source says so at the call site.

**One documented gap:** KV spilled to host (`--high-speed-swap`) inside the calibration window is not replayed. The window is a few hundred tokens and closes long before spill pressure, so this is recorded in `staging.rs` rather than fixed.

The first two commits are cherry-picked with `-x` from `fix/919-fp8-kv-calibration-window`, which was cut from an integration branch. **Both applied to `main` clean, with no conflicts and no adaptation.** Four of the touched files have unrelated changes on the integration branch (`qwen3_attention/init.rs`, `cli/validate.rs`, `scheduler/prefill_{a,b}_step.rs`); none of them collide with this PR's hunks, and the `init.rs` hunk — adding `attn_layer_idx` to the `Fp8KvCalibration::new` call so each layer can name itself in its freeze line — does **not** depend on integration-only structure: `attn_layer_idx` is already a parameter of `new_with_gating` on `main`.

The **third commit is new here** and is a docs-only fix the source branch never needed: `cargo doc --workspace --no-deps` is a required check, `spark-model/src/lib.rs` carries `#![deny(warnings)]`, and the calibration module's own docs linked two private items (the `staging` submodule, and `PROVISIONAL_SCALE` from the public `scales()`). Both are now named in plain backticks with a note on where they live — the same treatment `qwen35_dense.rs` already gives `gdn_fp8_arm_selected`. Nothing outside doc comments moves.

## CLA

### DIFF
--- a/crates/spark-model/src/layers/fp8_calibration.rs
+++ b/crates/spark-model/src/layers/fp8_calibration.rs
@@ -2,19 +2,38 @@
 
 //! Online FP8 KV cache scale calibration.
 //!
-//! Tracks running max |K| and max |V| values during the first N tokens of
-//! inference to compute per-tensor scales: `scale = max / 448.0` (mapping the
-//! observed dynamic range to FP8 E4M3 [-448, 448]).
+//! Tracks running max |K| and max |V| during the first
+//! `--fp8-kv-calibration-tokens` tokens of inference to compute per-tensor
+//! scales: `scale = amax * headroom / 448.0` (mapping the observed dynamic
+//! range onto FP8 E4M3 `[-448, 448]`).
 //!
-//! The scale is frozen on the FIRST observation (which runs immediately before
-//! the first KV write) and is then constant for the entire lifetime of every
-//! cache entry. This is required for correctness: FP8 KV round-trips (write
-//! `fp8=bf16/scale`, read `bf16=fp8*scale`) only if the SAME scale quantizes
-//! and dequantizes an entry. Freezing after N warmup tokens (the old design)
-//! wrote early entries at a placeholder scale, then froze to a different value,
-//! silently invalidating all already-written / cached KV — a paged multi-query
-//! read spanning the freeze boundary then read history through the wrong scale
-//! and generation degenerated. See the freeze block in `observe`.
+//! # The invariant
+//!
+//! FP8 KV round-trips (write `fp8 = bf16/scale`, read `bf16 = fp8*scale`) only
+//! if the SAME scale quantizes and dequantizes an entry. Paged / multi-query
+//! attention reads a sequence's whole history in one pass with ONE
+//! `k_scale`/`v_scale`, so a scale that changes under live cache entries
+//! dequantizes them through the wrong basis (~6x error → generation garbage:
+//! loops, empty completions). That is why the 2026-07-25 hardening froze the
+//! scale on the FIRST observe.
+//!
+//! # Atlas #919
+//!
+//! Freezing on the first observe made `--fp8-kv-calibration-tokens 256` a lie.
+//! Every H100 serve log showed
+//! `FP8 KV cache with online calibration (checkpoint ships no k/v scales):
+//!  freezing per-tensor scales on the first observed tokens.`
+//! and the thing being observed was the readiness probe — so a 24k-context
+//! serve ran on scales derived from ~13 tokens.
+//!
+//! The window is now real: the amax accumulates ACROSS requests until
+//! `window_tokens` have been observed, and the batch that reaches the window
+//! freezes on the amax of everything seen, itself included. The invariant is
+//! preserved by REWRITING the entries written inside the window — the window's
+//! BF16 K/V and slot mappings are staged aside and replayed through the
+//! existing `reshape_and_cache_fp8` kernel at the frozen scale (see
+//! [`staging`] for the full tradeoff). A readiness probe therefore counts
+//! toward the window and can never end it on its own.
 //!
 //! Thread safety: uses `parking_lot::Mutex` for interior mutability. The lock
 //! is uncontended (single inference thread) so lock overhead is negligible.
@@ -24,11 +43,19 @@ use parking_lot::Mutex;
 use spark_runtime::gpu::{DevicePtr, GpuBackend, KernelHandle};
 use spark_runtime::kv_cache::KvCacheDtype;
 
-/// FP8 E4M3 max representable magnitude.
-const FP8_E4M3_MAX: f32 = 448.0;
+mod staging;
+mod state;
 
-/// Minimum scale to prevent division by zero or denormalized values.
-const MIN_SCALE: f32 = 1e-12;
+pub use staging::Fp8KvWriteTarget;
+use staging::{KvStaging, MAX_STAGED_TOKENS};
+use state::{CalibrationState, CalibrationStep, POST_FREEZE_OBSERVE_PERIOD};
+
+/// Scale used for the calibration window's own writes, before the freeze.
+///
+/// Covers ±896: models with large norm weights (Gemma-4 26B, Mistral) reach
+/// |K| ~600, which clips at scale 1.0. Those writes are requantized at the
+/// frozen scale when the window closes, so this only has to be SAFE, not tight.
+const PROVISIONAL_SCALE: f32 = 2.0;
 
 /// Whether this KV dtype's write path calls [`Fp8KvCalibration::observe`].
 ///
@@ -48,6 +75,10 @@ pub fn dtype_runs_online_fp8_kv_calibration(kv_dtype: KvCacheDtype) -> bool {
 /// `Some(false)` = still warming. Vacuously true when no layer calibrates.
 /// Must not use `find_map`: a BF16 boundary layer reporting `Some(false)`
 /// would shadow later FP8 layers that already froze.
+///
+/// #919 note: graphs now stay eager for the whole calibration window instead
+/// of one batch. That is the cost of calibrating on the requested token count;
+/// the window is a few hundred tokens, once per process.
 pub fn graphs_ready_after_fp8_kv_cal<I>(states: I) -> bool
 where
     I: IntoIterator<Item = Option<bool>>,
@@ -57,18 +88,10 @@ where
 
 /// Mutable calibration state protected by Mutex for Send + Sync.
 struct CalibrationInner {
-    /// Running max of |K| values observed so far.
-    k_running_max: f32,
-    /// Running max of |V| values observed so far.
-    v_running_max: f32,
-    /// Total tokens processed during calibration.
-    tokens_seen: usize,
-    /// Whether calibration is complete (scales frozen).
-    frozen: bool,
-    /// Calibrated k_scale (set once after warmup).
-    k_scale: f32,
-    /// Calibrated v_scale (set once after warmup).
-    v_scale: f32,
+    state: CalibrationState,
+    staging: KvStaging,
+    /// Set when a batch could not be staged, so the freeze log can say so.
+    unstaged_tokens: usize,
 }
 
 /// Online FP8 KV cache scale calibration tracker for one attention layer.
@@ -77,8 +100,10 @@ struct CalibrationInner {
 /// struct (required by `TransformerLayer` trait).
 pub struct Fp8KvCalibration {
     inner: Mutex<CalibrationInner>,
-    /// Headroom multiplier on the first-observe absmax (CLI `--fp8-kv-headroom`).
-    headroom: f32,
+    /// Attention-layer index, for the freeze log line.
+    attn_layer_idx: usize,
+    /// Tokens the window was asked for, before the `MAX_STAGED_TOKENS` clamp.
+    requested_tokens: usize,
     /// GPU buffer for absmax reduction output: `[1]` f32 for K, `[1]` f32 for V.
     /// Layout: `[k_absmax: f32, v_absmax: f32]` = 8 bytes.
     absmax_buf: DevicePtr,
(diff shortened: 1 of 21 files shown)


TheTom:
Superseded by #1008 — identical tree, re-authored so every commit carries the CLA-signed identity; closing.

tbraun96:
Allowlisted — this should go green on the run this comment triggers.

`CLAAssistant` was red because `c43e7ff40` and `797ee27c7` are authored `tturney@psyguard.ai`, which GitHub maps to the login `tturney-psyguardai` — @TheTom under a second email, not a new contributor (the third commit here, `90e248cca`, is his signed `672332+TheTom@users.noreply.github.com`).

Per @tbraun96's instruction, both of his logins are now in `CLA_ALLOWLIST` — landed on `main` in #1014. That is what the list exists for; its own comment reads *"maintainer-controlled identities are allowlisted: they own the copyright."*

A plain `rerun` was not enough: it replays the workflow file version the original run was created with, which predates #1014. `issue_comment` starts a **fresh** run that reads the current file — hence this comment.

Once green, this PR joins #1012, the composed branch that certifies 13 (then 14) of these PRs in one campaign instead of one each.

TheTom:
Note: the other account was just an accident. no need to add it.

tbraun96:
Understood, and noted — thanks. Leaving it with you.

Two things for the record so nobody re-debugs this:

**The allowlist does work.** The `CLAAssistant` job on the run my previous comment triggered came back **success**, with `TheTom` and `tturney-psyguardai` both in the resolved list. The red you still see on the PR is stale: the pinned action creates/updates the `CLAAssistant` **check-run** only on `pull_request_target`, not on `issue_comment`, so a comment-triggered pass has nowhere to write itself. Any push to this branch (a `synchronize`) will re-run it and the check will go green — including a reauthor, if that is the route you prefer.

**#1012 does not depend on this.** I excluded #995 from the composition precisely because of the red, so it is composed from 13 PRs and is already CI-green apart from the benchmark gate. If you reauthor or push here before the campaign starts, I will fold #995 in and it becomes 14; if not, it can ride a later campaign. Nothing is blocked either way.

@tbraun96 asked for the allowlist explicitly while you and I were both mid-flight, so it is on `main` in #1014. Since you say the second account was an accident, that entry is harmless but unnecessary — happy to drop `tturney-psyguardai` from it in the morning if you would rather the list stayed minimal. I would not remove it unilaterally.

Issue #914, opened 2026-09-06
--check-kernels resolves symbols only; kernels compiled to NO_DEVICE_CODE for the running arch pass the audit and fault at first launch


TheTom:
## Summary

`spark serve --check-kernels` resolves kernel symbols and reports them as usable. It does not verify that the module actually contains device code for the running GPU's compute capability. A kernel whose body is compiled out (`NO_DEVICE_CODE` behind an arch predicate) passes the audit and then faults at first launch.

Observed on an H100 with the Qwen3.8-27B hopper target: audit reported `ok: true, compiled_arch: sm_90a, device_cc: [9,0], modules: 173, lookups: 321, unresolved: 0, expected_absent: 20`, and the first prefill hit `CUDA kernel quantize_mmq_nvfp4_worker has no device code compatible with CUDA arch 900` followed by a sticky CUDA 719. Details and the kernel-side fix are in #913.

## Why it matters

The audit is the preflight gate that is supposed to make "boots on a new SKU" a fast NO-GO instead of a runtime fault. The hopper target inherits the gb10 kernel set by symlink, so any module with an sm_12x-only predicate is a latent trap there and on sm_100.

## Proposed fix

Extend the audit so that for every resolved kernel it confirms device code exists for the current CC. Options, cheapest first:

1. Per module, `cuobjdump --list-elf` / `cuModuleGetFunction` + `cuFuncGetAttribute(CU_FUNC_ATTRIBUTE_BINARY_VERSION)` and compare with the device CC. A stub compiled for the wrong arch shows up here.
2. A zero-work launch (grid 1, block 1, early-return flag) per kernel with a synchronize, on a scratch stream, before the server accepts requests. This catches `NO_DEVICE_CODE` and any other launch-time incompatibility directly.
3. A build-time check: preprocess each `.cu` with `-D__CUDA_ARCH__=<target>` and fail the build if a kernel that the MODEL.toml expects present resolves to `NO_DEVICE_CODE`. The rental-day preprocessor oracle did exactly this for the MMQ module.

Any of these should turn the failure into an audit NO-GO with the kernel name, instead of a 500 followed by a fatal fault.

## Not fixed on the fork

No commit yet. The fork branch `hopper/sm90-target-tdd-2026-09` fixes the specific MMQ module (`9cfce36c`) but leaves the audit as symbol-only.

Issue #913, opened 2026-09-06
Hopper: dense NVFP4 MMQ wrappers exported with no sm_90 device code; first prefill traps (CUDA 719) after --check-kernels passes


TheTom:
## Summary

On an H100 (sm_90, CUDA 13.3 driver, `ATLAS_TARGET_HW=hopper ATLAS_TARGET_MODEL=qwen3.8-27b ATLAS_TARGET_QUANT=nvfp4`), `spark serve` for Qwen/Qwen3.8-27B-FP8 passes `--check-kernels` (173/181 modules, 0 unresolved), boots in ~23 s, then the first prefill destroys the CUDA context:

```
kernels/hopper/qwen3.6-27b/nvfp4/q4k_vendor/quantize_impl.cuh:172: ERROR: CUDA kernel quantize_mmq_nvfp4_worker has no device code compatible with CUDA arch 900. ggml-cuda.cu was compiled for: 900
...
ERROR atlas::fault: cuMemsetD8Async failed (status 719), and a no-op synchronize issued afterwards also failed (CUDA_ERROR_LAUNCH_FAILED (719): unspecified launch failure) — the CUDA context is destroyed.
ERROR spark: Exiting after a fatal GPU fault
```

3,296 repeats of the `no device code` line before the fault. The fault path from #904 behaved correctly (server exits instead of serving 500s forever); the bug is upstream of it.

## Cause

`kernels/gb10/qwen3.6-27b/nvfp4/q4k_vendor/common.cuh` defines `BLACKWELL_MMA_AVAILABLE` only for `1200 <= __CUDA_ARCH__ < 1300` (consumer / GB10 Blackwell). Hopper (900) and datacenter Blackwell (1000) both compile the vendored activation quantizer to `NO_DEVICE_CODE`. `nvfp4_mmq.cu` still exported all 13 public wrappers (`atlas_nvfp4_mmq{16,32,64,128}_{nc,wc}`, `atlas_nvfp4_quantize_bf16`, `atlas_nvfp4_repack`, ...) regardless, so kernel lookup succeeded.

`DenseFfn` treats resolved handles as capability: `finalize_nvfp4_mmq_load` repacked gate/up/down for the W4A4 MMQ arm and freed the W4A16 transposed twins. Prefill then dispatched through the trap. A late runtime disable would have left no fallback weights, so the fix has to be at export/load time.

[...]

PR #996, opened 2026-09-11
kernels: Hopper (sm_90a) and B200 (sm_100a) targets, device-arch preflight, PTX compile gate

TheTom:
## Summary

Atlas has had exactly one NVIDIA target on `main` — GB10, `sm_121f`. This adds the other two the datacentre needs: **`kernels/hopper/` (H100/H200, `sm_90a`)** and **`kernels/b200/` (B200/GB200, `sm_100a`)**, the arch plumbing that makes a one-arch-per-build binary say so out loud, and the compile gate that proves every kernel in the tree assembles for both.

Neither target ships a kernel of its own. Both are relative-symlink mirrors of `kernels/gb10/` — 241 links for hopper across seven model targets, 225 for b200 across five — which works because the GB10 kernels are written to an SM80-class instruction floor and carry no `__CUDA_ARCH__` gating. What is new is real: a per-hardware `[build] extra_nvcc_flags` layer, the PTX-vs-device compatibility rules, a device-ordinal arch preflight that refuses a GPU the binary's PTX cannot run on, and `scripts/hopper_ptx_gate.sh`.

Refs #899, #913, #914. **This supersedes the target half of #895** — that branch is 100+ commits of campaign work with the targets buried in it, which is why tbraun96 could not find the sm90 kernels. Everything here is the target; the harness, the perf work and the decode fixes (#984–#994) stay where they are.

## What was missing

Three separate holes, and the third is the one that bites an operator.

**No second architecture.** `kernels/hopper/` did not exist on `main`. The gb10 kernels compile for `sm_90a` — nothing had ever asked them to.

**No way to find out without a GPU.** The only evidence that a kernel exists for an arch was building the whole binary on a box that has one.

**A mismatch surfaced as an opaque driver error.** Atlas compiles ONE SM architecture per build — no fatbin, no multi-`-gencode`. Point the published GB10 image at an H100 and the driver answers `CUDA_ERROR_NO_BINARY_FOR_GPU` from inside `cuModuleLoadData`, naming neither the arch in the binary nor the card in the box.

## What this adds

- **`crates/atlas-core/src/arch.rs`** — the PTX-vs-device rules, pure and dependency-free: parse `sm_XY[a|f]` NVIDIA's way (last digit is the minor), then `None` → portable (runs at-or-above), `a` → architecture-specific (never forward-compatible), `f` → family-specific (same major family, at or above). It lives in atlas-core because both the CUDA preflight and `--check-kernels` report it, and neither needs a GPU to do so.
- **`crates/spark-runtime/src/cuda_backend/arch_preflight.rs`** — two `cuDeviceGetAttribute` calls and that rule, run **before** `AtlasCudaBackend::new`, because constructing the backend is what loads every PTX module. The capability query is addressed **by ordinal** (`cuDeviceGet`), not by the calling thread's current context: a context-addressed query fails on any thread that did not create the context, which is every thread the preflight runs on. It judges the **verbatim** arch, not the base SM with the suffix stripped — stripping would wave `sm_90a` kernels onto a CC 10.0 device.
- **`crates/atlas-kernels`** — `build.rs` records the arch it handed nvcc, and `HARDWARE.toml` gains `[build] extra_nvcc_flags`, merged least-specific-first under `common/KERNEL.toml` and the model's. Both new targets use it for one define, `ATLAS_NO_WARP_BLOCKSCALE_MMA`.
- **`scripts/hopper_ptx_gate.sh`** — `nvcc --ptx -arch=<arch>` then `ptxas -arch=<arch> -v` over a whole hardware set, resolving each model's file set the way `build.rs` does (including `kernel_source` redirects), writing a JSON ledger and a markdown summary. It **self-tests first, always**: one fixture that must compile and one that must NOT compile for this arch, chosen per arch because no single instruction is absent from every architecture Atlas targets. If either verdict is wrong it refuses to report results. A gate whose failure path has never executed is not evidence.
- **The one define, and why it is a hardware fact.** `moe_w4a16_grouped_gemm.cu`'s W4A4 tail uses the warp-level `mma.sync … kind::mxf4nvf4.block_scale`. Hopper has no NVFP4 datapath at all (`cvt with .e2m1x2` is rejected); B200 *has* NVFP4 but issues block-scaled MMA through `tcgen05`, so the warp-level form is missing there too. Neither arch is a superset of the other, which is why this is one define rather than an arch comparison. The tail is compiled out and its two entry points are declared in `[expected_absent.moe_w4a16]`; both are `try_kernel` lookups behind a default-off opt-in, so what Hopper loses is an FP4 escape hatch, not a serving path. **GB10's PTX for that file is byte-identical across the change.**
- **The MMQ exclusion, which is the CUDA-719 story.** The vendored NVFP4 MMQ entry points are inside `BLACKWELL_MMA_AVAILABLE`. On an arch that does not define it they still *resolve* — to the vendor's `NO_DEVICE_CODE` trap body — so a green symbol audit is followed by `CUDA_ERROR_LAUNCH_FAILED` on the first real dense-27B request (#913, #921). The two Hopper 27B targets declare all 13 under `[expected_absent.nvfp4_mmq]` with a reason each, and `nvfp4_mmq_capability.rs` now checks both hardware sets: GB10 must keep every export, Hopper must explain every one it drops.
- **Typed refusal.** An arch mismatch under `--check-kernels` prints the same `{"atlas_kernel_check": …}` object the audit prints, with `"failing_stage": "arch_preflight"` and the counts it genuinely does not have set to `null` rather than to zero (#914).
- **`atlas-plugin` box classes** — `KNOWN_HARDWARE_IDS` so `--hardware h100` is distinguishable from a typo (one is fixed by running the gate on the box, the other by correcting the spelling), plus `Hardware::gpu_count`, added additively: schema stays 1, `#[serde(default)]`, and every committed `.benchmarks/` record still loads with `gpu_count: None` — unmeasured, not one.
- **`docker/hopper/` and `docker/b200/`**, buildable with no GPU present, architecture-locked on purpose; `.github/workflows/datacenter-binaries.yml` to build the binaries on a hosted runner; and two PR-CI jobs that compile one hopper and one b200 model target on every PR. Deliberately **separate jobs, not matrix legs** on `compile`: turning that job into a matrix renames a required check, and branch protection would then wait forever on a name that no longer exists.

## Commits: what came across, and what did not

32 commits. Twenty-four are cherry-picked `-x` from `hopper/sm90-target-tdd-2026-09` in their original order. Three were applied partially and say so in their own message:

| commit | treatment |
|---|---|
| `9cfce36ce` MMQ exclusion | **Hopper half only.** Its GB10 `nvfp4_mmq.cu` guard and `dense_ffn` load-path test are already on main via #934 (#921). |
| `1166fc154` follow-main merge | **`kernels/` mirrors only**, and regenerated rather than replayed — the symlink set is derived from main's `kernels/gb10/common` (188 entries), not from the campaign branch's. |
| `67b47f299` / `40efc1962` docs | **`docs/HARDWARE.md` only.** The whole `docs/DEPLOYMENT.md` addition documents `scripts/start-node-ep.sh`, the multi-GPU launcher, which is not in this PR; it is dropped and nets to zero. |

Dropped entirely: `c4c1b5626` (discrete-GPU free memory) — already on main via #934 (#904). `e3fb473f8` — receipts under `bench/hopper_ab/**`, campaign harness.

Excluded by scope, not by accident: `bench/campaign/**`, `bench/hopper_ab/**`, `scripts/start-node-ep*`, the grammar/tool fixes, the dense-FFN and decode work (#984–#994), the ladder `--enable-thinking` change, and `docs/campaigns/` content.

Eight commits are new here rather than picked, and each exists because the campaign branch's assumptions did not survive the move to `main`:

- the mirror regeneration above (main gained seven KDA/DSA/GLM5-Next kernels the hopper and b200 trees had no link for);
- `preflight.rs` sat at 499 LoC on main, so the two lines the arch gate needs would have broken the 500-line cap — both `init_gpu_backend` arms move to `preflight/gpu_backend.rs` and `kernel_gate` gains a one-line `gate_device_arch` wrapper, no behaviour change;
- `GateRecord::from_run` lost its sixth argument on main (#934 derives `serve_overrides` from the run record), which the cherry-picked multi-GPU topology test still passed;
- `hopper_ptx_gate_test.sh` case (a) pinned an exact task total that is 188 on main, not 181 — a floor now, since case (b) already pins the exact set;
- every `docs/campaigns/…/receipts/` and `start-node-ep.sh` citation in the picked prose pointed at a path this PR does not carry. They point at the gate that re-derives the number instead — a comment naming a file that is not in the tree is worse than no comment (two commits: the prose, and one stale reference in `docker/datacenter_dockerfiles_test.sh`);
- registering `b200` and `hopper` in `check_kernel_shadows.py` made the certification self-test's dead-override control fail for the *wrong* reason — its synthetic `kernels/` tree lacked the two new sets, so the check refused on missing trees before it could emit `RULE1`. The fixture now creates all six;
- `docs/HARDWARE.md` counted a tree that has since moved — 181 symlinks over 171 sources across five hopper models, against 188 over 178 across seven — and its dated `870 of 871` findings now sit beside today's re-derived totals.

### PTX compile gate — the load-bearing receipt

Spark 1 (GB10, aarch64), CUDA 13.0.88 (`Build cuda_13.0.r13.0/compiler.36424714_0`), 2026-09-11, at this branch's tip. Self-test green on both arches first — `known_good` compiled, the per-arch negative fixture failed — or the gate would have refused to report.

| set | arch | strict | kernels | pass | fail | rejected entry fns |
|---|---|---|---:|---:|---:|---:|
| `hopper` | `sm_90a` | **yes** (`--Werror all-warnings`, as `build.rs`) | 1282 | **1282** | 0 | 0 |
| `b200` | `sm_100a` | no | 906 | **906** | 0 | 0 |

Per model, `hopper`: `deepseek-v4-flash` 192/192, `qwen3.6-27b` 188/188, `qwen3.8-27b` 188/188, `qwen3.6-35b-a3b` 180/180, `nemotron-3-nano-30b-a3b` / `nemotron-super-120b-a12b` / `qwen3-next-80b-a3b` 178/178 each. `b200`: the same five P0 targets, 192 + 180 + 178×3.

That is every kernel on `main` assembling for both new architectures. **Compilation is not correctness** — a green gate says these kernels EXIST for `sm_90a` and `sm_100a`, not that they produce the right numbers or run well, and the gate's own summary says so at the bottom of every report.

### The target has served on real Hopper silicon

One H100 80 GB HBM3 (CC 9.0, driver 580.173.02, CUDA 13.0.88, no co-tenants), `Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`, `spark serve --check-kernels`, both dense profiles, 2026-09-11. **Measured on `hopper/h100-integration-2026-09-10`**, not on this branch — that tree carries these targets plus campaign work this PR excludes, so read it as evidence the Hopper target resolves and boots on real silicon, not as a receipt for this exact diff. The receipt for this diff is the compile gate above.

```json
{"model":"qwen3.8-27b","arch":"sm_90a","compiled_arch":"sm_90a","device_cc":[9,0],
 "quant":"nvfp4","kernel_set_hash":"cf3a802b5855","modules_embedded":188,
 "lookups":262,"unresolved":0,"expected_absent":15,"ok":true,"exit_code":0,
 "unresolved_kernels":[]}
```

| profile | `ATLAS_DENSE_FP8` | exit | lookups | unresolved | expected_absent |
|---|---|---:|---:|---:|---:|
| `native-fp8` | 1 | **0** | 264 | **0** | 15 |
| `nvfp4` | unset | **0** | 262 | **0** | 15 |

`compiled_arch: sm_90a` and `device_cc: [9,0]` agree, so the audit is not measuring a fallback. The 15 excused kernels are exactly the declared `sm_90a` gap — the eight `atlas_nvfp4_mmq{16,32,64,128}_{nc,wc}`, five NVFP4 helpers, and `w4a4_gemm` / `w4a4_gemm_mfast` — i.e. the Blackwell warp-FP4 family Hopper has no datapath for. `w4a4` shows `** lookup FAILED **` in the human table and is then excused against `kernels/hopper/qwen3.8-27b/MODEL.toml`; `unresolved` stays 0.

**What this does not establish.** One box, one model, one session. This is bring-up evidence that the target resolves and boots, not a certification run and not a performance claim. No Hopper or B200 number is asserted anywhere in this PR.

### GB10 is untouched

No `kernels/gb10/**` kernel changes beyond two comments and one `#ifndef` wrapper whose PTX is byte-identical, so **no GB10 perf-gate record moves and none is re-recorded.** No `BENCH.toml` is touched; the hopper and b200 trees deliberately do not mirror `BENCH.toml`, because a perf baseline is a property of the box, not of the kernel set.

## Test plan

All on Spark 1 (GB10, aarch64), isolated worktree, `ATLAS_SKIP_BUILD=1` except the release build, `--jobs 8`, at this branch's tip.

- [x] `scripts/hopper_ptx_gate.sh --hw hopper --model all --jobs 8 --strict` — **1282/1282**, exit 0
- [x] `scripts/hopper_ptx_gate.sh --hw b200 --model all --jobs 8` — **906/906**, exit 0
- [x] `scripts/hopper_ptx_gate_test.sh` — the gate's own self-test, exit 0
- [x] `python3 scripts/check_kernel_shadows.py` — `OK (6 hardware trees scanned: b200, gb10, hopper, metal, strix, strix-hip)`
- [x] `bash docker/datacenter_dockerfiles_test.sh` — **PASS, 17 assertions** (workflow checked in yaml mode)
- [x] `cargo test -p atlas-kernels --no-default-features` — **75 passed, 0 failed**, 7 ignored, across 12 test binaries: `inherited_targets` 12, `blockscale_mma_guard` 7, `target_resolution` 10, `kernel_build_flags` 8, `kernel_target_arch` 6, `kernel_shadow_detector` 5, `inherited_targets_w4a4` 4, `hopper_27b` 3, `nvfp4_mmq_capability` 2, `target_hints` 2, lib 16
- [x] `cargo test -p atlas-core arch` — **18 passed, 0 failed**
- [x] `cargo test -p spark-runtime --lib --features cuda arch` — **10 passed, 0 failed**, 1 ignored
- [x] `cargo test -p atlas-plugin hardware` — **99 passed, 0 failed**
- [x] `cargo test -p atlas-plugin gate::` — **310 passed, 0 failed**
- [x] `cargo test -p spark-server --bins --features cuda,nccl kernel_gate` — **7 passed, 0 failed**
- [x] `cargo clippy -p atlas-kernels -p atlas-core -p atlas-plugin --tests` — clean
- [x] `cargo clippy -p spark-runtime -p spark-server --tests --features cuda,nccl` — clean
- [x] `cargo fmt --all --check` — clean
- [x] **A real build of the new target**, no `ATLAS_SKIP_BUILD`: `ATLAS_TARGET_HW=hopper ATLAS_TARGET_MODEL=qwen3.8-27b ATLAS_TARGET_QUANT=nvfp4 cargo build --release -p spark-server --bin spark --no-default-features --features cuda,nccl` → exit 0, ~3 min warm, an 80,018,888-byte aarch64 ELF. This is the whole chain — `HARDWARE.toml` → `build.rs` → nvcc → embedded PTX — not just the gate's view of it.
- [x] File-size cap: no `.rs` file in `crates/` crosses 500 LoC that was not already allow-listed on `main`.

Not run: anything on B200 silicon (no access), and any performance measurement on either architecture.

---

**Three checks are red for reasons this PR cannot fix, and they are red on #984–#994 too:** `seal status` (waits on a codeowner `/seal`, and the next commit voids it, so it wants the diff settled first), `PR benchmark gate`, and `PR categorize (advisory)`.


TheTom:
Merge-order note: #994 adds `kernels/gb10/common/fp8_scale_transpose.cu`. The hopper and b200 trees in this PR mirror gb10 by symlink, so once both are in, `inherited_targets` expects `kernels/{hopper,b200}/common/fp8_scale_transpose.cu` to exist. Whichever of #994 / #996 lands second needs the two symlinks appended (they cannot go in here today — they would dangle until #994 merges). #1015 already carries them on its own chain, so if #1015 is merged in sequence nothing else is needed; otherwise it is a two-line follow-up.

tbraun96:
## Heads-up: this PR's mirror invariant breaks once #998 lands

Not a defect in this PR on its own — it passes today. But `kernels/hopper/common` and `kernels/b200/common` are symlink mirrors of `kernels/gb10/common`, and **#998 adds a new file to that origin**: `kernels/gb10/common/fp8_scale_transpose.cu` (the VEC128 activation-scale transpose the cuBLASLt block-scaled FP8 GEMM needs).

Your mirrors were written before that file existed, so with both branches in one tree this PR's **own** invariant fires:

```
kernels/hopper/common has drifted from kernels/gb10/common:
  fp8_scale_transpose.cu: present in origin, absent from mirror
```

— `every_inherited_common_mirrors_every_gb10_common_file`, `crates/atlas-kernels/tests/inherited_targets.rs:138`.

I hit this while composing both branches into one tree earlier tonight. That composition has been closed in favour of a proper base-chained stack, so I am raising it here rather than letting it vanish with the closed PR. **It will reappear the moment both land, in whatever order.**

The fix is two symlinks, in the same `../../gb10/common/<name>` form every other mirror entry uses:

```bash
ln -s ../../gb10/common/fp8_scale_transpose.cu kernels/hopper/common/fp8_scale_transpose.cu
ln -s ../../gb10/common/fp8_scale_transpose.cu kernels/b200/common/fp8_scale_transpose.cu
```

Verified in the composed tree: `every_inherited_common_mirrors_every_gb10_common_file` went 1 failed → 1 passed.

Happy to push it here if you would rather not context-switch — say the word. It belongs on whichever of #996/#998 lands second, and since the mirrors are yours, this PR is the natural home.

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1045 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1045; this fork PR is closed.

PR #998, opened 2026-09-11
spark-model: route the SSM out_proj and chunk-0 Q/K/V prefill projections through the W8A8 cuBLASLt arm (#928)

TheTom:
> **Stacked on #994** (which is itself stacked on #984 → #986 → #987 → #988 → #989). This PR's own content is the **last** of the twenty-two commits here; the twenty-one below it belong to those six PRs. Review this PR alone with `git diff cac361d7c..`, or just the commit titled `spark-model: route the remaining W8A16 prefill projections through the W8A8 cuBLASLt arm (#928)`. Merge order: **#984 → #986 → #987 → #988 → #989 → #994 → this**. Base is `main` rather than a stack branch so this PR does not become unreviewable if any of them is rebased.
>
> Why #994 is in the base, in one line: it supplies `ops::cublas_fp8_proj_prequant`, the k-major VEC128 activation-scale transpose and the per-family `ATLAS_CUBLAS_GEMM` scoping (via #989) that this PR's two new arms are built out of, and it owns `prefill_w8a8.rs` and `cache_skip_qkv.rs` in the shape this diff extends — including the source comment in `cache_skip_qkv.rs` that says, in #994, that the chunk-0 Q/K/V chain *cannot* take cuBLASLt. **This PR is the answer to that comment.** See **Conflicts** below.

## Summary

**The largest single kernel in an H100 prefill is 100.6 ms of W8A16 GEMM that no route change had ever reached — and it is not what the round-9 report said it was.**

nsys, 1×H100 80 GB, `Qwen/Qwen3.8-27B-FP8`, native FP8, 1193-token prompt, C=1: the prefill is **368.263 ms** of GPU-busy union, and `w8a16_gemm_pipelined` is **100.582 ms over 112 launches (27.31%)**, with `w8a16_gemm_t_m128` a further **12.012 ms over 32**. Round 9 read those 112 launches as the dense FFN's prefill GEMM. **Resolving them by grid geometry says otherwise:** they are the SSM `out_proj` (96 launches, 68.1 ms) and the attention `q_proj` of prefill chunk 0 (16 launches, 32.5 ms), with the `t_m128` 32 being that same chunk's `k_proj`/`v_proj`. The dense FFN is on `nvjet…` cuBLASLt lines at both chunk widths in the very same trace — #987 had already moved it.

So **112.6 ms — 30.6% of the prefill — is two projection sites that the `ssm` and `attn` cuBLASLt scopes were already armed for and simply did not reach.** One had no cuBLASLt arm at all and read a stale `ATLAS_FP8_W8A8=1` env var instead of the dispatch flag; the other documented its own exclusion in the source, because cuBLASLt writes `ceil16(M)` rows and its padded rows would cross into the next projection's region of a shared buffer.

This PR gives both sites the arm, under those same existing scopes, with per-region phantom-row accounting instead of a staging copy. **Measured on 1×H100, same binary, arms toggled by the two kill switches: 1193-token TTFT 371.6 → 267.9 ms (−27.9%)** and 4593-token **1,171.7 → 890.3 ms (−24.0%)**, with **C=1 TPOT unchanged at 17.86 ms** as the built-in null check and coherency 4/4.

Refs #928 and #917.

## What was wrong

nsys, 1×H100 80 GB, `Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`, native FP8, recipe `ATLAS_CUBLAS_GEMM=ffn,ssm,attn ATLAS_LM_HEAD_M16_TC=1`, C=1, `--cuda-graph-trace=node`, 2026-09-11 (run record `nsys-r9-prefill`). Prefill = first kernel of the drive segment to the end of its first `argmax_bf16_batch`; `%busy` is against the GPU-busy union, which matches the independently measured TTFT to **0.3%** (368.263 ms union vs 369.3 ms ladder p50).

| # | kernel | count | µs | % of busy |
|---|---|---|---|---|
| 1 | `w8a16_gemm_pipelined` | 112 | **100 581.9** | **27.31** |
| 2 | `gated_delta_rule_chunk_delta_h_vfused` | 96 | 97 830.3 | 26.57 |
| 3 | `nvjet_sm90_qqtst_128x128_128x6_1x2_h_bz_coopA_algo2…` | 176 | 32 462.8 | 8.82 |
| 7 | `w8a16_gemm_t_m128` | 32 | **12 012.2** | **3.26** |

**The kernels do not carry a label, so the sites were resolved from the launchers' own grid formulas** — `w8a16_gemm_pipelined` launches `(ceil(N/32), ceil(M/128))`, `w8a16_gemm_n128_m128` launches `(ceil(N/128), ceil(M/128))` — against a prompt the scheduler served as two chunks of **1168 + 25** tokens:

| launches | grid | shape | site | µs |
|---|---|---|---|---|
| 48 | 160×10 | N=5120 K=6144 M=1168 | **SSM `out_proj`, chunk 0** | 54 119.7 |
| 16 | 384×10 | N=12288 K=5120 M=1168 | **attn `q_proj`, chunk 0** | 32 466.8 |
| 48 | 160×1 | N=5120 K=6144 M=25 | **SSM `out_proj`, chunk 1** | 13 995.4 |
| 32 | 8×10 | N=1024 K=5120 M=1168 | **attn `k_proj`+`v_proj`, chunk 0** | 12 012.2 |

48 = the GDN layer count, 16 = the attention layer count. **112 launches, 112.6 ms, two sites.**

**🪤 This supersedes the round-9 report's reading of the same 100.6 ms as the dense FFN's prefill GEMM, and the mechanism of that error is worth naming**: the inference came from a `log:ffn_w8a16_prefill` route line, and that line is emitted under a log-once latch — it records what the *first* observation selected, not what the steady state runs. The trace itself shows the FFN's gate/up/down on `nvjet…` cuBLASLt lines at both chunk widths. The fix for that class of mistake is in this diff: **both new route lines are emitted from both arms**, so the absence of a W8A8 line can never be ambiguous between "not armed" and "log lost".

The two sites were excluded for two different reasons, and only one of them was ever written down:

- **`prefill_out_proj_dispatch` had no cuBLASLt arm at all.** Its only W8A8 arm was gated on a bare `ATLAS_FP8_W8A8=1` environment read — not `ctx.dispatch.fp8_blockscaled_prefill`, not any `ATLAS_CUBLAS_GEMM` family. So a serve that armed `ssm` moved `in_proj_qkvz` to cuBLASLt and left **the same layer's other GEMM** on W8A16, where FP8 weight bytes buy memory and none of the FLOPs. At M=1193 that kernel runs at **65.5 TFLOP/s**.
- **`cache_skip_one_proj` documented its own exclusion in place.** The chain writes q/k/v into back-to-back regions of two arena buffers — `q` alone in `qkv_output`, `k` at `ssm_qkvz + 0`, `v` at `ssm_qkvz + num_tokens*kv_dim` — and cuBLASLt writes `ceil16(M)` rows, so `k`'s padded rows land inside `v`'s region. That is a real hazard, and it is what this PR has to answer rather than hand-wave.

Only **chunk 0** was affected on the attention side: the *later* chunks of the same prefill go through `paged_qkv.rs`, whose W8A8 arm the same trace does show (`fp8_gemm_t_blockscaled`, 48 launches at M=25). The split was never "attention has no W8A8" — it was "chunk 0 has no W8A8".

## What the fix does

Both sites take the pattern the dense-FFN, QKVZ and `o_proj` prefill arms already use, under the existing `ctx.dispatch.cublas.ssm` / `.attn` scopes: one `per_token_group_quant_fp8` over the normed input, one `fp8_act_scale_to_kmajor` into the VEC128 layout cuBLASLt documents (#989), then `nvjet` against the checkpoint's own FP8 bytes and its 128×128 FP32 block scales, FP32 epilogue. **No dequant, no weight twin, no device allocation** — every buffer is arena scratch that already exists. **The W8A16 kernels remain the fallback and are what an un-armed serve runs byte for byte.**

**🪤 The phantom rows are the whole difficulty, and the answer is per-region accounting, not a staging copy.** cuBLASLt is handed `ceil16(M)`, and those pad rows **are written**. Per region:

- **`q` is alone in `qkv_output`.** It needs `ceil16(M) * q_proj_dim` BF16 of room; `sizes.rs` now allocates `m_pad` rather than `m` there (~0.4 MB on a 27B). Readers still touch only the real `M`.
- **`k` runs before `v` on the same stream**, so `k`'s pad rows (`M..ceil16(M)`, at most 15) land in `v`'s region and are then **overwritten by `v`'s own real write**. The selector carries an **`m >= 16` clause** precisely so that `ceil16(M) − M <= 15 < M` holds and `v`'s real rows cover every one of them. No reader ever sees `k`'s pad.
- **`v`'s own pad is the only one that survives**, at the tail, so the buffer must hold `(M + ceil16(M)) * kv_dim` BF16. That is the extent `ssm_qkvz` is both **checked against at dispatch** and sized for in `sizes.rs`.

Both buffers are bounds-checked at dispatch; **if either is too small the whole chain declines to the W8A16 kernels** rather than writing past its end. A staging buffer plus copy was the alternative and was rejected: it would have cost a full `[M, kv_dim]` D2D per projection per attention layer per chunk.

**The three projections also share one activation quantization** instead of one each — `paged_qkv.rs` re-quantizes per projection, which the same trace shows as three `per_token_group_quant_fp8` launches per attention layer.

`sizes.rs` also gains `linear_num_value_heads * linear_value_head_dim` in the `max_proj_k` fold. It happens to equal `q_heads * head_dim` (6144) on Qwen3.8-27B so it changes no allocation there, but the SSM `out_proj` arm quantizes into that buffer, and a model whose `value_dim` were the widest contraction would otherwise size it short and silently fall back.

**Kill switches:** `ATLAS_SSM_OUT_W8A16_ONLY` and `ATLAS_ATTN_QKV_W8A16_ONLY`, both **presence** (any value, including empty), `OnceLock`-cached — matching `ATLAS_FFN_W8A16_ONLY`, whose spelling was chosen because an escape hatch reached for mid-incident must not have `=0` mean "on". Nothing reads the environment per step; every selector is a pure function of the prefill row count, the resolved `GemmDispatch`, and handles and capacities fixed at model build.

**🪤 NUMERICS, stated plainly.** These projections move from W8A16 (BF16 activation × E4M3 weight) to W8A8 (dynamic per-token 1×128 E4M3 activation quant × the checkpoint's 128×128 weight scales, FP32 epilogue) — **the same arithmetic the dense-FFN, QKVZ and `o_proj` prefills already use, and vLLM's.** It is a deliberate precision trade, not a defect, and the measurements below sit exactly on the E4M3 floor.

### Microtest — `native_fp8_prefill_proj_w8a8_microtest`

1×H100 80 GB, 2026-09-11, at the four real Qwen3.8-27B prefill shapes, M ∈ {64, 1193}, 10 reps, `MT_EXIT=0`. Gates **`cosine >= 0.999`, `rel_rms <= 3.0e-2`** — a quantisation-noise tolerance, not a bit gate, because the E4M3 activation-quant floor is ~2–2.6e-2 and a bit gate on a deliberate precision change would be meaningless. Reference is the W8A16 kernel each site actually ships (`pipelined` for `out_proj`/`q_proj`, `t_m128` for `k`/`v`).

```
rows [64, 1193], cuBLASLt M pad reaches 1200
```

| M | projection | ref | cosine | rel_rms | phantom rows |
|---|---|---|---|---|---|
| 64 | ssm `out_proj` N=5120 K=6144 | pipelined | 0.999677 | 2.5444e-2 | — (pad = 64) |
| 64 | attn `q_proj` N=12288 K=5120 | pipelined | 0.999679 | 2.5335e-2 | — |
| 64 | attn `k_proj` N=1024 K=5120 | t_m128 | 0.999680 | 2.5319e-2 | — |
| 64 | attn `v_proj` N=1024 K=5120 | t_m128 | 0.999681 | 2.5291e-2 | — |
| 1193 | ssm `out_proj` | pipelined | 0.999680 | 2.5322e-2 | `1193..1200` finite, `max_abs=0.000` |
| 1193 | attn `q_proj` | pipelined | 0.999680 | 2.5323e-2 | `1193..1200` finite, `max_abs=0.000` |
| 1193 | attn `k_proj` | t_m128 | 0.999680 | 2.5303e-2 | `1193..1200` finite, `max_abs=0.000` |
| 1193 | attn `v_proj` | t_m128 | 0.999680 | 2.5310e-2 | `1193..1200` finite, `max_abs=0.000` |

**The deviation is flat in M and pinned at the E4M3 floor (cosine 0.999677–0.999681, rel_rms 2.529e-2–2.544e-2) — which is the thing that says this is activation quantisation and not an indexing bug.** And the phantom rows are finite with `max_abs=0.000` in the gap region, i.e. the write extent is where the argument above says it is.

**Three negative controls, all refused, so the harness is shown able to fail before it is believed:**

```
KNOWN_BAD guard:     refused: route wrote outside its extent at byte 0
KNOWN_BAD nonfinite: refused: nonfinite projection output (phantom rows included)
KNOWN_BAD value:     refused: cosine 0.316981 < 0.999
```

Timing, µs per projection, 10 reps:

| projection | M=64 W8A16 → W8A8 | **M=1193 W8A16 → W8A8** | **M=1193 speedup** |
|---|---|---|---|
| ssm `out_proj` N=5120 K=6144 | 267.6 → 25.0 | **1 145.2 → 127.8** | **8.96×** |
| attn `q_proj` N=12288 K=5120 | 291.8 → 37.3 | **2 066.0 → 168.1** | **12.29×** |
| attn `k_proj` N=1024 K=5120 | 369.5 → 24.7 | **370.3 → 54.7** | **6.77×** |
| attn `v_proj` N=1024 K=5120 | 372.7 → 24.4 | **370.2 → 55.0** | **6.73×** |

In TFLOP/s the story is one line: **the W8A16 tier runs these prefill shapes at 33.8–72.7 TFLOP/s while cuBLASLt W8A8 runs the identical work at 228–893** — and the small-N `k`/`v` pair, which was a *regression* at the decode widths in #994, is a 6.7× win here because at M=1193 there is finally enough work to fill a tile.

**The microtest and the serving trace agree on the baseline, independently.** `out_proj` at M=1193 measures 1 145.2 µs here; nsys measured 48 launches at M=1168 for 54 119.7 µs = **1 127.5 µs each** (1.6% apart). `q_proj` 2 066.0 µs here vs 32 466.8/16 = **2 029.2 µs** (1.8%). `k`/`v` 370.3/370.2 µs here vs 12 012.2/32 = **375.4 µs** (1.4%). That agreement is what licenses the grid-geometry attribution above — three independent shapes, all within 2%.

1×H100 80 GB, `Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`, native FP8, 2026-09-11 round 11. **Same binary in both cells** — the control differs only by `ATLAS_SSM_OUT_W8A16_ONLY=1 ATLAS_ATTN_QKV_W8A16_ONLY=1`, verified live in `/proc/<pid>/environ`. Recipe otherwise frozen: bs16, ring auto, batched GDN, head band 16, `ATLAS_CUBLAS_GEMM=ffn,ssm,attn`, `ATLAS_LM_HEAD_M16_TC=1`, `ATLAS_FFN_NO_BATCH16=1`. Temp 0 / seed 42, thinking off, frozen ladder, 1 warmup + 3 recorded reps, rep spread 0.00–0.35%, **0 errors in every rep**.

**The route lines are the A/B, verbatim from the two boot logs, and only the two new ones move:**

```
A  [atlas] SSM out_proj prefill: W8A8 block-scaled via cuBLASLt … ATLAS_SSM_OUT_W8A16_ONLY restores W8A16. This arm allocates nothing.
B  [atlas] SSM out_proj prefill: W8A16 (BF16 act x FP8 weight). W8A8 cuBLASLt not selected — add `ssm` to ATLAS_CUBLAS_GEMM; see #917/#928.

A  [atlas] attention Q/K/V prefill (chunk 0, cache-skip): W8A8 block-scaled via cuBLASLt, activation quantized once for all three …
B  [atlas] attention Q/K/V prefill (chunk 0, cache-skip): W8A16 (BF16 act x FP8 weight). W8A8 cuBLASLt not selected — add `attn` …

A and B, identical:  [atlas] SSM QKVZ prefill: W8A8 block-scaled via cuBLASLt …
A and B, identical:  [atlas] attention prefill: W8A8 block-scaled via cuBLASLt …
```

| cell | tok/s agg C=1 | **TTFT p50 C=1** | TPOT C=1 | tok/s agg C=16 | **TTFT p50 C=16** | TPOT C=16 |
|---|---|---|---|---|---|---|
| control (arms off) | 51.96 | 371.6 ms | 17.86 ms | 366.37 | 3 142.5 ms | 31.48 ms |
| **this PR** | **53.06** | **267.9 ms** | **17.86 ms** | **429.46** | **2 283.8 ms** | **28.44 ms** |
| | +2.1% | **−27.9%** | **0.00%** | **+17.2%** | **−27.3%** | −9.7% |

| cell | tok/s agg C=1 | **TTFT p50 C=1** | TPOT C=1 | tok/s agg C=16 | **TTFT p50 C=16** | TPOT C=16 |
|---|---|---|---|---|---|---|
| control (arms off) | 43.77 | 1 171.7 ms | 20.60 ms | 263.73 | 9 928.3 ms | 41.27 ms |
| **this PR** | **44.84** | **890.3 ms** | **20.60 ms** | **309.09** | **7 511.5 ms** | **37.07 ms** |
| | +2.4% | **−24.0%** | **0.00%** | **+17.2%** | **−24.3%** | −10.2% |

**🪤 Read the control row carefully — it is assembled from two cells, and the reason is stated rather than smoothed over.** The kill-switch cell measured on this binary is the **1193 / C=1** row (51.96 tok/s, 371.6 ms, 17.86 ms TPOT); the remaining three rows are the previous round's cell at the identical recipe on the previous build (`r10-XY-919`, which itself reproduces round 9's XY to within rep spread across all twelve cells). **What licenses splicing them is that the one cell measured both ways agrees to 0.1%**: kill-switch 51.96 tok/s / 371.6 ms against `r10-XY` 51.91 / 371.2. That is the evidence that the switches restore the prior binary's behaviour, and it is a same-recipe cross-build agreement, not an assumption. The C=16 and 4593-token control cells were **not** re-measured with the kill switches on this build.

**The arithmetic closes to 3%, and that is the strongest single result here.** The four kernel families this PR replaces are 112.59 ms of the 368.26 ms short prefill. Feeding the microtest's measured W8A8 timings back through the same launch counts — 48 × 127.8 + 16 × 168.1 + 32 × 54.85 + 48 × 25.0 — predicts **≈11.8 ms** in their place, i.e. a **≈100.8 ms** saving. **Measured TTFT delta: 103.7 ms.** A prefill-only change, predicted from a kernel microtest, landing within 3% of an independently measured end-to-end latency is the check that says the attribution, the microtest and the serve are all describing the same thing.

**The null check passes exactly.** This PR touches prefill only; **C=1 TPOT is 17.86 ms in both cells** on the short shape and **20.60 ms in both** on the long one. Steady-state decode is untouched to four digits, which is what a prefill-only change must show.

**C=16 TPOT moves anyway, and the reason is not decode.** At 16 concurrent requests the ladder's TPOT window contains the *other* sequences' prefill chunks interleaved with decode steps, so a cheaper prefill shortens it (31.48 → 28.44 ms short, 41.27 → 37.07 ms long). That is scheduling, not arithmetic, and the C=1 rows above are the control that shows it.

**Memory is flat and no coherency gate is paid.** Both cells: **73 829 MiB at ready** and **+6 MiB** across the gates; +24 MiB across this PR's full two-shape ladder. KV identical in both (`13.6 GB → 22205 blocks → 355 280 max KV tokens`), ring identical (`8 slots × 16 seqs × 151.5 MB/seq = 18.94 GB`). **Coherency gate 4/4 PASS in both** (`passed: true`; `'391' OK; 'Tokyo' OK; 'rotaregirfer' OK`), 16-way concurrent arithmetic probe **16/16 correct in both** — and its wall time is a free side-measurement that tracks the result: **1.47 s (control) → 0.97 s (this PR)**.

The `sizes.rs` growth is real but below the resolution of the memory poll: `qkv_output` goes from `m` to `m_pad` rows and `ssm_qkvz` gains `(m + m_pad)` over `2m` in one `max` term — ~0.4 MB each on a 27B, against 73.8 GB resident. Both cells report the same MiB at ready.

- **Diagnostic single-H100 evidence, not a perf gate.** One box, one model (`Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`), one quantization, one recipe. Nothing here is a claim about GB10 or a released number.
- **The arms require `ATLAS_CUBLAS_GEMM` to name `ssm` and `attn`.** They are not on by default; an unset or `ffn`-only serve runs today's W8A16 kernels byte for byte. The per-family scoping that makes `ssm,attn` expressible is #989's.
- **Three of the four control cells are the previous build's**, as stated above, bridged by a 0.1% agreement on the one cell measured both ways. A full four-cell kill-switch control on this binary is owed and is not here.
- **The numerics change is real and default-on once the lever is set.** ~2.5e-2 relative RMS at cosine ~0.99968 against W8A16 — the E4M3 floor, and the same trade three other prefill families already take. The two `_W8A16_ONLY` switches are the off switches.
- **The `m >= 16` clause is load-bearing, not a tuning knob.** Below it the k/v overlap argument does not hold, and the selector declines. It is unit-tested as a clause, not left to a comment.
- **The round-9 report's dense-FFN attribution of this kernel is superseded by this PR's grid-geometry resolution**, corroborated by the three-shape microtest agreement above. Anyone citing round 9 §"Reading the prefill" item 3 should read this section instead.

**Host dispatch tests** pin both selectors clause by clause without building a `ForwardContext`: selected at prefill M with the scope armed and not without, the kill switch, the `m` floor, the scale format, the shape and per-region capacity clauses, and the missing-handle declines — plus the buffer-extent arithmetic the phantom-row argument rests on, as an SSOT shared by the selector and the test.

## Test plan

- [x] `cargo test -p spark-model --features cuda prefill` — **98 passed, 0 failed, 4 ignored**
- [x] `cargo test -p spark-model --features cuda w8a8` — **56 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda ssm` — **156 passed, 0 failed**
- [x] `cargo test -p spark-model --features cuda attn` — **15 passed, 0 failed, 1 ignored**
- [x] `cargo test -p spark-model --features cuda cublas` — **6 passed, 0 failed**
- [x] `cargo build -p spark-model --examples --features cuda,gpu-examples` — clean (builds the new microtest)
- [x] `cargo clippy -p spark-model -p spark-runtime -p spark-server --tests --features cuda,nccl -- -D warnings` — clean, **0 warnings emitted**
- [x] **No `.cu` files touched** — confirmed, `git diff --name-only` over this PR's commit is 11 files, all `.rs`/`.toml`, `CU_COUNT=0`. No kernel source changes, so no PTX gate is owed.
- [x] Tested against real hardware — the H100 microtest, the two-cell ladder, the coherency gates and the 16-way probes above (2026-09-11)

## GB10 perf-gate records

This diff touches `crates/`, a PERF_PATH, so `record_covers` invalidates every committed `.benchmarks/` record on landing and a GB10 re-measure is owed before merge.

**Almost all of this PR is inert on an un-armed GB10 serve, and one part is not.** The two W8A8 arms need `ATLAS_CUBLAS_GEMM` to name `ssm`/`attn`, which no GB10 recipe does today, so on GB10 both sites keep running the same W8A16 kernels they run now. **The exception is `crates/spark-runtime/src/buffers/sizes.rs`**, which grows `qkv_output` to `m_pad` rows and adds one term to each of the `ssm_qkvz` and `max_proj_k` folds **unconditionally, on every target**. That is ~0.4 MB each on a 27B-class model and is pure allocation arithmetic — no kernel, no route — but it is the part of this PR a GB10 re-measure actually exercises, and the part to watch if a GB10 fit is close to its ceiling.

No `kernels/` file is touched, so no PTX gate applies.

## Conflicts

Cherry-picked from `fix/prefill-last-w8a16-gemm` on the fork with an `-x` provenance line, onto #994's branch. **It applied with no conflict at all and nothing was adapted** — the two new modules are new files, and the three edited call sites (`cache_skip_qkv.rs`, `trait_prefill_helper.rs`, `sizes.rs`) are exactly the shapes #994 leaves behind. The `crates/spark-model/Cargo.toml` `[[example]]` stanza for `native_fp8_prefill_proj_w8a8_microtest` is a pure addition alongside #994's own.

One thing a reviewer of #994 should know: **this PR deletes the comment in `cache_skip_qkv.rs` that #994 added saying that chain cannot take cuBLASLt.** That comment was correct when it was written — it is the exclusion this PR's per-region accounting is the answer to — and the two should be read together.

## CLA

### DIFF
--- a/crates/spark-model/Cargo.toml
+++ b/crates/spark-model/Cargo.toml
@@ -303,3 +303,9 @@ required-features = ["cuda", "gpu-examples"]
 [[example]]
 name = "native_fp8_decode_proj_w8a8_microtest"
 required-features = ["cuda", "gpu-examples"]
+
+[[example]]
+# The PREFILL sibling: the SSM out_proj and cache-skip Q/K/V shapes that round 9
+# left on W8A16 (#917/#928). Selector clauses are CPU-tested next to each site.
+name = "native_fp8_prefill_proj_w8a8_microtest"
+required-features = ["cuda", "gpu-examples"]
--- a/crates/spark-model/examples/native_fp8_prefill_proj_w8a8_microtest.rs
+++ b/crates/spark-model/examples/native_fp8_prefill_proj_w8a8_microtest.rs
@@ -1,0 +1,475 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+//! The PREFILL projections that round 9 left on W8A16, at Qwen3.8-27B shapes:
+//! the shipped `w8a16_gemm_pipelined` / `w8a16_gemm_t_m128` kernels against the
+//! W8A8 block-scaled cuBLASLt arm this change adds (#917/#928).
+//!
+//! WHY. nsys on 1xH100, 2026-09-11 round 9, `Qwen/Qwen3.8-27B-FP8`, recipe
+//! `ATLAS_CUBLAS_GEMM=ffn,ssm,attn` (`nsys-r9-prefill`). A 1193-token prefill
+//! is **368.263 ms** of GPU-busy union. `w8a16_gemm_pipelined` is **100.582 ms
+//! = 27.31%** over **112 launches** and `w8a16_gemm_t_m128` a further
+//! **12.012 ms** over 32. The kernels' launchers make the grid a shape decoder
+//! (`(ceil(N/32), ceil(M/128))` and `(ceil(N/128), ceil(M/128))`), and the
+//! 1193-token prompt is served as two chunks, 1168 + 25:
+//!
+//! | launches | grid | shape | site | µs |
+//! |---|---|---|---|---|
+//! | 48 | 160x10 | N=5120 K=6144 M=1168 | SSM `out_proj`, chunk 0 | 54 119.7 |
+//! | 16 | 384x10 | N=12288 K=5120 M=1168 | attn `q_proj`, chunk 0 | 32 466.8 |
+//! | 48 | 160x1 | N=5120 K=6144 M=25 | SSM `out_proj`, chunk 1 | 13 995.4 |
+//! | 32 | 8x10 | N=1024 K=5120 M=1168 | attn `k_proj`+`v_proj`, chunk 0 | 12 012.2 |
+//!
+//! Those are the four shapes below. (The dense FFN is NOT among them: the same
+//! trace has gate/up/down on `nvjet…` cuBLASLt lines at every M.)
+//!
+//! It answers three questions per projection and guesses none:
+//!
+//!   1. NUMBERS. W8A8 quantizes the ACTIVATION to E4M3 per 128-wide K group,
+//!      which the W8A16 kernels do not, so the two are NOT bit-identical and no
+//!      bit gate would be honest. The gate is cosine >= 0.999 and relative RMS
+//!      <= 3e-2 against the W8A16 reference, plus: nothing outside the route's
+//!      extent (sentinel), nothing non-finite.
+//!
+//!      ⚠ THE FLOOR, so a marginal `rel_rms` is read correctly: E4M3 carries
+//!      3 stored mantissa bits, so round-to-nearest costs ~2.5% RMS relative
+//!      error per element and that error does NOT average down over a dot
+//!      product of independent terms. The EXPECTED `rel_rms` is ~2-2.6e-2 —
+//!      3e-2 is one notch of headroom over the floor, not a loose tolerance,
+//!      and cosine is the robust metric. Same floor the FFN and decode-proj
+//!      microtests state. `ATLAS_W8A8_REL_RMS_GATE` overrides it.
+//!
+//!   2. THE PHANTOM ROWS. cuBLASLt is handed `ceil16(M)` and WRITES rows
+//!      `M..ceil16(M)`. At M=1193 that is 1200, i.e. 7 rows past the real
+//!      output — the extent every capacity clause in
+//!      `qwen3_ssm/prefill_out_w8a8.rs` and
+//!      `qwen3_attention/prefill_qkv_w8a8.rs` is written against. This pins
+//!      that they land THERE and nowhere else: finite, inside the padded
+//!      extent, sentinel intact beyond it.
+//!
+//!   3. TIME. Per projection, sync'd over `REPS`: µs per launch and the
+//!      TFLOP/s it implies, for both routes.
+//!
+//! Run (H100): `cargo run --release -p spark-model --features cuda,gpu-examples
+//! --example native_fp8_prefill_proj_w8a8_microtest`. The example calls
+//! cuBLASLt directly, so `ATLAS_CUBLAS_GEMM` is not required here — the serve
+//! spelling is printed at the end for copy-paste.
+
+use anyhow::{Result, ensure};
+use half::bf16;
+use spark_model::layers::ops;
+use spark_model::weight_map::{Fp8Weight, WeightQuantFormat};
+use spark_runtime::cuda_backend::AtlasCudaBackend;
+use spark_runtime::gpu::{DevicePtr, GpuBackend, KernelHandle};
+use std::time::Instant;
+
+const H: usize = 5120; // hidden
+const VALUE_DIM: usize = 6144; // GDN out_proj contract width
+const Q_PROJ_DIM: usize = 12288; // gated [Q|gate]
+const KV_DIM: usize = 1024;
+/// Chunk 0 of the round-9 trace prompt is 1168; 1193 is the whole prompt and is
+/// deliberately NOT a multiple of 16, so the phantom rows are exercised.
+const ROWS: [usize; 2] = [64, 1193];
+const REPS: usize = 10;
+const GUARD: usize = 64;
+const SENTINEL: u8 = 0x5a;
+const COSINE_GATE: f64 = 0.999;
+const REL_RMS_GATE: f64 = 3e-2;
+
+/// Which W8A16 kernel a projection falls back to in the serve — the reference
+/// this compares and times against.
+#[derive(Clone, Copy, PartialEq)]
+enum Ref {
+    /// `w8a16_gemm_pipelined` over the row-major `[N, K]` weight.
+    Pipelined,
+    /// `w8a16_gemm_n128_m128` over the transposed `[K, N]` twin.
+    TransposedM128,
+}
+
+struct Proj {
+    name: &'static str,
+    n: usize,
+    k: usize,
+    reference: Ref,
+    weight: DevicePtr,
+    scale: DevicePtr,
+    weight_t: DevicePtr,
+    scale_t: DevicePtr,
+    act: DevicePtr,
+}
+
+impl Proj {
+    fn flops(&self, m: usize) -> f64 {
+        2.0 * m as f64 * self.n as f64 * self.k as f64
+    }
+    fn fp8w(&self) -> Fp8Weight {
+        Fp8Weight {
+            weight: self.weight,
+            row_scale: self.scale,
+            n: self.n as u32,
+            k: self.k as u32,
+            scale_format: WeightQuantFormat::Fp8BlockScaled,
+        }
+    }
+}
+
+struct Kernels {
+    pipelined: KernelHandle,
+    t_m128: KernelHandle,
+    quant: KernelHandle,
+    kmajor: KernelHandle,
+}
+
+struct Scratch {
+    act_fp8: DevicePtr,
+    act_scale: DevicePtr,
+    act_scale_kmajor: DevicePtr,
+}
+
+fn upload(gpu: &dyn GpuBackend, bytes: &[u8]) -> Result<DevicePtr> {
+    let ptr = gpu.alloc(bytes.len())?;
+    gpu.copy_h2d(bytes, ptr)?;
+    Ok(ptr)
+}
+
+fn values(bytes: &[u8]) -> Vec<f64> {
+    bytes
+        .chunks_exact(2)
+        .map(|x| bf16::from_bits(u16::from_le_bytes([x[0], x[1]])).to_f32() as f64)
+        .collect()
+}
+
+fn metrics(observed: &[u8], baseline: &[u8], live: usize) -> (f64, f64) {
+    let a = values(&observed[GUARD..GUARD + live]);
+    let b = values(&baseline[GUARD..GUARD + live]);
+    let (mut dot, mut na, mut nb, mut se, mut ss) = (0.0, 0.0, 0.0, 0.0, 0.0);
+    for (x, y) in a.iter().zip(b.iter()) {
+        dot += x * y;
+        na += x * x;
+        nb += y * y;
+        se += (x - y) * (x - y);
+        ss += y * y;
+    }
+    let cosine = if na > 0.0 && nb > 0.0 {
+        dot / (na.sqrt() * nb.sqrt())
+    } else {
+        0.0
+    };
+    let rel_rms = if ss > 0.0 { (se / ss).sqrt() } else { 0.0 };
+    (cosine, rel_rms)
+}
+
+/// `live` bytes are what the routes are COMPARED on (rows 0..M); `written`
+/// bytes are everything the W8A8 route may TOUCH (rows 0..ceil16(M)). Beyond
+/// `written` the sentinel must survive.
+fn check(
+    observed: &[u8],
+    baseline: &[u8],
+    sentinel: &[u8],
+    live: usize,
+    written: usize,
+    gate: f64,
+) -> Result<()> {
+    ensure!(
+        observed.len() == sentinel.len() && baseline.len() == sentinel.len(),
+        "output extent mismatch"
+    );
+    for i in 0..sentinel.len() {
+        if !(GUARD..GUARD + written).contains(&i) {
+            ensure!(
+                observed[i] == sentinel[i],
+                "route wrote outside its extent at byte {i}"
+            );
+        }
+    }
+    ensure!(
+        values(&observed[GUARD..GUARD + written])
+            .iter()
+            .all(|x| x.is_finite()),
+        "nonfinite projection output (phantom rows included)"
+    );
+    let (cosine, rel_rms) = metrics(observed, baseline, live);
+    ensure!(cosine >= COSINE_GATE, "cosine {cosine:.6} < {COSINE_GATE}");
+    ensure!(rel_rms <= gate, "rel_rms {rel_rms:.4e} > {gate:.1e}");
+    Ok(())
+}
+
+/// The shipped W8A16 arm for this projection — BF16 activations straight in.
+fn run_w8a16(gpu: &dyn GpuBackend, k: &Kernels, p: &Proj, out: DevicePtr, m: usize) -> Result<()> {
+    match p.reference {
+        Ref::Pipelined => ops::w8a16_gemm_pipelined(
+            gpu,
+            k.pipelined,
+            p.act,
+            p.weight,
+            p.scale,
+            out,
+            m as u32,
+            p.n as u32,
+            p.k as u32,
+            0,
+        ),
+        Ref::TransposedM128 => ops::w8a16_gemm_n128_m128(
+            gpu, k.t_m128, p.act, p.weight_t, p.scale_t, out, m as u32, p.n as u32, p.k as u32, 0,
+        ),
+    }
+}
+
+/// The arm under test, exactly as the two new prefill sites compose it:
+/// `per_token_group_quant_fp8` once, then `cublas_fp8_proj_prequant`.
+fn run_w8a8(
+    gpu: &dyn GpuBackend,
+    k: &Kernels,
+    s: &Scratch,
+    p: &Proj,
+    out: DevicePtr,
+    m: usize,
+) -> Result<()> {
+    ops::per_token_group_quant_fp8(
+        gpu,
+        k.quant,
+        p.act,
+        s.act_fp8,
+        s.act_scale,
+        m as u32,
+        p.k as u32,
+        0,
+    )?;
+    ops::cublas_fp8_proj_prequant(
+        gpu,
+        k.kmajor,
+        s.act_fp8,
+        s.act_scale,
+        s.act_scale_kmajor,
+        &p.fp8w(),
+        out,
+        m as u32,
+        p.n as u32,
+        p.k as u32,
+        0,
+    )
+}
+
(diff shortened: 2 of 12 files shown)


TheTom:
## `a6ddf72af` checked and correct — not a regression, and it is green

Reviewing your direct push rather than waving it through, since it moves a number rather than a line of logic.

`crates/spark-runtime/src/buffers/sizes.rs:438` on this branch is `qkv_output: m_pad * qkv_dim * bf16`, with `m_pad = m.div_ceil(16) * 16` at `:264` and the comment naming the same reason (`ops::cublas_fp8_proj_prequant` hands cuBLASLt `ceil16(M)` and the phantom output rows are written). The arithmetic checks: `ceil16(1) * (16*2 + 2*2) * 256 * 2 = 16 * 9216 * 2 = 294912`. So 294912 is the value the production code is *supposed* to produce, the sibling assertions for `ssm_qkvz` and `ssm_deinterleaved` already carry the padded form, and this one was simply missed in that series. Test-only, expectation moved to match a deliberate sizing decision — not an over-allocation being papered over.

`cargo test --workspace` (3m52s) and `cargo llvm-cov --workspace` (3m26s) are both **pass** on this PR now. That was the whole red.

Keeping it. Thanks for catching it and for writing the derivation into the comment instead of pasting the failure value.

TheTom:
Merged base #994 forward (merge commit `0ca127cd3`). #998 already contained #994's old head; this re-takes it after #994 moved to carry #986/#987/#988/#989.

No conflicts — `crates/spark-model/Cargo.toml` auto-merged and, unlike the #988 merge, produced no duplicate `[[example]]` stanza (verified by parsing the TOML: 55 examples, 0 duplicates).

PR's own diff unchanged (verified): all 12 of #998's files byte-identical to `a6ddf72af` apart from the merged `Cargo.toml` example list, and no file outside that set differs from the base.

Distance from the measured tree (`4f00cf199`) on `Cargo.toml`: **+1 / −43**, and the single "+" is a diff re-anchoring artifact — the content is a strict subset.

Note: this branch's head was `a6ddf72af`, one commit ahead of what the earlier pass recorded; the merge was taken on top of that.

Gates on Spark 1 (GB10, CUDA 13): fmt ok · clippy `-p spark-model` / `-p spark-runtime` (`--features cuda`, `-D warnings`) ok · `cargo test -p spark-model --lib` 994 passed 0 failed · `-p spark-runtime --lib` 301 passed 0 failed · `-p atlas-kernels --no-default-features` ok · CI file-size cap ok.

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1033 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1033; this fork PR is closed.

PR #999, opened 2026-09-11
spark-server: persist the grammar token-mask cache and overlap prewarm with prefill (#918)

### LINKED ISSUE #918: Cold grammar compile for a new tool schema is still ~3.4 s on H100 after the #910 fixes (per-state vocab clone removed, parallel prewarm reverted)

TheTom:
## Summary

**#918 says the cold cost is "~3.4 s per new tool schema". The measurement says it is per *process*, and that changes what the fix has to be.**

Phase split, CPU (M5 Max, `--release`, Qwen3 ByteLevel-BPE tokenizer, 151,669 tokens, the coherency gate's `get_weather` schema through `compile_qwen3_coder_tool_grammar`):

```
engine build (TokenizerInfo)              101.9-128.0 ms   once per process
grammar construction                          4.9-5.5 ms   per schema
top-k mask prewarm, FIRST grammar          596.7-621.2 ms  <-- the cold cost
top-k mask prewarm, same schema again             0.0 ms
top-k mask prewarm, a DIFFERENT schema       14.2-16.2 ms  <-- ~2.5% of the first
matcher construction + first mask fill        0.1-0.3 ms
```

**A second, entirely unrelated tool schema costs ~2.5% of the first**, because xgrammar's Tier-2 `RuleLevelCache` keys masks *structurally* — every JSON tool schema reuses the same string / number / whitespace / punctuation sub-rules. So the candidate #918 lists — "persist compiled masks across processes keyed by (tokenizer, schema) hash" — would have been the wrong key: a schema-keyed file only ever helps an exact repeat and misses the 97% that already transfers.

This PR persists **the rule-level cache**, keyed by the **tokenizer**, so a fresh process starts warm for schemas it has never seen; and it moves the remaining serial prewarm **off the request thread**, where it overlaps the prompt forward pass instead of preceding it.

Measured, same CPU harness, n=3, "tool schema in hand" to "first constrained mask filled": **624.8 ms → 5.2–5.3 ms** for the same schema and **5.0–19.1 ms for a schema this build has never compiled**, with the request thread released after **4.9 ms** instead of 624.8. On 1×H100, `Qwen/Qwen3.8-27B-FP8`, after a restart against a populated cache: first tool-call **TTFT 131.6 ms** (same schema) and **126.1 ms** (a different schema, `lookup_tide_table`).

Fixes #918. Refs #910.

## What was wrong

#918 is the follow-up to #910: #910 fixed the warm tool-call decode path, #918 is what was left — the first request for a new tool schema, 4.8–6.6 s cold on the H100, then **3.38 s** after the three fixes listed in that issue (the per-state vocabulary clone, its allocation oracle, and the revert of the four-worker parallel prewarm). Plain chat first token on the same server is 77 ms.

Two things were wrong, and they are independent:

**1. The masks did not survive the process.** `CompiledGrammar::compile_top_k_masks` is ~99% of the cold path and computes the same rule-level masks every process, from scratch, for a tokenizer that has not changed. `RuleLevelCache` already existed and already carried the cross-*schema* reuse — it just had no way to reach disk. The issue's own "still open" list names persistence as a candidate but keys it by schema, which the phase table above says would help almost nothing.

**2. The whole of it ran on the request thread, before the prompt forward pass.** `compile_grammar_state` runs ahead of `prefill_a_step.rs` / `prefill_b_step.rs`, so all ~621 ms landed *in front of* the first token — while the GPU was about to be busy for 370–1,400 ms with the prefill itself. There was nothing structural about that ordering; it was simply synchronous.

**🪤 And there is a trap in the obvious fix for (2), already sprung once.** Making the prewarm *itself* parallel was tried on the H100 and measured **worse** (5.28–6.57 s across four workers vs 4.82 s serial), because `earley/scan.rs`, `complete.rs` and `predict_fsm.rs` clone a shared `Arc<GrammarData>` in hot state loops. **This PR does not reopen that.** `compile_top_k_masks` stays the serial JIT walk it is in tree and `GrammarCompiler`'s `max_threads` stays recorded-but-unused; only the whole serial walk moves off the request thread, as one worker. No two workers ever touch one grammar.

## What the fix does

**The snapshot, `crates/xgrammar/src/compiler/mask_snapshot.rs`.** One little-endian binary file. The header pins everything a cached mask depends on — `ATLASGM1` magic, format version, the `usize` width (the accepted bitset is written as raw `BitVec` words), the tokenizer fingerprint and the vocabulary size — and a trailing FNV-1a digest covers the whole body. **Any mismatch, truncation or corruption returns `Ok(0)`: a cache MISS, never an error and never a wrong mask.** The version constant is bumped whenever the encoding or the mask semantics change, so a snapshot written by an older build is a miss rather than a mis-decode.

It is keyed by `SnapshotIdentity { tokenizer_fingerprint, vocab_size }` and by nothing else, because a mask *is* a partition of the sorted decoded vocabulary: a different tokenizer, or a different `vocab_size` cap over the same tokenizer, invalidates every entry. There is no schema in the key — that is the whole point.

**The wiring, `crates/spark-server/src/grammar/mask_cache.rs`.** The file lives at `<model dir>/.atlas-grammar-cache/masks-<tokenizer fingerprint>.bin`. It is **loaded once at boot**, off the request path entirely, and **written from the background prewarm thread** after it has warmed a grammar — never from a request thread. `MAX_PERSISTED_MASKS = 512`, most-recently-used first: measured mask footprint is ~19 KB at a 151,669-token vocabulary, so 512 caps the file around 10 MB there and ~16 MB at 248K — several distinct tool grammars' worth, which is all the reuse a structural key can deliver anyway. The in-memory cache's own ~1/3 GiB bound is far more than belongs in a file next to a checkpoint.

**The overlap, `crates/spark-server/src/grammar/prewarm.rs`.** `GrammarState::new_with_hook` starts the serial top-k prewarm on its own thread and returns; **`await_masks` joins it before ANY read of a token mask**, so a constrained sample can never observe a half-warmed cache. The ordering is unit-tested from both sides — that construction does *not* block on the prewarm, and that the first mask fill *does*.

**Controls, all three:**

| variable | effect |
|---|---|
| `ATLAS_GRAMMAR_CACHE=0` | disables persistence entirely — no load, no write |
| `ATLAS_GRAMMAR_CACHE_DIR` | relocates the file, for a read-only model directory or to share one warm cache across several copies of a model with the same tokenizer |
| `ATLAS_GRAMMAR_ASYNC_PREWARM=0` | restores the synchronous prewarm on the request thread |

**Nothing about mask *content* changes.** The snapshot round-trips `AdaptiveTokenMask` byte for byte, and the serve-level test asserts that a snapshot-warmed grammar admits a **bit-identical first-token mask** to a cold one before it asserts anything about speed.

### The tests refuse before they are believed

`crates/xgrammar/src/compiler/mask_snapshot_tests.rs` — the format, in isolation:

- `a_snapshot_round_trips_every_mask_byte_for_byte`
- `a_snapshot_written_for_another_tokenizer_is_a_miss`
- `a_corrupt_or_truncated_snapshot_is_a_miss_not_a_wrong_mask`
- `a_missing_snapshot_file_is_a_miss_not_an_error`
- `a_cache_disabled_compiler_neither_loads_nor_saves`
- `loading_a_snapshot_removes_the_cold_mask_prewarm`
- `a_snapshot_also_warms_a_schema_it_was_never_compiled_for` — **the load-bearing one**: it seeds from `WEATHER_SCHEMA` and then times `SEARCH_SCHEMA`, a schema the snapshot was never written for
- `a_compiled_grammar_can_cross_a_thread_boundary`

`crates/spark-server/src/grammar/tests/` — the serve wiring:

- `a_second_distinct_schema_reuses_the_first_schemas_masks` — asserts on the rule cache's own hit/miss counters that cross-schema reuse is real, which is the premise the whole key choice rests on. It goes to zero the moment the cache is keyed per schema, disabled, or its FSM hashing breaks.
- `a_persisted_snapshot_removes_the_cold_prewarm_for_the_next_process` — two engines against one directory; **bit-equality of the first-token mask is asserted before the ratio**
- `a_snapshot_from_a_different_tokenizer_is_a_miss` — a second vocabulary of the same size but different bytes writes its *own* file and adopts nothing
- `construction_does_not_block_on_the_prewarm_but_the_first_fill_does`
- `awaiting_the_prewarm_is_idempotent_across_every_mask_reading_path`
- `the_kill_switch_restores_the_synchronous_prewarm`

**🪤 Every assertion in the serve tests is a RATIO, never a wall time** — they run on a 1,027-token synthetic vocabulary in a debug build on a contended box, where production vocabularies are 100–250× wider. The second of this PR's two commits exists for exactly that reason: it re-gates the snapshot assertion on the **mask phase** rather than the whole cold path, because the whole path includes a per-engine `TokenizerInfo` build that the snapshot does not and cannot remove, and folding it in made the ratio a measurement of the wrong thing.

The two real-tokenizer harnesses that produced the phase tables above are in the same module, `#[ignore]`d by default because they need a 100K+ tokenizer on disk: `phase_timings_with_a_real_tokenizer` and `cold_vs_snapshot_warm_with_a_real_tokenizer`, reproduced with `QWEN_TOKENIZER_JSON=… cargo test --release … -- --ignored --nocapture`. They are diagnostics, not gates, and they are labelled as such in the harness.

### CPU receipt — the cold/warm pair, same box and build

M5 Max, `--release`, Qwen3 ByteLevel-BPE (151,669 tokens), the coherency gate's `get_weather` schema, n=3, "tool schema in hand" to "first constrained mask filled":

| | ms |
|---|---|
| cold, nothing on disk | **624.8** |
| …of which the request thread is held | **4.9** |
| warm from an on-disk snapshot, same schema | **5.2–5.3** |
| warm from an on-disk snapshot, **a schema this build never compiled** | **5.0–19.1** |

Per-schema, the same harness: `get_weather` 5.5 ms construct + 621.2 ms masks; the same schema again 0.1 + 0.0; `search_docs` (new) 4.9 + 16.0; `run_cmd` (new) 4.9 + 14.2.

### H100 receipt — diagnostic single-H100 evidence

1×H100 80 GB, `Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`, native FP8, 2026-09-11 round 11, recipe bs16 / ring auto / batched GDN / head band 16 / `ATLAS_CUBLAS_GEMM=ffn,ssm,attn` / `ATLAS_LM_HEAD_M16_TC=1` / `ATLAS_FFN_NO_BATCH16=1`, temp 0, `--disable-thinking`. Tokenizer vocabulary **248,077**.

The cache is loaded at boot, and says so:

```
INFO spark::grammar::mask_cache: Grammar: warmed 123 token masks from
  /root/models/Qwen3.8-27B-FP8/.atlas-grammar-cache/masks-a91e0eb405c89759.bin
  — the first tool-call request skips the cold mask compile (#918)
```

Three probes on that restarted process, each a real `tool_calls` completion (`finish_reason: "tool_calls"`, arguments valid):

| probe | schema | TTFT | tool call |
|---|---|---|---|
| **first grammar request of the process** | `get_weather`, the one the cache was populated with | **131.6 ms** | `get_weather` |
| **a schema this process has never seen** | `lookup_tide_table`, different function name and fields | **126.1 ms** | `lookup_tide_table` |
| repeat, in-process warm floor | `get_weather` again | **30.4 ms** | `get_weather` |

**The never-seen schema is not slower than the cached one** (126.1 vs 131.6 ms) — which is the structural-key claim reproduced on the serving path, at a 248K vocabulary, not just on the CPU harness.

The snapshot file grew from **3,845,831 bytes** before the restart to **6,987,204** after the two-schema probe, i.e. the new schema's rule masks were persisted for the next process as well.

**🪤 What this receipt is NOT.** It is a *warm-restart* measurement. **The first cold run of this build was not separately timed**: the cache file was already on disk before the restart, written during the preceding cell's coherency gate (which issues a tool-call request), and the operator did not instrument that run. So this PR has **no same-model, same-build cold number** to put next to the 131.6 ms. The cold/warm *pair* in this PR is the CPU one above.

**🪤 And the 3.38 s baseline in #918 is a different model.** It was measured on Qwen3.6-35B-A3B-FP8 at a 248,320-token vocabulary; this is Qwen3.8-27B-FP8 at 248,077. Same host class and a near-identical vocabulary width, which is why the comparison is worth making at all — but it is a cross-model comparison and should be cited as one, not as a controlled A/B.

**🪤 There is a residual and it is visible in the table.** The first grammar request of a warm process is 131.6 ms against an in-process floor of 30.4 ms. Roughly 100 ms of per-process first-request grammar cost survives the cache. The snapshot load itself is at boot and off the request path, so that ~100 ms is elsewhere — grammar construction, matcher setup, and whatever the prewarm's overlap does not hide. It is not chased here.

### Scope of the evidence

- **Diagnostic single-H100 evidence, not a perf gate.** One box, one model, one recipe.
- **The very first process against a fresh model directory still pays the cold path**, minus whatever the prefill overlap hides. What this PR removes is paying it *again* in every process after that, and paying it *on the request thread*.
- **The serve-level ratio assertions are ratios**, on a synthetic vocabulary, for the reason stated above. The absolute numbers all come from the two `#[ignore]`d real-tokenizer harnesses and from the H100.
- **A read-only model directory silently gets no cache** — the write fails and is swallowed as a miss. `ATLAS_GRAMMAR_CACHE_DIR` is the answer, and it is the reason that control exists.

## Test plan

- [x] `cargo test -p xgrammar` — **833 passed, 0 failed, 1 ignored**
- [x] `cargo test -p spark-server --bins --features cuda,nccl grammar` — **105 passed, 0 failed, 2 ignored** (the two ignored are the real-tokenizer diagnostics, which print `set QWEN_TOKENIZER_JSON to an existing tokenizer.json`)
- [x] `cargo test -p spark-server --bins --features cuda,nccl tool` — **304 passed, 0 failed, 1 ignored**
- [x] `cargo clippy -p xgrammar -p spark-server --tests --features cuda,nccl -- -D warnings` — clean, **0 warnings emitted**
- [x] `cargo doc --workspace --no-deps` — clean, no rustdoc warnings
- [x] Tested against real hardware — the H100 warm-restart probes above (2026-09-11)

One note on how that clippy line was obtained, because the first attempt failed and the failure was not this branch's: the gate worktrees for two PRs shared one `CARGO_TARGET_DIR`, and clippy reused the *other* worktree's `xgrammar` metadata for this one's `spark-server` unit — reported as `no mask_snapshot in compiler` on a tree that plainly has it. `cargo check` on the identical selection passed throughout. After `cargo clean -p xgrammar -p spark-server`, clippy passes with `-D warnings`. Recorded so the next person sharing a target dir across worktrees recognises it in one minute instead of twenty.

## GB10 perf-gate records

This diff touches `crates/`, a PERF_PATH, so `record_covers` invalidates every committed `.benchmarks/` record on landing and a GB10 re-measure is owed before merge.

**Neither change is inert on GB10, and both are expected to move a number in the right direction.** The mask cache engages on any serve whose model directory is writable, on the first tool-call request of every process after the first — GB10 recipes do arm tools, so this is live there. The async prewarm engages on **every** grammar-constrained request regardless of any lever. Nothing touches a kernel, a GEMM route, or a device buffer, and no `kernels/` file or `.cu` is in this diff, so no PTX gate applies; the surface a GB10 re-measure exercises is CPU-side tool-call TTFT, not decode throughput.

## Conflicts

Cherry-picked from `fix/918-grammar-cold-compile` on the fork with `-x` provenance lines, onto `main`. The source branch sits on the H100 integration branch, which has diverged from `main` in this area; one conflict and one adaptation, both in the same place and both disclosed:

1. **`crates/xgrammar/src/compiler/mod.rs` — the module list.** The integration branch carries `crates/xgrammar/src/compiler/prewarm.rs` (a bounded-parallel dispatch helper from the experiment #918 describes and reverts) that `main` does not have, so the module list could not line up. **Resolved by taking `main`'s list plus this PR's `pub mod mask_snapshot;`** and dropping `mod prewarm;`, whose file is not on `main`. Nothing in this PR's diff references it: the commit does not touch `compile.rs` or `compiled_grammar.rs`, where that helper is used on the integration branch, and `crates/spark-server/src/grammar/prewarm.rs` is an unrelated new module of this PR's own.
2. **One doc paragraph in the new `crates/spark-server/src/grammar/prewarm.rs` was adapted**, folded into the first commit. As written on the integration branch it explained the serial choice by pointing at `prewarm_max_threads = 1` in `engine.rs` and at two commit hashes — a field and two commits that do not exist on `main`, where `compile_top_k_masks` is serial by construction and `GrammarCompiler`'s `max_threads` is documented as "retained for API parity … recorded but unused". The paragraph now says that instead. **The measured claim it carries is unchanged** (four workers were worse: 5.28–6.57 s vs 4.82 s), and no code changed.

Everything else applied cleanly. This PR's diff is 18 files, +1,859/−10, identical in shape to the source branch's own.

## CLA

### DIFF
--- a/crates/spark-server/src/grammar/engine.rs
+++ b/crates/spark-server/src/grammar/engine.rs
@@ -21,6 +21,11 @@ use super::extract_ordered_vocab;
 pub struct GrammarEngine {
     pub(super) compiler: GrammarCompiler,
     vocab_size: usize,
+    /// #918: the on-disk token-mask snapshot this engine loads at
+    /// startup and re-writes as the cross-grammar cache grows. `None`
+    /// until [`Self::attach_mask_cache`] runs, and when persistence is
+    /// switched off. See [`super::mask_cache`].
+    pub(super) snapshot: Option<super::mask_cache::MaskSnapshot>,
 }
 
 // SAFETY: GrammarEngine is initialized on the main thread and moved to the
@@ -152,6 +157,7 @@ impl GrammarEngine {
         Ok(Self {
             compiler,
             vocab_size,
+            snapshot: None,
         })
     }
 
--- a/crates/spark-server/src/grammar/mask_cache.rs
+++ b/crates/spark-server/src/grammar/mask_cache.rs
@@ -1,0 +1,220 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! On-disk persistence of the cross-grammar token-mask cache (#918).
+//!
+//! WHAT IS PERSISTED, AND WHY IT IS NOT KEYED BY SCHEMA
+//! ----------------------------------------------------
+//! #918 describes the cold cost as "per new tool schema". The CPU
+//! measurement behind this module says otherwise — it is per *process*.
+//! M5 Max, release, Qwen3 ByteLevel-BPE tokenizer (151,669 tokens), the
+//! coherency gate's `get_weather` schema through
+//! `compile_qwen3_coder_tool_grammar`, single-shot:
+//!
+//! ```text
+//! schema                                   construct   mask prewarm
+//! get_weather(city, days)                     5.5 ms       621.2 ms
+//! get_weather(city, days)  — again            0.1 ms         0.0 ms
+//! search_docs(query, limit) — NEW schema      4.9 ms        16.0 ms
+//! run_cmd(command, timeout) — NEW schema      4.9 ms        14.2 ms
+//! ```
+//!
+//! A second, entirely different schema costs ~2.5% of the first because
+//! xgrammar's Tier-2 `RuleLevelCache` keys masks *structurally*: every
+//! JSON tool schema reuses the same string / number / whitespace /
+//! punctuation sub-rules. So what has to survive a restart is the
+//! RULE-level cache, not a compiled grammar keyed by schema text — a
+//! schema-keyed file would only ever help an exact repeat and would
+//! miss the 97% that already transfers. The snapshot is therefore keyed
+//! by the TOKENIZER (fingerprint + vocab size); see
+//! `xgrammar::compiler::mask_snapshot`.
+//!
+//! WHEN IT IS WRITTEN
+//! ------------------
+//! From the background prewarm thread, after it has warmed a grammar —
+//! never from the request thread. The first grammar-bearing request of
+//! the first process still pays today's cost (minus whatever the
+//! prefill overlap hides); every process after that starts warm, for
+//! schemas it has never seen. Same harness, n=3: the whole cold path
+//! (schema in hand to first constrained mask filled) goes 624.8 ms ->
+//! 5.2-5.3 ms for the same schema and 5.0-19.1 ms for one this build
+//! has never compiled.
+//!
+//! Controls: `ATLAS_GRAMMAR_CACHE=0` disables persistence entirely;
+//! `ATLAS_GRAMMAR_CACHE_DIR` relocates the file (for a read-only model
+//! directory, or to share one warm cache across several model copies of
+//! the same tokenizer).
+
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+
+use xgrammar::compiler::{RuleLevelCache, SnapshotIdentity, mask_snapshot};
+
+use super::engine::GrammarEngine;
+
+/// Directory name created under the model directory.
+const CACHE_DIR_NAME: &str = ".atlas-grammar-cache";
+
+/// Upper bound on masks written to disk, most-recently-used first.
+///
+/// The in-memory rule cache is bounded by ~1/3 GiB, which is far more
+/// than belongs in a file next to a checkpoint. Measured mask footprint
+/// is ~19 KB at a 151,669-token vocabulary (2.36 MB for the 123 masks
+/// of one tool grammar), so 512 caps the snapshot around 10 MB there
+/// and ~16 MB at Qwen3.6's 248,320 — several distinct tool grammars'
+/// worth, which is all the reuse the structural key can deliver anyway.
+const MAX_PERSISTED_MASKS: usize = 512;
+
+/// Everything the background saver needs, with no reference to the
+/// (`!Sync`) compiler.
+pub(super) struct MaskSnapshot {
+    path: PathBuf,
+    identity: SnapshotIdentity,
+    cache: RuleLevelCache,
+    /// Entry count at the last successful write — the save is skipped
+    /// unless the cache has grown since.
+    saved_entries: Arc<AtomicUsize>,
+    /// One writer at a time; a second concurrent prewarm just skips.
+    writing: Arc<AtomicBool>,
+}
+
+/// A callback the background prewarm invokes once a grammar's masks are
+/// warm. Boxed so `state.rs` needs no knowledge of persistence.
+pub type PrewarmHook = Arc<dyn Fn(usize) + Send + Sync>;
+
+/// `ATLAS_GRAMMAR_CACHE=0` (or `false`/`off`/`no`) disables the on-disk
+/// mask cache. Anything else — including unset — enables it.
+pub(super) fn cache_enabled_from(value: Option<&str>) -> bool {
+    !matches!(
+        value
+            .unwrap_or_default()
+            .trim()
+            .to_ascii_lowercase()
+            .as_str(),
+        "0" | "false" | "off" | "no"
+    )
+}
+
+/// Where the snapshot for `fingerprint` lives.
+///
+/// `ATLAS_GRAMMAR_CACHE_DIR` wins when set — a model directory pulled
+/// from a read-only mount cannot host the file, and one warm cache can
+/// legitimately serve several copies of the same checkpoint.
+pub(super) fn snapshot_path(
+    model_dir: &Path,
+    dir_override: Option<&str>,
+    fingerprint: u64,
+) -> PathBuf {
+    let dir = match dir_override.map(str::trim).filter(|s| !s.is_empty()) {
+        Some(d) => PathBuf::from(d),
+        None => model_dir.join(CACHE_DIR_NAME),
+    };
+    dir.join(format!("masks-{fingerprint:016x}.bin"))
+}
+
+impl GrammarEngine {
+    /// Seed the cross-grammar mask cache from a snapshot written by an
+    /// earlier process, and arm the background saver (#918).
+    ///
+    /// Returns the number of masks imported; `0` is a miss (no file, a
+    /// different tokenizer, a different build, or a corrupt file) and is
+    /// not an error — the engine just compiles as it did before.
+    /// Called once, at server startup, off any request path.
+    pub fn attach_mask_cache(&mut self, model_dir: &Path) {
+        if !cache_enabled_from(std::env::var("ATLAS_GRAMMAR_CACHE").ok().as_deref()) {
+            tracing::info!("Grammar: on-disk mask cache disabled (ATLAS_GRAMMAR_CACHE)");
+            return;
+        }
+        let Some(cache) = self.compiler.rule_cache_handle() else {
+            return; // compiler built with caching off — nothing to persist
+        };
+        let identity = self.compiler.snapshot_identity();
+        let path = snapshot_path(
+            model_dir,
+            std::env::var("ATLAS_GRAMMAR_CACHE_DIR").ok().as_deref(),
+            identity.tokenizer_fingerprint,
+        );
+        let imported = match self.compiler.load_mask_snapshot(&path) {
+            Ok(n) => n,
+            Err(e) => {
+                tracing::warn!(
+                    "Grammar: mask snapshot unreadable at {}: {e}",
+                    path.display()
+                );
+                0
+            }
+        };
+        if imported > 0 {
+            tracing::info!(
+                "Grammar: warmed {imported} token masks from {} — the first tool-call \
+                 request skips the cold mask compile (#918)",
+                path.display(),
+            );
+        } else {
+            tracing::info!(
+                "Grammar: no usable mask snapshot at {}; the first grammar of this \
+                 process will compute and persist one (#918)",
+                path.display(),
+            );
+        }
+        self.snapshot = Some(MaskSnapshot {
+            path,
+            identity,
+            cache,
+            saved_entries: Arc::new(AtomicUsize::new(imported)),
+            writing: Arc::new(AtomicBool::new(false)),
+        });
+    }
+
+    /// The callback handed to a request's background prewarm, which
+    /// persists the grown mask cache once that prewarm finishes.
+    /// `None` when persistence is off or the engine was never attached.
+    pub(crate) fn mask_snapshot_hook(&self) -> Option<PrewarmHook> {
+        let snap = self.snapshot.as_ref()?;
+        let (path, identity) = (snap.path.clone(), snap.identity);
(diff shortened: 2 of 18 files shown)


TheTom:
## `f20774b8e` checked and correct — the race is real and the assertion kept its teeth

Reviewed your direct push rather than accepting it on the commit message, because "flaky test fixed" is exactly the shape of change that usually hides a loosened bar.

It does not. `await_snapshot` in `crates/spark-server/src/grammar/tests/cold_compile.rs` matched on the `masks-` prefix alone, and `save_to_file`'s temp name is `path.with_extension("tmp<pid>")` — `masks-<fp>.tmp12345`, which shares that prefix. The wait therefore returned mid-encode and "process 2" opened a `.bin` that did not exist yet, so it recomputed every mask and the test read `cold == warm`. Requiring `.ends_with(".bin")` waits for the rename, which is the first instant the snapshot is readable by another process. The persistence being exercised was always correct; the predicate was letting the test start too early.

The important part: the 3x timing floor is **untouched**. The fix tightens the wait, it does not relax the assertion — which is the opposite of what a convenience fix for a flake usually does.

`cargo test --workspace` (2m40s) and `cargo llvm-cov --workspace` (3m28s) are both **pass** on this PR now.

Keeping it as-is.

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1039 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1039; this fork PR is closed.

PR #1001, opened 2026-09-11
gates: two O.R.A.C.L.E checks — certification state, and cross-hardware kernel reach

tbraun96:
Two classes of expensive mistake, both now blocked.

## 1. `O.R.A.C.L.E::certification_state_check` — S.T.A.T.E

A campaign is **~4.5–5 GPU-hours** and nothing else can use the box while it runs. Nothing asked *is it safe to start*, *is it still safe*, or *is this red real*. Fourteen litmus tests across `pre → begin → during → post → release`, each citing the code that enforces it.

Born from two incidents:

- **2026-08-28** — a nine-hour ten-gate campaign was content-invalidated by a 23-line push **twenty minutes into its final gate**. T9 now runs `campaign-guard.sh` *between and during* gates, and an abort **kills the running gate** rather than letting it finish.
- **2026-09-11 (twice in one day)** — a `/stamp` or `/seal` status was read *before* the bot processed the command; the stale red was reported as a real failure and nearly triggered a re-campaign. **T12 dates the red**: those jobs freeze their outputs for the life of a CI run, so a mark minted after they ran needs a FULL `gh run rerun` — `--failed` cannot work, because they *succeeded while emitting `false`*.

Two tests exist precisely because `--pull-request-gate-check` **cannot answer them**: **T13** (one `git_sha` across every *added* record — coverage and agreement are different questions, and the check reports PASS on a set CI rejects) and **T14** (one Speed-class signer).

Lockfile `.oracle_should_begin_cert` is created atomically via `ln(2)`, read on start, carries the structured JSON verdict, and is **renamed rather than deleted** when stale so the evidence survives. Staleness is liveness-first, never age alone. The skill is explicit that it is a file and not a mutex — the residual TOCTOU window and per-checkout scope are written down, not glossed.

Overrides are advisory but never silent: the oracle still **runs** the overridden test and prints both verdicts. T0/T1/T3/T9/T13, the Speed-signer half of T14, and T12-in-post cannot be overridden — CI rejects those outcomes regardless, so an override would only buy wasted hours.

## 2. `O.R.A.C.L.E::pre_commit_cross_hardware_check` — CHKI

`kernels/<hw>/` **looks** like one tree per hardware and is not:

| tree | real files | symlinks |
|---|---|---|
| strix | **7** | **105** (97 → `gb10/common`) |
| strix-hip | 22 | 87 |

So a gb10 edit changes what AMD compiles — and `taxon::hardware_of()` cannot see it, being a path prefix.

Static where provable, judgement where not. `scripts/check_cross_hardware.py` resolves all three sharing mechanisms and subtracts hardware where a real stem shadow wins. Includes resolve **lexically**, because nvcc is handed the symlink path — which is why strix-hip's real `prefill_paged_compute.cuh` (BR64 32) shadows gb10's (64).

### Negative controls — both real incidents replay as failures

| control | result |
|---|---|
| `d584c0c50` (gb10 edit broke hipcc; caught only by a compile leg) | **rc=1** `CHKI (undeclared: strix, strix-hip)` |
| `17fe989ec` (Strix workaround in gb10 common; invalidated 27 targets) | **rc=1** CHKI |
| no kernel change | rc=0 |
| strix-owned file | reaches strix only — no false positive |
| mutated self-fixture | **rc=2** — refuses to scan rather than pass green |

Fail-closed: rc 2 ("cannot see the inputs") is as red as rc 1, matching `campaign-guard.sh`'s exit 2.

★ **Deliberately not done:** making `taxon::affected()` symlink-aware. It would render `gate/closure.rs::excuses()` unsatisfiable — records exist for gb10 only (1700/1700), so any edit to one of the 97 shared files would be permanently un-excusable. Attribution and invalidation are different questions.

## Verification

`PREFLIGHT CLEAN` — file-size cap, fmt, doc, clippy (0 lints), `cargo test` (83 suites), certification self-test. `.gitignore` now tracks `.claude/agents/` alongside `.claude/skills/`: an agent file is a blocking review contract, and a gate nobody can read is not a gate.

### DIFF
--- a/.claude/agents/oracle_cert.md
+++ b/.claude/agents/oracle_cert.md
@@ -1,0 +1,207 @@
+---
+name: oracle_cert
+description: O.R.A.C.L.E::certification_state_check — the S.T.A.T.E examiner (Sha, Tree, Agreement, Timing, Environment). Decides whether a benchmark certification campaign (~4.5-5 GPU-hours) may START, may CONTINUE, or may be SEALED — before the first gate, between and during gates, and after the last one. Blocking. Spawned by /oracle_certification_state_check; never call it on a summary. It rules on the STATE of one branch, one box and one PR at one moment — not on stack order (that is `oracle`), not on whether the numbers pass their bars (that is the gate).
+model: opus
+tools: Bash, Read, Grep, Glob
+---
+
+# O.R.A.C.L.E::certification_state_check — S.T.A.T.E: Sha, Tree, Agreement, Timing, Environment
+
+You rule on **one** thing: at this moment, on this box, for this sha, is it safe to spend
+(or keep spending, or bank) a certification campaign?
+
+You exist because a campaign is the most expensive thing this repo does and the cheapest
+thing to waste. Every row below is a measured loss, not a plausible one.
+
+| what happened | cost | the test that now catches it |
+|---|---|---|
+| 2026-08-28: ten gates ran nine hours at `1d30d5d5ff`; a 23-line push to `ffn.rs` landed twenty minutes into the last gate; every record was content-invalidated | 9 h | T9, run between AND during gates |
+| `$ATLAS_HOME` unwritable; every gate died `recipe … not in the local index (0 cached)` with the cause nowhere in the message | hours | T4 |
+| TTFT gates on a box with no stored baseline recorded `info`, not a verdict; a ten-gate campaign came back eight | 2 gates | T6 |
+| BFCL produced all 995 responses, died in scoring on a lazy `soundfile` import, wrote no record, and **exited 0** | 1.6 h | T7, and "judge by the record, never by rc" |
+| a campaign split across three boxes carried three signing keys; CI rejected the lot; seven gates re-measured | one night | T14 |
+| 2026-09-06: `seal status` finished 13:37-13:41, the Seal was minted 13:40-13:42 — three green `Seal` checks beside three red `seal status`, and nobody watching | 3 PRs stalled | T12 |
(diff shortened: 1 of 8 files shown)


tbraun96:
## Evidence for the seal

**No campaign owed.** The diff touches zero `PERF_PATHS` — eight files, all tooling:

```
.claude/agents/oracle_cert.md          .github/workflows/ci.yml
.claude/agents/oracle_chki.md          .gitignore
.claude/skills/oracle_certification_state_check/SKILL.md    AGENTS.md
.claude/skills/oracle_pre_commit_cross_hardware_check/SKILL.md
scripts/check_cross_hardware.py
```

No records added, so `Disagreement::Commits` has nothing to judge and no signer question arises. No `BOUNDARY_FILE` touched.

### The new check proves it can fail
| control | result |
|---|---|
| `d584c0c50` — gb10 edit that broke hipcc | **rc=1** `CHKI (undeclared: strix, strix-hip)` |
| `17fe989ec` — Strix workaround in gb10 common | **rc=1** CHKI |
| no kernel change | rc=0 |
| strix-owned file | reaches strix only — no false positive |
| mutated self-fixture | **rc=2**, refuses to scan rather than pass |

`PREFLIGHT CLEAN` — file-size cap, fmt, doc, clippy (0 lints), `cargo test` (83 suites), certification self-test.

### Known limits, stated rather than implied
- The script implements **R1** (reach + declaration), **S1** (new cross-hardware symlink) and **S4** (dangling). **S2** (de-share must pin base bytes) and **S3** (new AMD arm in an nvidia-owned file) are specified in the skill and agent but **not yet implemented** — the negative controls above exercise R1/S1/S4 only.
- The certification lockfile is **per-checkout, not per-box**, and a residual TOCTOU window remains between `begin` and the driver's first serve. Both are written into the skill rather than glossed.

Sealing on the repo owner's explicit approval.

PR #1002, opened 2026-09-11
spark-server: batch fresh prompts into one varlen prefill step and never drop a response on fallback

### LINKED ISSUE #1000: Batched varlen prefill fallback returns HTTP 200 with zero tokens and no finish_reason at C=16
## Summary

With batched (varlen) prefill armed, a wave that the model declines mid-forward
fails every request in it **silently**: the scheduler frees each sequence and
drops its `ResponseSink` without sending anything. The HTTP body for a streaming
request is already committed with a 200 by then, so the client sees a stream that
ends after the role delta — no content, no `finish_reason`, no usage frame, and
no error. At C=16 that was sixteen empty 200s.

This is filed independently of the throughput question it was found next to: the
failure is silent truncation, which is indistinguishable at the client from a
correct empty answer.

## Environment

- 1xH100 80GB (`atlas-testing-jawn`), 2026-09-11, round 11.
- Atlas `hopper/h100-integration-2026-09-10` @ `5404a2b14`, kernel count 193.
- Model `Qwen/Qwen3.8-27B-FP8`.
- Serve: the round's best recipe plus `ATLAS_PREFILL_VARLEN=1`
  (`--max-prefill-tokens 8192`, `--max-batch-size 16`); boot line shows
  `prefill_varlen_batch=true`.

## Repro

1. Serve with `ATLAS_PREFILL_VARLEN=1` (or `--prefill-varlen-batch`).
2. Send a 16-way concurrent burst of ~1193-token prompts, 256 output tokens each.

Coherency 4/4 and the 16-way probe 16/16 both pass first, and C=1 is identical to
the flag-off arm (53.04 tok/s, 268.3 ms TTFT) — the flag costs nothing until it
engages.

## Observed

Three consecutive C=16 reps, all sixteen streams returning nothing:

```
[r11-E-varlen-i1024o256] C= 16 rep0  tok/s=0.00  wall=0.34s  ctok=0  ptok/req=[0]  ttft_p50=235ms  err=0
[r11-E-varlen-i1024o256] C= 16 rep1  tok/s=0.00  wall=0.34s  ctok=0  ptok/req=[0]  ttft_p50=238ms  err=0
[r11-E-varlen-i1024o256] C= 16 rep2  tok/s=0.00  wall=0.32s  ctok=0  ptok/req=[0]  ttft_p50=247ms  err=0
```

HTTP 200 on all sixteen, a first-token event at ~235 ms, then the stream ends
with no content, no `finish_reason` and no usage frame. `err=0`: the client
cannot tell anything went wrong.

The server log names the cause — twelve such lines on this cell, zero on the four
cells that never arm the flag:

```
ERROR spark::scheduler::phase_continue_prefills::run_batched_prefill:
  Batched prefill error (wave of 6 streams, 6 prefilling): prefill_inner: batched mode
  requires seq_len_start > 0 (paged path); got seq_len_start=0.
  Caller must fall back to per-stream for this chunk.
```

**It is intermittent, which makes it worse.** A re-run of the identical rung on
the same live process gave `rep0 tok/s=438.20 ctok=4096` (a clean full run, ~2%
*faster* than the flag-off arm) and `rep1 tok/s=391.13 ctok=3328
ptok/req=[0, 1193]` — thirteen of sixteen completed, **three silently dropped**.
So the same binary and the same request shape produce a correct run, a
partially-dropped run, or a wholly-dropped run.

## Expected

A prefill that fails for any reason must terminate its request visibly: an error
frame on a streaming sink, `Err` on a blocking one. A request must never be
retired by closing its channel. Separately, when the batched path declines
*before* mutating stream state — which is what the log line above asks for in so
many words — the scheduler should run the per-stream path for every member of the
wave rather than failing them.

## Where

- `crates/spark-server/src/scheduler/phase_promote_prefills.rs` — the
  `maybe_token == None` arm frees the sequence and lets `p.sink` drop.
- `crates/spark-server/src/scheduler/phase_continue_prefills/run_batched_prefill.rs`
  — a wave error pushes `(i, None)` for every member instead of falling back.
- `crates/spark-model/src/layers/qwen3_attention/trait_impl/prefill_inner.rs` —
  the chunk-zero guard that raises the refusal, out of step with the admission
  predicate in `batch_kernel/eligible.rs` that accepted the wave.

Refs #927.

TheTom:
## Summary

**A 16-way burst of fresh prompts was prefilled strictly one prompt at a time, and the flag that was supposed to fix it returned sixteen empty HTTP 200s instead.** Both halves are one cause: four call sites decide whether chunk-zero streams may co-admit into a batched prefill, and one of them read a different boolean than the other three.

This PR makes that predicate a single function, re-runs a wave the model declines *before touching any stream* per-stream instead of dropping it, and gives every failed prefill an error frame instead of a silently closed SSE channel. The batching lever itself stays **opt-in and default OFF** (`--prefill-varlen-batch`, legacy `ATLAS_PREFILL_VARLEN=1`) until the H100 A/B is on the table.

Fixes #1000. Refs #927.

## What was wrong

Round 11 on 1×H100, `Qwen/Qwen3.8-27B-FP8`, two back-to-back 16-way bursts of 1193-token prompts, `--max-prefill-tokens 8192` (`prefill_budget=8192, max_batch_tokens=8196`), temp 0.

**Cell A — the default path.** nsys over the window counted `rms_norm_residual` launches per forward: **64 prefill forwards for 32 requests, strictly `1168, 25, 1168, 25, …`**, every one of them a single prompt.

| forward `GrdX` | count | busy ms | ms/forward | µs/token |
|---|---|---|---|---|
| 1168 (prompt head) | 32 | 7064.4 | 220.76 | 189.0 |
| 25 (prompt tail) | 32 | 936.2 | 29.26 | **1170.2** |

A 16-row decode step on the same box costs 1234.5 µs/token — **the 25-token tail chunk is per-token indistinguishable from decode**, and it is 11.7% of prefill GPU time for 2.1% of the tokens. Window: PREFILL 8.0 s (39.8%), DECODE 10.1 s (50.2%), IDLE 2.0 s (10.0%). C=16 TTFT p50 2.28 s / p99 4.22 s. **The budget was never binding**: 2048 and 24576 both moved the aggregate ±0.4%.

The scheduler was not the serialiser. `continue_in_progress_prefills` already hands all sixteen streams to `Model::prefill_batch_chunk` in one call. `kernel_batched_eligible` then rejects any batch with `chunk_start == 0` unless `allow_chunk_zero`, and a burst of fresh prompts is all chunk 0 — so it fell through to the per-stream loop: sixteen forwards at M=1193 instead of one at M≈7000, every projection GEMM at a sixth of the arithmetic intensity available.

**Cell E — the documented fix, enabled.** `ATLAS_PREFILL_VARLEN=1` at C=16:

```
ERROR ...run_batched_prefill: Batched prefill error (wave of 6 streams, 6
prefilling): prefill_inner: batched mode requires seq_len_start > 0 (paged
path); got seq_len_start=0. Caller must fall back to per-stream for this chunk.
```

**Sixteen HTTP 200s carrying zero tokens, no `finish_reason`, no usage frame, `err=0` at the client — 3/3 reps**, plus a recheck rep that dropped 3 of 16. Silent truncation, indistinguishable from a correct empty answer.

## Root cause

**1. Four readers, one boolean, three agreements.** `check_kernel_batched_eligible` (admission), the paged-upload force in `batch_kernel.rs` and `prefill_attention_paged_attn_batched` all read `codispatch || varlen`. The fourth — `Qwen3AttentionLayer::prefill_inner` — read `codispatch` **alone**. So under `--prefill-varlen-batch` admission said yes and the attention layer said no at layer 0, with Phase A already done: KV blocks allocated, prefix reservations taken, hidden staged for every member of the wave.

**2. The fallback the error asked for did not exist.** The bail text says "Caller must fall back to per-stream for this chunk." The caller pushed `(i, None)` for every member, and `promote_completed_prefills` freed each sequence while **dropping its `ResponseSink`**. The SSE body had already been committed with a 200 when the handler subscribed, so closing the channel produced a stream ending after the role delta. A blocking request fared no better: the dropped `oneshot` surfaces as the misleading "Inference cancelled".

## What the fix does

**One predicate, `ops::prefill_batched_chunk_zero_allowed()`.** All four sites call it; nobody re-derives the disjunction. `eligible::first_chunk_batched_enabled` is deleted rather than left as a second definition.

**`BatchedPrefillDeclined`, a refusal raised before any stream was mutated.** New marker error in `spark_model::traits`, detected with `downcast_ref` through the `anyhow` chain. The distinction is load-bearing: once a batched forward has entered Phase A it owns KV blocks and prefix reservations for every member, so re-running those streams would double-allocate — the caller must fail them *with a response*. A refusal raised earlier leaves the streams exactly as the scheduler handed them over, so the caller can and must re-run the whole wave per-stream. `batch.rs`'s arena-capacity check — which runs before a single stream is touched and used to take the whole wave down for one oversized member — is now a decline.

**Per-stream fallback, `phase_continue_prefills/prefill_fallback.rs`.** A declined wave is re-run one stream at a time (at `n == 1` the model's dispatcher delegates to the same single-stream path the non-batched scheduler runs), so a member that still fails fails alone. The advance/sample bookkeeping both paths share moved into the same module.

**Every failed prefill now answers.** `phase_promote_prefills` calls `lifecycle::send_error_to_sink` on the hard-error path — the same helper `prefill_request` already uses for a failure between taking the sink and building an `ActiveSeq`. This is that same window, one phase later.

**Tail pre-split, `Model::prefill_tail_cut`.** `prefill_chunk_dispatch` splits a prompt's FINAL chunk once at `((total-1)/bs)*bs - bs` (1168 for a 1193-token prompt at `bs=16`) to land an SSM tail checkpoint a later turn's block-floored prefix match can use. It is unconditional on hybrid-SSM models with the prefix cache active and must stay so — a shape that differs cold vs warm flips a temperature-0 argmax, since BF16 accumulation is not associative. The batched path does not split, so handing it a whole prompt as one `is_last` chunk would give a co-admitted stream a **different forward shape than the same prompt takes per-stream**. The model now reports where it would cut and the scheduler pre-splits there under VARLEN: geometry stays identical per sequence, and the tails **batch** — sixteen tails of one burst share `chunk_start = 1168`, so the planner puts all of them in ONE forward instead of sixteen standalone 25-token passes. The arithmetic is a single free function, `prefill_b::tail_split_cut`, called by both the splitter and the reporter.

**Admission arithmetic, for the record.** `wave_cap = min(--max-prefill-tokens, max_batch_tokens) = 8192`. Sixteen whole 1193-token prompts pack 6 per wave (6×1193 = 7158; a seventh is 8351) → **3 dispatches, not 16**. Pre-split 1168-token heads pack 7 (7×1168 = 8176). Both pinned in `prefill_waves::tests`.

**The boot route line prints the resolved decision**, not the raw argument: `prefill_chunk_zero_batch=` joins `prefill_varlen_batch=` on the `kernel flags:` line, so a log says which way the four-site predicate actually resolved.

## Tests

**`crates/spark-model/src/layers/ops/chunk_zero_ssot_tests.rs` (3)** — a source-scanning test over the tree, so a fifth reader cannot quietly re-derive the disjunction:

- `every_chunk_zero_reader_calls_the_one_predicate`
- `no_chunk_zero_reader_rederives_the_disjunction`
- `the_predicate_is_defined_once`

**`.../prefill_b/batch_kernel_chunk_zero_tests.rs` (2)** — the admission half, at the #927 shape (sibling file because `batch_kernel_tests.rs` was at the 500-LoC cap):

- `admits_the_h100_c16_chunk_zero_wave` — six fresh 1193-token prompts admitted; the six 25-token tails sharing `chunk_start == 1168` admitted as one wave; and the control, that with the predicate OFF the same wave is refused **at admission, before any stream is touched**
- `varlen_admits_ragged_chunk_zero_lengths`

**`crates/spark-server/src/scheduler/prefill_fallback_tests.rs` (3)** — end-to-end through the scheduler tick against a stub model:

- `a_declined_wave_is_re_run_per_stream_and_every_request_completes`
- `a_hard_batched_failure_still_answers_every_request` — the anti-#1000 assertion: an error frame, never a closed channel
- `the_control_the_batched_path_completes_everyone_when_it_works`

**`.../phase_continue_prefills/prefill_waves.rs` tests (+2)** — the burst geometry, pinned as numbers:

- `the_h100_c16_burst_packs_six_prompts_per_prefill_step` — 16 whole prompts → 3 waves of 6/6/4; 1168-token heads → 7 per step; sixteen 25-token tails → **one** wave
- `flag_off_still_admits_every_prompt_in_one_call` — the control, that the planner does not start splitting with the lever off

**`.../trait_impl/prefill_b.rs::tail_split_cut_tests` (4)** — `1193 @ bs=16 → 1168` (tail 25), a prompt sitting on a boundary, short prompts saturating to zero, and `bs == 0` not being a division.

**`crates/spark-server/src/scheduler/test_support_prefill.rs`** — the stub model and harness the fallback tests need, extracted out of `prefill_fifo_tests.rs` (which shrinks by 161 lines and keeps its five cases unchanged) so both suites share one stub.

**`crates/spark-model/examples/native_prefill_varlen_batch_microtest.rs`** — GPU receipt, `required-features = ["cuda", "gpu-examples"]`, exit 1 on failure. Three gated contracts on four ragged chunk-zero sequences: (1) **BITS** — the batched kernel's output equals the single-stream kernel's byte for byte (both arms run the same `prefill_paged_compute.cuh` body at BR=32 with the same operand order, so anything but `unequal = 0` is a bug; the test stays under `chunk_len = 256` so neither arm takes the BR=64 twin); (2) **KV** — the K/V pool is byte-identical before and after both arms, per-sequence digest, because the chunk-0 batched path READS the pages a fresh sequence just wrote; (3) **EXTENT** — guard bands untouched, and stream `b`'s output region written only by stream `b`, checked by perturbing one stream's Q and asserting only its rows move. It is built by this PR's gate but **has not been run on an H100**; see below.

## Test plan

All on Spark 1 (GB10), `ATLAS_SKIP_BUILD=1`, `--jobs 8`, in an isolated worktree against this branch.

- [x] `cargo clippy -p spark-model --all-targets --features cuda -- -D warnings` — **0 warnings**, `Finished \`dev\` profile [unoptimized] target(s) in 13.41s`
- [x] `cargo clippy -p spark-server --all-targets --features cuda,nccl -- -D warnings` — **0 warnings**, `Finished \`dev\` profile [unoptimized] target(s) in 25.30s`
- [x] `cargo test -p spark-model --features cuda --lib` — **897 passed, 0 failed, 11 ignored**
- [x] `cargo test -p spark-server --bins --features cuda,nccl` — **2460 passed, 0 failed, 12 ignored** (first run; the grammar cold-compile timing test did not flake)
- [x] `cargo build -p spark-model --examples --features cuda,gpu-examples` — clean, the microtest binary links
- [x] The CI file-size cap check (`.github/workflows/file-size-cap.yml`, run locally against this tree) — `✓ All .rs files in crates/ are ≤500 LoC (or in the allow-list)`. Nothing was added to the allow-list; the largest new file is the 400-LoC example.

Each of this PR's fourteen new unit tests was confirmed present in the run, by name, not just in the aggregate count.

- **No H100 A/B has been run.** The lever stays opt-in and default OFF. Everything above about throughput is *admission arithmetic* — how many prompts fit in a wave — plus the round-11 receipt for the behaviour being fixed. Nobody has measured the batched path's tok/s or TTFT against the default path on an H100 in this state.
- The measurement cells still owed, before the default can move: (a) C=16 burst, lever OFF, the round-11 cell-A control re-taken on this branch; (b) C=16 burst, lever ON — tok/s, TTFT p50/p99, and the nsys forward histogram, which should read 3 head waves + 1 tail wave rather than 64 single-prompt forwards; (c) the same pair with `ATLAS_NO_TAIL_SPLIT=1`, which is the A/B for the pre-split by itself; (d) a temperature-0 answer-equality check between lever ON and OFF, since the whole reason the tail split is pre-applied is that a differing forward shape flips an argmax; (e) the GPU microtest above, run on the H100.
- **The 25-token tails are not eliminated**, only batched. Removing the tail split entirely is `ATLAS_NO_TAIL_SPLIT=1` and costs the warm-turn checkpoint; that trade is not decided here.
- Nothing here touches the MLA path. `prefill_inner_hc` keeps its hard `seq_len_start > 0` failure, because `check_kernel_batched_eligible` rejects MLA configs wholesale via `config_is_mla` before any stream is touched — it is unreachable from the batched dispatcher, and a comment now says so rather than leaving it looking like a missed site.

`PREFILL-BATCHING-ATTRIBUTION.md` carries the full phase-1 attribution: the nsys anchors, why chunk 0 needed nothing structural, and why a tail cannot batch with another prompt's head.

## CLA

### DIFF
--- a/PREFILL-BATCHING-ATTRIBUTION.md
+++ b/PREFILL-BATCHING-ATTRIBUTION.md
@@ -160,3 +160,86 @@ projection GEMM at a sixth of the arithmetic intensity it could have") is
 refuted by measurement: the M=1168 GEMMs are already at full efficiency on an
 H100. Varlen repackages the GDN work; it does not reduce it. The lever stays
 default OFF, and the C=16 gap to vLLM is not a prefill-batching deficit.
+
+## 6. The slot fix held; the SCHEDULING of the waves was the loss (#1002, H100 round 15)
+
+Round 15 rebuilt at `8a6f50b61` with #1002's slot-aliasing fix in. **Every
+structural check passes**: cell V15 logged **0 `SSM pool slot SHARED`, 0
+`release_slot … already free`, 0 content-loop / fuzzy / SimHash stops, 16/16
+responses at the full 256 tokens**, and every `Captured CUDA graph … slots=`
+list carries distinct real slots (index 16 is the padding sentinel). Against
+round 13's cell V — 24 content-loop fires, 2 fuzzy stops, 5 SimHash stops, 6/16
+probe responses cut at 49 tokens — that is a clean reversal. On the LONG shape
+`deferral SKIPPED` fires and the numbers land on the no-lever cell's exactly
+(398.34 vs 401.28 tok/s, 4 115.2 vs 4 118.8 ms TTFT): round 13's +83% TTFT
+regression is gone, replaced by a null.
+
+**And the lever was still a loss on the SHORT shape, for a scheduling reason,
+not a kernel one.** Whenever the 16-stream varlen batching engaged:
+
+| | A15 (no lever) | **V15 (engaged)** | Δ |
+|---|---:|---:|---:|
+| `1024x256` C=16 aggregate | 513.86 | **427.53** | **−16.8%** |
+| TTFT p50 | 1 359.4 ms | **4 141.2 ms** | **3.05×** |
+| TTFT p99 | 2 504.5 ms | 4 142.2 ms | +65% |
+| **TPOT** | 25.90 ms | **21.28 ms** | **−17.8%** |
+| e2e p50 | 7.97 s | 9.57 s | +20% |
+
+**TPOT moved the right way by 17.8%** — batching the prefill up front genuinely
+removes the prefill/decode interference round 11 traced — so the batching works.
+What did not work was that **p50 = p99**: `16 streams -> 3 wave(s)` ran
+back-to-back inside ONE tick. Promotion is a phase, not a callback: a stream's
+first token reaches its client in `promote_completed_prefills`, after
+`continue_in_progress_prefills` returns, and decode runs later still in the
+tick's own decode step. So a stream that finished in wave 1 waited for waves 2
+and 3 before anyone heard from it, and every TTFT in the burst landed on the
+slowest.
+
+**The fix is ONE WAVE PER TICK** (`prefill_waves::waves_this_tick`), not a
+second guard condition. Wave 1 runs, its finished streams are promoted at the
+end of that tick, decode interleaves from the next one, and the streams that did
+not fit re-plan next tick — where they batch among themselves, because the
+planner is re-run from the live geometry every tick. On the round-15 shape:
+sixteen 1193-token prompts plan 7 / 7 / 2, seven prefill on tick 1, and on tick
+2 those seven stand first in FIFO order and their tails complete as one
+175-token forward before any further head wave is issued.
+
+The alternative — promoting inside the wave loop — buys nothing here.
+`promote_completed_prefills` removes from `prefilling`, which invalidates the
+wave indices mid-loop, and decode still would not run until the prefill phase
+returned; the promoted stream's token would leave at the same wall time. The
+tick boundary is where the scheduler already interleaves, so that is where the
+cut belongs.
+
+FAIRNESS is unchanged in kind. The planner is first-fit in FIFO order, so wave 1
+always contains stream 0: the head of the queue advances exactly one chunk per
+tick, which is the guarantee the single-stream `prefilling.first_mut()` path
+gives — and it now carries everyone who fits with it. With the flag OFF the
+planner returns one wave holding every stream and this cap returns it whole, so
+that path stays byte-identical. Pinned in `prefill_waves_tick_tests`.
+
+**Why engagement varied burst to burst, and what is done about it.** Round 15
+saw two of four otherwise identical short-shape bursts engage (427 tok/s) and
+two not (514 tok/s — the no-lever numbers), a 17.52% rep spread that is not
+measurement noise. The planner is not the source: `plan_prefill_waves` is a pure
+function of the geometries. The ADMISSION is: it requires `active.is_empty()`
+and two chunk-0s co-admitted in the SAME tick, and a burst whose first request
+has already been promoted to decode by the time the rest arrive fails the first
+test. That is arrival timing against the scheduler's tick period — **inherent,
+not pinnable** — so `varlen_admission` now returns the one input that decided
+each verdict and `phase_start_prefills` logs it per burst:
+
+```
+Varlen prefill admission: defer=false reason="decode already active this tick (arrival timing)" new_reqs=9 prefilling=0 active=7 smallest_chunk0=[1193, 1193] cap=8192
+```
+
+A run's engagement is now readable from its serve log instead of inferred from
+its throughput. The `deferral SKIPPED` line is unchanged and still fires on the
+long shape.
+
+**The lever stays default OFF.** §5's finding stands: `M=8176` (7 prompts fused)
+measured **190.4 µs/token against 188.8 µs/token at M=1168** — 0.8% WORSE —
+so varlen repackages the GDN work rather than reducing it. This change removes
+the TTFT collapse that made the lever a 16.8% aggregate loss on top of that; it
+does not make batching pay, and the C=16 gap to vLLM is still not a
+prefill-batching deficit.
--- a/PREFILL-BATCHING-ATTRIBUTION.md
+++ b/PREFILL-BATCHING-ATTRIBUTION.md
@@ -98,3 +98,65 @@ max_batch_tokens) = 8192`. Sixteen whole 1193-token prompts pack 6 per wave
 (`--prefill-varlen-batch`, legacy `ATLAS_PREFILL_VARLEN=1`, default OFF) until
 the H100 A/B; the boot route line prints the resolved chunk-zero decision beside
 it, and `ATLAS_NO_TAIL_SPLIT=1` stays the A/B for the split itself.
+
+## 5. Why the lever then degenerated its output (#1002, H100 round 13)
+
+Phase-2 attribution. The lever now engages as §1-§4 describe — `16 streams ->
+3 wave(s), M per wave [50, 8176, 8176] (cap 8192)`, `Q12 kernel-batched prefill
+dispatched (fused large-M) n=7 total_tokens=8176`, every response carrying
+tokens and a `finish_reason`. And cell V logged **24 content-loop-watchdog
+fires, 2 fuzzy-repetition stops and 5 SimHash stops against ZERO on cells A, D,
+T1 and T2** — same binary, same prompts, temp 0, seed 42 — with 6/16 probe
+responses cut at 49 tokens.
+
+**Not the varlen kernel path. The scheduler hands two live sequences one SSM
+pool slot.** A sequence claims its slot at admission, so a stream parked in
+`prefilling` owns one; `compact_survivors_into_range` derived its free targets
+from the DECODING set alone. Varlen is the first configuration that parks
+slot-owning streams there (`want_varlen_defer`: 96 deferrals on cell V, zero on
+A/D/T1), so the per-tick compaction migrated an active survivor onto a
+prefilling stream's slot — visible in the log 30 ms apart as
+`slots=Some([0, 7])` then `slots=Some([0, 1])`. `compact_sequence` calls
+`ssm_pool.claim_specific` and discards its false return, so the collision is
+silent, and two sequences then share one GDN `h_state`/`conv_state`.
+
+It outlives the burst: both owners release that index, `release_slot`'s only
+guard was a `debug_assert` (a no-op in the shipped `--release` binary), and the
+duplicated free-list entry makes `claim_slot` issue it to two fresh sequences
+for the life of the process. That is why the 16-way probe four minutes later —
+**with varlen not engaging at all**, every prefill logging `Prefilled (single
+chunk)` — still decoded at `slots=[0, 0, 0, 1, 1, 2, 2, 3, ...]` and lost 6/16
+responses. Duplicate REAL slots appear in 16 of cell V's 24 batched-decode
+captures and in none of any other cell's in this campaign.
+
+**The `50`-token wave is not a truncated chunk 0.** `[50, 8176, 8176]` is two
+25-token TAILS (from streams that arrived one tick early, their heads dispatched
+as the `2336` forward) plus fourteen 1168-token heads: `2 x 25 + 14 x 1168 =
+16402`. The budget can never shrink a chunk 0 — `plan_stream_chunk` budgets
+against `max_prefill_tokens` and the WAVE cap belongs to the planner, which
+opens a new wave rather than trimming a member. Pinned in `prefill_waves_tests`
+alongside a tick-by-tick replay of the burst's arrival pattern.
+
+**The long shape's +83% TTFT was deferral with no batching.** 4593 pre-splits
+to `4576 + 17` and `2 x 4576 = 9152 > 8192`, so the planner emitted 14 waves for
+16 streams; waves run back-to-back inside one tick, so no stream is promoted
+until all of them have run and every TTFT collapses onto the p99 (7 524.7 ->
+13 756.1 ms while p99 barely moved). `varlen_defer_pays` now declines the
+deferral when the two smallest chunk-0s cannot share a wave.
+
+**And a watchdog stop is no longer only a `"length"`.** The wire
+`finish_reason` still reports `"length"` for every non-timeout guard — that
+mapping is a measured contract, and minting a new enum value hard-fails typed
+clients — but the guard's name now rides beside it in the `stop_reason`
+extension field, on the streaming chunk and the blocking choice alike (the
+round-13 probe ran `stream=false`).
+
+**Still open after this:** the batching itself does not pay. nsys on cell V
+measured `M=8176` (7 prompts fused) at **190.4 us/token against 188.8 us/token
+at M=1168** — 0.8% WORSE — with total prefill busy moving 8 000.5 -> 7 925.2 ms
+(-0.9%) for the same 38 176 tokens, and total GDN prefill identical at 54% of
+prefill either way. Round 11's arithmetic-intensity hypothesis (§1, "every
+projection GEMM at a sixth of the arithmetic intensity it could have") is
+refuted by measurement: the M=1168 GEMMs are already at full efficiency on an
+H100. Varlen repackages the GDN work; it does not reduce it. The lever stays
+default OFF, and the C=16 gap to vLLM is not a prefill-batching deficit.
--- a/PREFILL-BATCHING-ATTRIBUTION.md
+++ b/PREFILL-BATCHING-ATTRIBUTION.md
@@ -1,0 +1,100 @@
+# Why concurrent prefills ran one at a time (#927, H100)
+
+Phase-1 attribution for the batched (varlen) prefill lever. **Headline: the
+scheduler was already co-admitting all sixteen streams; the serialisation is one
+crate down, and the documented fix could not engage because four sites disagreed
+about one boolean.**
+
+## Anchors (measured, 1xH100, Qwen/Qwen3.8-27B-FP8, round 11 cells A and E)
+
+nsys over two back-to-back 16-way bursts of 1193-token prompts,
+`--max-prefill-tokens 8192` (`prefill_budget=8192, max_batch_tokens=8196`).
+`rms_norm_residual` launches once per layer per forward, `GrdX` = that forward's
+token count; all 576 detected forwards have exactly 64. **64 prefill chunks for
+32 requests, strictly `1168, 25, 1168, 25, …`.**
+
+| forward `GrdX` | count | busy ms | ms/forward | **µs/token** |
+|---|---|---|---|---|
+| 1168 (prompt head) | 32 | 7064.4 | 220.76 | 189.0 |
+| 25 (prompt tail) | 32 | 936.2 | 29.26 | **1170.2** |
+
+A 16-row decode step costs 1234.5 µs/token — the tail chunk is per-token
+indistinguishable from decode. Window: PREFILL 8.0 s (39.8%), DECODE 10.1 s
+(50.2%), IDLE 2.0 s (10.0%). C=16 TTFT p50 2.28 s / p99 4.22 s (vLLM 1.05 / 1.17,
+same box). Budget at 2048 or 24576 moved the aggregate ±0.4% — **the budget was
+never binding**, which is what sent this to the trace.
+
+## 1. Why the default path prefills one prompt per step
+
+Not the scheduler. `continue_in_progress_prefills` takes
+`can_batch_prefill_only` (N≥2 prefilling, no active decode, not EP) and hands ALL
+sixteen streams to `Model::prefill_batch_chunk` in one call. The serialisation is
+inside that call: `prefill_batch_chunk_dispatch` asks `kernel_batched_eligible`
+first, and that predicate rejects any batch with `chunk_start == 0` unless
+`allow_chunk_zero` — false with neither `ATLAS_PREFILL_CODISPATCH` nor
+`--prefill-varlen-batch` set. A burst of fresh prompts is all chunk 0, so it falls
+through to the per-stream loop: sixteen sequential forwards at M=1193 instead of
+one at M≈7000, every projection GEMM at a sixth of the arithmetic intensity it
+could have. The `seq_len_start > 0` precondition exists because the
+batched attention kernels are **paged-only** and chunk 0 historically took the
+non-paged `prefill_attention_with_cache_skip` arm. VARLEN v1 lifted that (chunk-0
+streams get a forced paged upload in `batch_kernel.rs`), so the precondition was
+already stale — for three of its four readers.
+
+## 2. Why `--prefill-varlen-batch` returned empty 200s instead
+
+Four sites decide chunk-zero co-admission. Three read `codispatch || varlen`:
+`check_kernel_batched_eligible`, the paged-upload force, and
+`prefill_attention_paged_attn_batched`. The fourth,
+`Qwen3AttentionLayer::prefill_inner`, read `codispatch` **alone**. Admission said
+yes; the layer said no — at layer 0, with Phase A already done (KV blocks
+allocated, prefix reservations taken, hidden staged for every member).
+
+```
+ERROR ...run_batched_prefill: Batched prefill error (wave of 6 streams, 6
+prefilling): prefill_inner: batched mode requires seq_len_start > 0 (paged
+path); got seq_len_start=0. Caller must fall back to per-stream for this chunk.
+```
+
+The error asked the caller to fall back. The caller did not: it pushed
+`(i, None)` for every member and `promote_completed_prefills` freed each sequence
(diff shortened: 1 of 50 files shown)


TheTom:
H100 A/B taken (round 13, tip `3c0379030` on the #895 line, same binary for every cell, frozen ladder, C=16). Converting to draft: the fix does what it says — the lever arms, the planner packs `16 streams -> 3 wave(s), M per wave [50, 8176, 8176]`, and 16/16 responses now carry tokens and a `finish_reason` (the #1000 symptom is gone) — but two things block it.

**Output degenerates under the batched wave.** Content-loop watchdog fired 24 times on this cell (plus 2 fuzzy-repetition and 5 SimHash stops) against zero on four other cells of the same binary; 6/16 responses in a 16-way probe were cut at 49 tokens and reported `finish_reason: "length"`. Coherency is 4/4 and md5-identical to the control at C=1 (the path does not engage there), so the divergence is confined to co-admitted chunk-0 streams. The trace also shows odd leftover packing (`2336 = 2 prompts`, a `350`-token and a `50`-token wave for streams whose per-stream shape is 1168 + 25), which is the first place to look. Intermittent: two of three ladder reps matched the control, one lost 780 tokens.

**It does not buy throughput.** nsys: the batched M=8176 forward costs 190.4 µs/token vs 188.8 µs/token at M=1168 — the single-prompt GEMMs are already at full efficiency, so the arithmetic-intensity argument in the PR body is refuted by measurement; total prefill busy moved −0.9% for the same 38 176 tokens. On the long shape (4593 → 4576 + 17) no two chunk-0s fit the 8192 cap, so the planner emits 14 waves for 16 streams and pays the deferral for nothing: TTFT +83%, aggregate −31%.

Also on this branch: the GPU microtest's stream-isolation control perturbs the first causal query row, whose output is Q-independent, so the control can never trip (the bit-exact comparison itself passed, 0 unequal of 3.17 M). Fix coming as an appended commit; the fallback-correctness and SSOT parts stand.

TheTom:
Fix for the degeneration appended (`35cb06a93`, `466ff2df6`, `5aeb65fb5`, `6ffe19c00`) — cherry-picked onto the frozen branch, nothing rewritten. Attribution §5 carries the long form.

**Root cause — the scheduler, not the varlen kernel path.** A sequence claims its SSM pool slot at admission, so a stream parked in `prefilling` already owns one, and `compact_survivors_into_range` derived its free targets from the *decoding* set alone — correct only while decoders own every slot, and `want_varlen_defer` is the first thing in the tree that parks slot-owning streams in `prefilling` (96 deferrals on cell V, zero on A, D and T1, which is exactly the set that degenerated). So the per-tick compaction migrated a decoding survivor onto a prefilling stream's slot — `18:54:34.514 Captured CUDA graph for batch size 2 (n=2, slots=Some([0, 7]))`, then 30 ms later `18:54:34.544 Captured CUDA graph for batch size 2 (n=2, slots=Some([0, 1]))`, with slot 1 held by a prefiller — and `compact_sequence` discards the false return of `ssm_pool.claim_specific`, so two sequences shared one GDN `h_state`/`conv_state` in silence. It outlived the burst because `release_slot`'s only guard was a `debug_assert`, a no-op in the `--release` binary that shipped: both owners released the same index, the duplicated free-list entry made `claim_slot` hand it to fresh sequences, and the 16-way probe four minutes later — with varlen not engaging at all, every prefill logging `Prefilled (single chunk)` — still decoded at `slots=[0, 0, 0, 1, 1, 2, 2, 3, ...]` and lost 6/16 responses. Duplicate real slots appear in 16 of cell V's 24 batched-decode captures and in none of any other cell's; the duplicates elsewhere are all the `max_slots` padding sentinel.

**What the appended commits change.**

- The target rule is extracted pure as `mod_helpers::slot_targets::plan_slot_compaction`, and it takes the slots held by non-survivors as an explicit input so it can never choose one. A survivor with no legal target keeps its slot: non-contiguous slots cost the batched-GDN recurrence fast path (the model already logs `SSM batched recurrent DECLINED`), shared recurrent state costs correctness. `retire_finished_sequences` takes `reserved_slots`, computed from `prefilling` via the RAII guard (`ssm_slot_idx`) rather than `slot_idx`, which can be the reuse sentinel or stale after a migration. `swap_out_sequence` migrates only when the victim actually owns the target, which is the ownership-transfer premise that call has always assumed.
- `ssm_pool::release_slot` refuses a duplicate push at runtime instead of asserting in debug only. A double release stays a bug upstream; its blast radius is now the sequences already involved rather than the pool for the life of the process.
- `batch_decode_graph_key` checks the live slot vector for duplicates on every batched decode step and logs one ERROR naming the shared slots — a few dozen u32 compares at n ≤ 32, against a multi-ms forward. That is the diagnostic that would have replaced this serve-log archaeology.
- `varlen_defer_pays` skips the deferral when the two smallest chunk-0s cannot share a wave. On the long shape (4593 → 4576 + 17, `2 × 4576 = 9152 > 8192`) the planner emitted 14 waves for 16 streams; waves run back-to-back inside one tick and nothing is promoted until all of them have run, so every TTFT collapsed onto the p99 (7 524.7 → 13 756.1 ms, aggregate 308.86 → 214.25 tok/s) and bought zero batched forwards. Candidates are chunk-0 lengths only — a wave's members must share `chunk_start` and `is_last`, so counting a mid-prefill tail would claim company the planner cannot seat.
- A guard cut now names itself in `stop_reason`, beside the `finish_reason` it keeps reporting. The wire value does not move: every non-timeout guard still maps to `"length"`, because relabelling to `"stop"` cost 2/10 then 6/10 episodes of the agentic gate, and a new enum value hard-fails typed clients. The extension field is `skip_serializing_if`, so an ordinary stop serialises byte-identically — key absent, not null. Both surfaces produce it; the round-13 probe ran `stream=false`, so blocking is the one it was measured on.

Also lifts the per-stream chunk geometry into `prefill_waves::plan_stream_chunk`, unchanged, so the chunk sequence can be replayed with no model and no GPU. That was the open question from round 13: BF16 accumulation is not associative, so a stream whose chunk shape depends on who it batched with is a different answer to the same request at temperature 0.

**Tests.** 6 on `plan_slot_compaction`, including `a_reserved_slot_is_never_a_target` and `targets_are_distinct_and_never_occupied_or_reserved`; 3 on the pool guard, `double_release_is_refused_not_pushed`, `a_duplicated_free_entry_issues_one_slot_to_two_sequences` and `take_neutralizes_drop_no_double_release`; 15 in `prefill_waves_tests`, which replay the cell-V receipt — the short shape packing 7/7/2 to `[8176, 8176, 2336]` (the nsys prefill histogram), `chunk_zero_is_never_truncated_by_the_wave_budget` (the logged `50` is two 25-token tails plus fourteen heads, `2 × 25 + 14 × 1168 = 16402`), the tick-by-tick arrival replay, `round13_long_shape_does_not_defer_because_no_two_heads_fit`, and the cap exact at `4096 + 4096` fits / `4096 + 4097` does not; 4 in `stop_reason_tests` plus 2 blocking-wire and 1 lifecycle test, asserting absence through `serde_json::to_value(..).get(..)` rather than a string search, and `no_new_nonstandard_finish_reason_value`. The wave tests moved to `prefill_waves_tests.rs` — the replays roughly tripled the module and it was going over the 500-LoC cap.

**Gates**, GB10, warm target, tip `6ffe19c00`:

```
cargo fmt --all -- --check                                      exit 0
clippy -p spark-model  --features cuda      --all-targets -D warnings   Finished `dev` profile [unoptimized] target(s) in 20.27s
clippy -p spark-server --features cuda,nccl --all-targets -D warnings   Finished `dev` profile [unoptimized] target(s) in 41.31s
cargo test -p spark-model  --features cuda --lib
  test result: ok. 899 passed; 0 failed; 11 ignored; 0 measured; 0 filtered out; finished in 2.98s
cargo test -p spark-server --bins --features cuda,nccl
  test result: ok. 2481 passed; 0 failed; 12 ignored; 0 measured; 0 filtered out; finished in 1.97s
examples: spark-model / spark-runtime / spark-storage             exit 0, 0, 0
file-size cap: 0 violations outside the allow-list
```

**What this does not prove.** There is no H100 re-measurement yet — the degeneration is diagnosed from the serve logs and pinned by unit tests, not re-run on the hardware, so "the watchdog stops go away" is a prediction until cell V is repeated. And the throughput case is unchanged and still refuted: nsys measured the batched `M=8176` forward at 190.4 µs/token against 188.8 µs/token at `M=1168`, total prefill busy −0.9% for the same 38 176 tokens, so round 11's arithmetic-intensity argument does not survive contact with the profiler and varlen repackages the GDN work rather than reducing it. What this PR is worth is therefore narrower than the body claims: the fallback never drops a response, the lever is correct when someone turns it on, and the double-release hole it exposed was latent behind a `debug_assert` for every configuration, not just this one. `--prefill-varlen-batch` stays default OFF.

TheTom:
Appended `cd1fca271` — `spark-server: dispatch one varlen prefill wave per scheduler tick`. Append-only fast-forward onto `6ffe19c00`; nothing earlier on the branch was touched, and the PR stays draft.

### What it changes

The wave planner already grouped fresh prompts into waves; the dispatcher then ran **every** wave back-to-back inside a single scheduler tick. `prefill_waves::waves_this_tick` caps that at the FIRST wave whenever varlen is on.

The reason a per-wave promotion would not have worked instead: **promotion is a phase, not a callback.** A stream's first token reaches its client in `promote_completed_prefills`, which runs *after* `continue_in_progress_prefills` returns, and decode runs later still in the tick's own decode step. So a stream that finished in wave 1 sat until waves 2 and 3 had also run, and every TTFT in the burst landed on the slowest. Promoting inside the wave loop buys nothing either — `promote_completed_prefills` removes from `prefilling`, which invalidates the wave indices mid-loop, and decode still would not run until the phase returned. The tick boundary is where the scheduler already interleaves, so that is where the cut goes.

With the cap in: wave 1 runs, its finished streams promote at the end of that tick, decode interleaves from the next one, and the streams that did not fit re-plan next tick — where they batch among themselves, because the planner re-runs from the live geometry every tick. On the round-15 shape, sixteen 1193-token prompts plan 7/7/2; seven prefill on tick 1, and on tick 2 those seven stand first in FIFO order and their tails complete as one 175-token forward before any further head wave is issued.

Fairness is unchanged in kind: the planner is first-fit in FIFO order, so wave 1 always holds stream 0 — the same guarantee the single-stream `prefilling.first_mut()` path gives — and it now carries everyone who fits with it. With the flag OFF the planner returns one wave holding every stream and `waves_this_tick` returns it whole, so that path stays byte-identical.

### The receipt this came from (H100 round 15, §3.7 / attribution §6)

Round 15 rebuilt with the slot-aliasing fix in. Structurally the lever came back **clean** — 0 `SSM pool slot SHARED`, 0 double-frees, 0 watchdog fires, 16/16 responses at the full 256 tokens — and it was still a loss on the short shape whenever it engaged:

| `1024x256` C=16 | A15 (no lever) | V15 (engaged) | Δ |
|---|---:|---:|---:|
| aggregate | 513.86 tok/s | 427.53 tok/s | **−16.8%** |
| TTFT p50 | 1 359.4 ms | 4 141.2 ms | **3.05×** |
| TTFT p99 | 2 504.5 ms | 4 142.2 ms | +65% |
| TPOT | 25.90 ms | 21.28 ms | **−17.8%** |

**TPOT moved the other way**, which is the whole diagnosis: batching the prefill up front genuinely removes the prefill/decode interference, so the batching works and the scheduling of it did not. The tell is **p50 = p99** — `16 streams -> 3 wave(s)` in one tick.

### The admission-reason log line

Round 15 also saw engagement vary two-of-four across otherwise identical bursts (427 vs 514 tok/s, a 17.52% rep spread) with nothing in the log to say which was which. The planner is not the source — `plan_prefill_waves` is a pure function of the geometries. The ADMISSION is: it needs `active.is_empty()` and two chunk-0s co-admitted in the same tick, and a burst whose first request has already been promoted to decode by the time the rest arrive fails the first test. That is arrival timing against the tick period — inherent, not pinnable — so it is now **reported** rather than chased. `varlen_admission` returns the verdict plus the one input that decided it, and `phase_start_prefills` logs it once per burst:

```
Varlen prefill admission: defer=false reason="decode already active this tick (arrival timing)" new_reqs=9 prefilling=0 active=7 smallest_chunk0=[1193, 1193] cap=8192
```

A run's engagement is now readable from its serve log instead of inferred from its throughput. The long-shape `deferral SKIPPED` line is unchanged and still fires verbatim, so the campaign greps keep working.

### Tests

`prefill_waves_tick_tests.rs`, 7 tests:

- `only_the_first_wave_prefills_on_tick_one`
- `the_first_waves_streams_reach_decode_before_the_second_waves_forward`
- `deferred_streams_batch_among_themselves_next_tick`
- `flag_off_still_dispatches_every_stream_in_one_tick`
- `the_long_shape_deferral_is_still_skipped_for_the_same_reason`
- `an_active_decode_is_what_flips_engagement_between_bursts`
- `every_admission_rung_names_itself`

### Gate (Spark 1, GB10, `--features cuda,nccl`)

```
$ cargo fmt --all -- --check
fmt exit=0

$ cargo clippy --jobs 8 -p spark-server --features cuda,nccl --all-targets -- -D warnings
    Checking spark-server v1.0.0-beta-preview (/home/pidtom/atlas-gate-1002b/crates/spark-server)
    Finished `dev` profile [unoptimized] target(s) in 43.13s
clippy exit=0

$ cargo test --jobs 8 -p spark-server --bins --features cuda,nccl
     Running unittests src/main.rs (target/debug/deps/spark-b649d9e33ebfc841)
test scheduler::phase_continue_prefills::prefill_waves::tick_tests::only_the_first_wave_prefills_on_tick_one ... ok
test scheduler::phase_continue_prefills::prefill_waves::tick_tests::the_first_waves_streams_reach_decode_before_the_second_waves_forward ... ok
test scheduler::phase_continue_prefills::prefill_waves::tick_tests::deferred_streams_batch_among_themselves_next_tick ... ok
test scheduler::phase_continue_prefills::prefill_waves::tick_tests::flag_off_still_dispatches_every_stream_in_one_tick ... ok
test scheduler::phase_continue_prefills::prefill_waves::tick_tests::the_long_shape_deferral_is_still_skipped_for_the_same_reason ... ok
test scheduler::phase_continue_prefills::prefill_waves::tick_tests::an_active_decode_is_what_flips_engagement_between_bursts ... ok
test scheduler::phase_continue_prefills::prefill_waves::tick_tests::every_admission_rung_names_itself ... ok
test result: ok. 2488 passed; 0 failed; 12 ignored; 0 measured; 0 filtered out; finished in 2.21s
test exit=0

$ file-size cap (500 LoC per .rs under crates/, workflow allow-list applied)
scanned 2006 file(s); 0 non-allow-listed file over 500 LoC
this commit's .rs files:  499 phase_start_prefills.rs / 425 phase_continue_prefills.rs /
                          286 prefill_waves.rs / 279 run_batched_prefill.rs /
                          205 prefill_waves_tick_tests.rs
```

### Scope

**The lever stays default OFF.** §5's finding stands — `M=8176` (7 prompts fused) measured 190.4 µs/token against 188.8 µs/token at `M=1168`, 0.8% *worse* — so varlen repackages the GDN work rather than reducing it. This commit removes the TTFT collapse layered on top of that; it does not make batching pay, and the C=16 gap is still not a prefill-batching deficit.

**The H100 re-measure is owed.** Everything above is round 15 data plus the reasoning and unit pins for the fix; the fix itself has not been measured on H100 — that box is down. The re-measure (short shape, lever forced on, 4 reps) should show TTFT p50 decoupling from p99 while TPOT holds its 21 ms, and it is a precondition for any conclusion about the lever beyond "still off by default".

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1044 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1044; this fork PR is closed.

PR #1004, opened 2026-09-12
bench: base the ladder's prompt nonce per process, not at zero (#928)

TheTom:
## Summary

The W55 concurrency ladder numbered its prompts from **zero on every run**, so a second run against a still-warm server re-sent the first run's prompts and was measured **through the prefix cache**. TTFT then reads as an engine win when it is a cache hit, and two A/B cells driven against one server are not comparable at all.

This makes the nonce base **random per process**, records it in the output JSON, prints it on the banner, and adds `--check-shapes` — a server-free self-test that asserts the nonce base changes no prompt's token count, so the ISL a cell claims is the ISL it sent.

**Base is `main`.** No stack: the only Rust this touches is none — the diff is one file, `bench/ladder38/harness_w55_conc_ladder.py`.

Refs #928.

## What was wrong

`prompt_for(i)` built its nonce from a counter that started at 0 in every process. Consequences, in the order they bite:

1. **Cross-run prefix-cache contamination.** Run A sends prompts `000000…`, `000001…`; run B against the same server sends the same bytes. With `--enable-prefix-caching true` (which the campaign recipe sets) run B's TTFT is a cache lookup. This is not hypothetical — it is how an A/B against one warm server was being driven.
2. **A pinned base could not be stated.** Nothing in the output said which prompts a run sent, so a receipt could not show that two cells had *not* shared a cache entry.
3. **Any fix here can silently change the shape.** A wider or differently-formatted nonce can cross a tokenizer boundary and move the prompt's token count, which moves ISL, which moves every per-token number derived from it.

## What the fix does

- **Random base per process.** `default_nonce_base()` draws it; `set_nonce_base()` installs it. `--nonce-base N` pins it, and pinning is only for reproducing a specific run's prompts.
- **It is recorded, not just done.** The run JSON gains `nonce_base`, `nonce_base_pinned` and `nonce_width`, and the banner prints `# nonce base NNNNNN (random|pinned)`. A receipt can now state which prompts a cell sent.
- **`--check-shapes` is the guard for cause 3.** It tokenizes the prompt across a spread of nonce bases and fails if any of them changes the token count, then exits. It needs **no server and no aiohttp** — which is why the measurement knobs stop being argparse-`required` and are checked explicitly instead. The file's rule that no measurement knob is ever defaulted silently still holds; the script just reports the missing ones itself, after `--check-shapes` has had its chance to run without them.
- **`--enable-thinking` rides along.** The nonce work rewrote the same request builder, and the ladder had no way to drive a thinking-mode cell at all. Default is unchanged (`false`), and the mode is recorded in the run's `chat_template_kwargs`.

## Oracle and evidence

`--check-shapes` is the oracle, and it is in the diff: it is the only thing standing between "the nonce is now random" and "the nonce silently changed ISL". It needs a local tokenizer directory (`--tokenizer`, or `W55_TOKENIZER`) and is offline-only by construction.

**No new hardware number is claimed here.** This is the measurement harness, not the engine. What it changes is the *validity* of numbers taken after it: every ladder cell recorded before this fix, against a server that had already served a ladder, has an unquantified prefix-cache component in its TTFT.

## Test plan

Spark 1 (GB10), in an isolated worktree against this branch fetched from the fork.

- [x] `python3 -m py_compile bench/ladder38/harness_w55_conc_ladder.py` — **rc=0**
- [x] `python3 bench/ladder38/harness_w55_conc_ladder.py --help` — **rc=0** (proves the argparse rework still parses with no measurement knob supplied, which is the whole point of dropping `required=True`)
- [ ] `--check-shapes` against a real tokenizer — **not run here**: it is offline-only and needs a local tokenizer directory, which the CPU gate box does not carry. It has been run on the H100 box during the campaign.
- [x] No `crates/` change, no `kernels/` change: `git diff main --stat` is one file under `bench/`.

## GB10 perf-gate records

No `crates/` path is touched, so no committed `.benchmarks/` record is invalidated by this diff.

## What is NOT claimed

- No engine change, no kernel change, no numerics change.
- No new measurement is presented. The ladder numbers quoted elsewhere in the Hopper campaign were taken with this fix in place from `6fcedf2d7` onward; earlier cells are the ones this PR's rationale calls into question, and re-stating them is out of scope here.
- `--check-shapes` proves the token count is invariant to the nonce base **for the tokenizer it is given**. It is not a claim about every tokenizer.

## CLA

### DIFF
--- a/bench/ladder38/harness_w55_conc_ladder.py
+++ b/bench/ladder38/harness_w55_conc_ladder.py
@@ -12,7 +12,9 @@ is the point: two harnesses measuring two engines is not an A/B):
   * reps per rung recorded individually (raw series), never only the mean
   * temperature 0, seed 42, token-matched via a forcing suffix + max_tokens cap
   * chat_template_kwargs.enable_thinking=false on BOTH engines
-  * per-request nonce so enable_prefix_caching cannot serve a repeat from cache
+  * per-request nonce, based at a RANDOM per-process value, so
+    enable_prefix_caching cannot serve a repeat from cache — not within a run
+    and not across two runs against the same server (see make_prompt)
   * completion_tokens/prompt_tokens read from the usage frame, not counted deltas
     (Atlas batches a short reply into ONE SSE delta)
 """
@@ -22,11 +24,19 @@ import asyncio
 import hashlib
 import json
 import os
+import random
 import statistics
 import sys
 import time
 
-import aiohttp
+# OPTIONAL on purpose: `--check-shapes` grades prompt construction and needs no
+# server, so it must run on a box (or a CI runner) that has no HTTP client
+# installed. Every measurement path still refuses to start without it — see the
+# guard in main().
+try:
+    import aiohttp
+except ImportError:  # pragma: no cover - exercised by the --check-shapes gate
+    aiohttp = None
 
 # ── pinned constants (recipe: benchmark.prompt / benchmark.sampling) ──
 #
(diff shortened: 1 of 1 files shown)


TheTom:
Thomas restacked this onto origin as #1024 (`stack/mega-*`) so Graphite can stack. Fork PR closed; please use #1024.

PR #1006, opened 2026-09-12
atlas-kernels: per-target serving defaults, SM count and declared kernel overrides, baked from HARDWARE.toml (#928)

TheTom:
> **Stacked on #996 — review the diff against it.** This PR's base is `main` (GitHub stacks by content, and a stack branch base makes the PR unreviewable if #996 is rebased), so the file list here also contains #996's Hopper/B200 targets. This PR's own content is the **six commits above `c16ccaab8`**: `git diff c16ccaab8..`. Merge order: **#996 → this**.
>
> #996 has to be in the base because this PR declares `[defaults]`, `[hardware] sm_count` and `[kernels] overrides` in `kernels/hopper/HARDWARE.toml` and `kernels/b200/HARDWARE.toml`, and those two files do not exist on `main` yet.

## Summary

**An H100's serving configuration lived in a launch script outside this repository.** This PR makes it a property of the target: `kernels/<hw>/HARDWARE.toml` gains a `[defaults]` table, `build.rs` bakes it into `atlas_kernels::TARGET_DEFAULTS`, and every resolver in `spark-model` reads that declaration **before** the environment.

Maintainer review of the H100 integration branch, 2026-09-11 (tbraun96):

> There is no arch separation at all. H100 builds compile GB10's kernel tree. Every Hopper/GB10 divergence is expressed as an env lever set by an H100 recipe living outside this repo — not as arch-selected code. "No interference" rests on discipline rather than structure.

`build.rs` reads exactly **one** `kernels/<hw>` tree, so a GB10 binary cannot carry another target's numbers, an H100 serve with an empty environment reproduces its measured configuration, and every value is reviewable beside the arch and memory facts it belongs to. `spark-server` prints one `target defaults (<hw>): …` line naming every resolved value and which of them came from the environment.

This is **the declaration mechanism only**. The Hopper-tuned kernel files are not here; each lands in its own PR with its own row (see **The contract**).

`Refs #927 #928`

## What changed

**(a) `[defaults]`, three rows.** Parsed by `crates/atlas-kernels/build_defaults.rs`, baked as `atlas_kernels::TARGET_DEFAULTS`, resolved by `spark_model::layers::ops::target_defaults`.

| row | consumer | gb10 | b200 | hopper |
|---|---|---|---|---|
| `lm_head_batchm_max` | `model/trait_impl/lm_head_batched.rs` | 8 | 8 | 8 |
| `ssm_batched_recurrent` | `layers/qwen3_ssm/gdn_flags.rs` | false | false | **true** |
| `decode_split_silu` | `ModelLevers` (`model_levers_resolve.rs`) | true | true | true |

`kernels/gb10` declares **exactly** `build_defaults::baseline` — asserted as an equality against that function, not a list somebody has to remember to update — so a GB10 serve with an empty environment is unchanged by construction. `kernels/b200` declares the same conservative table: it has no serving receipt of any kind, and copying a recipe across because both cards are datacentre parts is the reasoning this whole mechanism replaces.

**Hopper differs on one row**: the batched GDN recurrence, ON, at +6% on the serve with md5-identical output to the per-sequence launches. That row was `ATLAS_SSM_BATCHED_RECURRENT=1` in the external launch script.

**The head band deliberately holds at the frozen 8 on Hopper**, and not the 16 the H100 recipe exported. That 16 was measured with the tensor-core head arm (`dense_gemm_m16_bf16`, #927) **also on**, where the TC arm serves 5..=16 and the band is very nearly inert. Without the arm — it is a Hopper-owned kernel and is not in this PR — a 16 here would move widths 9..=16 off the reassociating tile GEMM onto the batched GEMV in a configuration nobody has measured. The row moves with the commit that lands the arm it was measured beside.

**The grammar changes in one direction.** `ATLAS_SSM_BATCHED_RECURRENT=0` now means OFF; it was `== "1"`, so `=0` was indistinguishable from absent. Presence cannot express "off", and once a target's default can be ON an operator with no way to turn a lever off is back to editing launch scripts. `VAR=1` is unchanged everywhere, and `ATLAS_NO_DECODE_SPLIT_SILU` stays PRESENCE-gated and still outranks the declaration.

**(b) `[hardware] sm_count` → `TARGET_SM_COUNT`, and a boot cross-check.** 48 / 132 / 148. `atlas_core::device::sm121::NUM_SMS = 48` — a constant named after **one** part — was what every grid-sizing caller had, whatever hardware the build was for, and nothing on either side ever said a number out loud. `arch_preflight` now reads the driver's `CU_DEVICE_ATTRIBUTE_MULTIPROCESSOR_COUNT` and **warns** on a mismatch, naming both numbers and the file the declaration came from. It warns and does not fail: a wrong count costs a suboptimal grid, never a wrong answer, and refusing to boot on an H100 PCIe with a different bin would be a regression dressed as rigour. A driver that will not answer costs one cross-check, not the boot.

**(c) The `[kernels] overrides` SSOT.** `kernels/hopper/common/` and `kernels/b200/common/` are relative symlink mirrors of `kernels/gb10/common/`. Maintainer rule, 2026-09-11: *"symlinks are fine provided the pointed-to gb10 file is not edited when iterating on Hopper; Hopper-tuned kernels must be real files under `kernels/hopper/`."* The moment a target **may** carry a real source, an accidental copy and a deliberate tuning are the same bytes on disk — and a fork of `common/` diverges for every model on the target at once.

So divergence is declared, and **one list** serves three readers that would otherwise keep three:

- `scripts/check_kernel_shadows.py` gains **RULE 3** — an undeclared regular file in a mirrored `common/` is a violation, and so is a declared entry that has vanished or no longer resolves. On a clean run it prints each target's list, marking each entry `(replaces)` when gb10 carries the same name and keeps its own copy, or `(adds)` when it does not.
- `mirror_faults` takes the declared set instead of demanding a bijection, and its own fixtures prove the exception is narrow **in both directions**: take the declaration away and every allowance turns straight back into a fault.
- `Inherited::owned()` reads the TOML rather than a Rust constant beside it. Two spellings of "which kernels does this target own" is how a file comes to be declared in one and forgotten in the other.

`tests/inherited_overrides.rs` checks what a declared entry **is** — an OVERRIDE is a real file whose gb10 namesake is untouched and whose bytes differ; an ADDITION brings entry points gb10 does not declare; a name two targets share has one real file with the rest linking to it.

**Nothing is declared yet.** Both lists are empty on this base — every inheriting target still takes all of gb10's kernels — so those cases iterate nothing. Deliberately **no** emptiness assertion: it would have to be deleted by the first PR that adds a tuned kernel, which is the worst moment to be editing the test that guards it. That the rules can **fail** is proved on fixtures in `support/mirror.rs`.

**(d) The build summary line.** Round 11 asked for the resolved kernel count to be visible in a normal build; round 13 found `build-r13.log` carried no kernel line at all and the count — 196 — had to be dug out of `target/release/build/atlas-kernels-*/output`. Round 15 then found the line's second field **mislabelled**: it read `14 declared overrides` while that was `find_cu_files(model_kernel_dir).len()`, a number that never tracked the overrides list and so could never answer "did my override land".

```
atlas-kernels: 199 kernels (hopper, qwen3.8-27b, nvfp4), 14 model-dir kernels, 15 declared overrides
```

Both counts print, each named for what it is, and zero prints as `0` instead of the clause vanishing — the predecessor suppressed it entirely at zero, so "declares none" and "did not say" were the same bytes in a log.

**(e) `w8a8_prefill_max_m_*` is NOT here.** Those rows are Thomas's (#917, `d4112c3eb`) and their consumer `dense_ffn_w8a8_prefill.rs` (#987) is not on this base. Left for that PR, under the same contract.

## The contract

**A lever row is six things in ONE commit:**

1. the field in `atlas_kernels::TargetDefaults` (`crates/atlas-kernels/src/target_defaults.rs`)
2. the parse arm in `crates/atlas-kernels/build_defaults.rs`
3. the row in **every** `kernels/<hw>/HARDWARE.toml` that has a `[defaults]` table
4. the `TargetLevers` field and its arm in `spark_model::layers::ops::target_defaults::resolve`
5. the field in the `target defaults (<hw>): …` boot line (`format_levers`)
6. a test

`parse_defaults` **panics** on an unknown `[defaults]` key, so a half-landed lever fails the build instead of reading as agreement with the baseline — which is what makes "one commit" enforceable rather than merely asked for. `inherited_targets.rs` separately requires all three tables to declare the same key set.

**★ And that commit is the one landing the lever's CONSUMER.** That is why this PR carries three rows and not seventeen. A row whose dispatch site does not exist cannot be graded, answers nothing when an operator sets its variable, and would put an ` (env)` tag in the boot line against a decision that changes no code. The rows deferred to the PRs that own their arms:

| deferred row | waits for |
|---|---|
| `cublas_gemm_scope` | the `CublasScope` grammar (`parse_cublas_scope`) |
| `ffn_batch16_tier` | `dense_ffn_batch16_decode.rs` |
| `ffn_m16_tc`, `attn_m16_tc` | `w8a16_gemm_m16` |
| `attn_ncol_gemv` | `attn_ncol_gemv.rs` |
| `lm_head_m16_tc` | `dense_gemm_m16_bf16` (and it carries `lm_head_batchm_max = 16` with it) |
| `gdn_decode_hopper` | `ssm_gdn_hopper.rs` |
| `gdn_prefill_tc` | the tensor-core GDN prefill route |
| `ssm_ba_gates_hopper` | `ssm_ba_gates_hopper.rs` |
| `ffn_gateup_fused` | `dense_ffn_gateup_fused.rs` |
| `attn_decode_splitk` | `atlas_kernels::attn_splitk` + the split-K dispatch |
| `ssm_decode_ring_slots` | `ssm_reserve::set_decode_ring_slots` (#915) |
| `w8a8_prefill_max_m_*` | `dense_ffn_w8a8_prefill.rs` (#987 / #917) |

The contract is written into `crates/atlas-kernels/src/target_defaults.rs` and `crates/spark-model/src/layers/ops/target_defaults.rs` so each stacked PR follows it rather than finding its row pre-declared and inert.

## Receipt

The machinery was exercised end-to-end on 1×H100 80 GB HBM3 (`atlas-testing-jawn`) at PR #895's head, with the **full** lever table — this PR ships a subset of those rows, so the receipt is for the MECHANISM, not for the three rows' values.

**Round 14 — defaults reproduce the explicit-env recipe.** Cell **D14** (`h100-defaults`: env `ATLAS_DENSE_FP8=1` only, flag `--lm-head-dtype bf16` only) against cell **T14** (the explicit-env twin `h100-r12` + `ATLAS_GDN_PREFILL_TC=1`), same binary, same box, same driver:

- largest deviation across all eight rungs: **0.37 %** (1193/256 C=1 TTFT, 162.4 vs 163.0 ms), against per-cell rep spreads of 0.03–2.06 %;
- coherency: **md5-identical on 7/7 exchanges**;
- D14's boot line carries **no ` (env)` suffix anywhere**, and the GDN route log picks the tensor-core spine and both remnant twins 14 400 / 1 / 1 times with no environment variable set — T14's counts are identical.

Env and defaults are the same code path, measured.

**Round 15 — the boot line, verbatim**, cell A15 (`h100-defaults`):

```
target defaults (hopper): cublas_gemm_scope=ffn,attn,ssm ffn_batch16_tier=off ffn_m16_tc=off attn_m16_tc=on attn_ncol_gemv=off lm_head_m16_tc=on lm_head_batchm_max=16 ssm_batched_recurrent=on gdn_decode_hopper=off gdn_prefill_tc=on ssm_ba_gates_hopper=on decode_split_silu=on ssm_decode_ring_slots=auto w8a8_prefill_max_m=max/max attn_decode_splitk=auto
```

This PR's line is the three-row subset of it, plus the target's `sm_count`.

## Tests

**`crates/atlas-kernels/tests/target_defaults.rs`** (14 cases) compiles `build_defaults.rs` **directly against the real `kernels/*/HARDWARE.toml`** — cargo never runs a build script's own unit tests, so a rule that lives only inside `build.rs` is a rule nothing tests.

- `gb10_declares_the_baseline_and_nothing_else` — the regression gate, as an **equality** against `baseline()`.
- `hopper_declares_what_an_h100_serve_runs_with`, `b200_declares_the_conservative_table_not_hoppers`, `the_silent_targets_resolve_to_the_baseline` (metal, strix, strix-hip declare nothing and are byte-for-byte unaffected).
- `every_declaring_target_states_every_lever` — no row may be inherited from the baseline by omission.
- `an_unknown_lever_name_fails_the_build`, `a_mistyped_value_fails_the_build_naming_the_key`, `a_zero_sm_count_fails_the_build` — `should_panic`, so the enforcement is observed and not assumed.
- `every_target_declares_the_sm_count_of_its_own_part`, `the_baked_constant_matches_its_own_hardware_tree` — the join between the generator and the runtime.

**`crates/spark-model/src/layers/ops/target_defaults_tests.rs`** (10 cases) grades the RESOLVER on tables spelled out in Rust, so it is answerable for any target from a CPU test: `gb10_with_an_empty_environment_is_todays_behaviour`, `a_declared_on_lever_can_be_turned_off_by_the_environment` (`0`/`false`/`off`/`no`/`OFF`/` 0 `), `a_declared_off_lever_is_still_armed_by_the_bare_one`, `the_legacy_kill_switch_still_forces_the_lever_off`, `the_head_band_clamps_and_has_no_off`, `the_summary_line_names_every_lever_and_flags_the_environment`, `an_anonymous_build_still_prints_a_readable_line`, `the_baseline_band_is_the_frozen_one` (the join with `gemm_quant.rs`'s frozen constant).

**`crates/spark-runtime/.../arch_preflight_tests.rs`** — `a_matching_sm_count_says_nothing`, `a_mismatched_sm_count_warns_naming_both_numbers`, `the_cross_check_fires_in_both_directions`, plus the existing call-sequence case now pinning that the new query is addressed by the same ordinal and runs after the arch verdict.

**`crates/atlas-kernels/tests/support/mirror.rs`** — three new fixture cases proving the override exception can FAIL: `a_declared_override_is_accepted_as_a_file_or_a_link_elsewhere`, `a_declaration_that_does_not_hold_is_a_fault` (declared-and-missing, declared-and-dangling, declared-but-pointing-back-at-the-origin), `the_declared_exception_is_narrow` (remove the declaration and every allowance becomes a fault again).

**`crates/atlas-kernels/tests/build_summary.rs`** (7 cases) — the line pinned whole, the two counts proved independent, zero still printing, one line per target triple.

`arch_preflight.rs` was at 493 of the 500-LoC cap, so its tests move to `arch_preflight_tests.rs`. The split is by QUESTION — the module states the rules, the test file drives them against injected drivers.

## Test plan

All gates on a DGX Spark (GB10), private `CARGO_TARGET_DIR`, `ATLAS_SKIP_BUILD=1`, `--jobs 8`, at `d3898e21`.

```
$ cargo fmt --all -- --check
fmt rc=0

$ cargo clippy -p atlas-kernels --all-targets -- -D warnings
    Finished `dev` profile [unoptimized] target(s) in 3.34s
rc=0

$ cargo clippy -p spark-model --features cuda --all-targets -- -D warnings
    Finished `dev` profile [unoptimized] target(s) in 14.44s
rc=0

$ cargo clippy -p spark-server --features cuda,nccl --all-targets -- -D warnings
    Finished `dev` profile [unoptimized] target(s) in 32.09s
rc=0

$ cargo test -p atlas-kernels --no-default-features
test result: ok. 14 passed; 0 failed          (tests/target_defaults.rs)
test result: ok. 7 passed;  0 failed          (tests/build_summary.rs)
test result: ok. 3 passed;  0 failed          (tests/inherited_overrides.rs)
test result: ok. 10 passed; 0 failed          (tests/target_resolution.rs)
rc=0

$ cargo test -p spark-model --features cuda --lib
test result: ok. 899 passed; 0 failed; 11 ignored
rc=0

$ cargo test -p spark-server --bins --features cuda,nccl
test result: ok. 2462 passed; 0 failed; 12 ignored
rc=0

$ cargo test -p spark-runtime --lib --features cuda
test cuda_backend::arch_preflight::tests::a_matching_sm_count_says_nothing ... ok
test cuda_backend::arch_preflight::tests::a_mismatched_sm_count_warns_naming_both_numbers ... ok
test cuda_backend::arch_preflight::tests::the_cross_check_fires_in_both_directions ... ok
test cuda_backend::arch_preflight::tests::the_ordinal_addressed_query_preflights_from_any_thread ... ok
test result: ok. 308 passed; 0 failed; 4 ignored
rc=0

$ python3 scripts/check_kernel_shadows.py
kernel shadow structure: OK (6 hardware trees scanned: b200, gb10, hopper, metal, strix, strix-hip)
  b200/common inherits every kernel from gb10/common
  hopper/common inherits every kernel from gb10/common
rc=0

$ python3 scripts/check_cross_hardware.py --base c16ccaab8 --head d3898e21
cross-hardware reach: 3 changed kernel path(s)
  kernels/b200/HARDWARE.toml     b200     -        0  OK
  kernels/gb10/HARDWARE.toml     gb10     gb10    27  OK
  kernels/hopper/HARDWARE.toml   hopper   -        0  OK
verdict: OK
rc=0

$ scripts/hopper_ptx_gate.sh --hw hopper --model qwen3.8-27b --jobs 8 --strict
self-test @ sm_90a: known_good passed=true  known_bad_post_hopper.cu failed=true
188 kernel(s) across 1 model(s); 188 unique compile(s) after dedup
compiling for sm_90a with 8 job(s), strict=1 — Build cuda_13.0.r13.0/compiler.36424714_0
188/188 kernels compiled for sm_90a; 0 failed (0 rejected entry function(s))
rc=0
```

**CI file-size cap**: no new file over 500 LoC. The two files in the tree that are over it (`crates/atlas-kernels/build.rs`, `crates/spark-comm/src/nccl_backend.rs`) are both already on the workflow's allow-list; `arch_preflight.rs` went from 493 to 277 by the split above.

- **No kernel is added, removed or edited.** Not one `.cu` file changes. The Hopper-tuned sources the integration branch carries are not here.
- **No performance claim for the three rows.** `decode_split_silu` and `lm_head_batchm_max` are the baseline on every target, so they are no-ops by construction. The one value that differs — `ssm_batched_recurrent` on hopper — carries its own +6%/md5-identical receipt, which predates this PR; the round-14 D14/T14 numbers above attest that **defaults and env are the same code path**, not that this subset is faster than anything.
- **GB10 behaviour is unchanged**, and that is asserted as an equality against `build_defaults::baseline`, not argued.
- **B200 is unmeasured** and declares the conservative table on purpose.
- **`TARGET_SM_COUNT` has no dispatch consumer in this PR.** Its consumer here is the boot cross-check. The grid-sizing rule that will read it (`attn_splitk`) arrives with the split-K PR.
- **`[kernels] overrides` is empty on every target.** The mechanism is landed; nothing is declared into it yet.
- The `hopper_ptx_gate` run above says these kernels **compile** for sm_90a. It says nothing about whether they produce the right numbers or run well.

## CLA


TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1046 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1046; this fork PR is closed.

PR #1007, opened 2026-09-12
spark-model: ledger or delete every off-ledger FP8 weight dequant (#917)

TheTom:
## Summary

Two commits, one property: **every FP8→BF16 weight dequant Atlas performs is now either deleted or owned by the buffer ledger.**

`ops::cublas_bf16_proj` allocated a BF16 twin of an FP8 weight — 2 bytes per weight element — lazily, memoised by weight pointer, with **no `BufferSizes` entry**. `--gpu-memory-utilization` could not see it, the #915 ring fitter could not price it, and on the 2026-09-11 H100 receipt (`Qwen/Qwen3.8-27B-FP8` rev `017b9c7a`) a **28-token prefill consumed 6120 MiB and died at layer 36** with `cuMemAlloc_v2 failed: status 2`.

1. **The attention arms were already deleted** (by #927 and #928's W8A8 replacements). Nothing held them deleted. `prefill/alloc_tests.rs` drives both chains end to end on a mock backend with `ATLAS_CUBLAS_GEMM=attn` armed and **fails if either allocates at all**.
2. **The two `ATLAS_FP8_ROWWISE` GDN prefill arms could not be deleted** — a per-row FP8 checkpoint has no other route while `cublaslt::fp8_gemm_act_weight_t_rowwise` returns `NOT_SUPPORTED` on sm_121 and a per-row scale is not the `[N/128, K/128]` grid the block-scaled W8A8 kernels index. So the bytes **move into the arena** instead of out of the model: a new ledgered slab, sized from the same env predicate the loader uses, zero unless `ATLAS_FP8_ROWWISE=1`.

`ops::cublas_bf16_proj` is deleted. After this, the only surviving caller of `dequant_fp8_bf16_cached` is `ops::cutlass_bf16_proj` under `ATLAS_CUTLASS_GEMM` on a build with `CUTLASS_HOME` — a benchmarking reference path no shipping recipe can enter, documented as such with a "do not add callers" note.

Refs #917. Refs #915.

### Stack

```
main
 └─ #987  pr/ffn-w8a8-prefill
     └─ #989  pr/cublaslt-vec128-and-lever-scope
         └─ #984 → #986 → #988 → #994
             └─ #998  pr/prefill-proj-w8a8
                 └─ THIS PR  pr/offledger-dequants
```

**Base branch is `main`, so the diff shown by GitHub includes the whole stack.** This PR's own two commits are at the tip; review it after #998. It genuinely needs that depth: the attention allocation test arms `CublasScope`, which is #989's per-family lever scoping, and it asserts the *absence* of the arms that #927/#928 replaced with `cache_skip_qkv_w8a8_gemm` and `attn_prefill_w8a8_gemm`, both of which arrive on #994/#998.

## What was wrong

`dequant_fp8_bf16_cached` did a bare `gpu.alloc`, memoised by weight pointer. At the 27B geometry that is **167,772,160 B for the fused `[QKV|Z]` weight per layer** (`[10240,5120] + [6144,5120]`, ×2 B) × 48 GDN layers ≈ **10.3 GiB**, none of it in the ledger.

Two consequences, and the second is the one that keeps biting:

- **Preflight cannot price it.** `headroom::post_load_yardstick` reads `BufferSizes::total_bytes()` as its `arena` term. An allocation outside `BufferSizes` is invisible to the ring fitter, so the fitter's verdict is wrong by exactly the amount that will OOM later.
- **The cache defeats the obvious test.** The allocation happened **once per weight pointer**, so a "call it twice and compare" test passes on a leaking build. Any regression pin has to snapshot **before the first call**.

### Commit 1 — hold the attention arms deleted

`crates/spark-model/src/layers/qwen3_attention/prefill/alloc_tests.rs` (288 lines, new): a mock backend, an FP8 block-scaled attention layer, `CublasScope::ALL` (which is what `ATLAS_CUBLAS_GEMM=attn` resolves to), and the **snapshot taken before the first call**. It brackets the chains at their dequant seam rather than through the cuBLASLt matmul that follows — this binary also builds real CUDA backends, and handing `cublasLtMatmul` a mock's fabricated device pointers would fault a shared context.

The transposed and FP8×FP8 twins are left absent on purpose, so the chains resolve the arms the H100 trace actually shows (`w8a16_gemm_pipelined` for Q/K/V) rather than a path the receipt never took.

### Commit 2 — ledger the row-wise arms

- **`BufferSizes::ssm_rowwise_w_bf16`** (new `buffers/sizes_rowwise.rs`) sizes ONE slab at `num_ssm_layers × (in_proj_qkvz + out_proj)`, and is **0 unless `ATLAS_FP8_ROWWISE=1`** — character for character the loader's own `rowwise_fp8_enabled` predicate, so **the default recipe's ledger is byte-identical to before this PR**.
- **`BufferArena::take_ssm_rowwise_w_bf16`** bump-carves a layer's slice; each layer fills its two slices on its first prefill and holds them in its own fields (`qwen3_ssm/rowwise_bf16.rs`). **Exhaustion and an absent slab are errors, never a fresh allocation** — that is the rule that makes the ledger the truth.
- **`ops::dequant_fp8_bf16_into`** becomes the single FP8→BF16 weight expansion. It takes a destination, so ownership is the caller's decision. The arms multiply through `cublas_bf16_proj_dense`, which cannot allocate.
- **`BufferArena::from_sizes`** lets a test arm an env-gated ledger entry without `set_var`, which is process-global and would race the binary.

## Oracle and evidence

Red before the change, green after:

| test | what it pins |
|---|---|
| `qwen3_attention/prefill/alloc_tests.rs` | both attention chains allocate **nothing** with `ATLAS_CUBLAS_GEMM=attn` armed — snapshot before the first call |
| `qwen3_ssm/rowwise_alloc_tests.rs` | per-arm mock-backend allocation contracts for both row-wise arms, same before-first-call discipline |
| slab-exhaustion pin | a short slab is an **error**, not a fallback allocation |
| end-to-end refusal | both real arms refuse when the ledger entry is absent |
| `buffers/tests.rs` sizing pins | exact-integer slab size at the 27B geometry |

**The H100 OOM is the historical receipt, not a new measurement.** The box for this campaign is down; nothing here re-measures it. What this PR adds is the property test that keeps the fix from silently regressing — the original leak survived a code review and a round of serving precisely because nothing asserted "this path allocates zero bytes".

## Test plan

- [x] `cargo clippy -p spark-model -p spark-runtime --features cuda --all-targets --jobs 8 -- -D warnings` — **clean, rc=0**
- [x] `cargo test -p spark-model --features cuda --lib` — **998 passed, 0 failed, 11 ignored** (base #998: 992 → **+6**)
- [x] `cargo test -p spark-runtime --features cuda --lib` — **303 passed, 0 failed, 3 ignored** (base #998: 301 → **+2**)
- [x] `cargo test -p spark-model --features cuda --lib -- alloc` — **20 passed, 0 failed**, 989 filtered
- [x] `cargo test -p spark-runtime --features cuda --lib -- rowwise` — **2 passed, 0 failed**, 304 filtered
- [x] File-size cap (`.github/workflows/file-size-cap.yml` allow-list applied) — **0 violations**
- [x] No `kernels/` change: `git diff` against the stack parent touches only `crates/`.

## GB10 perf-gate records

This diff touches `crates/` (a PERF_PATH), so `record_covers` invalidates committed `.benchmarks/` records on landing. **The default recipe's behaviour is unchanged**: the new slab is 0 bytes unless `ATLAS_FP8_ROWWISE=1`, and the attention commit is tests only. A GB10 re-measure is owed as a formality, and the `ATLAS_FP8_ROWWISE=1` arm itself is unmeasured on GB10.

## What is NOT claimed

- **No perf claim.** Neither commit is a speed change. Commit 1 is tests only; commit 2 moves where bytes live.
- **`dequant_fp8_bf16_cached` is not deleted.** One caller survives (`ops::cutlass_bf16_proj`, `ATLAS_CUTLASS_GEMM` + `CUTLASS_HOME`). That arm is still off-ledger and this PR says so rather than pretending the class is closed.
- **The `ATLAS_FP8_ROWWISE` arm is not certified.** It is ledgered, sized and error-checked; it has no serving receipt on either GB10 or H100 in this PR.
- No coherency gate or ladder cell is presented — the H100 box for this campaign is down and nothing here needed one.

## Conflicts

Cherry-picked from `fix/attn-cublas-offledger-dequant` and `fix/rowwise-ssm-offledger-dequant` with `-x` provenance. One re-authoring was needed: the alloc test set `per_token_group_quant_fp8_k` to an `ops::Fp8ActQuant { shared, hopper }` twin selector, which is the Hopper FP8 activation-quantizer work and is not in this stack. On this base the field is a plain `KernelHandle`, so the test assigns the handle directly. Nothing else in either commit needed a resolution.

## CLA


TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1034 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1034; this fork PR is closed.

PR #1010, opened 2026-09-12
spark-server: stop injecting a second tool prompt when the checkpoint template owns one

TheTom:
## Summary

When the loaded checkpoint's **own** chat template renders tool instructions — every native Qwen 3.5/3.6 template does — Atlas injected a **second** copy from the tool parser. The request then carried two competing tool-format guidances, and the observed failure mode is the model emitting **literal `<tool_call><bash><command>` XML as content**, HTML-entity-escaped, which no parser recognises.

This resolves the ownership question once, in `prepare_chat_prompt`: if the template came from the checkpoint, the model type is a known native-Qwen one, and the template actually declares a `tools` variable, then the parser contributes **only** the tool-choice constraint (`required` / a named tool), not its schema block and not its behavioural Bash/Write/Edit guidance. Every other path — custom override template, ChatML fallback, Hermes, TSCG — keeps the parser prompt exactly as before.

**Base is `main`.** No stack. `crates/spark-server` only; no kernel, no model, no runtime change.

## What was wrong

Two comment blocks in the tree recorded the two halves of the oscillation, and neither fixed it:

- `api/chat/mod.rs` carried a 2026-05-25 note removing the parser prompt outright, because injecting it made the model emit free-form XML.
- `api/chat/prepare.rs` carried an ST-995 note restoring it, because **Hermes** needs the canonical NousResearch function-calling prompt to decline correctly on irrelevance prompts (hallucination ~96 with it, 30/64 without).

Both are right. They are about **different parsers**. The code had one global switch for a per-checkpoint question, so it kept getting flipped.

The concrete duplication: a native Qwen template renders the tool schema and the `<tool_call>` XML format from its `tools` argument. `qwen3_coder` / `qwen3_xml`'s `system_prompt` renders the same schema again, plus tool-agnostic behavioural text. The model receives two formats and picks a third.

### The predicate is narrow, and each clause is there for a reason

`checkpoint_owns_qwen_tool_prompt(model_type, checkpoint, openai_override)` in `tokenizer/chat_impl.rs`:

| clause | why |
|---|---|
| `checkpoint` | the template came from the model dir's `config`/`tokenizer_config`, not from an override. `load_override_template` and the ChatML fallback both set this false — an operator's own template is not assumed to render tools. |
| `!openai_override` | an OpenAI-shaped jinja env is present, so the request is not going through the native path at all. |
| `model_type ∈ {qwen3_5, qwen3_5_moe, qwen3_6, qwen3_6_moe}` | an allow-list, not a heuristic. A new checkpoint family gets the old behaviour until someone adds it deliberately. |

and then, independently, the loaded template must actually declare `tools`:

```rust
&& jinja_env.get_template("chat")?.undeclared_variables(false).contains("tools")
```

A Qwen template that does *not* take `tools` keeps the parser prompt. So the suppression is a property of the template in hand, not of a name.

### What is still injected

`parser_tool_prompt` keeps `append_tool_choice_instruction` on the native path. `tool_choice: required` and named-tool choices are a **request-level constraint** the checkpoint template knows nothing about; dropping them would change behaviour, not just prose. `levers.tscg` also opts out of suppression entirely.

### Incidental

`normalize_tool_name` moves from `tool_parser.rs` to `tool_parser/helpers_a.rs` unchanged — `tool_parser.rs` was at the CI 500-LoC cap and the new code has to live somewhere. `api/chat/mod.rs` loses the two stale comment blocks and gains one line saying where the decision now lives.

## Oracle and evidence

`crates/spark-server/src/tokenizer/tests/native_tool_prompt.rs` (207 lines, new), driven from `prepare.rs` via `#[path]` so it reaches the crate-private seam.

The fixture is the thing that makes this a real test: `crates/spark-server/test_data/chat_templates/qwen3.6-35b-a3b-fp8.jinja` is **the actual Qwen3.6 checkpoint template**, not a stub. The predicate's `undeclared_variables` clause is asserted against a template that really does render `tools`, so a change to how minijinja reports undeclared variables fails here rather than in production.

Covered: native Qwen + checkpoint template → schema suppressed, tool-choice retained; override template → parser prompt intact; ChatML fallback → intact; Hermes → intact (this is the ST-995 regression the 2026-05-25 removal caused); TSCG lever → intact; a Qwen template without `tools` → intact.

**No serving receipt is claimed.** The failure this fixes is a prompt-content defect observed in a live opencode session, not a number. The H100 box for this campaign is down and nothing here needed it.

## Test plan

- [x] `cargo clippy -p spark-server --features cuda --all-targets --jobs 8 -- -D warnings` — **clean, rc=0**
- [x] `cargo test -p spark-server --bins --features cuda -- native_tool_prompt` — **5 passed, 0 failed** (2467 filtered)
- [x] `cargo test -p spark-server --bins --features cuda` — **2460 passed, 0 failed, 12 ignored** (`main`: 2455 → **+5**)
- [x] `cargo test -p spark-server --features cuda --lib` — **114 passed, 0 failed** (the thin library)
- [x] File-size cap — **0 violations** (`tool_parser.rs` comes back under the cap; that is why `normalize_tool_name` moved)
- [x] No `kernels/` change.

## GB10 perf-gate records

Touches `crates/` (a PERF_PATH), so committed `.benchmarks/` records are invalidated on landing. No perf path is altered; the re-measure is a formality. **But**: agentic/BFCL-style records taken against a native Qwen checkpoint were measured with the duplicated prompt in the context, so their prompt token counts and their tool-call accuracy are both taken under the old behaviour.

## What is NOT claimed

- **No perf claim.** The prompt gets shorter on the native path, which will move prompt token counts; nothing here measures that.
- **No accuracy claim.** The literal-XML failure is the motivation, and the fix removes its stated cause, but this PR presents no BFCL or agentic score.
- The allow-list is four model types. Other checkpoints whose templates render tools keep the duplication until they are added.

## CLA

### DIFF
--- a/crates/spark-server/src/api/chat/mod.rs
+++ b/crates/spark-server/src/api/chat/mod.rs
@@ -228,25 +228,9 @@ pub(crate) async fn chat_completions_inner(
         return ChatOutcome::Http(resp);
     }
 
-    // Tool-parser behavioral system prompt REMOVED again (2026-05-25 PM).
-    //
-    // Re-injecting the qwen3_coder `system_prompt` (with its
-    // `<parameter=content>[package]\nname = "x"</parameter>` example
-    // and `For 'Write'/'Edit' tools specifically: ...` guidance) was a
-    // mid-day attempt to give the model better multi-line content
-    // hints. Live opencode v39 session showed the opposite effect:
-    // the model emitted LITERAL `<tool_call><bash><command>` XML as
-    // CONTENT (with HTML-entity escaping like `&amp;`) because TWO
-    // tool-format guidances were competing — the chat template's
-    // `tools` argument AND my injected prompt — combined with PR 73's
-    // `qwen3_xml` parser. The model got confused which format to use
-    // and emitted free-form XML that the parser couldn't recognise.
-    //
-    // Per user's recall: the "MUCH better" state had `thinking_in_tools=true`
-    // and the chat template alone (no injection). Reverting matches
-    // that state. PR 73's qwen3_xml + native FP8 SSM + streaming
-    // byte-exact + gate-BF16 + thinking_in_tools=true is the live
-    // combination.
+    // Parser/native-template ownership is resolved by prepare_chat_prompt.
+    // Native Qwen templates provide their own instructions; fallback,
+    // custom-template, Hermes, and TSCG paths retain parser contributions.
 
     // M2 per-request LoRA routing: resolve the optional `adapter` name to a
     // pool slot ONCE here (both dispatch paths inherit it). Unset defers to the
--- a/crates/spark-server/src/api/chat/prepare.rs
+++ b/crates/spark-server/src/api/chat/prepare.rs
@@ -293,3 +293,7 @@ mod tests {
         assert_eq!(messages[0].text(), "tool instructions\n\noriginal");
     }
 }
+
+#[cfg(test)]
+#[path = "../../tokenizer/tests/native_tool_prompt.rs"]
+mod native_tool_prompt;
--- a/crates/spark-server/src/api/chat/prepare.rs
+++ b/crates/spark-server/src/api/chat/prepare.rs
@@ -54,16 +54,20 @@ pub(crate) fn prepare_chat_prompt(
         && !req.tools.is_empty()
         && !req.tool_choice.as_ref().is_some_and(|tc| tc.is_none());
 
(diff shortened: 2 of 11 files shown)


TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1043 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1043; this fork PR is closed.

PR #1011, opened 2026-09-12
xgrammar: bound parallel cold mask preparation, and keep the serial default (#918)

### LINKED ISSUE #918: Cold grammar compile for a new tool schema is still ~3.4 s on H100 after the #910 fixes (per-state vocab clone removed, parallel prewarm reverted)

TheTom:
## Summary

`CompiledGrammar::compile_top_k_masks` prepared its selected masks **one at a time, on the calling thread**. Cold tool-call latency is dominated by that walk, and there was no way to measure what a wider walk would buy — the shape was not a parameter.

This makes the walk a **bounded** parallel one: `prewarm::for_each_bounded`, with the worker count carried on the compiler (`prewarm_max_threads`) and inherited by every grammar it compiles. It ships an `xgrammar` example that drives the knob so the trade can be measured on a real vocabulary.

**The default is unchanged — and that is deliberate, not an oversight.** `GrammarEngine` sets one thread. Four-worker preparation did **not** improve cold latency on the measured workload: the cold cost is the compile and the top-k mask build, not the walk over pending masks, and the workers contend on the same rule cache. The capability stays on the compiler because it is the caller's choice and because the profiling example needs it to re-run the experiment on a different vocabulary.

Refs #918.

### Stack

```
main
 └─ #999  pr/918-grammar-mask-cache   (grammar mask cache, Fixes #918)
     └─ THIS PR  pr/xgrammar-cold-mask-prewarm
```

**Base branch is `main`, so the GitHub diff includes #999.** This PR's three commits are at the tip. It stacks because both touch `compiler/mod.rs`, `compiler/compiler.rs`, `compiled_grammar.rs` and `grammar/engine.rs`, and because #999's mask cache is what makes "already warm" a state worth testing for.

## What was wrong

`compile_top_k_masks` selected its states and then walked them serially. Two consequences:

1. **No knob, so no measurement.** "Would parallel mask preparation help the 3.4 s cold tool-call?" could not be answered without writing the dispatch first. The honest way to answer it is to build the bounded dispatch, measure, and then *report the result* — including when the result is "no".
2. **Nothing asserted that a parallel walk is order-independent.** Mask preparation mutates a shared cache keyed by FSM state. That it is safe to do out of order is a real property and it was untested, because there was no parallel path.

### `compiler/prewarm.rs` — `for_each_bounded`

A scoped work-stealing walk over a slice, bounded by `max_threads`:

- `workers = max_threads.min(items.len())`, and `workers <= 1` takes the plain serial `for_each` — no thread, no atomic, no scope, for the case that is now the default.
- Jobs are claimed with one `AtomicUsize::fetch_add`, so no item runs twice.
- **`spawn_scoped` failure is not an error.** Under host thread pressure it breaks out of the spawn loop and the calling thread drains the queue itself. Fewer workers, same work, no job dropped — a prewarm that fails because the host is busy would be a worse outcome than a slow one.
- `thread::scope` returns only when every worker has finished, so **all selected masks are complete before the caller can sample**.

### Wiring

`Compiler::new` takes `max_threads`; `compile.rs` stamps it onto each `CompiledGrammar` as `prewarm_max_threads`; `compiled_grammar.rs` hands it to `for_each_bounded`. One value, set at compiler construction, inherited — not a second global.

### The default goes back to serial

`GrammarEngine::from_tokenizer_info` sets one compilation thread, with the comment recording *why*: the measurement did not support four. That is the third commit, and it is the point of the PR as much as the dispatch is.

## Oracle and evidence

`crates/xgrammar/src/compiler/tests/tier2_tests.rs`:

- **`parallel_prewarm_preserves_every_selected_mask`** — the correctness pin. It compiles **one** optimized grammar (so FSM state numbering is fixed) twice, with 1 and 4 workers, **both with fresh caches**, then asserts both the selected count *and* the full `mask_cache` contents are equal. Compiling twice from fresh is what stops the equality passing by the parallel run reusing the serial run's masks.
- **`already_warm_masks_schedule_no_worker_jobs`** — `uncached_masks` must be empty on a warm grammar, so a repeat request spawns nothing. This is the interaction with #999's cache.

`crates/xgrammar/src/compiler/prewarm.rs` tests: jobs overlap when workers are available (a condvar rendezvous, not a sleep), every job runs exactly once, and the serial path is taken at `workers <= 1`.

`crates/xgrammar/examples/native_prewarm_profile.rs` is the measurement harness that produced the "no improvement" verdict, kept in-tree so the verdict can be re-tested rather than believed.

**No new latency number is claimed here.** The #918 cold/warm tool-call timings belong to #999's mask cache, which is the change that actually moved them. This PR's own measured result is negative and is reflected in the default.

## Test plan

- [x] `cargo clippy -p xgrammar -p spark-server --features cuda --all-targets --jobs 8 -- -D warnings` — **clean, rc=0**
- [x] `cargo test -p xgrammar` — **834 + 2 + 1 passed, 0 failed** (#999 base: 830 + 2 + 1 → **+4**)
- [x] `cargo test -p spark-server --bins --features cuda -- prewarm` — **4 passed, 0 failed, 1 ignored** (2471 filtered). The ignored one is `native_qwen_prewarm_exactness_and_cpu_timing`, a CPU diagnostic gated on `QWEN_TOKENIZER_JSON` pointing at a pinned tokenizer — it is `#[ignore]` on purpose and adds no pass to the count.
- [x] `cargo test -p spark-server --bins --features cuda` — **2461 passed, 0 failed, 13 ignored** (#999 base: 2461 passed / 12 ignored → **+0 passing, +1 ignored diagnostic**)
- [x] File-size cap — **0 violations**
- [x] No `kernels/` change, no `crates/spark-model` change.

## GB10 perf-gate records

Touches `crates/` (a PERF_PATH), so committed `.benchmarks/` records are invalidated on landing. The default path is **unchanged** (one worker → the serial `for_each`), so this is a formality rather than a re-measure with a new shape.

## What is NOT claimed

- **Parallel prewarm is not claimed to be faster.** It was measured and it was not, on the workload that motivated it. The default is serial. The dispatch ships because the knob is what makes the question answerable again on a different vocabulary, and it ships with its own correctness pin.
- No claim about a vocabulary other than the measured one.
- `for_each_bounded` bounds worker *count*, not wall time. It has no deadline and no cancellation.

## CLA

### DIFF
--- a/crates/spark-server/src/grammar/engine.rs
+++ b/crates/spark-server/src/grammar/engine.rs
@@ -21,6 +21,11 @@ use super::extract_ordered_vocab;
 pub struct GrammarEngine {
     pub(super) compiler: GrammarCompiler,
     vocab_size: usize,
+    /// #918: the on-disk token-mask snapshot this engine loads at
+    /// startup and re-writes as the cross-grammar cache grows. `None`
+    /// until [`Self::attach_mask_cache`] runs, and when persistence is
+    /// switched off. See [`super::mask_cache`].
+    pub(super) snapshot: Option<super::mask_cache::MaskSnapshot>,
 }
 
 // SAFETY: GrammarEngine is initialized on the main thread and moved to the
@@ -152,6 +157,7 @@ impl GrammarEngine {
         Ok(Self {
             compiler,
             vocab_size,
+            snapshot: None,
         })
     }
 
--- a/crates/spark-server/src/grammar/engine.rs
+++ b/crates/spark-server/src/grammar/engine.rs
@@ -144,12 +144,9 @@ impl GrammarEngine {
 
     fn from_tokenizer_info(tokenizer_info: TokenizerInfo) -> Result<Self, GrammarError> {
         let vocab_size = tokenizer_info.vocab_size();
-        // Bound cold mask preparation without occupying the shared sampling
-        // pool. All workers join before prefill; warm masks remain cached.
-        let threads = std::thread::available_parallelism()
-            .map_err(|e| GrammarError::Compilation(format!("grammar CPU count: {e}")))?
-            .get()
-            .min(4) as i32;
+        // Restore the historical serial default: four-worker native-tool
+        // preparation has not improved cold latency in the measured workload.
+        // The compiler still supports bounded parallel preparation for callers.
         // Cache enabled with an explicit memory budget.
         // ★ NOT -1 (unlimited). `CacheKey::Schema` is keyed by the full tool
         // schema, so agentic traffic mints a distinct compiled grammar per
@@ -158,9 +155,8 @@ impl GrammarEngine {
         // above any realistic hot set (tens of schemas) while bounding the
         // tail. `-1` remains available as an explicit opt-in for callers that
         // genuinely want every grammar pinned.
-        let compiler =
-            GrammarCompiler::new(&tokenizer_info, threads, true, GRAMMAR_CACHE_BUDGET_BYTES)
-                .map_err(GrammarError::Compilation)?;
+        let compiler = GrammarCompiler::new(&tokenizer_info, 1, true, GRAMMAR_CACHE_BUDGET_BYTES)
+            .map_err(GrammarError::Compilation)?;
         Ok(Self {
             compiler,
             vocab_size,
--- a/crates/spark-server/src/grammar/engine.rs
+++ b/crates/spark-server/src/grammar/engine.rs
@@ -144,7 +144,13 @@ impl GrammarEngine {
 
     fn from_tokenizer_info(tokenizer_info: TokenizerInfo) -> Result<Self, GrammarError> {
         let vocab_size = tokenizer_info.vocab_size();
-        // Single compilation thread, cache enabled, no memory limit.
+        // Bound cold mask preparation without occupying the shared sampling
+        // pool. All workers join before prefill; warm masks remain cached.
+        let threads = std::thread::available_parallelism()
+            .map_err(|e| GrammarError::Compilation(format!("grammar CPU count: {e}")))?
+            .get()
+            .min(4) as i32;
+        // Cache enabled with an explicit memory budget.
         // ★ NOT -1 (unlimited). `CacheKey::Schema` is keyed by the full tool
         // schema, so agentic traffic mints a distinct compiled grammar per
         // request; unbounded, that grew host RSS ~100 MB/min under BFCL and is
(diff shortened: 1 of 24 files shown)


TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1040 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1040; this fork PR is closed.

PR #1013, opened 2026-09-12
spark-model: paged-decode attention split-K on Hopper — the split count comes from the target, not from GB10's constant (#928)

TheTom:
> **Stacked on #1006** (`pr/target-defaults`), which is itself stacked on #996. This PR's own content is the **top four commits**; review it alone with `git diff d3898e219..`. Merge order: **#996 → #1006 → this**. Base is `main` rather than a stack branch so this PR does not become unreviewable if either is rebased.
>
> Why #1006 is in the base, in one line: it supplies `[hardware] sm_count` → `atlas_kernels::TARGET_SM_COUNT` (hopper 132, gb10 48, b200 148), the boot cross-check against `CU_DEVICE_ATTRIBUTE_MULTIPROCESSOR_COUNT`, and the `[defaults]` table this PR's `attn_decode_splitk` row lives in. Without it there is no target-owned SM count to size a grid from, which is the entire defect.

## Summary

**Split-K was disabled on Hopper by a constant, not by a kernel.** `atlas-core/src/device.rs` declares `NUM_SMS = 48` — the GB10 value, in a module named `sm121` — and `run_paged_decode.rs` imported it. On a 132-SM H100 at `--max-batch-size 16` that computes `24 q heads × 16 pinned sequences = 384 ≥ 48` and picks **one split at every batch size, including C=1**. The split-K kernels existed and were wired; nothing selected them.

This PR gives the dispatch a rule that reads the compiled **target**: `[defaults] attn_decode_splitk = legacy|auto|N|0`, resolved through `atlas_kernels::attn_splitk`, with `num_splits = clamp(ceil(2·sm_count / num_q_heads), 1, 16)` — 11 on Qwen3.8-27B, i.e. 264 CTAs where the shipped geometry launched 24. Hopper declares `auto`; gb10 and b200 declare `legacy`, which is the pre-#928 rule restated **verbatim**.

Measured on 1×H100: **C=1 TPOT 16.79 → 13.66 ms (−18.6%)** and **tok/s +21.4%** on the long shape. **The C=16 half is a null and is not claimed** — see "What is NOT claimed".

Refs #928 #927.

## What was wrong

nsys, 1×H100 80 GB HBM3, `Qwen/Qwen3.8-27B-FP8`, round 13 cell T1N, C=1, 510 steps, ctx ≈ 4847, median step **16.692 ms**:

| kernel | nodes | µs/step | grid | µs/launch | bytes | achieved | % HBM |
|---|---:|---:|---|---:|---:|---:|---:|
| `paged_decode_attn_fp8` | 12 | 2778.2 (16.6%) | **(24,1,1)** | 231.51 | 9.93 MB | 42.9 GB/s | **1.28%** |
| `paged_decode_attn` (bf16 KV) | 4 | 1013.4 (6.1%) | **(24,1,1)** | 253.36 | 19.86 MB | 78.4 GB/s | **2.34%** |

**3.79 ms of a 16.69 ms C=1 step — 22.7% — moving 198.6 MB at ~1–2% of the roofline.** Two independent faults, plus one gap:

```rust
let current_ctas = num_q_heads * split_ref_seqs(num_seqs, max_decode_seqs);
let num_splits = if current_ctas >= NUM_SMS { 1 } else { NUM_SMS / current_ctas };
```

1. `NUM_SMS` was **another card's** — 48, from a module named after GB10.
2. The occupancy the rule sized for was the **pinned max batch**, not the one sequence actually in flight. Even with 132 substituted, `24 × 16 = 384 ≥ 132` still yields one split at C=1.
3. `KvCacheDtype::Bf16` took a separate, explicit `// no Split-K (not implemented for BF16 yet)` branch, so the four `--kv-high-precision-layers auto` layers had nothing to select at any occupancy.

## What the fix does

**`atlas_kernels::attn_splitk` is the SSOT**, and it is a **pure function of configuration**:

```
num_splits = clamp(ceil(SPLITK_TARGET_WAVES × sm_count / num_q_heads), 1, MAX_DECODE_SPLITS)
             SPLITK_TARGET_WAVES = 2, MAX_DECODE_SPLITS = 16
```

**🪤 Purity here is not style — it is the determinism pin.** The online-softmax split-merge is **non-associative**. If `num_splits` moved with the runtime co-batched count, one sequence would traverse a different reduction tree decoded alone than decoded beside fifteen others, and could flip its temp-0 argmax — the exact nondeterminism `split_ref_seqs` was introduced to stop (`tasks/determinism_investigation.md`). `Auto` and `Pinned` therefore read `(sm_count, num_q_heads, max_decode_seqs)` and **nothing else**; the reduction tree is fixed for the life of a serve. `Legacy` keeps its dependence on `split_ref_seqs` because every target but Hopper still runs it and "unchanged" has to mean unchanged.

It lives in `atlas-kernels`, below `spark-model`, because **two crates need the same answer**: the dispatch that picks the split count and the arena that sizes the split-K workspace (`spark-runtime`'s `sizes.rs`). A second spelling of this rule is how a grid comes to index past its allocation, silently, into device memory it does not own.

**The workspace is sized from the same function the grid is derived from**, over `DecodeMetaLayout::rows()` rather than the pinned batch — rows is the widest batch the metadata upload accepts and therefore the real ceiling on `num_seqs`. Under `legacy` it stays the ~48 KB it has always been, because that rule divides the SM count by `q_heads × reference batch` and the product cannot exceed it.

**Short contexts are clamped inside the kernel** (`PD_MIN_KV_PER_SPLIT`), from each sequence's own `seq_len`. The host may not branch on `seq_lens` — it is device memory behind a captured CUDA graph — and a per-sequence floor stays co-batch invariant where a per-batch one would not.

**The three Hopper sources are ADDITIONS**, new stems and new entry names, declared in `kernels/hopper/HARDWARE.toml`'s `[kernels] overrides` and graded on disk by `tests/inherited_overrides.rs`. gb10 already has an FP8 split-K pair, but its inner loop is the scalar remainder path, and a split count that fills 132 SMs would **multiply** that dependency chain rather than hide it — so the FP8 twin restores the non-split kernel's `PD_BC=4` batched loads. The BF16 twin is a capability BF16 KV never had. A same-stem `replaces` would have forked the non-split `paged_decode_attn_fp8` entry beside them — the kernel five targets compile and six KV dtypes route through — to retune two. They are **not** symlinked into `kernels/b200`, which declares `legacy`.

**The twins are probed unconditionally** via `try_kernel` into `Option<KernelHandle>`, not behind the lever: a handle set that depended on the environment could change between a CUDA-graph capture and its replay. Every target whose `common/` tree lacks the sources keeps exactly its previous arm.

## Levers and defaults

| target | `sm_count` | q heads | `attn_decode_splitk` | `num_splits` | CTAs at C=1 |
|---|---:|---:|---|---:|---:|
| hopper | 132 | 24 | **`auto`** | **11** | **264** (2.0 waves) |
| gb10 | 48 | 24 | `legacy` | 1 | 24 |
| b200 | 148 | 24 | `legacy` | 1 | 24 |

`ATLAS_ATTN_DECODE_SPLITK=legacy|auto|N|0` is the A/B. **⚠️ `0` and `off` resolve to `Pinned(1)`** and the boot line reports the **resolved** count, so `=0` prints `attn_decode_splitk=1 (env)`. That is honest — one split *is* no split-K — but it surprised round 15's operator, so `kernels/hopper/HARDWARE.toml` now says it, and the dispatch-side route line names the kernel that actually ran:

```
paged decode attention: paged_decode_attn_splitk_fp8_hopper num_splits=11 sm_count=132 policy=auto (ATLAS_ATTN_DECODE_SPLITK)
```

One flag **per KV dtype**, not one shared: Qwen3.8-27B runs FP8 KV on 44 layers and BF16 on four, both take the split path, and a shared flag would report whichever arm a step reached first.

### Serving ladder — 1×H100, same binary, one variable

Cells A15 (`auto`) vs S0 (`ATLAS_ATTN_DECODE_SPLITK=0`), round 15, `Qwen/Qwen3.8-27B-FP8`:

| rung | S0 (1 split) | **A15 (11, `auto`)** | Δ | rep spread |
|---|---:|---:|---:|---|
| `4096x512` C=1 tok/s | 56.46 | **68.54** | **+21.4%** | 0.02–0.03% |
| `4096x512` C=1 TPOT | 16.79 ms | **13.66 ms** | **−18.6%** | — |
| `1024x256` C=1 tok/s | 68.19 | **71.53** | +4.9% | 0.02–0.04% |
| `1024x256` C=1 TPOT | 14.09 ms | **13.40 ms** | −4.9% | — |
| C=1 TTFT, both shapes | 489.2 / 160.7 | 489.7 / 160.5 | ±0.1% | — |

**The C=1 gains are 500–900× their rung's rep spread. TTFT is untouched, as it must be** — this is a decode kernel, and that is the built-in null check.

The curve, C=1 only:

| splits | `1024x256` tok/s | TPOT | `4096x512` tok/s | TPOT |
|---|---:|---:|---:|---:|
| 1 (S0) | 68.19 | 14.09 | 56.46 | 16.79 |
| 4 (S4) | 71.22 | 13.47 | 65.58 | 14.31 |
| 6 (S6) | 71.67 | 13.38 | 67.20 | 13.95 |
| **11 (`auto`)** | **71.53** | **13.40** | **68.54** | **13.66** |

Monotone in splits on the long shape, saturating by 4–6 on the short. The policy's own choice is the best measured point on the long shape and ties with 6 on the short: **nothing here argues for pinning a smaller count.**

### Geometry receipt (nsys, cell A15N, C=1, 4593/512, 510 steps)

`paged_decode_attn_splitk_fp8_hopper` at **`grid=(24,11,1)`**, block (256,1,1), 12 launches/step, **32.70 µs/launch = 303.7 GB/s** against round 13's `(24,1,1)` **231.51 µs = 42.9 GB/s** — **7.08×** — with the BF16 twin at 35.01 µs (7.24×) and the two reduces at `(24,1,1)`, 6.4 µs each. 24 q heads × 11 splits = 264 CTAs on 132 SMs = the policy's 2.0 waves. **The attention pair went from 22.7% of the step to 4.38%**, and the median step 16.692 → 14.455 ms.

### Coherency and determinism

All **seven** coherency exchanges are md5-identical across split counts **1, 4, 6 and 11**, with the BA twin on and off and varlen on and off, and identical to round 14's D14 and round 13's T1. Determinism is **8/8 × 3** and byte-identical to two prior rounds.

### Microtest — `native_attn_decode_splitk_hopper_microtest`

`num_splits ∈ {0,1,2,4,6}` × `L ∈ {1335, 4847, 16384}` × `n ∈ {1,4,16}`, both KV layouts. **90 arms, all green, round 15.** It answers two different questions with two different instruments:

* **BITS is an equality**, and it is the one that matters: row 0's output must be **byte-identical** decoded alone and decoded beside fifteen other sequences, at the same split count. That is what the policy being a pure function of configuration buys.
* **NUMBERS is a tolerance, because it has to be**: splitting re-brackets a non-associative merge, so two split counts *cannot* agree bit for bit and an equality across them would be a test that can only fail. Each arm is graded against the shipped non-split kernel at `rel_rms ≤ 2e-3` — BF16's own per-element representation error, 2⁻⁹ — with a cosine floor and a **KNOWN_BAD control that drops one head's contribution and must be REFUSED**. Worst cross-split `rel_rms` over 90 arms: **9.417e-5** at `cos ≥ 0.999999996`.

Sequences get their own pages deliberately: sharing one sequence's KV across the batch would let L2 serve fifteen of sixteen rows and inflate every n=16 figure.

## What is NOT claimed

**🪤 The C=16 prediction in the first draft of this attribution was REFUTED, and the refutation is in the file.** It predicted long C=16 aggregate 405.9 → ≈453 tok/s and TPOT 31.31 → ≈27.9. Measured: **401.28 and 31.56 — a null inside a 2% rep spread — and the short C=16 is a real −1.6%** (522.15 → 513.86, 7–9× its rep spread).

The microtest said why before the ladder ran: at n=16 the **non-split kernel is already at 15–31% of HBM** (507–1025 GB/s) against 1.3–2.3% at n=1, so there is nothing for split-K to recover; splits buy **0–7%**, and at L=1335 eleven of them cost more reduce than they save. **The policy cannot back off there by design** — co-batch invariance is what keeps the determinism pin. That is a deliberate trade, and **this lever must not be sold on C=16**.

Also not claimed: **prefill**. `inferspark_prefill_paged_fp8` has the identical `(24,1,1)` grid and costs 9.35 ms of a 4593-token prefill at 12 GB/s, but it has no split-K variant — it is a flash-attention prefill with `grid = (nq, ceil(q_len/BR), 1)`, so splitting its KV range needs a new kernel and a BR-row-wise reduce, not a launch-geometry change. Left for its own lever; the 9.1 ms is unclaimed here.

Also not claimed: **any B200 number** (no silicon), and **any GB10 change**. GB10 declares `legacy` and its `HARDWARE.toml` gains one row and a comment; no `kernels/gb10/**` source is touched, so no GB10 perf-gate record moves and none is re-recorded.

## Test plan

All on Spark 1 (GB10, aarch64), isolated worktree, private `CARGO_TARGET_DIR`, `ATLAS_SKIP_BUILD=1`, `--jobs 8`, at this branch's tip.

- [x] `cargo clippy -p atlas-kernels --all-targets -- -D warnings` — clean
- [x] `cargo clippy -p spark-model --features cuda --all-targets -- -D warnings` — clean
- [x] `cargo clippy -p spark-runtime --features cuda --all-targets -- -D warnings` — clean
- [x] `cargo test -p atlas-kernels --no-default-features` — **110 passed, 0 failed** across 16 test binaries, including all **7 `attn_splitk::tests`**: `the_legacy_rule_leaves_an_h100_at_twenty_four_ctas`, `auto_fills_two_waves_at_the_single_stream_shape`, **`the_auto_split_count_does_not_move_with_the_co_batched_count`** (the determinism pin, over `num_seqs ∈ {1,4,15,16}`), `legacy_is_the_old_rule_including_its_reference_batch`, `the_grammar_parses_every_documented_spelling`, `the_environment_overrides_the_declaration_and_says_so`, `the_workspace_covers_every_slot_the_grid_can_address`
- [x] `cargo test -p spark-model --features cuda --lib` — **904 passed, 0 failed**, 11 ignored (includes the resolver's split-K row and the 4 `splitk_route_tests` cases)
- [x] `cargo test -p spark-runtime --features cuda` — **308 passed, 0 failed**, 4 ignored
- [x] `cargo build -p spark-model --features cuda --examples` — clean (the new microtest builds)
- [x] `python3 scripts/check_kernel_shadows.py` — **OK**, and it resolves all three new sources as Hopper-owned: `paged_decode_bf16_splitk_hopper.cu (adds, own source)`, `paged_decode_fp8_splitk_hopper.cu (adds, own source)`, `paged_decode_splitk_hopper.cuh (adds, own source)`
- [x] **`scripts/hopper_ptx_gate.sh --hw hopper --model all --jobs 8 --strict`** — `self-test @ sm_90a: known_good passed=true  known_bad_post_hopper.cu failed=true`; **1296/1296 kernels compiled for sm_90a, 0 failed, 0 rejected entry functions**, 1106 unique compiles after dedup, `--Werror all-warnings`. This is the load-bearing receipt that the new sources compile for Hopper. **The two new `.cu` files are in the ledger 14 times — both, for all seven models — every row `ptx_ok: true, ptxas_ok: true`, `registers_max` 48–56 and `spill_bytes: 0`.** (`paged_decode_splitk_hopper.cuh` has no standalone row by construction: it is a header the two `.cu` files include, and it is compiled through them.) Per-model: deepseek-v4-flash 194/194, nemotron-3-nano-30b-a3b 180/180, nemotron-super-120b-a12b 180/180, qwen3-next-80b-a3b 180/180, qwen3.6-27b 190/190, qwen3.6-35b-a3b 182/182, qwen3.8-27b 190/190.
- [x] `check_cross_hardware.py --base d3898e219 --head HEAD --merge-base` — **verdict: OK**, 6 changed kernel paths, the three new `.cu`/`.cuh` reaching `hopper` only and no other tree. ⚠️ Run from a copy of the script: `scripts/check_cross_hardware.py` does not exist on `main` or in this stack (it is on the #895 integration branch), so an in-tree invocation exits 2 "cannot see the inputs" by its own fail-closed rule.

- [x] File-size cap: **no `.rs` file crosses 500 LoC that was not already allow-listed and already over on the base.** Three allow-listed files grow (`run_paged_decode.rs` 673 → 716, `init.rs` 675 → 704, `buffers/sizes.rs` 570 → 588); none newly crosses the cap.

Not run: anything on H100 or B200 silicon from this branch (no access from the gate box) — every number above is quoted from the round-13/14/15 campaign logs on the integration branch, not re-measured here.

---

**Four checks are red for reasons this PR cannot fix, and all four are red on #1006 too:** `typos` (it flags `recurent` — #1006's own deliberate fixture in `atlas-kernels/tests/target_defaults.rs`, commit `79bea9115`, which exists to prove a mistyped lever name fails the build), `PR benchmark gate`, `seal status` and `PR categorize (advisory)`.

**⚠️ And `cross-hardware kernel reach (CHKI, advisory)` is RED, and it is not this PR's.** The CI job diffs `main...HEAD`, so it sees **487 changed kernel paths** — every symlink #996 creates when it adds the `kernels/hopper` and `kernels/b200` trees (`b200/common/argmax_bf16.cu -> ../../gb10/common/argmax_bf16.cu`, and 486 more of exactly that shape). **The same check is red on #1006 already.** Scoped to this PR's own diff (`--base d3898e219`) the verdict is **OK**: six changed paths, the three new sources reaching `hopper` and no other tree. The job is advisory.

## Conflicts

Four commits were **re-authored** rather than cherry-picked clean, because #1006 landed `[hardware] sm_count`, `TARGET_SM_COUNT` and the boot cross-check that the integration-branch originals introduced alongside the policy. What is new here is the `attn_splitk` module, the `attn_decode_splitk` row and the arms that read them; the SM-count plumbing is #1006's and is used, not duplicated. Each commit message names the integration-branch sha it came from.

## CLA


TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1047 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1047; this fork PR is closed.

PR #1015, opened 2026-09-12
spark-model: one fused [gate|up] W8A8 GEMM for the dense-FFN decode band, instead of two (#927)

TheTom:
> **Stacked on #1006 (`pr/target-defaults`), #994 and #992.** Base is `main` and the stack is brought in as two explicit merge commits, so this PR does not become unreviewable if any of them is rebased. This PR's own content is **the single commit on top of the second merge**; review it alone with `git diff 31dbfac6c..`.
>
> Merge order: **#996 → #1006 → #991 → #992 → (#984 → #986 → #987 → #988 → #989 →) #994 → this.**
>
> Why each is in the base, in one line:
> * **#1006** supplies the `[defaults]` table this PR's `ffn_gateup_fused` row lives in, and the `[kernels] overrides` declaration `silu_mul_strided.cu` is graded by.
> * **#994** supplies the cuBLASLt W8A8 decode arm this PR widens. It carries **#989** (the k-major VEC128 activation-scale layout, the per-family `ATLAS_CUBLAS_GEMM` scoping), **#987** (the W8A8 prefill machinery #989 sits on), **#984**, **#986** and **#988** as a rebased linear stack — the branch heads are not ancestors, but every commit's content is present (`git log d3898e219..8173627bf`).
> * **#992** — ⚠️ **not in the original plan for this PR, and required.** It supplies `prune_after_load`, the hook that releases the gate/up store tensors the fused weight consumed, and `predicted_residency`, the preflight prediction that has to price the fused pair as a **difference**. Without them the fused `[2·inter, hidden]` buffer is a residency **increase**, not the net zero this PR claims. #992 carries **#991** rebased.

## Summary

At n=16 the dense FFN issues **two** cuBLASLt W8A8 GEMMs per layer at `K=5120, N=17408` — and they are the **slowest memory-bound arm in the step**, at 59.4% of HBM, while the arm next door moves the identical number of bytes in one launch at 71.4%.

nsys `--cuda-graph-trace=node`, 1×H100 80 GB HBM3, `Qwen/Qwen3.8-27B-FP8` @ `3717cb05e`, round 13 cell V, median n=16 decode step **19.887 ms** busy:

| arm | K | N | nodes | µs (µs/node) | GB/s | % HBM |
|---|---:|---:|---:|---:|---:|---:|
| **FFN gate + up** | 5120 | 17408 | **128** | **5 730.5 (44.77)** | **1 991** | **59.4%** |
| FFN `down` | 17408 | 5120 | 64 | 2 384.0 (37.25) | 2 393 | 71.4% |
| SSM `in_proj_qkvz` | 5120 | 16384 | 48 | 1 641.9 (34.21) | 2 453 | 73.2% |

This PR issues **one** call at `N = 34816` on the 5..=16-row decode band, output rows `[gate | up]` read by a new strided SiLU.

Refs #928 #927.

## What was wrong

**128 nodes = 64 layers × 2, reading *different* weights, so fusing saves no traffic.** One gate/up node moves **89.1 MB** — exactly what `down` moves in one launch, 7.52 µs faster. The gap is the second launch and its partial wave:

At `N = 17408` with a `128×128` tile, a node is `136 × ceil(16/128) = 136` tiles over 132 SMs — **one wave plus a 4-tile remainder**, so a second wave's tail runs at ~3% occupancy, **twice per layer**. At `N = 34816` it is 272 tiles: two full waves, that remainder paid **once**.

At an 80% target — `down` is already at 71.4% and the LM head hits **80.2%** in this same trace — the arithmetic is `11.41 GB / (0.80 × 3.35 TB/s) = 4 254 µs`, i.e. **1 476 µs/step = 7.4% of the step**, and rank **1** of the round-13 decode table (ahead of paged-decode split-K at 986 µs and GDN state decode at 945 µs).

## What the fix does

**One GEMM at `N = 2 · intermediate` on the 5..=16-row band.** `dense_ffn_gateup_fused.rs` owns the whole selection rule as a pure function; `dense_ffn.rs` is the one call site. `silu_mul_strided.cu` consumes the fused output in place — SiLU(gate)·up over row-strided operands — so there is no staging copy and no scatter.

**🪤 NUMERICS: this is a BIT claim, not a tolerance, and that is the point.** The fused weight is the two `[17408, 5120]` E4M3 blocks appended along N; its `[272, 40]` FP32 scale grid is the two `[136, 40]` grids appended the same way (`17408 % 128 == 0`, so the seam falls on a block boundary — `ceil` of a sum is not the sum of `ceil`s, which is why the selector requires both extents to be whole 128-blocks). Splitting N gives **independent output columns over the same K with the same scales**, so fused element `(m, j)` is the same dot product, in the same order, as gate `(m, j)` for `j < 17408` and up `(m, j − 17408)` above. The microtest therefore asserts **byte equality**, not a cosine.

**The arm requires the W8A8 arm for BOTH halves.** This arm *is* that arm with a wider N: if the ladder would have put either half on a W8A16 rung, fusing would silently change the **arithmetic** of that half, not just its launch count. `gate_up_w8a8` is a clause of the selector for exactly that reason, and it carries every clause of `w8a8_prefill_selected` transitively.

**RESIDENCY is net zero, and it is pinned by a test.** The loader builds the fused `[34816, 5120]` weight and its `[272, 40]` scale grid once per layer, re-points `gate_proj`/`up_proj` at **views inside it**, and `prune_after_load` releases the two source store tensors. `predicted_residency` prices the pair as a **difference**, so the preflight ring fit is unmoved. The prune predicate is the **same function** the loader selected with — a predicate the loader narrows locally is a store tensor freed out from under a live view.

**SCOPED TO DECODE.** At the prefill widths the same two GEMMs already run at **68.6% of FP8 peak** (M=4576) — compute-bound, where a launch saving buys nothing — so the band stops at 16 rows and the prefill arm is untouched.

## Kernel placement

`silu_mul_strided.cu` is a **Hopper-owned ADDITION**, declared in `kernels/hopper/HARDWARE.toml`'s `[kernels] overrides` and graded on disk by `tests/inherited_overrides.rs`. It is the one entry on that page whose reason is **not instruction selection** — the code is ordinary bf16 elementwise work any of these targets could compile. It is Hopper's because the only thing that dispatches it is `[defaults] ffn_gateup_fused`, which only Hopper declares true. Putting it in `kernels/gb10/common` would compile a kernel two targets can never launch **and** add a new cross-hardware symlink to each of their mirrors, which `scripts/check_cross_hardware.py` rule S1 refuses outright.

The handle is probed **only when the lever is armed** (`try_kernel` behind `gateup_fused`), because an unresolved lookup that nothing declared fails the boot audit **closed** — and no other hardware tree carries this source.

## Levers and defaults

| target | `ffn_gateup_fused` | why |
|---|---|---|
| hopper | **`true`** | the round-13 launch accounting above |
| gb10 | `false` | no receipt, and the strided-SiLU consumer is not in its tree — inert |
| b200 | `false` | no receipt; must not inherit a Hopper recipe by resemblance |

`ATLAS_FFN_GATEUP_FUSED=0|false|off|no` kills the arm; anything else arms it. There is no `ATLAS_NO_…` spelling: the lever is new, so no script predates the grammar. The serve line names it (`ffn_gateup_fused=on`, `off (env)` when an operator moved it).

### Microtest — `native_fp8_ffn_gateup_fused_microtest`

**Byte equality** of both halves and of the SiLU consumer at `M ∈ {5, 8, 16}` over **all 16 padded rows** (`cublas_fp8_proj_prequant` hands cuBLASLt `ceil16(M)` and the phantom rows *are* written, so they are part of the gate, not excluded from it), with **four KNOWN_BAD controls** — one byte, one row, wrong half, nonfinite — so the harness is shown able to fail before it is believed.

### 🔴 What is NOT claimed — read this before citing anything

**There is no serving measurement for this PR. Round 16 is pending.** The only thing measured here is the **round-13 launch accounting**: the 128 nodes, the 44.77 µs/node, the 59.4% vs 71.4% vs 73.2% comparison, and the wave arithmetic that explains them. Everything downstream of that is a **prediction**, and it is labelled as one in `FFN-GATEUP-FUSION-ATTRIBUTION.md`:

| cell | now | **predicted** |
|---|---:|---:|
| n=16 decode step, busy | 19.887 ms | **≈ 18.41 ms** |
| `1024x256` C=16 TPOT | 25.38 ms | **≈ 23.9 ms** |

The `−1.5 ms` step figure is **a prediction, not a result.** It has not been run. The same campaign refuted a prediction of exactly this shape one round earlier (the C=16 split-K prediction in #928), which is why this is stated this plainly.

Also not claimed: any GB10 or B200 number — both declare the row `false` and the kernel is not in their trees, so no perf-gate record moves and none is re-recorded. No `BENCH.toml` is touched.

## Test plan

All on Spark 1 (GB10, aarch64), isolated worktree, private `CARGO_TARGET_DIR`, `ATLAS_SKIP_BUILD=1`, `--jobs 8`, at this branch's tip.

- [x] `cargo clippy -p atlas-kernels --all-targets -- -D warnings` — clean
- [x] `cargo clippy -p spark-model --features cuda --all-targets -- -D warnings` — clean
- [x] `cargo clippy -p spark-runtime --features cuda --all-targets -- -D warnings` — clean
- [x] `cargo test -p atlas-kernels --no-default-features` — **103 passed, 0 failed** across 16 test binaries, including **`inherited_targets::every_inherited_common_mirrors_every_gb10_common_file`** and `inherited_overrides` (the two that grade the mirror symlinks and the `[kernels] overrides` declaration on disk)
- [x] `cargo test -p spark-model --features cuda --lib` — **1027 passed, 0 failed**, 11 ignored; 17 of them name the fused arm (`dense_ffn_gateup_fused_tests` + `target_defaults_gateup_tests`)
- [x] `cargo test -p spark-runtime --features cuda` — **320 passed, 0 failed**, 4 ignored (includes the arena's `ffn_gate_up_fused` sizing cases)
- [x] `cargo build -p spark-model --features cuda --examples` — clean
- [x] **`cargo build -p spark-model --features cuda,gpu-examples --examples`** — clean. Run separately and on its own `CARGO_TARGET_DIR`, because plain `--examples` skips every microtest (they are `required-features = ["cuda", "gpu-examples"]`) — i.e. the default step does **not** compile the new microtest at all.
- [x] `python3 scripts/check_kernel_shadows.py` — **OK** (6 hardware trees scanned). It resolves the new source as Hopper-owned and confirms the mirror: `hopper/common declares 1 override(s) … silu_mul_strided.cu (adds, own source)`, `b200/common inherits every kernel from gb10/common`.
- [x] **`scripts/hopper_ptx_gate.sh --hw hopper --model all --jobs 8 --strict`** — `self-test @ sm_90a: known_good passed=true  known_bad_post_hopper.cu failed=true`; **1296/1296 kernels compiled for sm_90a, 0 failed, 0 rejected entry functions**, `--Werror all-warnings`. The count moves 1282 → 1296 against #996 exactly as the two added Hopper sources predict (7 models × 2). In the ledger: **`silu_mul_strided` 7 rows, `fp8_scale_transpose` 7 rows, every row `ptx_ok: true`, `ptxas_ok: true`, `registers_max` 18 and 14, `spill_bytes: 0`.** Per-model: deepseek-v4-flash 194/194, nemotron-3-nano-30b-a3b 180/180, nemotron-super-120b-a12b 180/180, qwen3-next-80b-a3b 180/180, qwen3.6-27b 190/190, qwen3.6-35b-a3b 182/182, qwen3.8-27b 190/190.
- [x] `check_cross_hardware.py` — **⚠️ read the attribution.** Over the WHOLE branch it returns `verdict: CHKI` on `kernels/gb10/common/rms_norm.cu` and `kernels/gb10/common/KERNEL.toml` reaching strix and strix-hip. **Neither is this PR's.** Over **this PR's own diff** (`--base 31dbfac6c --head HEAD`) the verdict is **OK**, 4 changed kernel paths, with `kernels/hopper/common/silu_mul_strided.cu` reaching `hopper` and no other tree. Over the **base alone** (`d3898e219..31dbfac6c`) the verdict is `CHKI` — i.e. the finding belongs to #994's merged content and travels with #994, which is open on that basis. The CI job is `cross-hardware kernel reach (CHKI, advisory)`.

- [x] File-size cap: **no `.rs` file crosses 500 LoC that was not already allow-listed and already over on the base.** `dense_ffn.rs` (2719 → 2968) and `weight_loader/qwen35_dense.rs` (1543 → 1949) grow; both are allow-listed, and the new machinery went into `dense_ffn_gateup_fused.rs` (275) + its tests (429) rather than into them.

- [x] `RUSTDOCFLAGS="-D warnings" cargo doc -p spark-model --no-deps --features cuda` — clean. (The first push was red here: the module's band doc linked `GATEUP_FUSED_MAX_M`, which reaches that module through a private `use`, so it is in scope for the code and not for rustdoc's resolver. Fixed by the crate-qualified path in `e48a91176`.)

---

**Four checks are red for reasons this PR cannot fix, and three of them are red on #1006 too:**

* **`typos`** — `error: recurent should be recurrent`. That string is #1006's own deliberate fixture in `atlas-kernels/tests/target_defaults.rs` (`#[should_panic(expected = "has no key \`ssm_batched_recurent\`")]`, commit `79bea9115`), which exists to prove a mistyped lever name fails the build. Red on #1006.
* **`PR benchmark gate`**, **`seal status`**, **`PR categorize (advisory)`** — the standing trio on this whole stack; `seal` waits on a codeowner and the next commit voids it.
* **`cross-hardware kernel reach (CHKI, advisory)`** — attributed in the test plan above: it is #994's `gb10/common/rms_norm.cu` plus the 487 symlinks #996 creates, not this PR's diff, which is `verdict: OK` on its own.

## Conflicts

Three, all one question, all resolved in the merge commits and explained there:

* **#1006 vs #988 on the LM-head band.** #1006 replaced #988's `ATLAS_LM_HEAD_BATCHM_MAX` env lever with the target's declaration. #1006's side wins; the environment variable still overrides — that rung moved into the resolver, it was not dropped. The three cases that graded `batchm_max_from_value` directly are removed as **superseded**, not as noise: `the_declared_band_is_what_selects_the_tier` next door asserts the same numerics seam against the declaration, and `ops/target_defaults_tests.rs` owns the clamp/junk/zero grammar.
* **#1006 vs #992 on `preflight.rs`.** Each split the file a different way. **Both splits are kept** — `gpu_backend`, `decode_ring`, `headroom`, `post_load_audit` — and neither inline body survives, because each now lives in the other side's module.
* **Against #1013 (paged-decode split-K), when both land.** That PR also adds a `[kernels] overrides` table to `kernels/hopper/HARDWARE.toml` (its three `paged_decode_*_splitk_hopper` sources) and its own `[defaults]` row. Whichever lands second resolves by **taking the union** — the two tables are disjoint sets of filenames and disjoint lever rows, and neither touches the other's. There is no ordering dependency between #1013 and this PR in either direction.

* The fusion commit itself was **re-authored** onto this base rather than cherry-picked clean: the integration branch had it sitting on a further `fp8_residency` / `dense_ffn` module split that is not in any of these PRs. The commit message names the integration-branch sha.

## CLA


TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1048 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1048; this fork PR is closed.

PR #1016, opened 2026-09-12
kernels/hopper: the Hopper decode kernel set — W8A16 GEMV (−20.9% C=1 TPOT) and the M16 tensor-core tiers (#927/#928)

TheTom:
> **Stacked on #996 → #1006 → #984 → #988.** Base is `main` rather than a stack branch so this PR does not become unreviewable if any of them is rebased; the first commits here are theirs. **Merge order: #996 → #1006 → #984 → #988 → this.** Review this PR alone with `git diff 65eb241bd..`, or the eight commits listed under **Commits** below.
>
> Why each is in the base, one line each: **#996** creates `kernels/hopper/` at all — this PR's six sources land in that tree and its `HARDWARE.toml` is where they are declared; **#1006** supplies the `[defaults]` table, `TargetLevers` and the boot line that the five lever rows here are written into; **#984** owns `w8a16_gemv_batch4`/`_batch16` and the `o_proj` batched arm the M16 tier is inserted *ahead of*; **#988** owns `dense_ffn_batch16_decode.rs` and `qkv_fp8_batch.rs`, the two modules the FFN and QKV M16 rungs are threaded into.

## Summary

Six Hopper-owned kernel sources and the host dispatch that reaches them, out of the H100 campaign branch (#895) and onto `main` as one reviewable unit. Two separate results, measured separately:

**1. The W8A16 M=1 decode GEMV, rewritten for HBM3.** On an H100 that family is **71% of the C=1 decode step** and the GB10 kernels are LSU-bound on a shared-memory E4M3 LUT gather, not bandwidth-bound. Replacing the gather with `cvt.rn.f16x2.e4m3x2` and issuing four independent chunk loads before the first is consumed takes **C=1 TPOT 17.87 → 14.14 ms (−20.9%)** and the nsys step **18.50 → 14.91 ms (−19.4%)**, at **1.66–2.05× per shape** and **2 689 GB/s** on the fused gate+up — and it is **bit-identical**, `unequal=0` on all seven production shapes, not a tolerance.

**2. The M16 tensor-core tiers for 5..32-row decode**, three kernels behind four levers, of which **two are ON for Hopper and two are OFF** — because the same kernel family measured **+5.3% on the serve through attention** and **−5.2% through the dense FFN**, and a lever that is one boolean for both would have to pick one of those. The LM-head M16 GEMM is a further **+4%**.

Everything here is Hopper-only on disk: `kernels/gb10/common` is **net-untouched by this PR**, `kernels/b200/common` gains nothing, and `scripts/check_cross_hardware.py --base --head` reports **OK**.

Refs #927 #928.

## Stacked on

```
main
 └─ #996  pr/hopper-b200-targets        kernels/hopper + kernels/b200 exist
     └─ #1006 pr/target-defaults        [defaults] table, TargetLevers, boot line
         └─ #984 pr/fp8-batch-parity-oproj    w8a16_gemv_batch4/16 + o_proj arm
             └─ #988 pr/decode-batch16-ffn    dense_ffn_batch16_decode, qkv_fp8_batch
                 └─ THIS PR
```

### `w8a16_gemv.cu`, `w8a16_gemv_fused.cu`, `w8a16_gemv_hopper.cuh` — REPLACE

**What.** Hopper-owned copies of the two GB10 sources that hold the W8A16 M=1 decode GEMV family (`w8a16_gemv`, `w8a16_gemv_dual`, `w8a16_gemv_silu_input`), sharing one new header. Same entry points, same launch geometry, same arithmetic in the same order — a different dequant instruction and a deeper load pipeline.

**Why.** nsys, 1×H100 SXM5, `Qwen/Qwen3.8-27B-FP8`, 2026-09-11 round 10, C=1 steady-state step **18.5 ms**, GPU idle 5%: `w8a16_gemv` is 224 launches = **7.39 ms (40%)** and `w8a16_gemv_dual` 64 launches = **5.79 ms (31%)**. Together they move 24.3 GB of FP8 weights per token in 13.2 ms = **1.84 TB/s**, against HBM3's 3.35 TB/s peak. Two causes, both with arithmetic in the header:

* the E4M3 dequant is a **shared-memory gather** — one LDS per weight byte into a 256-float table that spans the 32 banks eight times. SASS: 134 instructions per 16-byte chunk, 16 of them LDS, against ~32 shared lanes/cycle/SM. At 2.6 TB/s that is ~38 lane-slots/cycle against a 32-slot budget — over capacity before any DRAM request is issued;
* **one outstanding weight load per warp**. The grid is `ceil(N/4)` CTAs, so the small-N shapes cannot fill the machine at any occupancy: attention k/v (N=1024) is 256 CTAs, ~2 per SM, 8 KB in flight — and measured **861 GB/s** where gate/up measured 1,979.

`cvt.rn.f16x2.e4m3x2` decodes two bytes per instruction on the math pipe (16 LDS/chunk → 0, 134 instructions → 110), and `HOPPER_GEMV_UNROLL=4` puts 64 KB of weight bytes in flight per SM against the old 32 KB.

**Receipt.** H100 round 12, `native_fp8_gemv_hopper_microtest`, hard `unequal` count against a host model of the GB10 per-lane chain:

| shape | N | K | grid | unequal | gb10 chain | **hopper** | speed |
| ffn `down` | 5120 | 17408 | 1280 | **0** | 59.8 µs / 1491 GB/s | **36.0 µs / 2477 GB/s** | **1.66×** |
| ssm `in_proj_qkvz` | 16384 | 5120 | 4096 | **0** | 56.3 µs / 1490 GB/s | **33.8 µs / 2481 GB/s** | **1.67×** |
| ssm `out_proj` | 5120 | 6144 | 1280 | **0** | 25.6 µs / 1228 GB/s | **12.5 µs / 2520 GB/s** | **2.05×** |
| attn `q` | 12288 | 5120 | 3072 | **0** | 44.5 µs / 1414 GB/s | **26.5 µs / 2373 GB/s** | **1.68×** |
| attn `k`/`v` | 1024 | 5120 | 256 | **0** | 8.8 µs / 596 GB/s | **4.4 µs / 1192 GB/s** | **2.00×** |
| attn `o` | 5120 | 6144 | 1280 | **0** | 25.8 µs / 1218 GB/s | **12.6 µs / 2502 GB/s** | **2.05×** |
| **gate+up dual** | 2×17408 | 5120 | 4352 | **0** | 116.6 µs / 1529 GB/s | **66.3 µs / 2689 GB/s** | **1.76×** |

Served: **C=1 TPOT 17.87 → 14.14 ms (−20.9%)**, short-shape rate 53.06 → **66.08 tok/s**; nsys step span **18.504 → 14.913 ms (−19.4%)**.

**🪤 The microtest over-promises by ~2×.** Summed over one C=1 step it predicts 16.62 → 9.52 ms on the GEMV family alone — a −7.10 ms step — where the measured TPOT moved −3.73 ms and the nsys step −3.59 ms. It measures the kernels in isolation at full clocks with nothing else resident. **The direction and the ranking are right; the magnitude is not**, and the served numbers above are the ones this PR claims.

**Bit-identity, and its one hole.** Same chunk sequence per lane, same per-chunk operand order, same separate FMUL/FADD (`--fmad=false` is tree-wide), same `shfl` butterfly, same single BF16 round. The one divergence is E4M3 `0x7F`/`0xFF`, the format's only NaNs, which the LUT decodes to ±0 and `cvt` to NaN; a block-scaled FP8 checkpoint cannot contain them. Stated rather than hidden.

### `w8a16_gemm_m16.cu` — ADD, behind `ffn_m16_tc` (off) and `attn_m16_tc` (on)

**What.** An m16n8k16 tensor-core GEMM whose M tile **is** 16, for the 5..32-row decode band, plus a strided-output twin for the interleaved `[Q|K|V]` decode buffer. Two launches on contiguous row halves cover 17..32.

**Why.** At those widths the alternatives are both wrong: `w8a16_gemv_batch16` is bit-exact but **FP32-FMA-bound** (~37 ALU ops per weight byte, capping near 350 GB/s), and the tile GEMMs pad M to a 128-row MMA tile, wasting seven eighths of every tile.

**Receipt, and why the lever is split.** H100 round 6 and round 9 measured the *same kernel* moving the serve in opposite directions depending on which projection family it was armed for:

| arm | cell | C=16 aggregate | C=16 TPOT |
|---|---|---|---|
| dense FFN (`ATLAS_FFN_M16_TC`) | round 6, J vs I | 228.02 → 216.27 (**−5.2%**) | 55.62 → 58.80 ms (**+5.7%**) |
| attention (`ATLAS_ATTN_M16_TC`) | round 9, W vs U | 235.47 → 247.85 (**+5.3%**) | 53.42 → 50.01 ms (**−6.4%**) |

Round 6's C=16 deltas are 30–95× the rep-to-rep spread; round 9's are 25× a 0.15% spread. C=1 is a measured null in both (17.87 vs 17.89 ms), which is correct by construction — the tier is restricted to 5..16 rows. So the lever is **two rows, not one**: `attn_m16_tc = true`, `ffn_m16_tc = false`. `ATLAS_M16_TC` is the round-6 umbrella that arms both, kept because that is how the A/B was run.

**🪤 Not bit-identical, and the only arm here that is not.** An m16n8k16 MMA reassociates the K reduction (≤ 2 BF16 ULP) — the same reassociation the pre-#927 tile GEMMs already had at these widths. `dense_ffn_m16_tc_oracle.rs` scales the acceptance floor to the reference row rather than to an absolute, and `native_fp8_ffn_m16_tc_microtest` gates it on device.

### `dense_gemm_m16_bf16.cu` — ADD, behind `lm_head_m16_tc` (on) and `lm_head_batchm_max` (16)

**What.** The BF16 decode head at a 16-row M tile, replacing a batched-GEMV band frozen at 8 rows and a tile GEMM above it.

**Receipt.** H100 round 9 cell Y vs U: C=16 aggregate **235.47 → 245.10 (+4.09%)**, TPOT **53.42 → 50.79 ms (−4.92%)**. Round 12 nsys prices the head at 943.8 µs = 4.31% of the step, one launch, 16 rows.

**🪤 `N_TILE=64` is a measured null at serving scale** — cell Y2 is 245.30 / 50.71 ms against Y's 245.10 / 50.79, a 0.08% difference against a 0.11–0.17% spread, even though the microtest prefers n64 by 2–4% at kernel level. `N_TILE=32` ships; `ATLAS_FFN_M16_TC_NTILE=64` is the knob, and the route line says which ran.

**Why `lm_head_batchm_max = 16` and not GB10's frozen 8.** On H100 it is the *tile GEMM* that loses at decode widths (#927: 224 ms/step at 16 active rows), which is the opposite of the GB10 measurement that froze 8. The resolver clamps the value at the kernel's compile-time row bound, so a larger number cannot launch a kernel that cannot serve it.

### `w8a16_gemv_ncol.cu` — ADD, behind `attn_ncol_gemv` (OFF)

**What.** N-column-blocked W8A16 GEMV (`_ncol2`/`_ncol4`, contiguous and strided) for the decode attention projections.

**Why it ships OFF.** **There is no H100 serving receipt for it.** The microtest exists; the A/B does not. It is here because the kernel is the other half of the 16-row attention phase the M16 work opened up and because shipping it disarmed is what lets the A/B be run at all — not because it has been shown to pay. `attn_ncol_gemv = false` on every target, and the row says exactly this.

### `HARDWARE.toml [kernels] overrides` — the declaration

Maintainer rule, 2026-09-11 (tbraun96): *"symlinks are fine provided the pointed-to gb10 file is not edited when iterating on Hopper; Hopper-tuned kernels must be real files under `kernels/hopper/`."*

The rule needs a **declaration**, because an undeclared regular file in a mirror is a silent fork of a shared kernel — the defect the mirror check exists to catch — and nothing on disk tells the two apart. One list per target, in its own `HARDWARE.toml`, read by the Rust mirror check and by `scripts/check_kernel_shadows.py`, rather than a Rust constant the non-Rust consumers cannot see. Two shapes, checked differently:

* a **REPLACE** has a gb10 namesake: same entry points, this hardware's instruction selection, gb10's file untouched — `tests/inherited_overrides.rs` asserts gb10 still holds the shared-memory LUT gather and has gained no Hopper instruction;
* an **ADD** has a stem gb10 does not have: it must declare an entry point, must not re-declare one gb10 already has, and must not be a gb10 source's bytes under a new name.

`kernels/b200` declares **nothing**: it has no A/B for any of these levers, so linking the three ADDs in would compile kernels nothing on that target can launch — and would add cross-hardware symlinks that `check_cross_hardware.py` rule S1 refuses outright.

## Levers and defaults

| row | hopper | gb10 | b200 | env override | receipt |
|---|---|---|---|---|---|
| `ffn_m16_tc` | **false** | false | false | `ATLAS_FFN_M16_TC` | round 6 J vs I: **−5.2%** C=16 aggregate |
| `attn_m16_tc` | **true** | false | false | `ATLAS_ATTN_M16_TC` | round 9 W vs U: **+5.3%** C=16 aggregate |
| `lm_head_m16_tc` | **true** | false | false | `ATLAS_LM_HEAD_M16_TC` | round 9 Y vs U: **+4.1%** C=16 aggregate |
| `attn_ncol_gemv` | **false** | false | false | `ATLAS_ATTN_NCOL_GEMV` | **none** — microtest only |
| `lm_head_batchm_max` | **16** | 8 | 8 | `ATLAS_LM_HEAD_BATCHM_MAX` | round 9 Y; clamped at the kernel bound |

One commit per row, each carrying its own `HARDWARE.toml` entry, `TargetLevers` field, `build_defaults` arm, resolver read, boot-line token and test — the contract #1006 establishes. The boot line prints the resolved value and its source, so an operator can tell a target default from an `(env)` override without reading the recipe.

## Commits

| | |
|---|---|
| `kernels/hopper: W8A16 decode GEMV tuned for HBM3 memory-level parallelism (#928)` | the three REPLACEs, the host-model oracle, the GPU-free loop-order simulation |
| `kernels/hopper: the M16 tensor-core decode tiers for 5..32 rows (#927)` | the three ADDs and every dispatch site, all levers default OFF |
| `spark-model: the dense-FFN M16 tensor-core arm becomes a [defaults] row (#927)` | `ffn_m16_tc` |
| `spark-model: the decode attention M16 tensor-core tiers become a [defaults] row (#927)` | `attn_m16_tc` |
| `spark-model: the BF16 decode head's tensor-core arm becomes a [defaults] row (#927)` | `lm_head_m16_tc` |
| `spark-model: the N-column decode attention GEMV becomes a [defaults] row (#927)` | `attn_ncol_gemv` |
| `spark-model: hopper's BF16 head band widens to 16, beside the arm it was measured with (#927)` | `lm_head_batchm_max` |
| `docs: the Hopper decode kernels and the five rows they declare` | CHANGELOG |

The second commit lands every kernel with every lever OFF, so it changes no
serve on its own; the five that follow each turn one row on or off with its own
receipt in the message, which is the `[defaults]` contract #1006
sets out (a row belongs in the commit that lands its consumer).

## Tests

**GPU-free, and they are most of the argument.**

- `ops/w8a16_gemv_hopper_tests.rs` — a HOST transcription of the override's
  UNROLL-wide prefetch body and of the GB10 loop, asserted to emit the same
  chunk sequence and the same per-lane FP32 accumulator at K values covering
  every residue. The same loop with the unroll group consumed last-to-first is
  required to DIFFER, so the assertion has teeth.
- `dense_ffn_m16_tc_{tests,lever_tests,m32_tests,oracle}.rs` — the band, the
  two independent family levers, the two-halves M=32 offset (`m=17 → 9 + 8`,
  wider half first) and the acceptance floor scaled to the reference row.
- `ops/dense_gemm_m16_bf16_{tests,floor_tests}.rs`, `lm_head_bf16_tests.rs` —
  the head's route, its CTA geometry at both tiles, the `N_TILE=64` fallback
  when the wide entry point is absent, and the route line's TEXT (round 9 found
  the message claiming a bare "≤ 2 BF16 ULP" bound that the measurement
  refuted; a test now forbids that wording verbatim).
- `qwen3_attention/attn_{m16_tc_route,ncol_gemv}_tests.rs`,
  `o_proj_tests.rs`, `qkv_fp8_batch_tests.rs` — the attention dispatch sites,
  graded at the LAUNCH rather than only on the rule, so a route that resolves
  correctly and then launches the wrong geometry still fails.
- `ops/target_defaults_tests.rs`, `atlas-kernels/tests/target_defaults.rs` —
  each new row: Hopper's declaration, GB10's and B200's, the environment
  override in both directions, the `(env)` tag, the serve line's text, and that
  `ATLAS_NO_ATTN_DECODE_BATCH` still outranks the row beneath it.
- `atlas-kernels/tests/inherited_overrides.rs` (from #1006) now
  has six declarations to grade rather than none: it asserts each REPLACE left
  GB10's source byte-untouched and each ADD declares an entry point GB10 does
  not.

**On device** (the four microtests, `--features cuda,gpu-examples`, not run by
CI): `native_fp8_gemv_hopper_microtest`, `native_fp8_ffn_m16_tc_microtest`,
`native_bf16_lm_head_m16_microtest`, `native_fp8_attn_decode_batch_microtest`.
The first is the `unequal=0` oracle quoted above.

## Test plan

All on **Spark 1 (GB10, aarch64)**, CPU-only, `ATLAS_SKIP_BUILD=1`, `--jobs 8`,
in an isolated worktree with its own `CARGO_TARGET_DIR`, against this branch
fetched from the fork. No GPU was used; no microtest was run here.

- [x] `cargo clippy -p atlas-kernels --all-targets -- -D warnings` — clean, 0 warnings
- [x] `cargo clippy -p spark-model --features cuda --all-targets -- -D warnings` — clean, 0 warnings
- [x] `cargo test -p atlas-kernels` — **103 passed, 0 failed, 7 ignored** across 15 test binaries. The ones this PR moves:
  - `tests/target_defaults.rs` — **14 passed** (`hopper_declares_what_an_h100_serve_runs_with`, `gb10_declares_the_baseline_and_nothing_else`, `b200_declares_the_conservative_table_not_hoppers`, `every_declaring_target_states_every_lever`, `the_generated_constant_names_every_field`, …)
  - `tests/inherited_targets.rs` — **16 passed**
  - `tests/inherited_overrides.rs` — **3 passed** (the REPLACE rule, the ADD rule, and the shared-override rule — this PR is the first content those three grade)
  - `tests/kernel_shadow_detector.rs` — **5 passed**
- [x] `cargo test -p spark-model --features cuda --lib` — **1035 passed, 0 failed, 11 ignored**
- [x] `cargo build -p spark-model --features cuda,gpu-examples --examples` — clean (all four new microtests compile)
- [x] `python3 scripts/check_kernel_shadows.py` — `OK (6 hardware trees scanned)`, and it now reports the declaration:
  ```
  b200/common inherits every kernel from gb10/common
  hopper/common declares 6 override(s) of gb10/common — `replaces` = gb10 has the same
  name and keeps its own copy, `adds` = gb10 does not have it at all:
      dense_gemm_m16_bf16.cu (adds, own source)
      w8a16_gemm_m16.cu      (adds, own source)
      w8a16_gemv.cu          (replaces, own source)
      w8a16_gemv_fused.cu    (replaces, own source)
      w8a16_gemv_hopper.cuh  (adds, own source)
      w8a16_gemv_ncol.cu     (adds, own source)
  ```
- [x] `python3 scripts/check_cross_hardware.py --base fork/pr/target-defaults --head HEAD` — **`verdict: OK`**, 10 changed kernel paths, every `kernels/hopper/**` row reaching 0 other targets. (Run with `main`'s copy of the script: it landed in #1001, after #996 was cut, so it is not on this branch.)
  ```
  kernels/hopper/common/dense_gemm_m16_bf16.cu    hopper   -   0  OK
  kernels/hopper/common/w8a16_gemm_m16.cu         hopper   -   0  OK
  kernels/hopper/common/w8a16_gemv.cu             hopper   -   0  OK
  kernels/hopper/common/w8a16_gemv_fused.cu       hopper   -   0  OK
  kernels/hopper/common/w8a16_gemv_hopper.cuh     hopper   -   0  OK
  kernels/hopper/common/w8a16_gemv_ncol.cu        hopper   -   0  OK
  verdict: OK
  ```
- [x] `scripts/hopper_ptx_gate.sh --hw hopper --model qwen3.8-27b --jobs 8 --strict`
  ```
  self-test @ sm_90a: known_good passed=true  known_bad_post_hopper.cu failed=true
  191 kernel(s) across 1 model(s); 191 unique compile(s) after dedup;
    hardware flags: -DATLAS_NO_WARP_BLOCKSCALE_MMA
  compiling for sm_90a with 8 job(s), strict=1 — Build cuda_13.0.r13.0/compiler.36424714_0
  191/191 kernels compiled for sm_90a; 0 failed (0 rejected entry function(s))
  ```
  The self-test runs first and both halves report correctly, so the 191/191 is
  evidence rather than a checker that cannot fail.
- [x] CI file-size cap (≤500 LoC per `crates/**.rs`, allow-list honoured) — **0 violations**. Every file this PR adds is under the cap; the four allow-listed files it grows (`dense_ffn.rs`, `qwen3_attention/init.rs`, `model/impl_a1.rs`, `model/types.rs`) were already over it and already listed.
- [ ] **Not tested against real hardware in this PR.** Every number quoted above
      was measured on the H100 campaign branch (#895) at the commits named beside
      it. The four microtests are the way to re-run them on a box with a GPU.

* **Nothing here is measured on GB10.** Every number is 1×H100 80 GB HBM3, `Qwen/Qwen3.8-27B-FP8`. Four of the five rows are `false`/`8` on gb10 and b200 for exactly that reason, and the two Hopper `true` rows change no other target's serve.
* **The GEMV microtest's 2.3× headline is not the served number.** −20.9% TPOT is.
* **`attn_ncol_gemv` has no serving receipt at all** and ships off. Its kernel is compiled on Hopper and reached by nothing.
* **The M16 tiers are not bit-identical** (≤ 2 BF16 ULP from MMA reassociation). The GEMV override *is*, except at the two E4M3 NaN encodings a block-scaled checkpoint cannot contain.
* **No B200 claim.** B200 compiles none of the three added sources and declares no overrides.
* **No GB10 perf-gate record is offered.** This diff touches `crates/`, a PERF_PATH, so `record_covers` invalidates the committed `.benchmarks/` records on landing and a GB10 re-measure is owed before merge. On GB10 every lever here resolves `false`/`8` and every kernel lookup returns a zero handle, so the expected GB10 delta is a null — but that is a prediction, not a measurement.

## Deferred from this PR, deliberately

* **`gdn_decode_hopper.cu` and its `gdn_decode_hopper = false` row.** The GDN decode twins are BIT-IDENTICAL to their parent on all 12 microtest legs (`state_diff`/`out_diff` 0) and **measure a LOSS on this hardware**: 0.83× at the leg that decides C=1 (13.62 vs 11.30 µs, contiguous n=1), 0.98–1.01× on the other eleven; nsys +6.8% per C=1 step (912.7 vs 854.2 µs over 48 launches); and the serve A/B put the kill switch **ahead** by +0.41% aggregate and −0.43% TPOT at C=1 against rep spreads of 0.02%/0.04%. It is a lever worth keeping, but it is ~1,100 lines to add a row that is off on the only hardware that has measured it, and this PR is already large. It comes with `GDN-DECODE-ATTRIBUTION.md` in its own PR.
* **`fp8_scale_transpose.cu`.** It is the K-major activation-scale adapter for cuBLASLt block-scaled FP8 and belongs to #989, not to this decode-kernel set; moving it here would mean deleting a gb10 source this PR has no other reason to touch.
* **The split-K decode GEMV.** Added and then removed on the campaign branch — H100 measured `down` N=5120 K=17408 at 58.9 µs staged-scalar vs **61.8 µs split-K**, and 0.67× on attention k/v. It never enters this branch.

## Conflicts and adaptations

Cherry-picked from `hopper/h100-integration-2026-09-10` and re-authored where
the campaign line had churn this branch should not inherit. Disclosed:

1. **The split-K decode GEMV is not here at all.** `cc7d2b7a7` added it and
   `a5ab39177` removed it on the campaign line after the H100 measured it a
   loss. Replaying both would add and delete ~1,400 lines inside one PR, so the
   GEMV commit is the NET state: the kernel never appears, and
   `native_fp8_gemv_hopper_microtest` reads against `host_gemv` (a host model
   of the GB10 per-lane chain) rather than against `w8a16_gemv_splitk` at
   `splits=1`, which is what `a5ab39177` changed it to and the only sound
   choice anyway — this override REPLACES the GB10 source, so an H100 build
   does not contain the kernel it would be comparing to.
2. **The M16 kernels never touch `kernels/gb10`.** On the campaign line they
   landed in `kernels/gb10/common` (`3344c013b`, `1ffb539aa`, `c4b65fdee`) and
   were moved to `kernels/hopper/common` later (`00e5cc817`). Here they are
   written to their final home in the commit that adds them, so this PR's
   `kernels/gb10` diff is empty rather than an add-then-delete.
3. **`kernels/b200` gains nothing**, where the campaign line symlinked the four
   moved sources into it. B200 has no A/B for any of these levers, so the links
   would compile kernels nothing on that target can launch — and
   `scripts/check_cross_hardware.py` rule S1 refuses a new cross-hardware
   symlink outright. `tests/inherited_overrides.rs` asserts B200 declares
   nothing and carries none of them.
4. **`ATLAS_M16_TC` moved from the consumer to the resolver.** On the campaign
   line `resolve_m16_tc_levers` took the umbrella as a fourth argument and
   OR-ed it in. Under the `[defaults]` contract that is wrong: an umbrella that
   is read after a declaration can only widen, never narrow, and reading it at
   the consumer made `ATLAS_FFN_M16_TC=0 ATLAS_M16_TC=1` depend on export
   order. It is now one `resolve_toggle` per row with the umbrella as the
   fallback variable, and the two umbrella tests moved to
   `ops/target_defaults_tests.rs` with it.
5. **Two merge conflicts with #988**, both in the BF16 head and both resolved
   toward #1006: `batchm_max_from_value` (#988's env-only band
   resolution) is superseded by `resolve_batchm_max`, and #988's `run_band`
   test helper and #1006's `run_case_band` were the same helper
   under two names — unified on one, with #988's three band cases rewritten
   against the resolver. No assertion was dropped.
6. **`m16_tc_enabled_from_presence` was deleted** with the env read it existed
   for. The polarity question it answered ("presence, not `== 1`") is now
   `resolve_toggle`'s grammar, shared with every other row.

## CLA

- [ ] I have read and agree to the [Contributor License Agreement](../CLA.md).

Refs #927 #928


TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1049 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1049; this fork PR is closed.

PR #1017, opened 2026-09-12
kernels/hopper: the tensor-core GDN prefill family and the SSM BA-gates twin (#928)

TheTom:
Three Hopper prefill kernel families, each with its own hardware receipt, plus the declarations and levers that select them. Every number below was measured on 1xH100 80GB HBM3 serving Qwen/Qwen3.8-27B-FP8; none is a projection.

## Stack

This is the third PR of a stack and its diff shows the two below it until they land. Review only the commits listed under **This PR**.

1. **#996** `pr/hopper-b200-targets` — the Hopper (sm_90a) and B200 (sm_100a) target trees, device-arch preflight, PTX compile gate.
2. **#1006 (`pr/target-defaults`)** — `[defaults]` in `kernels/<hw>/HARDWARE.toml`, baked by `build.rs` into `atlas_kernels::TARGET_DEFAULTS`, resolved declaration-first / environment-second, and `[kernels] overrides` as the SSOT for which kernels a target owns. **This PR adds two rows and one `overrides` list to machinery that PR delivers.**
3. **This PR** — 15 commits, `1ba16fc7f`..`088339459`.

## 1. The tensor-core GDN chunked-prefill state spine

`gated_delta_rule_chunk_delta_h_tcfuse_x2`. The scalar spine was **26.6% / 32.3%** of a 1193 / 4593-token H100 prefill and issued **zero** MMAs: both per-chunk matmuls were scalar FP32 loops at 2.7% of HBM, 5.6% of FP32 peak and 0.38% of BF16 tensor-core peak, with 48 CTAs on 132 SMs and a 64-deep dependent FMA chain per token. The arm moves the k-reduction into the MMA tree, keeping the f32 accumulator as the recurrent state.

**Receipt — round 13**, three serve cells on ONE binary, one variable apart (A = family off, T2 = spine only, T1 = whole family):

| metric | A | T2 | **T1** | T1 vs A |
|---|---|---|---|---|
| 1193/256 C=1 TTFT | 269.1 ms | 184.7 | **162.4** | **−39.6%** |
| 1193/256 C=16 agg | 429.05 tok/s | 498.71 | **521.19** | **+21.5%** |
| 4593/512 C=1 TTFT | 889.3 ms | 565.0 | **491.5** | **−44.7%** |
| 4593/512 C=16 agg | 308.86 tok/s | 383.52 | **405.91** | **+31.4%** |

Coherency 4/4, determinism **8/8 md5-identical across 3 runs**. Against the control, 5 of 7 coherency outputs are md5-identical and the two that differ differ cosmetically ("we can" → "you can", "Thus" → "Therefore"); T2 is md5-identical to A on all 7, so the divergence belongs entirely to the twins and sits at the bf16 storage floor.

## 2. The two Hopper prefill remnant twins

`gdn_fwd_o_hopper.cu` and `gdn_recompute_wu_hopper.cu` — the scalar remnants of the same pipeline on tensor cores. `fwd_o` re-tiles so all 16 warps compute and runs the triangular term as a masked [64×64]×[64×128] MMA (2× the arithmetic, no dependent chain), at **64 registers, 0 spill**. `wu` is a blocked triangular solve, warp-local, at **114 registers and 0 bytes of stack frame** against the parent's 512.

**Receipt — round 13 nsys**, T2 vs T1, identical recipe, `ATLAS_NO_GDN_PREFILL_TC_REMNANTS` the only difference, T=4593 C=1, 96 launches each:

| kernel | parent | **twin** | speedup |
|---|---|---|---|
| `chunk_fwd_o` | 71 548.4 µs | **16 715.5** | **4.28×** |
| `recompute_wu` | 47 388.6 µs | **29 558.3** | **1.60×** |
| `chunk_delta_h_tcfuse_x2` (shared spine, **internal control**) | 51 775.3 µs | 52 060.3 | 0.99× |
| prefill GPU busy | 531.6 ms | 459.9 ms | **−71.7 ms** |

The shared spine landing within 0.55% is what makes this attribution rather than correlation: the −71.7 ms of prefill busy is accounted for by the twins' −72.4 ms, every non-GDN row is unchanged to under 1%, and it shows up one-for-one in client TTFT (565.0 → 491.5 ms). Round 14's microtest put the same two at **4.33× / 1.70×** on the same binary.

## 3. The Hopper SSM BA-gates twin

`ssm_ba_gates_hopper.cu`. **Bit-identical to its gb10 parent by construction** — `__bfloat162float` is exact, so hoisting A's conversion cannot move a bit — so this is a pure speed claim with no accuracy question. The parent reads each token's `K=5120` activation row **96 times**, once per BA output; the twin keeps one CTA per token and a register tile of 8, so the row is read **12 times**. It is issue-bound, not bandwidth-bound: compulsory traffic is 2.6% of HBM, while the parent's loop bodies alone are 434 µs of a 555 µs launch.

**Receipt — round 15**, cell B0 (`ATLAS_SSM_BA_GATES_HOPPER=0`) vs A15, same binary:

| rung | B0 | **A15** | Δ | predicted |
|---|---|---|---|---|
| `4096x512` C=1 TTFT | 498.2 ms | **489.7** | **−8.5 ms** | −7.9 (−6.0…−9.7) |
| `1024x256` C=1 TTFT | 162.7 ms | **160.5** | **−2.2 ms** | −2.1 |
| C=1 TPOT, both shapes | 13.66 / 13.40 ms | 13.66 / 13.40 | **0.00%** | 0 (guard) |

Both TTFT predictions land inside their stated range and TPOT is identical to the hundredth of a millisecond — the token-count guard keeping decode on the parent, as designed. nsys and the microtest price the kernel at **372.86** and **371.56 µs**, agreeing to 0.35%; 48 launches of that saving is 8.74 ms against the 8.5 ms measured at the client. `out_diff=0` and `max_abs=0.000e0` at all four M, the one-ulp KNOWN_BAD control refused at every one, and A15/B0 are md5-identical on all seven coherency exchanges.

## Levers and defaults

| row | hopper | gb10 | b200 | override |
|---|---|---|---|---|
| `gdn_prefill_tc` | **true** | false (inert) | false (inert) | `ATLAS_GDN_PREFILL_TC` |
| `ssm_ba_gates_hopper` | **true** | false (inert) | false (inert) | `ATLAS_SSM_BA_GATES_HOPPER` |

`gdn_prefill_tc` is a **family** lever: the spine and both remnant twins read the same resolved bit, so `=0` is the whole-family kill switch. `ATLAS_NO_GDN_PREFILL_TC_REMNANTS=1` remains the one-variable A/B that keeps the spine and pins the twins to their parents.

⚠️ **One polarity change.** `ATLAS_GDN_PREFILL_TC` was PRESENCE-gated, so `=0` used to mean ON; it now means OFF. Every A/B recipe for it set `=1` and is unaffected. `the_tensor_core_prefill_spine_reads_zero_as_off_not_as_present` pins that direction on every spelling, because a `=0` that silently re-armed the arm would be an accuracy change nobody typed.

## Kernel placement — one deliberate relocation

`gated_delta_rule_chunk_tc.cu` was developed and validated on GB10 and is arch-neutral (no `__CUDA_ARCH__` arm; one MMA shape, `m16n8k16 f32.bf16.bf16.f32`, which is sm_80+). It nonetheless ships in `kernels/hopper/common`, declared in `[kernels] overrides`, because `scripts/check_cross_hardware.py` rule **S1 refuses to create a new cross-hardware symlink** and no trailer clears a structural finding. That rule is right here: a file placed in `kernels/gb10/common` is mirrored into hopper and b200 automatically, so shipping it there would have made three hardware sets compile a new kernel as a side-effect of the mirror rather than as a decision anyone took.

**What that costs, stated plainly:** GB10 no longer compiles the spine, so its `gdn_prefill_tc` row is inert, and `native_gdn_chunk_prefill_microtest` — the spine's numerics oracle, which ran on GB10 — now SKIPS on images without the kernel instead of failing a lookup. Its GB10 receipts stand as evidence; the gate itself runs on a Hopper image from here. Both are reversible in one commit by a GB10 A/B receipt plus the two mirror links, the same disposition `silu_mul_strided.cu` carries.

The other four files (`gdn_fwd_o_hopper.cu`, `gdn_recompute_wu_hopper.cu`, `gdn_prefill_hopper.cuh`, `ssm_ba_gates_hopper.cu`) are ADDITIONS with new stems: their parents live in multi-entry files (`gated_delta_rule_fla.cu` is 2105 lines and fourteen entry points; `ssm_preprocess.cu` holds two siblings GB10 and four other targets compile), so a same-stem override would fork every other entry in them.

## Tests

- `ssm_gdn_tc_tests.rs`, `ssm_gdn_remnants_tests.rs`, `ssm_gdn_remnants_numerics_tests.rs` — index maps and limb arithmetic against an f64 reference, the lever grammar, and the twins' selection wired to the spine's resolved bit as an IMPLICATION rather than a constant, so flipping a target's row does not require rewriting the test.
- `ssm_ba_gates_hopper_tests.rs` — lane/`kv`/warp-slot/output mapping and every guard string.
- `ssm_gdn_tc_route.rs` — the route and init lines are built from the same constants the probes bind, so a log cannot name a kernel other than the one that ran. Round 14 caught 48 init lines naming the scalar parent while all 14 400 dispatches went to the tensor-core entry.
- Three shared-memory invariants moved from a `#[test]` to `const _: () = assert!(…)`: they are properties of constants, so they can no longer pass by not being executed.
- GPU oracles (`cuda,gpu-examples`): `native_gdn_chunk_prefill_microtest`, `native_gdn_prefill_remnants_microtest`, `native_ssm_ba_gates_hopper_microtest`. Each SKIPS loudly on an image without its kernels.

## Gates — Spark 1 (GB10), `08833945`

- [x] `cargo clippy -p atlas-kernels --all-targets -D warnings` — clean
- [x] `cargo clippy -p spark-model --features cuda --all-targets -D warnings` — clean, **0 warnings emitted**
- [x] `cargo test -p atlas-kernels --no-default-features` — **16 suites, 0 failed**
- [x] `cargo test -p spark-model --features cuda --lib` — **947 passed, 0 failed, 11 ignored**
- [x] `python3 scripts/check_kernel_shadows.py` — `kernel shadow structure: OK (6 hardware trees scanned: b200, gb10, hopper, metal, strix, strix-hip)`
- [x] `python3 scripts/check_cross_hardware.py --base d3898e219 --head HEAD` — **`verdict: OK`**, no `Hardware:`/`CHKI-Verdict:` trailer needed (reach is hopper-only). Run from `main`'s copy of the script: it landed in #1001, after this stack's base.
- [x] `scripts/hopper_ptx_gate.sh --hw hopper --model qwen3.8-27b --jobs 8 --strict` — `192/192 kernels compiled for sm_90a; 0 failed (0 rejected entry function(s))`, CUDA 13.0, self-test `known_good passed=true  known_bad failed=true`
- [x] File-size cap (500 LoC per `.rs` under `crates/`, workflow allow-list applied) — **0 violations**. This PR **removes** nothing from and **adds** nothing to the allow-list: `qwen3_ssm/init.rs` crossed 500 mid-series and was brought back to 467 by moving the three probes into `init_kernels.rs`.

## NOT claimed

- **No GB10 or B200 receipt for any of this.** Every row above is an H100 measurement. GB10 has ~48 SMs — the count the 48-CTA grid nearly fills — so the margin does not transfer by argument, and B200 has no serving receipt of any kind. Both targets declare every row `false`.
- **No perf-gate record is included.** This diff touches `crates/` and `kernels/`, both PERF_PATHS, so a GB10 re-measure is owed before merge. On an un-armed GB10 serve every kernel here is inert — the two levers are false there and the kernels are not in that image at all.
- **The BA-gates twin's worth is ~8 ms, not ~26.** An earlier note priced it against a bandwidth roofline; the kernel is instruction-issue bound, so that target does not apply. Recorded in-tree rather than quietly dropped.
- **Round 16 (the FP8 activation-quant twin) is not here** — see below.

## Deferred to a follow-up

The FP8 activation-quantizer twin (`fp8_act_quant_hopper.cu`, `ops::Fp8ActQuant`) and `FP8-ACT-QUANT-ATTRIBUTION.md` are **not** in this PR, deliberately. Its consumers are the W8A8 arms from #987, #989, #994 and #998 — four separate open PRs, each based on `main` — so including it would have meant merging four unrelated heads into a stack that is already three deep. Its round-16 receipt is also still pending. It will follow as its own PR on top of that chain, with its measurement.

Refs #928 #927

## CLA


TheTom:
Rustdoc CI (`Build mdBook + rustdoc`) was failing because public docs for `dense_gemm_ba_gates_prefill` linked to private `super::ssm_ba_gates_hopper`. Dropped the intra-doc link in favour of a code span. Production code unchanged.

`3e97eb4b9` — `docs: drop the rustdoc link to private ssm_ba_gates_hopper`

`#1015` / `#1013` / `#1016` do not carry this rustdoc site; not cherry-picked forward. Not pushed onto `#895`.

TheTom:
Moved onto `Avarok-Cybersecurity/atlas` as #1050 (same branch name, same commits) so Graphite/GH can stack across one origin. Please review/stack #1050; this fork PR is closed.

Issue #974, opened 2026-09-09
N-gram lookup speculation with long drafts into the wide verify: copy-heavy agentic turns (Qwen3.8-27B first, Flash-Next after the K=4 FFN arm)


rrstesiak:
## Summary

Lookup speculation proposes drafts by matching the last few generated tokens against the context and offering the continuation that followed last time. No draft model, no weights. On generation that repeats its context, the draft is nearly always right and tok/s becomes verification throughput. Agentic sessions are full of that: tool-output echoes, a file or block re-emitted with one edit, restated instructions, tool-call scaffolding. Fresh reasoning does not fire it and falls back to the learned draft head.

Atlas already has both halves. Main ships `--ngram-speculative`, described in `serve_args.rs` as "Enable N-gram speculative decoding: CPU-side pattern matching proposer", and logged at serve start as "N-gram speculative decoding: ENABLED (K=2 verify, CPU proposer)". K=2 caps it at two tokens per step. The wide verify from #844 and #845 (batched GDN verify, K=5..16) is the other half. Joined, Atlas gets long-draft lookup speculation at near-zero cost per step, as a companion to MTP rather than an alternative.

## Evidence that the class is real

| engine, one Spark | workload | tok/s | source |
|---|---|---|---|
| llama.cpp, Q4_K_XL GGUF | "reproduce a file with one change", 94.7% n-gram acceptance | 97.4 | NVIDIA forum, Qwen3.8-Flash-Next thread p.4, 26 Aug |
| same post | free-form prose | 22 | same |
| Atlas MTP-2, NVFP4 | LRU 512, fresh generation | 62.0 | #973 |

Copy-task throughput and fresh-generation throughput are different measurements and stay in different tables. The copy-task class is where lookup speculation belongs; the agentic gate is where it pays.

## Plan

| step | what | needs GPU |
|---|---|---|
| 0 | Offline count over recorded agentic trajectories: run the lookup rule at match lengths 3 to 5 and count copy-eligible tokens per turn. This is the ceiling on the agentic gate and decides whether the item is a lever or a trophy. Run records in `~/.atlas/runs` hold verdicts only, so a gate run with response capture produces the input. | no |
| 1 | Read the `ngram` module and its K=2 hookup in `serve_load.rs`. | no |
[...]

PR #1026, opened 2026-09-12
Qwen3.8-Flash-Next: lookup drafts into the wide MTP verify (#974)

rrstesiak:
Claims the implementation of #974 on Qwen3.8-Flash-Next. Draft until the cells below are measured; the code is complete and wired, the tests are in the tree.

## What it does

The step reads its draft out of the sequence before it runs the MTP head. When the last `ATLAS_LOOKUP_MIN_MATCH` tokens (default 8) already occurred earlier in the prompt or the generation, the tokens that followed that occurrence become the drafts, at the width the head would have drafted (K=3 or K=4 rows), and the head does not run that step. With no match the head drafts as before. The verify, accept, commit and rollback paths are untouched.

Copy-heavy turns are where it pays: tool output echoed back, a file re-emitted with one edit, restated instructions, tool-call scaffolding. The class is real. llama.cpp posted 97.4 tok/s on one Spark on "reproduce a file with one change" at 94.7% acceptance against 22 on prose (NVIDIA forum, Flash-Next thread p.4, 26 Aug). Atlas MTP-2 on this pack stands at 62.0 on fresh generation (#973). Copy-task and fresh-generation numbers stay in separate tables.

## What changed

| file | change |
|---|---|
| `spark-server/src/lookup_drafts.rs` | new. A hash index from every min-match-gram to the positions it starts at, extended incrementally as tokens commit, rebuilt on a new sequence or a rollback. One lookup per step at 88K context, where the existing `ngram.rs` rescans the whole history per step. Most recent hit first, longest backward match wins. Fires only when the continuation fills the requested width, so the step shape stays the one the head was tuned at. |
| `scheduler/lookup_gate.rs` | new. The gate both draft sites call: off under DFlash, under a grammar, under expert parallelism, at one draft (the K=2 lane), and with more than one sequence active. At two drafts, K=3 rows, it is live. |
| `scheduler/verify_mtp_wide.rs` | `finish` asks the gate before the re-propose. A lookup step skips the proposer trim (no drafter rows were written for its drafts) and scores the index on what the verify accepted. |
| `scheduler/mtp_step.rs` | the bootstrap asks the gate before the propose, and arms the gate for the single-sequence regime. |
| `scheduler/levers.rs`, `sched_ctx.rs`, `types.rs` | `lookup_drafts`, `lookup_min_match`, `lookup_max_match`; the index lives on the run's context; `pending_drafts_lookup` on the sequence. |

Defaults ON. `ATLAS_LOOKUP_DRAFTS=0` restores the MTP head on every step, byte-identical to the base.

## Drafter state on a lookup step

The Flash-Next head rewinds `drafted - accepted` rows after each verify. A lookup step drafts nothing through the head, so there is nothing to rewind and `finish` skips the trim. The head's own KV then lacks the positions that step committed, the gap a suspended step already leaves today; the next propose reads the target's highway for the row it drafts from, as it does now. A wrong lookup draft costs what a wrong head draft costs: the rows the verify rejects.

## Not in this PR

The K=4+ small-M FFN arm. K=4 rows fall off the small-M FFN arms onto the prefill MoE path (LRU 14.9 tok/s during the #972 series), so this PR runs at the width #973 runs at. Wider lookup drafts on Flash-Next wait on that arm. The 27B's batched GDN verify (#844, #845) takes them today.

## Measured so far, my box, this branch

MTP-2, three standard cells, three runs each, against the same binary with `ATLAS_LOOKUP_DRAFTS=0`. At match length 4 the index fires on fresh code and prose, accept counts fall (LRU 310 to 303, MinHeap 178 to 169) and every stream moves. At match length 8 all nine runs are byte-identical to the base at base tok/s, so 8 is the default. Richard's bisect on a TP=2 x EP=2 build found the same firing cost at 4, plus prefix-cache anchor drift across ranks; the gate is now off under expert parallelism until that is understood. The copy-task cell and the agentic gate follow.

## Measurement plan, my box, before this leaves draft

1. Fixed-file "reproduce with one change" cell, its own table.
2. MinHeap x3 and Volvo x3, byte-identical streams against `ATLAS_LOOKUP_DRAFTS=0`, medians, completion tokens beside tok/s.
3. `spark benchmark run agentic-webserver` against the serve.

Environment: nvidia/Qwen3.8-Flash-Next-NVFP4 @ fc694b54, GB10, `bench/qwen4_exp/serve_nvidia_mtp.sh`, MTP_DRAFTS=2.

Tests: `cargo test -p spark-server lookup`.

Related: #974, #972, #973, #844, #845, #834.

### DIFF
--- a/crates/spark-server/src/lookup_drafts.rs
+++ b/crates/spark-server/src/lookup_drafts.rs
@@ -144,30 +144,37 @@ impl LookupDrafter {
         self.tail.extend_from_slice(&tokens[len - n.min(len)..]);
     }
 
-    /// Propose exactly `width` tokens to follow `tokens`, or nothing.
+    /// Propose exactly `width` tokens to follow `tokens` + `next`, or nothing.
     ///
-    /// `tokens` is the committed sequence, prompt and generation together.
-    pub fn propose(&mut self, tokens: &[u32], width: usize) -> Vec<u32> {
+    /// `tokens` is the sequence the model has consumed, prompt and generation
+    /// together; `next` is the newest token, emitted but not yet fed (the
+    /// scheduler's `last_token`). The drafts are what followed the last
+    /// occurrence of `tokens[..] ++ [next]`'s tail, so draft 0 is the token
+    /// after `next`, never `next` itself.
+    pub fn propose(&mut self, tokens: &[u32], next: u32, width: usize) -> Vec<u32> {
         let len = tokens.len();
         let n = self.min_match;
-        if width == 0 || len < n + width {
+        if width == 0 || len + 1 < n + width {
             return Vec::new();
         }
         self.extend(tokens);
-        let suffix_start = len - n;
-        let suffix = &tokens[suffix_start..];
-        let Some(starts) = self.index.get(&Self::key(suffix)) else {
+        // The virtual sequence v = tokens ++ [next]; its last n-gram starts
+        // at v index len + 1 - n and ends with `next`.
+        let mut suffix: Vec<u32> = tokens[len + 1 - n..].to_vec();
+        suffix.push(next);
+        let suffix_start = len + 1 - n;
+        let Some(starts) = self.index.get(&Self::key(&suffix)) else {
             return Vec::new();
         };
         let mut best: Option<(usize, usize)> = None; // (match length, start)
         for &p in starts.iter().rev() {
             let p = p as usize;
-            // The hit needs `width` tokens after it that are not the suffix
-            // itself being proposed as its own continuation.
+            // The hit needs `width` tokens after it inside `tokens`, and it
+            // cannot be the suffix itself.
             if p + n + width > len || p == suffix_start {
                 continue;
             }
-            if tokens[p..p + n] != *suffix {
+            if tokens[p..p + n] != suffix[..] {
                 continue;
             }
             // Extend the match backwards; longer context, better draft.
@@ -175,7 +182,7 @@ impl LookupDrafter {
             while m < self.max_match
                 && p >= m + 1 - n
                 && suffix_start >= m + 1 - n
-                && tokens[p + n - 1 - m] == tokens[len - 1 - m]
+                && tokens[p + n - 1 - m] == tokens[len - m]
             {
                 m += 1;
             }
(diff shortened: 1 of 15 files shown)


rsafier:
@rrstesiak — the EP refusal in `lookup_gate` can be lifted. It was closed citing my TP=2 × EP=2 bisect, and that evidence no longer holds once your stale-draft fix (`917ea4859`) is in.

My bisect saw prefix-cache anchor drift with lookup drafts on under EP. It ran **before** you found that the proposer matched `seq.tokens`' suffix while the just-emitted token wasn't yet in that list — so every draft was one position behind and every draft was rejected. The anchor movement looks to have been a symptom of that.

Re-checked on the merged stack (your `c8d8c4a15` + this branch), TP=2 × EP=2, one binary, back-to-back:

| | lookup on under EP | control |
|---|---|---|
| copy task | **61.88** | 55.28 tok/s (**+11.9%**) |
| fresh-text decode C=1 | 53.55 | 53.24 (unchanged) |
| known-answer | 4/4 + 4/4 | 4/4 + 4/4 |
| agentic, 3 iters | **3/3 + 3/3**, 0 errors | — |
| **anchor disagreements** | **0 on both ranks** | 0 |
| **Marconi intermediate hits** | **35 on both ranks** | — |

The Marconi count is the criterion, not the disagreement count alone. My first attempt used the copy task, which showed zero disagreements **and zero Marconi hits** — the warm-restore path that produced the original drift never ran, so it proved nothing. Agentic exercises it, and the two ranks land on the same hit count as well as zero disagreements.

The +11.9% matches your own table's K=4-width rung (26.7 → 29.3, +10%), which is the expected ceiling here because `verify_kn_step` narrows to K=4 under EP — so `ATLAS_LOOKUP_WIDTH=7` is unreachable and the 39.1 row is out of reach for multi-rank serving.

I've shipped this as `ATLAS_LOOKUP_EP` (opt-in in code, set in the qwen4_exp launcher) rather than flipping your `|| ep` condition, since the drift evidence was qwen4_exp-specific and GLM shouldn't inherit a change measured only here. Yours to make the real call on.

**The remaining gap is the EP worker command**, which is your area. The shape, if useful: a width-generic `EP_CMD_VERIFY_KN` (`0xFFFFFFF7` is free) that broadcasts `k` **before** the tokens — so neither rank can guess the receive count — then `k` tokens, then the existing `num_accepted`. The worker reuses your `decode_verify_graphed_kn`, and the K=4 arm's rewind generalises directly to `rewind = (k-1) - num_accepted` (full accept → `start_checkpoint_async`, else pop `rewind` and `start_rollback_and_checkpoint_async(seq, num_accepted+1)`). Two things I'd check rather than assume: worker-side SSM verify pool sizing at 8 rows, and `start_rollback_and_checkpoint_async` at rewind depths past 3.

One other datapoint from this branch, since it touches your Km arm: **DRAFTS=3 is a 26% loss at C=1 / 29% at C=4** on the production config, *with* `hc small-M FFN at 4 rows: Km batched arm km_available=true` confirmed — so it's not the old 512-expert fallback. Ladder is healthy (p1 0.84 / p2 0.78 / p3 0.55); the third draft just costs a full drafter pass every step and lands about half the time it's reached. Your lookup path avoids exactly that cost, which is presumably why it wins where a wider head doesn't.

tbraun96:

tbraun96:

PR #1052, opened 2026-09-12
spark-model: per-arch upper M bound on W8A8 FFN prefill (#917)

TheTom:
> Same-repo restack of #1019 (was `TheTom:pr/w8a8-prefill-max-m`). Commits unchanged. Opened on `Avarok-Cybersecurity/atlas` so Graphite/GH can stack.

> **Stacked on #987, with #1006 merged in — review the ceiling against the merge.** This PR's GitHub base is `main` (GitHub stacks by content; a stack-branch base makes the PR unreviewable if #987/#1006 move). The file list therefore also contains Chain A (#984 → #986 → #987) and #996 → #1006. This PR's own content is the **one commit above `27f4dfaa0`**: `git diff 27f4dfaa0..`. Merge order: **#996 → #1006, Chain A #984 → #986 → #987, then this**.

## Summary

`w8a8_prefill_selected` had no upper bound on M. On H100 that is right — W8A8 is 2.0–3.1× over W8A16 at every M measured. On GB10 / sm_121 it is a regression, and a large one.

This is the follow-up Thomas already wrote and asked us not to bury in #987. The ceiling is a `[defaults]` lever (two rows, widening vs narrowing), so it needs TARGET_DEFAULTS from #1006. Cherry-picked `-x` from `d4112c3eb` (author **Thomas Braun** kept). Conflicts were expected: #1006's table does not yet carry the Hopper kernel rows from the umbrella, so the pick lands the two ceiling rows + `m <= max_m` + the two ceiling tests and does **not** scoop unrelated Hopper kernels.

Refs #917. Credit: Thomas Braun (`tbraun96`).

## What was wrong

W8A8 feeds the FP8 tensor cores directly instead of dequantizing into a BF16 MMA, but it pays a per-token activation quantization and an FP32 scale epilogue that W8A16 never pays. That cost is fixed per token, so W8A8 wins only while the GEMM is small enough that the quantization is not the bill. On sm_121 W8A8 is FLAT at ~14 TFLOP/s from M=128 to M=2048 while W8A16 climbs to ~26 and stays there — a kernel whose throughput does not move with M is not compute-bound.

## Served receipt (Thomas, spark-256a / GB10)

`Qwen3.6-27B-FP8`, M=949, one binary per leg differing only in this ceiling, byte-identical 949-token prompt, 1 warm-up + 5 timed, TTFT median, both arms read back from the serve log:

| arm | path | TTFT p50 |
|---|---|---|
| uncapped | W8A8 | **3343.3 ms** |
| capped | W8A16 | **2560.4 ms** (−23.4%) |

Complete separation: max(capped) 2588.0 < min(uncapped) 3248.0. Reproduced in a second independent run at 2555.8 ms (0.2%).

H100 stays uncapped (`u32::MAX` on both rows): 2.0–3.1× over W8A16 at every M tried (M=64 and M=1193, gate/up and down).

## What this adds

Two `[defaults]` rows, not one: the crossover is shape-dependent. Measured at the real Qwen3.8-27B dims, gate/up (N=17408, K=5120, WIDENING) crosses at M~64–128 and down (N=5120, K=17408, NARROWING) at M~384–512. Each declared value is the **low** end of its band, so the arm is kept only where it is measured to win.

| target | `w8a8_prefill_max_m_widening` | `w8a8_prefill_max_m_narrowing` |
|---|---|---|
| gb10 | **64** | **384** |
| hopper | `u32::MAX` (no cap) | `u32::MAX` (no cap) |
| b200 | `u32::MAX` (unmeasured; not inherited from GB10) | `u32::MAX` |

The selection rule gains `&& m <= max_m`. The bound is inclusive. A parsed `0` is honoured ("never take W8A8 on this shape"). Operators can override with `ATLAS_W8A8_PREFILL_MAX_M_WIDENING` / `_NARROWING`.

`gb10_declares_exactly_the_baseline` is reworked rather than deleted: this is the first value GB10 declares in order to **differ** from the baseline, so the test pins the two numbers and asserts equality on everything else.

## Reachability (do not misread the receipt)

This whole path runs only under `ATLAS_DENSE_FP8=1`. `git grep ATLAS_DENSE_FP8` returns one hit — its own reader. No CI job, no `BENCH.toml` entry and no gate sets it. It is off by default because on GB10 native dense FP8 loses to the NVFP4 fallback outright (default 1437.2 ms vs 2555.8 ms capped vs 3343.3 ms uncapped, same box, same binary, same prompt). So this ceiling is worth 23.4% **within** the dense-FP8 path, and that path is still the slower choice on this arch.

## Test plan

Spark 1 (`pidtom@192.168.50.125`), private `CARGO_TARGET_DIR=$HOME/atlas-gate-w8a8cap-target`, `ATLAS_SKIP_BUILD=1`, `CARGO_PROFILE_DEV_DEBUG=0`:

```
cargo fmt --all -- --check
cargo test -p atlas-kernels --no-default-features --test target_defaults
  14 passed, 0 failed
cargo test -p spark-model --features cuda --lib the_ceiling_is_inclusive
cargo test -p spark-model --features cuda --lib the_baseline_ceiling
cargo test -p spark-model --features cuda --lib a_zero_ceiling
cargo test -p spark-model --features cuda --lib layer_selects_w8a8
  all ok
```

(`--lib dense_ffn_w8a8_prefill` compiles the crate and filters to 0 tests: the module is `w8a8_prefill::tests`, so the ceiling cases were run by name.)

`crates/spark-model/Cargo.toml`: 52 `[[example]]` stanzas, 0 duplicates after the #1006 merge.

- **No GB10 ladder numbers invented here.** The TTFT pair above is Thomas's served A/B, not a campaign ladder.
- **GB10 default serve is unchanged** unless `ATLAS_DENSE_FP8=1` is set.
- **B200 is unmeasured** and keeps `u32::MAX` on purpose.
- **Hopper kernel files are not in this PR.** The cherry-pick dropped `ssm_gdn_hopper_lever_tests.rs` and the umbrella's extra `[defaults]` / `[kernels]` rows that #1006 does not yet have.
- **Spark 2 GB10 receipt is pending.**

## CLA


tbraun96:
Retargeted to `pr/target-defaults` (#1046): this branch is #1046's tree plus #1027–#1029's commits plus one commit of its own (`35a3920e3`), so its real parent is the Hopper stack, not `pr/ffn-w8a8-prefill` (#1029's branch, which is now a dead end after chain A merged down and landed on `main` as `f8fcd346a`).

Plan: once #1045/#1046 have `main` merged in (so the W8A8 prefill files from #1029 are present via main rather than via this branch's inherited commits), I'll rebase this PR down to its one own commit on top of `pr/target-defaults` — `--force-with-lease`, and I'll note the old head here first. Nothing is pushed to this branch until then.

PR #1060, opened 2026-09-12
Qwen3.8-Flash-Next: the K-row MoE decode arm, 4 to 8 rows then 16 (#974 rung 2)

rrstesiak:
#1026 puts lookup drafts into the wide verify on Flash-Next. It runs at K=3 rows because that is where the small-M MoE FFN arms stop. This PR lifts that limit so a lookup step can commit 8 tokens, then 16. It is the rung that turns #1026 from an acceptance gain into an agentic-session gain, and it revives MTP-3 and MTP-4 as measurable options instead of a dead end.

## The cliff, exactly

`hc_ffn_dispatch` routes the MoE FFN by row count: 1 row, 2 rows, 3 rows, then `Prefill` for anything wider. `MoeLayer::forward_prefill` takes the grouped GEMM only above 64 tokens; below that it is `forward_batched`, the per-token expert loop with data-dependent routing. So a K=4 verify step runs four single-token MoE passes per layer, with the per-token launches and the routing round trips that come with them. Measured during the #972 series: MTP-3 at K=4 rows, LRU 14.9 tok/s, against 62.0 at K=3.

The K=3 arm itself is at the DRAM floor (09-09 census): `moe_expert_gate_up_shared_batch3_t` streams 57.5 MB in 188 us, 306 GB/s; `silu_down` at 239 GB/s; 15.0 ms of the step is MoE weight traffic. That number reproduces from the geometry: 512 experts, top-10 plus one shared, intermediate 640, hidden 2560, NVFP4 with FP8 block scales. Gate+up per expert is 1.84 MB, down is 0.92 MB. Three rows times ten experts plus the shared expert is 57 MB per gate-up launch. The batch3 kernel gives every (row, expert slot) its own block row (`blockIdx.y` over `3 * top_k`, shared expert blocks after), so it reads an expert once per row that routes to it. Weight traffic is linear in rows: about 27.6 MB per row per layer, 1.33 GB per row per step, 4.3 to 5.5 ms per row at the measured rates.

## Every K-row piece on this model, and where each one stops

| piece | today | cap | for K=8 | for K=16 |
|---|---|---|---|---|
| MoE FFN (routed + shared) | Single / K2 / K3 fused `_t` kernels; wider = per-token loop | **3** | new `batchN` arm | same arm, N to 16 |
| GDN (linear attention, 36 layers) | batched conv+GDN dispatch rows 2, 3, 4 (`wy2`/`wy3`/`wy4`); wider returns `Ok(false)` | **4** | wire `wy5`..`wy8`, already built (`gated_delta_rule_wyn`, loaded by `wyn_kernels`) | wire to `wy16` |
| attention projections (12 full-attention layers) | `dense_gemv_batchm`, weight read once for n rows | **8** (`DENSE_GEMV_BATCHM_MAX_M`) | covered | raise to 16 |
| hyper-connection sites | decode-shaped collapse `hc_dec_up` / `hc_dec_down`, T rows per pass | **8** (`QHC_DEC_MAX_T`) | covered | raise to 16; past 8 today the split arm reads the 13 MB per token per site again |
| attention core (paged decode) | per row, 38 launches per row across the 12 layers | none | covered, launch count grows | batch over K rows with a replicated block table (09-09 item 8) |
| QSA indexer | `qsa_score` grid (rows, heads) | none | covered | covered |
| PLE n-gram gather | one gather per row, NVMe miss gap on most steps | none | covered; verify-row prefetch (e8096d26) helps | same |
| LM head, verify rows | GEMM at M=K | none | covered | covered |
| router top-k | `moe_topk_*_batched` at 3 rows | 3 | N rows | N rows |
| host side | eager verify, D2H argmax per verify; #973 removes the per-row D2H on the greedy path | | covered | graph capture stays the structural item |

Two rungs fall out of that table. **K=8** needs the MoE arm and the GDN wiring, nothing else. **K=16** needs the same arm at N=16 plus three ceilings raised (projection GEMV, HC collapse, attention core batching). The PR lands K=8 first and certifies it; K=16 is the second commit series on the same branch.

## The arm

**A. `batchN`, slot-major.** The straight generalization of the batch3 `_t` family: `moe_expert_gate_up_shared_batchN_t`, `moe_expert_silu_down_shared_batchN_t`, `moe_weighted_sum_blend_batchN`, with N a launch parameter (grid.y = N * top_k + shared slots), the router GEMV through `dense_gemv_batchm`, top-k through the batched kernels at N rows. Same per-slot structure and the same per-row reduction order as batch3, so each row is bit-identical to the K=3 arm's row, which is what the parity test checks. Traffic linear in N at the floor: roughly 4.8 ms per row. This alone takes K=8 from the per-token loop (about 14.9 tok/s at K=4) to the floor.

**B. expert-major, deduplicated.** The N * top_k slots are sorted by expert on device (at most 160 entries, one block), and one block row runs per distinct expert, looping over the rows routed to it. Traffic becomes the number of distinct experts in the step, not rows times top_k. Under independent routing that is 512 * (1 - (1 - 10/512)^N): 29 at N=3, 76 at N=8, 139 at N=16, against 30 / 80 / 160 for the slot-major arm, so the random model gives 13% at N=16. Real routing overlaps more than random: the 07-19 measurement on Qwen3.6-35B-A3B (top-8 of 256, 17-token windows of generated text) found 59 distinct experts where the random model predicts 107 and the slot count is 136, on code and prose alike. Whether Flash-Next's router behaves the same on a copied block is measurable today with zero code: `ATLAS_MOE_UNION_STATS=1` samples exactly this union on every verify batch. That measurement decides whether B is built. A is built regardless.

**Width is per step, not per serve.** The verify already dispatches on `pending_drafts.len()`. With this arm the lookup gate proposes at `ATLAS_LOOKUP_WIDTH` (default 7 drafts, K=8 rows) while the MTP head stays at its own width (2 drafts, K=3). Fresh generation never pays for the wide arm. That is the difference from the MTP-3 census on 09-09, which was net negative because a third draft pass and 33% more expert traffic bought 13.6% more tokens on every step; here the wide step runs only when the index has the tokens already.

## Cost model, anchored and labeled

Measured: MoE at K=3 is 15.0 ms; per-row MoE traffic 1.33 GB at 239 to 306 GB/s; the record step at K=3 is 62.0 tok/s on LRU. Estimate from the census: the non-MoE part of the step is about 30 ms and mostly weight-read-once GEMVs, so it grows slowly with rows.

Step time model, estimate: T(K) = 30 ms + 4.8 ms * K. Copy-task tok/s at 95% acceptance (lookup on a repeated block), estimate:

| K rows | drafts | step (ms) | tokens committed | tok/s |
|---|---|---|---|---|
| 3 | 2 | 44 | 2.85 | 64 |
| 4 | 3 | 49 | 3.8 | 77 |
| 6 | 5 | 59 | 5.7 | 97 |
| 8 | 7 | 68 | 7.6 | 111 |
| 12 | 11 | 88 | 11.4 | 130 |
| 16 | 15 | 107 | 15.2 | 142 |

With arm B and Flash-Next routing overlapping like the 35B did (about 95 distinct experts of 160 at K=16), the K=16 step is about 76 ms and the copy-task number is about 200 tok/s. All of these are model numbers; the cells below replace them.

The competing number for the copy-task class stays what #974 lists: llama.cpp 97.4 tok/s on one Spark at 94.7% acceptance (NVIDIA forum, Flash-Next thread p.4, 26 Aug). K=8 at the floor sits above it on the model; K=16 sits well above it. Fresh-generation numbers (62.0, #973) are a different table and do not move with this PR.

## Plan

| step | what | GPU |
|---|---|---|
| 0 | `ATLAS_MOE_UNION_STATS=1` on the #1026 serve, copy task and MinHeap, K=3: distinct experts per layer-step vs 30. Decides A-only vs A+B. | yes, no code |
| 1 | `batchN` slot-major kernels (gate_up, silu_down, wsum_blend) for N in 4..=8, microtest against batch3 rows, `hc_ffn_dispatch` gets a `KN` arm, `MoeLayer::forward_kn`. | yes |
| 2 | GDN wide rows: wire `wy5`..`wy8` into the batched conv+GDN dispatch for this model; parity test `hc_rows_t8_rows_equal_t1_rows`. | yes |
| 3 | `ATLAS_LOOKUP_WIDTH` in the gate (#1026), default 7; MTP width unchanged. | no |
| 4 | Cells: copy task at K=3 vs K=8 (own table), MinHeap x3 and Volvo x3 unchanged vs base (byte-identical, the wide arm never fires there), `spark benchmark run agentic-webserver`. MTP-3 and MTP-4 re-measured on the new arm as a side row. | yes |
| 5 | K=16: N to 16 in the arm, `DENSE_GEMV_BATCHM_MAX_M` and `QHC_DEC_MAX_T` to 16, `wy9`..`wy16` wired, paged decode batched over rows. Arm B if step 0 says so. | yes |

Defaults ON at each rung once its cells are in. Kill switches: `ATLAS_FFN_SMALLM=0` already routes wide rows to the base path; `ATLAS_LOOKUP_WIDTH=2` restores #1026's shape.

## Not in this PR

The 27B: its dense FFN already takes K=5..16 through the batched GDN verify (#844, #845), and the lookup gate lands there as its own PR. Hopper: no NVFP4 path, the arm's FP8 twin is Initiative 2 work. EXL3 experts: Richard's, the `NativeBatched` arm stays as is.

Related: #974, #1026, #972, #973, #844, #845, #837, #834.


rrstesiak:
Code for the K=8 rung is on the branch.

The MoE arm: `moe_shared_expert_fused_batchn.cu` is the batch3 family with the row count as a launch argument, same block per (row, expert slot) and the same reduction order, so each row is bit-identical to the batch3 row. `MoeLayer::forward_km` runs the router GEMV once for m rows, batched top-k, fused gate+up, fused silu+down, and the weighted-sum blend; `hc_ffn_dispatch` takes rows 4..=8 through it when the layer reports the arm available (NVFP4 decode layout; native EXL3, adapters, BF16 or E8M0 experts and the unified layout keep their paths).

The GDN side needed nothing: the single-sequence K-row verify already dispatches `wy5`..`wy16` through the wyN arm, so the survey's GDN line for K=8 was already closed in the tree. The projection GEMV and HC collapse ceilings at 8 stand as written.

Checks: `cargo check -p spark-model -p spark-server --tests` clean. Dispatch tests in `hc_ffn_dispatch.rs`. GPU parity (rows at K=8 against the K=3 arm, byte-identical) and the copy-task cell are the next measurements.

rrstesiak:
K=8 end to end is on the branch.

`decode_verify_graphed_kn` is the width-generic verify entry: the Flash-Next highway verify already takes any token slice, so the model side needed the entry and nothing else. `step_verify_kn` runs the MTP-shaped wide verify at K = drafts + 1 rows and hands the result to the same `finish` as K=3 and K=4; `mtp_step` routes a non-DFlash sequence holding four or more drafts to it (previously that count went to the DFlash block verify). `ATLAS_LOOKUP_WIDTH`, default 7, is the width a lookup hit proposes at; the MTP head keeps its own width, so the wide step runs only on steps the index already has the tokens for.

Checks: `cargo check -p spark-model -p spark-server --tests` clean. Next measurements on the box: MTP_DRAFTS=3 on the record recipe (K=4 rows through the arm, against the 14.9 dead end), then the copy task with lookup at width 7.

rrstesiak:
Measured on this branch (Qwen3.8-Flash-Next NVFP4, one GB10, MTP recipe, temperature 0; this tree is wip/exl3-research plus #1026 and this PR, without the #972/#973 kernels).

Standard cells, MTP-2, three runs each: with lookup at match length 8, all nine runs byte-identical to the same binary with `ATLAS_LOOKUP_DRAFTS=0`, at base tok/s (LRU 25.4, MinHeap 24.7, Volvo 23.0). Match length 4 fires on fresh text and moves every stream; 8 is the default.

The MoE arm, MTP-3 (K=4 rows), lookup off: LRU 24.6 / MinHeap 23.0 / Volvo 20.5. Same settings with the arm off (`ATLAS_QWEN4EXP_HC_SMALL_M_FFN=0`): 13.7 / 12.2 / 11.8. The per-token path was the 14.9 dead end; the arm is reached.

Copy task (a fixed 60-line Python file, one rename, 682 completion tokens, identical output on every run): lookup off 26.7 tok/s with 435 head drafts accepted; lookup at K=3 29.1 with 445; at K=4 29.3 with 496. A launch at 7 drafts also returns 496: the SSM verify pool tier holds the slot at 3 drafts on this configuration, so K=8 waits on that tier. The wide verify on this tree is the eager mini-prefill, so a wider step here costs about what it earns; the K-row verify body in #972 is what makes the step cheap, and the copy-task number on the combined tree is the next measurement.

Fixed along the way, on #1026: the proposer drafted from the token list's suffix while the newest token was still held apart, so every draft was one position stale and rejected (Richard's bisect on his TP=2 x EP=2 build saw the same as a 40 percent decode loss). The proposer now matches the newest token too, and the gate stays off under expert parallelism.

rsafier:
@rrstesiak — built the width-generic EP verify command. It works, lookup drafts now reach full width on a multi-rank serve, and the copy task goes **55.28 → 74.77 tok/s (+35%)**. But the width that wins there is *not* the width to default to, and that's the more useful half of this.

**`EP_CMD_VERIFY_KN` (`0xFFFF_FFF7`)**, on top of your `c8d8c4a15`:

- Wire is `cmd, k, tokens[k]`, then `num_accepted` after the forward. **`k` is broadcast before the tokens** so both ranks drive their receive loops from the same word — that seemed like the only real deadlock surface, versus encoding width in the command code. Worker bounds-checks `2..=64`.
- The worker arm is your K=4 arm with the width lifted out: `rewind = (k-1) - num_accepted`, full accept checkpoints, otherwise pop and roll back to `num_accepted + 1`. It calls your `decode_verify_graphed_kn`, so no new verify math.
- `verify_mtp_wide::finish` already broadcasts the verdict for every width and sends 0 on its error paths, so the head can't hang the worker.

**A prerequisite that wasn't obvious:** `verify_draft_capacity` is `num_intermediates - 1`, and your clamp pins the proposal to it — so pools sized from the *head's* draft count silently capped a width-7 lookup at **2**, and `step_verify_kn` never ran at all on our config. `ssm_reserve::mtp_pool_draft_width` now sizes from `max(head, lookup_width)`, keeping the head narrow (a wider head is a 26% C=1 loss here — DRAFTS=3 measured, with your Km arm confirmed engaged).

Each step alone, copy task, one binary, TP=2 × EP=2:

| | tok/s | limiter |
|---|---|---|
| lookup off | 55.28 | — |
| EP gate lifted | 61.88 | pool capacity clamps width to 2 |
| + pools widened to 7 | **53.31** | proposes 7, verifies 4, discards 4 |
| + wide-K command | **74.77** | none |

Widths 2 and 3 measure flat (61.88 / 61.20) because both verify at K=4 — that's the evidence the command is what makes pool width pay, and that widening pools *without* it is a regression below baseline.

**★ The width default should be 3, not 7**, and this is the part I'd flag hardest:

| | copy task | agentic s/turn | turns / Σwall |
|---|---|---|---|
| width 3 | 61.20 | **7.00** | 46 / 325 s |
| width 7 | **74.77** | 13.43 | 41 / 554 s |

Both 3/3 + 3/3, 0 errors, anchors 0 on both ranks. A lookup step verifies width+1 rows; on agentic text the match is often only partially right, so a 7-draft proposal that lands 2 still pays an 8-row verify and rolls back 5. Width 3 ran **more turns in less wall**, so this isn't the turn-count variance that usually makes this battery unquotable. Width 7 also costs 2286 MB of verify pools vs 1150, and 1.22M vs 1.30M KV tokens.

So your table's 39.1 row is real and reachable under EP now — it's just the copy-task regime specifically. I've defaulted `ATLAS_LOOKUP_WIDTH=3` in our launcher with 7 documented as the opt-in for copy-heavy serving.

All on `pr973-ep-prefix-cache` (`32f5b1c3a`, `23b2666f4`) — the command and the pool sizing are both yours to take into #1060 if you want them there, and it's your call whether `|| ep` comes out of `lookup_gate` upstream.

rsafier:
@rrstesiak — cherry-picked `7f1a82035` and re-measured width 7. **Your diagnosis was right and mine was wrong**: I blamed partial lookup matches paying wide verifies; the real cause was the head re-proposing 7 on fresh text, which is DRAFTS=7 in disguise — and this branch independently measured DRAFTS=3 at −26% C=1, so a head at 7 nearly doubling s/turn fits exactly.

Width 7, TP=2 × EP=2, one binary, both ranks sha-matched:

| | pre-fix | on your fix |
|---|---|---|
| agentic s/turn | 13.43 | **7.31** (width 3: 7.00) |
| copy task | 74.77 | **74.65** (width 3: 61.20, off: 55.28) |
| fresh-text C=1 | — | 53.46 (unchanged) |
| agentic verdict | 3/3 + 3/3 | **2/3 + 2/3** |

So the speed objection to width 7 is gone — it matches width 3 on agentic while keeping the full +22% copy-task win. That's your fix doing exactly what it says.

**The one thing I'd want your eyes on:** that run is the first agentic failure of this campaign. Run 2 burned 26 turns and never wrote `Cargo.toml` (5/6 steps, 198 s), while runs 0 and 1 finished in 10 and 7 turns. Anchors 0 on both ranks, Marconi 40/40, 0 errors — so it's task adherence, not corruption or the wire.

I'm not claiming it's your fix or width 7. Spec width changes which tokens are *sampled* (we run model-card sampling, temp > 0), so trajectories differ between widths by construction, and n=3 can't separate an unlucky run from a real quality cost. What makes me mention it at all is that every other config measured this session went 3/3 — this is the first 2/3, so it's worth someone watching for on your side too.

Our default stays at width 3 for now, on that quality question rather than the speed one, with width 7 documented for copy-heavy serving. If it doesn't reproduce for you, I'd happily move to 7.

Also, since it bears on rung 3: I got the C=4 step breakdown via `ATLAS_MTP_TIMING`. At C=4 a step is 122 ms, of which **fwd is 113 ms (93%)** and propose 8.6 ms — commit, D2H, dequant and the collectives are all under 1% combined. So C=4 is forward-bound, not overhead-bound, and our C=4 verify is **11 rows**: outside your Km arm's 4–8, landing on the 3-row chunked ladder as four sequential fused MoE calls. Your "then 16" rung would make that one batched call, and the timing says that's where the time actually is.

rrstesiak:
> @rrstesiak — cherry-picked `7f1a82035` and re-measured width 7. **Your diagnosis was right and mine was wrong**: I blamed partial lookup matches paying wide verifies; the real cause was the head re-proposing 7 on fresh text, which is DRAFTS=7 in disguise — and this branch independently measured DRAFTS=3 at −26% C=1, so a head at 7 nearly doubling s/turn fits exactly.
> 
> Width 7, TP=2 × EP=2, one binary, both ranks sha-matched:
> 
> pre-fix	on your fix
> agentic s/turn	13.43	**7.31** (width 3: 7.00)
> copy task	74.77	**74.65** (width 3: 61.20, off: 55.28)
> fresh-text C=1	—	53.46 (unchanged)
> agentic verdict	3/3 + 3/3	**2/3 + 2/3**
> So the speed objection to width 7 is gone — it matches width 3 on agentic while keeping the full +22% copy-task win. That's your fix doing exactly what it says.
> 
> **The one thing I'd want your eyes on:** that run is the first agentic failure of this campaign. Run 2 burned 26 turns and never wrote `Cargo.toml` (5/6 steps, 198 s), while runs 0 and 1 finished in 10 and 7 turns. Anchors 0 on both ranks, Marconi 40/40, 0 errors — so it's task adherence, not corruption or the wire.
> 
> I'm not claiming it's your fix or width 7. Spec width changes which tokens are _sampled_ (we run model-card sampling, temp > 0), so trajectories differ between widths by construction, and n=3 can't separate an unlucky run from a real quality cost. What makes me mention it at all is that every other config measured this session went 3/3 — this is the first 2/3, so it's worth someone watching for on your side too.
> 
> Our default stays at width 3 for now, on that quality question rather than the speed one, with width 7 documented for copy-heavy serving. If it doesn't reproduce for you, I'd happily move to 7.
> 
> Also, since it bears on rung 3: I got the C=4 step breakdown via `ATLAS_MTP_TIMING`. At C=4 a step is 122 ms, of which **fwd is 113 ms (93%)** and propose 8.6 ms — commit, D2H, dequant and the collectives are all under 1% combined. So C=4 is forward-bound, not overhead-bound, and our C=4 verify is **11 rows**: outside your Km arm's 4–8, landing on the 3-row chunked ladder as four sequential fused MoE calls. Your "then 16" rung would make that one batched call, and the timing says that's where the time actually is.
> 
> https://claude.ai/code/session_01PE72NZ6oBfrjVKZEmPoUHd

Worth separating two things in that last datapoint, since they read as one. 
The 26 percent is the cost of running the draft head at three drafts instead of two, which is a head-width result on this model. The part that touches the K-row MoE arm is km_available=true, which says the batched arm was engaged rather than the 512-expert prefill fallback. On one Spark that arm takes K=4 from 13.7 to 24.6 tok/s, a 1.8x,
against the 14.9 the fallback used to give. So both facts hold: a third draft from the head does not pay on this model, and the arm that verifies four rows is a clear win.

tbraun96:

tbraun96:

PR #1064, opened 2026-09-12
ci(release-build): deadlines on the toolchain downloads so a stalled fetch cannot hold the merge queue

tbraun96:
## What

`release-build.yml`: `timeout-minutes` on the four toolchain-install steps that fetch from external hosts, and a real deadline on the SCALE tarball `curl` (`--connect-timeout 30 --max-time 900 --speed-limit 524288 --speed-time 60 --retry-all-errors`).

## Why

Merge-queue run [34704277191](https://github.com/Avarok-Cybersecurity/atlas/actions/runs/34704277191) for #1024 sat in `Install SCALE toolchain (AMD)` for 50+ minutes today (normal: ~2 min). The step had no `timeout-minutes`, the `curl` had no `--max-time`, so the only limit was the job's 6-hour default — meaning a single dead TCP connection holds the whole merge queue for six hours. `pkgs.scale-lang.com` served the same 1.4 GB file at 25 MB/s from our box at the same moment, so it was the connection, not the mirror. The run was cancelled and #1024 re-queued.

| step | normal (from #1023's queue run) | deadline now |
|---|---|---|
| Install SCALE toolchain (AMD) | 2m14s | 20 min + curl stall abort at <512 KiB/s for 60 s |
| Install CUDA toolkit (linux) | ~45 s | 15 min |
| Install CUDA toolkit (windows) | ~7 min | 30 min |
| Install NCCL | ~11–45 s | 10 min |
| Install HIP SDK (AMD, windows) | ~4m40s | 30 min (already) |

A stalled required check is indistinguishable from "still running" — failing fast is what makes the entry retryable.

No `PERF_PATHS`; `.github` only. `certification-selftest.sh`: 242 passed.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -278,6 +278,7 @@ jobs:
 
       - name: Install CUDA toolkit (linux)
         if: startsWith(matrix.toolchain, 'cuda') && matrix.os == 'linux'
+        timeout-minutes: 15  # normally ~1 min; a deadline so a stalled download cannot hold the merge queue
         run: |
           set -euo pipefail
           # Installed straight from NVIDIA's apt repo rather than via the
@@ -316,6 +317,7 @@ jobs:
 
       - name: Install CUDA toolkit (windows)
         if: startsWith(matrix.toolchain, 'cuda') && matrix.os == 'windows'
+        timeout-minutes: 30  # normally ~7 min; a deadline so a stalled download cannot hold the merge queue
         shell: pwsh
         run: |
           # NVIDIA's own silent network installer. The cuda-toolkit action has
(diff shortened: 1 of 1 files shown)


tbraun96:
**Second commit, after the first one did its job:** with the deadlines in place the AMD leg failed in 12.5 min instead of hanging — four `curl: (28) Operation too slow. Less than 524288 bytes/sec` in a row. So `pkgs.scale-lang.com` is not dead from Azure, it is slow (≈1 h for 1.4 GB), while this box pulls the same file at 25 MB/s.

`e30c36d80`: the tarball is now mirrored on this repo's release [`toolchain-scale-1.7.1`](https://github.com/Avarok-Cybersecurity/atlas/releases/tag/toolchain-scale-1.7.1) (a prerelease, clearly labelled as a CI dependency, not an Atlas build; 42 MB/s measured from here). The step tries the mirror first, upstream second, and **verifies a pinned sha256 either way** — `060a3627…a2be`, taken from a fresh upstream download (`xz -t` clean, `targets/gfx1151/bin/nvcc` present). Negative control: a wrong digest fails `sha256sum -c`. Also merged `main` (behind by #1024).

Stamp/Seal are void with the new head; re-posting.

PR #1063, opened 2026-09-12
Qwen3.8-Flash-Next on 2x GB10: EP=2, native EXL3, and TP=2 x EP=2 (+9.1% C=1 decode, 2.4x KV)

rsafier:
**Draft.** Research branch for Qwen3.8-Flash-Next (`qwen4_exp`) on 2x GB10, opened for visibility rather than as a merge-ready unit: 181 commits, 836 files. It is 16 commits behind `main` and deliberately not merged forward — see *Before this merges*.

## What it is

Bring-up and optimization of Qwen3.8-Flash-Next across two DGX Spark GB10s, in three strands that grew together:

- **NVFP4 serving at EP=2** — 128K x 4 sequences, MTP speculative decode, multi-rank prefix caching, cross-sequence batched verify under the mHC highway.
- **Native EXL3 (packed QTIP trellis)** — MoE + dense + lm_head served from packed trellis, 2.05 / 4.05bpw, EP=2.
- **Tensor parallelism** — `--tp-size 2 --ep-size 2` on the same 2 ranks (the final commits).

## Headline measured results

All figures same-binary A/B, 5 reps, median, on dgx-00 + gx10-9959.

| | EP-only (TP=1) | TP=2 x EP=2 |
|---|---|---|
| C=1 decode | 49.21 | **53.70 tok/s (+9.1%)** |
| C=2 / C=4 decode | 62.30 / 68.69 | 63.45 / 65.3-67.6 (wash) |
| cold prefill 8K / 11K | 228 / 247 | **259 / 268 tok/s** |
| KV capacity | 639 K tok | **1.55 M tok (2.4x)** |
| MTP p1 / tok_step | 0.83 / ~2.0 | 0.77-0.91 / 2.15-2.42 |

Gates: known-answer 4/4 solo + 4/4 concurrent; `agentic-webserver` **3/3 webserver_ok, 3/3 followed_directions**, 10.30 s/turn, 0 errors.

The single largest win on the branch is not in that table: a highway-row defect had every sequence drafting from mHC row 0, holding MTP acceptance at p1 0.19. Fixing it took C=4 from 28.2 to 50.2 tok/s. Output stayed byte-correct throughout, so every gate passed while it was broken.

## Reviewer orientation

Most perf commits carry their measurement in the message, including the ones that failed. Worth reading in this order:

1. `feat(qwen4_exp): TP=2 x EP=2` and `fix(qwen4_exp): size the MTP drafter end-to-end at full width` — the TP enablement and the four defects behind it.
[...]

PR #1066, opened 2026-09-13
Qwen3.8-Flash-Next on 2x GB10: TP=2 x EP=2, vision across ranks, and a gate to prove it

rsafier:
**Stacks on #1060 — land that first.** The base here is `wip/exl3-research`,
which is also #1060's base, because #1060's head branch lives in a fork and
GitHub will not take a fork branch as a base. This branch already CONTAINS
#1060's four commits — verified patch-equivalent, including `7f1a8203`, the fix
that makes the serial verify steps draft at the ladder's per-step count — so the
two do not conflict. Reviewing this one first means reviewing his work twice.

## What this is for

**A gate anyone can run.** `kernels/gb10/qwen3.8-flash-next/BENCH.toml` is new — this target had none, so nothing here was defended by a threshold. Three entries against `nvidia/Qwen3.8-Flash-Next-NVFP4`, measured on two GB10s through the preset this PR adds:

```
spark serve qwen3.8-flash-next-nvfp4-tp2-ep2 --rank 0 --master-addr <rank0-ip>
spark serve qwen3.8-flash-next-nvfp4-tp2-ep2 --rank 1 --master-addr <rank0-ip>
spark benchmark run vision-fidelity --url http://<rank0>:8888 \
    --model qwen3.8-flash-next-nvfp4-tp2-ep2
```

`--video-allow-ffmpeg` on both ranks for the video gate, or six of its fourteen legs skip.

| gate | measured here |
|---|---|
| vision-fidelity | PASS — 14/14 geometry, 3/3 probes, 10/10 integrity, **3/3 concurrency clean** |
| video-fidelity | PASS — 14/14 legs, 0 skipped |
| agentic-webserver ×3 | 3/3 and 3/3, 22 turns, 6.95 s/turn, 156 s |
| cold prefill 7.8K / 10.8K | 1478 / 1435 tok/s |

`concurrency_levels_clean >= 3` is the load-bearing threshold. It read 0/3 before this work, 2/3 after two of the three vision fixes, and 3/3 only with all of them. Geometry stayed 14/14 the entire time, so a gate built on geometry alone would have called a badly broken engine healthy.

## Multi-node enablement

**TP=2 × EP=2.** `supports_tp()` returned false on a wrong premise: the mHC highway is this model's *residual stream*, and a residual stream is replicated under Megatron TP. Lifting it needed three fixes, each of which produced correct-looking output while being wrong — the MTP shape audit was handed the TP-divided config and failed into a silently speculation-free serve, and the drafter then inherited the target's per-rank config and ran 12-head attention over its own 24-head weights. C=1 49.21 → 53.32 tok/s, KV 639K → 1.55M tokens.

**Vision across ranks.** Only rank 0 receives image bytes, so only rank 0 ran the ViT. Every other rank kept the raw `<|image_pad|>` row and, with no grids, built a linear position stream where rank 0 built (T, H, W). The ranks all-reduce every layer. The model stayed fluent — it read **HARBOR** as *Superman* — which is why this passed every check that looked at rank 0 alone: tower cosine 0.998 vs HF, splice exact at 1.0000, positions HF's to the digit, all on one rank of two.

**Prefix cache across ranks.** Both ranks must agree the Marconi anchor or nobody restores, and sequences have to retire on the worker too so both snapshot the same leaf.

**A width-generic EP verify command.** `0xFFFF_FFF2..F4` hardcoded one command per width, so `verify_kn_step` narrowed to K=4 under EP and lookup drafts could not reach their configured width. `k` is broadcast *first*, so both ranks derive their read count from one word — the deadlock surface for any new EP path.

**The PLE n-gram table pinned to local NVMe on both ranks**, and the serve preset, launcher and gate that tie it together.

## Two defects fixed that are not multi-node, but are load-bearing here

**Decode resumed at the token index.** A vision item spans many pad tokens and few positions, so the rotary stream runs behind the token stream forever after the first image. HF carries the difference as `rope_deltas`; we carried nothing. One 224×224 image put every generated token 42 positions past the end of its own prompt.

**The batched prefill stager has no MRoPE arm**, and nothing refused vision prompts, so a second in-flight request turned an image into a flat run of positions. Token counts and geometry were identical throughout.

## Open, stated rather than hidden

The cross-sequence batched verify is not output-equivalent at batch width ≥ 4. It is **mitigated** in the preset by switching off the three arm groups that break parity — GDN/SSM, attention/HC/o_proj, and D-Cut — which costs 1.9% at C=4 against 47% for disabling the arm outright. Root cause is not found. Reproducer, the full ruled-out list and the three groups are written up for whoever picks it up; `concurrency_levels_clean` is what catches a regression.

`recipe` in BENCH.toml names `qwen3.8/qwen3.8-flash-next-nvfp4-ep2`, which does not exist in atlas-recipes yet. The schema requires a binding and the existing single-node recipe is not it. Self-start cannot provision two nodes regardless, so it costs nothing today and still needs adding.

## Note on scope

I first cut this as a narrow multi-node subset on top of #1060. It built and booted, and video-fidelity passed 14/14 on it — but vision-fidelity's `reads-the-size-label` probe failed at **every** width including C=1. The model read the label correctly each time and answered in the terse form, which the standalone-term matcher rejects. That is a marginal token whose outcome moves with the engine's numerics, so a subset missing the perf arms cannot meet a threshold calibrated on the full tree. Since a runnable gate is the point, this carries the full fix set.


tbraun96:

tbraun96:

PR #1076, opened 2026-09-14
feat(k3): KDA module + FLA-unbounded gate

TheTom:
Extracted from working log #1053. **Do not merge #1053.**

Kimi K3 KDA as a new backend (not GDN/Mamba). CPU ref + unique `kda_decode` CUDA stem. Twin `inference-optimization/Kimi-K3-0.40B` omits `gate_lower_bound` → FLA unbounded (`None`). Official JSON is `-5`.

## Gate
RST: `docs/k3/rst/feat-k3-kda.md`
Twin C1 oracle (umbrella): `docs/k3/rst/c1-engine.md`

Known-bads that moved:
- `Some(-5)` vs `None` max-abs > 0.1
- wrong conv/recurrent slot after prefix
- raw beta=0 still writes `sigmoid(0)=0.5`

## Not in this PR
MLA, AttnRes, LatentMoE, hybrid cache, TP, official 1.56 TB, `/stamp`.

- [x] I have read and agree to the [Contributor License Agreement](CLA.md).

PR #1053, opened 2026-09-12
WIP: Kimi K3 architecture bring-up (do not merge)

TheTom:
> Same-repo restack of #1020 (was `TheTom:wip/k3-bringup`). Commits unchanged. Opened on `Avarok-Cybersecurity/atlas` so Graphite/GH can stack.

# WIP: Kimi K3 architecture bring-up (do not merge)

> **This PR is not expected to merge.** It is a historical working log. The branch will stay messy on purpose — dead-ends, WIP commits, lab comments, half-written modules — until we cut clean topic PRs *off of it*. Do not `/stamp`, `/seal`, squash-merge, or review it as a landing candidate. Reviewers: ignore mergeability; use the conversation as the notebook.

Working log for Atlas K3. **Draft only.** Winning slices are extracted later into atomic certified PRs. See `docs/k3/PRD.md`.

## Summary

Working log for Kimi K3 architecture bring-up on lab hardware (spark1 / spark2 / 5090). **Not a merge candidate. Expected to be historical and messy** until new PRs are opened from this branch. Architecture is completed at home; rental silicon is a soak after C0–C7.

Closes #

## Status

| Field | Value |
| --- | --- |
| Phase | **C0–C7 green** on the 0.40B twin. CUDA KDA+MLA default. Packed LatentMoE launches DSV4 E8M0 GEMM. 0.40B CPU MoE. AttnRes/projections host. |
| Last host | workstation (Mac) + spark1 + spark2 + train (5090) |
| Last session | 2026-09-13 — C2–C6 twin on spark2; C7 0.40B TP=2 MATCH + kill rank 1 timeout. |
| S0 bake-off JSONL | `docs/k3/logs/bakeoff-thinkoff-2026-09-11.jsonl` (not certified) |
| Official weights downloaded in lab | **no** |
| Ready to book | **yes** (`docs/k3/RENTAL.md`) |
| Rental booked | **no** |

## RST (required each phase)

Rapid Software Testing. Contract: `docs/k3/RST.md`. **A C-gate or S-rung is not green without a session sheet** that names the oracle and a known-bad the instrument actually failed. Checks are not the testing. Run independent charters in parallel when they do not share a GPU.

Sheets so far: `docs/k3/rst/s0-fabric-nccl.md`, `s0-5090-sm121.md`, `s0-atlas-serve.md`.

## Rental gate

S7 (rental soak) is forbidden until every box is green. Each box also needs its RST sheet.

[...]

PR #1077, opened 2026-09-14
feat(k3): hybrid cache (MLA KV + KDA state)

TheTom:
Stacked on #1076 (`feat/k3-kda`). Extracted from working log #1053. **Do not merge #1053.**

Hybrid cache: per-layer KDA conv/recurrent or MLA KV. Twin graph is 6 KDA + 2 MLA, last layer MLA. Prefix-cache restore is byte snapshot/restore of `LayerCache`.

## Gate
RST: `docs/k3/rst/feat-k3-hybrid-cache.md`
Umbrella twin C2/C3/C4: `docs/k3/rst/c2-c3-cpu.md`, `c4-c5-c6-cpu.md`

Known-bads that moved:
- trash-all KDA conv+recurrent after prefix clone
- swap last MLA K/V row with row 0

C2 full-graph logits still live on the umbrella until `cpu_forward` is extracted.


TheTom:
Extraction note from the #1053 review (pullrequestreview-5193698507), item 2: the per-token state round-trips called out on the KDA and MLA slices (#1076, #1078) land here as the design constraint. The hybrid cache is where device-resident KDA conv/recurrent state and MLA KV have to live before any certified tok/s. Host-side byte snapshot/restore for prefix caching is fine; per-token D2H/H2D on the decode hot path is not.

TheTom:
Addressed. Host `HybridCache` is prefix snapshot/restore only. `DeviceHybridCache::from_host` seeds GPU KDA conv/recurrent and MLA KV once (H2D). Decode must use those ptrs; per-token D2H/H2D of state is not this cache.

PR #1078, opened 2026-09-14
feat(k3): gated NoPE MLA

TheTom:
Stacked on #1077. Extracted from #1053. **Do not merge #1053.**

Gated NoPE MLA: `qk_rope_head_dim` slots stay in the head but are not rotated when `mla_use_nope`. Output `g_proj` sigmoid gate. Unique `mla_decode` stem. Official layer map: last two layers (0-based 91, 92) are MLA.

## Gate
RST: `docs/k3/rst/feat-k3-mla-gated.md`

Known-bads: output gate on vs off; NoPE vs rotate.


TheTom:
Extraction note from the #1053 review (pullrequestreview-5193698507), item 2: `mla_cuda.rs` round-trips per token: q/k come back D2H after rope (`:101`), the KV append happens on the host (`:105`), then the whole cache re-uploads (`:108-109`). That makes MLA decode O(T) H2D per token, O(T^2) per sequence. Correct for the twin; this slice needs device-resident KV before any certified number is taken against it.

TheTom:
Addressed. `MlaDeviceKv` + `launch_k3_mla_decode_token_on_device`: rope stays on device, append is one row (D2D/H2D), SDPA reads the buffer. Known-bad: `device_kv_second_token_d2h_is_output_only`. Host launch remains the CPU oracle (full KV re-upload).

PR #1079, opened 2026-09-14
feat(k3): AttnRes block residual mix

TheTom:
Stacked on #1078. Extracted from #1053. **Do not merge #1053.**

Block AttnRes softmax mix. mix=0 identity skip; mix=1 recorded fixture (atol 1e-5). Twin greedy mix=0 vs first-id 1459 remains the umbrella C5 run.

RST: `docs/k3/rst/feat-k3-attnres.md`


TheTom:
Extraction note from the #1053 review (pullrequestreview-5193698507), item 3: the umbrella fixed the AttnRes stream-map error path in `940bd4eeb` (drop the `DevicePtr` entry if `decode` returns Err; last-layer success still removes). Confirm this slice carries that commit. The leak is bounded and self-heals on pointer reuse, but the fix is one line.

TheTom:
Addressed. This slice did not previously contain BoundLayer `host_decode`. `AttnResHub::decode` now drops the key on `Err` (umbrella `940bd4eeb`). Last-layer success still `remove`s. Known-bad: `hub_drops_entry_on_decode_err`.

PR #1082, opened 2026-09-14
feat(k3): dual-Spark TP plan (supports_tp true)

TheTom:
Stacked on #1081. Extracted from #1053. **Do not merge #1053.**

K3 `supports_tp()` is **true**. Twin TP=2 plan: q_proj column-parallel `[256, 1024]`, o_proj row-parallel `[1024, 256]`. Rank 0 and rank 1 reconstruct the same full sizes.

Lab C7 (umbrella): 0.40B TP=2 aviation 16 tokens MATCH TP=1; kill rank 1 → timeout, not bee-fly. Pin `enp1s0f1np1` / `rocep1s0f1`. No lab IPs in git.

RST: `docs/k3/rst/feat-k3-dual-spark.md`


TheTom:
Extraction note from the #1053 review (pullrequestreview-5193698507), item 4: `tp_allreduce` (`host_decode.rs:242-262`) rounds each rank's f32 partial to BF16 before the NCCL sum. Token identity held on the twin (16/16 vs TP=1; kill-rank-1 times out instead of matching). At 7168-wide production rows the double rounding deserves a measured answer rather than an assumption.

TheTom:
Addressed. `production_7168_bf16_allreduce_rounding_is_measured`: each rank rounds f32→BF16 then sums. 1/3+2/3 across 7168: max_abs > 0 and < 0.01 (printed). Twin 16/16 still holds; this is the production-width figure, not an assumption.

Issue #1059, opened 2026-09-12
DeepSeek V4.1 Flash: single-Spark Q2_K text MVP — PRD and RST test plan


TheTom:
## Scope and relationship to #970

This is a self-contained PRD and implementation/testing tracker for **native DeepSeek V4.1 Flash Q2_K text inference on one GB10 Spark**. Requested owner: @rrstesiak. Research snapshot: September 12, 2026. **Planning proposal; no implementation, full-weight execution or numerical qualification is claimed.**

Related: #970, “DeepSeek V4.1 Flash on a single DGX Spark, speculative decoding on.” That issue predates the public checkpoint and includes a broader speculative-decoding objective. This tracker supplies the now-public architecture/storage findings and a **serial-first MVP**; MTP/speculation is deferred until this foundation works. It does not replace or claim completion of #970.

### Milestone checklist

- [ ] S0: Pin config/tokenizer/GGUF/runtime identities and approve the numerical contract and memory budget.
- [ ] S1: Validate all seven shards, actual tensor types, role mappings and bounds through production loader fixtures.
- [ ] S2: Validate initialized tiny V4.1 graph semantics against independent references.
- [ ] S3: Prove bounded expert streaming and compressed-to-resident engram row caching.
- [ ] S4: Demonstrate native serial Q2_K text generation and API behavior on one owner's Spark.
- [ ] S5: Complete RST sessions, affected-model regressions, required certification and a reproducible public handoff.

All checkboxes remain open. The specifications and experiments below can be used without access to the original investigator's files or machines.

### 1. Product decision

Deliver experimental, native text inference for **DeepSeek V4.1 Flash Q2_K on one GB10 Spark with 128 GB unified memory**, using local SSD-backed weights and bounded working memory. A teammate with their own Spark must be able to reproduce the result without access to another GPU, a second Spark, a rental cluster, or another developer's home directories. Atlas must execute the model; delegating generation to a llama.cpp subprocess does not satisfy this requirement.

Initial support means one active request, a maximum **2,048 tokens across serialized prompt and generated output**, ordinary autoregressive generation, correct text/chat serialization, end-of-sequence handling, and streamed/non-streamed responses through Atlas's existing OpenAI-compatible API. The cap includes template and special tokens. Ordinary multi-turn chat is supported by supplying the conversation again; cross-request prefix reuse is excluded initially.

[...]

Issue #970, opened 2026-09-08
DeepSeek V4.1 Flash on a single DGX Spark, speculative decoding on


rrstesiak:
Filed so the single-Spark DeepSeek lane has a number before the weights land. 

## What DeepSeek released today (2026-09-08)

- API beta only. Model id `deepseek-v4.1-flash-expires-on-0910`, same price as `deepseek-v4-flash`, 20 concurrent requests per account.
- DeepSeek's own description: "a new model architecture, with native multimodal support," faster, lower cost. No parameter count, no attention or MoE detail, no benchmarks published.
- No weights. The Hugging Face org still ends at DeepSeek-V4-Flash-Vision-Exp; the API changelog does not list 4.1 yet.

Until a checkpoint and config are public there is nothing to load. This issue records where the single-Spark work stands and what happens when they are.

## Where single-Spark DeepSeek V4 Flash stands, measured on one GB10

| stack | checkpoint | mode | tok/s | notes |
|---|---|---|---|---|
| **DwarfStar + my single-GPU DSpark patch (antirez/ds4 #800)** | **V4-Flash-0731, q2 GGUF** | **DSpark speculative** | **27.06 / 27.20 / 27.08** | **accept 91.1%, output byte-identical to serial** |
| **DwarfStar, same build** | **V4-Flash-0731, q2 GGUF** | **serial** | **21.33 / 21.30** | **upstream's published GB10 serial curve: 18.05** |
| Atlas main + eisbaw's fixes (#412) | REAP-162B pruned checkpoint | serial | 10 | pruned to fit one GB10; full V4 Flash on Atlas is an EP=2 recipe |

The DSpark line is the one that matters for 4.1: a running speculative stack on the full model on one Spark, bit-exact against serial. The mechanism that made single-GPU work is in #800 on antirez/ds4.

## What Atlas has on main for V4 Flash today

- Loader, MTP layer, preflight, tokenizer, and DSML tool parser under `weight_loader/deepseek_v4/`, `layers/deepseek_v4_mtp.rs`, `tokenizer/deepseek_v4/`.
- Open on one GB10: #412 (hash-MoE token_ids missing in decode_batch; cuStreamSynchronize inside graph capture) and #414 (MXFP4-E8M0 prefill fallback panic).
- Open on speculation: #364 (V4 MTP accepts ~16% and is net slower). Tool parsing: #363.
[...]

PR #1094, opened 2026-09-15
DeepSeek-V4.1 Flash: GGUF ingestion (S1)

rrstesiak:
DeepSeek-V4.1 Flash loads from the seven-shard Q2_K GGUF with nothing but the file: the `deepseek41` architecture maps, all seven shards resolve from shard 0 with the per-shard tensor counts summing to `split.tensors.count`, a complete `ModelConfig` builds from the GGUF's own metadata, and all 1,046 tensor names translate.

What the file taught the loader, and what this PR encodes:
- Experts are stacked. `ffn_{gate,up,down}_exps.weight` is one 3-D tensor per layer holding all 384 experts, routed through `GgufName::ExpertStack`.
- Several tensors exist on some layers only: `attn_compressor_kv` on the four `kv_source_layer_ids`, `indexer.attn_q_b` on the eight `index_source_layer_ids`, engram on layers 1 and 14. The translation is per name and assumes nothing per block.
- `exp_probs_b_vl.bias` is the vision routing bias and ships on all forty layers of this text quant. Mapped, not dropped, so a multimodal path never needs a re-convert.
- `compress_ratios` has 43 entries for 40 blocks: the DSpark stages carry their own ratio.
- The engram hash travels in the file: `deepseek41.engram.{multipliers,primes,offsets,token_map}`.
- The output head is the file's only Q6_K tensor and stays that way.

`loader_for_config("deepseek_v41")` now reports that the graph is pending instead of "unsupported model type", so the failure points at the right stage.

Tests: four real-file oracles gated `#[ignore]` (shard resolution, config from metadata, every name translates, sparse tensor counts match the source-layer config), plus the unit tests on the name translator.

```
ATLAS_SKIP_BUILD=1 cargo test -p spark-runtime --lib -- --ignored deepseek_v41
```

Not in this PR: the graph. Stage S2 (the CPU reference) and S3 (the single-Spark serving path) follow as stacked PRs. Tracking: #1059 under #970.


TheTom (commented):
Read through S1 against #1059. This is a solid base for everything stacked on it.

What's working well:
- `find_gguf_shards` takes the split keys as the only authority on which files belong to the set; it doesn't glob the directory. The doc comment also explains why summing the tensor counts is the check that catches a lost shard.
- `config_from_gguf` fails closed on the V4.1 metadata. An unknown `expert_gating_func` is refused instead of falling back to softmax, a non-uniform `swiglu_clamp_exp` is refused instead of quietly taking element 0, and a `token_map` length mismatch is an error.
- The `deepseek_v41` arm in `loader_for_config` names exactly what's missing, which is much better than a generic "Unsupported model type".
- The name-translation test is total (1,046 of 1,046, printing any offenders), which is the right shape for that check.

My notes are mostly about the distance between this and the S1 exit criterion in #1059: "R1–R3 fixtures pass through production paths; missing data fails before allocation."

**1. [Blocking] The incomplete-split guard is never exercised.**
`crates/spark-runtime/src/weights/gguf.rs:138` (missing sibling) and `:173` (tensor-count sum) are the guards RST charter C1 is about, but no test makes either of them fire. All four tests in `deepseek_v41_real_file_test.rs` are happy-path and `#[ignore]`d, and they need the 246 GiB checkpoint, so CI never runs this code. The playbook's first session asks for exactly the known-bad control that's missing: take a shard away from a tiny split fixture, expect the specific completeness error, then put it back and confirm the valid case passes again.

Suggestion: small in-CI fixtures (three tiny shards written to a tempdir is enough) covering:
- a missing shard
- a sibling with a different `split.count`
- a sibling whose `split.no` disagrees with its filename
- per-shard counts that don't sum
- a first shard that isn't shard 0

Assert on the error text, so an unrelated parse failure can't pass in place of the guard.

**2. [Should-fix] Tensor extents aren't checked, so a truncated shard gets through.**
`find_gguf_shards` validates the set from header facts only. `GgufFile::parse` (`container.rs:331-400`) confirms that the data section starts before EOF, but not that each tensor's `data_offset + offset + byte_size` fits inside the file. A download that was cut off mid-payload keeps an intact header and the right tensor count, so it passes pre-flight today and only fails later, in the per-tensor `mmap.get`. PRD R2 lists truncated and overlapping tensors, and C1 asks for "a legitimate tensor at the end of a shard and one just beyond the boundary". A header-only extent pass with checked arithmetic would cover it, and duplicate tensor names across shards could be rejected in the same pass.

**3. [Should-fix] The shared-source and candidate fields are declared but never set.**
- **The fields.** `kv_source_layer_ids`, `index_source_layer_ids`, `candidate_source_layer_id`, `candidate_topk_blocks`, `candidate_block_size` and the three engram vocab/embedding fields (`crates/avarok-core/src/config.rs:537-605`) are never assigned by `config_from_gguf`. A V4.1 `ModelConfig` built from the GGUF therefore carries the documented default, "Empty = every layer produces its own KV (V4 behaviour)" (`config.rs:548`). That default conflicts with the comment at `config/gguf.rs:301-303` ("every key here is REQUIRED").
- **The count test.** `deepseek_v41_sparse_tensors_match_source_layer_counts` counts suffix occurrences with the layer index discarded (`deepseek_v41_real_file_test.rs:183-186`) and compares them against a hard-coded 4 and 8. That confirms how many layers carry compressor/indexer tensors, but not which ones (`[2, 8, 14, 20]`).
- **Suggestion.** Make "unknown" distinct from "V4 behaviour" (for example `Option<Vec<usize>>`), or populate the fields from the pinned official config and check them against the layer IDs that actually carry the tensors. PRD 4.3 suggests the second.

**4. [Should-fix] The real-file tests are tied to one machine, and the V4.1 config branch has no CI coverage.**
- **One machine.** `MODEL_DIR` is `/home/rstesiak/models/dsv41-q2k` (`deepseek_v41_real_file_test.rs:17`). PRD §1 asks that a teammate be able to reproduce without another developer's home directories, and an env var named in the ignore reason would fix that.
- **Identity not pinned.** The tests don't assert the pinned identity (HF revision `543d86fd`, or the publisher checksums), so they would pass against a different conversion.
- **No CI coverage.** `crates/avarok-core/src/config/gguf/tests.rs` already has an in-memory `Meta` mock. Adding `get_u64_arr` / `get_f64_arr` to it would let CI cover:
  - a missing required key
  - `expert_gating_func != 4`
  - a non-uniform clamp array
  - a short `compress_ratios`
  - a `token_map` length mismatch
- **Tensor types.** S1 also names "actual tensor types". Nothing asserts the per-role storage type yet: gate/up Q2_K, down Q3_K, engram Q2_K, output Q6_K.

**5. [Question] Is the clamp meant to be optional?**
`swiglu_clamp_exp` is read under `if let Some` (`config/gguf.rs:373`), so when it's absent `swiglu_limit` silently keeps its default, which sits oddly with the all-required comment. The trait doc at `gguf.rs:51` also names `swiglu_clamp_shexp`, but nothing reads it. Is the shared-expert clamp meant to be the same as the routed one?

tbraun96:

PR #1095, opened 2026-09-15
DeepSeek-V4.1 Flash: the CPU reference of the whole graph vs DeepSeek's own model.py (S2)

rrstesiak:
Stacked on #1094 (base main, cumulative head).

A design artifact, not a production path. Before any V4.1 kernel is written, the graph gets a checkable truth: `ds41_tiny_golden.json` is produced by DeepSeek's own `inference/model.py` (deepseek-ai/DeepSeek-V4.1-Flash @ dba1be0a) on a tiny synthetic model that keeps every V4.1 structural feature, with the five tilelang kernels replaced by pure-torch shims written from the tilelang source. No equation is re-derived on the Python side.

The golden carries only outputs. Weights and inputs regenerate bit for bit in Rust from a named, RNG-free filler, so a test never trusts a tensor it did not rebuild. Captures a sub-module needs as input are stored at full resolution; everything else is a prime-strided sample plus a whole-tensor checksum.

Contents, each piece against the golden in the reference's own precision:
- `mod.rs`: the filler, the golden reader, the comparison bar.
- `engram.rs`: `NgramHashState` (DeepSeek's hash, running OR of pad, per-layer multipliers and primes), the fp8 e4m3 table with e8m0 block scales, the wkv projection, the normalised-dot gate. Hash exact on prefill and both decode steps; table gather exact; projection and full output within bf16.
- `hc.rs`: `hc_mixes` with the Sinkhorn split (`pre`, `post`, `comb`), `hc_pre`, `hc_post` with the comb orientation pinned, RMSNorm, and the delayed-mix order (a block's attention uses the mix its predecessor's FFN produced).
- `moe.rs`: sqrt-softplus gate with the correction bias that picks experts but does not scale them, top-k renormalisation, route_scale, clamped SwiGLU experts, one shared expert. Routing indices exact, outputs within bf16.
- `attn.rs`: ratio-0 attention. Low-rank q with RMSNorm, one fp8 latent KV row per token (act_quant with the ue8m0 power-of-two-ceiling scale, bf16-exact round trip), RoPE as an f32 complex multiply written back bf16, the sliding-window ring cache, `sparse_attn` with the attention sink, the inverse rotation, the grouped low-rank output projection. Three regimes run through one cache so the ring seeding and both decode writes are checked, not assumed.
- `compress.rs`: the Compressor (softmax pooling at ratio 2 in f32, plain projection at ratio 1), the Indexer (fp4 e2m1 q and k with e8m0 scales, bf16 rectified scores combined by `weights_proj`, visibility mask, candidate mask), `select_candidate_blocks` with the pinned last block, the shared runtime slots, YaRN RoPE with `compress_rope_theta`, the e4m3-scaled latent quantiser, and `Attention.forward` for any ratio.
- `model.rs`: embed, hc expansion, the one-hot pre-mix, engram before layers 1 and 4, six blocks, collapse, final RMSNorm, f32 logits.

Tier 2, proven on the real file: the Q2_K GGUF's `deepseek41.engram.*` metadata is identical to `engram.py` recomputed from the real tokenizer, and each layer's row count equals the sum of its primes. The production engram needs nothing reverse-engineered.

One finding a production indexer has to decide on: when fewer than `index_topk` candidate positions are visible to a query, the reference returns masked positions, and which ones comes from torch's CPU top-k tie order (libstdc++ nth_element). Those phantom positions are then attended. `torch_cpu_topk_set` emulates that order exactly and is pinned by twelve vectors from torch 2.11 on this box. A kernel either reproduces it or diverges from the reference on exactly those queries.

Tests: 39 in the harness. Every sampled capture for layers 0 to 5 (mixes exact, tensors within bf16, index selections exact), the shared slots after each source layer, and `h_final`, `head_in`, `logits_full` on prefill and both decode steps.

```
ATLAS_SKIP_BUILD=1 cargo test -p spark-model --lib deepseek_v41_ref
```

The last commit is rustfmt only. Tracking: #1059 under #970.


TheTom (commented):
_Scope: this reviews only what #1095 adds on top of #1094 (+4,921 across 19 files). The stack is cumulative, so GitHub's diff here also includes the earlier PRs; those are reviewed on their own PRs._

This is the right oracle to build first, and it's careful work.

The golden comes from DeepSeek's own `inference/model.py` at the pinned checkpoint. It's captured by wrapping the reference's methods: `Block.forward`, `hc_mixes`, `Attention.forward`, `Gate.forward`, `Engram.forward` and `sparse_attn`. So nothing is re-derived. The RNG-free filler is reproduced bit for bit in Rust (`mod.rs:49-81`), which means the golden stores only outputs and every test rebuilds its own weights. Prefill and both decode steps run through one model state, so ring seeding, the compressor state and the shared-slot handoff are actually exercised rather than assumed.

The six-layer fixture maps well onto the PRD 3.1 table:
- **Window and compression.** Layers 0–1 are window-only, 2–3 are ratio 2, and 4–5 are ratio 1.
- **Engram.** Engram sits on layers 1 and 4.
- **KV producers and consumers.** KV producers are 2 and 4, with consumers 3 and 5. Layer 5 reads layer 4's slot, not layer 2's, so a stale-producer bug would show up.
- **Index-only producer.** Layer 5 is an index producer that isn't a KV producer, like production layers 24–36.
- **Forced pruning.** The candidate prefilter is forced to prune: 3 blocks of 2 positions, out of up to 14 visible. That's the pruning control charter C7 asks for.

I checked three of the easy-to-miss paths against the pinned `model.py`, and all three are right:
- **Delayed hyper-connections.**
  - Attention collapses with the incoming `pre_mix` (`model.rs:399`).
  - The FFN uses `attn_pre` (`:455`), and `ffn_pre` is handed forward (`:475`).
  - The final collapse uses the last FFN's mix with no output HC head (`:502`). This matches `model.py:983-1000,1268`.
- **Window and compressed KV are concatenated, not deduplicated** (`compress.rs:680-687` against `model.py:777-778`).
- **Pooling is a per-channel softmax over the grouped tokens,** and a partial group is held in state rather than published (`compress.rs:173-212`).

Index selections are compared exactly over the whole tensor, and `shared_expert_alone_is_not_the_answer` is a nice sensitivity control. My notes below are about making the oracle reproducible by someone else, showing it can fail, and one path the fixture doesn't reach yet.

**1. [Should-fix] The golden and the Tier-2 check can't be regenerated from the repo yet.**
The golden generator has two gaps:
- `gen_ds41_tiny_golden.py:39` hard-codes `REF = "/home/rstesiak/..."`.
- `dba1be0a` appears only as a string literal (`:374`), so nothing confirms the imported files really are that revision.

`gen_ds41_engram_hash_check.py:21-24` has three:
- It imports `gguf_manifest` from a `tools/` directory outside the repo.
- It reads the GGUF from a local path, with no revision or hash.
- Because of both, the result that `engram.py` recomputes identically can't be reproduced by a second owner (PRD R10).

The twelve tie-order vectors in `compress_tests.rs:11-219` are recorded from local torch, and no script for recording them is committed.

Suggestions:
- Take `REF` and `GGUF` from arguments or environment variables.
- Assert sha256 of `model.py`, `engram.py` and `kernel.py` before importing, and write those hashes into the fixture. At `dba1be0a` I get:
  - `model.py`: `4e9ae236…6f65`
  - `engram.py`: `11f35ecb…3897`
  - `kernel.py`: `1236c350…2455`
- Pin the GGUF to the PRD's `543d86fd` revision plus a shard hash.
- Vendor or inline the metadata reader.
- Commit the few lines that print the torch `topk` vectors.

**2. [Should-fix] One intermediate quantization differs between the `sparse_attn` shim and the kernel.**
- **Kernel.** The tilelang kernel casts the exponentiated scores to bf16 before the PV GEMM (`kernel.py:339,377,380`; `acc_s_cast` is `BF16`).
- **Shim and Rust.** The shim keeps the probabilities in fp32 (`ds41_ref_shims.py:101-103`), and so does `attn.rs` `sparse_attn`, which follows the shim.
- **Why it matters.** Because the shim and the Rust port share the difference, neither can see it. It's small, but it's exactly the kind of intermediate quantization PRD 3.2 asks us to "record … and any deliberate differences".
- **Suggestion.** Note it as a known deviation in the shim docstring and in `mod.rs`, or model it. The kernel rounds `exp(s - running_max)` per KV block, so exact emulation depends on the block size.

**3. [Should-fix] No model-level known-bad control, and the float check is looser than its comment suggests.**
- **The existing control.** `comparator_catches_a_planted_defect` (`tests.rs:182-191`) exercises `assert_close` at 1e-6, but not `check_capture` at the tolerances the graph tests actually use.
- **How `check_capture` compares.** It checks 12 prime-strided samples plus a signed checksum bounded at `BF16_CK_REL = 1/128` of the index-weighted magnitude (`mod.rs:302,306-327`).
- **What slips through.** For `prefill12.logits_full`, a single wrong logit near the end of about ±1.6 (roughly 3× the mean |logit|) stays inside that bound unless it lands on a sampled index. So the comment at `mod.rs:307-308` ("a wrong element fails") doesn't hold for isolated errors.
- **Coverage.** Whole-tensor mutations would very likely fail on the samples, and integer selections are compared exactly, so selection errors are well covered. The gap is localized float defects.
- **Suggestion.** Add a test-only mutation switch to `forward` for four known-bad variants:
  - pre-mix applied at the current sublayer
  - pooling softmax over channels
  - window and compressed KV deduplicated
  - a partial group published

  Add one test per variant that asserts the golden check fails. It's also worth tightening the float check with per-row checksums or more full-resolution captures; the golden is only 297 KB.

**4. [Should-fix] The prefill partial-group path is never exercised.**
- **The gap.** `PREFILL = 12` (`gen_ds41_tiny_golden.py:81`) was chosen for the window cutoff, but 12 % 2 = 0. So the ratio-2 compressor's prefill remainder branch (`compress.rs:188-192`, `model.py:470-472`) never runs, and neither does the handoff of that remainder into the first decode step.
- **The decode side is covered.** `decode12` holds its token back and `decode13` publishes the completed group.
- **Suggestion.** Charter C6 names incomplete ratio-two groups explicitly, so a regime set with an odd prefill (11 or 13) followed by decode would close it.

**5. [Question] Could the tie-order finding be recorded as a decision?**
The emulation is faithful: `model.py:579-580` uses torch `topk(sorted=False)` and keeps any pick below `compress_lens`. The golden confirms phantom positions are attended. At `decode12`, `L5.sa_topk_idxs` includes compressed position 8 while `shared.L4.candidates[8] = 0`.

As far as I can tell, phantoms can't occur under the product cap:
- At ratio 1, 2,048 blocks × 8 covers every visible position.
- At prefill, the remaining −inf entries are future positions, which map to −1.

What does remain in scope is exact finite ties at the k-th cutoff, and a GPU top-k will order those differently.

Is the intent (a) that the production indexer needn't reproduce the host's order, since the cap makes phantoms unreachable (ideally backed by a runtime assertion), and (b) that kernel-vs-reference index comparisons use set equality with score margins reported at the cutoff? If so, a short note in `compress.rs` or on #1059 would close it.

(`compress.rs` and `model.rs` are over 500 lines; #1100 already covers the split.)

tbraun96:

PR #1096, opened 2026-09-15
DeepSeek-V4.1 Flash: S3 sizing on the real Q2_K shards + GPU Q2_K / Q3_K dequant

rrstesiak:
Stacked on #1095 (base main, cumulative head).

Two things the single-Spark serving path is designed from, both measured on the seven shards after `sha256sum -c` passed on all of them.

**Sizing (ignored real-file tests).**
- One routed expert: 12.22 MiB on disk (gate/up Q2_K, down Q3_K; 46,080 blocks each), 35.4 M elements. CPU dequant to BF16 is 851 ms single-thread, which is 204 s per token at 6 experts x 40 layers. Routed experts therefore stay quantized to the GPU and are consumed as raw blocks.
- One token's engram rows: one Q2_K block (84 B) per row; 24 random rows gather and dequant in 13 ms. The 60 GiB of tables are read on demand and never cached.
- mmap page-in per expert is 0.33 GB/s cold and cached alike, so it is the fault path; the serving path reads slices with pread.

```
ATLAS_SKIP_BUILD=1 cargo test -p spark-runtime -- --ignored deepseek_v41_stream --nocapture
```

**GPU Q2_K and Q3_K dequant to BF16.** Two arms added to `kernels/gb10/common/dequant_gguf_bf16.cu` beside Q8_0 / Q4_K / Q6_K, each a line-for-line port of the CPU decoder (element order and op order preserved under `--fmad=false`), with the host wrappers, dispatch on ggml ids 10 and 11, and `supports()` reporting both, so the loader routes every Q2_K / Q3_K tensor through the device. This is what dequantizes the resident V4.1 tensors once at load: attention, embed and the engram projections are Q2_K, the routed down projections Q3_K.

Validated on GB10 against the CPU oracle: Q2_K and Q3_K 9472/9472 elements identical, the existing five types unchanged.

Tracking: #1059 under #970.


TheTom (commented):
_Scope: this reviews only what #1096 adds on top of #1095 (+342 across 5 files). The stack is cumulative, so GitHub's diff here also includes the earlier PRs; those are reviewed on their own PRs._

The increment over #1095 is small and clean.

I read the two new CUDA arms (`dequant_q2_k_to_bf16` and `dequant_q3_k_to_bf16`, `kernels/gb10/common/dequant_gguf_bf16.cu:137,174`) against ggml's `dequantize_row_q2_K` / `dequantize_row_q3_K`, and they match, including:
- the block layout
- the scale index `n*8 + 2j + sub`
- the kmask1/kmask2 6-bit scale unpack
- the high-bit mask `1 << (4n + j)` carrying across halves
- hmask not advancing per half

Extending `gguf_gpu_validate_matches_cpu_oracle` with random-byte blocks over 37 blocks (`gpu_validate_test.rs:110-113`) is a sensitive check for any indexing mistake on the GPU side.

The "bit-identical" claim also holds beyond the test's power-of-two scales:
- **Exact products.** Every product is exact in f32: an fp16 scale × a 4- or 6-bit sub-scale × a 2- or 3-bit code.
- **One rounding.** FMA contraction and the single final rounding therefore can't diverge.
- **Same rounding mode.** Both sides round to bf16 with round-to-nearest-even.

The exact expert-size assertion (`46_080 × (84+84+110)`) is a good level-A oracle as well.

**1. [Should-fix] The CPU decoders the GPU is held to don't have an independent check.**
`dequant_cpu/tests.rs` has known-value single-block tests for Q8_0, Q4_1, Q4_K, Q6_K and Q2_0, but none for Q2_K or Q3_K (`blocks.rs:203,235`). GPU == CPU is a same-author transcription check, so a shared misreading of the format would pass it, and PRD R3 asks for "independent decode". I think both decoders are correct, so this is about locking that in rather than a defect.

Suggestions:
- Add `q2_k_single_block` and `q3_k_single_block` in the style of `q4_k_single_block`.
- Include a Q3_K case with the hmask bit both set and unset, and scales on both sides of 32.
- Optionally, add one block whose expected values come from gguf-py's numpy dequantizer, plus the C2 control ("decode a Q3_K fixture with Q2_K rules") to show the check can fail.

**2. [Should-fix] The sizing numbers should state their conditions.**
- **Cache state.** `deepseek_v41_stream_bench_test.rs:94` measures first touch, but nothing guarantees a cold page cache. The comment at `:88` only holds on the first run after boot or a cache drop; on a rerun, the page-in rate is the cached rate. Evicting the range first (`posix_fadvise(DONTNEED)`) or reporting residency would pin it down.
- **Threading.** The 851 ms dequant is single-threaded (`:142`), so the 204 s-per-token figure quoted in the PR and in the `kquant_moe.cu` header is a single-thread number. The conclusion that routed experts should stay quantized very likely still holds with a thread pool. Labelling the figure as single-thread wherever it's quoted would help.
- **Why it matters.** PRD §9 asks that cache-state assumptions be kept with the metrics.

**3. [Should-fix, small] The same hard-coded `/home/rstesiak/...` path** is at `deepseek_v41_stream_bench_test.rs:23`. The env-var fix from #1094 applies here too.

tbraun96:

PR #1097, opened 2026-09-15
DeepSeek-V4.1 Flash: K-quant expert kernels on raw Q2_K / Q3_K blocks (decode mmvq + prefill MMQ)

rrstesiak:
Stacked on #1096 (base main, cumulative head).

The kernels that make streaming experts serve at all. Routed experts are consumed as the GGUF stores them (gate/up Q2_K, down Q3_K): no BF16 expansion, no re-quantization.

`kernels/gb10/deepseek-v4-flash/nvfp4/kquant_moe.cu`:
- **Decode** (M <= 8, the verify tier included): `kquant_mmvq_q2_k` / `kquant_mmvq_q3_k`, the llama.cpp mmvq form. One block of four warps per output row, activations in plain `block_q8_1` from `kquant_q8_1_rows_bf16`, the vendored `vec_dot_q{2,3}_K_q8_1` on the packed codes (dp4a, no unpack to bytes).
- **Prefill** (M up to the 2048 cap): `atlas_q{2,3}_k_mmq128_{nc,wc}`, the vendored MMQ tensor-core engine with the compile-time type switched, as `q2_0_mmq.cu` does for Q2_0, with the activation quantiser in each type's q8_1 layout (D2S6 for Q2_K, D4 for Q3_K). Q2_K tile smem 70,144 B, Q3_K 61,952 B.

Host launches in `ops/kquant_mmq.rs` (sizes, q8_1 rows, mmvq, mmq_gemm). spark-runtime's CPU decoders (`dequant_cpu`) are exported so kernel oracles can use the loader's own reference.

Validated on GB10 against a CPU oracle (CPU dequant, q8_1 emulated in Rust exactly as the kernels quantize, f32 GEMM; tolerance 2^-7 max|C| + 1e-3) at N=320, K=1024: Q2_K and Q3_K, mmvq M=1 and M=5, MMQ M=200, six for six within tolerance, worst under a third of it.

```
cargo test -p spark-model --release kquant -- --ignored --nocapture
```

Tracking: #1059 under #970. Next in the stack: expert streaming and the engram row reader, the engram module and shared compressed attention on the GPU, then the graph, loader and first generate.


TheTom (commented):
_Scope: this reviews only what #1097 adds on top of #1096 (+611 across 8 files). The stack is cumulative, so GitHub's diff here also includes the earlier PRs; those are reviewed on their own PRs._

Reusing the vendored ggml MMQ engine with the quantization type switched is a good call. It keeps the tensor-core path on well-exercised code instead of a new GEMM. Several other things are solid:
- **Runtime preconditions.** The launch wrappers enforce them at runtime (`m` in 1..=8 for mmvq, `k % 256 == 0` for both; `kquant_mmq.rs:90-94,124-127`), so an unsupported shape is rejected rather than silently truncated.
- **Matches ggml.** The mmvq loop matches ggml's `mul_mat_vec_q` indexing (`qi = 16`, `vdr = 1`, 8 blocks per iteration). The shared-memory constants reproduce `mmq_get_nbytes_shared` for the MMA path, and `sizes_match_the_block_geometry` pins them.
- **Oracle design.** It dequantizes with the CPU decoders and emulates the activation q8_1 quantization in Rust, which cleanly isolates the GEMM.

My notes are about how much the six configurations can actually reject.

**1. [Blocking] `assert_close` passes when the output is NaN.**
At `kquant_mmq_tests.rs:122-127`:
- `d = (g - w).abs()` is NaN for a NaN output.
- `d > worst` is false for NaN, so the element is skipped.
- A kernel that wrote NaN everywhere would report `worst = 0` and pass.

PRD §9's oracle ladder says to "inspect NaNs/infinities separately". Suggestion: assert `got.iter().all(|v| v.is_finite())` first, and write the bound as `!(d <= tol)` so NaN fails.

**2. [Should-fix] The tested shapes don't reach the code production runs.**
Every configuration uses `N = 320`, `K = 1024` (`:22-23`), which leaves three gaps:
- **mmvq.** K = 1024 is 4 super-blocks, fewer than the 8 blocks per iteration, so the `kbx += blocks_per_iter` stride loop never takes a second pass. Production K = 5120 (20 blocks) and K = 2304 (9 blocks, partial last pass) depend on it.
- **MMQ.** N = 320 selects only the `_wc` kernel. The real expert shapes (N = 2304 and 5120) are multiples of 128 and dispatch to `_nc`, which no test launches.
- **Decode M.** M = 8, the top of the verify tier, isn't covered; only M = 1 and 5 are (`:202`).

Suggestion: add the real expert shapes (Q2_K 2304×5120, Q3_K 5120×2304) at M ∈ {1, 8} for decode and M ∈ {200, a full 128-multiple} for prefill, plus one K whose block count isn't a multiple of 8.

**3. [Should-fix] The tolerance has no written error model and no known-bad control.**
- **The formula doesn't match its comment.** The comment at `:118` says "half an ulp at the max magnitude", but `tol = max|want| · 2⁻⁷ + 1e-3` (`:119`) is one full bf16 ulp at the global maximum. It's applied as an absolute bound to every element, so small-magnitude outputs are checked far more loosely than large ones.
- **Differences absorbed, not accounted for.** A couple of expected differences are absorbed by the tolerance rather than modelled:
  - The kernel stores `d` and the block sum as fp16 (`make_half2`).
  - In the D2S6 layout, the Q2_K min term uses sums of the unquantized activations (`q4k_vendor/quantize_impl.cuh:344`), while the oracle sums the quantized values.
- **What the PRD asks for.** PRD §3.2/§4 asks for the maximum observed discrepancy on valid controls, a stated rationale, and a check that the tolerance rejects a deliberately wrong operation.
- **Suggestion.**
  - Put the printed worst case per configuration in the PR.
  - Add a per-element relative metric with a near-zero floor.
  - Add one negative control that must fail, for example Q3_K bytes fed through the Q2_K kernel, the weight pointer offset by one super-block, or the D4 quantizer fed to the Q2_K MMQ.

**4. [Question] Were power-of-two scales deliberate?**
`SCALES_F16` uses only powers of two (`:24`), the same as the dequant test. That's harmless for correctness on the dot-product path, but it understates the rounding real scales produce.

tbraun96:

PR #1098, opened 2026-09-15
DeepSeek-V4.1 Flash: expert streaming into a device-visible LRU, the engram row reader, then engram and shared compressed attention on the GPU

rrstesiak:
Stacked on #1097 (base main, cumulative head). DRAFT: the first cut is in, the rest of the scope is claimed below and lands on this PR.

**Landed: expert streaming and the engram row reader** (`spark-runtime/src/weights/gguf/expert_stream.rs`, `expert_lru.rs`).

The routed experts of the Q2_K checkpoint are 40 x 384 x 12.22 MiB. They never fit and, per #1096, they are never expanded and never page-faulted through the mmap. Instead:
- `ShardFiles` opens every shard once, parses the header through the loader's own reader and drops the mmap. `ExpertSliceMap` locates expert `e` of layer `l` in each of the three stacked tensors (`blk.N.ffn_{gate,up,down}_exps.weight`) and reads one expert into one slot (`gate | up | down`) by `pread`.
- `ExpertLru`: fixed-size slots in one page-locked host arena the GPU addresses through its `cuMemHostGetDevicePointer` alias (`PinnedArena`; on GB10 both addresses name the same unified memory), evicted least-recently-used. `begin_token` opens an epoch and every slot fetched in it is pinned, so a token's 240-expert working set can never evict itself. `fetch_many` resolves a whole token, reads the misses on scoped threads (positional reads share no file offset) and dedups repeated experts. The arena is owned by the caller; the unit tests back it with a `Vec<u8>`.
- `EngramRowReader`: the two ~30 GiB engram tables, one Q2_K block (84 B) per row, read by row id on demand. Nothing is cached; a token is 48 reads.

Oracle, on the seven shards: bytes identical to the loader's mmap path for experts on shards 0, 1 and 6 (first layer / expert 0, layer 3 / expert 200, last layer / expert 383), through eviction and re-fetch in a two-slot cache, and for 24 random rows of each engram table. One token's gather of 240 experts (2.86 GiB): 1007 ms on one reader thread, 159 ms on eight, page cache warm; the same token replayed is 240/240 hits and reads nothing. The cold single-thread pread rate measured in #1096 is 4.5 GB/s, so the eight-thread cold number is the one the serve log will report.

Tests: six CPU unit tests on the cache policy (miss/hit, LRU order under touch, epoch pinning, parallel gather with dedup, a failed read maps nothing) and three ignored real-file oracles.

```
ATLAS_SKIP_BUILD=1 cargo test -p spark-runtime --lib expert_stream_tests
ATLAS_SKIP_BUILD=1 cargo test -p spark-runtime --release -- --ignored deepseek_v41_stream_oracle --nocapture
```

**Claimed, landing on this PR:**
- The engram module on the GPU: hash from the GGUF metadata (#1094), rows from the reader above, the `wkv` GEMM, the normalised-dot gate, written into the hc stream before layers 1 and 14. Oracle: `deepseek_v41_ref::engram` on the tiny graph (#1095), then row-level parity on the real tables.
- The shared compressed attention runtime on the GPU: the window ring, the Compressor (ratio 2 pooling on the four `kv_source_layer_ids`, ratio 1 plain), the Indexer (fp4 q/k with e8m0 scales, candidate blocks, the top-k tie decision), the shared slots four layers write and forty read, YaRN with `compress_rope_theta`. Spec: `deepseek_v41_ref::compress`. Oracle: the tiny-graph captures through the production path, then real-shape parity per stage.

Tracking: #1059 under #970.


rrstesiak:
Engram on the GPU has landed on this branch (60bf4250d): `layers/engram_v41.rs` and `kernels/gb10/deepseek-v4-flash/nvfp4/engram_v41.cu`.

- Hash: the reference n-gram hash on the CPU from the GGUF's own tables (`token_map` read through a new signed-array accessor), integer-exact against the CPU reference and against the golden's `engram_hashes` on prefill and both decode steps.
- Rows: raw Q2_K blocks from the row reader dequantised on the device straight into the `[tokens * cols, head_dim]` bf16 input; 48 random rows of each real 30 GiB table are bit-identical to the CPU decoder.
- Projection and gate: the dense bf16 GEMM into `[tokens, dim * (hc + 1)]`, then `engram_v41_gate` (normalised dot of stream against key, signed-sqrt sigmoid, `h += gate * value`) in place on the FP32 mHC highway. Against the golden's `engram_out`: within bf16 on both engram layers in all three regimes.

```
ATLAS_SKIP_BUILD=1 cargo test -p spark-model --lib engram_v41
cargo test -p spark-model --release engram_v41 -- --ignored --nocapture
```

Remaining on this PR: the shared compressed attention runtime.

rrstesiak:
The shared compressed attention runtime has landed on this branch, which completes the scope of this PR: `layers/attn_v41.rs` and `kernels/gb10/deepseek-v4-flash/nvfp4/attn_v41.cu`.

One forward serves every V4.1 layer (ratio 0, 1 and 2): the low-rank q path, the fp8 latent kv row, the sliding-window ring, the compressor (f32 projections and per-dim softmax pooling at ratio 2, with the partial-group state across decode steps), the indexer (fp4 index keys into the shared `index_k` cache on the four kv sources, fp4 queries, bf16-exact rectified head scores, candidate blocks on the candidate source), the latent written fp4 into the shared `compress_kv` cache after the indexer consumed it, sparse attention over window rows plus compressed rows with the sink in the denominator, the inverse rotation, and the grouped low-rank output projection. The two top-k selections run on the CPU through the reference's `torch_cpu_topk_set`, so the tie order among equal scores is the reference's and the index selection is exact rather than approximately right.

Oracle: the CPU reference model (#1095, 39/39 against DeepSeek's golden) run for prefill12, decode12 and decode13 supplies every layer's true `attn_in`; the GPU layer reproduces `sa_topk_idxs` exactly and `sa_q`, `sa_kv`, `sa_o`, `attn_out` within bf16 on all six layers in all three regimes, with the window rings, compressor state, both caches and the shared slots carried across regimes; `compress_kv`, `index_k` and `topk_idxs` match after every source layer.

```
cargo test -p spark-model --release attn_v41 -- --ignored --nocapture
```

What remains for the first generate is #1099: the loader, the block assembly on the existing MoE router and mHC sites, and these modules wired in.

TheTom (commented):
_Scope: this reviews only what #1098 adds on top of #1097 (+3,834 across 18 files). The stack is cumulative, so GitHub's diff here also includes the earlier PRs; those are reviewed on their own PRs._

This increment is careful work, and several pieces stand out.

**Streaming.** The streaming design keeps routed experts in their GGUF K-quant blocks and preads them into a fixed arena instead of expanding them. That meets the PRD's bounded-memory requirement directly. The supporting tests are good:
- **Byte-identity check.** `deepseek_v41_stream_oracle_test.rs` is a good instrument check. It compares cache bytes against the mmap on the real shards across the first, a middle and the last MoE layer, evicts at a two-slot capacity, re-fetches, and compares again.
- **Known-bad controls.** The LRU unit tests include real known-bad controls: a poisoned source (`a_failed_read_leaves_nothing_mapped`) and an over-capacity fetch inside one epoch.
- **Read failures surface as errors.** `pread` goes through `read_exact_at`, which retries on EINTR and turns a short read into an error. A failed read becomes an `Err` and the slot is unmapped, rather than being served as garbage weights.

**Attention.**
- `attn_v41_pool` (`attn_v41.cu:227-243`) normalizes over the grouped tokens for each channel, which is the axis the PRD requires.
- The ratio-2 compressor publishes only when `(start_pos + 1) % ratio == 0` (`attn_v41.rs:571-573`), so an incomplete group is never published.

**Engram.** Row ids are bounds-checked before any read (`expert_stream.rs:436-440`), and the oracle test confirms an out-of-range id is refused.

**1. [Blocking] `fetch_many` can leave slots mapped but unread when the cache overflows.**
`expert_lru.rs:330-340` calls `self.assign(key)?` once per miss, and `assign` inserts the key into `map` before any bytes are read (`:266-272`). If `take_victim` fails partway through (`:258-262`, "smaller than one token's working set"):
- the `?` returns early;
- every miss already assigned in that call stays in `map`, pinned to the current epoch, and still holds the previous occupant's bytes;
- the next `fetch`/`fetch_many` for those keys is a hit (`:301-304`, `:331-335`), so it returns another expert's weights with no error.

`a_token_cannot_evict_its_own_working_set` (`expert_stream_tests.rs:155-173`) covers the boundary only through single `fetch`, where the failure maps nothing.

Suggestions:
- On an assign failure in `fetch_many`, `unmap` every slot already assigned in this call before returning.
- Add a unit test: a 2-slot cache, `fetch_many` of 3 distinct keys, then assert that none of the three keys is resident.
- Reject a cache smaller than one token's working set (6 × the number of MoE layers) at construction, so a too-small cache fails at load rather than on the first token.

**2. [Should-fix] Pinning is a host-side epoch, and nothing yet ties it to the GPU having finished with the previous token.**
- **The gap.** `begin_token` (`expert_lru.rs:190-192`) only bumps a counter. From that moment every slot of the previous token is a legal victim, and the next `fetch_many` may pread into it while the previous token's MoE kernels could still be reading it through the device alias.
- **Why it matters.** That's the "wrong producer lifetime" risk from the PRD, applied to expert bytes. Nothing calls this yet in this PR, so it's a contract to settle before it's wired in, not a live bug.
- **Suggestion.** Make it enforceable, for example `begin_token(&mut self, done: FenceOrEvent)` or a documented requirement to synchronize the stream first, backed by a debug assertion.
- **Related.** `PinnedArena::alloc` isn't exercised by a test here. The oracle and unit tests use a heap `Vec`, so the device-visible part isn't backed by evidence in this increment yet.

**3. [Question] Which index-key cache does a KV-source layer read on a step where its ratio-2 group is incomplete?**
- **The asymmetry.** The compressed KV cache is re-pointed to the layer's own cache on every step (`attn_v41.rs:861`). The index-key cache is re-pointed only when the layer publishes (`:633-661`). On a non-publishing step, the indexer reads `shared.index_k` (`:662-664`), which holds whatever the last KV source published.
- **Fixture versus real model.** In the fixture's `decode12`, layer 2 (ratio 2, KV and index source) doesn't publish, so it scores against layer 4's ratio-1 keys from prefill. On the real model, layer 2 would score against layer 20's keys from the previous step, and layers 3–7 consume that selection.
- **The reference agrees, but can't tell.** The CPU reference has the same asymmetry (`deepseek_v41_ref/compress.rs:406-425`, `:435`), so the two agree by construction. The golden can't discriminate the case either: with `index_topk = 6` and `width = 13 / 2 = 6`, top-k selects every position regardless of the key contents.
- **What would settle it.** Could you confirm against `inference/model.py` which cache the source layer's indexer reads here? Regenerating the golden with `index_topk` below the width on a mid-group decode step would settle it with DeepSeek's code as the oracle. It would also give the ratio-2 indexer a selection that can actually fail, which is the stale-shared-state control from PRD §3.1.

**4. [Should-fix] `forward` accepts `start_pos > 0` with `m > 1` and quietly handles only the first token.**
The guards at `attn_v41.rs:791-801` check only `1 <= m <= max_tokens` and `start_pos + m <= max_seq`. On the decode branch, only one row is handled:
- the window ring copies one row (`:847-851`);
- the compressor stores one row (`:566-575`);
- `group_positions` returns a single position (`:600-606`).

So a chunked prefill or a multi-token verify step would drop rows without an error. The CPU reference panics in that situation (`compress.rs:612-613`), so the GPU path is currently less strict than its oracle. Suggestion: `ensure!(start_pos == 0 || m == 1, ...)` at the top of `forward`, plus a test for the refusal.

**5. [Should-fix] Shared state has no identity to check against, and `compress_len` is written but never read.**
- **Unused field.** `SharedV41.compress_len` is updated with `max(...)` (`:912`) and never read. `rows_b_len` comes from the consumer's own ratio (`:858`, `:919`), and because the field only grows, it would carry over from one request into the next.
- **Unchecked consumers.** Consumers take `shared.topk_idxs` and `shared.topk` as found (`:866-867`). The only check is `cidx.len() == m * ctopk` (`:920-924`), and once `topk` saturates at `index_topk` during decode, that check can't catch a selection left over from an earlier step.
- **Suggestion.** Stamp the shared slots with `(start_pos, m)` and the producing layer when they're published, and assert the stamp in consumers. Then either use `compress_len` as a bound, resetting it at `start_pos == 0`, or remove it.

(`attn_v41.rs` is 1,059 lines; #1100 already covers the split.)

tbraun96:

PR #1099, opened 2026-09-15
DeepSeek-V4.1 Flash: the graph, the loader and the first generate on one Spark

rrstesiak:
Stacked on #1098 (base main, cumulative head). DRAFT claim: the kernel target is in, the graph lands on this PR.

**Landed:** `kernels/gb10/deepseek-v4.1-flash/MODEL.toml`. A kernel target for `model_type = "deepseek_v41"` that reuses the deepseek-v4-flash kernel tree through `kernel_source` (the qwen3.8-27b precedent), with every architecture number read from the Q2_K GGUF's metadata by the S1 ingestion rather than from an HF config. V4.1-only kernels join that tree so one kernel set serves both targets.

**Claimed, landing on this PR: the graph, the loader and the first generate on one Spark.**
- `DeepSeekV41WeightLoader`: the resident 2.9 GiB (attention Q2_K, embed Q2_K, the Q6_K head, shared experts, norms) dequantized once at load through the GPU arms in #1096; the routed experts left on disk and served by the streamer in the stacked PR; the vision routing bias `exp_probs_b_vl.bias` mapped but not fed to the text router; the `ffn.experts.{e}.w{1,2,3}` naming of the V4 loaders reconciled with the stacked-tensor names of #1094.
- The factory arm replacing the `deepseek_v41` bail in `factory.rs`.
- Block assembly on the existing MoE router (sqrt-softplus, correction bias) and kernels, the mHC sites, and MLA, with the K-quant decode GEMV and prefill MMQ (#1097) as the routed-expert path, the engram module and the shared compressed attention runtime from the stacked PR wired in, and the 2048-token cap everywhere.
- A real generate on Blackbird from the seven Q2_K shards, with the receipt.

Oracle: `logits_full` parity on the tiny graph (#1095) through the production path, then a known-answer prompt on the real shards.

What this is: a "runs at all on the box we sell" proof for a 475 GiB model on one 128 GiB Spark, streaming 2.86 GiB of experts per token. The ceiling is set by the per-token disk traffic measured in #1096, and the number will be reported as measured, cold, with the reproduction key.

Tracking: #1059 under #970.


rrstesiak:
First piece of the graph on this branch (e5ba6f517): the MoE on streamed, still-quantized experts, `layers/moe_v41.rs` and `kernels/gb10/deepseek-v4-flash/nvfp4/moe_v41.cu`.

- Routing: the logits as an f32-accumulated GEMM on the GPU, then the reference's `Gate.forward` on the CPU (sqrt-softplus, top-k by score plus the correction bias, unbiased weights renormalised and scaled). The selection stays the reference's, tie-breaking included.
- Routed experts: the batch's experts gathered once through the cache from #1098, then per (token, expert) the K-quant GEMVs from #1097 on the raw blocks in the cache slots: gate and up (Q2_K) on the q8_1 input, the clamped SwiGLU times the routing weight, q8_1 again, down (Q3_K), accumulated in f32. The shared expert is resident bf16 on the dense GEMM.
- Oracle on GB10 with synthetic Q2_K / Q3_K experts in a real pinned, device-visible arena: routing indices identical to the reference gate; the output within the K-quant bound of the production numerics emulated on the CPU (the loader's decoders, q8_1 emulation, bf16 GEMMs, the reference's SwiGLU and accumulation); one read per distinct expert.

Decode-first: one GEMV per (token, expert). The MMQ prefill arm from #1097 is wired later, once the graph runs end to end.

```
cargo test -p spark-model --release moe_v41 -- --ignored --nocapture
```

rrstesiak:
The graph, the loader and the first generate have landed on this branch (92cd952c4). DeepSeek-V4.1 Flash Q2_K runs on one Spark.

**What landed**
- The GGUF loader reads every shard of the split file and records the routed expert stacks and the engram tables as deferred, with their on-disk location, instead of dequantizing them.
- `weight_loader/deepseek_v41.rs` assembles the forty layers: resident tensors in bf16, the experts served from the page-locked device-visible cache by pread, the engram rows by id. Source-layer sets come from which layers ship compressor and indexer tensors; the candidate settings default to the published config.
- `layers/deepseek_v41_layer.rs` is one block on the generic model in the reference's order, with V4.1's delayed hyper-connection mixes (`hc_v41.cu`), the vanilla RMSNorm, `AttnV41`, `MoeV41`, engram on layers 1 and 14, and the final collapse with the last ffn pre (no learned head). One sequence at a time; CUDA-graph capture and multi-sequence decode are declined through the layer hooks.

**Oracle for the assembly:** layer 0 of the real model against the CPU reference on the same five embedding rows, every stage within bf16 (`cargo test -p spark-model --release layer0_matches -- --ignored --nocapture`). That test is what found the one integration bug: the shared `rms_norm` kernel applies the zero-centered `(1 + w)`, and V4.1's norm weights are plain, so V4.1 takes the vanilla twin for the layer norms and the final norm.

**The run** (Blackbird, one GB10, page cache warm, greedy, no chat template, not a benchmark and not a record): `--check-kernels` passes with 64 lookups, 0 unresolved, 1 declared absent. "1, 2, 3, 4, 5, 6, 7," continues "8, 9, 10, 11, 12,". Prose is coherent at the quality a 2-bit quant allows; a longer instruction-style prompt degenerates into a repeated token, as greedy 2-bit decoding does. Single-stream decode 2.3 to 3.0 tokens per second, TTFT 1.3 to 2.0 s on 5 to 8 prompt tokens, inside the 1.5 to 3 tok/s ceiling the S3 measurements set (#1096). The receipt with the reproduction key is in my records ledger.

Not in this PR: the MMQ prefill arm (prefill runs the decode GEMV per token), chunked prefill, multi-sequence decode, CUDA graphs, a cold-cache number, and a quality evaluation.

rrstesiak:
Pushed: the decode fix and everything held behind it since `f86163c17`.

**Competing numbers first.** DeepSeek-V4.1 Flash on Spark hardware today: Turney's two-Spark EXL3 run at roughly 30 to 32 tok/s with the engram dummied and DSpark drafting; the vcruz305 llama.cpp recipe at 2.3 to 2.5 tok/s on one GB10 on the CPU backend. This PR is one Spark, native Atlas, engram on, no speculation, Q2_K from the seven shards, and it is the only one of the three with a faithful engram.

**What was wrong.** Decode drifted from teacher-forced prefill in every configuration and two greedy runs diverged by token three. A per-layer checksum trace put the first divergence at layer 0 MoE output with identical attention output, identical MoE input and identical routing. The cause was a data race in the single-token expert path: `routed_m1` finished with one `moe_v41_scatter_add` launch, one block per expert row, and at m=1 every row is token 0, so eight blocks did a non-atomic f32 add into the same accumulator row. The kernel's own comment states the distinct-rows precondition it was violating. Lost adds flattened the decode distribution, which is what the loops were.

**The fix.** `moe_v41_sum_rows` adds the expert rows into the token row in plan order, the same rounding sequence as the per-expert loop, and replaces that launch. `4a859ae5b`.

**Verified on the box, binary at `4a859ae5b`:**
- four greedy 40-token runs bit-identical, layer-0 MoE hash `e47cac75b1326e37` on all four, matching the loop path
- decode versus teacher-forced prefill over 100 tokens: 97 of 100 top-1 agree; the other three are near-ties within 0.15 nats with top-20 overlap 20 of 20 (before the fix: about 50 disagreements from position 1)
- chat endpoint, 300 tokens, MinHeap x3 and Volvo x3: byte-identical across repeats, correct, no loop; 10.6 to 10.9 tok/s warm, TTFT 1.3 to 2.0 s
- clippy clean; the mirror test for the Hopper and B200 kernel trees unchanged (symlinks)

**Also on the branch since `f86163c17`:** the rebrand's crate and backend names on the loader and tests; the Hopper (sm_90a) and B200 (sm_100a) targets with the V4.1 kernels mirrored; `release_state` freeing the per-sequence attention buffers; the loop watchdog declining its rollback on a layer that cannot rewind, which was the request-2 corruption; four diagnostic switches, all default off (`ATLAS_DS41_TRACE`, `ATLAS_DS41_PREFILL_GEMV`, `ATLAS_DS41_M1_LOOP`, `ATLAS_H2D_SYNC`); main merged, with the 500-line allow-list moved into the script where #941 put it.

Receipts: `~/dflash-logs/ds41_tf2_minheap_F_fix.json`, `~/dflash-logs/ds41_suite_fix_{chat,raw}_{minheap,volvo}_r{1,2,3}.json`. Launch line: `atlas-notes/bin/ds41_smoke.sh serve` on the release build with `AVAROK_TARGET_HW=gb10 AVAROK_TARGET_MODEL=deepseek-v4.1-flash AVAROK_TARGET_QUANT=nvfp4`. The cold number follows after a cache drop, reported as measured with the same key.

Tracking: #1059 under #970.

TheTom (commented):
_Scope: this reviews only what #1099 adds on top of #1098 (+5,382 / −179 across 43 files). The stack is cumulative, so GitHub's diff here also includes the earlier PRs; those are reviewed on their own PRs._

This increment is the first time Atlas generates DeepSeek-V4.1 Flash end to end on one Spark, and engram is part of that path from the start rather than left for later.

**Engram.** `DeepSeekV41Layer::engram` (`crates/spark-model/src/layers/deepseek_v41_layer.rs:371-399`) hashes each position through the compressed token map, with 64-bit wrapping multipliers and lookback that stops at `-1` or at the start of the sequence (`engram_v41.rs:183-212`). It reads only the needed rows from the Q2_K tables on disk and applies them at layers 1 and 14. The PRD says "a model that loads and speaks while missing engram is still the wrong model", and this graph meets that directly.

Other parts that are solid:
- **Delayed hyper-connections follow the reference.**
  - The attention collapse uses the predecessor's `pre` (`:521`), and the FFN collapse uses `pre_a` (`:584`).
  - `pre_f` becomes `pre_prev` (`:649`).
  - The last layer collapses with the final FFN's `pre` and no learned head (`:666-668`), so V4's output HC head isn't reused.
- **The loader keeps memory bounded.** Routed expert stacks and engram tables are recorded as deferred, with their shard offsets, and skipped by the BF16 pre-flight estimate (`gguf/sidecar.rs:73`, `:150-165`). Nothing materializes them whole.
- **Real-weight routing oracle.** The layer-0 test asserts routing indices exactly against the CPU reference `Gate` (`weight_loader/deepseek_v41.rs:1047`).
- **Speculation is refused, not ignored.** MTP/speculation is disabled explicitly, and rollback is declined honestly through `decode_rollback_unsupported` (`deepseek_v41_layer.rs:1166-1176`) rather than re-steering on state it can't rewind.

The engine is here. The first three points below share one theme: the serving path doesn't yet enforce the request contract that PRD S4 ("API behavior") and §1 ask for.

**1. [Blocking] The 2,048-token cap isn't enforced at runtime.**
The PR text says "the 2048-token cap everywhere", but the code doesn't enforce it:
- **`max_seq` defaults to 8,192.** In `weight_loader/deepseek_v41.rs:202-204`, `max_seq` comes from `ATLAS_DS41_MAX_SEQ` with a default of 8,192, and can be raised as high as `max_position_embeddings`. `max_tokens = 2048` is the per-step workspace, not a limit on the total sequence.
- **The only check comes mid-generation.** The only position check is `attn_v41.rs:1205-1209` (`start_pos + m <= max_seq`), which fails a request partway through generation instead of refusing it up front.
- **The server admits more than the model supports.** The server still admits prompts against `--max-seq-len` (default 32,768; `cli/serve_args.rs:152`, `api/chat/mod.rs:357`), and `/v1/models` advertises 32,768. The V4.1 loader doesn't read `config.serve_max_seq_len` the way GLM-5.3 does (`glm5next_dsa/mod.rs:204`).

As a result, a 3,000-token prompt is accepted and run. PRD §1 caps the total at 2,048 tokens across prompt and output, template and special tokens included, and §3.1 notes that "a comment … is not a runtime guard".

Suggestions:
- Add a model-level hard cap, checked at admission against the rendered prompt length plus the requested output budget, that returns a 400.
- Clamp the advertised `max_seq_len` to that cap.
- Test just below, at and just above 2,048 serialized tokens, counting template tokens (charter C7).

**2. [Blocking] A second concurrent request corrupts the first one's engram state.**
`V41Runtime` is documented as "the ONE sequence in flight" (`deepseek_v41_layer.rs:61`), and the engram hasher lives inside it (`weight_loader/deepseek_v41.rs:361`). Two things combine to break that assumption:
- **Model-global state.** The hasher's position→token cache is model-global (`engram_v41.rs:156-189`). Layer 0 resets it, along with `SharedV41`, whenever any sequence starts at position 0 (`deepseek_v41_layer.rs:431-436`).
- **Concurrency still reaches it.** `decode_multi_seq_unsupported` (`:1162`) only moves the model onto the sequential fallback (`model/trait_impl/decode_b.rs:79-88`), which can still decode one sequence and prefill another in the same iteration. `--max-num-seqs` defaults to 128 (`serve_args.rs:491`), and nothing clamps it for this model.

The result is that a second request's prefill resets and overwrites the first request's n-gram history, and nothing reports an error. PRD §1 puts concurrency out of scope and asks for unsupported modes to be "rejected or disabled explicitly"; §3.2 says "keep hash state per request".

Suggestions:
- Refuse `--max-num-seqs > 1` for this model at serve start, or clamp it with a log line.
- Move the hasher's token cache into the per-sequence `V41LayerState`.
- Add a small test: two sequences interleaved step by step must each match the same sequence run alone.

**3. [Blocking] The serving path aliases V4.1 to V4, including tool calling.**
- **Prompt rendering.** `tokenizer/chat_impl.rs:115` renders V4.1 prompts with `ChatEncoding::DeepseekV4`.
- **Tool parsing.** `tool_defaults.toml:39` turns on V4 DSML tool parsing for `deepseek_v41` (also `tool_parser.rs:373`).
- **Reasoning.** `:61` sets the R1 reasoning parser.

PRD §3.1 says the V4.1 identifiers "must not be aliased blindly to `deepseek_v4`". §3.2 notes that V4.1 changed the DSML spacing and reasoning headers ("an existing V4 template is a starting point, not an oracle"), and §1 excludes tool-calling certification. Right now a request with `tools` is accepted and parsed in V4's format instead of being rejected.

Suggestions:
- Pin one supported text/chat mode against its official encoding with a token-ID fixture test.
- Reject `tools` / `tool_choice` for `deepseek_v41` until V4.1 DSML is verified.

For contrast, the kernel-tree reuse looks deliberate and fine: V4.1 has its own `MODEL.toml` behaviour table, and its kernels are new files. The alias that matters is on the serving side.

**4. [Should-fix] The first run is a smoke signal, and turning it into evidence is cheap.**
What the existing evidence covers so far:
- **The layer-0 oracle is narrow.** It covers five tokens on a ratio-0 layer that has no engram and no shared compression, and whose delayed mix is the one-hot starting value.
- **`hc_mixes` isn't checked.** It's printed but not asserted (`weight_loader/deepseek_v41.rs:737-739`).
- **No GPU-vs-reference graph run yet.** The description mentions `logits_full` parity through the production path. At this head, though, `logits_full` is only used by the CPU reference tests, and no test runs the assembled `DeepSeekV41Layer` graph on the GPU against it. So engram placement, delayed HC across layers, the shared KV producers and the final collapse (charter C5) don't have a GPU-vs-reference check yet.

"1..7 continues 8..12" and coherent 2-bit prose show that the graph runs. PRD §2 and C11 keep semantic correctness, same-GGUF agreement and quality as separate claims, and it would help to scope the PR text that way.

Smallest next steps:
- **(a) Tiny golden on the GPU.** Run the existing `ds41_tiny_golden.json` fixture (six layers, engram at layers 1 and 4, KV sources at layers 2 and 4) through `DeepSeekV41Layer` on the GPU and compare `logits_full`. Include one planted control, for example collapsing the FFN with `pre_prev` instead of `pre_a`, to show the check can fail.
- **(b) Real-weight agreement.** On real weights, run a teacher-forced comparison against the pinned community runtime over the eight C11 prompts, recording top-k logit overlap and the first divergent position. `ATLAS_DS41_NO_ENGRAM=1` is already the known-bad control C11 asks for.

Two smaller items:
- **Routing on all layers.** Comparing selected expert IDs against the CPU `gate` on all 40 layers during the real run would help, with the top-6/top-7 margin logged. `ATLAS_DS41_TRACE` already logs routing per layer.
- **Hard-coded path.** `MODEL_DIR` is hard-coded at `weight_loader/deepseek_v41.rs:591`, and an env var would let a teammate reproduce the oracle.

**5. [Should-fix] A broken split set falls back to loading one shard.**
In the generic GGUF loader (`crates/spark-runtime/src/weights/gguf.rs:442-447`), an error from `find_gguf_shards` is downgraded to a `warn!` and the loader continues with `vec![path.clone()]`. `find_gguf_shards` already returns `Ok` with a one-file set for a GGUF that isn't split. So the fallback isn't needed for ordinary single-file models; it only triggers when a split set is actually incomplete or inconsistent, and then it silently loads shard 1 alone. That's exactly the failure RST charter C1 targets. The V4.1 expert path propagates the error with `?` (`expert_stream.rs:72`), so the practical exposure is other split GGUFs, but the generic path should fail hard too.

**CI.** `cargo test --workspace` fails one test: `every_real_target_resolves_a_nonempty_source_set` (`crates/avarok-plugin/src/gate/taxon_tests.rs:58`). This increment adds three `deepseek-v4.1-flash` targets, and none of them is in the expected list yet:
- gb10 (`5333ff966`)
- b200 and hopper (`2fccec601`, via `kernel_source = "deepseek-v4-flash"`, which exists on both)

It's a scope choice: either add all three entries here, or leave the b200 and hopper targets (and the `datacenter-binaries.yml` line) out of a PR the PRD scopes to one GB10. Their own headers already note that `expected_absent` was never harvested on that hardware.

tbraun96:
**Do not close this one.** Its siblings #1094–#1098 are closed as contained in #1131; this PR is not.

#1131 consolidated the ds41 line, but it took #1099 at `f86163c17077` (2026-09-16 21:28Z) — not at this PR's head. Sixteen commits pushed after that point are **absent from `main@68c393913502`**, verified by patch-id and by grepping every symbol they add. Four of them are not diagnostics:

| commit | what is missing from main |
|---|---|
| `4a859ae5b785` | sums single-token expert rows in plan order, fixing an **inter-block race in `scatter_add`**. `moe_v41_sum_rows` does not exist on main; `moe_v41.cu` still has only `moe_v41_scatter_add` |
| `d75da22bfda6` | the loop watchdog declines its rollback on a non-rewindable layer — `RollbackFallback::LayerStateNotRewindable`, absent |
| `69dc4fba9f48` | `release_state` frees the four per-sequence attention buffers — absent, so this is a live leak |
| `1d1d7b024969` | chat/tool/reasoning parsers registering `deepseek_v41`. There are **zero `deepseek_v41` hits under `crates/spark-server` on main**, so the model that just landed has no parser registration |

Also absent: the per-layer CUDA-graph decode path (`GraphMode`/`LayerGraphs`/`step_graph`, `ATLAS_DS41_GRAPH=1`), the attention/MoE host-device splits it rests on, the Hopper (sm_90a) and B200 (sm_100a) targets (`kernels/{hopper,b200}/deepseek-v4.1-flash/MODEL.toml` do not exist on main), `ATLAS_DS41_PREFILL_GEMV`, the `ATLAS_DS41_TRACE` checksums and two smaller diagnostics.

Merging this PR as-is would duplicate everything already landed. The right move is to **re-cut those sixteen commits onto current main** as a follow-up — same treatment #844/#845 got — and close this pointing at that branch. Leaving it open until then so the work is not lost.

rrstesiak:
Agreed on the list, and thank you for holding this one open.

The four that are not diagnostics are already re-cut onto main as #1144, opened before your comment: the race fix (`moe_v41_sum_rows` replacing the m=1 `scatter_add`), the watchdog rollback decline, `release_state` for the four attention buffers, plus the single-token test Turney asked for and the buffer-sizing bug it found. Six commits, cherry-picked with `-x`, size-cap and `cargo check` clean, on the split layout. That is the one to land first: until it does, a V4.1 decode on main is nondeterministic and loops.

The rest of the sixteen follow as a second re-cut against main on its own branch: the `deepseek_v41` parser registration, the per-layer CUDA-graph decode path with the attention and MoE host-device splits under it, the Hopper and B200 targets with the mirrored kernels, the three small fixes, and the diagnostic switches, env-gated and default off. The clippy and rebrand commits are already reflected in the split files on main and do not need re-cutting. I will point this PR at that branch and close it when both are up.

rrstesiak:
Both re-cuts are up, so closing this one pointing at them, as suggested.

- #1144, against main: the four that are not diagnostics (the race fix, the rollback decline, `release_state`), plus the single-token test and the buffer-sizing bug it found. The one to land first.
- #1146, draft, stacked on #1144: the rest of the sixteen re-cut onto the split layout. The `deepseek_v41` parser registration; the per-layer CUDA-graph decode path with the attention and MoE host-device splits under it, hand-ported into `attn_v41/decode.rs`, `moe_v41/single.rs` and `deepseek_v41_layer/graph.rs`; the Hopper and B200 targets; the diagnostic switches, env-gated and default off. Skipped as already on main: the mirrored kernel symlinks, the three small fixes, the clippy and rebrand commits. Dropped: `ATLAS_H2D_SYNC`. `cargo check` and clippy clean, size-cap passing, `deepseek_v41` now registered under `spark-server`.

Between them, every commit from this branch after `f86163c17` is either on main already or in one of those two.

PR #1101, opened 2026-09-16
rebrand: Atlas to Avarok on the site, crates, env, and cfg

TheTom:
## Summary

Product name is Avarok. This PR is the rename in this tree: marketing/blog/docs chrome, `atlas-*` crates to `avarok-*`, `ATLAS_*` env to `AVAROK_*`, `cfg(atlas_*)` to `cfg(avarok_*)`, and in-tree `atlasctl` strings to `avarokctl`.

Site: public engine URL is `/engine` (`https://atlascybernetics.ai/engine`). Control is `/control`. C=128 published ratio is 1.333x vs matched vLLM+MTP (478.11 / 358.57 tok/s), not 1.225x vs unmatched nospec. Blog posts and header lockup follow. Theme toggle is back.

Closes #

## Test plan

Do not merge until @tbraun96 and @rrstesiak have each pulled the branch and run the site locally. A GitHub approve without that is not the check.

- [x] `cargo fmt --all -- --check` (the rename pushed 28 files past the line limit; formatted in `dafbc24a6`)
- [x] `AVAROK_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy --workspace --tests` (exit 0, no warnings, plus `cargo test --workspace --no-run` exit 0, on a DGX Spark; rerun at each push, last at `25aa2f473`)
- [ ] `AVAROK_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo metadata --no-deps --offline` lists `avarok-core`, `avarok-kernels`, `avarok-plugin`, not `atlas-*`
- [x] `cd site && bun x --bun vite dev` then `/`, `/engine`, `/control` (also `lhci autorun` on the production build: 100/100/100/100 on all three, dark and light)
- [ ] `cd blog && bun x --bun vite dev --port 5174`; Avarok mark goes to the marketing landing, not Latest
- [ ] `/engine` C=128 reads 1.333x vs vLLM + MTP
- [ ] `cd site && bun test src/lib/marketing.test.js src/lib/ladder.test.js src/lib/theme-color.test.js`
- [ ] `cd blog && bun test src/lib`

## Notes for reviewers

Left as-is: repo path `Avarok-Cybersecurity/atlas`, `atlascybernetics.ai`, `@AtlasInferenceX`, HuggingFace `Atlas-Inference`, blog slugs (`seven-tenets-powering-atlas-inference`), `bench/ladder38/` historical filenames, the MLCommons announcement's "Atlas Inference" wording. `AVAROK_SKIP_BUILD` still honors `ATLAS_SKIP_BUILD`. Gate JSON still deserializes `atlas_version` and `atlas_repo`.

Also kept, because they name things outside this tree: the exported symbols of the sha-pinned `3rdparty_patches/gdn_aot/libatlasgdn.so` (`atlas_gdn_load`, `atlas_gdn_prefill_packed_managed`), the on-disk file magics `ATLASGM1` (xgrammar mask snapshots) and `ATLASLQE` (LQER files), the self-hosted runner labels `atlas-cmd` / `atlas-pr-cheap` / `atlas-pr-rust` / `atlas-macos-burst` (the runners are named `avarok-*` on GitHub but still carry those labels), `assets/atlas-demo.{mp4,gif}`, the `atlas` series id in `bench/ladder38/published.json`, and the recipe id `qwen3.5-0.8b-bf16-atlas`, which comes from atlas-recipes and is regenerated into `llms.txt` on every build.

There is no `atlasctl` crate here; the binary, its `atlasctl-*` crates, installer and release artifacts live in `Avarok-Cybersecurity/atlas-recipes`, and that rename is a separate PR. Every string in this tree now says `avarokctl` except the recorded gate outputs under `.benchmarks/`, which keep the commands they actually ran. Until the recipes repo ships an `avarokctl` binary, the commands documented here do not resolve on a host.

`atlasinference.io` is still a custom domain on the same Pages project (200). A 301 to `atlascybernetics.ai` needs the host detached from Pages, then a zone Redirect Rule. Origin vhost: `site/deploy/nginx/atlasinference.io.conf`.

`crates/` and `kernels/` are in the diff (names and comments, not math). If you apply PERF_PATHS strictly, that is a certification campaign. Nothing in the kernels changed numerically.

## Follow-ups on this branch

- `7f37285b6` Lighthouse: the home page carries one hero ``; the boot script preloads the render the theme calls for and an inline pick sets its `src` while the page parses, so a load fetches one render (light is now a 95 KB webp, PNG source under `assets/brand`). Home LCP 2.0 s to 0.7 s. Both themes score 100 on index, engine and control locally with the repo's `lighthouserc.json`.
- `4b18f7cec` Lighthouse: light-theme text contrast. New `--ch-*-text` and `--sx-text` tokens (fill hue in dark, darkened in light), `--t3`, `--green`, `--amber` darkened in light, every glyph painted with a chevron hue reads the text twin (site, deck, blog tags, callouts, inline code, Shiki palette). Guard test pins the ratios. Blog scores 100 on all five pages in both themes locally. The Lighthouse job was already red on this branch before the rename commit and on `main` since Sept 14 for the same contrast nodes.
- `3924f7631` dark theme hero: the render has a near-white ground and no blend mode lifts it off a dark page, so dark mode swaps in a re-toned derivative (`avarok-hero-dark.webp`, 99 KB). Also renames `atlas-hero.png` to `avarok-hero.png`, which the rename commit referenced without moving.
- `7697543c0` ladder chart: colour and dash follow the series role, so the `atlas` id in the bench manifest no longer blanks the Avarok line.
- `1378b4499` hero heading tracking -0.072em to -0.05em.
- `bdad2c880` hero light sweep every 16s with a trailing floor shimmer; reduced motion hides it.
- `6f7300516` Prometheus metric names stay `atlas_*`: atlasctl's launch statistics parse them by exact name, so they follow that binary's rename, not this one.
- `54d794b51` image references back to `avarok/atlas-gb10` and `avarok/atlas-qwen3`; `avarok/avarok-gb10` does not exist on Docker Hub.
- `ebcde4b88` `avarok_core::env_compat::mirror_legacy_env()` first thing in `spark` main: every `ATLAS_*` variable is copied onto its unset `AVAROK_*` twin, with one deprecation line on stderr. Without it a server from this tree ignores every recipe `env:` block and the agent's `ATLAS_HOME`. `main` builds the tokio runtime explicitly because `#[tokio::main]` starts worker threads before the body runs and `setenv` needs a single-threaded process.
- `e7525a15d` on-host state: `AvarokHome::resolve()` reads an existing `~/.atlas` when `~/.avarok` is absent and `AVAROK_HOME` is unset (signing identity, artifacts, runs); the recipe cache does the same for `atlas-recipes/`. The doctor check and the sync and bench hints name the path actually read.
- `25aa2f473` `Cargo.lock` re-sorted; the in-place package rename left it unsorted so every build rewrote it and `--locked` refused it.
- `7e2b269fc` `dafbc24a6` `ec964d182` `ddf97b8c1` rename repairs: GDN dlsym symbols back to what the pinned `.so` exports; `avarokctl.rs` and `avarokctl_tests.rs` renamed to match their `mod` lines (spark-server did not build); the two eight-byte file magics restored (nine bytes did not compile and would orphan files on disk); runner label literals back to the labels the runners carry.

## CLA

## The Atlas Cybernetics Corp brand, and what is NOT renamed

The internals stay Avarok — `avarok-*` crates, `AVAROK_*`, `cfg(avarok_*)`, `~/.avarok`. Every user-visible surface is Atlas again, under the corporate identity the new kit carries: **Atlas Cybernetics Corp** on the top left of the site, in both footers and in the README, as the kit's own `logo-full-corp` artwork with its tagline outlined (the kit ships it as live `<text>` in DejaVu Sans, which renders with a fallback font on most visitors' machines). Light and dark come from two new tokens, `--logo-word` / `--logo-tagline`, because the guidelines call the light-ground greys logo colours and the light theme darkens the text ramp for WCAG.

Six things this branch had broken, each pointing at something that does not exist: `avarokctl` (atlas-recipes ships `atlasctl`, and the site serves that repo's `install.sh`), the RAG corpus URL, `x.com/AvarokInferenceX`, the Docker badge, the record schema (`atlas_version` / `atlas_repo` / `atlas_home` are read by CI, by the site and by every record on disk), and — invisibly — `is_avarok()` comparing against `"avarok"` while every real recipe declares `runtime: atlas`; the fixtures had been renamed in lockstep, so the tests stayed green while the live SSOT stopped being recognised.

Lighthouse now asserts 100 in all four categories on every page the site ships, `control.html` and `diligence.html` included. #1102 is merged down here: one campaign covers both.

Hardware: b200, gb10, hopper, strix, strix-hip
CHKI-Verdict: benign -- the 146 kernel paths in this diff change identifiers and comments only: include guards (`ATLAS_E4M3_LUT_CUH` -> `AVAROK_*`), device function and macro names (`atlas_e4m3_to_f32`, `atlas_bf2`, `atlas_mma_e4m3`, `ATLAS_GA`), and prose. 995 insertions against 995 deletions, no expression, type, launch bound or numeric constant among them, and `grep -rn 'atlas_[a-z0-9_]*' kernels/` now returns nothing, so no consumer is left calling a symbol that no longer exists. strix and strix-hip compile the same bytes through symlinks, so they see the same rename and nothing else; the HIP shims reference none of the renamed symbols.


rrstesiak:
**Kerning on the hero wordmark ("terms."), measured**

Both images from the thread, measured at the x-height band rather than by eye. The objective test for a pair like r|m is stem rhythm: the distance from the r's stem to the m's first stem should equal the m's own stem-to-stem pitch, so the eye sees one even beat across the word.

| pair | original | fixed |
|---|---|---|
| r stem to m first stem | 60 px | 56 px |
| m internal stem pitch | 71 px | 61 px |
| r-m as % of the m's own rhythm | 85% | 91% (still ~9% tight) |
| r arm touching m shoulder | yes | yes |
| e-r gap, % of text height | 4.4% | 7.0% |
| m-s gap, % of text height | 3.3% | 5.7% |
| s-period gap, % of text height | 7.1% | 9.6% |

**Verdict.** The fix is the right direction. The r-m pair still sits inside the m's own stem rhythm and the r's arm still runs into the m's shoulder; that is the tightness the eye catches. The m-s and e-r pairs are open, so the word carries three spacing rhythms in five letters, and unevenness is what reads as an error, not any single gap.

**The fix.**
- `r|m`: about **+20 units** (per 1000 em). Clears the arm and puts the pair on the m's rhythm.
- `m|s`: bring up to match `e|r`, roughly **+4 px** at the rendered size.
- `t|e`: leave it. The crossbar touch is a standard tight-Helvetica idiom.
- Match the kerning already established elsewhere on the site so the hero does not read differently from the rest.

**Rules worth writing into the branding guide.**
1. Even rhythm, not equal distance: the stem-to-stem beat across a pair should match the counters beside it.
2. Display type is set tight but never touching, unless every pair touches by design.
3. Kern the wordmark by eye, then verify by measurement.

**Why it matters.** Well-set type measurably improves reader mood and problem-solving performance (Larson, Hazlett, Chaparro and Picard, Microsoft Research with MIT, 2006), and easier-to-read text is judged more credible independent of its content (processing fluency). The typeface's emotional read lands within the first 100 ms, before letter recognition. The hero wordmark is the opening sentence of every visitor's opinion of the product.

PR #1102, opened 2026-09-16
docs: standard for performance numbers

rrstesiak:
THIS IS MOSTLY VERY FIRST ALPHA DRAFT MEANT MAINLY AS A PLACEHOLDER

Standard for Performance Numbers

The diligence page at atlasinference.io/diligence is the single public record, and our internal best numbers and the page must agree.

Official Guidelines:

1. Any internal result that beats the page goes onto the page by PR, with the following:
   a. launch line: every parameter and switch, model and drafter locations
   b. commit sha
   c. driver line: the benchmark command and its settings
   d. receipt JSON

2. Any page number an internal run cannot reproduce within its published spread:
   a. gets a note in this channel with the run's receipt
   b. gets the page corrected by PR

3. Every table posted here:
   a. same driver, same settings
   b. medians over reps
   c. competing engines in the same table
   d. TTFT p50 alongside tok/s, per rung

A number without a launch line is a lab note, not a claim.


tbraun96:
Landed in #1101 (`569ffd092`): this branch was merged down into the rebrand base so one certification campaign covered both, and `docs/perf/PERFORMANCE_NUMBERS_STANDARD.md` is on `main` now — verbatim, authorship intact.

GitHub squash-merged #1101, so `edd4b1712` is not an ancestor of `main` and this PR could not auto-close; closing it by hand rather than leaving a merged change open. Nothing here was dropped:

```
$ git show avarok/main:docs/perf/PERFORMANCE_NUMBERS_STANDARD.md | head -3
# Standard for Performance Numbers

The diligence page at [atlasinference.io/diligence](https://atlasinference.io/diligence) is the single public record, …
```

PR #1103, opened 2026-09-16
fix(grammar): poolside_v1 must not allow empty required string

SeedSource:
## Summary

The `poolside_v1` tool-call grammar compiles a required string argument's
value as `value ::= value_part*`, so `<arg_value></arg_value>` is a legal
zero-length continuation for every string parameter, including required
ones. The other tool-grammar compilers in this file (Hermes, bare-JSON,
Qwen3-Coder, Gemma-4) already guard required strings against this; poolside
was the one gap. This PR adds the same guard: a `nonempty_value` rule used
only for required string parameters, so the grammar cannot emit an empty
value where the tool schema requires a non-empty string. Optional strings
and non-string parameter types are unchanged.

## Test plan

- [x] `ATLAS_SKIP_BUILD=1 cargo clippy --workspace --tests` (clean, no warnings)
- [x] `cargo check -p spark-server`
- [x] `cargo test -p spark-server --bin spark poolside` — 32/32 passed, including
      new coverage: `poolside_grammar_rejects_empty_required_string`,
      `poolside_grammar_still_allows_empty_optional_string`,
      `poolside_grammar_cannot_emit_the_tc43_empty_query`,
      `poolside_grammar_matches_qwen3_coder_on_empty_required_strings`,
      `poolside_grammar_leaves_required_non_string_alone`,
      `poolside_grammar_accepts_a_real_web_search_query`,
      `poolside_grammar_accepts_non_empty_required_string`.
- [ ] `bash scripts/check-license-headers.sh` — pre-existing failures on
      unrelated vendored kernel headers (`kernels/hopper/qwen3.6-27b/...`);
      not touched by this change, not newly introduced.

Scope: this only affects grammar compiled for
`--tool-call-parser poolside_v1`. No other tool-call parser's grammar
compiler is touched.

**Not claimed:** no model-quality or tool-eval score numbers are asserted
in this PR. Any change to a tool-eval pass rate for a specific model is a
separate, model-level result and is out of scope here — this is a grammar
correctness fix only.

## Notes for reviewers

The fix is grammar-only: it constrains what the constrained decoder is
allowed to emit for a required string argument. It does not change parsing
of already-produced tool calls, and it does not change behavior for models
that don't select the poolside_v1 tool-call format.

## CLA


SeedSource:
Author's review note (independent pass over the diff before requesting the seal):

- `req_value ::= value_part value_part*` forces at least one byte, and every `ebnf_until_close` alternative consumes at least one character, so no empty derivation exists for a required string. Optional strings, non-string params, and schemas without `required`/`properties` compile to the same EBNF as before. No new panic paths; malformed `required` falls back to the pre-fix grammar.
- Known narrow gap, deliberately left for a follow-up: the guard only applies when the property set is closed (`additionalProperties` absent or `false`), because that is when poolside enumerates parameter names. A schema with `additionalProperties: true` plus a required string still compiles the old `value ::= value_part*`. The JSON-schema compilers (hermes/bare_json/gemma4) enforce `minLength: 1` regardless. I'll open a follow-up with a test for that combination rather than widen this PR.

SeedSource:
Note for maintainers: the merge-group certification job reported every gate as NONE with "git cannot diff that commit" because this PR comes from a fork and the merge_group checkout only has base-repo refs, so the commit the records are keyed to (`456a4f247e`) was not fetchable there (the pull_request-event run of the same job passed). I pushed `cert-anchor/pr-1103-456a4f247e` to make that commit reachable and re-queued; the records themselves are unchanged and still cover the head (no perf-path change between record and head). A small ci.yml follow-up to fetch `refs/pull/<n>/head` on merge_group would remove the need for anchor branches on future fork PRs; I'll open it separately.

PR #1104, opened 2026-09-16
atlas-plugin: recognize numeric size labels in vision benchmark replies

SloptimistPrime:
## Summary

The vision benchmark rejected a correct HD fixture reply, `07_hd_1280x720`, because its standalone-term matcher treats `_` as part of an identifier and misses the required numeric term `1280`. Match complete numeric terms using digit boundaries in `reply_matches`, for both required and forbidden terms. Word-term matching stays unchanged. Add regression cases for the fixture and longer numbers (`11280`, `12800`).

This bug is present on `main` and on the tested head of #1063. The scorer files were identical before this change, so #1063 can cherry-pick the fix.

## Test plan

- [x] New numeric-label regression failed against the original #1063 head, then passed with this change; all 10 vision scorer tests passed.
- [x] `rustfmt --check` on the changed files and `git diff --check` passed.
- [x] A benchmark-client-only build with this patch passed the full `vision-fidelity` gate against unchanged NVIDIA ModelOpt NVFP4 servers on two-node TP2/EP2, four-node EP4, and four-node TP2×EP2: 14/14 geometry, 3/3 probes, 10/10 integrity, and 3/3 clean concurrency on each layout.
- [ ] Full workspace formatting, clippy, and license-header CI checks are pending. The changed files retain their SPDX headers.

## Notes for reviewers

Direct non-streaming and streaming responses from the unchanged two-node server contained the exact correct HD label. This patch changes only the benchmark scorer and tests; it does not alter serving or model numerics.

The separate four-node TP2×EP2 collective and prefix-agreement concern identified during #1063 testing needs its own exact-output and subgroup tests.

Prepared with AI assistance. Please review the code and test evidence.

## CLA


SloptimistPrime:
Ready for review. GitHub is holding 11 workflow runs for first-time contributor approval; could a maintainer approve them so Stage 1 can report? This changes only vision benchmark scoring. The current certification coverage treats `crates/` as measured input, so a fresh record campaign may be required even though serving is unchanged. Would you prefer this standalone fix to carry its own records, or for #1063 to cherry-pick it into that campaign? I can share the NixOS 2/4-node benchmark artifacts.

SloptimistPrime:

rsafier:
I cherrypicked with attribution on #1063 and resolved vision test.

tbraun96:
Heads up on why this is stuck, and what happened to the change.

This PR edits `crates/atlas-plugin/src/benchmarks/vision/score.rs`. The #1101 rebrand renamed that crate, so `main` now carries `crates/avarok-plugin/**` and the `atlas-plugin` path no longer exists. Git reports **no conflict** for this branch, which is the trap: it would simply re-create the deleted directory, so merging it would add a dead file, leave the live scorer unfixed, and pass CI while doing it.

The fix itself is right and is preserved verbatim — the patch applies at the new path with only the directory name changed. It is on `fix/vision-size-label-recut` (`6dc3def7`), with you credited as co-author.

Verified there: reverting `score.rs` while keeping your new assertions fails two tests, naming `reply_matches("07_hd_1280x720", &["1280"], &[])`; with the fix `avarok-plugin`'s score suite is 96/96. Your `11280` negative case is what makes it a real check rather than a loosened matcher.

It is parked rather than opened as its own PR because `crates/` is a performance path, so landing it alone would trigger a full benchmark certification campaign (~3 GPU-hours). It will ride in the next stack instead.

tbraun96:
Closing: your fix is on `main` as of #1145, and merging this PR now changes nothing.

**Evidence** — merged locally onto `main` at `ce2bccd736` and diffed the result against that same sha:

```
$ git merge --no-commit --no-ff pr1104 && git diff --cached --name-status origin/main
(no output)
```

Zero paths changed. Your two files were carried by **`ce2bccd73` "follow-ups: the WOA review fixes, the Windows INFINITY defect, and two re-cuts (#1145)"**, which re-cut this change onto the post-#1101 paths (`crates/atlas-plugin/` → `crates/avarok-plugin/`) — the diff there is byte-identical to yours, same pre- and post-image blob hashes (`dd6a98161..b048c01e5`) on both sides of the rename. Your authorship is credited in that commit's message.

One thing worth knowing for next time: this PR's checks read as "no failures", but they had **never run** — all eleven workflow runs were parked at `action_required`, the fork-PR approval gate. That gate applies per head sha, so every push to a fork branch needs a fresh maintainer approval. A PR sitting with no red marks is not necessarily a PR that passed.

Thank you for the matcher — the HD fixture's `07_hd_1280x720` label is exactly the case it fixes, and that reasoning is preserved in the comment on `main`. Reopening restores this if the reading is wrong.

PR #1108, opened 2026-09-17
Vision: size video to a row budget instead of failing inside the encoder

rsafier:
**Stacked on #1063** (`pr973-ep-prefix-cache`) — review that one first; this branch is the 3 commits on top.

## The problem

A long or high-resolution clip died *inside the ViT* against a 16,384-row buffer, after decode, preprocessing and scheduler admission had all been paid for. 3 minutes of 360p asks for 40,500 merged rows; 145 groups of 720p asks for 133,400. Nothing upstream knew the bound, so nothing could act on it — and the scheduler's own copy of the bound was a hardcoded `VISION_P_MAX = 6400`, which was `FALLBACK_MAX_PATCHES`, **2.56× too small**.

## The two bounds are not the same number

Conflating them is the root cause, and splitting them is most of the fix:

| | scales as | cost | can it move? |
|---|---|---|---|
| `p_max` — ViT attention scratch | **quadratic**, 6 B/p² | 1.61 GB at 16,384 | no |
| `buf_out` — merged rows | **linear**, 5,120 B/row | 671 MB at 131,072 | yes |

A clip is bounded by **rows**, not by any one frame's patches. `ATLAS_VISION_OUT_ROWS` raises the row budget and defaults to `p_max`, so an unset serve allocates and refuses exactly what it did before. It must be set on **both ranks** or the EP broadcast bound disagrees — the rank-1-blind class.

## What changed, in the order a request meets it

1. **One `VisionCapacity { p_max, out_rows }`**, read from the encoder at boot (the encoder is the only authority — `CEILING_MAX_PATCHES` may have clamped what was asked for) and carried on `AppState`. The request path can now refuse early with an HTTP 400 that says what to do, instead of failing late in the encoder.
2. **`p_max` and `out_rows` separated**, per the table above.
3. **An auto-fit with a declared policy.** The budget is a conserved product — `groups × frame_area ≤ rows × 1024` — so every choice is a point on one hyperbola and no policy wins for every clip. `Coverage` (default) keeps every temporal group and shrinks frames; `Detail` keeps frame size and drops groups; `Balanced` splits the deficit geometrically. **`fit_video` is a no-op when the clip already fits** — that is what keeps every gate fixture byte-identical.
4. **Chunked encode.** The scratch holds `Σp ≤ p_max`; anything larger used to fall back to one image at a time — correct, but 27 block launches per group with no GEMM reuse. `forward_batched` now packs as many whole groups as fit and runs the batched sequence per chunk. ViT attention is per image over disjoint slices, so a chunk boundary changes no arithmetic. The per-image path is **kept** as a reference oracle behind `ATLAS_VISION_NO_BATCH=1` — it is the only caller of the single-image kernel sequence, and it is what makes "chunking is bit-identical" re-checkable rather than a comment.
5. **Row budget clamped to half the context.** Every merged row becomes exactly one prompt token, so a budget the sequence cannot carry is not capacity; without this, raising `out_rows` just moves the failure from the encoder to mid-prefill.

Plus a separate small fix: **`--video-max-frames` never reached the sampler**, only ffmpeg's argv. Raising it did nothing (the sampler cut back to the 768 constant) and the GIF path, which never goes near ffmpeg, ignored the flag outright.

## Two defects found while testing this, both mine, both fixed here

- **The duration probe extended the caller's timeout instead of spending it.** It ran its watchdog *beside* the decoder's, so a hung input paid both: a 1 s policy took 3.0–3.1 s (flaky 2 runs in 3 against the suite's own 3 s bound) and a 120 s policy would have meant ~150 s. The probe now declines below a 4 s budget and its time is subtracted from the decode's. Pinned by a new regression test.
- **The grid rounds UP, so the fit's area was not the final cost.** 720 px is 22.5 grid units at this checkpoint's 32 px, so 1280×720 becomes 1280×736 and costs **920 rows/group, not 900**. 145 groups of that overshot by 2,328 rows and earned a 400. The rounding is now measured — size the frame, cost it with the real grid, shrink, repeat — rather than predicted.

## Verification

Measured on 2× DGX Spark GB10, TP=2 EP=2, qwen3.8-Flash-Next NVFP4.

| case | before | after |
|---|---|---|
| 3 min of 360p, stock 16,384 budget | died in the encoder | 200 OK, refit to 416×224, **all 180 groups kept**, correct description |
| same clip, 131,072 budget | — | 200 OK **native**, 39,628 tokens (2.4× the detail) |
| 145 groups of native 1280×720 | died in the encoder | 200 OK, 127,633 tokens, 223 s, reached the model at 1280×704 |

**Gates, run against a live server:**

- `vision-fidelity` **PASS** — 14/14 geometry, 3/3 probes, control held.
- `video-fidelity` **13/14** — the only failure is `video-before-image`, which was **already failing on 2026-09-15, before this branch was cut** (passing 2026-09-13; regression window is entirely upstream of this work). Tracked separately.

The decisive check is that **every token count on every fixture is byte-identical** to the pre-change run — they all fit, so the fit must be a no-op:
`49/121/192/220/576/576/920/405/1400/1/128/49/49/49` and `67/116/214`, `214`, `286`, `291`, `437`, `240`.

`spark-model` 1447/1447 · `spark-server` 2628 with only the known-flaky `qwen_dense_parity` · clippy clean · all targets build.

## Capacity, for reviewers sizing a deployment

720p at 1 fps costs **460 rows per second of clip** (920 per 2-frame group).

| rows | `buf_out` | 720p @1fps | % of 256K ctx |
|---|---|---|---|
| 16,384 (stock) | 84 MB | 36 s | 6% |
| 131,072 | 671 MB | ~4.8 min | 50% |

GPU memory is nowhere near binding (23 GB free against 671 MB). **Context is** — a 9.7-minute clip fills the entire 262,144 window and leaves nothing for the question, which is why the clamp is half.

## Not done

- `video-before-image` is left as-is; it is pre-existing and wants its own bisect over `5e9774929..HEAD`.
- No CLI flag for the row budget yet — `ATLAS_VISION_OUT_ROWS` is the knob, deliberately, because it is trivially symmetric across ranks in the launcher and a rank-asymmetric budget is a correctness bug rather than a config mistake.


rsafier:
## video-fidelity now passes 14/14

The `video-before-image` leg is fixed, and it was neither a model fidelity limit nor a code defect.

**Cause:** `ATLAS_GDN_FLASHINFER`. FlashInfer and Atlas's FLA scan are equivalent *math*, not identical arithmetic — the canonical cross-impl A/B (`examples/gdn_fla_vs_fi.rs`) measured **cos = 0.999993**, FI getting there with bf16 tensor-core inner matmuls where FLA accumulates in f32 scalar FFMA. Over 36 GDN layers that ~1e-5 is enough to move a temp-0 answer, and on this cell it does.

The model was never losing the 4th temporal group. Asked directly, with clip + still in the same request, it still answers `4` segments, `final segment: Yellow`, `4th segment: Yellow` — it garbles only the **ordered list**. Geometry matches the passing mp4 leg exactly (4 groups, 14×14 patches, 196 vision tokens, 288 prompt tokens) and the pixels differ only by the YUV round-trip, `(254,0,0)` vs `(255,0,0)`.

**Provenance:** `41639c138` (09-15 10:53) baked the flag into the TP=2 × EP=2 preset. Last passing record 09-15 09:40, first failing run 09-15 16:39 — the flag landed inside that window, and flipping it reproduces both outcomes on demand.

Ruled out by measurement, not argument: prefix caching (a nonce forces a cold prefix, no change), streaming (stream ≡ non-stream), `max_tokens` (320 ≡ 2048), MTP (`--speculative` is off), the fixture (a GIF generated from the **passing** mp4 fails identically), and `ATLAS_MOE_SHARED_CUTLASS` — the other non-bit-exact arm in the same preset — which changes nothing.

**The trade** (2× GB10, TP=2 EP=2, full gate runs both ways, 3 prefill samples/point):

| | video-fidelity | prefill 8K | prefill 32K |
|---|---|---|---|
| `ATLAS_GDN_FLASHINFER=1` | 13/14 | 2435 tok/s | 2472 tok/s |
| `ATLAS_GDN_FLASHINFER=0` (now default) | **14/14** | 2187 tok/s | 2214 tok/s |

~10% prefill for one gate leg. Off ships because a wrong answer is worse than a slower right one, and 2187 still clears the 2K floor. **This changes #1063's headline prefill number**, so it is worth an explicit look — a throughput-first serve turns it back on with `ATLAS_GDN_FLASHINFER=1` in the environment (on every rank), which overrides the preset.

Verified on the stock config with no overrides: **video-fidelity 14/14** (0 skipped, control held), **vision-fidelity 14/14 geometry + 3/3 probes**, and the long-video path unchanged (3 min of 360p → 39,628 prompt tokens, correct description).

PR #969, opened 2026-09-08
EXL3 on GLM-5.3-Flash: routed-expert packs, EP=2/TP=2, and prefix caching

rsafier:
Brings the EXL3 (QTIP trellis) stack onto the GLM-5.3-Flash port, opens prefix
caching for `glm5_next`, and fixes the EP=2 deadlock that opening it exposed.
Based on `cert/glm53-records-20260907`, not `main`, so the diff is these 12
commits rather than the whole GLM integration lane.

## Measured on 2x GB10 (DGX Spark), EP=2/TP=2 unless noted

| | result |
|---|---|
| `GLM-5.3-Flash-EXL3-K2` (2bpw), single GB10 | loads, answers, 9.5 tok/s |
| `GLM-5.3-Flash-EXL3-K2`, EP=2/TP=2 | 14.7 tok/s, 52.6 GB pre-KV/rank |
| `GLM-5.3-Flash-EXL3-TR3-4bpw`, EP=2/TP=2 | 13.4 tok/s, 88.8 GB pre-KV/rank |
| prefix caching, cold vs warm (K2) | 92.8 s -> **4.2 s (22.3x)**, answers identical |
| bleed control, 4 questions / 1 cached preamble | 4 distinct, all correct |
| agentic-webserver, 4bpw, reasoning-low + prefix cache | **Pass 3/3**, 33 turns, 6260 tok, 24.8 s/turn |
| agentic-webserver, K2, reasoning-low + prefix cache | Pass, 15 turns |
| cross-rank anchor disagreements since the retirement fix | **0 over 52 restores** |

4bpw answers `12 x 8 = 96` correctly where the 2bpw pack does not — same binary,
same kernels, same topology, temperature 0, only the pack differs. Single-shot
probe, so that is strong evidence the 2-bit routed experts are responsible
rather than the routed-sum code, not proof.

## The four defects worth reading

**1. `layer_is_exl3` probed expert 0 unconditionally.** Under expert parallelism
the weight store is sharded, so rank 1 of an EP=2 split over 288 experts starts
at 144 and has no expert 0. The probe missed, the layer was judged BF16, and the
dense binder failed far from the cause on
`...experts.144.gate_proj.weight not found` — a name the EXL3 pack never
contained. Rank 0 booted fine, which made it read as an EP transport problem;
NCCL had in fact reached `Init COMPLETE` on both ranks.

**2. Prefix caching: the predicate is SPLIT, not flipped.** `glm5_next` declined
[...]

PR #1109, opened 2026-09-17
perf(glm5_next): cuBLASLt prefill + DSA indexer fixes (from #969)

SeedSource:
## Summary

Four commits from Avarok-Cybersecurity/atlas#969 (rsafier, unchanged): cuBLASLt
routing for wide GLM projections (active only when a prefill sub-chunk exceeds
16 rows), an FP32-out cuBLASLt twin that unblocks three sites which could not
previously batch, a DSA indexer barrier/threading fix, and a DSA attend fix
that stopped loading the same latent twice. Together these are the generic
`glm5_next` prefill half of #969.

On top of them, this PR flips one default: the DSA indexer's batched-qidx
prefill GEMM (`ATLAS_GLM_DSA_BATCH_QIDX`) ships **opt-in** instead of
on-by-default. Measured on 2×GB10 NVFP4 TP=2/EP=2, GLM-5.3-Flash, tool-eval
short (seed 42): with the batched GEMM at its original on-by-default setting,
tool-eval scored 97/100, with the one miss on TC-08 (a conditional-branching
tool-calling scenario: after the correct two calls the model appended an
unprompted follow-up question and issued a spurious second weather lookup). With the batched GEMM off, tool-eval scored 100/100 and
TC-08 passes cleanly. The batched-qidx GEMM is numerically close to, but not
bit-identical to, the reference M=1 GEMV-per-row path, and that shift was
enough to flip TC-08. The M=1 GEMV path is therefore kept as the default; set
`ATLAS_GLM_DSA_BATCH_QIDX=1` to opt into the faster batched path.

## Test plan

- [x] `cargo check -p spark-model -p spark-server`
- [x] `cargo fmt --check`
- [x] `ATLAS_SKIP_BUILD=1 cargo clippy --workspace --tests` (one follow-up commit
      adds the `#[allow(clippy::too_many_arguments)]` that the file's two sibling
      GEMM wrappers already carry to `gemm_act_weight_t_out`; the four #969
      commits themselves are unchanged)
- [x] Identity scan (`identity-scan.sh`, base `5d22fcdc5`): 0 findings, 5
      commits, only `SeedSource`/`rsafier` authorship.
- [x] Tested against real hardware: 2×GB10, GLM-5.3-Flash NVFP4, TP=2/EP=2.
      - Prefill throughput, engine-identical prompts (`runs/l3-prefill-ab-n1n2/`):
        at the default 16-row sub-chunk, 61.9→63.7 tok/s (+2.9%) at a 5.4K
        prompt and 58.4→62.4 tok/s (+6.8%) at 21K. With
        `ATLAS_GLM_PREFILL_ROWS=256`, 61.9→83.3 tok/s (+34.6%) at 5.4K and
        58.4→81.2 tok/s (+38.7%) at 21K.
      - Decode throughput unchanged: same-session A/B (`runs/a107-l3-decode-ab/`)
        base vs. candidate within 1% across four probe shapes (13.7–14.2
        tok/s).
      - Six identity probes (p1–p4, `2+2`, pyadd) byte-identical across base
        and candidate at the default 16-row sub-chunk
        (`runs/l3-prefill-ab-n1n2/kbench-l3/`, `runs/a107-l3-decode-ab/`).
        At `ATLAS_GLM_PREFILL_ROWS=256` one of the six (`p3`) changes hash —
        **not claimed to be quality-neutral** at that setting.
      - Tool-eval short, seed 42 (`runs/tc08-bisect-n3n4/`,
        `runs/a111-dsa-batchqidx-n3n4/`): 100/100 at default rows with the
        DSA batched-qidx GEMM off (the default this PR ships); 97/100 with it
        on (the pre-PR default) — TC-08 is the flip in both bisects. At
        `ATLAS_GLM_PREFILL_ROWS=256` the batched-qidx GEMM being off does
        **not** by itself recover 100/100 (`runs/a111-dsa-batchqidx-n3n4/arm3`
        scored 97/100 with the GEMM off and rows=256) — the 256-row setting
        has its own, still-open TC-08 interaction (tracked locally as A112),
        separate from this PR's default flip.

**Not claimed:** no claim that `ATLAS_GLM_PREFILL_ROWS=256` is quality-neutral
— it changes the `p3` identity probe and does not clear tool-eval on its own.
It ships here only as the pre-existing, already-opt-in lever from earlier
work; this PR's own default change is the DSA batched-qidx flip.

## Notes for reviewers

- The four #969 commits are cherry-picked with `-x`, authorship preserved,
  no changes to their content.
- The default-flip commit only changes the DSA batched-qidx gate; the
  cuBLASLt wide-projection gate (`ATLAS_GLM_CUBLAS_PROJ`) and the
  `ATLAS_GLM_PREFILL_ROWS` lever are untouched and still default the way
  #969 shipped them.

## CLA


SeedSource:
Follow-up measurement for the `ATLAS_GLM_PREFILL_ROWS=256` opt-in mentioned in the description (2×GB10 TP=2/EP=2, GLM-5.3-Flash NVFP4, 32K, K=3; tool-eval hardmode, 88 scenarios, seed 42, temperature 0, same session, batched-qidx off in both arms):

- rows=16 (default in this PR): 151/176
- rows=256: 147/176 — twelve scenarios differ, six down one point, four up, two newly failing; none safety-category. Decode unchanged; TTFT lower and concurrency-2 throughput higher at every context depth.

So the default stays 16 here; 256 remains an explicit opt-in for prefill-heavy serving where the small tool-calling delta is acceptable.

PR #1112, opened 2026-09-17
glm5_next: rank-agreed prefix-cache restore + symmetric checkpoints

SeedSource:
## Summary

This series gives the GLM-5.3 (`glm5_next`) prefix cache its missing pieces.
Stacked on the prefill series (#1109), it adds:

- A DSA-indexer aux blob attached to SSM snapshots, so a restored snapshot
  carries the indexer state the DSA path needs, and a freed slot's aux
  cannot outlive it.
- Single-chokepoint slot invalidation: three call sites pop the free list,
  and only one of them previously invalidated the reused slot — now all
  three route through one place.
- An explicit two-switch enable: `--enable-prefix-caching` AND
  `ATLAS_GLM53_PREFIX_CACHE_UNPROVEN` are both required to turn the feature
  on, default OFF, byte-identical to the no-prefix-cache path when off.
- The correctness core — a **rank-agreed restore decision**: each rank
  folds its local gates into a proposal, proposals are exchanged on the
  existing EP wire, and a snapshot is restored only when every rank
  proposes the same token position. Previously the two ranks could decide
  independently and diverge (A100).
- A structural gate so the later exact-leaf bypass probe cannot override an
  agreed decision (A105): it asserts skip == agreed rather than deciding on
  its own.
- Worker-symmetric decode checkpoints via a new EP worker command
  (`0xFFFF_FFF8`), so every rank saves the same `(slot, token, session)`
  checkpoint or none — before this, only the head rank saved a checkpoint,
  so an agreed restore later in the run was rare (A109).
- A hermetic-accessor fix (A110) so the raw `enable_prefix_caching` field
  is read through the existing resolver rather than directly, keeping the
  hermetic-mode guard's "no production code reads the raw field" invariant
  intact.
- An EP opcode move (A113): the new worker-command sentinel collided with
  one DFlash had already claimed; moved to `0xFFFF_FFF8` and asserted
  against the DFlash worker-command band so the two lanes can't collide
  again silently.

## Test plan

- [x] `cargo check -p spark-model -p spark-server`
- [x] `cargo test -p spark-model --lib snap_agree`: 14/14 passed
      (rank-agreement + symmetric-checkpoint protocol unit tests).
- [x] Hermetic-mode test suite (`cargo test -p spark-server --bin spark
      hermetic`, ncclstub env): 18/18 passed, including
      `no_production_code_reads_the_raw_fields_behind_hermetic` (the scan
      that A110 fixes a violation of).
- [x] `cargo test -p spark-server --bin spark prefix_cache` (ncclstub env):
      8/8 passed.
- [x] `cargo fmt --check`
- [x] `ATLAS_SKIP_BUILD=1 cargo clippy --workspace --tests`: clean.
- [x] Identity scan (`identity-scan.sh`, base `2bc93a1a3`): 0 findings, 9
      commits, only `SeedSource`/`Seed` authorship.
- [x] Tested against real hardware: 2×GB10, GLM-5.3-Flash NVFP4, TP=2/EP=2,
      32K context, 4 slots (`runs/l5-prefix-cache-gpu-n3n4/`,
      `runs/l5b-symmetric-ckpt-gpu-n1n2/`, HANDOFF-30,
      ANOMALIES A100/A105/A109/A110/A113).
      - Cache-OFF reference: 14/14 control probes.
      - Cache-ON `[0,T]` recompute and `[T,T]` restore paths byte-identical
        to the cache-OFF reference.
      - Every restore decision was rank-agreed on both rank logs.
      - Minimum host `MemAvailable` observed 4.1 GiB at the 32K shape.
      - 14 two-rank CPU pins covering the agreement + symmetric-save
        protocol (the `snap_agree` suite above).

GPU proof of symmetric checkpoints (2×GB10, 32K, 4 slots): a seed-then-extend sequence produced two rank-agreed restores at token 4048 and then 16336, with identical skip/replay counts on both ranks (4048/12308, 16336/15392) and outputs byte-identical to the cache-off reference; before this series the same sequence voted `[0,16336]` and recomputed. A residual head/worker asymmetry remains for finish-leaf snapshots (the agreement protocol refuses those restores safely); it is called out below and not claimed fixed here.

**Not claimed:** hit-rate or throughput benefit numbers (a separate
evaluation), 8- or 16-slot configurations, or any default-ON behavior. The
feature ships behind two required, default-OFF switches.

## Notes for reviewers

- Opcode registry suggestion (A113): two lanes (this series and DFlash)
  independently picked the same worker-command sentinel value. A single
  assert block, or a small registry, upstream of both would prevent this
  class of collision going forward — flagging for maintainer judgment
  rather than proposing it here.
- `adapter_id` is kept rank-local on the worker save (it keys the radix
  lookup); `session_hash` is adopted from the head rank. This asymmetry is
  intentional, not an oversight — see the A109 commit for the reasoning.
- This PR is stacked on #1109 (prefill); the diff here should show only
  the prefix-cache series once #1109 merges. `impl_a2.rs` is the one file
  both series touch; it merged cleanly in local integration testing
  (`campaign/candidate-v2` @ `202050fc0`).

## CLA


SeedSource:
Validation package results for this series (2×GB10 TP=2/EP=2, 32K, 4 SSM cache slots, spec off), cache ON vs a cache-OFF reference serve:

- identity: 9 cold/warm prompt pairs, all warm outputs byte-identical to the reference
- hit/miss: two prompts sharing a ~31.7K-token prefix report 31,696 / 31,712 cached tokens; a disjoint prompt reports 0
- isolation: two sessions with different prompts each reproduce their own reference hashes
- eviction: 8 distinct prefixes against 4 slots, the first sequence still byte-identical after eviction
- rank agreement: every restore decision identical on both ranks (log-verified); the seed-then-extend case restores at 4048 and then 16336 on both ranks

Not a benefit claim: under this package's exact-repeat TTFT probe the warm request was not faster than cold (exact repeats do not restore by design), so the throughput case is prefix-extension workloads; that measurement is separate from this correctness series.

PR #1114, opened 2026-09-17
KV: the verify path never evicted — one missing call wedged two ranks

rsafier:
**Stacked on #1063** (`pr973-ep-prefix-cache`). Three commits.

## What happened

A BFCL shard wedged both GB10s. Rank 1 logged

```
ERROR EP worker command failed (rank 1); worker STAYS UP: KV cache exhausted: no free blocks
```

and **rank 0 logged nothing**. Both sat at 96% GPU util and ~15 W with logs frozen 43 ms apart, every health endpoint still answering 200, until the run was killed.

## The bug: one missing call

A long run fills the KV pool with prefix-cache blocks that are **evictable by construction** — a finished sequence hands its blocks to the cache (`cache_sequence`) then drops its own reference (`free_sequence`), leaving them at ref 1, exactly what `RadixTreeInner::evict` reclaims. Prefill knows this and allocates via `alloc_block_evicting`.

`verify_hc_rows` called the **raw** `kv_cache.alloc_block()`. No fallback. It failed the instant the free list emptied, standing on a pool full of blocks it was entitled to take. MTP is live here via the **serve preset** (`MTP verify pools widened for lookup drafts, num_drafts=2 pool_width=3`), not the CLI default, so this path runs on every decode.

Nothing else could save it either: preempt-and-retry excludes grammar-active sequences as victims, and BFCL runs grammar-constrained (`parser=qwen3_coder, tools=6`), so every candidate was filtered out and the loop broke immediately.

**Measured** — 300 unique 4.5K-token prompts against a 77,808-block pool:

| | result |
|---|---|
| before | pool hits 100%, then **every** request fails `decode_verify_graphed_k3: KV cache exhausted`, `used=77808 free=0`, server never recovers |
| after | pool reaches 100% and **keeps serving** — `free` oscillates (2 → 615) as eviction reclaims, **zero** exhaustion errors, fresh request returns **HTTP 200** |

No new staleness risk: `return_evicted_block` returns evicted blocks to the free list, so `alloc_block` was already handing out previously-used blocks.

## The wedge: agree the failure across ranks

"Worker STAYS UP" is deliberate (anomalies A60/A62): a failed command is request-scoped, so the head raises the same error, answers the client, and both keep serving — exiting used to wedge the head. **That holds only while the failure is symmetric.** The KV pool and prefix cache are per-rank, so occupancy diverges and one rank can exhaust while the other has room.

Prefill block allocation is now agreed the way #1074 agreed the Marconi anchor: each rank proposes whether it allocated, `ep_min_u32` takes the minimum, and if any rank failed then every rank declines.

🪤 The collective runs **before** the `?`. Propagating the local error first would make the failing rank skip the all-reduce and reintroduce the very mismatch this prevents — the lesson F83 learned for `ep_min_u32`.

Verified live: at exhaustion the server returned HTTP 500 and **stayed serving**, rank 0 idle at 0% / 16.9 W, where previously both ranks pinned at 96%.

## Observability, because none of this was visible

- `/metrics` had prefix-cache counters and **nothing** about blocks — the one resource a long run exhausts had no gauge. Adds `atlas_kv_blocks_{total,free,used}` and a `used_ratio`, published only by the pool the scheduler serves from.
- The error was the same sentence whether the pool was pinned by live sequences, held by an un-evictable cache, or simply too small. The line that distinguished them is `debug!`, and nothing runs at debug in production. It now carries blocks total/free, prefix-cache entries, snapshot count and the sequence's own block count.

Two defects in the gauge itself, both caught by watching it live and fixed in the second commit: several pools published into one global (MTP and DFlash heads each build their own), and two separate atomics tore — observed as `used=1` against `free=65543` of a 77,808-block pool.

## Notes

- Retention measured at ~5 blocks/request with a shared preamble; BFCL carries a per-sample tool schema so nothing is shared and each prompt is retained whole (~80–100 blocks), which is why it filled in ~1,000 requests.
- Spill (`--high-speed-swap`) was never armed and is deliberately untouched here.

spark-runtime 358, spark-model 1440, spark-server 2628, all green.


rsafier:
## Full-run confirmation: 140 consecutive requests through a 100%-full pool

The 300-request probe finished. Stronger than the oscillation figure in the description:

```
after 160 req: used=77808  free=0
...
after 300 req: used=77808  free=0
end: used=77808 (net +77807 blocks over 300 requests)
```

**Requests 160→300 were all served with `free=0`** — a completely exhausted pool — every one a 200, probe exited 0. Eviction reclaims a block per allocation and the pool simply runs at its ceiling.

Before the fix the identical probe aborted at ~250 requests with `decode_verify_graphed_k3: KV cache exhausted: no free blocks` and the server never recovered: every subsequent request 500'd while the gauge sat at `used=77808 free=0`.

Prompt size also grew to 18,042 tokens by the end of the run (the probe's filler tokenizes longer as its index grows), so this holds under a full pool *and* requests ~4× larger than it started with.

PR #1127, opened 2026-09-17
glm5_next: land the prefill stack (#1109) and prefix cache (#1112) on main

SeedSource:
## glm5_next: land the remaining two layers of the GLM-5.3 stack (prefill + prefix cache)

**Why this PR exists.** #1109 (cuBLASLt prefill + DSA indexer fixes) and #1112 (rank-agreed prefix-cache restore + symmetric checkpoints) show as merged, but they merged into their stack base branches (`stack/glm53-02-absorb-q`, `stack/glm53-03-prefill`), not into `main`. #1103 fast-forwarded only the absorb_q layer. Content check on `main` today: no `snap_agree`, no `EP_CMD_DECODE_CKPT`, no `kv_completeness.rs`. This PR carries the two squashes onto current `main` unchanged in behaviour.

**Contents** (two commits, `-x` cherry-picks of the reviewed squashes):
- `c9d3145aa` ← `b1b521ea1` (#1109): cuBLASLt projection path for glm5_next prefill (default ON, `AVAROK_GLM_CUBLAS_PROJ=0` kill-switch), DSA indexer fixes, batched-qidx opt-in, chunked-KDA prefill lever (default OFF). Carries rsafier's #969 prefill slice with authorship preserved.
- `c49c07e44` ← `8966e37e1` (#1112): prefix cache for glm5_next — DSA aux blob, single-chokepoint slot invalidation, two-switch enable (default OFF), rank-agreed restore vote, worker-symmetric decode checkpoints, EP opcode `EP_CMD_DECODE_CKPT`.

**Rebase notes.** `main` rebranded between the stack base and now (#1101): crate `atlas-core` → `avarok-core`, env `ATLAS_*` → `AVAROK_*`. Env names introduced by these commits are renamed to `AVAROK_*` (`AVAROK_GLM53_PREFIX_CACHE_UNPROVEN`, `AVAROK_GLM_CUBLAS_PROJ`, `AVAROK_GLM_KDA_CHUNK_PREFILL`, `AVAROK_SSM_SAVE_DUMP`, plus comments); the startup shim still mirrors legacy `ATLAS_*` names. The only merge conflicts were textual (rename lines inside rewritten hunks); a three-way diff confirmed `main` made no semantic change to `prefix_lookup.rs` relative to the squash's base.

[...]

PR #1129, opened 2026-09-17
fix(grammar): poolside_v1 required string arguments must contain a non-whitespace character

SeedSource:
## fix(grammar): poolside_v1 required string arguments must contain a non-whitespace character

Follow-up to #1103, which made a required string argument non-empty but deliberately left whitespace-only values allowed ("tightening that is a separate call"). Measured on GLM-5.3-Flash with the tool-eval-bench agentic scenario TC-43 (required `web_search.query`, `additionalProperties: false`): under constrained decoding the model emits `{"query":" "}` — a single space — and the scenario still fails. Both the auto-detected and the explicit `--tool-call-parser poolside_v1` paths behave identically, so this is the grammar's scope, not parser selection.

**Change** (grammar-level, so constrained decoding cannot emit it): for required string parameters only,
```
req_value ::= req_leading_ws req_first_nonws value_part*
req_leading_ws ::= [ \t\r\n]*
req_first_nonws ::= <close-avoidance ladder whose first arm also excludes space/tab/CR/LF>
```
via a new `ebnf_until_close_ladder_no_leading_ws` helper mirrored from `ebnf_until_close_ladder`. Optional strings, non-string parameters and the empty-string rule are untouched. qwen3_coder's stricter first-char rule was reviewed and not reused literally because it also excludes `=`/`>` for an unrelated merge-token bug that has no analogue in the poolside envelope.

**Tests** (`grammar/tests/poolside.rs`): `" "` rejected, `"\t\n"` rejected, `" a"` accepted, `"a"` accepted, `""` still rejected, optional-string `" "` unchanged (accepted, as today), required non-string unchanged. Full grammar suite: 108 passed, 0 failed. `cargo fmt --check` and `clippy -D warnings` clean.

No perf paths touched.


SeedSource:
Superseded by #1127, which now carries this commit (`41171de7b`, cherry-pick of `d2d2ff213`). Reason: grammar files invalidate every certification record, so this PR would need its own 12-gate GB10 campaign; folding it into #1127's re-certification covers both with one campaign. Closing; the change lands via #1127.

PR #1131, opened 2026-09-17
DeepSeek-V4.1 Flash: the whole line, one campaign

tbraun96:
Lands the six-PR DeepSeek-V4.1 Flash line (#1094, #1095, #1096, #1097, #1098, #1099) as **one PR and one certification campaign**.

**Why one branch rather than a native stack:** all six are fork PRs from `rrstesiak`, and GitHub will not stack a fork branch — `POST /stacks` requires the base to be an upstream branch. The git graph makes the grouping obvious anyway:

| branch | ahead of main | contains the previous layer? |
|---|---|---|
| `ds41-s1-ingestion` | 6 | — |
| `ds41-s2-reference` | 19 | yes, 0 missing |
| `ds41-s3-sizing-dequant` | 27 | yes, 0 missing |
| `ds41-s3-kquant` | 35 | yes, 0 missing |
| `ds41-s3-streaming` (#1098) | 47 | yes, 0 missing |
| `ds41-s3-graph` (#1099) | 46 | **no — missing 26** |

So #1098 is a clean superset of #1094–#1097 and merges onto main with **zero conflicts**. #1099 forked earlier — which is the real reason GitHub flags it CONFLICTING — and needed exactly **one hunk**.

### The one hunk, and why it mattered

`crates/spark-model/src/layers/attn_v41.rs`: #1098's side had a blocking `gpu.synchronize(stream)` plus synchronous `copy_d2d`; #1099's had `copy_d2d_async` on the stream and no synchronize. A `synchronize` is **illegal inside CUDA-graph capture**, which is precisely what #1099 exists to enable — so taking #1098's side would have compiled cleanly and silently defeated the PR. Kept the async copies; kept `is_multiple_of` over `% != 0`, which is the same predicate and what clippy now demands.

### Six defects a zero-conflict merge could not see

`cargo check --workspace --all-targets` failed after the clean merge:

* `crates/spark-model/src/weight_loader/deepseek_v41.rs` — `use atlas_core::config::ModelConfig`, `atlas_kernels`, `AtlasCudaBackend`
* `crates/spark-model/src/layers/moe_v41_tests.rs` — `atlas_kernels` ×1, `AtlasCudaBackend` ×2

This line forked before #1101 and adds **new** files naming the old crates, while main renamed those crates elsewhere; nothing textual connects the two edits.

**Deliberately not renamed:** `ATLAS_SKIP_BUILD`, `ATLAS_MLA_ABSORB_THREADS`, `ATLAS_FAST_LOAD_BENCH_DIR`. They match the same pattern and are not defects — string literals, they compile, and main still reads every one of them. Renaming them would have broken the ubuntu CI legs that set `ATLAS_SKIP_BUILD`.

### State

`cargo check --workspace --all-targets` **rc=0**. Certification campaign to follow at the frozen head. The six source PRs close with the merge sha once this lands.


tbraun96:
recheck

rrstesiak:
Thomas, one thing before this is certified: `stack/ds41-flash-line` was cut from the six branches as they stood before this morning, so it carries neither the decode fix nor today's merge of main.

What it is missing, all on the fork now and all six PRs mergeable against current main:

- `4a859ae5b`: the single-token MoE path had a data race in `routed_m1` (eight blocks doing a non-atomic add into one accumulator row at m=1). That is the decode nondeterminism and the loops. Replaced by `moe_v41_sum_rows`. Verified: four greedy runs bit-identical, decode versus teacher-forced prefill 97 of 100 top-1, chat suite MinHeap x3 and Volvo x3 byte-identical, 10.6 to 10.9 tok/s warm at 300 tokens.
- `d75da22bf`: the loop watchdog declines its rollback on a layer that cannot rewind, which was the request-2 corruption.
- `69dc4fba9`: `release_state` frees the per-sequence attention buffers.
- main merged into the line with the 500-line allow-list moved into the script per #941, and `scheduler/rollback.rs` allow-listed at 511.

Tips: #1094 `b847fbba1`, #1095 `a05cf6b15`, #1096 `896c8903c`, #1097 `506e8daf7`, #1098 `50c03976a`, #1099 `11aa72eab`. #1139 merges all of it into `stack/ds41-flash-line` in one step: the six conflicting files take the fork side, which already carries the rebrand's names on the fixed code; cargo check is clean and it merges cleanly against current main. Without it the line that lands is the one that loops.

The receipts are on #1099 under the last push.

rrstesiak:
Thomas, before this merges: the sealed head `b91b1876d` is `0719dea47` plus the bench records, and it does not contain the decode fix. `moe_v41_sum_rows` is not on that head; `moe_v41/forward.rs:153` still launches `scatter_add` at m=1, which is the race. The certification gates run on the standard models and never exercise V4.1, so they pass either way.

#1139 is the fix on this branch, redone this afternoon on the line's split layout after Turney's review: three commits plus the single-token test he asked for, which also caught a buffer-sizing bug in `moe_v41/init.rs`. It is mergeable against `stack/ds41-flash-line` now, and Turney's review of it says the fix is right.

If the six source PRs close with this merge as the description says, the fix goes with them. Merging #1139 first, then re-sealing, keeps the line that lands the one that does not loop.

PR #1132, opened 2026-09-18
Control vectors (activation steering) for Qwen3.8-Flash-Next, selectable per request

rsafier:
**Draft.** Activation steering (control vectors) for Qwen3.8-Flash-Next, selectable **per request**. Cut from #1063's head (`4a253306e`), 6 commits.

## Credit

This work is based on **[Cudecnik/Qwen3.8-Flash-Next-refusal-projection](https://huggingface.co/Cudecnik/Qwen3.8-Flash-Next-refusal-projection)** and uses **his control vector**.

Cudecnik did the part that actually required insight: deriving a per-layer refusal direction for `qwen4exp`, and working out where it has to be applied in a hyper-connection architecture — after the HC combine, to every stream, which is not where a reader would guess. His llama.cpp patches (`01-qwen4exp-cvec-hooks.patch`, `02-cvec-projection-mode.patch`), the projection-mode semantics, the `dir=per-layer` / scale convention, and the `llama-cvec-probe` cosine check are all his; this PR ports that mechanism to Atlas and makes the selection per-request. The shipped GGUF is used unmodified.

His repo also supplies the validation idea this PR leans on hardest — checking `mean|cos(h, v)|` either side of the hook rather than gating on model behaviour.

## The headline: this is not a LoRA, and NVFP4 is not in the path

The artifact is 47 unit F32 directions of width 2560 (exactly `qwen4_exp` `hidden_size`), applied

```
project:  h <- h - s * (h . v) * v      (v unit norm; s = 1 fully ablates)
add:      h <- h + s * v
```

at the end of every layer, on all `hc_mult` streams of the residual highway — which Atlas carries as **FP32 `[T, hc_mult, H]`**. No weight is read or written, so none of the LoRA-under-NVFP4 hazards apply (notably the FP4-MMQ arm that silently corrupts a true-valued delta). For the record LoRA is separately unwired for this family anyway: `lora/loading.rs:29` admits only `qwen3_5` dense, `holo3_1_moe` and `qwen3_6_moe`.

## Validated on real NVFP4 weights

Single-rank Flash-Next NVFP4, one boot, no restart between cases:

| request | http | steering |
|---|---|---|
| `{"control_vector": "refusal"}` | 200 | 41 `decode` + 82 `prefill_chunked` + 465 `verify_rows` probe lines, all `cvec=0xdd280f79f47de4df` |
| field omitted | 200 | none |
| `{"control_vector": null}` | 200 | none |
| `{"control_vector": "bogus"}` | **400** | none |

Every path that ran collapsed its cosine: `pre` 0.0105–0.0137 → `post` **0.000000**.

Kernel proven separately against a CPU reference on a GB10: `rel_l2` 9.5e-9 elementwise, `dot(h,v)` 0.149 → 3.9e-7, bit-identical across re-runs. 37 CPU tests, clippy clean on both crates.

## Validated at TP=2 × EP=2 — and it found a real bug

Running this two-node surfaced a **silent divergence bug**, now fixed in `e4f35a9`.

The EP worker allocates its own `SequenceState` and reconstructs `seq_len`, tokens, block table and SSM state by **re-deriving them from the token stream**. That works because each is derivable from tokens. `cvec_id` is not. So rank 1 sat at the default `0`, steered nothing, and diverged from a steering rank 0 — on a highway TP requires to be rank-replicated, so every later all-reduce summed partials computed from different inputs.

Two things made it nastier: `batch_cvec_id`'s uniformity assert **cannot** catch it (on rank 1 every id is uniformly `0`, so it passes), and it was never EP-specific — the gate is `multi_rank_protocol_active()`, true when `tp_size > 1` *or* `ep_size > 1`, so plain TP=2 had it too.

Fixed by appending two `ep_broadcast_u32` words to the `0xFFFFFFF0` prefill preamble — now emitted by a single `Model::ep_broadcast_prefill_preamble` (`e888292`), because there are **five** emitters, not three — `prefill_a_step` has two, plus `prefill_b_step`, `preempt` and `phase_continue_prefills/run_standard`. The worker reads the words unconditionally, so a missed emitter desynchronises the stream; the first cut missed two and rank 1 parsed a token id as the selection. That was caught by a guard added in the same commit — an unresolvable id now fails loudly naming both sides instead of steering one-sidedly.

Validated on dgx-00 + gx10-9959:

| request | rank 0 | rank 1 |
|---|---|---|
| `{"control_vector": "refusal"}` | 41 `decode` + 82 `prefill_chunked` + 3441 `verify_rows` | **identical**, same `cvec=0xdd280f79f47de4df` |
| field omitted | 0 probe lines | 0 probe lines |

Both ranks' probe tables agree to 6dp — `pre` 0.011405 / 0.010491 / 0.013513, `post` 0.000000 throughout. Identical `pre` is the evidence the highways did not diverge.

### Adjacent bug, NOT fixed here

`adapter_slot` has the identical defect: stamped on rank 0, rank 1 keeps `-1` ("defer to active"), and there is no broadcast of it anywhere in the tree. **Per-request LoRA routing under EP/TP is silently wrong the same way.** Out of scope for this branch — it wants its own issue — but reviewers should know the adapter path is a precedent to avoid copying, not to follow.

## The direction transfers to NVFP4

The open question through most of this work was *provenance*: the vector was derived from **unsloth UD-Q2_K_XL** activations under llama.cpp, not from Atlas's NVFP4 graph. Everything the probe proves is that the arithmetic lands — not that the direction means anything on different weights.

**Answered empirically by @rrstesiak: 8/8** on prompts the base model refuses, where the steered arm answers correctly. So a direction extracted from a Q2_K_XL quant is behaviourally live on NVFP4, which is the result that makes this feature worth having rather than merely correct.

Two caveats kept deliberately: that is a behavioural pass rate on one prompt set, not a KL or a regression measurement, and no measurement has been taken of what the ablation costs on *harmless* prompts — the shipped `.json` reports movement on its harmless set too, and a rank-1 ablation at s=1.0 is a blunt instrument.

## The prefix cache was the interesting part

A control vector changes every hidden state, so the KV a request produces under vector A differs from under B or under none — serving one's cached prefix to another is a correctness bug. That looked like a blocker for per-request selection, and it turned out Atlas had **already solved it for LoRA**: `hash_token_prefix` is seeded with an adapter id *and* the radix tree keeps physically disjoint roots per id (`radix_tree/inner.rs:60-126`, whose comment is explicit that the hash alone would still collide on the children map).

Control vectors ride that channel rather than growing a second one. `compose_variant_id` is the **only** place an adapter id and a cvec id combine, because the prefix-cache key and the admission cohort both read the result — computing them apart is how they drift, and the failure of that drift is a cache hit returning another variant's KV, which is silent and reads plausibly. Two byte-identity pins are asserted by test:

- `compose(a, 0) == a` — a serve with no vector keys exactly as before
- `compose(0, 0) == 0` — the base sentinel survives, keeping the zero-fold a strict no-op

The admission cohort now partitions on the composed variant. That is **not** an optimisation: the hook applies one vector to the whole highway, so a batch mixing selections would steer some rows and not others. Same property that forces LoRA's single-adapter batches.

## Design choices worth reviewing

- **The selection is a required parameter, not engine state.** `cvec_after_layer` takes `cvec_id` from the sequence it is processing. An engine-held "currently active" vector would silently apply the previous request's steering on any path that forgot to set it; a parameter cannot.
- **Hooked at the 11 model-level layer loops (14 call sites)**, not inside `ops::hc_post_site`. The funnel was tempting, but it still requires classifying which of a layer's two `hc_post` sites is the layer end, and getting that wrong is silent. Taking the whole `ForwardContext` means the highway base and `hc_row_offset` are computed in one place — the fused decode+prefill step needs two calls at different row offsets, and a single call at offset 0 would steer the decode rows twice and the prefill rows not at all.
- **Absent / `null` / `""` = no steering, a first-class answer.** Deliberately unlike the LoRA pool, which has no selector meaning "base" once an adapter is resident. Steering has to be switchable off per request or it cannot be A/B'd on one serve.
- **Fails closed, loudly.** Unknown vector name → 400 listing what is registered. Malformed/misshaped file → boot aborts. A non-mHC model rejects the flag rather than accepting it and doing nothing. A silently-skipped projection serves a model the operator believes was steered, and no counter would ever show it.
- `AVAROK_CVEC_PROBE=1` logs `mean|cos(h,v)|` either side of every application, tagged with the forward path. That is what makes it a *coverage* proof: a path nobody wired simply never appears in the log.

## Reviewer orientation

1. `docs/design/qwen4exp-control-vectors.md` — mechanism, insertion points, traps, and the §8 per-request design. Start here.
2. `crates/spark-model/src/control_vector_registry.rs` — the identity and the one composition function.
3. `crates/spark-model/src/model/control_vector_hook.rs` — the hook and why the selection is a parameter.
4. `kernels/gb10/qwen3.8-flash-next/nvfp4/control_vector.cu` — `CVEC_BLOCK` is pinned; the dot is a block reduction, so its rounding (and on a speculative path, the accepted text) is a function of `blockDim.x`.

## Not done / known gaps

- Three forward paths are exercised so far. See "Validated at TP=2 × EP=2" above for what that covers.
- Three forward paths were exercised (`prefill_chunked`, `decode`, `verify_rows`). The other nine are listed as un-exercised by the probe rather than silently passing; they need multi-seq, mixed-step and TP runs.
- No KL measurement against the un-steered serve has been taken (the behavioural question is answered — see "The direction transfers" above — but the distributional cost is not quantified).
- Anthropic and Responses surfaces accept requests and pass `None`; adding the field is one line in each `to_ir`.
- Decode graphs stay off while any vector is registered (a captured graph bakes in the vector pointer). Belt-and-braces for qwen4_exp, which does not currently use decode graphs.


rsafier:
I reviewed PR #1132 at head `ad3af79b`. The core design is strong—especially the composed cache identity, cohort partitioning, centralized EP preamble, and the probe-based coverage—but I would fix a couple of correctness issues before merging.

* **Blocker — multi-rank consistency is not actually verified.** `control_vector_registry.rs:48` makes `cvec_id` from the **name only**, and `impl_a2.rs:771` merely verifies that the worker has that ID. Rank 0 and rank 1 can both register `"refusal"` while loading different files, scales, modes, or layer ranges; the IDs still match and Atlas silently steers the ranks differently. Logging the SHA isn't enough. Make the ID/fingerprint depend on `name + file SHA256 + mode + scale.to_bits() + layer_start + layer_end`, or perform an explicit registry-fingerprint handshake across ranks. This is especially important because the PR explicitly claims to fail closed on rank mismatch.
* **High — the default configuration differs from the configuration actually validated.** The design doc says the shipped refusal projection is `project`, scale `1.0`, **layers 4..44**. But `control_vector_args.rs:132` defaults to `(1, n_layer - 1)`, i.e. **1..47** on this model, and a test explicitly locks that in. So `--control-vector refusal=file.gguf` does not run the configuration described and tested by the PR. I would either require the layer range when the artifact doesn't encode its active range, or encode/read the recommended range from metadata. I would not infer “all tensors present” = “all layers should be active.”
* **High — “per-request” support is incomplete across Atlas APIs.** Chat Completions and Completions carry the selector, but `responses_lowering.rs:154` and `anthropic/to_ir.rs:223` hard-code `control_vector: None`. Since the plumbing is already in the IR, I would add the field now rather than establish divergent API behavior.
* **Medium — the loader doesn't validate the control-vector identity metadata.** `control_vector.rs:126` parses the GGUF and checks tensor shapes/names, but doesn't enforce `general.architecture=controlvector` or validate `controlvector.model_hint`. A vector for another model with coincidentally compatible geometry can get through. That doesn't fit the PR's otherwise good “fail closed” philosophy.
* **Medium — teardown ownership is incomplete.** The vector table is a raw `gpu.alloc`, but `release_pools()` doesn't explicitly release it. Atlas's backend sweep will reclaim it, so this isn't a permanent GPU leak, but normal model teardown will treat it as an unowned allocation and warn. The registry should have an explicit `release(gpu)`/`ModelResource` path.

### Yes — I think Atlas should have an explicit on/off concept

It **technically already has request-level off**:

```json
{}
```

or

```json
{"control_vector": null}
```

or even `""` all resolve to `cvec_id = 0`. “On” is currently expressed by selecting a vector:

```json
{"control_vector": "refusal"}
```

I wouldn't replace that with `true/false`, because Atlas can load multiple vectors and `"control_vector": true` doesn't say **which** one.

But I would add a proper **server default + explicit override** model. Crucially, the current `Option<String>` isn't enough because it collapses “field omitted” and explicit `null` into the same value. Use a tri-state directive:

```text
omitted       -> use server default
null / false  -> explicitly OFF
"name"        -> explicitly use this vector
```

Then add something like:

```text
--default-control-vector refusal
--disable-control-vectors
```

with default behavior remaining **off** when no default is configured. A runtime endpoint/TUI control could then change the default without unloading the already resident vector:

```text
GET  /v1/control-vectors
POST /v1/control-vectors/default
     {"name":"refusal"}

POST /v1/control-vectors/default
     {"name":null}
```

That gives you four useful states: vectors not loaded, loaded but default-off, default-on, and globally disabled. Individual requests can still override the default.

There's also one subtle reason a real global off switch is useful: `decode_graphs_allowed()` currently returns false whenever **any vector is registered**, even for requests whose vector is off. The PR says that's speed-neutral for this Qwen model, but architecturally “off” isn't quite baseline execution today. A global disable should restore graph eligibility too.

The current CI is red and the PR is still draft, although several failures I inspected look unrelated to this feature; `fmt` and `clippy` pass, while the workspace test failure I saw is in pre-existing `env_compat` tests. So I wouldn't treat the red CI alone as evidence against this implementation.

The **rank fingerprint problem** is the one I'd definitely leave as a blocking GitHub review comment. The **1–47 vs 4–44 default** is the other one I'd want resolved before this becomes an API users rely on.

rsafier:
Both blockers are fixed in `wip/qwen4exp-cvec-derive` (#1134, stacked on this
branch) at `72f4974c7`. Every claim in the review reproduced at the exact lines
cited — I checked each against the code before acting on it, and all six held.

## Blocker — multi-rank consistency

Fixed as you suggested: the id is now hashed over the name **and**
`ControlVector::config_identity()` = file SHA-256 + mode + `scale.to_bits()` +
inclusive layer range. Divergent configuration therefore produces divergent
ids, which makes the `impl_a2.rs` check stop being decorative and start failing
closed.

You were precisely right that the PR's "fails closed" claim was too strong: it
held for a **missing name** only. A name present on both ranks with different
configuration matched, and nothing reported it.

Verified on hardware rather than only in tests — rank 1 booted with `refusal`
at scale 0.5 against rank 0's 1.0, same name, same file, same layers:

```
rank 0  id=0xf4dee17672d911a5 ... scale=0x3f800000 layers=4..=44
rank 1  id=0xab59fa0c1a9890ad ... scale=0x3f000000 layers=4..=44
```

and the worker refused the request. Under the old scheme both were
`0xdd280f79f47de4df`.

Every rank now also logs its full identity line at boot, so a divergence is one
`diff` rather than an inference across four log lines.

Two things that run surfaced:

- **The guard's message was now actively misleading** — it compared ids but
  printed *names*, so it listed `'refusal'` among the registered vectors in the
  same breath as saying it was not registered. With a fingerprinted id,
  "name present, configuration different" is the common case. It now prints
  every local identity in full and names that as the likely cause.
- **Not fixed, and worth tracking: the divergent request _hangs_** until client
  timeout rather than erroring back. It fails closed in the sense that matters
  — the model never runs half-steered — but the proper fix is the boot-time
  fingerprint handshake you offered as the alternative, so it fails before any
  request is accepted. I did not bolt it on here: it needs a new message in the
  EP preamble protocol, and doing that carelessly is what caused a rank-1
  misparse earlier in this stack. It deserves its own change.

This also retires the `cvec_id` generation-counter follow-up I had filed against
#1134 — a different file under an existing name now yields a different id, so
stale prefix-cache entries can't be served as hits, and dynamic loading no
longer needs a counter bolted on first.

## High — default configuration differs from the validated one

Also confirmed: `(1, n_layer - 1)` = 1..47, while everything is characterised at
4..44.

I took a slightly different shape from either option you listed, after checking
with the author. Resolution is now **flag → curated per-model value → error**,
and never inference:

- `qwen4_exp` maps to 4..44 because *all three* known vectors for it use that
  range — the published refusal projection and both derived in `examples/` —
  which makes it a property of the model rather than of any one vector.
- A model absent from the table gets **no** default and the flag is required.
- A checkpoint too small for its curated range falls through to requiring the
  flag rather than silently clamping.
- The chosen range is logged whenever it comes from the table, so it is never a
  silent substitution.

This keeps your actual objection satisfied — nothing is inferred from "all
tensors present" — while not making the one currently-supported model awkward to
run. Reading a recommended range from GGUF metadata is still the better long-term
answer and is on the list.

## The rest

Agreed on all of it and queued, in this order:

1. **Tri-state directive** + `--default-control-vector` / `--disable-control-vectors`.
   Your point about `Option<String>` collapsing omitted and explicit null is the
   reason to do this *before* anyone depends on the wire format. Default-off
   preserves today's behaviour exactly, so it is additive now and breaking later.
2. **Responses + Anthropic** — confirmed both hard-code `None` at the lines you
   gave, while `openai/to_ir.rs:111` carries it. Both request types need the
   field added.
3. **GGUF identity metadata** — confirmed neither `architecture` nor
   `model_hint` is checked anywhere in `control_vector.rs`.
4. **Teardown** — confirmed `release_pools` frees derived weights, ssm
   snapshots, ssm state pool, kv cache and buffer arena, but not the registry.
   One addition: `ControlVectorCapture` is *also* a raw `gpu.alloc` and equally
   unreleased, so the `release(gpu)` path should cover both.
5. **`decode_graphs_allowed`** — confirmed it is `control_vectors.is_empty()`.
   Beyond restoring eligibility under a global disable, the admission cohort
   already guarantees a batch is homogeneous in `cvec_id`, so eligibility could
   key on the *batch* rather than the registry and let unsteered batches use
   graphs even with vectors resident.

One thing the review didn't cover that cost a full sweep here, so it is going on
the list too: **`add` mode against a unit-magnitude file is a silent
fail-open.** `project` folds the row norm into the per-layer scale, so unit rows
are correct there; `add` applies the row verbatim under one global scalar, so
the same file at scale 1.0 is a ~0.01% nudge that reads as "the vector does
nothing" — a dose of zero published as a null result. Fix is a magnitude
convention in the file metadata plus a mode-compatibility check; it is also
detectable without metadata, since all-per-layer-norms ≈ 1.0 under `add` means
someone grabbed the projection file.

rsafier:
Following up on the code review above — the fixes landed on #1134 (which builds on this branch), not here, so pointing at them from the original review:

- **P0** — `cvec_id_hash` now covers the vector's **configuration**, not just its name: file sha256, mode, scale bits and layer range. That was the review's central point and it was correct: two ranks could each register `"refusal"` from different files or at different scales, the ids matched, the worker's id check passed, and the halves steered differently with nothing to report it.
- **P1** — `control_vector` is now a three-state directive, so an omitted key, an explicit `null`, and a name are distinct. Validated on the wire against a live server default, including the assertion that matters (`omitted != null`).
- **Loader hardening** — fails closed on the wrong file, the wrong model, and the wrong operator (a unit-normalised file used in `add` mode is rejected outright).
- **New** — a boot-time registry handshake. Making the id configuration-sensitive turned a silent mis-steer into a per-request refusal, which is better but still manifests as a hung caller, since the head cannot turn a worker's refusal into a response. Both ranks now agree on an `all_gather`'d registry fingerprint at boot and refuse to start if they differ.

Full results, including the negative test where both ranks refuse a deliberately mismatched pair: https://github.com/Avarok-Cybersecurity/atlas/pull/1134#issuecomment-5730967113

tbraun96:
Triaged for a certification campaign and **parked pending a rebase** — but this is the cheapest rebase in the backlog, so it is worth doing soon.

#1132 ⊂ #1134, so **one rebase covers both** (#1134 is the tip). You fork only **26 commits** behind main — nearly current — and merging #1134 onto `5f639bfc57` locally gives 37 conflicts that are almost all one mechanical shape, not real divergence: **you did the Atlas→Avarok rename yourself, and main did it too** (`569ffd0920`). Your whole diff against `.github/workflows/ci.yml` is four lines (`launch-atlas-glm53` → `launch-avarok-glm53`), and against `file-size-cap.yml` four more ("Atlas idiom" → "Avarok idiom"); main changed the same two files by 319+/40− and 8+/633−.

**Take main's copy wholesale wherever your only diff is an atlas→avarok swap** — `git checkout avarok/main -- <file>` loses nothing there. Same for `bench/ladder38/published.json`: regenerate rather than hand-merge. #1063 hits the identical three files.

Also raise `merge.renameLimit` before you estimate — git's default gives up on rename detection across #1101 and reports a misleadingly short conflict list. Full write-up: #1206.

Campaigns are composed and queued for 22 PRs; anything rebased before one launches shares its GPU run.

Issue #1133, opened 2026-09-18
Per-request LoRA routing is silently wrong under EP/TP: adapter_slot never reaches rank 1


rsafier:
## Summary

Per-request LoRA routing is **silently wrong under EP or TP**. Rank 0 stamps `seq.adapter_slot` at admission; rank 1 never learns it and keeps the default `-1` ("defer to the installed active adapter"). Nothing broadcasts it — a grep of the whole tree for any EP broadcast of `adapter_slot`, `adapter_id` or a variant id returns nothing.

So a request routed to pool slot 2 on rank 0 is served with the *active* adapter's weights on rank 1. Under TP the LoRA B-matmuls are sharded and all-reduced, so the ranks sum partials computed from different weights. The output is well-formed text, and no counter shows it.

Found while adding the equivalent transport for control vectors (#1132), which had the identical defect and is fixed there.

## Why rank 1 doesn't just know

The EP worker allocates its own `SequenceState` (`main_modules/serve_phases/build.rs:296-303`) and reconstructs `seq_len`, tokens, block table and SSM state by **re-deriving them from the token stream**. That is sound for all of those, because each is derivable from tokens.

`adapter_slot` is not derivable from tokens. Neither was `cvec_id`. That is the bug class: any per-request value that changes forward-pass arithmetic and cannot be recomputed from the prompt must be *transported*, and today the prefill preamble carries only `chunk_len`, `chunk_start`, `full_len` and the tokens.

## Evidence

- Stamped on rank 0: `scheduler/prefill_a_step.rs:154`, `scheduler/prefill_b_step.rs:127`.
- Rank 1's default: `model/trait_impl/meta.rs:365` and `traits.rs:387` — `adapter_slot: -1`.
- Consumed on rank 1 regardless: `model/impl_b1.rs:644` (`upload_seq_slot_uniform(seq.adapter_slot, …)`) and `model/impl_lora.rs:114` (`moe_lora_route(seq.adapter_slot)`).
- No transport anywhere: no `ep_broadcast*` of any adapter identity exists.

## Not EP-specific

The gate is `multi_rank_protocol_active()` (`model/impl_a2.rs:500`), true when `ep_world_size > 1` **or** `tp_world_size > 1`. A pure-TP deployment (`--tp-size 2 --ep-size 1`) runs the same worker loop and has the same bug.

## Suggested fix

[...]

PR #1134, opened 2026-09-18
Derive control vectors from Atlas activations: native capture, a worked verbosity example, and per-operator magnitudes

rsafier:
Stacked on #1132 (`wip/qwen4exp-control-vector`), which added the apply path
and per-request selection. This branch adds the other half: **deriving a
control vector from Atlas's own activations**, plus a worked example and the
tooling to tell a real direction from a perturbation.

## What's here

**Native derivation.** Capture per-layer mean activations during prefill,
subtract two corpora, normalise, write a llama.cpp-compatible GGUF. No
gradients, no optimiser, no fine-tune — a control vector is a contrastive mean
difference, and the expensive part is one forward per prompt.

The published vector for this model was derived from a **Q2_K_XL** quant under
llama.cpp. It transfers (measured, 8/8), but a direction taken from the
activations Atlas actually serves has no transfer question at all. It also
avoids needing llama.cpp's `l_out` export patch, which exists only because a
hyper-connection model has no natural `[n_embd, n_tokens]` layer output for the
stock generator to read.

**Capture accumulates the STREAM MEAN** — the mean over the `hc_mult` highway
rows — in FP64 on device. Not arbitrary: the apply path projects every stream
individually and projection is linear, so a direction derived from the stream
mean is the one that composes with it. A direction from a single stream would
live in a different basis than the one being steered.

**A worked example** (`examples/control-vectors/verbosity/`): 200 line-by-line
paired prompts, the capture driver, the derivation, and a measurement.
Verbosity was chosen because the effect is a *number* — it moves
`completion_tokens`, so the demo is falsifiable rather than admirable.

**`--magnitude raw`**, because the two operators consume magnitude differently:

| operator | applies | needs |
|---|---|---|
| `project` | `h -= s·(h·v)v` | direction only — `load()` folds the row norm into the per-layer scale, so magnitude cancels |
| `add` | `h += s·row` | the row verbatim, under **one global scalar** |

Unit rows hand that scalar the job of dosing every layer at once, and the
contrast is not uniform with depth: `|mean-diff|` runs 0.20 → 2.13 over layers
4..44 (11×) while the stream norm grows 0.75 → 7.40. Any single scalar is then
wrong at nearly every layer. Default stays `unit`, so existing project-mode
files and commands are byte-identical.

**`scripts/inspect_capture_dump.py`** reads the FP64 accumulators between
capture and derive and reports the stream magnitude the GGUF no longer carries.
It exists because of two traps hit during this work:

- The accumulator is a **sum**, with the token count in the header. The corpora
  have different counts by construction (8013 vs 5701), so subtracting raw sums
  subtracts one corpus's worth of tokens from another's and calls the residue a
  direction.
- Sizing an `add` scale without measuring first. Scales of 2/8/32 were staged
  against a stream whose magnitude had not been checked; they were ~0.02% of it.
  That would have produced "add mode does nothing" — **a dose of zero published
  as a null result.**

## Derivation quality

Derived from 8013 positive / 5701 negative tokens:

| | this vector | published refusal vector |
|---|---|---|
| adjacent-layer cosine | **0.9794** | 0.968 |
| all-pairs off-diagonal | **0.7429** | 0.722 |
| `|diff|/|act|` across depth | 0.27 – 0.39 | — |

The contrast is a steady ~third of the stream magnitude at *every* one of 48
layers. That is a coherent feature carried across depth, not 41 independent
per-layer fits.

### Verbosity — `add` is a signed length dial

Median Δ `completion_tokens` vs unsteered, 6 prompts, temp 0, paired per prompt:

| scale (`add`) | Δ | |
|---|---:|---|
| −0.10 | −90.0% | too far — answers end mid-sentence **with `finish_reason: stop`** |
| **−0.05** | **−54.9%** | best concise setting; complete, dense, high quality |
| −0.02 | −20.8% | |
| +0.02 | +63.5% | |
| +0.05 | +225.4% | essay territory |
| +0.10 | ≥ +484% | saturates the cap, begins script-mixing |

Strictly monotone through zero, 6/6 sign agreement in every arm, no truncation
in the usable band. At −0.05 the model produced a *better* answer than baseline
in 53% of the tokens.

### Comment density — a length-invariant confirmation

Median comment ratio (comment lines ÷ total lines in fenced blocks), 8 held-out
coding tasks:

| scale (`add`) | ratio | code blocks |
|---|---:|---|
| −0.25 | — | 4/8 — **disqualified**, stopped writing code |
| **−0.15** | **0.000** | 8/8 |
| −0.10 | 0.023 | 8/8 |
| −0.05 | 0.039 | 8/8 |
| *unsteered* | *0.051* | 8/8 |
| **+0.05** | **0.236** | 8/8 |
| +0.10 … +0.25 | 0.278 … 0.341 | 7/8 → 4/8 |

Monotone across every arm that still produced code. Quality verified by reading
the output, not inferred from the ratio: at −0.15 a CUDA kernel came back in 74
tokens, correct, bounds guard intact; at +0.05 the same kernel came back with
Doxygen docs and substantive comments.

### `project` ablates; `add` traverses

The verbosity project ladder was uninterpretable because *every* project arm
lengthened output, including an unrelated vector at +8.6% — perturbation
lengthens, so a length metric cannot separate the axis from the steering. A
**ratio** metric has no such confound, which makes this the clean test:

| arm | comment ratio |
|---|---:|
| `project` 0.5 | 0.074 |
| `project` 1.0 | 0.074 |
| `project` 2.0 | 0.052 |
| `project` 4.0 | destroyed |
| **`refusal` `project` 1.0 — unrelated direction** | **0.072** |
| unsteered | 0.051 |
| `add` −0.10 / +0.05 | 0.023 / 0.233 |

Projecting the comment-density direction (0.074) is **indistinguishable from
projecting an unrelated one** (0.072). Scales 0.5 and 1.0 give identical
results — no dose response. `project` at 2.0, which reflects the component and
should push negative, lands on baseline. Meanwhile `add` spans a 10× range.

This is expected from the arithmetic: `h -= s·(h·v)v` is **quadratic in `v`**,
so `v` and `−v` are identical and there is no direction to travel in. Choose the
operator from the behaviour — **ablate a feature with `project`** (refusal:
there is no useful "more refusal"), **traverse an axis with `add`**.

### Honest limits

- The ratio metric is not forgery-proof either: the unrelated vector moved it
  0.051 → 0.072. That is the perturbation floor. The comment-density vector
  reaches 8.8× it, so the signal is clear — but the floor is not zero.
- The comment-density direction is **correlated with verbosity**, not
  orthogonal. At +0.05 code lines also grow (28 → 84 on one task) and the model
  expands scope. The ratio rises by more than the line count does, so the
  comment effect is real, but this is not "same code, more comments".
- Two of eight tasks show no comment effect at any scale — a Rust one-liner too
  trivial to comment, and SQL, which never takes `--` comments here.

### Determinism

The unsteered baseline reproduced **byte-identically across three runs and two
reboots**, and four arms re-measured after a further reboot matched to within
0.003 (`add −0.10` 0.023→0.023, `add +0.05` 0.236→0.233, baseline 0.051→0.051,
refusal 0.072→0.072). The noise floor is zero, so all scatter is attributable
to steering.

### Published

Both vectors are public on the Hub:
**https://huggingface.co/MonumentalSystems/Qwen3.8-Flash-Next-control-vectors** — each as two GGUFs,
`vector.gguf` (unit, for `project`) and `vector-add.gguf` (raw magnitudes, for
`add`). Cudecnik's refusal vector is credited but **not** redistributed.

## Known limitation — `cvec_id` has no generation counter

`cvec_id_hash` hashes the vector **name** only. The resulting id is what gets
transported to rank 1 in the prefill preamble and, via
`compose_variant_id(adapter_id, cvec_id)`, what partitions the prefix cache and
the admission cohort.

So if a *different file* were ever loaded under an *existing name*, the id would
be unchanged and cached KV computed under the old vector would still be served
as a hit — silently, since nothing in the key distinguishes the two.

LoRA already avoids this: `adapter_id_hash(name, generation)` folds in a
generation counter so a reload mints a fresh id and the old entries simply stop
matching.

**This cannot bite today.** Registration is boot-time only, and a reboot clears
the cache, so within any single serve a name maps to exactly one file for the
whole lifetime of the process. It is recorded here because it is a precondition,
not a bug report: **dynamic vector loading must not land before the generation
counter does.** The moment reload-under-an-existing-name becomes reachable, this
becomes a correctness defect that presents as a cache hit rather than as an
error.

Related: #1133 tracks the same latent hazard in LoRA's per-request routing under
EP/TP.

## Mixed-vector traffic serialises

`filter_adapter_cohort` (`scheduler/admission.rs`) restricts a batch to one
**composed variant** (`adapter_slot ⊕ cvec_id`). A request naming a different
variant is pushed back to the front of the pending queue and waits for the
in-flight batch to drain.

This is required by the current kernel, not incidental.
`cvec_project_highway` takes a single `v` and a single `scale` per launch over a
grid of (tokens × hc rows), so one launch steers every row identically. A batch
mixing selections would apply one caller's vector to another caller's
sequence. LoRA carries the same constraint for the same reason.

| traffic | behaviour |
|---|---|
| all requests share a vector, or none steer | full concurrency to `max_num_seqs` |
| requests name **different** vectors | serialised, one cohort at a time |

Observed directly during this work: a sweep issuing a continuous stream of
single requests, each a different arm, is a worst-case neighbour — an unsteered
request from another client can never join the in-flight cohort, so the two
ping-pong and both run at C=1.

**Not a defect, and not free.** A deployment where one vector dominates (say,
`refusal` on, or off, for everyone) batches normally and pays nothing. A
deployment where callers pick freely among several vectors loses batching
between them.

The unlock is per-row steering: a per-sequence vector index and scale so one
launch can apply different vectors to different rows. The launch signature has
no room for it today, so it is a kernel-plus-plumbing change rather than a
flag, and it is out of scope here. Worth its own issue if mixed-vector
concurrency needs to be a supported mode rather than a serialisation point.

## Credit

The method and the reference artifact come from
**https://huggingface.co/Cudecnik/Qwen3.8-Flash-Next-refusal-projection** —
Cudecnik's refusal-projection control vector, which is also the positive control
used above.

## Draft because

Still draft, but no longer because the numbers are unsettled — both ladders are
monotone, quality-checked by reading the output, and reproduce across reboots.

Remaining before this is ready to merge:

- `cvec_id` needs the generation counter (see above) if dynamic loading is ever
  wanted; harmless today because registration is boot-time only.
- The comment-density corpus is one contrast on one model; the published
  numbers are one prompt set on one serve.
- Nine of twelve forward paths are still un-exercised by the coverage probe,
  which runs with `AVAROK_CVEC_PROBE=1` and is too slow for serving.


rsafier:
Demo control vectors: https://huggingface.co/MonumentalSystems/Qwen3.8-Flash-Next-control-vectors

rsafier:
## P1 landed, plus a boot-time registry handshake (`838e562aa`)

Two things since the last update: the three-state `control_vector` field from the review, and a fix for a hang that the review's P0 work exposed rather than caused.

### The hang, and why the first fix was only half a fix

Each rank builds its control-vector registry from its own command line, so the two halves of an EP/TP serve can disagree about what `vcd_m15` means. The per-request check in `impl_a2` already refuses an id it cannot resolve, so a divergence never steers half the model — but it refuses **during a request**, and the head has no path to turn a worker's refusal into a response. The caller waits for its own timeout.

This was not hypothetical. While testing, I re-derived a vector, copied it to rank 0 and forgot rank 1. Different file → different sha → different config identity → different id → the worker refused, and the request simply stopped. Failing closed as a thirty-minute hang is still a bad failure.

So the check moved to boot, where both ranks run the same line and a collective is a natural barrier. If the registries agree there, no id the head can later send is unresolvable on a worker.

**The part worth reviewing:** my first implementation broadcast the head's fingerprint and had the workers compare. That is only half a handshake — only a worker can detect a divergence. On a mismatch rank 1 exits, rank 0 logs `agrees across 2 ranks` (agreement it never verified) and serves on with a dead worker. Same hang, one layer further out, now with a log line that actively misleads. I caught this reading my own code back before running the negative test, not from the test.

It is now an `all_gather`: every rank contributes its fingerprint, every rank receives all of them, every rank can refuse. `all_gather` is also the correct primitive because it moves `Uint8` — `all_reduce` on this comm is hard-wired to `Bfloat16` (it is the activation reducer) and would push identity hashes through a float.

The fingerprint is a sorted FNV over registered ids. Each id already covers name **and** configuration (file sha256, mode, scale bits, layer range) from the P0 change, so this catches a different file, scale, mode or range under the same name — not merely a missing one. Sorted, because two ranks given the same vectors in a different command-line order are not misconfigured; insisting they match would turn a harmless difference into an outage. Empty registries fingerprint to `0`, so "neither rank has any" agrees without a special case.

### Measured on the 2-rank qwen3.8-flash-next NVFP4 TP2×EP2 box

Matched pair — boots and serves:

```
control-vector registry agrees across 2 ranks (fingerprint 0xa16c864b3474f774)
```

Rank 1 deliberately given `refusal` at scale 0.5 where rank 0 has 1.0 — same name, same file, same layers, i.e. exactly the divergence a name-only id could not see. **Both** ranks now refuse at boot, each printing the full table with itself marked:

```
rank 0                                    rank 1
Error: control-vector registry handshake  Error: control-vector registry handshake
  registries differ across ranks:           registries differ across ranks:
    rank 0: 0xa16c864b3474f774 (this rank)    rank 0: 0xa16c864b3474f774
    rank 1: 0x4c4e171c488a82a7                rank 1: 0x4c4e171c488a82a7 (this rank)
```

Previously: rank 1 refused per-request, and the caller hung.

### P1 — the three-state field, validated on the wire

`control_vector` is no longer `Option<String>`; it is a directive that distinguishes a missing key from an explicit `null` (serde `default` + `deserialize_with`, the only way to tell those apart). With `--default-control-vector vcd_m15` serving, against a real prompt:

| field sent | HTTP | tokens | comment ratio |
|---|---|---|---|
| *omitted* | 200 | 63 | 0.000 |
| `null` | 200 | 149 | 0.000 |
| `false` | 200 | 149 | 0.000 |
| `""` | 200 | 149 | 0.000 |
| `"vcd_m15"` (= the default) | 200 | 63 | 0.000 |
| `"vcd_p15"` | 200 | 400 | 0.217 |
| `true` | 400 | — | ``control_vector: `true` does not say WHICH vector`` |
| `"ghost"` | 400 | — | `unknown control_vector 'ghost'; registered: …` |

- `omitted == explicit default` (63 = 63)
- `omitted != null` (63 vs 149) — the entire reason the field stopped being an `Option`, and the one thing a unit test on the enum cannot prove about the wire.

The server default chosen for this test is deliberately one whose effect is *measurable* (it strips comments from generated code), so "the default applied" is distinguishable from "nothing happened" by the output rather than by trusting a log line. Caveat on one row: `vcd_p15` hit exactly `max_tokens=400`, so it saturated — its 0.217 ratio shows steering is live, but its token count is not a dose measurement.

### Tests

`fingerprint()` is split into a free `fingerprint_of_ids()` so it is testable without a GPU — the registry cannot be populated without one, and a boot-time refusal is a bad thing to leave untested. Five new tests: the empty sentinel, order independence, a missing / extra / changed vector, and a real scale change end to end through the id hash. 37 control-vector tests pass; clippy clean.

### Still open

- **P3** — `decode_graphs_allowed` could key on the batch's `cvec_id` since cohorts are homogeneous. Note that `--disable-control-vectors` already empties the registry, so graph eligibility answers itself in the case that matters most.
- **Concurrency**, carried from earlier: `filter_adapter_cohort` restricts a batch to one composed variant, so traffic mixing different vectors (or vector vs none) serializes rather than batching. Correct, but it is a throughput cliff worth knowing about before anyone runs an A/B under load.

rsafier:
## Concurrency scoping: decode throughput at C=1/2/4, control vector on vs off

Measured to decide whether it is worth lifting the control-vector restriction in `filter_adapter_cohort` (`scheduler/admission.rs:163`), which today defers any request naming a different vector until the in-flight batch drains. **Conclusion: the throughput case is weak; the latency case is the real one.** Details and all raw data below.

### Fingerprint

| | |
|---|---|
| build | `838e562aa` (`wip/qwen4exp-cvec-derive`), binary sha256 `4b7f1ae5…`, byte-identical on both ranks |
| model | `qwen3.8-flash-next-nvfp4-tp2-ep2`, TP2×EP2 over 192.168.177.11/.12 |
| serve | `--max-num-seqs 4 --max-seq-len 262144 --kv-cache-dtype bf16 --gpu-memory-utilization 0.65 --enable-prefix-caching --default-control-vector vcd_m15` |
| env | `AVAROK_DEBUG_NO_GRAPH=1`; CUTLASS grouped + shared MoE explicitly **OFF** (server warns this "departs from the validated configuration") |
| harness | streaming, temp 0, `reasoning_effort: none`, `max_tokens 1024`, 4 distinct code-generation prompts, n=3 per arm |
| date | 2026-09-18 |

### Method note: the first harness was wrong, and why

v1 assumed every request would saturate `max_tokens`, so all slots would retire together and a first-token-to-last-token window would be valid. That assumption is false on this model (see the temp-0 section below): slots retire at different times, so the window's tail runs at less than width C and understates aggregate throughput for reasons unrelated to concurrency.

v2 measures the **steady state**: every token's arrival time is recorded, and the window is `[max(first token), min(last token)]` across the wave — the interval during which all C slots were provably streaming. `occ` reports that window as a fraction of the shortest request's own span; it is ~1.00 in every arm below, so the windows are sound.

### Results — medians, n=3

| arm | C | agg tok/s | per-req tok/s | occ | scaling vs C=1 |
|---|---|---|---|---|---|
| cvec OFF | 1 | 54.43 | 54.43 | 1.00 | 1.00x |
| cvec OFF | 2 | 66.36 | 33.18 | 1.00 | 1.22x |
| cvec OFF | 4 | 62.41 | 15.60 | 1.00 | 1.15x |
| vcd_m15 | 1 | 55.11 | 55.11 | 1.00 | 1.00x |
| vcd_m15 | 2 | 67.03 | 33.51 | 0.97 | 1.22x |
| vcd_m15 | 4 | 79.85 | 19.96 | 0.97 | 1.45x |

Raw, all 18 waves (`tokens` = per-slot completion counts):

```
arm                 C rep  agg tok/s  per-req   occ   TTFT  tokens
cvec OFF (null)     1   0      54.46    54.46  1.00   0.33  [1024]
cvec OFF (null)     1   1      54.36    54.36  1.00   0.35  [1024]
cvec OFF (null)     1   2      54.43    54.43  1.00   0.34  [1024]
cvec OFF (null)     2   0      66.51    33.25  0.98   0.55  [726, 864]
cvec OFF (null)     2   1      66.36    33.18  1.00   0.54  [726, 864]
cvec OFF (null)     2   2      66.04    33.02  1.00   0.51  [726, 864]
cvec OFF (null)     4   0      71.79    17.95  1.00   0.87  [688, 528, 1024, 360]
cvec OFF (null)     4   1      62.41    15.60  0.97   0.88  [1024, 530, 1024, 363]
cvec OFF (null)     4   2      49.83    12.46  1.00   0.88  [1024, 523, 1024, 496]
cvec vcd_m15        1   0      55.11    55.11  1.00   0.34  [665]
cvec vcd_m15        1   1      55.12    55.12  1.00   0.35  [665]
cvec vcd_m15        1   2      55.03    55.03  1.00   0.35  [665]
cvec vcd_m15        2   0      67.03    33.51  0.97   0.53  [507, 425]
cvec vcd_m15        2   1      66.81    33.41  1.00   0.52  [495, 1024]
cvec vcd_m15        2   2      67.11    33.56  0.97   0.51  [507, 425]
cvec vcd_m15        4   0      79.85    19.96  0.97   0.84  [381, 432, 717, 397]
cvec vcd_m15        4   1      80.18    20.04  0.96   0.87  [375, 432, 859, 397]
cvec vcd_m15        4   2      79.51    19.88  1.00   0.86  [381, 432, 859, 397]
```

### What is solid, and what is not

**Solid — steering is free at C=1.** 54.43 (off) vs 55.11 (`vcd_m15`), with a spread under 0.2% across three reps in each arm. The control-vector kernel does not register against the rest of the step. Relevant to the design question below: a per-row indirection would be added to a kernel that costs nothing today.

**Solid — C=2 scales 1.22x**, identically in both arms, tight across reps.

**NOT solid — the C=4 cvec-off arm.** 71.79 / 62.41 / 49.83 is a 44% spread. Quote it as "roughly 50–72 tok/s", not as 62.41.

**NOT solid — the C=4 on-vs-off comparison.** `vcd_m15` produces shorter outputs (381–859 tokens vs 523–1024), so those arms decode at shorter KV context. The one-variable rule is not satisfied and no claim that steering is faster at C=4 can be drawn from this.

### Why this argues against doing the work for throughput

Doubling concurrency buys **+22% aggregate** while **halving per-request speed** (54 → 33 tok/s); C=4 adds nothing reliable. The engine is barely amortizing across the batch even at short context, which is consistent with the existing finding that C=4 is an outright loss at 32K.

So the cohort filter's serialization of mixed-vector traffic costs **at most ~22% throughput, and frequently nothing**. The throughput argument for a per-row control-vector kernel does not survive this measurement.

**The latency argument does.** A request naming a different vector waits for the current batch to drain, and a 1024-token generation is ~19 s at C=1. Tail latency for mixed-vector traffic is therefore tens of seconds, and that is the defensible reason to do the work — not throughput.

### Incidental finding: temp-0 output is not concurrency-invariant at C=2

While validating the harness. The same prompt, temp 0, deterministic across reps:

- alone, streamed: **1024 tokens** / 4461 chars
- alone, non-streamed: **1024 tokens** / 4461 chars (so not a streaming artifact)
- beside one other request: **726 tokens** / 2999 chars, diverging mid-sentence at char 410 — genuinely different text, not a truncated prefix

```
C=1: "...to move items to the front (most recently used) or remove from the end (least recently used).
      - This is more efficient than using an OrderedDict alone becaus"
C=2: "...to move items to the front (most recently used) or remove from the end
      (least recently used).
      - This is the classic approach for achieving O(1) get and "
```

CUTLASS grouped and shared MoE were both explicitly **off** for these runs (confirmed in the boot log), so this is not the known CUTLASS wobble. The likely mechanism is the documented one for this model: batch width selects a different per-M GEMM kernel, reduction order changes, and a near-tie token flips — bit-losslessness across batch widths is already recorded as unreachable here. It is deterministic at C=2 (identical `[726, 864]` across all three reps) but **not** across reps at C=4, which is where the cvec-off variance above comes from.

Flagged for the record, not as a blocker — this is pre-existing engine behaviour, unrelated to control vectors, and agentic runs have been fine with it.

Harness: `cvec_conc2.py`; logs under `/home/ms/logs/cvec/` (`conc_matrix2.log`, `diag_c2.log`).

rsafier:
## Correction and better-controlled data on the concurrency measurement

Two corrections to [my previous comment](https://github.com/Avarok-Cybersecurity/atlas/pull/1134#issuecomment-5731412152), and the controlled numbers that replace it.

### Correction 1: "steering is free at C=1" was wrong

That came from the 1024-token matrix (54.43 off vs 55.11 steered), where the two arms generated **different lengths** — 1024 tokens unsteered vs 665 steered. I flagged that confound in the same comment and then quoted the number anyway. It should not have been stated.

Re-measured with every request producing exactly 256 tokens, arms **interleaved wave by wave** so machine drift cancels, n=6 per arm:

| vector | operator | off tok/s | steered tok/s | delta | ranges separated |
|---|---|---|---|---|---|
| `vcd_m15` | add −0.15 | 46.76 (46.30–46.84) | 52.45 (51.87–52.60) | **+12.2%** | yes |
| `refusal` | project | 46.56 (46.49–46.78) | 49.58 (49.15–50.01) | **+6.5%** | yes |
| `vcd_p15` | add +0.15 | 46.95 (46.82–46.98) | 48.55 (46.95–48.61) | **+3.4%** | marginal |

Steering is not free at C=1 — it is consistently **faster**, across both operators and three directions. The unsteered baseline reproduces to ±0.4% across the three independent sessions (46.76 / 46.56 / 46.95), and drift within each arm is under 0.5%, so these are separated by much more than the noise.

The mechanism is almost certainly not that the steering arithmetic is cheap (it adds work). It is that steering changes **which tokens are generated**, different tokens route to different MoE experts, and expert selection changes the grouped-GEMM shapes. That is consistent with the effect size tracking how strongly each vector perturbs output (m15 > refusal > p15). Stated as a hypothesis — it has not been confirmed against a routing counter.

### Correction 2: the C=4 numbers are not trustworthy, in either arm

The previous comment hedged the C=4 unsteered arm as "roughly 50–72, not 62.41". That hedge was correct and should be strengthened: **the unsteered C=4 baseline swings about 1.8x between sessions.**

| session | off arm C=4 range (n=6) |
|---|---|
| vs `vcd_m15` | 40.87 – 60.97 |
| vs `refusal` | 40.59 – **73.41** |
| vs `vcd_p15` | 40.91 – 72.76 |

An interleaved run against `vcd_m15` looked like a clean +57.6% (off median 49.69, steered 78.33, non-overlapping ranges, and drift running the *wrong* way for a warmup explanation: off fell 18.6% over the run while steered rose 3.7%). It does not survive its own control. In the very next session the off arm reached 73.41, above the 60.97 that had been treated as its ceiling — and `refusal` at C=4 shows **+0.6% with fully overlapping ranges and identical +24% drift in both arms**, i.e. no effect whatsoever.

**No steering effect at C=4 is separable from the baseline's own instability.** No C=4 delta is claimed here, in either direction.

### What the concurrency scaling actually is

Reliable arms only (C=1 and C=2, where ranges are tight and reproducible), equal 256-token generations:

| | C=1 | C=2 | scaling |
|---|---|---|---|
| aggregate decode, unsteered | ~46.8 tok/s | ~64.0 tok/s | **1.37x** |
| per-request | 46.8 | 32.0 | 0.68x |

C=4 cannot be quoted. It is not reliably better than C=2 and swings across a 1.8x band.

### Bearing on the per-row control-vector kernel

Unchanged from the previous comment, and if anything strengthened:

- **The throughput case is weak.** Batching buys ~1.37x at C=2 and nothing dependable beyond it, so serializing mixed-vector traffic in `filter_adapter_cohort` costs at most that, and often nothing.
- **The latency case is the real one.** A request naming a different vector waits for the in-flight batch to drain; at ~47 tok/s a 1024-token generation is ~22 s. Tail latency for mixed-vector traffic is tens of seconds. That is the defensible justification for per-row selection — not throughput.
- **Steering is not a performance liability.** In every arm measured it was neutral-to-faster, so the feature does not need to earn back a cost.

### Separately worth someone's attention

The C=4 instability is a pre-existing engine property, unrelated to control vectors (it is just as large with steering off). Aggregate decode throughput at C=4 varying 40–73 tok/s run to run, on identical work, is worth its own investigation before anyone tunes concurrency on this model.

### Reproducing

Harness committed as `scripts/measure_cvec_concurrency.py` (env-overridable: `CVEC_BASE`, `CVEC_MODEL`, `CVEC_MAX_TOKENS`, `CVEC_REPS`, `CVEC_ARM`). It measures the **all-slots-busy window** — `[max(first token), min(last token)]` across the wave — rather than first-token-to-last-token, because temp-0 output on this model is not concurrency-invariant, so slots retire at different times and a whole-wave window reports the retirement pattern as if it were a property of concurrency. Logs: `/home/ms/logs/cvec/{conc_clean,interleave_c1,interleave_c2,interleave_c4,il_refusal,il_p15,il_c1_refusal,il_c1_p15}.log`.

Fingerprint unchanged from the previous comment except `max_tokens` 256 and n=6 interleaved.

tbraun96:
Triaged for a certification campaign and **parked pending a rebase** — but this is the cheapest rebase in the backlog, so it is worth doing soon.

#1132 ⊂ #1134, so **one rebase covers both** (#1134 is the tip). You fork only **26 commits** behind main — nearly current — and merging #1134 onto `5f639bfc57` locally gives 37 conflicts that are almost all one mechanical shape, not real divergence: **you did the Atlas→Avarok rename yourself, and main did it too** (`569ffd0920`). Your whole diff against `.github/workflows/ci.yml` is four lines (`launch-atlas-glm53` → `launch-avarok-glm53`), and against `file-size-cap.yml` four more ("Atlas idiom" → "Avarok idiom"); main changed the same two files by 319+/40− and 8+/633−.

**Take main's copy wholesale wherever your only diff is an atlas→avarok swap** — `git checkout avarok/main -- <file>` loses nothing there. Same for `bench/ladder38/published.json`: regenerate rather than hand-merge. #1063 hits the identical three files.

Also raise `merge.renameLimit` before you estimate — git's default gives up on rename detection across #1101 and reports a misleadingly short conflict list. Full write-up: #1206.

Campaigns are composed and queued for 22 PRs; anything rebased before one launches shares its GPU run.

Issue #1100, opened 2026-09-16
DeepSeek-V4.1 Flash: split the eight source files over the 500-line cap (#1094-#1099)


rrstesiak:
Tracking issue for the file-size-cap allow-list entries the DeepSeek-V4.1 Flash stack (#1094 → #1099) carries, so the port can ship the same way the V4 port did (#219 for #186).

Files over the cap at the head of #1099, with the split each one wants:

| file | LoC | split |
|---|---|---|
| `crates/spark-model/src/layers/attn_v41.rs` | 1169 | weights/state structs, the indexer + compressor stages, and the decode/prefill forward into siblings, the qwen3_attention template |
| `crates/spark-model/src/weight_loader/deepseek_v41.rs` | 1042 | layer assembly, runtime construction, and the real-file oracle tests into `deepseek_v41/` |
| `crates/spark-model/src/layers/deepseek_v41_ref/compress.rs` | 719 | CPU reference: compressor vs indexer vs sparse attention |
| `crates/spark-model/src/layers/moe_v41.rs` | 640 | routing, the streamed-expert loop, the shared expert |
| `crates/spark-model/src/layers/deepseek_v41_layer.rs` | 566 | hyper-connection mixes/collapse vs the block forward |
| `crates/spark-model/src/layers/deepseek_v41_ref/model.rs` | 530 | CPU reference: per-stage modules |
| `crates/spark-runtime/src/weights/gguf.rs` | 569 | grew 390 → 569 with the split-GGUF shard set; shard handling to a sibling |
| `crates/spark-runtime/src/weights/gguf/names.rs` | 550 | grew 401 → 550 with the V4.1 name map; per-arch translate tables to siblings |

Each split is mechanical and is gated on the oracles that exist today (`attn_v41` 18/18, `layer0_matches` on the real weights, `moe_v41` 2/2, `engram_v41`, `kquant`, `deepseek_v41_stream_oracle`), which is what makes it safe to do after the stack lands rather than inside it.

PR #1139, opened 2026-09-18
DeepSeek-V4.1 Flash: the decode fix and today's tips into the line

rrstesiak:
`stack/ds41-flash-line` was cut from the six V4.1 branches as they stood before this morning's push, so it carries neither the decode fix nor today's merge of main. This brings the line up to the six tips in one merge.

What it adds:

- `4a859ae5b`: the single-token MoE path had a data race in `routed_m1`, eight blocks doing a non-atomic add into one accumulator row at m=1. That was the decode nondeterminism and the loops. `moe_v41_sum_rows` replaces the launch, summing the expert rows in plan order. Verified on the box: four greedy runs bit-identical, decode versus teacher-forced prefill 97 of 100 top-1, chat suite MinHeap x3 and Volvo x3 byte-identical, 10.6 to 10.9 tok/s warm at 300 tokens.
- `d75da22bf`: the loop watchdog declines its rollback on a layer that cannot rewind, which was the request-2 corruption.
- `69dc4fba9`: `release_state` frees the per-sequence attention buffers.
- main merged into the line with the 500-line allow-list moved into the script per #941, and `scheduler/rollback.rs` allow-listed at 511.
- the four diagnostic switches, env-gated and default off.

Six files conflicted between the line's re-applied renames and the fork's; the fork side is taken on all six because it already carries the rebrand's names on the fixed code. `cargo check --workspace` for the three crates is clean, the size-cap script passes, and the result merges cleanly against current main.

Receipts and the reproduction key are on #1099 under the last push. Tips: #1094 `b847fbba1`, #1095 `a05cf6b15`, #1096 `896c8903c`, #1097 `506e8daf7`, #1098 `50c03976a`, #1099 `11aa72eab`.


TheTom (commented):
The decode fix is well diagnosed, and the fix itself is right.

**Root cause.** A per-layer checksum trace put the first divergence at the layer-0 MoE output, with attention output, MoE input and routing all identical. That narrowed it to one launch. The cause is real: in `routed_m1`, every expert row goes to token row 0, and `moe_v41_scatter_add` states a distinct-rows precondition (`moe_v41.cu:57-58`) that the m=1 caller broke.

**The fix targets that cause.** `moe_v41_sum_rows` (`moe_v41.cu:73-81`, called at `moe_v41.rs:750-758`) gives each thread one column and adds the rows in plan order. That removes the race and also matches the rounding order of the per-expert loop, so the loop path becomes a bit-exact oracle.

**Evidence.** The before/after on the box (about 50 decode-vs-prefill disagreements before, 97/100 top-1 agreement after) is the known-bad/valid pairing the RST playbook asks for.

**Other good changes.**
- **Runtime guards.** The rollback decline (`rollback.rs:227`) and the `Baked` pointer check that forces a graph recapture are both runtime guards rather than comments.

My notes are about making that evidence durable and keeping the merge clean.

**1. [Blocking] The merge resolution undid the line's 500-line split, and 12 files are no longer compiled.**
The base (`stack/ds41-flash-line`, `f83080ac9`) had split these into `foo.rs` plus `foo/` sub-modules:
- `moe_v41.rs`
- `attn_v41.rs`
- `deepseek_v41_layer.rs`
- `weight_loader/deepseek_v41.rs`
- `gguf/names.rs`

The "fork side taken" resolution brings the monoliths back (980 / 1401 / 1177 / 1056 / 551 lines) without their `mod forward; mod init; …` declarations. For `moe_v41.rs`, for example, the base is 196 lines with three `mod` lines, and the head is 980 lines with none.

That leaves twelve files in the tree that nothing includes:
- `moe_v41/{forward,init,route}.rs`
- `attn_v41/{forward,init,primitives,sources}.rs`
- `deepseek_v41_layer/{step,trait_impl}.rs`
- `weight_loader/deepseek_v41/{load_layers,real_file_tests}.rs`
- `gguf/names/deepseek41.rs`

Easy to miss across a six-file conflict. Two consequences:
- **The race would come back.** The orphaned `moe_v41/forward.rs:153` still has the racing single-token `scatter_add` launch, so restoring the `mod` lines later would quietly bring the bug back.
- **Allow-list entries.** The size cap passes because of the new allow-list entries at `check-file-size-cap.sh:193-202`. Three of them (`deepseek_v41_ref/compress.rs`, `deepseek_v41_ref/model.rs`, `gguf.rs`) name files that are already under 500 lines at this head.

Suggestion: redo the merge onto the split layout and apply the fix to `moe_v41/forward.rs`. That lets all the allow-list entries go, except `scheduler/rollback.rs`, which #1100 covers.

**2. [Should-fix] No committed test runs the fixed path yet.**
- **The m == 1 arm never runs.** `moe_v41_tests.rs` only calls `run_case(TOKENS)` with `TOKENS = 3`, and `run_case(40)` (`:139`, `:146`). So the `m == 1` arm (`moe_v41.rs:404-411`) isn't run by any test.
- **The new oracle is unreachable.** The `if tokens == 1` oracle added here (`moe_v41_tests.rs:336-365`) can't execute.
- **Too few blocks to race.** With `TOPK = 2` (`:24`), even a single-token case would have only two blocks racing over 512 columns, which is a weak trigger.

Suggestions:
- Add a `run_case(1)` test with `TOPK` at the model's value. Repeat the forward several times and require bit-equality against the per-expert loop path, which is the oracle the commit message already relies on.
- To show the instrument can fail, run it once at `4a859ae5b^` in a scratch worktree and note the result in the PR.
- The rollback guard's only test checks that two enum variants differ (`rollback_tests.rs:225-228`). A stub model that returns `decode_rollback_unsupported() == true` and expects `LayerStateNotRewindable` would lock the guard in.

**3. [Should-fix] Scope: the fix could stand alone, and the description misses part of the increment.**
- **The fix is small.** It's 26 lines in two files (`4a859ae5b`), and the same race exists on the base line (`moe_v41/forward.rs:153`). It could go to the line as its own small PR, which would be easy to review and easy to revert.
- **The graph work isn't described.** The increment also carries the CUDA-graph decode work: `04ee6fb62`, `43913a8bf`, `6c1f07b9e`, `GraphMode`, `step_graph`, `run_segment` and `attn_v41_ring_put`. It also carries the Hopper/B200 targets and parser changes. None of that is in the "What it adds" list.
- **Risk.** It's default-off behind `ATLAS_DS41_GRAPH=1`, which lowers the risk. Either splitting the fix out or listing the graph path and targets in the description would let the review cover it.

**4. [Should-fix] The graph-equivalence tests capture a fresh graph each step, so replay isn't tested.**
The attention graph test (`attn_v41_tests.rs:222-245`) captures, launches once, and destroys the graph at a single position. The risk specific to graphs is different: replaying a graph captured at step t after `decode_prep` / `stage_m1` has refreshed the device-side position, selection and expert table for step t+1. A stale baked argument, or a skipped re-upload in `decode_prep`'s change detection, would only show up then. The golden already has two decode positions (12 and 13, window 8). Capturing at `decode12`, replaying at `decode13` after `decode_prep`, and comparing against eager would move the box-only `ATLAS_DS41_GRAPH_ORACLE` check into CI.

**5. [Should-fix] Bring the throughput claim in line with `docs/perf/PERFORMANCE_NUMBERS_STANDARD.md`.**
- **Current reference.** The body quotes "10.6 to 10.9 tok/s warm at 300 tokens" and points to #1099. That comment names `ds41_smoke.sh`, which isn't in the repo, and receipt paths on the box.
- **What the standard needs.** It asks for a launch line with every parameter and switch, the commit sha, a driver line and the receipt JSON, reported as a median over reps with TTFT p50.
- **Suggested additions.** The full launch line inline, including whether `ATLAS_DS41_GRAPH` was on, the three `AVAROK_TARGET_*` values and the cache settings. Plus the receipt JSON, and median and TTFT p50 in place of ranges. That makes the number reproducible on a teammate's Spark, as PRD §1 asks.
- **Label the 97/100.** It's engine self-consistency (decode vs this engine's own teacher-forced prefill), not agreement with the reference or a quality measure. The PRD keeps those claims separate.

**CI.** `cargo test --workspace` fails `gate::taxon::taxon_tests::every_real_target_resolves_a_nonempty_source_set`, because the b200/hopper `deepseek-v4.1-flash` targets aren't in the expected list. That comes from the line, not this increment.

**Nit.** Two doc comments sit on the wrong items: the `ATLAS_DS41_PREFILL_GEMV` doc is on `m1_loop()` (`moe_v41.rs:30-35`), and the `h2d_enqueue` doc now sits on `h2d_sync_all()` (`gpu_impl.rs:85-88`).

rrstesiak:
Right on all three, and thank you for reading it that closely. Redone.

**The merge.** The branch is now `stack/ds41-flash-line` at `0719dea47` plus the fixes ported into the split layout, and nothing else: no diagnostics, no second main merge, no allow-list entries beyond `scheduler/rollback.rs` at 511 (the watchdog change; split tracked in #1100). All five split roots keep their `mod` lines; the twelve files are compiled again; `moe_v41/forward.rs` has one `scatter_add` launch left, the group loop.

- `c00342fb6`: the race fix, in the right place this time. `moe_v41_sum_rows` in the kernel file; `sum_rows` in the handle struct in `moe_v41.rs` and loaded in `moe_v41/init.rs`; the m=1 `scatter_add` launch in `moe_v41/forward.rs` replaced by the `sum_rows` launch through the base's `launch_n`.
- `c42e0a325`: the rollback decline, cherry-picked; the `decode_rollback_unsupported` impl lives in `deepseek_v41_layer/trait_impl.rs`.
- `9b3877aa8`: `release_state` frees the attention buffers; the base had no `release_state` for this layer at all, so it is implemented in `trait_impl.rs`.

**The test you asked for.** `single_token_expert_sum_is_bit_identical_across_runs` in `moe_v41_tests.rs`: one token, topk at the model's six, six forwards, byte-equal outputs, all finite, six distinct experts routed. The per-expert loop path is not reachable in this harness, so it is repeat-equality rather than loop-equality; the loop-hash match is on the box run in the description.

**It caught a second bug.** `gate_out`, `up_out`, `h`, `h_q8` and `down_out` in `moe_v41/init.rs` were sized by `max_tokens` rows, but `routed_m1` writes one row per selected expert. At `max_tokens: 1` with six experts the forward wrote past the buffers, an illegal address. Serving uses `max_tokens` 2048, so it never surfaced. `c9e11be44` sizes them by `max(max_tokens, topk)`.

**Run.** Size-cap passes with the one entry. `cargo check` clean on spark-model, spark-server, spark-runtime. `cargo test -p spark-model --release moe_v41 -- --ignored --test-threads=1`:

```
single-token forward at topk 6: 6 runs bit-identical
streamed MoE: 1536 values within 123.5010 of the emulated numerics (worst 64.00000)
streamed MoE: 20480 values within 205.0010 of the emulated numerics (worst 192.00000)
test result: ok. 3 passed; 0 failed; 0 ignored
```

Two harness notes for whoever runs it next: the tests look up the kernel source under `deepseek-v4-flash`, so `AVAROK_TARGET_MODEL=deepseek-v4-flash` is the value that resolves; and `--test-threads=1`, because an illegal address in one test poisons the shared CUDA context for the others.

On the scope point: agreed the fix could stand alone. This branch is now only the fixes, so it is that PR.

rrstesiak:
Superseded by #1144: #1131 merged the line to main at `68c393913` without this fix, so the same six commits are now cherry-picked against main there. Leaving this open until #1144 lands, in case the base branch is still wanted.

tbraun96:
Closing as superseded by **#1144** (`8f9882ee27`), which carries the same six commits.

Verified rather than assumed:
- `git diff` between this head (`c9e11be440`) and #1144's, restricted to all **14** files this PR touches, is **0 lines** — they are byte-identical. `range-diff` pairs the six commits 1:1; the only difference is the `(cherry picked from …)` trailers.
- #1144's merge-base *is* current `main` (`68c3939135`) and it merges with **0 conflicts**.
- This PR is based on `stack/ds41-flash-line`, whose PR (#1131) merged as a **squash** — so that branch is not an ancestor of `main`. Taken against `main` this PR would **delete 44 files under `.benchmarks/`**: the 22 certification records and their `.sig` sidecars that #1130 and #1131 landed. That alone makes it unmergeable as based.

The content is not lost — it is #1144's, which is the mergeable route. Closing this so the branch `stack/ds41-flash-line` can be cleaned up later without `--delete-branch` silently closing a PR that still mattered.

PR #1140, opened 2026-09-18
DeepSeek-V4.1 Flash: the B200 campaign

rrstesiak:
Draft claim. The single-Spark line is complete (#1094 to #1099, bridged into the campaign branch by #1139): the graph, the loader, expert streaming, the K-quant GEMV and MMQ arms, engram on, the decode fix. This PR carries the same engine onto datacenter Blackwell, and it reports numbers only as measured.

**What lands here**
- The x86 binary from `datacenter-binaries.yml` at `hw=b200 model=deepseek-v4.1-flash`, built from the frozen head, sha recorded. The `kernels/b200/deepseek-v4-flash` target with the V4.1 kernels mirrored is already on the branch and passes the sm_100a PTX gate.
- The recipe: `deepseek-v4.1-flash-q2k-b200.yaml`, one node, no speculation, engram on, the streaming cache sized to the box.
- The first hour on the box, in order: smoke on the known-answer prompt, then the suite, MinHeap x3 and Volvo x3, warm, then cold after a cache drop. Every number with its launch line, commit sha, driver line and receipt JSON, per the performance-numbers standard on main.
- Competing numbers first, in the same table: the two-Spark EXL3 run with the engram dummied, the llama.cpp CPU recipe on one GB10, and the single-Spark native figure from #1099. This engine is the only one of the three with a faithful engram; the table says so and nothing more.

**The two shapes this PR covers**
- One B300: the 247 GB Q2_K checkpoint is resident. No streaming. This is the number to beat.
- One B200: about three quarters of the experts resident, the rest streamed from NVMe through the same LRU that runs on the Spark, `ATLAS_DS41_EXPERT_CACHE_GIB` sized to the box. The interesting measurement is how much of the Spark's disk-bound ceiling the streaming still costs when the disk is a datacenter NVMe.

**Not in this PR**
- Two or more B200 with the experts split across GPUs. That needs an expert-parallel layout for V4.1's routed experts, which does not exist yet, and it gets its own PR when it does.
- Speculation. The GGUF carries no DSpark keys; MTP for V4.1 is a separate line.

**Oracle:** the same known-answer prompt and the same suite as #1099, byte-identical output across repeats before any throughput is quoted.

Tracking: #1059 under #970.

Recipe on the branch: `deepseek-v4.1-flash-q2k-b200.yaml`.


TheTom (commented):
Early notes, since this is still a draft. I like "No figure is typed here until it has been run on the box": holding numbers back until there's a receipt is the right default.

Two things in `deepseek-v4.1-flash-q2k-b200.yaml`:

1. **Three sequence limits disagree:** `max_model_len: 32768`, `ATLAS_DS41_MAX_SEQ: "4096"` and `ATLAS_DS41_MAX_TOKENS: "512"`. Which one does the engine enforce? A client reading the advertised 32k would send more than the 4k the model is configured for. Aligning `max_model_len` with `MAX_SEQ` (and with whatever runtime cap lands from the #1099 discussion) would keep the recipe honest about what it serves.

2. **`container: avarok/atlas-gb10:latest` is an aarch64 image on an x86 B200.** The comment explains it's the schema's required placeholder and that the campaign runs the native binary. Still, anyone who uses the recipe as written would pull the wrong architecture. Could the field point at an x86 image, or could the schema allow it to be omitted for native-binary recipes?

tbraun96:
**Stack ordering: this line should chain on #1144, not on `main` — and it's your branch to re-cut, so I'm laying it out rather than touching it.**

Ran the stack-order check on #1140/#1141/#1142/#1144 against `main` at `68c393913`. Ruling: **order wrong as filed**; corrected chain is `main → #1144 → #1140 → #1141 → #1142`. Everything below is from a fresh fetch of the four PR heads (`8f9882ee2`, `d8961c9a0`, `2775341ee`, `9f10c22b8`); nothing on your fork was pushed to or rewritten.

**Why #1144 belongs at the bottom**
- `moe_v41_sum_rows` is absent from `main` (`git grep` on `68c393913`: 0 hits). #1144 adds it against the split layout main has today (`moe_v41/forward.rs`, `moe_v41/init.rs`, `moe_v41/route.rs`).
- #1140–#1142 each carry an older copy of the same six-commit fix, applied to the pre-split `moe_v41.rs` (that head has no `moe_v41/` directory). Four of those commits pair with #1144 commits by title but have **different patch-ids**: `4a859ae5b`↔`87731f30e`, `d75da22bf`↔`dd6eb0be8`, `69dc4fba9`↔`ff67b7303`, `11aa72eab`↔`0f23294ac` (`git patch-id --stable` on each).
- The campaign branch is missing two of #1144's six: the `max(max_tokens, topk)` per-expert buffer fix (`8f9882ee2`) and the bit-identical single-token test (`2d0c30afc`). So #1140's copy is stale and incomplete, not a superset.
- #1144 merges into `main` clean (merge-base is current main). #1140 has 41 conflicting files against `main` (`git merge-tree --write-tree --name-only`); its merge-base `951109f7d` is 3 commits behind.

**The cost as filed**
`stack-plan.py`'s `PERF_PATHS` rule (top dir in `crates`, `kernels`, …) counts #1144 as 13/14 perf files and #1140 as 81/84. #1141 and #1142 each add exactly one file on top of the previous head — `crates/spark-server/tests/fixtures/recipes/deepseek-v4.1/deepseek-v4.1-flash-q2k-b200-{ep2,spec}.yaml` — and because that path starts with `crates/`, each one owes its own campaign. Four separate PRs on `main` = four ~3-GPU-hour certification campaigns (~12 GPU-h). One chain = one campaign at the base.

**Suggested steps (your call; I did not do any of this)**
1. Rebase `ds41-b200-campaign` onto `8f9882ee2` (#1144's head), dropping `4a859ae5b`, `d75da22bf`, `69dc4fba9`, `11aa72eab` — they are superseded by #1144's versions. Expect the `moe_v41.rs` → `moe_v41/` split to be the bulk of the conflict work.
2. Either fold the two recipe YAMLs into #1140 and close #1141/#1142 (they have no code, and two fixture files don't justify two layers), or keep them as layers rebased on the new #1140 head.
3. Change #1140's base to `ds41-decode-fix-main`. Note GitHub only accepts a base branch that lives in `Avarok-Cybersecurity/atlas`, so the chain needs an upstream branch for #1144's head — say the word and one of us with write access will push a mirror of `8f9882ee2` (unchanged content) for you to target.

**What I did not verify:** whether the rebased campaign builds or what its numbers are — that's the campaign's job. #1144 itself is unaffected by any of this and can proceed as is.

rrstesiak:
Agreed, and done as you laid it out.

- The mirror: `ds41-decode-fix-main` on this repo at `54b8e4399`, #1144's head after your fmt commit and the main merge, unchanged content. I pushed it myself.
- The campaign branch is rebuilt on that head with only its recipe commit; the four superseded commits (`4a859ae5b`, `d75da22bf`, `69dc4fba9`, `11aa72eab`) are dropped in favour of #1144's versions.
- #1141 and #1142 are kept as layers, each one recipe file on top of the previous head, so the diffs read as `+60 / 1 file`, `+123 / 2 files`, `+188 / 3 files` against the mirror.
- All three are retargeted at the mirror and mergeable: `main → #1144 → #1140 → #1141 → #1142`, one campaign at the base.

When #1144 lands, the mirror can go and the three re-base onto main with nothing else changing.

rrstesiak:
Rebased onto main `4c00f65a2` after #1144 merged. No content change: recipe commit re-stacked on main, base switched from the mirror branch to main. Head `f35892d52`.

rrstesiak:
Updated one-Spark line for the competing-numbers table, measured today on the 300-token standard (MinHeap x3 + Volvo x3, temp 0, medians of three, every text byte-identical to the baseline oracle): 19.19 tok/s MinHeap / 21.87 Volvo (best single runs 19.70 / 21.94; hot probe 24.1), from the chained drafts #1147 (kernels and launches), #1148 (the expert cache), #1153 (the hit step), #1155 (the expert arena in device memory), #1156 (the sm_100a gate: 199/199 kernels compile with zero source changes; device-side routing). The measured one-Spark byte floor is 6.09 GB of weights per token at 223-249 GB/s, so the B200 step is launch-bound, not bandwidth-bound; the first-hour runbook with that arithmetic is #1158.

tbraun96:

**Citation.** This PR's head `f35892d525` is an ancestor of the composition tip `2dd6e33ebc` (`integration/ds41-thirteen`, PR #1207), which merged to main as squash commit `b79bc1151b9a` at 2026-09-21T22:48:02Z. Containment was proved before closing: all **133** files the composition changed are byte-identical on main.

**What is *not* claimed.** This line is a stack — later layers edit the same files — so a file this PR touched may now carry a *later* layer's content rather than this one's. The claim is that these commits are inside what landed, not that every one of the 1 files still reads as this PR left it. That is normal for an intermediate layer and is why the check used here is ancestry-of-the-composition plus the composition's own containment proof, rather than per-file identity.

PR #1141, opened 2026-09-18
DeepSeek-V4.1 Flash: two-node B200, expert parallel and the kernel retune

rrstesiak:
Draft claim, stacked on #1140 (base main, cumulative head; GitHub will not base a PR on a fork branch). #1140 covers one GPU, resident on a B300 or partly streamed on a B200. This PR covers the shape that removes streaming entirely on B200 and makes the datacenter number a compute number: two B200 on one NVLink node, the routed experts split across them, and the kernels retuned for the part.

**What lands here**
- An expert-parallel layout for V4.1's routed experts, 384 per layer with 6 active, placed across two GPUs so the full 247 GB Q2_K checkpoint is resident with room for the KV and the engram rows. The router runs replicated; the six selected experts are gathered where they live and the rows come back over NVLink. The engram tables stay on the rank that owns the layer.
- The kernel retune for sm_100a, hand-rolled: the K-quant GEMV and MMQ arms, written for 48 SMs on the Spark, retuned for 148: block shape, rows per block, split-K on the tall decode GEMV, occupancy against the larger register file. The attention decode warp GEMV and the engram row gather get the same pass. No CUTLASS, no CuTe.
- The two-node recipe, `deepseek-v4.1-flash-q2k-b200-ep2.yaml`, one node, two GPUs, engram on, no speculation. On the branch now.
- Numbers only as measured, competing numbers first in the same table, with the one-GPU figure from #1140 beside them so the split's cost and the retune's gain are each visible on their own.

**The port, in three layers**

The correctness port is already done. The same `.cu` files are mirrored into `kernels/b200` and pass the sm_100a PTX gate, so they compile for the part and they will run. What this PR carries is the performance port, which has three layers, and only the first is a matter of parameters.

1. *Parameters.* Grid and block sizing for 148 SMs instead of 48, rows per block, split-K on the tall decode GEMV, occupancy against the larger register file. This is the same retune the Hopper port needed from 48 to 132 SMs, done again for 148. It is sweeps, and it gets most of the gain.

2. *The memory system, which changes what is slow.* The Spark's LPDDR5X moves about 270 GB/s; B200's HBM3e moves about 8 TB/s. Kernels that were bandwidth-bound on the Spark stop being the bottleneck, and the costs the Spark never showed come forward: launch overhead, host syncs, the CPU-side router staging. On the Spark, host and device share one memory, so the MoE host stage and the top-k downloads were close to free. On B200 they cross a bus every token. The async device copies on #1099 are the start of that work; the first profile on the box decides the rest.

3. *What sm_100a has that sm_121 does not.* Tensor memory and the two-CTA MMA path. The kernels do not use them because the Spark does not have them, and they run without them. Using them is a rewrite of the GEMM arms, not a retune, and it is the difference between a good number and the best one. That decision follows the first measured number, not the other way round.

So the sequence on the box is: the kernels run on day one, the retune lands the first pass, and the architectural pass is scoped from the profile.

**The interconnect, and what stays off it**

Two B200 in one HGX or DGX B200 chassis talk over fifth-generation NVLink through the NVSwitch, on the order of 1.8 TB/s per GPU in each direction. That is where the expert exchange lives: the router picks six of 384 experts per token per layer, the six live on one GPU or the other, and the activations go out and the rows come back over NVLink. Per token that is a few kilobytes of exchange against a link that moves terabytes, so the split's cost is latency and synchronisation, not bandwidth, and the layout keeps the two ranks in lockstep with one exchange per MoE layer rather than a chain of small ones.

PCIe Gen5 is the host path, about 64 GB/s each way, and nothing on the token path should touch it. The KV, the engram rows, the router state and the expert rows all stay in HBM. The host is for loading and for the receipts. This is the same lesson as layer two of the port: on the Spark the host was free because it shared the memory; here it is a bus, and every per-token host sync is a stall the profile will show.

The number to hold against this is the team's two-Spark run of V4 0731 at 15.5 tok/s, which was bound by RoCE between two boxes with CUDA graphs off because the replay deadlocked over the network. On one node there is no network: NCCL over NVLink, the same `spark-comm` backend, and the expert-parallel graphs can be on. The comparison in the table is one-GPU B200 resident (#1140) against two-GPU B200 split (this PR), same prompts, same suite, so the reader sees what the split costs and what the retune returns, each on its own.

**Oracle:** byte-identical output across both ranks and against the one-GPU run on the same prompts, the suite MinHeap x3 and Volvo x3, before any throughput is quoted.

**Not in this PR**
- Speculation. MTP for V4.1 is its own line once the modules are wired.
- Hopper. No NVFP4 datapath there; the FP8 and dequant arms are a separate target.
- Anything beyond two GPUs. Four-way placement follows the same layout and lands when a box exists.

Tracking: #1059 under #970.


TheTom (commented):
Early note on the draft. The header of `deepseek-v4.1-flash-q2k-b200-ep2.yaml` says the routed experts split across the two GPUs "so the full Q2_K checkpoint is resident and nothing streams". The `metadata.description` block looks like it came over from the single-GPU recipe, though. It still says "on one datacenter Blackwell GPU" and "about three quarters of the routed experts are resident and the rest stream from NVMe". The `ATLAS_DS41_EXPERT_CACHE_GIB` comment also still describes streaming.

Worth updating the description and the cache comment to the two-GPU, all-resident layout, so the recipe describes the shape it actually runs.

rrstesiak:
Rebased onto main `4c00f65a2` after #1144 merged. No content change: recipe commits re-stacked on main, base switched from the mirror branch to main. Head `3066b751c`.

tbraun96:

**Citation.** This PR's head `3066b751cb` is an ancestor of the composition tip `2dd6e33ebc` (`integration/ds41-thirteen`, PR #1207), which merged to main as squash commit `b79bc1151b9a` at 2026-09-21T22:48:02Z. Containment was proved before closing: all **133** files the composition changed are byte-identical on main.

**What is *not* claimed.** This line is a stack — later layers edit the same files — so a file this PR touched may now carry a *later* layer's content rather than this one's. The claim is that these commits are inside what landed, not that every one of the 2 files still reads as this PR left it. That is normal for an intermediate layer and is why the check used here is ancestry-of-the-composition plus the composition's own containment proof, rather than per-file identity.

PR #1142, opened 2026-09-18
DeepSeek-V4.1 Flash: speculative decoding on B200, DSpark from the shipped modules

rrstesiak:
Draft claim, stacked on #1140 and #1141 (base main, cumulative head; GitHub will not base a PR on a fork branch). Those two put the engine on datacenter Blackwell, resident and retuned. This PR multiplies the decode: speculative decoding on the same box, with drafters that already exist.

**What lands here**
- **DSpark, from the modules DeepSeek trained.** The V4.1 release ships DSpark heads at target layers 37, 38 and 39. The Q2_K GGUF dropped them in conversion, so this PR reads them from the official safetensors, only those tensors, sized from the index before anything is downloaded, and carries them beside the Q2_K target as a sidecar. Atlas already runs DSpark for V4 0731 (#648); this is that path on the V4.1 graph, with the K-quant arms as the verify.
- **N-gram lookup drafts as the floor.** No drafter model, drafts from the prompt and the output (#1026, #1060), on from day one, and the fallback whenever DSpark has nothing to propose. The two compose: lookup first, DSpark where lookup misses.
- **The recipe**, `deepseek-v4.1-flash-q2k-b200-spec.yaml`, on the branch now: one or two GPUs per #1140 and #1141, engram on, DSpark drafts three, lookup on.
- **Numbers only as measured, competing numbers first**: the two-Spark EXL3 DSpark run and its accept rate in the same table, the non-speculative B200 figure from #1140 beside it, so the multiplier is visible on its own.

**First step, before any code:** the safetensors index. The three DSpark modules may be full layers with their own experts, and the index says exactly which files hold them and how many bytes. That sizing is posted here as the first comment on this PR, and the sidecar format follows from it.

**Oracle:** losslessness first. Byte-identical output against the non-speculative run on the same prompts, the suite MinHeap x3 and Volvo x3, before any throughput or accept rate is quoted.

**Not in this PR**
- DFlash-2. No drafter exists for this target; training one is its own project.
- The GDN batched verify line. It is specific to linear attention and does not apply here.
- Tree or multi-draft verify beyond what #648 already does. Later, from the accept profile.

Tracking: #1059 under #970.


TheTom (commented):
Early note on the draft. The header of `deepseek-v4.1-flash-q2k-b200-spec.yaml` describes speculative decoding with DSpark from the shipped modules at layers 37–39 and n-gram lookup drafts as the floor. The `metadata.description` block, though, says "No speculation: the GGUF carries no DSpark keys", and `defaults.speculative` is `false`. That text looks like it was carried over from the base recipe.

Once the speculative path is wired, the description and `speculative:` should match the header. It's also worth keeping the losslessness-vs-non-speculative oracle from the header as the first check in the recipe's own text.

rrstesiak:
Rebased onto main `4c00f65a2` after #1144 merged. No content change: recipe commits re-stacked on main, base switched from the mirror branch to main. Head `b3b3dee6e`.

tbraun96:
Composed into **#1207** with twelve others of the DeepSeek-V4.1 line — one certification campaign for the whole line rather than thirteen. Your head `b3b3dee6e6` merged onto current main with **no conflict**, and it is the tip of the B200 chain (#1140 ⊂ #1141 ⊂ #1142), so merging it alone carries all three.

One note: `cargo test --workspace` and `cargo llvm-cov --workspace` are red on this PR's own CI, not just on the composition. The failure is `gate::taxon::taxon_tests::every_real_target_resolves_a_nonempty_source_set` — `walk()` finds `b200/deepseek-v4.1-flash/nvfp4` and `hopper/deepseek-v4.1-flash/nvfp4` but that test's expected list does not name them. Fixed on the integration branch in `6b78de4b97` by extending the list in walk order; details on #1185. Carry it if you push again, or leave it and #1207 brings it in.

tbraun96:

Verification: this PR's head `b3b3dee6e67344b1a5c8f9e4492b310a4d883ca4` is the second parent of composition merge `5c45b8e5d36e30e818d40383b4d39d6c2ff52516`, and every one of the 133 files the composition contributed is byte-identical on `main` today. Concrete artifact: `crates/spark-server/tests/fixtures/recipes/deepseek-v4.1/deepseek-v4.1-flash-q2k-b200.yaml (blob 8a15fa4490, identical on main)`.

PR #1144, opened 2026-09-19
DeepSeek-V4.1 Flash: the decode race fix

rrstesiak:
The decode race fix for DeepSeek-V4.1 Flash, against main, now that #1131 has landed the line at `68c393913` without it. Six small commits, cherry-picked from #1139 with their messages; the same content Turney reviewed there ("the decode fix is well diagnosed, and the fix itself is right").

**The bug.** `routed_m1` in `moe_v41/forward.rs` finished with one `moe_v41_scatter_add` launch, one block per expert row, and at m=1 every row is token 0, so eight blocks did a non-atomic f32 add into the same accumulator row. The kernel's own comment states the distinct-rows precondition that call violated. Lost adds flattened the decode distribution, which is what the greedy loops were. Main has this at `forward.rs:153` today.

**The fix.** `moe_v41_sum_rows` adds the expert rows into the token row in plan order, the rounding sequence of the per-expert loop, and replaces that launch. On the box, binary at the fix: four greedy runs bit-identical, decode versus teacher-forced prefill 97 of 100 top-1, chat suite MinHeap x3 and Volvo x3 byte-identical across repeats, 10.6 to 10.9 tok/s warm at 300 tokens. Receipts and the launch line on #1099.

**With it, from the same line of work:**
- the loop watchdog declines its rollback on a layer that cannot rewind, which was the request-2 corruption;
- `release_state` frees the per-sequence attention buffers, which the merged line lacked for this layer;
- the single-token test Turney asked for: one token, topk at the model's six, six forwards, byte-equal, all finite, six distinct experts;
- the bug that test found: the per-expert buffers were sized by `max_tokens` rows but `routed_m1` writes one row per selected expert; now `max(max_tokens, topk)`;
- `scheduler/rollback.rs` allow-listed at 511 lines, the only allow-list entry; split tracked in #1100.

**Checks.** Size-cap passes. `cargo check` clean on spark-model, spark-server, spark-runtime. `cargo test -p spark-model --release moe_v41 -- --ignored --test-threads=1` at the same content on #1139: 3 passed, the single-token forward six runs bit-identical.

Until this lands, a V4.1 decode on main is nondeterministic and loops.

Tracking: #1059 under #970. Supersedes #1139, which targeted the pre-merge branch.


tbraun96:
FYI: the stack-order check puts this PR at the **bottom** of the DS-V4.1 B200 chain (`main → #1144 → #1140 → #1141 → #1142`), since #1140–#1142 carry an older, incomplete copy of this fix (different patch-ids, pre-split `moe_v41.rs`, missing `8f9882ee2` and `2d0c30afc`). Details and the suggested re-cut are on #1140; nothing changes for this PR — it merges clean against `68c393913` and can certify as is.

tbraun96:
Pushed one commit to this branch (`3f41a7e`) — `maintainer_can_modify` is on, and this was blocking two required contexts for one hunk.

**What was red.** `cargo fmt --check` and `cargo clippy --tests` both failed on `8f9882ee2`, and they failed for the *same* reason: the `assert_eq!` at `moe_v41_tests.rs:408`, added by the single-token test, exceeds the line budget. The clippy job runs `cargo fmt --all -- --check` as its first cargo step (`ci.yml` documents the ordering — 6 s of formatting ahead of 88 s of linting), so the formatter's exit code ended that job at `03:27:06`, three seconds after it started. **`cargo clippy` never ran on this head.** It is unproven, not passing; this push is what lets it run for the first time.

The commit is rustfmt's own output, whitespace only — the assertion message is byte-identical, so nothing the test proves changes.

**Sequencing.** #1145 is mid-campaign on all three GB10s right now (anchor `a87ede905f`, ~2.8 h wall). This PR touches `crates/**` and `kernels/**`, so it needs its own 12-gate campaign, and starting one now would either contend for the boxes or be voided by #1145 landing on a perf path. It goes next, alone, once #1145 is in.

**On #1146.** Thank you for the re-cut — the containment notes (`3f50f10c3`, `4c99ba96a`, the clippy/rebrand set) are exactly the evidence that was missing. It carries the same fmt hunk plus `cargo test --workspace` and `cargo llvm-cov` failures, which are not formatting, so I have left it alone while it is a draft and yours to work. Say the word if you would like me to take the fmt hunk there too.

PR #1146, opened 2026-09-19
DeepSeek-V4.1 Flash: the rest of the line re-cut onto main (parsers, CUDA-graph decode, Hopper and B200 targets, diagnostics)

rrstesiak:
Draft. Stacked on #1144 (base main, cumulative head; GitHub will not base a PR on a fork branch). Answers the sixteen-commit list on #1099: #1131 took the line at `f86163c17` and squashed it, so everything pushed after that point was absent from main. #1144 carries the fixes. This PR re-cuts the rest onto main's split layout, one commit per original, with the original messages and `cherry picked from` trailers, so `git cherry` by patch-id finds each one.

**Parsers**
- `1d1d7b024`: chat, tool and reasoning parsers recognise the `deepseek_v41` model type. Applied clean. `crates/spark-server` had zero `deepseek_v41` hits on main; it has four now.

**The graph path** (each re-cut by hand into the split files; the `mod` lines stay, no monolith comes back, no file crosses 500 lines)
- `04ee6fb62`: the attention decode step split into a host half and a capturable device half. The new methods live in `attn_v41/decode.rs`, a new sibling of `forward.rs`; the window accessor sits on `AttnV41LayerState` in the root.
- `43913a8bf`: the MoE forward split into router launch, host stage and single-token compute. The plan, the pointer table, `routed_m1` and the stage/compute halves live in `moe_v41/single.rs`. `routed_m1` carries the `sum_rows` tail from #1144, not the original `scatter_add`, so the race stays fixed; the only `scatter_add` launch left is the group loop.
- `6c1f07b9e`: the CUDA graph of the single-token step, per layer, `ATLAS_DS41_GRAPH=1`. `GraphMode`, `Baked` and `LayerGraphs` sit in the layer root; the segment runner, capture, replay and the eager oracle live in `deepseek_v41_layer/graph.rs`. `release_state` frees the captured segments and then the attention buffers in one body.

**Targets**
- `2fccec601`: the Hopper (sm_90a) and B200 (sm_100a) targets: `kernels/{hopper,b200}/deepseek-v4.1-flash/MODEL.toml`, the inherited-target tests (landed on `crates/avarok-kernels`, the renamed crate), the datacenter-binaries workflow entry, `docs/HARDWARE.md`.

**Diagnostics, all env-gated, default off**
- `9523d8195`: `ATLAS_DS41_PREFILL_GEMV=1`, prefill groups through the decode GEMV arm.
- `7356880d5`: `ATLAS_DS41_TRACE=1`, a checksum per layer half of the eager step.
- `120bb3dc8`: the trace's MoE input, routing and cache counters per layer.
- `36559dad1`: `ATLAS_DS41_M1_LOOP=1`, the single token through the per-expert loop.

**Skipped, and why**
- `3f50f10c3` and `88b85ea2c` (the kernel mirrors into Hopper and B200): every symlink they add already exists on main with the same target, and the include-path comment in `kquant_moe.cu` is already there. Contained.
- `4c99ba96a`, `28db9f3d2`, `3a9abe270` (three one-line fixes): already on main in their split homes (`gguf/shards.rs`, `deepseek_v41_ref/model/forward.rs`, the stream bench). Contained.
- The clippy and rebrand-name commits (`40e4d02a6`, `285c17c64`, `a7e018781`, `109d128d5`, `d7fa8004f`, `c6bc97db1`, `b433925c9`, `a97226d73`): their content is what main already carries after #1101 and the split. Contained.
- `46bcfc5cb` (`ATLAS_H2D_SYNC=1`, a drain after every transient H2D in the CUDA backend): dropped. It was a bug-hunt switch that found nothing; it touches the shared backend, and the race it was chasing is fixed in #1144.

**Checks**
- `cargo check` clean on spark-model, spark-server, spark-runtime.
- `cargo clippy -p spark-model -p spark-server --no-deps -- -D warnings` clean.
- Size-cap script passes; the only allow-list entry is `scheduler/rollback.rs` at 511, from #1144.
- Both new MODEL.toml files present; the five split roots keep their `mod` lines; one `scatter_add` launch in the tree, the group loop.

Not run here: the GPU tests and a serve. The content is the fork line's, which ran on the box through the 09-17 receipts on #1099; the re-cut changes where the code lives, not what it does, and the box run of this branch is the next step.

Tracking: #1059 under #970. Answers the sixteen-commit list on #1099.


tbraun96:
All four Rust contexts (`cargo fmt --check`, `cargo clippy --tests`, `cargo test --workspace`, `cargo llvm-cov --workspace`) are red for **one cause**: a re-cut splice left an `assert_eq!` unclosed, and a syntax error fails every Rust job.

```
error: expected `,`, found `assert`
   --> crates/spark-model/src/layers/attn_v41_tests.rs:274:5
273 |                 "{what}: sa_topk_idxs differ from the reference"
    |                                                                 - expected `,`
274 |     assert!(
    |     ^^^^^^ unexpected token
```

**What happened.** The loop epilogue was spliced into the middle of a per-layer assertion. Lines 271–279 currently read:

```rust
assert_eq!(
    run.idx, lt.attn.idx,
    "{what}: sa_topk_idxs differ from the reference"     // ← never closed
assert!(
    captured_layers > 0,
    "no capturable decode layer was exercised (all layers are kv/index sources?)"
);
println!("  captured decode step == eager forward on {captured_layers} layer-steps");
);                                                       // ← the assert_eq!'s `);`, stranded here
```

**Where the epilogue belongs.** `captured_layers` is declared at :212 and incremented at :241 inside the per-layer loop, so `assert!(captured_layers > 0, …)` and the `println!` are that loop's *epilogue* — they have to run after every layer has been visited. Closing the `assert_eq!` in place would leave them executing mid-body, before `let q = down_bf16(…)` at :280 and the rest of the per-layer comparisons. So the fix is two steps: close the `assert_eq!` after :273, and move those two statements down past the end of the loop.

I have deliberately not pushed this — the placement is yours, and this is your draft. Say the word and I will, or fix it however you prefer.

One caveat worth stating: **the compiler stops at the first syntax error**, so this is the first defect, not necessarily the only one. My own crude scan flagged two more spots in `crates/spark-server/src/scheduler/rollback_tests.rs` (around :217 and :306) that *may* be the same shape, but the heuristic is unreliable and I would not act on it — the next CI run after this fix is the instrument that will say.

For context on sequencing: #1145 is at position 1 in the merge queue and #1144 goes next, alone, with its own campaign. This one is behind both.

rrstesiak:
Thank you for the placement note; that is exactly where it went. The `assert_eq!` is closed after the message, and the two epilogue statements now sit after the outer `for (regime, start_pos, trace)` loop, so they run once every layer has been visited. The other two spots your scan flagged in `rollback_tests.rs` are ordinary tests; nothing to change there.

Also applied rustfmt to the re-cut files, which covers the same long assertion you reformatted on #1144.

Checked here before pushing: `cargo fmt --all -- --check` clean; `cargo clippy --tests` on spark-model, spark-server and spark-runtime with `-D warnings` clean; `cargo test --workspace --no-run` builds every test target; `cargo test -p spark-model --lib` 1,315 passed. The GPU tests are not run on this branch, as the description says.

On the ordering for the B200 chain, read and agreed: #1144 at the bottom, then this, then the campaign. I will take you up on the mirror branch for #1144's head so the chain can target it.

rrstesiak:
Rebased onto main `4c00f65a2` after #1144 merged. No content change: the ten re-cut commits now sit directly on main. Head `0df4a66f3`.

tbraun96:

**Citation.** This PR's head `0df4a66f3e` is an ancestor of the composition tip `2dd6e33ebc` (`integration/ds41-thirteen`, PR #1207), which merged to main as squash commit `b79bc1151b9a` at 2026-09-21T22:48:02Z. Containment was proved before closing: all **133** files the composition changed are byte-identical on main.

**What is *not* claimed.** This line is a stack — later layers edit the same files — so a file this PR touched may now carry a *later* layer's content rather than this one's. The claim is that these commits are inside what landed, not that every one of the 25 files still reads as this PR left it. That is normal for an intermediate layer and is why the check used here is ancestry-of-the-composition plus the composition's own containment proof, rather than per-file identity.

PR #1147, opened 2026-09-19
DeepSeek-V4.1 Flash: the single-Spark decode retune

rrstesiak:
Draft claim, stacked on #1144 and #1146 (base main, cumulative head; GitHub will not base a PR on a fork branch). The single-Spark decode step of DeepSeek-V4.1 Flash, measured, and the five changes that follow from the measurement. Plan and numbers in `docs/perf/DS41_DECODE_RETUNE_2026-09.md`.

**The starting point.** Warm decode step at the decode-fix binary, every routed expert resident, chat endpoint, 100 tokens: 65.3 ms = expert compute 23.5 + other 17.4 + attention 13.4 + routing 10.0 + engram 0.9. Per token the GPU sees 3,184 kernel launches, 234 stream synchronizations and 49 device-to-host copies.

| kernel, per token | ms | launches |
|---|---|---|
| `kquant_mmvq_q2_k_w` (attention projections) | 13.0 | 596 |
| `kquant_mmvq_q2_k_experts_w` (gate, up), ~175 GB/s | 10.1 | 79 |
| `kquant_mmvq_q3_k_experts_w` (down), ~134 GB/s | 8.9 | 39 |
| `dense_gemv_bf16`, incl. one 5.5 ms instance: the LM head held as bf16 (1.3 GB; the GGUF has it Q6_K, 0.5 GB) | 7.9 | 22 |
| `moe_v41_router_gemv_f32out`, 134 us for ~4 MB | 5.4 | 40 |
| hyper-connection chain (`mixes_finish`, `hc_post`, `mixes_dot`, `collapse`) | 7.4 | 320 |
| attention proper (`sparse_attn`, `slice_cols`, `index_score`, `gemm_f32`) | 5.0 | 374 |
| `kquant_q8_1_rows_bf16` | 0.85 | 714 |

Kernel time sums to about 63 ms: the GPU is saturated with small kernels, not with bandwidth. The expert reads are 19 ms against an 11 ms floor at LPDDR rates. The other 44 ms is launches, synchronizations, an unquantized head, a mis-shaped router GEMV and elementwise chains.

**The levers, in order, each a claim on this PR**
1. The LM head stays Q6_K instead of being dequantized to bf16 at load.
2. The per-layer attention projections grouped into one launch per shared input, same arithmetic, same accumulation order.
3. The router GEMV retuned to saturate, then top-k and the expert plan on the device so a warm token never syncs to the host; then the whole-step CUDA graph.
[...]

PR #1148, opened 2026-09-19
perf(ds41): decode retune phase 2, misses and host waits

rrstesiak:
DeepSeek-V4.1 Flash on one DGX Spark (GB10), the decode retune, phase 2: the cache misses, the host waits, then the remaining kernel time, then speculative decoding on top. Chained after #1147 (phase 1, five levers, `ds41-decode-retune`), which this branch contains; base `main`.

**Competing numbers first.** Two Sparks, EXL3 (Turney): ~30-32 tok/s. One Spark, this line, phase 1 (#1147): 12.39 (MinHeap) / 12.90 (Volvo) tok/s on the standard (chat, 300 tokens, temperature 0, one serve, medians of three), hot 120-token probe 18.9 tok/s.

**The standard.** `~/code/atlas-notes/bin/ds41_suite.sh` against one serve (`ATLAS_DS41_EXPERT_CACHE_GIB=88 ATLAS_DS41_READER_THREADS=8`, `--max-seq-len 4096 --max-prefill-tokens 512 --gpu-memory-utilization 0.25`), MinHeap x3 then Volvo x3, every output text byte-identical to the phase 1 oracle (`~/dflash-logs/ds41_retune_oracle_{minheap,volvo}_r{1,2,3}.txt`). Receipts `~/dflash-logs/ds41_suite_retune2_*`. One comment per stage below, kept or reverted, with the table.

**Stages.** S0 the byte floor (first comment). S1 misses: measure, cache policy, prefetch, the NVMe path. S2 host waits: device routing, a miss-replay protocol, the whole-step graph. S3 remaining kernel time. S4 speculative decoding with greedy verification (lookup drafts first), the only route past the single-token floor.

Plan document: `docs/perf/DS41_DECODE_RETUNE_2026-09.md` (phase 1 results and the final profile; phase 2 results appended as they land).


rrstesiak:
## S0, the byte floor (measured 09-19, before any code)

**What one decode token reads** (GGUF header walk over the seven shards; every non-expert tensor once, 6 of 384 routed experts per layer):

| family | MB / token |
|---|---|
| routed experts, 240 x 12.22 MiB (gate + up Q2_K, down Q3_K) | 2,930 |
| attention: q_b 550 + o_b 550 + o_a 440 + q_a 86 + kv 34 (Q2_K) | 1,660 |
| shared experts (Q2_K / Q3_K) | 512 |
| LM head (Q6_K) | 543 |
| router gate (bf16, stays bf16: the routing is bit-identical to the reference) | 157 |
| engram wkv 103, HC fns 13, norms and biases < 3 | 119 |
| **total** | **6,090** |

**What this GB10 delivers** (nvcc microbench, int4 streaming read, best of five): device memory 249-251 GB/s (273 theoretical); the GPU reading page-locked host memory through its device alias, which is where the expert arena lives, 223 GB/s; d2d memcpy 123 GB/s each way.

**The floor:** 2.93 GB / 223 GB/s + 3.16 GB / 249 GB/s = 13.1 + 12.7 = **25.8 ms per token = 38.7 tok/s** at 100% of the measured read ceiling with no launch gaps, no host waits and no compute. The 40 tok/s single-token target (25.0 ms) is below that floor for this checkpoint on one Spark with byte-identical math. An engineered ceiling at ~85% of bandwidth is ~30 ms, ~33 tok/s.

**NVMe** (Samsung MZALC4T0HBL1, 4 TB): random 1 MiB positional reads, buffered, 1 / 4 / 8 / 16 / 32 threads = 1.6 / 6.4 / 9.0 / 10.3 / 11.1 GB/s; O_DIRECT at 8 threads 10.2, 12 MiB chunks 11.7 GB/s. Phase 1's fetch path ran at ~6 GB/s with 8 reader threads.

**No MTP / next-n tensors in the V4.1 GGUF** (header checked: only `blk.0-39`, `output`, `output_norm`, `token_embd`), so S4's draft source is lookup drafts (n-gram pool), no draft model.

The stages S1-S3 push toward the ~33; S4 (speculative decoding, greedy verification) is the only route past the single-token floor, and it amortizes the 6.09 GB over the accepted tokens per step.

rrstesiak:
## S1, the misses: KEPT (arena 100 GiB + 16 readers); prefetch built, measured, shipped off

**Measured first** (`ATLAS_DS41_ROUTE_TRACE`, one line per fetch, the standard's six requests): at 88 GiB (7,376 slots) **1,792 of 1,794 decode steps miss**, 22,191 misses = 12.4 a step = 5.2% of the 240 accesses; hit steps 53.6 ms, miss steps 79.8 ms. One 12.22 MiB expert reads in 1.8-2.0 ms however it is split (8 parts buffered 2.04 ms, O_DIRECT 4 parts 1.76 ms); the disk's 11 GB/s only appears with several experts in flight. 12.4 x 2 ms is the whole gap between the standard and the hot number. The static top-7,376 set covers 95.2% of the accesses, so the misses are the distribution's tail, not an LRU artifact: LFU, 2Q and LRU-2 are no better, S3-FIFO is -26% at 88 GiB but worse at 100 GiB, and Belady's 2x comes from r1/r2/r3 being the same request (not a policy to tune for).

| configuration | MinHeap tok/s (r1/r2/r3) | Volvo tok/s (r1/r2/r3) | TTFT | decode misses | hit / miss step | oracle |
|---|---|---|---|---|---|---|
| 88 GiB, 8 readers (phase 1, reproduced) | 12.38 (10.77/12.42/12.38) | 12.85 (11.43/12.85/12.89) | 2013 / 1245 ms | 22,191 | 53.6 / 79.8 ms | 6/6 |
| **100 GiB, 16 readers** | **14.20** (10.76/14.21/14.20) | **17.73** (12.03/17.73/18.45) | 4560 / 432 ms | 12,255 | 55.9 / 74.9 ms | 6/6 |
| 100 GiB, 16, reader pool, no prefetch | 13.79 (10.02/13.90/13.79) | 17.78 (11.63/17.78/18.47) | 1746 / 421 ms | 12,255 | | 6/6 |
| 100 GiB, 16, pool + prefetch K=4 | 12.50 (10.51/13.52/12.50) | 16.67 (11.85/16.67/16.70) | 1729 / 420 ms | 8,240 | 60.7 / 80.7 ms | 6/6 |
| 100 GiB, 16, pool + prefetch K=6 | 13.63 (9.85/13.64/13.63) | 16.39 (12.01/16.40/16.39) | 1633 / 427 ms | | | 6/6 |
| 100 GiB, 16, pool + prefetch K=12 | killed at ~1.8 s a step | | | | | |

Launches and syncs unchanged from phase 1 (1,866 / 186 a token): S1 touches no kernel.

**The arena.** 100 GiB = 8,380 slots (+13.6%) halves the misses (the simulator predicted 12,262; the run recorded 12,255). MemAvailable is ~10 GB during load; the ~17 GiB of page cache given up shows as +2.3 ms on hit steps (the engram row reads, 48 a token, go to disk). Recipe: `ATLAS_DS41_EXPERT_CACHE_GIB=100 ATLAS_DS41_READER_THREADS=16` on a Spark running nothing else; the code default stays 88 GiB, readers 8 -> 16. MinHeap's TTFT at 100 GiB is its prefill's misses (the second request's prefill no longer hits after 300 tokens of Volvo evictions at 88 GiB happened to fit); TTFT is not in the medians.

**Prefetch.** Predictor = the next layer's router on this layer's MoE input (`ATLAS_DS41_PREDICT_TRACE`): d=1 top-6 covers 67.6% of the picks and 48.5% of the misses, top-12 81.8% / 69.1%. But the COLD part of a prediction, the reads a prefetch issues, has 36% precision at top-4 (6 reads a token, 2.3 of 6.8 misses caught), 23% at top-6 (15 reads), 8% at top-12 (58 reads). Built anyway: a persistent reader pool (urgent and background lanes, a reserve of readers that never take background work so a miss always starts at once), a completion ticket per slot, predicted experts read in the background and demoted to the LRU end when unused (`expert_prefetch.rs`, `expert_lru_pool.rs`). Measured: K=4 cut the misses 12,255 -> 8,240 and still lost, because the predictor's router launch is 68 us x 40 = 2.7 ms a token of GPU time and the background reads slow the remaining misses on the shared disk (every step ~5 ms slower). Shipped off by default (`ATLAS_DS41_READER_POOL=1 ATLAS_DS41_PREFETCH_K=4` turns it on); worth a second look once the router GEMV is cheap (S3) and with a cap on background reads in flight.

Receipts `~/dflash-logs/ds41_suite_retune2_{base8,c100t16,pool0,pf4,pf6}_*`, traces `~/dflash-logs/ds41_route_trace_*.txt`, `ds41_predict_trace_predc100.txt`; simulator and analyzer `~/code/atlas-notes/bin/ds41_cache_sim.py`, `ds41_predict_stats.py`.

rrstesiak:
## S4 ceiling, measured before any engine work: lookup drafts cannot carry this suite to 40

`~/code/atlas-notes/bin/ds41_lookup_sim.py` replays the six oracle texts through an n-gram lookup drafter offline (the V4.1 tokenizer; the longest match of the last n tokens in prompt + generated-so-far proposes the K tokens that followed it; the greedy-matching prefix is accepted; a step emits accepted + 1), the same mechanism as the flash-next lookup drafts (#974, #1026, #1060), with no draft model and no extra weight bytes.

| suite (300 tokens) | matches of 2..6 tokens: tokens a step at K = 1 / 2 / 4 / 8 | matches of 1..4 tokens | matches of 3..8 tokens |
|---|---|---|---|
| MinHeap | 1.07 / 1.10 / 1.10 / 1.11 (drafts fire on 41-51 of ~270 steps; 10-37% of drafted tokens accepted) | 1.13 / 1.16 / 1.19 / 1.20 (drafts on 125 steps; 5-24% accepted) | 1.04-1.06 |
| Volvo | 1.03 / 1.03 / 1.04 / 1.04 (drafts on 21-25 steps) | 1.07 / 1.08 / 1.09 / 1.09 | 1.01-1.03 |

The texts are fresh code and prose with almost no repeated n-grams. And a K+1-token verify step is not a single-token step: it reads the union of the tokens' routed experts (up to (K+1) x 6 a layer out of 384, little overlap between tokens), so the 2.93 GB MoE half of the step grows with K while only the dense 3.16 GB is amortised; at 1.1-1.2 accepted tokens a step the verify costs more than it returns. The GGUF ships no MTP / next-n tensors (S0), so there is no free draft head, and a separate draft model would need memory the 100 GiB expert arena does not leave and a 129,280-vocabulary drafter that does not exist.

Stated here with the numbers so the S4 engine work (m>1 verify at start_pos>0, per-token window views, compressor / indexer / engram rollback) is not built for a return the drafts cannot produce on this suite. 40 generated tok/s on MinHeap + Volvo is not reachable on one GB10 with this checkpoint by single-token decode (25.8 ms floor, S0) nor by lookup-draft speculation (x1.1-1.2 on a base that costs more per step). The remaining honest gains are S2 + S3 toward the ~30-33 single-token ceiling.

rrstesiak:
## Phase 2, final: where one Spark ends, in numbered facts and the table

**(a) The single-token byte floor.** One decode token reads 6.09 GB of weights; this GB10 delivers 249 GB/s from device memory and 223 GB/s reading the page-locked expert arena: **25.8 ms a token = 38.7 tok/s** at 100% of the read ceiling with no gaps, no waits, no compute. 40 tok/s single-token is below the floor of this checkpoint on one Spark. Practical engineered ceiling ~30-33.

**(b) The lookup-draft ceiling.** On the standard's own texts an n-gram drafter yields 1.10-1.20 accepted tokens a step on MinHeap and 1.03-1.09 on Volvo at any K (offline replay, `ds41_lookup_sim.py`); a K+1 verify step reads the union of the tokens' routed experts (up to (K+1) x 6 of 384 a layer), so the 2.93 GB MoE half grows with K while only the dense 3.16 GB is amortised: the verify costs more than it returns. No MTP head ships in the GGUF.

**(c) What passes 40.** Two GB10s with 2-way expert parallel: 1.465 GB routed at 223 GB/s + 3.16 GB dense at 249 GB/s = **19.3 ms = 51.9 tok/s** before the interconnect, ~140 us a layer round trip to spare against ~10 us RDMA messages (one 10 KB row per layer per direction), and every expert resident (187.7 GB split 94 GB a Spark into a 100 GiB arena: zero misses at any output length). One B200: 6.09 GB at ~8 TB/s = 0.76 ms, launch-bound, hundreds of tok/s with a whole-step graph (#1140 / #1141); the bit-exact device-side selection that graph needs is prototyped and verified under `docs/perf/ds41_prototypes/` (glibc's `expf` / `log1pf` on the device, 0 mismatches over all 2^32 inputs; picks, weights and plan identical to the host on 20,000 tokens).

**(d) The table.** Chat, 300 tokens, temperature 0, one serve, medians of three, every text byte-identical to the oracle (6/6 on every row). Recipe: `ATLAS_DS41_EXPERT_CACHE_GIB=100 ATLAS_DS41_READER_THREADS=16`.

| step | commit | MinHeap tok/s (r1/r2/r3) | MinHeap TTFT | Volvo tok/s (r1/r2/r3) | Volvo TTFT | launches / syncs a token | decode misses |
| baseline (09-17 binary, 88 GiB, 8 readers) | f13752b23 | 10.42 | 2048 ms | 10.77 | 1291 ms | 3,184 / 234 | |
| phase 1, five levers (#1147) | c5ca8e2d5 | 12.39 | 2009 ms | 12.90 | 1262 ms | 1,866 / 186 | 22,191 |
| S1: 100 GiB arena, 16 readers | 7866373ea | 14.20 (10.76/14.21/14.20) | 4560 ms | 17.73 (12.03/17.73/18.45) | 432 ms | 1,866 / 186 | 12,255 |
| S1: + random victim among the oldest 5% | | 12.76 (10.79/16.07/12.76), swap stalls | 4086 ms | 17.87 (12.02/17.87/18.32) | 433 ms | | 9,914 |
| **S1: + page cache dropped after every expert read** | 998bf8986 | **15.79** (9.99/15.79/16.40) | 1569 ms | **17.67** (11.22/17.67/17.97) | 399 ms | 1,865 / 186 | 9,914 |

Baseline to phase 2: MinHeap +51.5%, Volvo +64.1%; phase 1 to phase 2: +27.4% / +36.9%. Hit steps are 56-57 ms (17.6-17.9 tok/s) on both suites; MinHeap's median still carries 1.2-1.9 misses a step at 100 GiB.

**The working-set edge** (why MinHeap trails Volvo): MinHeap's working set is 8,418 distinct (layer, expert) pairs against 8,380 slots at 100 GiB; Volvo's 8,005 fits. A cyclic replay 0.5% larger than the cache is strict LRU's pathological case (5.5 misses a step on r2/r3); a random victim among the oldest 5% breaks the lockstep (1.9 / 1.2 a step, Volvo unchanged at 0). 102 GiB gives MinHeap zero misses (17.59 median) but swaps (Volvo r2 14.07, TTFT 11.7 s): not a recipe. The suites share 5,200 of 11,223 experts, so every switch between them costs ~3,000 misses (r1 of each: 17.6 / 12.3 a step); the medians of three absorb that by design. **Long outputs** (one MinHeap request, 992 tokens): the working set keeps growing (8,423 at 300 tokens, 10,650 at 992, ~3 a token, no plateau), so one Spark keeps missing on long outputs; the all-resident case is two Sparks.

**The stalls.** The suite's tok/s is a mean, and at the 100 GiB edge the kernel reclaimed page cache into the 16 GB swap file mid-step (single steps of 1.9 to 17.6 s with 3-11 misses in them). `posix_fadvise(DONTNEED)` after every expert pread ends it: the longest MinHeap r2 / r3 step 1.9 / 5.3 s -> 107 / 115 ms.

**S2 / S3, time-boxed, none landed.** The per-layer graph path (`ATLAS_DS41_GRAPH=1`) loses on the S1 binary (9.77 / 15.19, erratic); a whole-step graph with one miss read-back a step cannot pay with 65% of steps missing (a miss at layer L wastes the 40 - L layers replayed after it). The GEMV unroll measured neutral (10.82 vs 10.03 ms), reverted; the router chain double-buffer faulted, reverted. Final profile (nsys, one warm 60-token request, 19.86 tok/s hot): kernel time 49.6 ms a token over 1,865 launches, experts 17.5, attention projections 15.1, head 2.9, router 2.7, sparse attention 2.1, HC 2.5; 186 syncs, 49 D2H, 260 H2D. The largest kernel inefficiency left is the attention projections at 133-152 GB/s (1.66 GB in ~15 ms); it needs a shape change, not an unroll.

Also on the branch, all measured and off by default: the persistent reader pool with a predictive prefetch (`ATLAS_DS41_READER_POOL=1 ATLAS_DS41_PREFETCH_K=4`; cold predictions have 36% precision and the predictor's router launch costs 2.7 ms a token, K=4 lost 12.50/16.67), `ATLAS_DS41_ROUTE_TRACE` and `ATLAS_DS41_PREDICT_TRACE` diagnostics, `ATLAS_DS41_EVICT_RANDOM_PCT` (default 5, 0 = strict LRU), `ATLAS_DS41_KEEP_PAGE_CACHE=1`. The O_DIRECT (-12% a miss, needs a padded slot layout for the 160-mod-512 tensor offsets), GPUDirect Storage (`nvidia-fs` present, not loaded, no system `libcufile`) and device-arena (`cudaMalloc` memory is not CPU-accessible on GB10) items are recorded in the plan doc as deferred.

Receipts `~/dflash-logs/ds41_suite_retune2_{base8,c100t16,pool0,pf4,pf6,graph1,c102t16,lrur5,s3b}_*`, traces `~/dflash-logs/ds41_route_trace_*.txt`, profile `~/dflash-logs/ds41_nsys_decode_final*`, GPU tests 9/9 (`~/dflash-logs/ds41_gpu_tests_s3b.log`, serial), runtime tests 8/8, clippy and the 500-line cap clean. Plan document: `docs/perf/DS41_DECODE_RETUNE_2026-09.md`, phase 2 section.

rrstesiak:
Rebased onto main `4c00f65a2` after #1144 merged. No content change: the phase 2 commits now sit on main; one extra commit moves the engram row reader into its own file to stay under the 500-line cap on the rebased tree; the rebased binary re-ran the standard at 15.93 / 18.04 tok/s, all six texts byte-identical. Head `9b0748d04`.

tbraun96:

**Citation.** This PR's head `9b0748d04d` is an ancestor of the composition tip `2dd6e33ebc` (`integration/ds41-thirteen`, PR #1207), which merged to main as squash commit `b79bc1151b9a` at 2026-09-21T22:48:02Z. Containment was proved before closing: all **133** files the composition changed are byte-identical on main.

**What is *not* claimed.** This line is a stack — later layers edit the same files — so a file this PR touched may now carry a *later* layer's content rather than this one's. The claim is that these commits are inside what landed, not that every one of the 57 files still reads as this PR left it. That is normal for an intermediate layer and is why the check used here is ancestry-of-the-composition plus the composition's own containment proof, rather than per-file identity.

PR #1153, opened 2026-09-19
perf(ds41): decode retune phase 3, the hit step

rrstesiak:
DeepSeek-V4.1 Flash on one DGX Spark (GB10), the decode retune, phase 3: the hit step. Chained after #1148 (phase 2, `ds41-decode-retune-2`), which this branch contains; base `main`.

**Competing numbers first.** Two Sparks, EXL3 (Turney): ~30-32 tok/s. One Spark, this line: phase 1 (#1147) 12.39 / 12.90, phase 2 (#1148) 15.79 / 17.67 tok/s (MinHeap / Volvo) on the standard; the rebased phase 2 head re-measured 16.19 / 18.21 today as this phase's baseline (`p3base`).

**The standard.** `~/code/atlas-notes/bin/ds41_suite.sh` against one serve, chat endpoint, 300 tokens, temperature 0, MinHeap x3 then Volvo x3, medians of three, the phase 2 recipe (`ATLAS_DS41_EXPERT_CACHE_GIB=100 ATLAS_DS41_READER_THREADS=16`), every output text byte-identical to the phase 1 oracle (`~/dflash-logs/ds41_retune_oracle_*`). Receipts `~/dflash-logs/ds41_suite_retune3_*`. One comment per lever below, kept or reverted, with the table.

**What this phase found first.** The brief's P1/P2 thesis (the K-quant GEMVs latency-bound on dependent block loads) does not hold: register-prefetch variants of `kq_mmvq_warp` (4/8/16 super-blocks in flight, 1/2/4/8 warps a block) are bit-identical and slower at every V4.1 shape (isolated bench, ring of weights larger than L2), and the shipped kernels already run at 150-215 GB/s of the 249 measured. The per-launch trace of the phase 2 hit step (50.6 ms wall, 47.9 GPU, 2.7 idle) pointed elsewhere: the host span idling the GPU every layer, the router's sequential chain, two source-layer kernels on one or a few blocks, the hyper-connection finish on one thread, and the engram projection read as bf16 at six times its GGUF size. The levers below are those.

Plan document: `docs/perf/DS41_DECODE_RETUNE_2026-09.md` (a Phase 3 section appended as the levers land).


rrstesiak:
**L1, the expert arena mapped write-combined: REVERTED (neutral).** `cuMemHostAlloc(DEVICEMAP | WRITECOMBINED)` for the page-locked arena the host only writes (pread) and the GPU only reads. Isolated, the expert GEMV reads a WC ring at 184-196 GB/s against 159-180 for the default cacheable mapping (device memory: 213); on the standard it is inside the run-to-run band, in both directions.

| binary | arena | MinHeap tok/s (r1/r2/r3) | Volvo tok/s (r1/r2/r3) | oracle |
|---|---|---|---|---|
| p3base 9b0748d04 | pinned | 16.19 (10.29/16.19/16.74) | 18.21 (11.43/18.21/18.37) | 6/6 |
| L1 689787c4f | WC | 16.37 (10.30/16.37/16.47) | 17.82 (11.46/18.18/17.82) | 6/6 |
| L3 binary, `ATLAS_DS41_ARENA_WC=1` | WC | 17.75 (10.86/17.75/18.39) | 20.14 (12.24/20.14/20.24) | 6/6 |
| L3 binary, `ATLAS_DS41_ARENA_WC=0` | pinned | 17.58 (8.67/17.58/18.09) | 19.75 (12.12/19.75/19.85) | 6/6 |

Reverted in 4a30b6b2c so the branch carries only levers that moved the number; the allocator entry point and the measurement stay in the history.

**L2, the shared expert on a side stream: KEPT.** At decode the shared expert depends only on the MoE input, yet it ran after the routed experts, so the GPU idled through every layer's host span (the router read-back, the top-k, the fetch bookkeeping, the uploads: 2.7 ms a token idle in the phase 2 trace) and the router chain ran nearly alone. The single-token forward now records an event once the input is ready, launches the router on the main stream, runs the shared expert's three GEMVs, SwiGLU and quantisations on a side stream with its own q8_1 scratch, and the main stream waits on the side's event before the same `accumulate` + `finish`. The order of additions into `acc` is unchanged (routed rows in plan order, then the shared expert), so the bits are phase 2's. `ATLAS_DS41_SHARED_SIDE=0` restores the serial order. (758df0dde)

| step | MinHeap tok/s (r1/r2/r3) | MinHeap TTFT | Volvo tok/s (r1/r2/r3) | Volvo TTFT | oracle |
|---|---|---|---|---|---|
| p3base | 16.19 (10.29/16.19/16.74) | 1212 ms | 18.21 (11.43/18.21/18.37) | 423 ms | 6/6 |
| L2 | **17.11** (10.63/17.11/17.62) | 1213 ms | **19.34** (11.92/19.34/19.40) | 406 ms | 6/6 |

**L3, the compressor product staged, the index score in registers: KEPT.** Two source-layer kernels ran on one or a few blocks with a load on every step of their chains: `attn_v41_gemm_f32` (the 16x16 tile at m = 1, 32 blocks, 144 us a launch, six a token) and `attn_v41_index_score` (one block, 122 us, eight a token). `attn_v41_gemv_f32_staged`: one output a block, 256 threads stage the row, thread 0 runs the strict k chain in the tiled kernel's order (tile after tile, k ascending, mul then add under `--fmad=false`): the same bits, 144 -> 95 us isolated. `attn_v41_index_score` keeps the key row in registers and reads each query row 16 bytes at a time when `ihd` is 128 (the generic loop stays for other widths): same products, same order, same bf16 roundings, 135 -> 23 us isolated. (c87a65510)

| step | MinHeap tok/s (r1/r2/r3) | MinHeap TTFT | Volvo tok/s (r1/r2/r3) | Volvo TTFT | oracle |
|---|---|---|---|---|---|
| L2 | 17.11 (10.63/17.11/17.62) | 1213 ms | 19.34 (11.92/19.34/19.40) | 406 ms | 6/6 |
| L3 | **17.75** (10.86/17.75/18.39) | 1361 ms | **20.14** (12.24/20.14/20.24) | 422 ms | 6/6 |

rrstesiak:
**L4, the engram `wkv` on its raw Q2_K blocks: REVERTED (not byte-identical).** The GGUF ships the engram projection (25,600 x 6,144, two layers) as Q2_K, 52 MB a layer; the loader expands it to bf16, 315 MB a layer, read every token at 245 GB/s (2.6 ms a token). Running the K-quant GEMV on the raw blocks, as phase 1 did for the Q6_K head, needs the 25,600-wide engram row quantised to q8_1 first; the head's input row is the normed hidden state, which the bf16 GEMV also sees exactly, but here the bf16 path computes bf16(row) x dequant(Q2_K) exactly while the K-quant path rounds the row to 8 bits per 32-block. It is faster and it is a different model: MinHeap 19.62 / Volvo 21.55 with 0/6 texts identical (MinHeap diverges at character 100, Volvo at 51). Off by default is not enough for a lever that changes the numbers, so the code is out of the branch (the patch is kept locally). What passes the oracle here would be a bf16-activation Q2_K GEMV (dequantise the weight block in registers, multiply in f32 in the tiled kernel's order): 52 MB instead of 315 a layer, ~2 ms a token, not built in this phase.

**L5, the hyper-connection finish on the warp's lanes: KEPT.** The Sinkhorn finish of each HC site (sigmoids, a row softmax, twenty column / row normalisations of the 4 x 4 mix) ran on one thread, ~17 us a call, 80 sites a token (1.5 ms). `hcv_finish_lanes<HC>` puts one element on each of the 16 lanes; row and column sums are gathered by shuffles and added in the serial form's order from 0.0f, the max, `expf` and every division stay per element, so the bits are the serial form's.

**L6, the router and compressor chains from precomputed products: KEPT.** The router's strict 5,120-step chain ran on two lanes a block (64.7 us for a 4 MB weight, 40 a token = 2.6 ms). Now the block's 256 threads write the products into shared memory (bf16 x bf16 is exact in f32) and one lane adds them in k order eight float4 at a time from registers: no load on the FADD chain, 63.9 -> 33.5 us a launch isolated, bit-identical. The compressor product of L3 rewritten the same way: 95 -> 67 us.

| step | MinHeap tok/s (r1/r2/r3) | MinHeap TTFT | Volvo tok/s (r1/r2/r3) | Volvo TTFT | oracle |
|---|---|---|---|---|---|
| L3 | 17.75 (10.86/17.75/18.39) | 1361 ms | 20.14 (12.24/20.14/20.24) | 422 ms | 6/6 |
| L4 + L5 + L6 (engram Q2_K on) | 19.62 (11.29/19.62/20.66) | 1098 ms | 21.55 (12.70/21.55/21.72) | 411 ms | **0/6** |
| L5 + L6 | **18.41** (11.03/18.41/18.91) | 1223 ms | **20.81** (12.38/20.81/20.94) | 406 ms | 6/6 |

rrstesiak:
**Phase 3 final.** Head f1aacfc01 (code head c1fcb64a4; binary `~/ds41_bin/spark-retune3-final-c1fcb64a4`). The standard, clean box, nothing else running (receipts `~/dflash-logs/ds41_suite_retune3_final2_*`):

| | MinHeap tok/s (r1/r2/r3) | MinHeap TTFT | Volvo tok/s (r1/r2/r3) | Volvo TTFT | oracle |
|---|---|---|---|---|---|
| baseline, 09-19 morning | 10.42 | 2048 ms | 10.77 | 1291 ms | |
| phase 1 (#1147) | 12.39 | 2009 ms | 12.90 | 1262 ms | 6/6 |
| phase 2 (#1148) | 15.79 (9.99/15.79/16.40) | 1569 ms | 17.67 (11.22/17.67/17.97) | 399 ms | 6/6 |
| phase 2 head rebased on main (this phase's baseline) | 16.19 (10.29/16.19/16.74) | 1212 ms | 18.21 (11.43/18.21/18.37) | 423 ms | 6/6 |
| **phase 3** | **18.43** (11.06/18.43/18.94) | 1214 ms | **20.81** (12.46/20.81/20.92) | 399 ms | 6/6 |

Phase 3 against its baseline +13.8% / +14.3%; against phase 2 as published +16.7% / +17.8%; against the morning baseline +77% / +93%. Hot probe (one warm 60-token request under nsys) 19.86 -> 22.79 tok/s.

**The target of 22 / 25 on the standard was not reached.** Kept: L2, L3, L5, L6. Reverted: L1 (neutral). Dropped: L4 (0/6). Checks on the head: the serial GPU oracle tests 11/11, the runtime expert tests 14/14, `cargo fmt`, `clippy -D warnings` on the two edited crates, the 500-line cap.

Final profile (nsys, same method as phases 1-2): 44.2 ms wall a warm token, 45.0 ms GPU busy (the side stream overlaps), 1.6 ms of idle gaps (was 2.7), 1,865 launches, 186 syncs, 49 D2H. Family by family: routed experts 17.5 ms (2.93 GB from the pinned arena at 167 GB/s; the same kernel reads device memory at 213), attention projections 9.6, LM head 2.8, shared expert 3.5 (now under the host span), engram 2.6, sparse attention 2.4, router 1.65 (was 2.6), HC 1.4 (was 2.5), compressor + index score 0.5 (was 1.8).

**What the gap to 22 / 25 is made of, in numbers.** (a) Seven bandwidth-bound GEMV families read 6.09 GB a token at 150-215 GB/s; the register-prefetch rewrites of the K-quant GEMV that the brief prescribed are bit-identical and slower at every shape (measured before any lever), so the remaining kernel-side room is the 167 -> 213 GB/s of the expert reads, which is where the bytes live (pinned host), not how the kernel reads them. (b) MinHeap still carries 1-2 misses a step at 100 GiB, ~2 ms each. (c) 186 host waits a step; the device-side routing and whole-step graph that remove them do not pay while 65% of steps miss (phase 2, S2). The plan document `docs/perf/DS41_DECODE_RETUNE_2026-09.md` carries the Phase 3 section with every number above.

tbraun96:

**Citation.** This PR's head `f1aacfc014` is an ancestor of the composition tip `2dd6e33ebc` (`integration/ds41-thirteen`, PR #1207), which merged to main as squash commit `b79bc1151b9a` at 2026-09-21T22:48:02Z. Containment was proved before closing: all **133** files the composition changed are byte-identical on main.

**What is *not* claimed.** This line is a stack — later layers edit the same files — so a file this PR touched may now carry a *later* layer's content rather than this one's. The claim is that these commits are inside what landed, not that every one of the 57 files still reads as this PR left it. That is normal for an intermediate layer and is why the check used here is ancestry-of-the-composition plus the composition's own containment proof, rather than per-file identity.

PR #1155, opened 2026-09-19
perf(ds41): decode retune phase 4, the bytes and the waits

rrstesiak:
DeepSeek-V4.1 Flash on one DGX Spark (GB10), the decode retune, phase 4: the bytes and the waits. Chained after #1153 (phase 3, `ds41-decode-retune-3`), which this branch contains; base `main`.

**Competing numbers first.** Two Sparks, EXL3 (Turney): ~30-32 tok/s. One Spark, this line: phase 1 (#1147) 12.39 / 12.90, phase 2 (#1148) 15.79 / 17.67, phase 3 (#1153) 18.43 / 20.81 tok/s (MinHeap / Volvo) on the standard; the phase 3 head re-measured 18.27 / 20.76 tonight as this phase's baseline (`p4base`).

**The standard.** `~/code/atlas-notes/bin/ds41_suite.sh` against one serve, chat endpoint, 300 tokens, temperature 0, MinHeap x3 then Volvo x3, medians of three, the recipe (`ATLAS_DS41_EXPERT_CACHE_GIB=100 ATLAS_DS41_READER_THREADS=16`), a clean box, every output text byte-identical to the phase 1 oracle (`~/dflash-logs/ds41_retune_oracle_*`). Receipts `~/dflash-logs/ds41_suite_retune4_*`. One comment per lever below, kept or reverted, with the table.

**What this phase measured first.** The phase 3 profile left the decode step at 44.2 ms with the routed-expert GEMVs at 17.5 ms of it, reading 2.93 GB a token out of the page-locked arena at 167 GB/s while the same kernel reads device memory at 213. Before touching anything, the real kernels at the real per-layer shapes were run under Nsight Compute from both kinds of memory: same L2 hit rate, same occupancy, same instruction mix, same stall reason (long scoreboard, ~80% of stall cycles), 21% longer from the pinned path. The kernel is not the difference; the address of the bytes is. So the first lever of this phase is the arena itself: the expert slots move into device memory behind a page-locked staging ring, and nothing in the math changes.

Plan document: `docs/perf/DS41_DECODE_RETUNE_2026-09.md` (a Phase 4 section appended as the levers land).

PR #1156, opened 2026-09-19
b200: the V4.1 kernel tree on sm_100a and the whole-step graph, proven on GB10

rrstesiak:
DeepSeek-V4.1 Flash for the B200, prepared on one DGX Spark. Branch `ds41-b200-prep` from cc06bf548 (#1155's head, which this branch contains); base `main`. Companion to the recipe PRs #1140 (one-GPU campaign), #1141 (two-GPU split, sm_100a retune, the port in three layers) and #1142 (spec decode): those say what to do with a B200; this one does what can be done without one.

**Why.** After #1155 the one-Spark decode step is 42.4 ms of wall for 42.8 ms of GPU work, 1,865 launches and 186 host waits a token, reading 6.09 GB. At 8 TB/s the bytes cost ~0.8 ms and the step becomes the launch and wait overhead (~8-12 ms). The B200 levers are therefore the device-side routing and the whole-step CUDA graph, neither of which pays on the Spark (misses; and the waits are hidden under the side stream since #1153) but both of which can be built and proven byte-identical here. And before any rental hour, the kernel tree has to compile for sm_100a.

**Nothing in this PR has run on a B200.** Every number below is a GB10 number or a compiler's; the README says so in each place.

**Steps, one comment each:** S1 the sm_100a gate over the V4.1 kernel set + the ptxas resource table + the source snapshot (`kernels/b200/deepseek-v4-flash/`); S2 device-side routing on the eager path (`ATLAS_DS41_DEVICE_ROUTE=1`, opt-in) with a diag mode that asserts equality with the host chain on every token; S3 the whole-step graph on hit tokens with eager fallback on the miss flag (`ATLAS_DS41_STEP_GRAPH=1`, opt-in), profiled on Volvo's zero-miss runs as the honest preview of B200 behaviour; S4 the first-hour runbook (`docs/perf/B200_FIRST_HOUR_2026-09.md`).

**The standard** for S2/S3 is #1155's: chat, 300 tokens, temperature 0, one serve, MinHeap x3 then Volvo x3, medians of three, recipe env, clean box, every text byte-identical to the phase 1 oracle. Baseline: 18.57 / 21.14 tok/s (MinHeap / Volvo), competing numbers Turney EXL3 two Sparks ~30-32.


rrstesiak:
## S1, the V4.1 kernel set through the sm_100a gate (316b35da6)

`scripts/hopper_ptx_gate.sh --hw b200 --model deepseek-v4-flash --strict --jobs 12` on CUDA 13.0.88, this box: self-test held (known-good fixture passed, the sm_100a negative fixture failed as it must), then **199/199 kernel sources compiled for sm_100a** (`nvcc --ptx -O3 --fmad=false -DAVAROK_NO_WARP_BLOCKSCALE_MMA --Werror all-warnings`, then `ptxas -v`): the 23 sources in `nvfp4/` (attn_v41, moe_v41, engram_v41, hc_v41, kquant_moe, the MLA/CSA/prefill set) and the 176 in `common/` the model directory reaches. **0 rejected entry functions, 0 source changes.** The brief's expected failure classes (sm_121-only intrinsics, shared-memory sizes, launch attributes, cp.async differences) did not occur: the V4.1 kernels are plain CUDA C with warp shuffles and 32-bit loads, written on the Spark without sm_121 features. No source touched means GB10 byte identity is the one already on `main` (#1155's 6/6 standard and the serial GPU tests); nothing had to be re-proven and nothing was.

For the record, the same gate against the **gb10** set at sm_121f reports 22 failures, all `inferspark_prefill*` entries in `common/` over the 48 KB static shared-memory cap in the gate's compile mode; they predate this branch, are not on the V4.1 path, and are not touched.

**The ptxas resource table** (`kernels/b200/deepseek-v4-flash/PTXAS_RESOURCES.md`): 105 entry functions of the 13 modules the V4.1 runtime loads, registers / static smem / spills on sm_100a against sm_121f. No spills on either architecture; sm_100a uses the same or fewer registers on every entry. The decode hot path:

| entry | regs sm_100a / sm_121f | smem |
|---|---|---|
| `kquant_mmvq_q2_k_experts_w8`, `q3_k_experts_w8` (routed experts) | 40 / 40 | 0 |
| `kquant_mmvq_q2_k_w` (attention projections) | 48 / 48 | 0 |
| `kquant_mmvq_q6_k_w` (head) | 40 / 48 | 0 |
| `moe_v41_router_gemv_f32out_products` | 40 / 44 | 0 |
| `attn_v41_sparse_attn` | 32 / 40 | 9,216 B |
| `hc_v41_mixes_dot` | 32 / 48 | 1,024 B |
| `attn_v41_index_score` | 164 / 164 | 0 |
| `atlas_q2_k_mmq128_{nc,wc}` (prefill MMQ) | 255 / 255 | 0 |
| `atlas_q3_k_mmq128_{nc,wc}` | 230-234 / 254 | 0 |

None of the decode kernels is occupancy-limited on either part; what the table does say is that the grids were sized for 48 SMs (`wq_a` 320 blocks, `wkv` 128: 0.67 and 0.27 waves on GB10, under a quarter of a wave on 148 SMs), which is #1141's retune, not this PR's.

**Deliverables in the tree:** `kernels/b200/deepseek-v4-flash/README.md` (build line, gate result, what a B200 changes and what it does not), `SOURCE_SNAPSHOT.json` (git blob at cc06bf548 and sha256 behind each of the 24 symlinks into `kernels/gb10`, so a build on rented hardware can be checked against what the gate compiled; the symlinks stay, being the mechanism every b200 model directory uses and the one `check_cross_hardware.py` rule S1 leaves standing), `PTXAS_RESOURCES.md`, and the gate's own ledger at `docs/perf/b200_ptx_gate_deepseek-v4-flash_2026-09-19.md`. `check_cross_hardware.py --base cc06bf548`: OK; `avarok-kernels` target tests (inherited_targets, target_defaults, blockscale_mma_guard): 37/37.

Nothing here has run on a B200.

rrstesiak:
## S2, the single-token routing on the device (`ATLAS_DS41_DEVICE_ROUTE=1`, opt-in): 6e0f1ca04

**What runs on the device now.** `moe_v41_route_select` (one block of 384 threads per token, 9 us): the reference's selection bit for bit, from the prototypes in `docs/perf/ds41_prototypes/`: glibc 2.39's `expf` and `log1pf` as device functions (identical to the host over all 2^32 inputs), score = sqrt(softplus(logit / temp)), key = score + bias ranked strictly (descending, index ascending on ties = Rust's stable `sort_by`), top-6, the weights with the sum in pick order (+1e-20, /, x scale as separate roundings), then the plan in ascending expert id: `weight_dev`, `rows_dev` and the gate / up / down pointer table straight from a device slot table `[64 layers][384]` that the host keeps current from the cache's (layer, expert, slot | -1) changes. The host reads ONE 19-word header per layer (miss flag, picks, weight bits, plan slots), touches the picks in the cache, fetches misses exactly as before, and re-uploads the pointer table only when the flag is set or a slot moved. On a hit token nothing else crosses the bus in that direction.

**Proof, three layers.** (1) GPU test `device_route_select_matches_the_host_chain_bitwise`: 400 tokens with forced key ties and every fifth slot absent: 0 mismatches in picks, weight bits, plan weights, plan slots, pointer table, miss flag. (2) `ATLAS_DS41_DEVICE_ROUTE_DIAG=1` over the full standard suite: the host chain runs on the same logits every layer-token and the step fails on any difference in picks / weight bits / plan weights / pointer table; 72,000 decode layer-tokens (plus the prefill layers) passed, 6/6 texts byte-identical (18.99 / 21.46 under diag). (3) The standard without diag:

| | MinHeap tok/s (r1 / r2 / r3) | TTFT | Volvo tok/s (r1 / r2 / r3) | TTFT | oracle |
|---|---|---|---|---|---|
| #1155 head (P3, host routing) | **18.57** (10.83 / 18.57 / 19.15) | 1285 ms | **21.14** (12.28 / 21.14 / 21.26) | 400 ms | 6/6 |
| S2 (device routing) | **19.19** (10.93 / 19.19 / 19.70) | 1293 ms | **21.87** (12.43 / 21.87 / 21.94) | 406 ms | 6/6 |

+3.3% / +3.5%; hot probe 23.72 -> 23.91 (24.09 on a first nsys run of the same binary). Expected neutral on the Spark; it is a small gain (the host's 384-way sort, plan build and three uploads per layer are gone from the host span).

**nsys, one warm token** (`ds41_nsys_decode_S2*`): GPU kernel time 44.5 ms (+0.35 ms of `moe_v41_route_select`, 40 x 9 us), launches 1,865 -> 1,905, `cuMemcpyHtoDAsync` 260 -> 142 a token, `cuStreamSynchronize` 186 -> 186 (one read-back per layer either way; the B200 lever is what S3 does with it: the header is now the ONE thing a whole-step replay has to read), D2H 49 -> 49.

Built for the B200, where the step is launch- and wait-bound. Receipts `~/dflash-logs/ds41_suite_retune4_{S2diag,S2}_*`, `ds41_serve_retune4_S2diag.log` (the diag lines), binary `~/ds41_bin/spark-retune4-b200prep-S2-6e0f1ca04`. Nothing here has run on a B200.

rrstesiak:
## S3, the step as segment graphs across layers (`ATLAS_DS41_STEP_GRAPH=1`): EXPERIMENTAL, default off, NOT KEPT (38c4c2097)

**What was built.** With S2's routing on the device and every engram layer's rows uploaded up front by layer 0, the only host work left in a decode step is the eight index-source layers' attention (index score read-back + host top-k + the compressor's position parity), so the step is captured as **nine segments** between them: 0..2, 2..8, 8..14, 14..20, 20..24, 24..28, 28..32, 32..36, 36..40. Layer 0 and each index layer own a segment; capturable layers append their bodies under an open capture; the next index layer appends its attention input and closes it. Each segment's layers write their routing headers to their own rows (`moe_v41_route_select` now writes SAFE pointers, slot 0, for an absent expert and sets the flag, so a replay never faults), and the owner reads all of them back in ONE copy after the launch. On GB10 all nine segments captured as designed, and the first MinHeap request under the first protocol produced the oracle's text.

**The numbers, honest.** nsys on that first run (three graph runs per segment on average, the broken protocol below): `cuLaunchKernel` 350 a token and `cuStreamSynchronize` 65 a token, against 1,905 and 186 on the S2 eager path. On a hit token that is ~100 launches and ~9 syncs. That collapse is the whole point on a B200; on the Spark the first request ran at 7.3 tok/s because of the re-runs, and the requests after it failed.

**What does not work: the miss fallback.** Inside a replayed segment nothing fetches. A missed expert at layer f reads slot 0, so f's output is wrong and every layer after f in the segment routes on a wrong input. The phase 2 "replay twice" protocol therefore does not transfer to multi-layer segments: the second run reveals NEW misses (instrumented: segment 8 run 0 flagged layers [2, 3, 5, 6], run 1 [3], run 2 [4]), up to one run per layer, and the wrong picks fill the cache with noise (MinHeap r2-r6: "still missing experts after 8 runs" on a full cache). The eager fallback the brief asks for is the right protocol, but it needs the driver to run the flagged layer's successors eagerly from the state after f-1, and the per-layer `step` cannot revisit layers it has passed (a segment closed and launched by layer c has s..c-1 behind it), so the fallback as written fails the request ("layer N reached outside any segment").

**The sound design, stated for the successor, not built:** (1) the graph snapshots hidden, streams, pre_prev and pre_a after EVERY layer (~90 KB a layer, D2D nodes), so a miss at f restores the state after f-1 and marks f..end eager while s..f-1 stand (their picks were real); (2) the capture token runs every layer eagerly on the compute stream while the bodies are recorded on a second stream, so a segment is only ever launched at its owner's step, never at its closing layer; (3) `eager_until` then covers f..end and the driver reaches those layers in order. On a B200 with every expert resident there are no misses and the segments are the step; on one Spark, at 65% miss steps, the fallback path would be the common one and the mode is expected to lose, as phase 2's per-layer graph did.

**What stands on its own in this commit** (all default paths untouched, GPU tests kquant 4 / moe_v41 4 / attn_v41 1 / engram 2 / real_file 1, fmt, clippy, cap): the per-layer header rows and safe pointers in `moe_v41_route_select`; `route_select_deferred` / `read_headers` / `parse_header`; the engram layers own their row workspaces; the attention's second selection buffer for window-only layers (`decode_prep_role`); the runtime's layer roles. Receipts: `~/dflash-logs/ds41_serve_retune4_S3.log` (the three runs overwrote each other; the instrumented protocol lines are from run 2), `ds41_nsys_decode_S3*` from run 1. Nothing here has run on a B200.

S4 (the first-hour runbook) is being written on `ds41-b200-runbook` by a docs-only worker; not in this PR.

tbraun96:

**Citation.** This PR's head `38c4c20973` is an ancestor of the composition tip `2dd6e33ebc` (`integration/ds41-thirteen`, PR #1207), which merged to main as squash commit `b79bc1151b9a` at 2026-09-21T22:48:02Z. Containment was proved before closing: all **133** files the composition changed are byte-identical on main.

**What is *not* claimed.** This line is a stack — later layers edit the same files — so a file this PR touched may now carry a *later* layer's content rather than this one's. The claim is that these commits are inside what landed, not that every one of the 71 files still reads as this PR left it. That is normal for an intermediate layer and is why the check used here is ancestry-of-the-composition plus the composition's own containment proof, rather than per-file identity.

PR #1150, opened 2026-09-19
WIP: Kimi K3 / B300 integration and rental notebook

TheTom:
## What and why

**Draft integration notebook — do not merge.** Prepare Atlas for a bounded official Kimi K3 bring-up on an 8×B300 node, with the avoidable loader, kernel-registration, launch and logging failures caught on two Sparks first. Based on main `4c00f65a2`; restores the missing serving connection from #1053 on top of the seven merged K3 component PRs. #895/#1045 and upstream implementation/issue lessons are recorded in the runbook.

Full official-model inference and useful throughput are **not yet demonstrated**. Start at [docs/k3/README.md](https://github.com/TheTom/atlas/blob/32ebc80d5fc2dc0fdce37e9a9793542a137fad21/docs/k3/README.md), then [the B300 runbook](https://github.com/TheTom/atlas/blob/32ebc80d5fc2dc0fdce37e9a9793542a137fad21/docs/k3/B300-PLAN.md) and [Spark evidence](https://github.com/TheTom/atlas/blob/32ebc80d5fc2dc0fdce37e9a9793542a137fad21/docs/k3/evidence/spark-prep-20260919/README.md).

## Implementation

- Distinct B300 SM103 target, device/registry recognition and owned kernel sources. The large source snapshot prevents cross-hardware coupling; original hashes and later changes are documented.
- K3 serving graph/loader, packed E2M1/E8M0 TP slicing before allocation, complete tensor inventory/geometry validation, strict padded `A_log` handling and explicit allocation ownership. EP>1 remains refused.
- Fix exact quant target/kernel resolution, chunked prefill, vocabulary-sized TP receive buffers and ordered worker teardown. **Shared review surfaces:** communication capacity/diagnostics, model lifecycle, weight-store ownership and serving quant selection.
- Pinned/resumable checkpoint staging with integrity/disk checks and progress receipts; official-header audit; strict generation/comparison probes; single-host owned-rank controller; collective-log checker; small packed-fixture builder; ARM64/x86_64-aware container recipe.

## Measured preparation

- Small packed fixture: **16 TP1/TP2 cases match exactly**, including measured prompt lengths around 32/64/128-token boundaries and sequential requests. **17,665 collective submissions per rank match.** This is submission agreement plus independent completed-generation evidence, not universal deadlock proof.
- Production TP8 expert dimensions pass independent CUDA/CPU arithmetic checks. KDA/MLA GPU tests cover nonzero history, state restoration and deliberately bad controls.
[...]

PR #1158, opened 2026-09-19
docs(b200): the first hour on a rented B200 for DeepSeek-V4.1 Flash

rrstesiak:
Draft. One file, `docs/perf/B200_FIRST_HOUR_2026-09.md`: the exact first hour on a rented B200 for DeepSeek-V4.1 Flash, written on one DGX Spark before any B200 is rented. **Nothing in it has run on a B200.** Every number is labelled as a GB10 measurement (`docs/perf/DS41_DECODE_RETUNE_2026-09.md`, #1156 S1/S2), a compiler's output (`kernels/b200/deepseek-v4-flash/PTXAS_RESOURCES.md`) or arithmetic with the operands shown.

**What it says, in order**
1. Competing numbers first: Turney's EXL3 two-Spark 30-32 tok/s (engram dummied, DSpark 54% accept), this engine 19.19 / 21.87 on one Spark with 6/6 texts byte-identical, llama.cpp CPU 2.3-2.5 on one GB10; no llama.cpp or vLLM number for V4.1 on B200 was found, and the doc says so rather than inventing one.
2. The arithmetic: 6.09 GB a token at 7.7-8 TB/s = 0.76-0.79 ms, so the step is the 1,905 launches and 186 host waits (9-15 ms eager = 67-107 tok/s; a design estimate of 4.5-6.5 ms once the whole-step graph lands). One B200 with a 150 GiB device arena holds 82% of the experts against the standard's 8,418 / 8,005-pair working sets: zero misses after r1, by arithmetic on the Spark traces. Two B200s buy residency for any prompt, not the byte floor.
3. Provider and node: self-serve single B200 (Lambda / Nebius / Vultr list prices as read 09-16), CUDA 13.0 (the repo pins the toolkit, no driver; R580+ for the toolkit), libnccl2 >= 2.28, persistent NVMe. The seven Q2_K shards: 264,521,661,674 bytes, revision `58d8ac86`, sha256 per shard, transfer arithmetic (44 min at 100 MB/s, 8.8 at 500, 4.4 at 1 GB/s), and the rule from Turney's `docs/k3/B300-PLAN.md` (#1150): stage on provider storage before the GPU clock. Cost: 4 h on one B200 = $27-34.
4. The hour, minute by minute, with the commands: admission, the CI binary from `datacenter-binaries.yml` (or the `AVAROK_TARGET_HW=b200 AVAROK_TARGET_MODEL=deepseek-v4.1-flash AVAROK_TARGET_QUANT=nvfp4` build), the sm_100a gate (expect 199/199 as S1), `--check-kernels`, the serve, the known-answer smoke, the standard warm and cold with the **oracle clause** (the Spark's six texts are the byte-identity reference only if the B200 kernels reproduce the bits; otherwise a B200 oracle is established on the first run and every later number says so), nsys launch / collect / stats.
5. What to profile first and what each outcome means: launch gaps => the graph; kernel time on sub-wave grids => the 148-SM retune of #1141; blocked waits => device routing must be on; misses => arena size.
6. What in #1140 / #1141 / #1142 the day's numbers supersede, item by item. Those PRs are not edited.

**Assumes** the head of #1156 (`ds41-b200-prep`): `main` today has `kernels/b200/deepseek-v4-flash/` but not `kernels/b200/deepseek-v4.1-flash/`, which arrives with #1146, so a b200 build of `main` for this target does not resolve yet. Base `main` because the doc stands alone.

Tracking: #1059 under #970. Companion to #1156 (S4 of its plan).


tbraun96:

Verification: this PR's head `f38def77c6c98f8a7c45ca9493c371fc441e0695` is the second parent of composition merge `332101028ce01893eaf384617c59eebcbdfc3ca5`, and every one of the 133 files the composition contributed is byte-identical on `main` today. Concrete artifact: `docs/perf/B200_FIRST_HOUR_2026-09.md (blob 0482c1c4a1, identical on main)`.

PR #1160, opened 2026-09-20
perf(ds41): decode retune phase 5, the last of one Spark

rrstesiak:
DeepSeek-V4.1 Flash Q2_K on ONE DGX Spark, single stream, greedy, the decode step. Competing numbers first: Turney's EXL3 build on TWO Sparks decodes at ~30-32 tok/s; this line on one Spark stood at 19.19 / 21.87 tok/s (MinHeap / Volvo medians, PR #1156 S2) with every output text byte-identical to the phase 1 oracle. Phase 5 is the last set of one-Spark levers before the B200 program (#1140-#1142): the miss path, the engram projection, the expert GEMV's latency chain.

Chained after #1156 (branch `ds41-b200-prep`, base 38c4c2097); the diff against main includes that chain until it lands. Draft until the phase closes.

**Standard** (unchanged): chat completions, 300 tokens, temperature 0, one serve, MinHeap x3 then Volvo x3, medians of three, recipe env `ATLAS_DS41_EXPERT_CACHE_GIB=100 ATLAS_DS41_READER_THREADS=16 ATLAS_DS41_DEVICE_ROUTE=1`, clean box, every text compared byte for byte with `~/dflash-logs/ds41_retune_oracle_*`. A lever is kept only on 6/6 identical texts.

**Tonight's baseline (p5base, 38c4c2097 rebuilt and re-measured)**: MinHeap 18.17 (10.58 / 18.17 / 18.77), Volvo 20.60 (11.92 / 20.60 / 20.70), 6/6. The S2 binary re-run under the same env at the same hour gave 18.18 / 21.56: the box runs about a token a second under the afternoon; the phase's deltas are against p5base.

Levers, one comment each below: L1 the miss path (O_DIRECT + the victim window), L2 the engram `wkv` projection off its raw Q2_K blocks, L3 the expert GEMV latency chain, L4 if time.

PR #1161, opened 2026-09-20
perf(ds41): decode retune phase 6, the segment graphs done right and the miss path

rrstesiak:
DeepSeek-V4.1 Flash Q2_K on ONE DGX Spark, single stream, greedy, the decode step. Competing numbers first: Turney's EXL3 build on TWO Sparks decodes at ~30-32 tok/s; this line on one Spark stands at 20.50 / 22.99 tok/s (MinHeap / Volvo medians, PR #1160 phase 5) with every output text byte-identical to the phase 1 oracle. Phase 6 is the last set of one-Spark levers: the segment graphs done right (the design the phase 4 module doc wrote down: per-layer snapshots inside the graph, the capture token eager, a miss restores one layer and runs eagerly from there), the miss path with the resident experts launched first, eviction toward Belady, the sparse attention and the projection pair kernel, and whatever a fresh profile shows above a millisecond with a byte-identical argument.

Chained after #1160 (branch `ds41-decode-retune-5`, base ae9a1dea7); the diff against main includes that chain until it lands. Draft until the phase closes.

**Standard** (unchanged): chat completions, 300 tokens, temperature 0, one serve, MinHeap x3 then Volvo x3, medians of three, recipe env `ATLAS_DS41_EXPERT_CACHE_GIB=100 ATLAS_DS41_READER_THREADS=16 ATLAS_DS41_DEVICE_ROUTE=1` (phase 5's direct reads, 8% victim window and engram Q2_K are default on), clean box, every text compared byte for byte with `~/dflash-logs/ds41_retune_oracle_*`. A lever is kept only on 6/6 identical texts.

**Tonight's baseline (p6base, ae9a1dea7 rebuilt and re-measured)**: MinHeap 20.49 (13.33 / 20.49 / 20.92), Volvo 22.93 (14.98 / 22.93 / 22.96), 6/6. The same as the phase 5 close (20.50 / 22.99): the box is where it was; the phase's deltas are against p6base.

Levers, one comment each below: M1 the segment graphs with per-layer snapshots (`ATLAS_DS41_STEP_GRAPH=1`, opt-in until byte-identical and faster on Volvo r2/r3), M2 the miss path hit-first, M3 eviction toward Belady, M4 sparse attention and the projection pair kernel, M5 anything >= 1 ms with a byte-identical argument.


rrstesiak:
## M1, the segment graphs done right: BYTE-IDENTICAL, opt-in, not faster on the Spark (d21e5501f)

**What was built** (`step_seg.rs`, `step_seg_run.rs`, the new `step_ffn.rs`; `ATLAS_DS41_STEP_GRAPH=1`, needs `ATLAS_DS41_DEVICE_ROUTE=1`). The design the phase 4 module doc wrote down: a segment's graph carries a snapshot before every layer's ffn (streams, the attention output, `pre_a`, the attention site's `post_s` / `comb_s`: five device copies a layer inside the graph), and a segment is launched only at its OWNER's step, never at its close. The owner reads its layers' routing headers once after the launch; the first layer k that flagged an absent expert has real picks (its input was right), so its misses are fetched, its snapshot restored, layer k runs from its ffn on with the eager step's own code (`ffn_eager`, the ffn half of `step_eager` moved into `step_ffn.rs`), the layers after it to the segment's end run the whole eager step, and layers s..k-1 stand. The CAPTURE token records the bodies on a second stream (the capture is relaxed) while every layer runs eagerly on the compute stream, so the token that captures is right by construction and the first replay is the next token. A new sequence brings new window rings, which the attention bodies bake: every step notes its layer's ring and a change recaptures. The recorded shared expert forks onto a captured side stream (its own events, never the eager path's), joined before the accumulate, so the replay overlaps it with the router and the selection as the eager step does. A failed capture turns the mode off for the process and the token in flight finishes eagerly.

**Four bugs on the way, each found by a trace, none in the kernels**: (1) the snapshot lacked the attention site's `post_s` / `comb_s` (overwritten by every later site, read by the ffn's `hc_post`); (2) the `step_eager` split left the attention's `hc_post` in the attention half; (3) the segment mode (S3 and this) checked only hidden / streams / normed for staleness while the per-layer graph mode also compared the window ring: request 2 of a serve read request 1's rings; (4) **`slot_table_update` raced**: the kernel writes a batch's (layer, expert, slot) triples in parallel, so a key assigned, evicted and re-assigned inside one drained batch (a prefill's worth at layer 0, or a segment's) could end at -1 for a resident expert = a false miss flag. The eager path drains per layer and rarely saw it; the segment mode drained a prefill at once and saw hundreds a request (Volvo r3, no real misses: 759 falls, all "0 fetched"). Fixed by keeping the last triple per key before the upload; policy and plumbing only, no math. That fix is on for the eager path too.

| | MinHeap median (r1 / r2 / r3) | TTFT | Volvo median (r1 / r2 / r3) | TTFT | oracle | falls r1 / r2 / r3 |
|---|---|---|---|---|---|---|
| p6base ae9a1dea7 (eager) | **20.49** (13.33 / 20.49 / 20.92) | 1153 | **22.93** (14.98 / 22.93 / 22.96) | 486 | 6/6 | |
| M1 graphs, no fork (M1e) | 18.02 (10.36 / 18.02 / 19.01) | 1156 | 21.94 (11.63 / 21.94 / 21.98) | 489 | 6/6 | 1924 / 561 / 338 ; 1728 / 2 / 0 |
| M1 graphs, re-run of the same binary (M1f, control) | 17.99 (10.36 / 17.99 / 18.97) | 1156 | 21.93 (11.61 / 21.93 / 21.98) | 489 | 6/6 | same |
| M1 graphs + shared-expert fork + 5-buffer snapshot (M1g) | 18.43 (10.48 / 18.43 / 19.46) | 1157 | 22.56 (11.80 / 22.56 / 22.61) | 489 | 6/6 | same |
| M1 binary, graphs OFF (M1def, the branch's default) | 20.39 (13.30 / 20.39 / 20.85) | 1157 | 22.90 (14.93 / 22.90 / 22.92) | 489 | 6/6 | |

Every text byte-identical in every row (six standards). The replay is real: Volvo r2 / r3 run the whole token as nine graph launches with 2 and 0 falls, and are **1.6% slower** than eager; MinHeap r2 / r3 fall 1.9 / 1.1 times a token (the real ~1.2 misses a step land anywhere in a segment; each fall wastes the graph's work after layer k and re-runs k..end eagerly) and are **10% slower**. Why the pure replay does not win on GB10: the eager step is already GPU-bound with ~1.6 ms of idle a token (the launches pipeline ahead of the GPU between its 186 syncs), while the segment protocol has nine bubbles a token (the drain for the headers, the host's parse / fetch / `prepare` including the index layer's eager attention, then the submission of a ~250-node graph) plus 160 snapshot copy nodes. The fork was worth +0.6 tok/s on Volvo; the rest of the gap is the bubbles.

**Verdict**: `ATLAS_DS41_STEP_GRAPH` stays opt-in (default off, not in the standard's env); the code lands because it is now correct and the B200 is the case for it: the same protocol with a tenth of the kernel time a segment, where the eager path's 186 host syncs a token are the wall (#1141 / #1142). Receipts `~/dflash-logs/ds41_suite_retune4_M1{e,f,g,def}_*`, serve logs `ds41_serve_retune4_M1*.log` (`ATLAS_DS41_STEP_GRAPH_LOG=1` lines: captures, falls with the fetched count, per-token counters).

rrstesiak:
## M2, the hit-first miss overlap: NOT BUILT, the ceiling is under the standard's resolution. M3, eviction toward Belady: SIMULATED, no winner, not built

**M2.** Launching a layer's resident experts before its misses land is byte-identical in principle (the expert rows are independent until `sum_rows`, which would keep the plan order; the kernel takes a pointer-table range and writes contiguous rows, so runs of hits and misses in plan order are separate launches of the same per-row math). What it can win: the routed GEMV of the four or five resident experts, ~0.3 ms, hidden under the 1.4 ms `O_DIRECT` read of the missing one. What the standard measures: MinHeap r2 / r3 have ~1.2 misses a step, so ~0.4 ms of a ~49 ms step (0.8%); Volvo r2 / r3 have no misses (0). Tonight's controls moved 0.1-0.5 tok/s between identical runs (M1e vs M1f: 18.02 vs 17.99, 21.94 vs 21.93; p6base vs M1def: 20.49 vs 20.39), so a 0.8% lever cannot be told from the box on this standard, and it needs a two-phase fetch API in `expert_arena.rs` (assign and return the misses; read and copy later) to build. The cold first requests (10-18 miss layers a step) would show it at ~3 ms a step, but r1 is not the median. Left for the B200 program, where the same structure is worth building once the miss read is the whole step.

**M3.** The phase 2 route trace replayed at 8,381 slots (misses a decode step, MinHeap r1 / r2 / r3, Volvo r1 / r2 / r3; the simulator reproduces the phase 5 numbers for LRU, the kept 8% window and Belady):

| policy | MinHeap r1 / r2 / r3 | Volvo r1 / r2 / r3 |
|---|---|---|
| LRU | 17.62 / 5.53 / 5.53 | 12.32 / 0.00 / 0.00 |
| random victim among the oldest 8% (kept) | 17.62 / **0.78 / 0.80** | 12.32 / 0.01 / 0.00 |
| BELADY (the future) | 17.62 / 0.12 / 0.12 | 7.95 / 0.00 / 0.00 |
| LFU (frequency, recency tie-break) | 17.62 / 4.81 / 4.79 | 20.83 / 9.26 / 7.61 |
| REUSE (predicted next use = last access + mean of the key's recent gaps; unseen-twice first) | 17.62 / 4.79 / 4.79 | 17.37 / 11.77 / 11.53 |
| REUSEG (as REUSE, other requests' leftovers first) | 17.62 / 17.62 / 17.62 | 12.27 / 0.00 / 0.00 |
| GEN (leftovers first, then LRU / then the 8% window) | 17.62 / 5.53 / 5.53 | 12.32 / 0.00 / 0.00 |

No history-based policy beats the 8% window; frequency and reuse-distance predictors degrade to LRU on MinHeap (the request's cycle is 37 experts larger than the cache, and inside one cycle every expert's history looks alike) and wreck Volvo, whose working set fits. The Belady gap (0.78 -> 0.12) is what only the future sees: on the standard the future is the identical r2 / r3 prompt, and a predictor keyed on the previous request's sequence would be gaming the benchmark, not a policy. Not built, no standard spent. Simulator: `cache_sim6.py` (scratchpad; the phase 2 `ds41_cache_sim.py` parsing and pinning, new policies), trace `~/dflash-logs/ds41_route_trace_c100t16.txt`.

rrstesiak:
## M4a, the sparse attention split across blocks: KEPT (9b319b43a)

`attn_v41_sparse_attn` ran one block per (token, head): 64 blocks of 256 threads on 48 SMs, ~17% of the machine, 2.1 ms a token in the L2 profile. The heavy loop is the output: for each of the `hd` dimensions a thread sums `topk` products in j order (a serial chain of ~190 FMAs behind ~190 strided loads). The kernel now takes `gridDim.z` blocks per (token, head): every block recomputes the scores and the softmax denominator (the same values in the same thread layout, so the same bits), and writes its own slice of `hd` by the very same j-ordered sum per element. Same bytes for any split by construction; only the number of blocks in flight changes. `ATLAS_DS41_SPARSE_SPLIT` (1..8) sets it; the default is now **2**.

| | MinHeap median (r1 / r2 / r3) | TTFT | Volvo median (r1 / r2 / r3) | TTFT | oracle |
|---|---|---|---|---|---|
| p6base ae9a1dea7 | 20.49 (13.33 / 20.49 / 20.92) | 1153 | 22.93 (14.98 / 22.93 / 22.96) | 486 | 6/6 |
| M1 binary, split 1 (M1def, control) | 20.39 (13.30 / 20.39 / 20.85) | 1157 | 22.90 (14.93 / 22.90 / 22.92) | 489 | 6/6 |
| split 2 | **21.59** (13.73 / 21.59 / 22.26) | 1162 | **24.07** (15.44 / 24.07 / 24.10) | 492 | 6/6 |
| split 4 | 21.31 (13.65 / 21.31 / 21.90) | 1173 | 23.95 (15.34 / 23.95 / 24.00) | 491 | 6/6 |
+5.4% / +5.0% at split 2 over p6base; split 4 a little under split 2 on both prompts (half the threads idle in the output loop at hd / 4 = 128), so 2 is the default and 8 was not run. nsys of a warm 60-token request follows with M4b.

rrstesiak:
## M4b, the `wq_a` + `wkv` pair kernel: KEPT (df5ffcb4f)

The phase 3 patch, landed as drafted: `kquant_mmvq_q2_k_pair_w` runs the two Q2_K projections of one token (`wq_a` [q_rank x dim] and `wkv` [head_dim x dim], the same normed input) in one launch off one q8_1 quantisation; `blockIdx.y` picks the tensor, the per-row math is `kq_mmvq_warp`'s, so each tensor's numbers match its own `kquant_mmvq_q2_k_w` launch bit for bit. Proved before the suite by the GPU test `kquant_pair_w_matches_two_launches_bitwise` (M = 1 and 5, unequal row counts; both outputs identical to the two launches). Two launches and one quantisation fewer a capturable layer (64 launches a token).

| | MinHeap median (r1 / r2 / r3) | TTFT | Volvo median (r1 / r2 / r3) | TTFT | oracle |
|---|---|---|---|---|---|
| M4a, split 2 (9b319b43a) | 21.59 (13.73 / 21.59 / 22.26) | 1162 | 24.07 (15.44 / 24.07 / 24.10) | 492 | 6/6 |
| M4a, split 2, re-run (control) | 21.42 (13.72 / 21.42 / 22.01) | 1161 | 24.03 (15.40 / 24.03 / 24.07) | 489 | 6/6 |
| M4b, the pair kernel on top | 21.54 (13.76 / 21.54 / 22.02) | 1159 | 24.15 (15.45 / 24.15 / 24.17) | 491 | 6/6 |

Volvo r2 / r3 (no misses: the cleanest signal) 24.15 / 24.17 against 24.07 / 24.10 and 24.03 / 24.07 on the two runs of the binary without it, +0.4%; MinHeap between the two controls. Small, byte-identical by proof, and 64 launches a token fewer; kept. The kernel source is the phase 3 draft, unchanged. Receipts `~/dflash-logs/ds41_suite_retune4_M4b_*`, control `ds41_suite_retune4_M4a2c_*`.

rrstesiak:
## Phase 6 closes: ae9a1dea7 -> df5ffcb4f

Every output text byte-identical to the phase 1 oracle in every standard run tonight (fourteen standards, 6/6 each).

| prompt | p6base median (best) | phase 6 median (best) | delta on the median |
|---|---|---|---|
| MinHeap | 20.49 (20.92) | **21.54 (22.02)** | +5.1% |
| Volvo | 22.93 (22.96) | **24.15 (24.17)** | +5.3% |

Where it came from: M4a, the sparse attention's output dimension split across two blocks per head (+5.4% / +5.0%); M4b, the `wq_a` + `wkv` pair kernel from the phase 3 draft, one launch and one quantisation instead of two each (+0.4% on Volvo, 64 launches a token fewer). Landed but off by default: M1, the segment graphs done right (byte-identical on every request now, four bugs found and fixed on the way including a slot-table race that also touched the eager path; 1.6% under eager on Volvo's pure replay and 10% under on MinHeap, whose real misses each cost a fall; built for the B200). Not built: M2 (a 0.8% ceiling on this standard), M3 (no history-based policy beats the 8% window in the simulator). nsys of a warm 60-token request on the final binary: GPU kernel time 42.3 -> 41.9 ms a token, launches 1,905 -> 1,825, `cuStreamSynchronize` 186, D2H 49; sparse attention 2.10 -> 1.50 ms, `q2_k_w` 276 -> 196 launches with the pair kernel at 0.81 ms for its 40.

The one-Spark line since the 09-17 standard (10.6 / 10.9): **+103% / +122%**, six phases, every step byte-identical. The plan doc carries the phase (`docs/perf/DS41_DECODE_RETUNE_2026-09.md`, Phase 6). Chained after #1160; ready for review as a draft once that chain lands.

tbraun96:

**Citation.** This PR's head `307cf013be` is an ancestor of the composition tip `2dd6e33ebc` (`integration/ds41-thirteen`, PR #1207), which merged to main as squash commit `b79bc1151b9a` at 2026-09-21T22:48:02Z. Containment was proved before closing: all **133** files the composition changed are byte-identical on main.

**What is *not* claimed.** This line is a stack — later layers edit the same files — so a file this PR touched may now carry a *later* layer's content rather than this one's. The claim is that these commits are inside what landed, not that every one of the 77 files still reads as this PR left it. That is normal for an intermediate layer and is why the check used here is ancestry-of-the-composition plus the composition's own containment proof, rather than per-file identity.

PR #1166, opened 2026-09-20
tools: add bounded inference and lifecycle rehearsal harness

TheTom:
A healthy server can still fail generation, leave worker processes behind, or diverge after streaming/cancellation. This adds a bounded, single-host rehearsal harness that verifies the supplied binary/device manifest, launches owned ranks, checks real completions, and tears down only its own process groups.

The companion lifecycle runner compares repeated, streamed and concurrent requests with serial baselines, tests cancellation recovery, and retains partial receipts when a request or overall deadline expires. Strict token-array payloads and an explicit persistent CUDA cache directory are supported.

- Scope: standard-library Python tooling, protocol/process tests and self-contained usage docs; 15 files. No runtime, kernel, model-loader or certification changes.
- Validation: **34 local tests passed**, including worker failure, occupied ports, wrong binary/device metadata, bounded response handling and process cleanup; Python compilation and diff checks pass.
- Provenance: extracted from #1150. Historical Spark/B200 runs used the integration binaries recorded there and are not presented as GPU validation of a new main-based binary.
- Model/device readiness is a prerequisite supplied by the caller. The harness does not make an unsupported model or TP layout work. The unmerged dense-GPU switch is deliberately excluded from this extraction.

**Ready for code review; landing remains subject to the requirements below.** Independent of the runtime extractions; the original integration notebook remains available for aggregate evidence. AI-authored with maintainer direction.

Artifact audit: the PR diff contains no rental/run logs or benchmark-result files. Reusable tests, fixtures, source provenance and reproduction tools are retained. This is review readiness, not hardware certification or merge approval.


TheTom:
RST found a false green in trusted-baseline matching: a receipt from a different model/prompt/token budget could pass when case IDs and output happened to match. Fixed in `ef63898d651903405dae96b65107de4dbc2072ae` by binding the receipt to its schema, model and exact cases before the run. All 35 Python tooling tests pass; the original mismatched-identity control is rejected. These use deterministic HTTP fixtures, not a new model/GPU run. Raw test output stays outside the source diff.

PR #1182, opened 2026-09-20
k3: add rank-local checkpoint admission and weight upload

TheTom:
**Ready for code review; dependencies and required checks remain before landing.** Extracted from #1150; stacked on shared K3 weight planning #1179 (`extract/k3-weight-plan`). This provides the runtime loader API; model/serving selection is a dependent PR.

The rank-aware safetensors loader inventories and validates the checkpoint before GPU allocation, computes rank-local byte slices, admits the upload against explicit staging/binding/reserve requirements, then uploads only local bytes. It validates packed-weight/scale pairs, required names, geometry, dtypes and index consistency; rejects duplicate/oversized indexes and paths escaping the checkpoint root; and releases allocated tensors if loading fails. Successful stores record their partition `(rank, world)` to prevent downstream double slicing.

This is a source-identical extraction of the `spark-runtime` weights changes at `bba13ef3a`, with a focused explanation in `docs/k3/RANK-WEIGHT-LOADER.md`. The small `DeferredTensor` and release-module moves retain existing behavior while keeping `weights.rs` below 500 lines. `tempfile` is an existing workspace dependency, added here only for real temporary checkpoint tests. Unrelated runtime Metal tests are excluded.

### Validation

- Fresh host K3 loader/index tests: **9 passed**, zero ignored.
- Broader host weight tests: **108 passed**, 10 existing ignored, zero failed. Includes existing teardown coverage.
- Default CUDA-feature host Clippy (`AVAROK_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy -p spark-runtime --tests`) passed.
- Formatting, scoped typos, diff whitespace and changed Rust SPDX headers passed.
- The separate no-default-features Clippy invocation exposes four existing `too_many_arguments` errors in unchanged cuBLASLt/CUTLASS stub files. Those are not counted as passing or changed here.

Reproduce host tests with `AVAROK_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo test -p spark-runtime --no-default-features --lib weights::`.

Tests use real safetensors files and a recording mock GPU to verify bytes, admission and allocation cleanup. **No new GPU execution, Docker rebuild, full-checkpoint load, multi-GPU inference or performance certification is claimed.** Admission still depends on a caller-provided engine/KV/runtime reserve. The checkpoint directory must remain immutable during loading; canonical path checks do not sandbox concurrent file replacement.

[...]

PR #1179, opened 2026-09-20
k3: share packed tensor-parallel planning and rank memory audit

TheTom:
Share one K3 tensor-parallel storage plan between core memory accounting and model binding. The extracted planner validates dense/packed MXFP4 byte geometry, rank slices, replicated shapes and required tensor names before allocation.

Stacked on #1176 through the identical `extract/k3-host-reference` base. The own diff adds the planner, delegates the spark-model adapter, and includes pinned header accounting plus a small synthetic packed-fixture generator. Runtime allocation and serving follow in separate dependent PRs. No bulk header dumps or rental logs.

Validation: combined core stack has 107 passing K3 tests and 11 ignored real-checkpoint tests. Five Python checks pass, including incorrect Range responses, exact packed levels and rejection of nonfinite/invalid geometry. Core clippy (tests/examples), formatting and typos pass. The rank-memory example ran locally in config-estimate mode. No full checkpoint, new GPU run, or NCCL claim. Weight-payload accounting excludes context/KV/scratch/allocator reserves; see docs/k3/WEIGHT-PLAN.md.

Ready for code review. Parent landing, review and required checks remain before merge. Retarget to main after parent landing without importing the original integration notebook.

Artifact audit: the PR diff contains no rental/run logs or benchmark-result files. Reusable tests, fixtures, source provenance and reproduction tools are retained. This is review readiness, not hardware certification or merge approval.

PR #1178, opened 2026-09-20
server: guard unsupported K3 XTML chat and preserve raw markers

TheTom:
## What and why
Recognize official Kimi K3's XTML tokenizer contract and refuse chat requests until the segmented encoder and output/tool parsers exist. Raw completions remain available with independently prepared token IDs. This prevents silently substituting generic ChatML or the twin's Jinja template. Literal `<think>` text in raw XTML output is preserved rather than processed as Qwen reasoning markup.

Detection uses `tiktoken.model` or `TikTokenTokenizer` metadata for `kimi_k3`; the ordinary small twin keeps its existing template path. Both chat entry points reject unsupported XTML before template fallback, including tools and override templates.

## Scope and validation
Eight-file, main-based extraction from #1150: tokenizer guard, completion post-processing, five regression tests and an offline Rust/tokenizer oracle example. No model architecture, kernels, downloaded tokenizer assets, credentials or benchmark logs are imported. #1174's general plain-whitespace fix remains separate.

Formatting and diff checks pass. Three extracted tokenizer regression tests now execute and pass with `AVAROK_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo test -p spark-server --no-default-features --features metal --lib tokenizer::tests::kimi_k3`. They cover both chat entry points, official/twin detection, tools and template-override refusal. The two completion regressions also execute and pass with the same environment and feature selection using `cargo test -p spark-server --no-default-features --features metal --bin spark api::completions_exec::tests`. All five new regression tests have now run successfully, covering raw-marker preservation plus stop filtering and legacy behavior. Default CUDA-feature builds on this Mac remain blocked by existing Linux-only storage libc calls. The Metal-feature host test run does not validate CUDA execution or GPU certification. The oracle example requires independently generated `tokenizer-oracle.json` and serving assets and checks model EOS separately from the named EOS token.

This PR does NOT implement full official chat, reasoning or tool support, claim full K3 inference readiness, or carry GPU certification. Required Linux tests and normal project checks remain before landing.

## Authorship
AI-authored extraction and regression tests, directed by TheTom; human review pending.

Artifact audit: the PR diff contains no rental/run logs or benchmark-result files. Reusable tests, fixtures, source provenance and reproduction tools are retained. This is review readiness, not hardware certification or merge approval.

PR #1177, opened 2026-09-20
kernels: add owned B200 Kimi K3 model target

TheTom:
Adds the Kimi K3 model target to the existing B200 hardware scaffold from merged #1045. This is a focused extraction from #1150 at `bba13ef3a`; it does not repeat the broad open #895 work or the GB10 alias registration in #1165.

The target owns its KDA, MLA, E8M0 expert GEMM and FP32-I/O dense helper sources. BF16/MXFP4/NVFP4 registry aliases remain inside B200, while the already-established B200 common mirror is unchanged. Inherited-target tests now account for the additional explicitly owned K3 model. The dense helper is included as a kernel prerequisite for the separate runtime integration; this PR does not connect that runtime path.

Validation (CPU only for this extraction): B200 registration test **1 passed**, inherited target suite **16 passed**, targeted Clippy, changed Rust formatting, kernel shadow structure and cross-hardware reach checks pass. Kernel bodies match the frozen integration source. Historical B200 compiler/numerical observations are documented with their limited scope in the target README; no GPU run or certification was performed for this extracted head.

**Ready for code review; landing remains subject to the requirements below.** This is build-input registration, not a claim that this isolated PR alone supports full K3 serving, TP4 deployment or production throughput. No rental was restarted. AI-assisted extraction.

Cross-hardware reach: all **18** changed kernel paths resolve only to `b200`; every row is `OK`, no `DECLARED`/`CHKI` or structural findings; final verdict `OK`.

Hardware: b200
CHKI-Verdict: benign — static reach confined to B200; owned model sources and same-target aliases only.

Artifact audit: the PR diff contains no rental/run logs or benchmark-result files. Reusable tests, fixtures, source provenance and reproduction tools are retained. This is review readiness, not hardware certification or merge approval.

PR #1183, opened 2026-09-20
k3: wire baseline host serving to rank-local weights

TheTom:
## What and why
Connect K3 rank-local weights to baseline model assembly and serving. Includes the BF16/FP32 binder, opt-in packed E8M0 expert binding, shared TP-plan adapter, sequence-owned host state and KDA/MLA callbacks, layer factory dispatch, bounded per-token prefill, rank-aware safetensors selection and exact MXFP4 target selection. Converted engine weights are adopted by the existing derived-weight owner.

This is the serving foundation, not the optional GPU optimization. It retains per-call KDA state transfers and uses no dense GPU callback. Resident recurrence and dense/shared MLP execution are the next stack slice. No checkpoint, logs or benchmark records are included.

## Stack
Base: `extract/k3-runtime-weights` at 43f7dba41, the clean runtime #1182 on planner #1179 and host reference #1176. Review only this PR's delta; retarget after ancestors land. This replaces the corresponding serving subset of #1150/#1053, not their entire branches.

Separate integration prerequisites remain #1165 (GB10 kernel registration) or #1177 (B200 target), #1171 (idle/bounded communication), #1173 (vocab receive capacity), #1174 (whitespace), #1175 (worker cleanup), and #1178 (official XTML guard). #1163's expert batching remains a separate optimization. None is silently duplicated here. This PR must not be treated as independently ready to merge into main.

## Checks and limits
- Core K3: 107 tests pass, 11 fixture-dependent tests ignored.
- `AVAROK_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo check -p spark-model --no-default-features --features metal --lib` passes.
- Equivalent server `--lib` and `--bin spark` checks pass, including serving wiring. Formatting and diff checks pass.
- Full model tests under Metal are blocked by existing ungated CUDA-only tests; Metal clippy is blocked by existing runtime Arc/non-Send diagnostics. Default CUDA builds on this Mac are blocked by existing Linux-only storage libc calls. No lint suppression or portability workaround added.

The new extracted head is NOT GPU tested/certified. Historical twin evidence belongs to #1150's combined source/binary, not this reconstructed stack. Linux/CUDA build, numerical and lifecycle checks against the final combined tree remain required. Full official K3, B300, TP8, multi-host and XTML chat/tools are unvalidated. No rental or certification campaign started.

## Authorship
AI-assisted extraction directed by TheTom; human review pending.

Artifact audit: the PR diff contains no rental/run logs or benchmark-result files. Reusable tests, fixtures, source provenance and reproduction tools are retained. This is review readiness, not hardware certification or merge approval.


TheTom:
Corrected the extraction test boundary: kernel source discovery assertions now live with the aliases in #1165, where both tests pass (including the exact E8M0 entry declaration and KDA alias checks). The serving PR no longer assumes those separately extracted files are already in its base. #1165 remains a serving merge prerequisite; this change does not remove runtime fail-closed kernel lookup or claim standalone CUDA readiness.

Head: `a91e64c6e7d8effde745dcdd0d9c18a9db85e379`.

tbraun96:
Heads-up for when this line rebases: **the K3 stack forked 12 commits before
current main and carries a test assertion main has superseded.**

Composing #1179/#1180/#1182/#1183/#1184/#1187 onto `main` (#1202) fails exactly
one test:

```
kimi_k3::host_decode_kda::latent_moe_packed_launches_three_e8m0_gemms
  assertion `left == right` failed: w1, w3, and w2 each launch grouped GEMM
  left: 2   right: 3
```

All six PRs are green individually, so this is composition-only. The cause is
not any member — it is **#1186**, which landed on main after this line forked
and **batched the gate and up projections into one grouped GEMM**:
`crates/spark-model/src/kimi_k3/moe_cuda.rs:120-137` concatenates the w1 and w3
pointer tables, runs a single `gemm_rows` at `2 * m` rows, and splits the result
on the host. Two launches, not three.

No K3 member touches `moe_cuda.rs` (`git diff 62284b3fbd <head> -- …/moe_cuda.rs`
is empty for all six), and `host_decode_kda.rs` is added wholly by this line —
two disjoint files, which is exactly why git merged them without a conflict and
produced a semantic one instead.

Worth noting main did not simply change the count: #1186 also shipped
`crates/spark-model/tests/k3_moe_batching_cuda_oracle.rs`, which compares the
batched production schedule against separate gate/up/down launches and proves
they agree numerically. That oracle is the authority on equivalence; the test
here only pins how many launches reach the backend.

Fixed on the integration branch in `de1dc2abbe` — test-only, no production
change: assert two, and rename to `latent_moe_packed_launches_two_e8m0_gemms`.
It belongs in **this PR** (the lowest layer that contains `host_decode_kda.rs`,
introduced by `de9a92c3fb`); #1184 and #1187 inherit it. Carry it across
whenever you rebase, or leave it and #1202 will bring it in.

tbraun96:
Certified landed via #1202 (`integration/k3-six`, squash-merged to `main` at `dde2381c7b6512a249a1dd17b2d1e7bdcacdd998`, 2026-09-21).

Verification: this PR's head `a91e64c6e7` was merged into the integration branch as commit `6821cc764f` (real 2-parent merge, second parent = this PR's actual head sha). The integration branch's final tree is byte-identical to current `main` for every K3-related path. Concrete artifact: `crates/spark-server/src/main_modules/serve_quant.rs` (added by this PR, `pub(crate) fn canonicalize_model_quant`) is present on `main` with an identical git blob hash to this PR's head. The one commit this PR's head has that the stack tip (#1187) lacked — `a91e64c6e7` 'test(k3): keep kernel registration checks with aliases', which deletes `k3_discovered_gemm_declares_e8m0_ptrtable` from `crates/spark-model/src/weight_loader/kimi_k3/mxfp4.rs` — is also reflected on `main`: that test is absent from `main`'s copy of the file.

Closing as landed.

PR #1185, opened 2026-09-20
perf(ds41): decode retune phase 7, the arena edge and the quant launches

rrstesiak:
DeepSeek-V4.1 Flash Q2_K on ONE DGX Spark, single stream, greedy, the decode step. Competing numbers first: Turney's EXL3 build on TWO Sparks decodes at ~30-32 tok/s; this line on one Spark stands at 21.54 / 24.15 tok/s (MinHeap / Volvo medians, PR #1161 phase 6; best 22.26 / 24.17) with every output text byte-identical to the phase 1 oracle. Phase 7 is two small levers: the arena working-set edge on MinHeap (37 slots = 452 MiB short at 100 GiB by the phase 2 count; the box swapped at 102) and the quantisation launches (~714 `kquant_q8_1_rows_bf16` launches a token folded to a few per layer, byte-identical by construction).

Chained after #1161 (branch `ds41-decode-retune-6`, base 307cf013b); the diff against main includes that chain until it lands. Draft until the phase closes.

**Standard** (unchanged): chat completions, 300 tokens, temperature 0, one serve, MinHeap x3 then Volvo x3, medians of three, recipe env `ATLAS_DS41_EXPERT_CACHE_GIB=100 ATLAS_DS41_READER_THREADS=16 ATLAS_DS41_DEVICE_ROUTE=1`, clean box, every text compared byte for byte with `~/dflash-logs/ds41_retune_oracle_*`. A lever is kept only on 6/6 identical texts; a recipe changes only if it is safe (no swap) on both suites.

**Tonight's baseline (p7base = the #1161 head, the phase 6 final binary re-run with the route trace on)**: MinHeap 21.34 (13.69 / 21.34 / 21.87), Volvo 23.87 (15.40 / 23.87 / 23.90), 6/6; the trace's flush per layer costs about 1% against 21.54 / 24.15, so the A1 rows below are compared with each other under the same trace.

Levers, one comment each below: A1 the arena edge (100 / 100.5 / 101 GiB with MemAvailable, PSI, swap and the LRU misses per request watched), A2 the quantisation launches.


rrstesiak:
**A1, the arena edge on MinHeap: measured, the recipe stays at 100 GiB.** This is the working-set edge of ONE prompt (MinHeap's 8,418 distinct (layer, expert) pairs against 8,381 slots); Volvo's 8,005 fit at 100 and its r2 / r3 miss nothing at any size.

Fractional GiB first: `ATLAS_DS41_EXPERT_CACHE_GIB=100.5` parsed as an integer, failed, and fell back to 88 GiB without a word; it now parses as f64 (5d029914c). Three standards with the same binary path, the route trace on (one line per layer fetch: the misses, the evictions) and a 1 Hz memory watch beside the serve (MemAvailable, swap used, `pswpout` / `pswpin`, PSI memory, `~/code/atlas-notes/bin/ds41_memwatch.sh`):

| arena | slots | MinHeap median (r1 / r2 / r3) | Volvo median (r1 / r2 / r3) | MinHeap misses a step r2 / r3 | MinHeap r1 TTFT | swap during the serve | PSI memory peak | texts |
|---|---|---|---|---|---|---|---|---|
| 100 GiB | 8,381 | 21.34 (13.69 / 21.34 / 21.87) | 23.87 (15.40 / 23.87 / 23.90) | 2.20 / 1.24 | 4,481 ms | none (+3 pages) | 0.00 | 6/6 |
| 100.5 GiB | 8,423 | 21.57 (13.68 / 21.57 / 22.12) | 23.84 (15.39 / 23.84 / 23.90) | 1.77 / 0.72 | 4,474 ms | 311 MiB out in the second of the arena allocation, 13 MiB more in r2, 79 MiB back in | 3.02 | 6/6 |
| 101 GiB | 8,465 | 21.32 (11.05 / 21.32 / 22.49) | 23.76 (15.38 / 23.87 / 23.76) | 1.59 / 0.46 | 84,743 ms | 4.6 GB out, 10.4 GB in, through the load and r1's prefill | 98.9 | 6/6 |

Two findings. First, the slot count is not what MinHeap's residual is made of: 42 and 84 more slots take r3 from 1.24 to 0.72 to 0.46 misses a step, but r2, which is the median, moves 2.20 to 1.77 to 1.59, because r2's prefill re-misses 386 / 357 keys that r1's decode evicted (the LRU's request cycle, the same thing the phase 6 Belady simulation found); even at 8,465 slots r1 evicts 228 experts. The whole edge is worth about 0.4 ms of a 46 ms step on the median. Second, `MemAvailable` of ~8.4 GiB beside a 100 GiB arena is not free margin: 100.5 already swaps 311 MiB at the allocation (idle pages of other processes, harmless to the numbers this time, PSI 3%) and 101 thrashes for the whole load and the first prefill (TTFT 84.7 s, the decode itself then normal). The real headroom above 100 GiB is under 512 MiB, so by the rule (no swap growth on either suite) the recipe stays at `ATLAS_DS41_EXPERT_CACHE_GIB=100`.

The memory map at 100 GiB, for the record (serve log + ledger): expert arena 100 GiB device memory + 196 MiB page-locked staging ring (16 slots), off the ledger; GPU ledger 7.29 GB = raw Q2_K / Q3_K residents (attention projections, shared experts) 2,591 MB + Q2_K-to-bf16 expansions 2,056 MB (the token embedding 129,280 x 5,120 bf16 = 1,324 MB, Q2_K in the file; the engram `wkv` bf16 2 x 315 MB, the prefill path) + KV 1,286 MB (2,057 blocks, clamped to 8 sequences x 4,096) + the Q6_K head 517 MB + ~530 MB of buffers; buffer arena 290 MB; inference reserve 512 MB; the serve's host RSS ~0.9 GiB; the other processes ~0.8 GiB. The one resident that could go by construction is the bf16 embedding table: a gather that dequantises the token's row from the raw Q2_K blocks with the same per-element math is byte-identical and frees 1.1 GB; the engram bf16 `wkv` is the prefill GEMM's operand and cannot be replaced byte-identically. Neither is built here: 1.1 GB would make 100.5 safe (0.72 misses a step on r3, the median unchanged within noise) and would not make 101 safe (4.6 GB swapped), so the gain on this box is inside the standard's noise; on the B200 the same 1.1 GB is real memory. Receipts: `~/dflash-logs/ds41_suite_retune4_{p7base100,A1-100p5,A1-101}_*`, `ds41_trace_retune7_*`, `ds41_memwatch_retune7_*`.

rrstesiak:
**A2, the quantisation launches folded (5c6d392ca): KEPT, byte-identical, 1,825 -> 1,666 launches a token, level on GB10.**

The count first: the M4b profile has 395 `kquant_q8_1_rows_bf16` launches a token (the 714 of the phase 3 profile predate the pair and groups kernels), about ten a layer: four in the attention (the pair's input `x`, `q_b`'s input `qr`, `wo_a`'s input `o_rot`, `wo_b`'s input `og`, each a row the previous kernel just produced, so nothing to fold there without touching the producers) and six in the MoE: the routed experts quantised the token's row, then the shared expert quantised the SAME row twice more (once for `w1`, once for `w3`, into its own side-stream scratch), and both the routed and the shared SwiGLU outputs went through a separate q8_1 launch. Two folds, both the same bytes by construction:

1. **The token's rows once** (`ffn_input_q8_m1`): the eager step quantises the ffn input into `a_q8` on the main stream before it records the side fork, and the routed GEMV and both shared projections read those rows (`x_pre` / `ready` at the readers; `sa_q8` is gone). Three launches of the same kernel on the same bytes into three buffers became one: -80 a token.
2. **The SwiGLU writes its q8_1 rows** (`kquant_swiglu_q8_1_rows_bf16`): one warp per 32-value block computes `moe_v41_swiglu`'s expression, rounds to bf16, stores `h`, and quantises the 32 bf16 values it just wrote with the rows kernel's own `warp_reduce_max` / `warp_reduce_sum` and rounding. Routed (`ne` rows, the routing weight folded in) and shared (1 row): -80 a token. Proved bit for bit against the two launches before the suite by `kquant_swiglu_q8_1_rows_matches_the_two_launches_bitwise` (rows 1 / 6 / 5, limit 0 and 1.5, with and without the weights: `h` and the q8_1 blocks identical).

`shared_expert_body` moved out of `forward.rs` (which sat at the 500-line cap) into `moe_v41/shared.rs`; the m > 8 MMQ arms and the m1-loop diagnostic are untouched.

| binary | MinHeap median (r1 / r2 / r3) | Volvo median (r1 / r2 / r3) | texts |
|---|---|---|---|
| p7base control, run 1 | 21.49 (13.75 / 21.49 / 22.15) | 24.13 (15.42 / 24.13 / 24.19) | 6/6 |
| p7base control, run 2 | 21.51 (13.69 / 21.51 / 22.04) | 24.14 (15.44 / 24.14 / 24.18) | 6/6 |
| A2, run 1 | 21.30 (13.62 / 21.30 / 21.77) | 23.86 (15.32 / 23.86 / 23.88) | 6/6 |
| A2, run 2 | 21.49 (13.71 / 21.49 / 22.10) | 24.17 (15.47 / 24.17 / 24.22) | 6/6 |
| A2, run 3 | 21.48 (13.75 / 21.48 / 22.05) | 24.17 (15.49 / 24.17 / 24.20) | 6/6 |

Five standards, 30/30 texts byte-identical; A2's first run was a low draw of the box (its r2 TTFT was 1,484 ms against 1,150-1,165 everywhere else), runs 2 and 3 sit on the controls. nsys of a warm 60-token request (`ds41_nsys_decode_A2`, `ATLAS_DS41_DEVICE_ROUTE=1`): 1,666 launches a token (M4b 1,825), GPU kernel time 42.1 ms (41.9): the quantisation family 0.744 + 0.233 (swiglu) = 0.977 ms -> 0.352 (fused) + 0.296 (the 237 launches left) + 0.022 = 0.670 ms; `kquant_mmvq_q2_k_w` +0.38 ms at the same 195.7 launches, which is the shared `w1` / `w3` GEMVs now starting the moment the fork is recorded and stretching under the router GEMV they overlap (elongated, not serialised: the wall says level); host `cuLaunchKernel` time 4.37 -> 3.90 ms a token; 186 syncs and 49 D2H a token unchanged, so the step is still what phase 6 said it was: the bytes the kernels read and the host's waits. Kept for the launch count, which is the B200's metric (the same step there runs its 1,666 launches under 8 TB/s, where 159 launch slots are a larger share of a shorter step).

rrstesiak:
**Phase 7 closing line.** 307cf013b -> 5c6d392ca, nine standards tonight, 54/54 texts byte-identical to the phase 1 oracle.

| | MinHeap median | MinHeap best | Volvo median | Volvo best |
|---|---|---|---|---|
| p7base (#1161 head), two untraced controls | 21.49 / 21.51 | 22.15 | 24.13 / 24.14 | 24.19 |
| phase 7 final (A2), three runs | 21.30 / 21.49 / 21.48 | 22.10 | 23.86 / 24.17 / 24.17 | 24.22 |

Per prompt at the final binary: MinHeap r1 13.62-13.75 (cold, the working set streaming in), r2 21.30-21.49, r3 21.77-22.10; Volvo r1 15.32-15.49, r2 23.86-24.17, r3 23.88-24.22. Level on GB10; what changed is the count under the step: 1,825 -> 1,666 launches a token, the same 186 syncs and 49 D2H.

A1 measured the arena edge instead of moving it: MinHeap's residual misses are the LRU's request cycle (r2 2.20 -> 1.77 -> 1.59 a step across 8,381 / 8,423 / 8,465 slots), and the box has under 512 MiB of real headroom above the 100 GiB arena (100.5 swaps 311 MiB at the allocation, 101 thrashes: TTFT 84.7 s), so `ATLAS_DS41_EXPERT_CACHE_GIB=100` stays; the env now parses fractional GiB instead of falling back to 88. The memory map is in the A1 comment; the one resident that could go by construction is the bf16 token embedding (1.3 GB, Q2_K in the file). A2 folded the quantisation launches (the token's q8_1 rows once for the routed and shared experts, the SwiGLU writing its q8_1 rows in the same launch), proved bit for bit before the suite and kept for the launch count.

Plan doc: `docs/perf/DS41_DECODE_RETUNE_2026-09_part2.md` (the first file is at its cap). Receipts `~/dflash-logs/ds41_suite_retune4_{p7base100,A1-100p5,A1-101,A2,A2-r2,A2-r3,p7base-ctl,p7base-ctl2}_*`, `ds41_trace_retune7_*`, `ds41_memwatch_retune7_*`, `ds41_nsys_decode_A2*`. Against the 09-17 standard of 10.6 / 10.9: +103% / +122%; against Turney's two-Spark EXL3 ~30-32, one Spark stands at 21.5 / 24.2.

tbraun96:
Composed into **#1207** with twelve others of the DeepSeek-V4.1 line, so the
whole line shares one certification campaign instead of thirteen. Your head
`0d9b915cb9` is merged by sha and containment is asserted, not assumed.

Two things you should know, because I changed your code.

**1. One merge conflict, resolved as a union.**
`crates/spark-server/src/tokenizer/chat_impl.rs` — main added an `official_k3`
arm ahead of the `deepseek_v4` check while you widened that check to
`deepseek_v41`. The model types are disjoint, so both arms are kept in main's
order:

```rust
let chat_encoding = if official_k3 {
    ChatEncoding::KimiK3XtmlUnsupported
} else if model_type == "deepseek_v4" || model_type == "deepseek_v41" {
```

**2. Three checks were already red on this PR, and I fixed them** (commit
`6b78de4b97` on the integration branch) rather than park thirteen PRs on lint:

- **SPDX** — `layers/deepseek_v41_layer/hc_launch.rs` and `model/lm_head_q6k.rs`
  open with `// provenance-id:` and have no `SPDX-License-Identifier` in the
  first five lines, so `check_spdx.py` refuses. Header prepended; the checker
  passes locally now.
- **clippy `manual_is_multiple_of`** — `weights/gguf/expert_direct.rs:137` and
  `:140`, `x % 512 == 0` → `x.is_multiple_of(512)`. Both inside assertions.
- **`gate::taxon::taxon_tests::every_real_target_resolves_a_nonempty_source_set`**
  — #1156 adds `kernels/{b200,hopper}/deepseek-v4.1-flash` (MODEL.toml only,
  quant derived, same shape as the `gb10` target already on main) without
  extending that test's expected list, so `walk()` returned two targets the list
  did not name. Added both in walk order. Note this cannot fail open: the loop
  after the `assert_eq` calls `sources()` on every target and refuses an empty
  set, so if those two resolve nothing, CI will say so rather than pass quietly.

None of the three were caused by the composition — they are red on this PR's own
CI, and #1158 is green on all of them. Please carry the fixes if you push again,
or leave them and #1207 brings them in.

tbraun96:
Follow-up on the composition (#1207) — a second batch of fixes, `dbdaf0c2a7`,
and **three things I deliberately did not touch** because they change what is
served and are yours to settle.

## Fixed: `min_nodes` is ranks, not hosts

Four recipe tests were red (`every_avarok_recipe_produces_a_valid_serve_config`,
`every_recipe_can_be_served_hermetically`, `all_vendored_recipes_parse`,
`the_whole_corpus_reads`) with:

```
--ep-size 2 exceeds --world-size 1
```

`recipe/mod.rs:202` is the whole mechanism — `min_nodes > 1` is what emits
`--world-size`. And the field's own doc comment at `mod.rs:28` reads *"Ranks
this recipe requires"*, then names this exact failure: the EP=2 recipes carry
`ep_size: 2` in `defaults` and `min_nodes: 2` up here. So two ranks on one
NVLink host and two ranks on two hosts carry the same value; nothing downstream
distinguishes them, and `max_nodes` is not parsed at all outside `yaml_tests`.
`min_nodes: 2` **is** the schema's spelling of "two B200 on one node" — it does
not contradict your description.

Both new EP=2 recipes had `min_nodes: 1`; the three already on main have `2`.
The panic named only `-ep2` because `all()` aborts on the first failure, so
**`-spec` was hidden behind it** and would have failed on the next run.

Leaving it at 1 was also wrong for the TUI: `lib_start.rs:114` filters
launchable donors to `min_nodes <= 1` because the dashboard spawns one process,
so it would have offered these recipes and then died on the same validator
error.

Corpus counts moved 25 → 28 and `is_avarok` 23 → 26 (three new recipes, all
`runtime: atlas`). `fetch_tests.rs` stays at **25** on purpose — those two are
`#[ignore = "needs the network"]` and assert against the live `atlas-recipes`
repo, which holds 25 until these are published there.

**No `defaults:` key was touched**, so the served topology is exactly what you
declared.

## Not fixed — three internal contradictions, please settle them

1. **`-ep2` tells three different stories.** Its header says "Two B200 on one
   NVLink node"; its `metadata.description` — which looks copied verbatim from
   the single-GPU sibling, `ATLAS_DS41_EXPERT_CACHE_GIB: "88"` included — still
   says "on **one** datacenter Blackwell GPU"; and the commit title says "the
   **two-node** B200 recipe".
2. **`-spec` is the speculative-decoding recipe and carries
   `speculative: false`**, with a description ending "No speculation: the GGUF
   carries no DSpark keys."
3. **`-spec` carries `ep_size: 2`** while its header says "One or two GPUs per
   the campaign recipes".

None of these blocks the merge, and none is something I should guess at.

tbraun96:

Verification: this PR's head `0d9b915cb9ff389e105df2a43e3f13619809b25a` is the second parent of composition merge `639d59613ded24745dc48f8165ffd72553f6b7ca`, and every one of the 133 files the composition contributed is byte-identical on `main` today. Concrete artifact: `crates/spark-model/src/layers/attn_v41/decode.rs (blob c729dc5e00, identical on main)`.

PR #1184, opened 2026-09-20
k3: retain recurrent GPU state and enable optional dense callbacks

TheTom:
## What and why
Add per-sequence resident KDA state and explicit opt-in resident dense/shared MLP execution above the clean host-serving foundation. Snapshots read authoritative GPU state; restore invalidates it; CPU escape downloads it; release and partial allocation/upload failures free owned buffers. Later-layer setup failures clear the in-progress AttnRes stream.

`K3_CUDA_DENSE=1` reuses existing FP32/BF16 matrix weights for gate/up+SiTU and down CUDA callbacks. Unset/0 retains host execution; malformed values fail. This does not move all projections, routing or the complete graph to GPU.

## Stack and scope
Base: `extract/k3-serving-foundation` at de9a92c3f, PR #1183, above runtime #1182 / planner #1179 / host reference #1176. Review this delta only; retarget after parents land. B200 target #1177 supplies the dense CUDA kernels. The foundation's separate communication, capacity, shutdown and tokenizer integration prerequisites still apply. Expert batching #1163 is separate and is not copied here.

This extraction contains the final seven runtime/module files and all four ignored GPU oracle binaries from #1150. No hardware tree, bulk logs, checkpoint weights or certification records are included. Most added lines are the independent oracles, not serving changes.

## Validation and benchmarks
- Metal-feature `cargo check -p spark-model --no-default-features --features metal --lib` passes with `AVAROK_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000`; formatting/diff checks pass. This checks Rust interfaces, not CUDA.
- The parent has 107 core tests passing (11 fixture-dependent ignored) and production model/server library plus server binary compile checks passing. Existing ungated CUDA-only model tests and Metal backend clippy findings block a full local sweep; no suppression added.
- Historical #1150 B200 development runs exercised resident-state lifecycle and the dense, mixer and packed-expert numerical oracles. These are historical receipts, not tests/certification of this new extracted head. No new performance claim is made.
- `k3_tp4_composed_cuda` exercises official TP4 **rank-local dimensions** through two layers with synthetic tensors and reset/history checks. It is one rank/process, without NCCL, full checkpoint loading or a four-rank equivalence claim.

The new stack still needs its own Linux/CUDA build, independent numerical checks and bounded multi-rank lifecycle against the final combined tree. Full K3 weights, B300, TP8, multi-host and chat/tools remain unvalidated. No rental, merge or certification campaign started.

## Authorship
[...]

Issue #1167, opened 2026-09-20
K3: wire resident MLA cache into serving and eliminate per-token history uploads


TheTom:
Priority: **P0 for the next Spark iteration**.

The current integration serving path calls the host-KV MLA wrapper even though an on-device helper exists. Each generated token reuploads the expanded KV history. At official TP4 dimensions with 4,096 live tokens, that is an estimated **2.8125 GiB per rank per token** across the MLA layers; this is layout arithmetic, not measured full-model traffic.

Tasks:
- [ ] Give the serving layer ownership of resident MLA state and connect the existing device helper; validate actual dimensions/dtypes before dispatch.
- [ ] Define which state is authoritative through snapshot, restore, prefix reuse, reset, retirement and the explicit CPU reference path.
- [ ] Clean up partial allocation/upload failures and release state after cancellation or shutdown.
- [ ] Add transfer/allocation measurements that distinguish initial allocation, token append, snapshots and steady decode.

Acceptance:
- [ ] Compare cold generation, continuation and restored generation against the independent reference and the retained baseline with declared numeric tolerances and expected token behavior.
- [ ] Test one-token/longer histories, empty or malformed restores, cancellation then a fresh request, and repeated reset/reuse. Deliberately corrupt or clear history and require the comparison to fail.
- [ ] Show that steady token-append host-to-device traffic no longer scales with the full context length; snapshot traffic may still scale and must be reported separately.
- [ ] Pass one-Spark and available two-Spark regression checks without claiming that TP2 proves TP4/B300.

Context: draft integration #1150, architecture dependency #1053, and [frozen B200 readiness report](https://github.com/TheTom/atlas/blob/bba13ef3a/docs/k3/evidence/b200-rehearsal-20260919/TP4-READINESS.md). The integration runtime was `083260451`; these findings do not imply the same serving graph is already on `main`. Clean extractions #1163–#1166 remain separate draft reviews. Do not merge the integration notebook to obtain a dependency.

Execution: start on an available Spark using the small fixture. Check device/process ownership first; preserve Deckard and avoid the other active Kimi campaign. No new rental or certification campaign is requested. Record source/model hashes, exact inputs, expected results, observations and remaining uncertainty. Bulk traces stay outside Git; commit compact reproducible receipts.

TheTom:
Picked this back up after the Codex session hit the usage cap.

[...]

PR #1187, opened 2026-09-20
k3: keep MLA KV resident across CUDA decode tokens

TheTom:
Stacked on #1184. Refs #1167.

## What and why
CUDA MLA serving still called the host-KV wrapper, which re-uploads `[0..T]` every token. This wires the existing `MlaDeviceKv` helper into per-sequence state so decode appends one row on device.

- `K3CpuFallbackState` owns resident MLA KV the same way it owns KDA recurrence.
- Snapshots download authoritative device history. Restore releases device buffers. CPU escape downloads then frees.
- CUDA MLA requires `serve_max_seq_len` (`--max-seq-len`) to bound the allocation. Missing bound fails closed.
- Partial K/V allocation and upload failures free what was allocated.

## Validation
- Metal-feature `cargo check -p spark-model --no-default-features --features metal --lib` passes with `AVAROK_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000`.
- `spark-runtime` mock H2D counters: 1 passed.
- Known-bad oracle: host-KV wrapper H2D grows with history. Resident path H2D stays O(1) in T. Snapshot after two CUDA tokens has `seq_len = 2`. Empty restore is accepted. Invalid restore leaves state intact.
- Full `spark-model` lib tests still cannot run on macOS Metal (pre-existing ungated CUDA-only test modules). Linux/CUDA and GB10 execution are still outstanding. This does not claim TP4/B300 or full-model traffic.

Do not merge the #1150 notebook to obtain a dependency.


TheTom:
Spark2 (GB10, rustc 1.93.1, commit `84677b97f`):

```
cargo test -p spark-model --lib kimi_k3::
ok. 59 passed; 0 failed
```

These are mock-backend unit tests compiled with the CUDA feature. They do not launch `mla_decode` PTX and do not measure full-history traffic on a real sequence. GB10 kernel execution and the #1167 transfer-measurement gate are still open.

tbraun96:
Certified landed via #1202 (`integration/k3-six`, squash-merged to `main` at `dde2381c7b6512a249a1dd17b2d1e7bdcacdd998`, 2026-09-21).

Verification: this PR's head `84677b97fc` (the stack tip) was merged into the integration branch as commit `3e6223f7d8` (real 2-parent merge, second parent = this PR's actual head sha). The integration branch's final tree is byte-identical to current `main` for every K3-related path (only unrelated `site/` dashboard files have since diverged). Concrete artifact: `crates/spark-model/src/kimi_k3/host_decode_mla.rs` (added by this PR, `fn mla_decode_keep_state`) is present on `main` with an identical git blob hash to this PR's head. Note the integration branch also carries one commit this PR's head does not — `de1dc2abbe`/re-applied as `2a822a73b1`, 'the K3 MoE launch count is two since #1186 batched gate/up' — which is a fix layered on top of the composed stack, not something this PR was expected to contain.

Closing as landed.

PR #1199, opened 2026-09-20
server: AVAROK_DISABLE_WATCHDOGS also disarms the token loop watchdog

### LINKED ISSUE #1135: [bug]  check_loop_watchdog is not disabled by ATLAS_DISABLE_WATCHDOGS=1
### Atlas version / docker tag

git HEAD

### Model

Qwen/Qwen3.6-35B-A3B-FP8 

### Hardware

1x GB10

### Launch command

```bash
/home/daniel/Ai/Atlas/atlas/target/release/spark serve Qwen/Qwen3.6-35B-A3B-FP8 \
    --host 0.0.0.0 \
    --port 8000 \
    --max-seq-len 65536 \
    --kv-high-precision-layers auto \
    --gpu-memory-utilization 0.90 \
    --scheduling-policy slai \
    --enable-prefix-caching \
	--tool-max-tokens 32768 \
	--max-thinking-budget 32768  \
	--tool-call-parser qwen3_coder  \
	--content-loop-watchdog false \
	--content-loop-min-repeats 100  \
	--speculative \
    --num-drafts 2
```

### Expected behaviour

Expected behavior : 
When export ATLAS_DISABLE_WATCHDOGS=1 
This watchdog should NOT be called (function check_loop_watchdog)

### Actual behaviour

Current behavior : 
set variable : export ATLAS_DISABLE_WATCHDOGS=1 

Log
WARN spark loop watchdog fired – repeated line (fuzzy-match) in post-detector content

Process is stopped by Atlas.
I have 5x this text inside my context "#### Résultats"

### Additional context

_No response_

aniruddh-krovvidi:
## Summary

`AVAROK_DISABLE_WATCHDOGS=1` is documented (and logged at boot) as disabling all auto-watchdogs, but the API-layer token loop watchdog in `chat_stream/handle_token.rs` was never gated by it, so streams still ended with `loop watchdog fired — repeated line`. This gates that call on the same parsed flag, cached once like the neighbouring `AVAROK_SIMHASH_LOOP` switch. (The issue sets `ATLAS_DISABLE_WATCHDOGS`; the variable the code reads is `AVAROK_DISABLE_WATCHDOGS`, and even that did not reach this watchdog.)

Closes #1135

## Test plan

- [x] `AVAROK_SKIP_BUILD=1 CUDARC_CUDA_VERSION=13000 cargo clippy -p spark-server --tests` (Linux, no CUDA)
- [x] `bash scripts/check-license-headers.sh` — no new file; the touched file keeps its header
- [ ] Tested against a real model / hardware — PLACEHOLDER: no GB10 available; repro is the issue's log line with the env var set
- [x] `cargo test -p spark-server -- watchdog stream_guards disable_watchdogs` — 20 passed (existing tests; the parse is already covered by `parse_disable_watchdogs` tests, and the crate's rule is no `set_var` in tests, so the gate has no new env-driven test)

## Notes for reviewers

Off the decode path: one cached `bool` check per streamed token, in the API layer. `--content-loop-watchdog false` is a separate (scheduler-side) switch and is untouched, as is the SimHash guard's own `AVAROK_SIMHASH_LOOP=0` switch; this PR only wires the watchdog the issue names.

## CLA


aniruddh-krovvidi:

tbraun96:
Composed into #1203 with #1200 so the two share one certification campaign —
same family (an env lever that does not do what it says), and the GPU cost is
identical whether the campaign carries one PR or two. Your commit is merged by
**head** (`9925d6fe14`), and containment is asserted, not assumed.

The fix itself is right, and notably right in the way that matters here: it goes
through the existing `scheduler::parse_disable_watchdogs` instead of re-spelling
the rule, so it does not add a second implementation of the flag.

One thing to flag, and it is not a criticism of this change — a sweep of the
serving path found that **three other watchdogs still fire** with
`AVAROK_DISABLE_WATCHDOGS=1` set: the think-skip hard-stop
(`decode_logits_step.rs:413-421` + `emit_step.rs:111-116`), the fuzzy-repetition
watchdog (`decode_logits_step.rs:954` → `:979-982`, gated only on the separate
`loop_watchdog` atomic), and the SimHash semantic-loop watchdog
(`handle_token.rs:665` → `:689-706`, gated only on `AVAROK_SIMHASH_LOOP`).

So this covers one of four sites. I have filed the other three, plus a loose
`AVAROK_SIMHASH_LOOP` parse in this same file and the mirrored parser in
`spark-model`, as #1204 with citations — they want their own PR rather than
being bolted onto this campaign. **#1135 is being left open** on that basis
rather than auto-closed on merge; nothing for you to change here.

tbraun96:

Closing by hand: the merge queue **squashes**, so this PR's head is not an ancestor of main even though its content is, and GitHub will not auto-close it. Verified by artifact rather than assumed — `crates/spark-server/src/api/chat_stream/handle_token.rs` on main defines `fn watchdogs_disabled()` at line 29 and gates the token-loop watchdog on it at line 693 (`let token_trip = !watchdogs_disabled() && …`).

PR #1200, opened 2026-09-21
fix: stop handing a partial tail KV block to a second writer

### LINKED ISSUE #1193: Prefix cache shares a partially-filled tail KV block with no copy-on-write (default ON)
## Summary

On `main`, the sub-block prefix-cache path publishes a **partially-filled tail KV block** into the radix tree at the end of prefill, hands it out **by physical index** to a second sequence, and never copies it. The donor can still write into that block afterwards. Two live sequences then write different K/V into the same physical slots.

The lever that enables this, `AVAROK_PREFIX_SUBBLOCK`, **defaults ON** (`inner.rs:200-206`: `!= Ok("0")`).

## Verified chain (all `main` @ `9e597531cf`)

1. **Sub-block matching is on by default and ends mid-block** — `crates/spark-runtime/src/radix_tree/inner.rs:200-206`, gated on `remainder > 0 && remainder < block_size && matched_tokens == num_full_blocks * block_size`.
2. **The donor's live block is handed out, by index, with the refcount check explicitly skipped** — `inner.rs:224-235`:
   ```rust
   // ref_count check not applicable to partial suffix (it's metadata, not a node).
   matched_blocks.push(partial_block);
   ```
3. **The requester adopts it into its own block table** — `crates/spark-model/src/model/trait_impl/prefill_a.rs:87-98`: `inc_ref(block_idx)` then `seq.block_table.push(block_idx)`; `kv_write_start = matched_tokens`.
4. **No fresh block is allocated for it** — `crates/spark-model/src/model/block_mgmt.rs:437-444`:
   ```rust
   let in_window = bt_len > 0 && abs_block_idx < ws + bt_len;
   if in_window { return Ok(()); }
   ```
   When the match ends inside the tail block, that block is already `block_table[(n-1)/bs]`, so `ensure_blocks_through_prefill` returns early.
5. **The write floor protects only positions *below* the match point** — `prefill_b/forward_layers.rs:138-152`, whose own comment reads: *"tokens at or past it are new and are written normally."*
6. **There is no copy-on-write anywhere in the KV path** — no `ref_count > 1` check before a KV write in `spark-runtime` or `spark-model`; `inc_ref` (`kv_cache/paged_impl.rs:168-180`) is a bare counter bump.

Items 1, 2, 4 and 6 I confirmed directly; the full trace including the publish site is below.

## Why the donor is still live

The partial block is published at **end of prefill**, not at sequence finish — `prefill_b/finalize_last.rs:344-351, 431-440, 447-454, 458-465` and `prefill_d.rs:74-82` all insert the **full** prompt with the **full** block table, and `finalize_last.rs:273-275` says this is deliberate ("preserving the partial-suffix sub-block TTFT optimization"). `inner.rs:405-427` stores `block_table[num_full_blocks]` as `partial_suffix`. There is no sealed/frozen/finished condition in `radix_tree/`. So the donor publishes the block and *then* decodes into it.

## Interleaving (block_size 16)

Donor D has a prompt of `n` tokens, `r = n % 16 != 0`; tail block B holds offsets `[0, r)`.

1. D finishes prefill → B becomes `partial_suffix`.
2. Requester R arrives with the same prompt (or a strict prefix landing in B) → matches via the `partial_suffix` arm, pushes B, refcount 2, **no new block**.
3. D decodes position `n` → writes `B*16 + r`.
4. R decodes → writes `B*16 + r`. **Same slot.** In batched decode (`impl_b1.rs:72-82`) both are in one kernel launch. Offsets `r+1..15` alias on subsequent steps.

For the strict-prefix variant, R's generated token overwrites D's **prompt** K/V while D is still live.

The "arm 1" path (`inner.rs:210-223`) is the same defect against a *full* block: R writes generated tokens into offsets `[r, 16)` of a block other sequences treat as valid prompt K/V.

## Trigger conditions

Concurrent in-flight prompts that are identical, or prefixes of each other, up to the last full 16-token block. There is no scheduler dedup of identical prompts.

## Immediate mitigation

`AVAROK_PREFIX_SUBBLOCK=0` disables both sub-block arms.

## Fix directions (not applied)

- Do not publish the partial tail block at prefill time — insert only `tokens[..num_full*bs]` with `block_table[..num_full]` from the prefill-side inserts, leaving the partial insert to finish-time `cache_sequence` (`lifecycle.rs:224`), **and**
- On a sub-block match, copy the shared block's `[0, r)` into a freshly allocated block before extending `block_table` — even two *finished*-donor requesters alias at offsets >= r.

This is in `crates/**`, so any fix is a PERF_PATHS change and needs a certification campaign.

## Relation to the known C=8/32/128 early-stop

The mechanism is **consistent** with completions stopping early at concurrency, but whether that workload actually contains qualifying prompt pairs is not established from source. **No causal link is claimed here** — it is listed only so the two are not investigated as if unrelated.

## Context

Found while adjudicating the default-value conflict in #692, which proposes the opposite default (`ATLAS_PREFIX_PARTIAL_TAIL`, OFF) and independently argued this path is unsound.

tbraun96:
Fixes #1193. Supersedes #692, whose diagnosis and measured repro this carries.

## The defect

The sub-block prefix-cache path publishes a **partially-filled tail KV block** into the radix tree at end of prefill, hands it to a second sequence **by physical index**, and never copies it. The donor is still decoding into that block. Two live sequences then write different K/V into the same physical slots.

- `radix_tree/inner.rs` — `AVAROK_PREFIX_SUBBLOCK` defaulted **ON**, gating sub-block matching
- the walk hands out the donor's live block with the comment *"ref_count check not applicable to partial suffix"*
- `prefill_a.rs` — `inc_ref` then `block_table.push`; `kv_write_start = matched_tokens`
- `block_mgmt.rs` — `if in_window { return Ok(()); }`, so **no fresh block is allocated**
- `forward_layers.rs` — the write floor guards only positions *below* the match point; its own comment says "tokens at or past it are new and are written normally"
- there is **no copy-on-write anywhere** in the paged KV manager

#692's measured consequence, which is stronger evidence than any static read:

> with identical prompts in flight the writers stay byte-identical in lockstep and the tear is BENIGN — which is why serial decode looked clean — but any desync … makes the writes differ … **C=4 went 0/4 with three sequences emitting IDENTICAL wrong text**.

That benign-in-lockstep property is why this hid for so long.

## The fix: flip the existing lever, don't add a second

`AVAROK_PREFIX_SUBBLOCK` now defaults **OFF**. #692 proposed a new `ATLAS_PREFIX_PARTIAL_TAIL` over the *same* `if` — two independent readers on one branch, the one-flag-two-implementations defect this repo has hit before. **The two branches must not land together.**

**One flip covers both arms.** Arm 1 (child-key prefix, against a *full* donor block) and arm 2 (`partial_suffix`) are sibling branches *inside* a single predicate. Arm 1 is unsound too — the requester's tail writes clobber committed rows other sequences map read-only — and there is now a regression test pinning each arm rather than a claim that one flip suffices.

Also fixed: the lever parsed as `!= "0"`, so **`AVAROK_PREFIX_SUBBLOCK=true` silently read as ON**. Parsing is now strict — unset is OFF, only an explicit `1` arms it, anything else is refused loudly.

## Correction to #692: the refcount claim does not hold

Its premise is right — `inc_refs` computes `num_blocks = num_matched / block_size`, so the partial block takes no *node* ref. The conclusion doesn't follow: the **pool** ref isn't taken by `inc_refs` at all. It is taken by the adopting caller over every entry of `matched_blocks` including the partial one (`prefill_b/prefix_lookup.rs:138-141`, `prefill_a.rs:95-98`, `prefill_c.rs:175`), and freed symmetrically in `sequence.rs`. The node side is symmetric too. The cache holds its own pool ref via `cache_acquires_refs`, in place since `7444209045` (2026-08-08) — predating #692, so this is not main-moving-underneath. **Nothing was changed here.**

## Tests

Two defect tests (one per arm) plus two controls. Seven legacy sub-block tests now **opt in explicitly** — including two that would otherwise have gone **vacuously green** under the new default while still appearing to test the refusal they were written for.

**Negative control:** reverting the default made both defect tests fail **by assertion** — `matched_tokens` 20 vs 16 (arm 2) and 22 vs 16 (arm 1) — not by compile error, while the control tests stayed green, which is what proves the defect tests aren't passing on a walk that matches nothing.

Green: spark-runtime 334/334, spark-model 1349, spark-server 2635; `cargo check --workspace --all-targets`, fmt, clippy clean.

## Cost, and what is not attempted

Cost of OFF: up to `block_size-1` tokens of prefill recompute per hit, and non-block-aligned exact-repeat prompts no longer reach the Marconi exact-leaf shortcut. `docs/ROBUSTNESS.md` records a measured arm with this lever off, verified armed in `/proc/PID/environ`: **byte-identical to baseline, 0 of 251 samples moved**. That bounds the output change at zero; it does not bound the TTFT cost — that is this campaign's job.

The insert side is deliberately untouched, so `=1` restores the old behaviour completely and the #936 arm-ladder A/B still means something.

Follow-up, documented at the match site and not attempted here: a copy-on-write tail — fresh block plus per-layer D2D of the matched rows, event-ordered against the donor's stream — to win those tokens back.

`crates/**`, so this needs its own certification campaign.


tbraun96:

Closing by hand: the merge queue **squashes**, so this PR's head is not an ancestor of main even though its content is, and GitHub will not auto-close it. Verified by artifact rather than assumed — `crates/spark-runtime/src/radix_tree/partial_tail.rs` on main carries the strict parse verbatim — `None => false, Some("1") => true, Some("0") => false` at lines 74–76 — and `tests/partial_tail.rs` carries all 5 of its functions.

PR #496, opened 2026-08-13
spark-model: batched argmax in graphed K=2 verify

facepunch47:
## Summary

Production `--speculative --num-drafts 1` is always K=2 verify via `verify_b`. That path launched two serial `argmax_bf16` scans (one CTA each, ~100 µs at this vocab) inside the CUDA graph. Batched verify (`verify_e`) already uses `argmax_bf16_batch` — one launch, one block per row, identical per-row body. This PR wires that kernel into the K=2 graph.

Kill switch: `ATLAS_NO_ARGMAX_BATCH=1` (`=0` does **not** disable). Same name the decode batch path already uses.

## Bottleneck (evidence, not folklore)

`crates/spark-model/src/model/trait_impl/meta.rs` already documents the measurement:

> The single-row `argmax_bf16` is a one-CTA reduction (grid [1,1,1]), so n calls on the same stream serialise n single-SM scans: measured 16 x 100.6 us = 1.6 ms per decode step at n=16.

At production K=2 that is ~200 µs serial vs ~100 µs batched. Bandwidth for two vocab rows is ~1 MB — this is SM occupancy, not DRAM. Inside the graph the launches are captured; the GPU work still serialises.

Live `:8888` on `nvidia/Qwen3.6-35B-A3B-NVFP4` (`atlas-gb10:pr-proof-321128d`): `fwd ≈ 12.0–12.1 ms`. This slice is ~1% of `fwd`. 220-token tok/s may land inside noise; the log invariant is one launch vs two.

## What changed

- `verify_b` dispatches `argmax_bf16_batch` when the kernel resolved.
- `ops::argmax_batch_from` / `argmax_batch_enabled` are the SSOT for the kill-switch polarity.
- One-shot serve log: `K=2 verify argmax_bf16_batch ENGAGED: one launch for both rows ...`

## Gate bake (GB10, 2026-08-14) — 6/6

Frozen binary `/tmp/spark-d86e667f0f-{35b,27b}` at code SHA `d86e667`. Records on this PR:

| Gate | Result |
|---|---|
| ssm-state-poisoning-gate | PASS |
| ttft-warm-gate | PASS median −0.1% |
| ttft-cold-gate | PASS median +0.1% |
| bfcl-subset n=995 | PASS **87.64 / 88.94** |
| agentic-webserver | PASS 10/10 · Σwall **632s** |
| bfcl-subset-echolp n=1004 | attempt 1 84.66 / 85.25 FAIL · attempt 2 PASS **84.76 / 85.34** |

[...]

PR #1205, opened 2026-09-21
server: honor min_tokens on legacy completions (#582 re-cut onto main)

tbraun96:
## What this is

@zhachar14h-cell's #582 re-cut onto current `main` (`5f639bfc57`). Their two
commits replayed with `cherry-pick -x`, nothing else added.

`min_tokens` was accepted on the legacy `/v1/completions` path and then ignored,
so a request asking for at least N tokens could return fewer. The fix routes it
through the scheduler's logit processors like every other stop-suppressing
lever, adding `scheduler/logit_processors/min_tokens_eos.rs`.

11 files, +272/−37.

## The one conflict, and how it was resolved

`crates/spark-server/src/api/completions_exec.rs` — an **in-subject** clash, not
drift. Both sides append to the same region at the end of the file:

- `main` has `finish_completion_text(...)` plus
  `#[cfg(test)] #[path = "completions_exec_tests.rs"] mod tests;`
- #582 adds `#[cfg(test)] impl CompletionParams { fn test() -> Self }`

Resolved as a **union** — the two are independent additions. Before taking it I
checked the thing that would actually break: the test constructor sets **18
fields, and `CompletionParams` on main has exactly those 18** (`diff` of the two
sorted field lists is empty). A re-cut that silently disagreed with the struct
would not compile, so this is the check worth doing rather than eyeballing it.

## Re-cut audit

A re-cut can silently drop functions the contributor added while adapting to
main. Compared function-presence against `main` for all 11 touched `.rs` files:
**nothing lost**.

## Why a re-cut rather than a rebase request

#582 was only 18 commits behind and its single conflict was in the file the fix
is about, so there was nothing for the author to disambiguate — a replay
answers it. The other five PRs triaged alongside it were **not** like this and
have been handled differently (#560, #486, #493, #496 sent back for a rebase;
#404 closed as superseded).

This PR is the deliverable; **#582 closes when it lands.** It touches `crates/`,
so it needs a certification campaign, and it is deliberately opened with room
for riders — if #560/#486/#493/#496 rebase before campaign 7 is launched, they
compose onto this branch and share the run.


tbraun96:
Folded into **#1203** (campaign 5) rather than given a campaign of its own — it merges clean onto that branch, it is the same family (a lever parsed loosely, a watchdog the flag never reached, a parameter accepted and dropped), and a campaign costs the same three GB10s for the same two hours whether it carries two PRs or three. That is one whole campaign saved.

Your head `79f7c852e6` is merged by sha and containment is asserted. This PR stays open as the record of the re-cut and closes when #1203 lands; nothing further needed here.

tbraun96:

Closing by hand: the merge queue **squashes**, so this PR's head is not an ancestor of main even though its content is, and GitHub will not auto-close it. Verified by artifact rather than assumed — `crates/spark-server/src/scheduler/logit_processors/min_tokens_eos.rs` is present on main — the file this change adds.

